PR #4, opened 2018-08-02
Machine Config Proposal

csrwng:
Instead of repeating the proposed types in a doc, this proposal includes code changes to represent the initial proposed type definitions.

The goal is to come to a basic consensus on an initial implementation and generate some discussion around the open issues. Like anything else, this is expected to evolve as we evolve the implementation.

### DIFF
--- a/cloud/aws/providerconfig/types.go
+++ b/cloud/aws/providerconfig/types.go
@@ -14,25 +14,140 @@
 package providerconfig
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// AWSMachineProviderConfig is the type that will be embedded in a Machine.Spec.ProviderConfig field
+// for an AWS instance. It is used by the AWS machine actuator to create a single machine instance,
+// using the RunInstances call (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html)
+// Required parameters such as region that are not specified by this configuration, will be defaulted
+// by the actuator.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSMachineProviderConfig struct {
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta
+
+	// AMI is the reference to the AMI from which to create the machine instance.
+	AMI AWSResourceReference
+
+	// InstanceType is the type of instance to create. Example: m4.xlarge
+	InstanceType string
+
+	// AdditionalTags is the set of tags to add to an instance, in addition to the ones
+	// added by default by the actuator. These tags are additive. The actuator will ensure
+	// these tags are present, but will not remove any other tags that may exist on the
+	// instance.
+	// +optional
+	AdditionalTags map[string]string
+
+	// IAMInstanceProfile is a reference to an IAM role to assign to the instance
+	// +optional
+	IAMInstanceProfile *AWSResourceReference
+
+	// PublicIP specifies whether the instance should get a public IP.
+	// Precedence for this setting is as follows:
+	// 1. This field if set
+	// 2. Cluster/flavor setting
+	// 3. Subnet default
+	// +optional
+	PublicIP *bool
+
+	// AdditionalSecurityGroups is an array of references to security groups that should be applied to the
+	// instance. These security groups would be set in addition to any security groups defined
+	// at the cluster level or in the actuator.
+	// +optional
+	AdditionalSecurityGroups []AWSResourceReference
+
+	// Subnet is a reference to the subnet to use for this instance. If not specified,
+	// the cluster subnet will be used.
+	// +optional
+	Subnet *AWSResourceReference
+}
+
+// AWSResourceReference is a reference to a specific AWS resource by ID, ARN, or filters.
+// Only one of ID, ARN or Filters may be specified. Specifying more than one will result in
+// a validation error.
+type AWSResourceReference struct {
+	// ID of resource
+	// +optional
+	ID *string
+
+	// ARN of resource
+	// +optional
+	ARN *string
+
+	// Filters is a set of key/value pairs used to identify a resource
+	// They are applied according to the rules defined by the AWS API:
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html
+	// +optional
+	Filters []Filter
+}
+
+// Filter is a filter used to identify an AWS resource
+type Filter struct {
+	// Name of the filter. Filter names are case-sensitive.
+	Name string
+
+	// Values includes one or more filter values. Filter values are case-sensitive.
+	Values []string
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSClusterProviderConfig struct {
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta
 }
 
+// AWSMachineProviderStatus is the type that will be embedded in a Machine.Status.ProviderStatus field.
+// It containsk AWS-specific status information.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSMachineProviderStatus struct {
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta
+
+	// InstanceID is the instance ID of the machine created in AWS
+	// +optional
+	InstanceID *string
+
+	// InstanceState is the state of the AWS instance for this machine
+	// +optional
+	InstanceState *string
+
+	// Conditions is a set of conditions associated with the Machine to indicate
+	// errors or other status
+	// +optional
+	Conditions []AWSMachineProviderCondition
+}
+
+// AWSMachineProviderConditionType is a valid value for AWSMachineProviderCondition.Type
+type AWSMachineProviderConditionType string
+
+// Valid conditions for an AWS machine instance
+const (
+	// MachineCreated indicates whether the machine has been created or not. If not,
+	// it should include a reason and message for the failure.
+	MachineCreated AWSMachineProviderConditionType = "MachineCreated"
+)
+
+// AWSMachineProviderCondition is a condition in a AWSMachineProviderStatus
+type AWSMachineProviderCondition struct {
+	// Type is the type of the condition.
+	Type AWSMachineProviderConditionType
+	// Status is the status of the condition.
+	Status corev1.ConditionStatus
+	// LastProbeTime is the last time we probed the condition.
+	// +optional
+	LastProbeTime metav1.Time
+	// LastTransitionTime is the last time the condition transitioned from one status to another.
+	// +optional
+	LastTransitionTime metav1.Time
+	// Reason is a unique, one-word, CamelCase reason for the condition's last transition.
+	// +optional
+	Reason string
+	// Message is a human-readable message indicating details about last transition.
+	// +optional
+	Message string
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSClusterProviderStatus struct {
-	metav1.TypeMeta `json:",inline"`
+	metav1.TypeMeta
 }
--- a/docs/proposal/machine-config.md
+++ b/docs/proposal/machine-config.md
@@ -1,0 +1,61 @@
+# AWS Provider Machine Configuration and Status
+
+## Overview
+This proposal aims to provide an initial API for AWS machine configuration and status for the AWS machine actuator.
+Cluster configuration for the AWS provider is out of scope for this proposal.
+
+## AWSMachineProviderConfig
+Defines the spec portion of a Machine specific to AWS. It is meant to be embedded in the
+Machine.Spec.ProviderConfig field of a cluster API Machine.
+
+Following are design decisions around this type:
+
+- No AWS SDK types are included. For the following reasons:
+  - Clients would also need to have the AWS SDK vendored.
+  - The type would be tied to a specific AWS client version.
+
+- The machine ProviderConfig includes all the configuration needed to create a Machine.
+  Until we resolve issues related to linking a cluster and machine, it is not clear
+  whether the machine controller should be required to pass a Cluster resource to the actuator:
+  - https://github.com/kubernetes-sigs/cluster-api/issues/41
+  - https://github.com/kubernetes-sigs/cluster-api/issues/145
+  - https://github.com/kubernetes-sigs/cluster-api/issues/177
+
+- Secrets are referenced for UserData and Account credentials, but are optional.
+  If the secrets are not present, then no user data is set on the instance, and if no
+  account credentials are present, it is assumed the IAMRole of the machine where the
+  actuator is running has enough permission to create a new instance.
+
+- The list of ELBs that should be associated with a machine are set explicitly in the config
+  and are attached when the instance is created.
+
+## AWSMachineProviderStatus
+Defines the status portion of a Machine specific to AWS. It is meant to be embedded in the
+Machine.Status.ProviderStatus field of a cluster API Machine.
+
+Design decisions for this type:
+
+- Conditions are used to report on problems creating an instance, or attaching ELBs.
+  Reference: https://github.com/kubernetes/kubernetes/issues/7856
+  Although the issue seems to imply that we want to get rid of conditions, Brian Grant's
+  [conclusion](https://github.com/kubernetes/kubernetes/issues/7856#issuecomment-335687733)
+  states that conditions should be kept to report extended status. Knowing why we failed
+  to create an instance or whether attaching an ELB did not succeed is still very valuable
+  information.
+
+## Open Issues
+
+### Load Balancer Management
+- Currently we include a set of IDs for ELBs to attach to the machine on creation. However, what happens if the ELB
+  doesn't exist, or if after the machine is created the ELB fails to attach? Should this be handled by a separate
+  controller?
+
+### BlockDevice Mapping
+- The RunInstances EC2 call allows specifying a BlockDeviceMapping to allow changing volume size or type from what
+  is specified in the AMI. Is this something that we need? or is it enough to say that if you want something different
+  to simply create a different AMI?
+
+### Extensibility
+- If the current types don't meet all the needs for different providers, do we allow some kind of extension mechanism
+  to the types? How do we extend the actuator itself? Or should we simply allow the base implementation to be vendored
+  into an implementation that can augment it. What hooks would we provide to allow this?
--- a/cloud/aws/providerconfig/v1alpha1/types.go
+++ b/cloud/aws/providerconfig/v1alpha1/types.go
@@ -14,12 +14,82 @@
 package v1alpha1
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// AWSMachineProviderConfig is the type that will be embedded in a Machine.Spec.ProviderConfig field
+// for an AWS instance. It is used by the AWS machine actuator to create a single machine instance,
+// using the RunInstances call (https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html)
+// Required parameters such as region that are not specified by this configuration, will be defaulted
+// by the actuator.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSMachineProviderConfig struct {
 	metav1.TypeMeta `json:",inline"`
+
+	// AMI is the reference to the AMI from which to create the machine instance.
+	AMI AWSResourceReference `json:"ami"`
+
+	// InstanceType is the type of instance to create. Example: m4.xlarge
+	InstanceType string `json:"instanceType"`
+
+	// AdditionalTags is the set of tags to add to an instance, in addition to the ones
+	// added by default by the actuator. These tags are additive. The actuator will ensure
+	// these tags are present, but will not remove any other tags that may exist on the
+	// instance.
+	// +optional
+	AdditionalTags map[string]string `json:"additionalTags,omitempty"`
+
+	// IAMInstanceProfile is a reference to an IAM role to assign to the instance
+	// +optional
+	IAMInstanceProfile *AWSResourceReference `json:"iamInstanceProfile,omitempty"`
+
+	// PublicIP specifies whether the instance should get a public IP.
+	// Precedence for this setting is as follows:
+	// 1. This field if set
+	// 2. Cluster/flavor setting
+	// 3. Subnet default
+	// +optional
+	PublicIP *bool `json:"publicIP,omitempty"`
+
+	// AdditionalSecurityGroups is an array of references to security groups that should be applied to the
+	// instance. These security groups would be set in addition to any security groups defined
+	// at the cluster level or in the actuator.
+	// +optional
+	AdditionalSecurityGroups []AWSResourceReference `json:"additionalSecurityGroups,omitempty"`
+
+	// Subnet is a reference to the subnet to use for this instance. If not specified,
+	// the cluster subnet will be used.
+	// +optional
+	Subnet *AWSResourceReference `json:"subnet,omitempty"`
+}
+
+// AWSResourceReference is a reference to a specific AWS resource by ID, ARN, or filters.
+// Only one of ID, ARN or Filters may be specified. Specifying more than one will result in
+// a validation error.
+type AWSResourceReference struct {
+	// ID of resource
+	// +optional
+	ID *string `json:"id,omitempty"`
+
+	// ARN of resource
+	// +optional
+	ARN *string `json:"arn,omitempty"`
+
+	// Filters is a set of key/value pairs used to identify a resource
+	// They are applied according to the rules defined by the AWS API:
+	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html
+	// +optional
+	Filters []Filter `json:"filters"`
+}
+
+// Filter is a filter used to identify an AWS resource
+type Filter struct {
+	// Name of the filter. Filter names are case-sensitive.
+	Name string `json:"name"`
+
+	// Values includes one or more filter values. Filter values are case-sensitive.
+	Values []string `json:"values"`
 }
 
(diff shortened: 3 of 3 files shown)


docs/proposal/machine-config.md
```diff
@@ -0,0 +1,65 @@
...
+
+## AWSMachineProviderConfig
+Defines the spec portion of a Machine specific to AWS. It is meant to be embedded in the
+Machine.Spec.ProviderConfig field of a cluster API Machine.
+
+Following are design decisions around this type:
+
+- No AWS SDK types are included. For the following reasons:
+  - Clients would also need to have the AWS SDK vendored.
+  - The type would be tied to a specific AWS client version.
```
mvladev: This 100% correct and valid, but we'll provide only subset of configurations, which you can do, when launching a machine.

For example in our project we decided to more or less copy the structs which were used in aws-client-go. https://github.com/gardener/machine-controller-manager/blob/5976512b001db0b029a8982dc93e1065f099e4a7/pkg/apis/machine/v1alpha1/types.go#L763-L885 . I'm not 100% sure that our approach is 100% valid as well.

docs/proposal/machine-config.md
```diff
@@ -0,0 +1,65 @@
+# AWS Provider Machine Configuration and Status
+
+## Overview
+This proposal aims to provide an initial API for AWS machine configuration and status for the AWS machine actuator.
+Cluster configuration for the AWS provider is out of scope for this proposal.
+
+## AWSMachineProviderConfig
+Defines the spec portion of a Machine specific to AWS. It is meant to be embedded in the
```
mvladev: Are we focusing this API only for on-demand AWS instances? If we want to have spot instances support then we might want to include additional fields.

cloud/aws/providerconfig/types.go
```diff
@@ -17,19 +17,112 @@ import (
...
+	// these tags are present, but will not remove any other tags that may exist on the
+	// instance.
+	Tags map[string]string
+
+	// IAMInstanceProfile is the name of the IAM role to assign to the new instance
+	IAMInstanceProfile string
+
+	// UserDataSecret contains a local reference to a secret that contains the
+	// UserData to apply to the instance
+	UserDataSecret *corev1.LocalObjectReference
```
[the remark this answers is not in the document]
csrwng: a LocalObjectReference only contains a string (Name). But it is the more semantically correct type given that we are referencing another object in storage. See https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#references-to-related-objects
You can find a few examples in the core types like this one:
https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L1256

cloud/aws/providerconfig/types.go
```diff
@@ -17,19 +17,112 @@ import (
...
+	// SubnetID is the id of the subnet to use for this instance
+	SubnetID string
+
+	// AvailabilityZone is the availability zone where the instance should be
+	// placed. Exammple: us-east-1c
+	AvailabilityZone string
+
+	// LoadBalancerIDs is the IDs of the load balancers to which the new instance
+	// should be added once it's created.
+	LoadBalancerIDs []string
```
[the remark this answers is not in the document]
csrwng: Excellent question. I'm not sure that it's doable for machines that are part of the control plane, where we run into a chicken and egg problem trying to communicate with the control plane. However, I'm starting to think that maybe that's the responsibility of the cluster actuator and not the machine actuator. For everything else,  I think the LoadBalancer type of service should work. Wdyt?

docs/proposal/machine-config.md
csrwng: I would say that the we could initially focus on on-demand (because it's simpler), but we could eventually evolve to support spot instances.

docs/proposal/machine-config.md
csrwng: Can we adopt the approach of including as much of the API as we need for what we initially support and as the actuator evolves add more if we need to?

csrwng:
Thank you @mvladev for the review and great suggestions!

csrwng:
@dgoodwin @abutcher @joelddiaz @twiest

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,151 @@
...
+	Value string
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSClusterProviderConfig struct {
 	metav1.TypeMeta `json:",inline"`
 }
 
+// AWSMachineProviderStatus is the type that will be embedded in a Machine.Status.ProviderStatus field.
+// It containsk AWS-specific status information.
```
dgoodwin: -k.

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,151 @@
...
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// AWSMachineProviderConfig is the type that will be embedded in a Machine.Spec.ProviderConfig field
+// for an AWS instance.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSMachineProviderConfig struct {
```
enxebre: do we want to include an optional KeyName for ssh?

csrwng: definitely, thx!

csrwng: PR updated

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,154 @@
...
+
+	// SecurityGroups is an array of references to security groups that should be applied to the
+	// instance.
+	SecurityGroups []AWSResourceReference
+
+	// Subnet is a reference to the subnet to use for this instance
+	Subnet AWSResourceReference
+
+	// Placement indicates where to create the instance in AWS
+	Placement Placement
```
ingvagabund: I like this.

docs/proposal/machine-config.md
frobware: I think this is really the best and most pragmatic choice.

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,154 @@
...
+	PublicIP *bool
+
+	// SecurityGroups is an array of references to security groups that should be applied to the
+	// instance.
+	SecurityGroups []AWSResourceReference
+
+	// Subnet is a reference to the subnet to use for this instance
+	Subnet AWSResourceReference
+
+	// Placement indicates where to create the instance in AWS
```
frobware: It's more than just "indication", no? How about `specifies where...`?

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,154 @@
...
+	Placement Placement
+
+	// LoadBalancerIDs is the IDs of the load balancers to which the new instance
+	// should be added once it's created.
```
frobware: nit: "it is created"

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,154 @@
...
+	LoadBalancerIDs []string
+}
+
+// AWSResourceReference is a reference to a specific AWS resource by ID, ARN, or filters
```
frobware: How do you discriminate between the various choices? Which (of the optional) take precedence?

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,154 @@
...
+	// ARN of resource
+	ARN *string
+
+	// Filters is a set of filters used to identify a resource
+	Filters []Filter
+}
+
+// Placement indicates where to create the instance in AWS
+type Placement struct {
+	// Region is the region to use to create the insntance
```
frobware: typo: "create the instance".

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,154 @@
...
+// AWSMachineProviderCondition is a condition in a AWSMachineProviderStatus
+type AWSMachineProviderCondition struct {
+	// Type is the type of the condition.
+	Type AWSMachineProviderConditionType
+	// Status is the status of the condition.
+	Status corev1.ConditionStatus
+	// LastProbeTime is the last time we probed the condition.
+	// +optional
+	LastProbeTime metav1.Time
+	// LastTransitionTime is the last time the condition transitioned from one status to another.
```
frobware: Is this "from one _state_ to another"?

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,154 @@
...
+	// AvailabilityZone is the availability zone of the instance
+	AvailabilityZone string
+}
+
+// Filter is a filter used to identify an AWS resource
+type Filter struct {
+	// Name of the filter
+	Name string
+
+	// Values includes one or more filter values
```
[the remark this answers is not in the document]
csrwng: It's coming from the EC2 AWS API which says the same thing. I don't believe that specifying a filter name without a value is a valid selector. So it should at least have one value.

cloud/aws/providerconfig/types.go
csrwng: It's from one status to the other ... referring to the Status field above... when it transitions from corev1.StatusTrue to corev1.StatusFalse or viceversa.

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+	// CredentialsSecret is a reference to the secret with AWS credentials. Otherwise, defaults to permissions
+	// provided by attached IAM role where the actuator is running.
+	CredentialsSecret *corev1.LocalObjectReference
+
+	// KeyName is the name of the KeyPair to use for SSH
+	KeyName *string
+
+	// DeviceIndex is the index of the device on the instance for the network interface attachment.
+	// Defaults to 0.
+	DeviceIndex int
```
mvladev: this should be `*int` (optional)

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+	// PublicIP specifies whether the instance should get a public IP. If not present,
+	// it should use the default of its subnet.
+	PublicIP *bool
+
+	// SecurityGroups is an array of references to security groups that should be applied to the
+	// instance.
+	SecurityGroups []AWSResourceReference
+
+	// Subnet is a reference to the subnet to use for this instance
+	Subnet AWSResourceReference
```
errordeveloper: For this to map to an ASG, then you would want a slice here.

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// AWSMachineProviderConfig is the type that will be embedded in a Machine.Spec.ProviderConfig field
+// for an AWS instance.
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSMachineProviderConfig struct {
```
errordeveloper: Is this supposed to represent a single machine? What about ASGs?

mvladev: @errordeveloper single machine - we won't use ASG

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+// a validation error.
+type AWSResourceReference struct {
+	// ID of resource
+	ID *string
+
+	// ARN of resource
+	ARN *string
+
+	// Filters is a set of filters used to identify a resource
+	Filters []Filter
```
errordeveloper: Is this mostly used for requests inputs?

csrwng: Yes

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+	// Name of the tag
+	Name string
+
+	// Value of the tag
+	Value string
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 type AWSClusterProviderConfig struct {
 	metav1.TypeMeta `json:",inline"`
```
ingvagabund: No need for the `json...` annotation part given all the data types are internal.

cloud/aws/providerconfig/types.go
ingvagabund: Can a cluster have more subnets? If so, do we want to allow it? The subnet(s) can be automatically detected based on cluster's vpc ID.

cloud/aws/providerconfig/types.go
ingvagabund: `ec2.InstanceNetworkInterfaceSpecification` wants `*int64`

timothysc:
/assign @detiber 

Can you review please, and we can try to get @csrwng unblocked.

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+	// these tags are present, but will not remove any other tags that may exist on the
+	// instance.
+	Tags []TagSpecification
+
+	// IAMInstanceProfile is a reference to an IAM role to assign to the instance
+	IAMInstanceProfile *AWSResourceReference
+
+	// UserDataSecret contains a local reference to a secret that contains the
+	// UserData to apply to the instance
+	UserDataSecret *corev1.LocalObjectReference
```
chuckha: I think this is outside the scope of the MVP

cloud/aws/providerconfig/types.go
```diff
@@ -17,19 +17,112 @@ import (
...
+	// IAMInstanceProfile is the name of the IAM role to assign to the new instance
+	IAMInstanceProfile string
+
+	// UserDataSecret contains a local reference to a secret that contains the
+	// UserData to apply to the instance
+	UserDataSecret *corev1.LocalObjectReference
+
+	// CredentialsSecret is a reference to the secret with AWS credentials. Otherwise, defaults to permissions
+	// provided by attached IAM role where the actuator is running.
+	CredentialsSecret *corev1.LocalObjectReference
```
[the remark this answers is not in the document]
chuckha: Each machine shouldn't need the aws credentials, just the actuator. I think we can remove this

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+
+	// KeyName is the name of the KeyPair to use for SSH
+	KeyName *string
```
chuckha: Is SSH required? I'd be in favor of nixing SSH until we *really* need it and I don't think we'll be needing it for the MVP

cloud/aws/providerconfig/types.go
chuckha: I think it would be cleaner to remove this for the MVP and add it back in later when we need it.

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+
+	// DeviceIndex is the index of the device on the instance for the network interface attachment.
+	// Defaults to 0.
+	DeviceIndex int
+
+	// PublicIP specifies whether the instance should get a public IP. If not present,
+	// it should use the default of its subnet.
+	PublicIP *bool
```
chuckha: The MVP will be either entirely public or entirely private so each machine does not need this configuration

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+
+	// SecurityGroups is an array of references to security groups that should be applied to the
+	// instance.
+	SecurityGroups []AWSResourceReference
+
+	// Subnet is a reference to the subnet to use for this instance
+	Subnet AWSResourceReference
+
+	// Placement specifies where to create the instance in AWS
+	Placement Placement
```
chuckha: Placement is out of scope for MVP, this can go away for now

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+
+	// LoadBalancerIDs is the IDs of the load balancers to which the new instance
+	// should be added once it is created.
+	LoadBalancerIDs []string
```
chuckha: I think this will live on the cluster object which the machine actuator has access to

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+}
+
+// AWSResourceReference is a reference to a specific AWS resource by ID, ARN, or filters.
+// Only one of ID, ARN or Filters may be specified. Specifying more than one will result in
+// a validation error.
+type AWSResourceReference struct {
```
chuckha: This seems like it could get messy? How does this fit into the MVP?

cloud/aws/providerconfig/types.go
```diff
@@ -14,22 +14,156 @@
...
+
+	// ARN of resource
+	ARN *string
+
+	// Filters is a set of filters used to identify a resource
+	Filters []Filter
+}
+
+// Placement indicates where to create the instance in AWS
+type Placement struct {
```
chuckha: out of scope for MVP

chuckha:
To give a little more context here, I'm more than happy to keep this PR as is if this work has resources, but right now it might be a little cumbersome for the agreed upon MVP. If you're happy with it and want to continue with the PR as is, we can definitely make it work.

csrwng:
@chuckha Thank you for the review. I'm ok with paring it down if it'll reduce friction for an initial implementation and as we add features/function, add things back into it.

cloud/aws/providerconfig/types.go
csrwng: It allows you to specify references to things that might not exist yet when you create your machine/machineset config. For example, you may know what your security groups will be named, but you don't have id's for them yet. Your config can use a Filter for the name tag for each of your security groups.

chuckha:
That sounds good to me -- we should track the stuff that we nix with a ticket. I expect the backlog to get properly populated with tasks for the MVP this week and I don't want us to forget to add things back

csrwng:
@chuckha Addressed your comments. The Placement struct included the region for the machine. Are we assuming that in an initial implementation the actuator would have a default region that it uses?

csrwng:
squashed and updated versioned types

cloud/aws/providerconfig/types.go
```diff
@@ -14,25 +14,132 @@
...
+	AMI AWSResourceReference
+
+	// InstanceType is the type of instance to create. Example: m4.xlarge
+	InstanceType string
+
+	// Tags is the set of tags to add to apply to an instance, in addition to the ones
+	// added by default by the actuator. These tags are additive. The actuator will ensure
+	// these tags are present, but will not remove any other tags that may exist on the
+	// instance.
+	Tags []TagSpecification
```
mvladev: Isn't it better for this type to be `map[string]string`?
AWS client accepts `map[string]string` already and with this implementation, you have to loop over every single item to put them in a `map`

cloud/aws/providerconfig/types.go
```diff
@@ -14,25 +14,132 @@
...
+	Status corev1.ConditionStatus
+	// LastProbeTime is the last time we probed the condition.
+	// +optional
+	LastProbeTime metav1.Time
+	// LastTransitionTime is the last time the condition transitioned from one status to another.
+	// +optional
+	LastTransitionTime metav1.Time
+	// Reason is a unique, one-word, CamelCase reason for the condition's last transition.
+	// +optional
+	Reason string
```
detiber: Why wouldn't we bubble these up to the MachineStatus ErrorReason/ErrorMessage instead of having a separate Reason/Message here?

cloud/aws/providerconfig/types.go
csrwng: The only reason I did it this way is: https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#lists-of-named-subobjects-preferred-over-maps
But if there are no objections to reverting it back to a map, then I can do so.

csrwng: Actually now that I read the guidelines again... this is a pure map ... not a map of subobjects, so it doesn't make sense to make it an array.

cloud/aws/providerconfig/types.go
```diff
@@ -14,25 +14,132 @@
...
+	// these tags are present, but will not remove any other tags that may exist on the
+	// instance.
+	Tags []TagSpecification
+
+	// IAMInstanceProfile is a reference to an IAM role to assign to the instance
+	IAMInstanceProfile *AWSResourceReference
+
+	// DeviceIndex is the index of the device on the instance for the network interface attachment.
+	// Defaults to 0.
+	DeviceIndex int
```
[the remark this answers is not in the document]
csrwng: Sounds good, will remove

cloud/aws/providerconfig/types.go
csrwng: Conditions are meant to provide more detailed status, giving you time, condition, etc. and allowing the possibility of reporting multiple conditions at a time. If we want to provide a more concise status at the higher level, then I can add an ErrorReason and ErrorMessage there.

cloud/aws/providerconfig/types.go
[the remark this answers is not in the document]
csrwng: will do

cloud/aws/providerconfig/types.go
csrwng: So far all of the resources that we reference from the machine config have a corresponding Describe call that allows them to be looked up by the fields specified in the struct. I would say we could introduce separate types of references in the future if/when we need them. Is that ok?

detiber: Sounds good, thanks for the clarification.

cloud/aws/providerconfig/types.go
```diff
@@ -14,25 +14,132 @@
...
+
+	// ARN of resource
+	ARN *string
+
+	// Filters is a set of filters used to identify a resource
+	Filters []Filter
+}
+
+// Filter is a filter used to identify an AWS resource
+type Filter struct {
```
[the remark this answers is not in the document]
detiber: Makes sense, can we add a link to that to the comments to make it more clear for new users?

cloud/aws/providerconfig/types.go
detiber: There is already an ErrorReason/ErrorMessage on the MachineStatus object: https://github.com/kubernetes-sigs/cluster-api/blob/051f338bdacb76117d73a86258bc58c946add7b5/pkg/apis/cluster/v1alpha1/machine_types.go#L112-L133

It sounds like there could be a use case for duplicating the Reason/Message here as well for more granularity, though.

cloud/aws/providerconfig/types.go
errordeveloper: I though this subproject wasn't going to be so prescriptive, and it would be up to any given flavour to make a decision.

cloud/aws/providerconfig/types.go
```diff
@@ -14,25 +14,123 @@
 package providerconfig
 
 import (
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
+// AWSMachineProviderConfig is the type that will be embedded in a Machine.Spec.ProviderConfig field
```
errordeveloper: This doesn't actually say anything about what this struct is supposed to represent. Is it a single running instance, or template of an instance or what? The comment must clarify that, at the very least. Instead it talks about relationship between structs, which is something one can figure by reading the code anyway.

cloud/aws/providerconfig/types.go
errordeveloper: I think there confusion around what this struct represents, as I pointed out above, hence the meaning (and type) of `Subnet` field is unclear.

cloud/aws/providerconfig/types.go
errordeveloper: Perhaps, it'd make more sense to use a separate struct for inputs and outputs. That's what AWS SDK does and it's more convenient to not create a additional model that one has to thing about.

cloud/aws/providerconfig/types.go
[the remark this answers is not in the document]
errordeveloper: Why do we redefine the type here? Why not alias https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#Filter?

cloud/aws/providerconfig/types.go
errordeveloper: Also, I think it's very reasonable to ask where this should be just an alias to [`ec2.Instance`](https://docs.aws.amazon.com/sdk-for-go/api/service/ec2/#Instance)?

csrwng:
@detiber @errordeveloper thank you for your comments. I've now updated the PR.

cloud/aws/providerconfig/types.go
detiber: @errordeveloper That is a goal for the overall project, yes. That said, I think there is an inherent impedance mismatch between a Machine and an ASG-managed instance that we probably need to solve more generically in the great cluster-api.

To me it feels like the MachineSet controller would be a better manager for ASG-backed instances rather than the Machine controller. I am interested on hearing more of your thoughts on this topic, though.

csrwng:
@errordeveloper I mentioned the reasons for not using the EC2 types directly in the proposal doc. Having separate types also gives us some freedom to choose the EC2 client we use. For example, at some point we may want to switch to the [V2 client](https://aws.amazon.com/blogs/developer/aws-sdk-for-go-2-0-developer-preview/).

cloud/aws/providerconfig/types.go
```diff
@@ -14,25 +14,140 @@
...
+	// 2. Cluster/flavor setting
+	// 3. Subnet default
+	// +optional
+	PublicIP *bool
+
+	// SecurityGroups is an array of references to security groups that should be applied to the
+	// instance. These security groups would be set in addition to any security groups defined
+	// at the cluster level or in the actuator.
+	// +optional
+	SecurityGroups []AWSResourceReference
```
detiber: Can this be renamed to AdditionalSecurityGroups to align with the updated comments?

csrwng:
done

detiber:
/lgtm

@errordeveloper are you able to attend the implementation meeting (https://groups.google.com/d/msg/kubernetes-sig-cluster-lifecycle/__XIKigkxkA/j1BHBdznDwAJ https://docs.google.com/document/d/10dq54Fd-xa6P5Iy3p46VY1YTFqugGMd1PygDIpuRw6c/edit#) to discuss your questions/concerns further?

errordeveloper:
I would like to attend regularly, but it clashes with general implementer's
office hours, which I am currently chairing. Let's discuss how meetings
align/misalign in the WG call today.

errordeveloper:
As far as I am aware, the V2 client is not very different with respect to
these struct are implemented.

PR #9, opened 2018-08-15
Adds vendor directory

chuckha:


.gitignore
```diff
@@ -10,6 +10,3 @@
 
 # Output of the go coverage tool, specifically when used with LiteIDE
 *.out
-
-# Vendor director
-vendor
```
detiber: Any reason to explicitly add vendor instead of adding a `make depend` step to the failing scripts?

chuckha: no. that sounds nicer than checking in 3M lines of code though. Will update.

PR #10, opened 2018-08-15
fix tests

chuckha:

### DIFF
--- a/scripts/ci-build.sh
+++ b/scripts/ci-build.sh
@@ -19,5 +19,5 @@ set -o nounset
 set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
-
+cd "${REPO_ROOT}" && make depend
 go build  "${REPO_ROOT}"/...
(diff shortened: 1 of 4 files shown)


scripts/ci-build.sh
```diff
@@ -19,5 +19,4 @@ set -o nounset
 set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
-
-go build  "${REPO_ROOT}"/...
+cd "${REPO_ROOT}" && make build
```
detiber: I actually like the idea of using the go build command here, as it will catch build paths that aren't listed in the makefile.

scripts/ci-make.sh
```diff
@@ -19,5 +19,4 @@ set -o nounset
 set -o pipefail
 
 REPO_ROOT=$(dirname "${BASH_SOURCE}")/..
-
-cd $REPO_ROOT && make images
+cd "${REPO_ROOT}" && make images
```
detiber: It might be a good idea to update this to run make all, similar to the main cluster-api script here.

PR #22, opened 2018-08-22
Clusterctl

chuckha:


chuckha:
I'm having a little trouble catching your meaning...

do you mean that `make test` doesn't test the `clusterctl` directory?

detiber:
> I'm having a little trouble catching your meaning...
>
> do you mean that `make test` doesn't test the `clusterctl` directory?

I hadn't even noticed the missing coverage for `make test`... I was mainly trying to point out that we aren't building the binary during the job that runs `make all`.

PR #25, opened 2018-08-27
WIP: image stamping

vincepri:
This is an initial port from [wardroom](https://github.com/heptiolabs/wardroom) adding support for containerd in Ubuntu (RHEL to follow).


chuckha:
Thoughts on purging the wardroom name in favor of something more descriptive like 'kubernetes-image', 'kubeadm-image', or 'cluster-api-provider-aws-image'?

images/ansible/roles/kubernetes/defaults/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+kubernetes_version: "{{ kubernetes.version if kubernetes_version_defined else '1.9.3-00' }}"
```
detiber: Kill all the backwards compat for `kubernetes.*` variables here and switch to simple defaults.

Default for versions should probably be an empty string or undefined, and the role should default to latest available release version if not specified.

At some point we may want to consider a way to handle alpha/beta/rc releases as well, but that isn't a priority for the MVP.

images/ansible/roles/kubernetes/tasks/main.yml
```diff
@@ -0,0 +1,23 @@
...
+  service:
+    name: kubelet
+    state: started
+    enabled: true
+
+- name: cache docker images for kubeadm
+  docker_image:
+    name: "{{ item }}"
+  with_items: kubernetes_cached_images
+  when: kubernetes_enable_cached_images | bool
```
detiber: Lets kill the cached image handling for now, especially since it is docker specific.

images/ansible/roles/kubernetes/templates/etc/systemd/system/kubelet.service
```diff
@@ -0,0 +1,12 @@
...
+Documentation=http://kubernetes.io/docs/
+
+[Service]
+ExecStart=/usr/bin/kubelet
+Restart=always
+StartLimitInterval=0
+RestartSec=10
+
+[Install]
+WantedBy=multi-user.target
```
detiber: We shouldn't need this systemd unit file, it should be managed by the kubelet/kubeadm packages.

images/ansible/roles/kubernetes/templates/etc/systemd/system/kubelet.service.d/10-kubeadm.conf
```diff
@@ -0,0 +1,7 @@
+[Service]
+Environment="KUBELET_KUBECONFIG_ARGS=--kubeconfig=/etc/kubernetes/kubelet.conf --require-kubeconfig=true"
+Environment="KUBELET_SYSTEM_PODS_ARGS=--pod-manifest-path=/etc/kubernetes/manifests --allow-privileged=true"
+Environment="KUBELET_NETWORK_ARGS=--network-plugin=cni --cni-conf-dir=/etc/cni/net.d --cni-bin-dir=/opt/cni/bin"
+Environment="KUBELET_DNS_ARGS=--cluster-dns=10.96.0.10 --cluster-domain=cluster.local"
+ExecStart=
+ExecStart=/usr/bin/kubelet $KUBELET_KUBECONFIG_ARGS $KUBELET_SYSTEM_PODS_ARGS $KUBELET_NETWORK_ARGS $KUBELET_DNS_ARGS $KUBELET_EXTRA_ARGS
```
detiber: For kubernetes v1.11+, we should be setting values in `/var/lib/kubelet/kubeadm-flags.env` insetad of overriding the unit files here.

images/ansible/roles/kubernetes/vars/main.yml
```diff
@@ -0,0 +1,9 @@
+---
+kubernetes_version_defined: "{{ kubernetes is defined and 'version' in kubernetes }}"
+kubernetes_cni_version_defined: "{{ kubernetes is defined and 'cni_version' in kubernetes }}"
+kubernetes_enable_cached_images_defined: "{{ kubernetes is defined and 'enable_cached_images' in kubernetes }}"
+kubernetes_cached_images_defined: "{{ kubernetes is defined and 'cached_images' in kubernetes }}"
+kubernetes_apt_key_url_defined: "{{ kubernetes is defined and 'apt_key_url' in kubernetes }}"
+kubernetes_apt_repo_string_defined: "{{ kubernetes is defined and 'apt_repo_string' in kubernetes }}"
+kubernetes_yum_baseurl_defined: "{{ kubernetes is defined and 'yum_baseurl' in kubernetes }}"
+kubernetes_yum_gpgkey_defined: "{{ kubernetes is defined and 'yum_gpgkey' in kubernetes }}"
```
detiber: This is all backward compat crud.

detiber:
@chuckha good point on the naming. @vincepri we'll also need to migrate over the python tool for sync'ing images. I'm also thinking that it might be a good idea to use that python tool for wrapping the interaction with packer for common use cases as well.

vincepri:
@detiber thanks for the review! fixed most comments and updated the PR. I didn't initially migrate the wardroom tool because we should be able to use packer directly to copy images to different regions [this line](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/25/files/2098247736d87f7ee3f19e906c3c5b3ef6a243b1#diff-117e24ee13e28cd08e452ab02d5d7c88R9) is a list of regions we want to copy the images to. Thoughts?

detiber:
> @detiber thanks for the review! fixed most comments and updated the PR. I didn't initially migrate the wardroom tool because we should be able to use packer directly to copy images to different regions this line is a list of regions we want to copy the images to. Thoughts?

Nice, I think that would work for the general user use case, but we will still probably want some tooling for the community AMI building. The wardroom tool for example will sync to all available regions in the client config and ensure the images are public.

images/ansible/filter_plugins/kube.py
```diff
@@ -0,0 +1,28 @@
...
+            raise Exception("Version '%s' does not appear to be a "
+                            "kubernetes version." % version)
+        sub = match.groups(1)[1]
+        if len(sub) == 1:
+            if platform.lower() == "debian":
+                return "%s-%s" % (match.groups(1)[0], '{:02d}'.format(sub))
+            else:
+                return version
+        if len(sub) == 2:
+            if platform.lower() == "redhat":
```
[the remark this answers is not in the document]
vincepri: Just tested it, seems to be working with "redhat" as platform https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/25/commits/3856319315cc043929089ba85cd12b920593ee37

images/ansible/roles/common/tasks/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+- import_tasks: debian.yml
+  when: ansible_os_family == "Debian"
+
+- import_tasks: redhat.yml
+  when: ansible_os_family == "RedHat"
```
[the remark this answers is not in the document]
vincepri: See https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/25/commits/3856319315cc043929089ba85cd12b920593ee37

images/ansible/roles/kubernetes/defaults/main.yml
```diff
@@ -0,0 +1,3 @@
+---
+kubernetes_version: ""
+kubernetes_cni_version: ""
```
detiber: Does this behave correctly with the defaults here? It appears this may fail, since we are passing the value into the kube_platform_version filter.

images/packer/packer.json
```diff
@@ -0,0 +1,103 @@
+{
+  "variables": {
+    "aws_access_key": "",
+    "aws_secret_key": "",
+    "build_version": null,
```
detiber: I wonder if we should just nix build_version completely here and use `isotime` in it's place... @chuckha thoughts?

images/packer/packer.json
```diff
@@ -0,0 +1,103 @@
...
+    "kubernetes_version": "1.11.2-00",
+    "kubernetes_cni_version": "0.6.0-00",
```
detiber: Ideally, I'd like to see these variables default to "" and have the Ansible roles detect the latest available stable version, with these variables only being used for explicitly overriding the defaults. Other wise we're going to have to continually and explicitly rev this for every release.

images/packer/packer.json
vincepri: I'm +1 on this

images/packer/packer.json
vincepri: If we opt for those value to be empty, should we change the `ami_name` and the tags so they don't include the Kubernetes version?

detiber: doh, no, we still want to set those, otherwise we'll lose the ability to filter images by version.

vincepri: Sounds good, I'll leave these in then and make the ansible default to `latest`.

images/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+## Building Images
+
+### Build Variables
+
+The following variables can be overriden when building images using the `-var` option when calling `packer build`:
+
+| Variable | Default | Description |
+|----------|---------|-------------|
+| build_version | unset | A unique build version for the image |
```
detiber: Lets go ahead and kill build_version in favor of using a timestamp instead. That way we can avoid requiring a user to manually specify it for each image built.

We can always re-add it in the future if using a timestamp becomes an issue.

images/README.md
```diff
@@ -0,0 +1,110 @@
...
+
+For example, to build only the Ubuntu image:
+
+```sh
+packer build -var build_version=`git rev-parse HEAD` --only=ami-ubuntu packer.json
+```
+
+### Required Permissions to Build the AWS AMIs
+
+The [Packer documentation for the Amazon AMI builder](https://www.packer.io/docs/builders/amazon.html) supplies a suggested set of minimum permissions. However, Wardroom has been successfully tested with the following IAM permissions:
```
detiber: kill wardroom reference.

images/README.md
```diff
@@ -0,0 +1,110 @@
...
+                "ec2:DeleteSecurityGroup",
+                "ec2:AuthorizeSecurityGroupIngress",
+                "ec2:CreateTags",
+                "ec2:DeleteVolume"
+            ],
+            "Resource": "*"
+        }
+    ]
+}
+```
```
detiber: Have you been able to test using these limited permissions? I'm wondering if the use of `ami_groups` or `ami_regions` adds additional requirements.

images/packer/packer.json
```diff
@@ -0,0 +1,103 @@
...
+      "source_ami": "{{user `centos_7_4_ami`}}",
+      "ami_name": "ami-centos-7.4-{{user `kubernetes_version`}}-{{timestamp}}",
+      "ami_groups": "{{user `ami_groups`}}",
+      "ami_regions": "{{user `ami_regions`}}",
+      "access_key": "{{user `aws_access_key`}}",
+      "secret_key": "{{user `aws_secret_key`}}",
+      "ssh_username": "centos",
+      "tags": {
+        "build_version": "{{user `build_version`}}",
+        "source_ami": "{{user `centos_7_4_ami`}}",
```
detiber: I'm wondering if we can use `source_ami_filter` here instead to be able to automatically choose the newest base image rather than having to hard code it into a variable file like we are currently doing.

Thoughts?

vincepri: Yeah. I was actually going to change it to use the official versions. The reason I left it untouched is that with the current behavior it's easier for folks to provide their own base image, which might contain different drivers/tweaks/security tools.

images/README.md
vincepri: I've updated the permissions to match the official packer suggested ones. Which include permissions to modify images attributes, needed for ami_groups to function. The `ami_regions` should be good too.

images/README.md
vincepri: Done

images/README.md
vincepri: Sounds good

vincepri:
@detiber Fixed / replied to the other comments. I think this is ready to be merged in. 😊

chuckha:
Do these files need the licensing stanza at the top?

vincepri:
All done! Thanks @chuckha

images/ansible/filter_plugins/kube.py
```diff
@@ -1,3 +1,17 @@
+# Copyright 2018 Heptio Inc.
```
detiber: s/Heptio/The Kubernetes Authors/

We should probably have this on all the Ansible yaml files as well.

vincepri:
@detiber License notices added, squashed commits.

PR #26, opened 2018-08-27
MVP Proposal

detiber:
This brings in the MVP proposal that was discussed during the AWS Provider implementation meeting on August 20, 2018 as well as referenced docs.


detiber:
@chuckha no real changing of content, mostly re-organizing and markdown-ifying.

PR #27, opened 2018-08-27
basic machine actuator and aws instance implmentation

chuckha:
~~This is missing a lot of stuff and clearly isn't working, but it should show you the general way I'm thinking about building out the components. This provides a tiny skeleton to grow and add other bones onto as we flesh out the initial implementation.

Thoughts? I'm looking for high level feedback, not details in particular since those are not so hard to change.

Signed-off-by: Chuck Ha <chuck@heptio.com>~~

EDIT: This is a basic implementation ready for code review. This likely doesn't work and likely has bugs and isn't tested but should provide a framework to pick up a little more momentum.

cc @detiber

### DIFF
--- a/cloud/aws/services/ec2/instances.go
+++ b/cloud/aws/services/ec2/instances.go
@@ -1,0 +1,71 @@
+// Copyright © 2018 The Kubernetes Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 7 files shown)


cloud/aws/services/ec2/instances.go
```diff
@@ -0,0 +1,70 @@
...
+	"github.com/aws/aws-sdk-go/service/ec2"
+	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+type Instance struct {
+	state string
+	id    string
+}
+
+func (i *Instance) Running() bool {
```
chuckha: get rid of these functions

cloud/aws/services/ec2/instances.go
```diff
@@ -0,0 +1,70 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+type Instance struct {
+	state string
```
chuckha: export these

PR #55, opened 2018-08-29
Cluster actuator

chuckha:
/hold

based on #27 being merged first


cloud/aws/services/ec2/instances.go
```diff
@@ -0,0 +1,71 @@
...
+import (
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+// instances is used to scope down/organize the ec2 client.
+type instances interface {
```
vincepri: The interface name seems a little generic, should we pick something more specific like `ec2InstanceClient` or is this a common convention?

cloud/aws/services/ec2/instances.go
```diff
@@ -0,0 +1,71 @@
...
+			State: *out.Reservations[0].Instances[0].State.Name,
+			ID:    *out.Reservations[0].Instances[0].InstanceId,
+		}, nil
+	}
+	return nil, nil
+}
+
+// CreateInstance runs an ec2 instance.
+func (s *Service) CreateInstance(machine *clusterv1.Machine) (*Instance, error) {
+	input := &ec2.RunInstancesInput{}
```
vincepri: I assume the input here is going to be populated with the `machine` options?

cloud/aws/services/ec2/instances.go
```diff
@@ -0,0 +1,71 @@
...
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+)
+
+// instances is used to scope down/organize the ec2 client.
+type instances interface {
+	DescribeInstances(*ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error)
+	RunInstances(*ec2.RunInstancesInput) (*ec2.Reservation, error)
```
vincepri: Given that these methods are used within an interface, which gives the context, what do you think about dropping the `Instances` suffix at the end? For example [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/55/files#diff-f84006d2b31889959340918e85b391deR44) there is some repetition.

cloud/aws/services/ec2/vpc.go
```diff
@@ -0,0 +1,57 @@
...
+
+import (
+	"errors"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+type vpcs interface {
+	DescribeVpcs(*ec2.DescribeVpcsInput) (*ec2.DescribeVpcsOutput, error)
```
vincepri: Same as above, should we drop the `Vpcs` suffix?

cloud/aws/services/ec2/instances.go
chuckha: i figure it's a local interface so it doesn't have to be terribly descriptive

cloud/aws/services/ec2/instances.go
chuckha: I can't change the methods, those are coming from the ec2 client. The goal of this interface is to scope down the client. I'm using a *ec2.EC2 pointer to satisfy this interface I'm defining here. But because i'm defining exactly what it is I need out of the *ec2.EC2 I can mock this in tests by implementing two functions instead of like, 300

cloud/aws/services/ec2/vpc.go
chuckha: again, cannot because it comes from the ec2 sdk

Issue #53, opened 2018-08-29
[aws] Idempotent create instance


chuckha:
There was talk of tokens and tags and opinions and I'd like to capture those ideas here.

cc @kris-nova @csrwng

csrwng:
For creating instances we can use client tokens to ensure uniqueness of the instance to create. Info here: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/Run_Instance_Idempotency.html
One option would be to use the machine UID as the token. Or if we want to not rely on a specific UID, we could generate a hash based on Name + Namespace.

detiber:
Updating priority because I think we want to solve this before we get too far into the implementation

/priority important-soon

detiber:
I think we also want to get general agreement into the contract we want to provide about idempotent creation and deletion of AWS resources to ensure that we are not leaving users with orphaned resources.

kops relies heavily on tags applied to created resources
kubicorn relies on storing the resource ids as part of the state and relies on deleting resources when failures are hit.

Some conditions we'll need to take into account:
- resource create succeeds, but subsequent tagging fails
- resource creates succeeds, but update of cluster/machine object fails
- attempting to delete resource fails after a successful create it but before recording/tagging it
- the controller/actuator dies after creating a resource but before tagging and or recording the resource

I don't think we'll be able to fully solve all of the potential problems, but we need to target a best effort and leave a clean logging/event trail for users to be able to manually resolve the cases we can't.

PR #66, opened 2018-08-30
Start docs around idempotent AWS resource management

detiber:
Attempt to start documenting the guarantees that we will provide around idempontent resource handling for AWS.

Related to: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/53


detiber:
Pretty rough still, but I'd like to get some feedback in before we get too far into the client implementation @chuckha @vincepri @kris-nova @csrwng @d-nishi @davidewatson

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Existing solutions and drawbacks for our use
+
+### kops
+
+kops relies heavily on tagging of resources. This provides the benefit of recreating the state of the cluster resources by querying with filters. However, in many cases tagging requires a second API call and there are some resources that are not able to be tagged. If we succeed in creating a resource but fail to tag the resource, then we risk the chance of orphaning that resource for the user.
+
+### Kubicorn
+
+In contrast to the kops approach, Kubicorn mainly relies on recording the resource IDs as part of state. However, since we rely on an external API server for recording the state there is still a possibility of creating the resource and failing to record the resource ID, which still exposes the possible risk of orphaning that resource for the user.
```
detiber: @kris-nova I know I greatly over-simplified both the kops and Kubicorn approaches here, would love if you could help add some additional detail and clarification.

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,45 @@
...
+- resource create succeeds, but subsequent tagging fails
+- resource creates succeeds, but update of cluster/machine object fails
+- attempting to delete resource fails after a successful create it but before recording/tagging it
+- the controller/actuator dies after creating a resource but before tagging and or recording the resource
+
+## Proposed workflow
+
+Where possible use [client tokens](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/Run_Instance_Idempotency.html) in the create request so that subsequent requests will return the same response.
+
+### Create resource with tag support
```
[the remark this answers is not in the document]
detiber: @chuckha I think we would want to add additional tags to use for configuration metadata similar to what is done by the [AWS cloudprovider](https://github.com/kubernetes/kubernetes/blob/master/pkg/cloudprovider/providers/aws/aws.go#L83-L93)

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,45 @@
...
+
+### Kubicorn
+
+In contrast to the kops approach, Kubicorn mainly relies on recording the resource IDs as part of state. However, since we rely on an external API server for recording the state there is still a possibility of creating the resource and failing to record the resource ID, which still exposes the possible risk of orphaning that resource for the user.
+
+## Summary of edge cases for creating an individual resource
+
+- resource create succeeds, but subsequent tagging fails
+- resource creates succeeds, but update of cluster/machine object fails
+- attempting to delete resource fails after a successful create it but before recording/tagging it
```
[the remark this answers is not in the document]
detiber: Hopefully the new verbiage is more clear. The intent was to take into account that after a successful create, but a failure to record the resource ID that the resource delete also failed.

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,45 @@
...
+- the controller/actuator dies after creating a resource but before tagging and or recording the resource
```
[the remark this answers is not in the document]
detiber: How would that work for resources that we aren't applying known tags to?

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,200 @@
...
+
+### Kubicorn
+
+In contrast to the kops approach, Kubicorn mainly relies on recording the resource IDs as part of state. However, since we rely on an external API server for recording the state there is still a possibility of creating the resource and failing to record the resource ID, which still exposes the possible risk of orphaning that resource for the user.
+
+## Summary of edge cases for creating an individual resource
+
+1. resource create succeeds, but subsequent tagging fails
+2. resource creates succeeds, but update of cluster/machine object fails
+3. attempting to delete resource fails after an attempt to rollback due to a failure to record the ID of the created resource to the cluster/machine object for resources that do not support tagging on create.
```
chuckha: highlighting this since it's a good idea and maybe hasn't been explicitly stated: if the object store goes away and we cannot save references to the resources we create then we should rollback.

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,200 @@
...
+4. the controller/actuator dies after creating a resource but before tagging and or recording the resource
+
+## Misc TODOs
+
+- Incorporate retries for AWS or cluster-api requests?
```
chuckha: "Each AWS SDK implements automatic retry logic. " - https://docs.aws.amazon.com/general/latest/gr/api-retries.html

not sure about the cluster-api object store calls

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Using client tokens
+
+Where possible use [client tokens](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/Run_Instance_Idempotency.html) in the create request so that subsequent requests will return the same response.
+
+## Tagging of resources
+
+Resources that are managed by the controllers/actuators should be tagged with: `kubernetes.io/cluster/<name or id>=owned`
+
+Question: Since other tools/components interact and create resources using the `kubernetes.io/cluster/<name or id>=owned` tag, should we also add a cluster-api-provider-aws specific tag for filtering?
```
chuckha: that sounds good to me. I'll propose `sigs.k8s.io/cluser-api-provider-aws=true`. I left out the unique identifier so we could have a way to list all resources that are owned by this provider and because the other tag already has a unique (per cluster) identifier.

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,200 @@
...
+## Proposed workflows
+
+### Resources that support tag on create support
+
+#### Create
+
+- Query resource by tags to determine if resource already exists
+- Create the resource if it doesn't already exist
+- Update the cluster/machine object config and status
+  - If update fails return a retryable error to reque the create
```
chuckha: typo => requeue

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,200 @@
...
+- Enque cluster/machine update if not already available/ready
```
chuckha: typo => enqueue

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,200 @@
...
+
+## Handling of AWS api errors
+
+Each resource has specific error codes that it will return and these can be used to differentiate fatal errors from retryable errors. These errors are well documented in some cases (elbv2 api), and poorly in others (ec2 api). We should provide a best effort to [properly handle these errors](https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/handling-errors.html) in the correct manner.
+
+## Proposed workflows
+
+### Resources that support tag on create support
+
+#### Create
```
chuckha: The Reconcile case is going to be slightly different since we might also be able to query by ID before calling create

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,200 @@
...
+#### Update
+
+- Query resource by ID
+- Update object status
+- Enque cluster/machine update if not already available/ready
+
+#### Edge case coverage
+
+1. Yes - tagging is handled on creation
+2. Yes - since resources are tagged on creation, returning an error and requeing the create will find the tagged resource and attempt to retry the object update.
```
chuckha: typo => requeuing

docs/proposal/aws-resource-handling.md
detiber: This should only be necessary in the case that the operation does not support client tokens or tagging as part of create based on the workflows proposed below.

docs/proposal/aws-resource-handling.md
chuckha: ahh i see ok 👍

docs/proposal/aws-resource-handling.md
detiber: It looks like client-go does as well (at least in the latest version): https://github.com/kubernetes/client-go/blob/master/rest/request.go

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,203 @@
...
+4. the controller/actuator dies after creating a resource but before tagging and or recording the resource
+
+## Misc TODOs
+
+- Solicit feedback on whether aws-sdk-go and client-go retry defaults are sufficient:
+  - aws-sdk-go
+    - https://docs.aws.amazon.com/general/latest/gr/api-retries.html
+    - https://docs.aws.amazon.com/sdk-for-go/api/aws/client/#DefaultRetryer
+  - client-go
+    - https://github.com/kubernetes/client-go/blob/master/rest/request.go#L659
```
detiber: @kris-nova can you provide any context to why Kops leverages additional retries outside of the ones built into the aws client?

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,203 @@
...
+- Enqueue cluster/machine update if not already available/ready
+
+#### Edge case coverage - option 2
+
+1. Yes - If only the tagging fails, then we will reconcile tags on update. If the update also fails, then we attempt to delete the resource.
+2. Yes, with caveat - If only the object update fails, then we throw a retryable error that will requeue the create operation and attempt to update the object after discovering the existing resource. If the tagging fails as well, then we attempt to delete the resource and edge case 3 will still apply.
+3. Minor mitigation - If we fail to delete the resource, we will still orphan the resource, but output a log message for querying/followup and return a non-retryable error
+4. Minor mitigation - If the process dies before recording the ID the resource is orphaned. If the process dies after recording the ID, but before tagging it is reconciled through update.
+
+### Resources without tag support with client token support
```
chuckha: this is a redundant case with case: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/66/files#diff-ca0d5b46102c7cf8ceb425bb9f57cd95R76

detiber: I need to improve the wording here, the other case are resources that support tags, just not tagging on create.

detiber:
@chuckha I'm not sure we gain anything by running a separate workflow for resources that support tagging on create with or without client token support.

That said, instances w/ launch templates are a niche corner case that we'll need to be clear about which workflow they fall under.

docs/proposal/aws-resource-handling.md
detiber: I updated the wording on the other workflow to hopefully make it more clear.

chuckha:
ah yes, i think the wording was misleading me, it's clear after your most recent update

errordeveloper:
I wonder why CloudFormation is not being discussed in this context? It does make handling of resources as a group much easier. You don't need to use all of it's features, and no need to code in YAML/JSON, I've been using CloudFormation (driven by Go code) very successfully in eksctl, and happy to share what I've learned.

randomvariable:
Prior references:

- [Sep 4 Cluster API Provider AWS](https://docs.google.com/document/d/10dq54Fd-xa6P5Iy3p46VY1YTFqugGMd1PygDIpuRw6c/edit#heading=h.48yay0jtaorb) ([Video](https://youtu.be/YPHj6Cp1-DY))
- [Sep 7 SIG AWS](https://docs.google.com/document/d/1-i0xQidlXnFEP9fXHWkBxqySkXwJnrGJP9OGyP2_P14/edit#heading=h.6iifpfr9xro7)
- [Lessons learnt from kops, kubicorn and more](https://docs.google.com/document/d/10dq54Fd-xa6P5Iy3p46VY1YTFqugGMd1PygDIpuRw6c/edit#)

From the last couple of meetings, two separate issues emerged, and I thought they might have got conflated in the last SIG AWS meeting.

## AutoScaling groups vs. controller provisioning of instances
The constraint here is that EC2 limits will come into play as soon as a moderate number (of value x) of instances are provisioned. x can't be known ahead of time, and is a dynamic limit based on customer use, other customers and regional load but can be in the 2-digit per 2-digits of seconds range. [Every endpoint](https://docs.aws.amazon.com/general/latest/gr/rande.html) in each region has its own rate limiting, and some are more prone to rate limiting than others (e.g. Route53). 

Therefore, IMO, it's almost a given to use ASGs in preference to a controller using `EC2.RunInstances`, and to make the reconciliation work. It could be reconciliation itself requires a large number of calls to `EC2.DescribeInstances` or other bits, for example that would make this moot however.

## CloudFormation vs. self-managed basic or DAG-based reconciliation

### Points in favour of CloudFormation
- Almost certainly quicker to get going, i.e. eksctl
- [GoFormation](https://github.com/awslabs/goformation) makes programmatic generation less horrible than it was before
- Don't have to manage DAGs for AWS resources
- Retries are taken care of for us

### Points against
- Reconciliation will be split between two different things that will operate in different ways.
- Rollback can be problematic (though much less so than it used to be)
- We won't know the IDs of almost all of the resources created by CF at the time of firing off Update/CreateStack, and will have to tail the output of CF resources to get them.
- People don't like CF's naming conventions (but you almost always end up having to implement some naming schemes).
- Reasoning about what goes into each stack before passing back to the controller's reconciliation may be difficult?

docs/proposal/aws-resource-handling.md
randomvariable: Not sure I prefer using Tags over ResourceIDs by default. We did [rely on tags for resources in the past](https://github.com/scalefactory/build-cloud), but then never sorted out what to do for resources that didn't have them (@jtopper can speak more authoritatively on that one). At some point we need to track ARNs or other resource IDs, and it might make more sense to use that universally? 

Do we know the cases in which the controllers would lose access to the API Server as a result of a change it's made? If it's during cluster turndown, could we "unpivot" the controller instead?

randomvariable:
Additional comments from @mikemead:

1. ASG vs standalone instances:
- seems like tagging might be important; ASGs propagate these at launch, there’s a delay + eventual consistency which might cause issues with instance bootstrapping (hit EC2 API limits early on at one of our customers with instances polling the API for tagging info)
2. CloudFormation vs self-managed provisioning:
- Self-managed requires granting something (EC2 instances?) an IAM policy to create/destroy EC2 things. With CloudFormation this isn’t necessary, you just give the thing the ability to delegate a role to CloudFormation - so the security story is a bit neater

randomvariable:
More comments from @mikemead:

> 1. ASG vs standalone instances:
> - seems like tagging might be important; ASGs propagate these at launch, there’s a delay + eventual consistency which might cause issues with instance bootstrapping (hit EC2 API limits early on at one of our customers with instances polling the API for tagging info)
> 2. CloudFormation vs self-managed provisioning:
> - Self-managed requires granting something (EC2 instances?) an IAM policy to create/destroy EC2 things. With CloudFormation this isn’t necessary, you just give the thing the ability to delegate a role to CloudFormation - so the security story is a bit neater

docs/proposal/aws-resource-handling.md
jtopper: The approach we used in build-cloud was similar to that used when building a Puppet provider: we'd choose an attribute that looked the most like a "primary key" for that type of object to track, and if there wasn't an obvious one from the resource, we'd use a name tag. There are some resource types where that wasn't possible because tagging wasn't available for that type; and in some cases the parameter we chose as a "primary key" can't be guaranteed unique by the platform - but we could assume that it was in our own usage. I think tracking resource IDs is probably the more trustworthy approach here, even though it'll take more engineering effort.

randomvariable:
Now that I'm actively working on this, do we also need something on pagination of resources?

docs/proposal/aws-resource-handling.md
```diff
@@ -0,0 +1,203 @@
...
+  - If update fails return a retryable error to requeue the create
+- Tag AWS resource
+- Enqueue cluster/machine update if not already available/ready
+
+TODO: flowchart
+
+#### Update
+
+- Query resource by ID
+- tag resource if missing tags
```
johanneswuerbach: This might cause problems for `shared` resources like a VPC used for multiple clusters. kubernetes core is using `kubernetes.io/cluster/<name or id>=owned` and `kubernetes.io/cluster/<name or id>=shared` to differentiate and store whether it created the resource itself or uses a shared one.

Problems:
1. Resources can only have up to 50 tags, so the re-usability of a VPC might be limited if every cluster adds a `shared` tag
1. If a resource is encountered without tag is discovered, should cluster-api assume ownership or fail? [kops](https://github.com/kubernetes/kops/blob/af4df08b694e2a1f8814a7b3649060477be67c86/docs/run_in_existing_vpc.md#running-in-a-shared-vpc) and [kubespray](https://github.com/kubernetes-incubator/kubespray/blob/e3d562bcdb0da3bc31c079907c9e04e0c1a43c68/docs/aws.md#aws) require explicit tagging of shared resources, which solves the problem but might be bad user experience.

johanneswuerbach:
I don't want to derail this issue, but as it was already mentioned, here my 2ct

> AutoScaling groups vs. controller provisioning of instances

We have actually encountered the opposite behaviour and are heavily running into ASG rate-limits, but not EC2s. Currently we have ~ 50 clusters all using https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler and while v1.3 improved the situation, we are still frequently seeing them.

The other problem with ASGs is that instance creation failures like user limit of instance type x, AZ limit of instance type x, misconfiguration are not directly visible as the instance isn't created explicitly, which at least in the autoscaler is an unsolved problem https://github.com/kubernetes/autoscaler/issues/1263

Also keep in mind that with autoscaling the number of required ASGs quickly raises as you need always 3 ASGs per instance type and spot vs. non-spot. So using two instance types and having both utilising spot instances if cheaper will end up with 12 ASGs for a single cluster.

vincepri:
@detiber Should this be merged given that the MVP is out? I'd ok tracking the other comments in separate issues, if they are still valid.

randomvariable:
@detiber @vincepri Yeah, I agree. Let's iterate on this.

PR #67, opened 2018-08-31
Adds simple Exists implementation to machine actuator

chuckha:

I'm not sure if this makes sense to merge now, but it's a little bit of code/test implements a very basic existence check


cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -35,7 +35,8 @@ func (e *ec2) InstanceIfExists(id *string) (*ec2svc.Instance, error) {
 	}
 	if *id == "abc" {
 		return &ec2svc.Instance{
-			ID: "abc",
+			ID:    "abc",
+			State: "running",
```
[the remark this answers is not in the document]
chuckha: yeah. actually i should probably use a constant assuming the ec2 sdk defines instance states as constants.

cloud/aws/actuators/machine/actuator.go
```diff
@@ -13,6 +13,7 @@
 
 package machine
 
+// should not need to import the ec2 sdk here
```
chuckha: Is this excessive?

cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -66,3 +67,43 @@ func TestCreate(t *testing.T) {
...
+func TestExists(t *testing.T) {
+	testCases := []struct {
+		name    string
+		exists  bool
+		machine *clusterv1.Machine
+	}{
+		{
+			name:    "machine does not exist",
+			exists:  false,
+			machine: &clusterv1.Machine{},
```
detiber: Should there be a positive test case here as well?

chuckha: yeah, i think the pattern outlined in my other pr (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/74) makes this a lot nicer

chuckha:
Test deleted because the actuator test covered the test cases of the instance_test.go file. If we need to specifically test one piece of instance we can do that, but we should prefer the more integration style testing where we use the real actuator and the real service with a mocked out ec2.

cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -16,28 +16,28 @@ package machine_test
...
-	if id == nil {
-		return nil, nil
-	}
-	if *id == "abc" {
-		return &ec2svc.Instance{
-			ID: "abc",
-		}, nil
-	}
+
+func (i *instancesSvc) DescribeInstances(input *ec2.DescribeInstancesInput) (*ec2.DescribeInstancesOutput, error) {
```
chuckha: will convert to gomock later

PR #68, opened 2018-09-04
Add initial implementation for vpc

vincepri:


cloud/aws/services/ec2/vpc.go
```diff
@@ -17,41 +17,85 @@ import (
...
 	input := &ec2.DescribeVpcsInput{
 		VpcIds: []*string{
 			&id,
 		},
 	}
+
 	out, err := s.VPCs.DescribeVpcs(input)
 	if err != nil {
-		return nil, fmt.Errorf("failed to describe vpcs: %v", err)
+		return nil, ErrVPCNotFound
```
chuckha: This loses the underlying error, it needs to be logged or wrapped in this error message.

cloud/aws/services/ec2/vpc_test.go
```diff
@@ -14,37 +14,79 @@
...
 
-	id := "world"
-	vpc, err := s.ReconcileVPC(id)
-	if err != nil {
-		t.Fatalf("got an unexpected error: %v", err)
-	}
-	if vpc.ID != "world" {
-		t.Fatalf("Expected an id of %v but found %v", "world", vpc.ID)
-	}
+	t.Run("WithExistingVPC", func(t *testing.T) {
```
chuckha: please wrap these into a table test like this:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/67/files#diff-76dfa0346698cdb9b60bf6a38a07141bR71

Generally each error case will require its own TestFunction or there will be a ton of extra boolean flags and complex logic in the t.Run() function. Or the error cases can be untested if it's not worth it. That's a trade off I often make.

cloud/aws/services/ec2/vpc.go
```diff
@@ -17,41 +17,85 @@ import (
...
 }
 
+func (s *Service) createVPC(v *v1alpha1.VPC) (*VPC, error) {
+	input := &ec2.CreateVpcInput{
+		CidrBlock: aws.String(v.CidrBlock),
+	}
+
+	out, err := s.VPCs.CreateVpc(input)
+	if err != nil {
+		return nil, err
```
chuckha: Should this error be wrapped?

cloud/aws/services/ec2/vpc.go
```diff
@@ -17,41 +17,85 @@ import (
...
+		if err == ErrVPCNotFound {
+			return s.createVPC(v)
+		}
+
 		return nil, err
 	}
-	// If it doesn't exist, create it
+
+	if v.CidrBlock != vpc.CidrBlock {
+		// TODO(vincepri): delete and recreate?
```
chuckha: hmm, deleting and recreating the VPC doesn't sound like an ideal thing...maybe we hard error if someone tries to update a field that is not updatable? TODO is ok for now, but let's think about it/track it

cloud/aws/services/ec2/vpc.go
```diff
@@ -17,41 +17,85 @@ import (
...
+}
+
 func (s *Service) lookupVPCByID(id string) (*VPC, error) {
 	input := &ec2.DescribeVpcsInput{
 		VpcIds: []*string{
 			&id,
 		},
 	}
+
 	out, err := s.VPCs.DescribeVpcs(input)
```
chuckha: We should handle the real ec2 aws errors here and either return them or handle them. That way we won't need to define custom errors (probably)

cloud/aws/services/ec2/errors.go
```diff
@@ -0,0 +1,8 @@
+package ec2
+
+import "errors"
+
+var (
+	// TODO(vincepri): gather more information like vpc arn.
+	ErrVPCNotFound = errors.New("vpc not found")
```
chuckha: I would prefer errors to be defined more like how they are described here so we can use functions to check if it's a type of error we expect instead of relying on string matching. https://blog.golang.org/error-handling-and-go

vincepri: Yep, agree. Just wanted to get something up quickly to share the initial shape and get some early comments.

vincepri:
@chuckha thanks for the feedback! 

re:mock, I haven't gotten around that yet, it'd require a little more work and it might be necessary to see if it's worth it for the first iteration of this project.

cloud/aws/services/ec2/vpc.go
vincepri: I think we should probably have a flag to allow the VPC to be re-created. To delete a VPC each single item inside of it needs to be cleaned up first, for example instances, gateways, etc. 

Another thing to consider is that VPCs can now have additional cidr blocks, which are added through additional calls.

cloud/aws/services/ec2/vpc.go
vincepri: Fixed, this was a mistake.

cloud/aws/services/ec2/vpc.go
vincepri: I think returning the errors might be ok here? From the [AWS docs](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeVpcs.html) these calls can return generic client errors, which may require user action. Alerting the user might be the first approach, we can always revisit it later.

cloud/aws/services/ec2/service.go
```diff
@@ -17,6 +17,6 @@ package ec2
 // The interfaces are broken down like this to group functions together.
 // One alternative is to have a large list of functions from the ec2 client.
 type Service struct {
-	Instances instances
 	VPCs      vpcs
+	Instances instances
```
chuckha: Why the move here? This is probably going to get big and I was hoping to keep it alphabetical but I'm open to other ideas.

vincepri: No reason, it probably just slipped through. Will fix

cloud/aws/services/ec2/vpc.go
vincepri: I think this can be just bubbled up as it is. Clients might use `IsSDKError(err)` helper function to handle it.

chuckha:
Oh, I think this needs to be rebased

cloud/aws/services/ec2/vpc.go
```diff
@@ -14,44 +14,85 @@
...
 }
 
 type VPC struct {
-	ID string
+	ID        string
+	CidrBlock string
 }
 
-func (s *Service) ReconcileVPC(id string) (*VPC, error) {
+func (s *Service) ReconcileVPC(v *v1alpha1.VPC) (*VPC, error) {
```
chuckha: This is a pointer and the interface wants a value, i'm more inclined to pass the pointer around unless there is good reason to pass-by-value, but i don't see it here

vincepri: I left it as a value anywhere, I'm ok passing it as a reference though.

PR #74, opened 2018-09-05
clean up mocks

chuckha:
What do you think of this style of mocking? This avoids the if/else blocks in the mock logic.


chuckha:
there is a better way to do this

PR #76, opened 2018-09-06
refactor tests with gomock

vincepri:


cloud/aws/actuators/cluster/actuator.go
```diff
@@ -25,10 +25,6 @@ import (
 	client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
 )
 
-type ec2Svc interface {
```
chuckha: Why is this being removed?

cloud/aws/services/ec2/interface.go
```diff
@@ -0,0 +1,22 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ec2
+
+import (
+	providerconfigv1 "sigs.k8s.io/cluster-api-provider-aws/cloud/aws/providerconfig/v1alpha1"
+)
+
+type Interface interface {
```
chuckha: I avoided defining an interface here because this package can't know how the clients are going to be using the structs it defines.

For example, this interface will grow large and if we're using it in say, the machine actuator, the interface is going to be significantly larger than it needs to be.

vincepri: I see. The main problem I was facing, and why I moved it in here was because things can get out of sync frequently at this stage. I can change it back if required, or we can refactor later as things grow.

vincepri:
This should also allow us to use the aws client directly through the defined interface.

cloud/aws/services/ec2/interface.go
chuckha: it's more about being very specific about what the actuator needs. If we define the interface here then we're saying "all things that use this service will need all of these methods", but that is not true since the machine actuator and the cluster actuator are using different pieces of the same struct.

vincepri: I see what you mean. Would it make more sense then to define 2 structs/interfaces inside the ec2 package? If not, I'll revert it back.

chuckha: That's still defining the interface with the library which is my main concern. In this case because there is currently one client it will be ok for a while, but it will be annoying for other variants that want to reuse portions of this logic. Not defining the interface also allows us to define only one ec2 struct that clients can scope down themselves and thus we have less boilerplate to maintain. (n services, n interfaces)

Overall, I think the point I want to make is that the client knows how they want to use the methods we provide, the  library providing those methods does not.

vincepri:
@chuckha all done, ready for another review when you get a chance

vincepri:
@chuckha is this good to merge?

chuckha:
/approve

@detiber for lgtm

Issue #32, opened 2018-08-27
[machine actuator] Implement Delete


chuckha:

phyber:
I'm looking at implementing this. Will link a PR to it shortly.

phyber:
This may be closable now. I forgot to set an appropriate comment on my PR to close the issue automatically.

detiber:
@phyber indeed, thanks for working on this!

PR #80, opened 2018-09-17
machineactuator: Initial Delete implementation

phyber:
Connected to: #32 

Initial work on the machine actuator delete functionality.

 - [x] Added some `vim` temporary files to `.gitignore`
 - [x] Should be able to successfully call AWS API to delete an instance (machine).
 - [x] Basic tests

### DIFF
--- a/cloud/aws/actuators/machine/actuator.go
+++ b/cloud/aws/actuators/machine/actuator.go
@@ -20,6 +20,7 @@ import (
 	ec2svc "sigs.k8s.io/cluster-api-provider-aws/cloud/aws/services/ec2"
 
 	"github.com/golang/glog"
+	"github.com/pkg/errors"
 	"k8s.io/apimachinery/pkg/runtime"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 )
@@ -34,6 +35,7 @@ type machinesSvc interface {
 type ec2Svc interface {
 	CreateInstance(*clusterv1.Machine) (*ec2svc.Instance, error)
 	InstanceIfExists(*string) (*ec2svc.Instance, error)
+	TerminateInstance(*string) error
 }
 
 // codec are the functions off the generated codec that this actuator uses.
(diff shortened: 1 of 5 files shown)


cloud/aws/actuators/machine/actuator.go
```diff
@@ -109,7 +110,32 @@ func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 // Delete deletes a machine and is invoked by the Machine Controller
 func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	glog.Infof("Deleting machine %v for cluster %v.", machine.Name, cluster.Name)
-	return fmt.Errorf("TODO: Not yet implemented")
+
+	status, err := a.machineProviderStatus(machine)
+	if err != nil {
+		return err
```
chuckha: please wrap all of these errors with `errors.Wrap` from https://godoc.org/github.com/pkg/errors

This will allow us to trace the error if we ever see it.

Something like `return errors.Wrap(err, "failed to get machine provider status")`

cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -41,6 +42,14 @@ func (e *ec2) InstanceIfExists(id *string) (*ec2svc.Instance, error) {
 	return nil, nil
 }
 
+func (e *ec2) TerminateInstance(instanceID *string) error {
```
chuckha: this ec2 object needs to be replaced by the ec2 mock we're using like this: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cloud/aws/services/ec2/subnets_test.go#L31

But I'm ok if you want to leave as is. We can do this work later.

chuckha:
/ok-to-test

also please squash all commits into one commit before the final merge

phyber:
Thanks for the feedback, I should have those issues taken care of some time tomorrow :)

cloud/aws/actuators/machine/actuator_test.go
phyber: This isn't a direct EC2 call, but a wrapper for the real `ec2.TerminateInstances` in the same way that `CreateInstance` wraps `ec2.RunInstances`. I'm not sure if that changes what needs to be done here.

I think for this work right now, I'd be happier having this get merged and then working on improving the testing in a separate PR, as I think it will take me a little time to get up to speed on proper mocking.

phyber:
The errors have now been wrapped in the `Delete` method and some testing has been added for the `TerminateInstance` method.

If there are no further changes requested, I'll squash this into a single commit and push that before the merge.

chuckha:
@phyber lgtm, just squash this down and i'll get this merged!

roberthbailey:
FYI there is a setting in the mergebot to auto-squash when merging PRs.

phyber:
These commits have now been squashed down and should be good to merge.

PR #81, opened 2018-09-18
NAT Gateway reconciliation

### LINKED ISSUE #35: [aws] Add NAT gateway management
The MVP will require the ability to idempotently manage a NAT gateway for the private subnet(s)

randomvariable:
Fixes #35 

Roughly, the approach will be for creation (after subnet reconciliation):

1. Loop over all subnets in provider config
2. Exit if all subnets are public
3. Loop over public subnets and ensure there is A NAT gateway managed by us in there.
4. Loop over routing tables and ensure one exists for each AZ corresponding to the NAT gateway
5. Loop over private subnets
  a. Check the route table is the one we manage
  b. If not, associate it

### DIFF
--- a/cloud/aws/services/ec2/natgateways.go
+++ b/cloud/aws/services/ec2/natgateways.go
@@ -1,0 +1,96 @@
+// Copyright © 2018 The Kubernetes Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package ec2
+
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/pkg/errors"
+	"sigs.k8s.io/cluster-api-provider-aws/cloud/aws/providerconfig/v1alpha1"
+)
+
+func (s *Service) reconcileNatGateways(subnets v1alpha1.Subnets, vpc *v1alpha1.VPC) error {
+	if len(subnets.FilterPrivate()) == 0 {
+		return nil
+	}
+
+	existing, err := s.describeNatGatewaysBySubnet(vpc.ID)
+	if err != nil {
+		return err
+	}
+
+	for _, sn := range subnets.FilterPublic() {
+		if sn.ID == "" {
+			continue
+		}
+		if _, ok := existing[sn.ID]; ok {
+			continue
+		}
+
+		_, err := s.createNatGateway(sn.ID)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Service) describeNatGatewaysBySubnet(vpcID string) (map[string]*ec2.NatGateway, error) {
+	describeNatGatewayInput := ec2.DescribeNatGatewaysInput{
+		Filter: []*ec2.Filter{
+			{
+				Name:   aws.String("vpc-id"),
+				Values: []*string{aws.String(vpcID)},
+			},
+		},
+	}
+
+	gateways := make(map[string]*ec2.NatGateway)
+
+	err := s.EC2.DescribeNatGatewaysPages(
+		&describeNatGatewayInput,
+		func(page *ec2.DescribeNatGatewaysOutput, lastPage bool) bool {
+			for _, r := range page.NatGateways {
+				gateways[*r.SubnetId] = r
+			}
+			return !lastPage
+		})
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to describe NAT gateways with VPC ID %q", vpcID)
+	}
+	return gateways, nil
+}
+
+func (s *Service) createNatGateway(subnetID string) (*ec2.NatGateway, error) {
+
+	ip, err := s.allocateAddress()
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create IP address for NAT gateway for subnet ID %q", subnetID)
+	}
+
+	out, err := s.EC2.CreateNatGateway(&ec2.CreateNatGatewayInput{
+		SubnetId:     aws.String(subnetID),
+		AllocationId: aws.String(ip),
+	})
+
+	err = s.EC2.WaitUntilNatGatewayAvailable(&ec2.DescribeNatGatewaysInput{
+		NatGatewayIds: []*string{out.NatGateway.NatGatewayId},
+	})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create NAT gateway for subnet ID %q", subnetID)
+	}
+
+	return out.NatGateway, nil
+}
--- a/cloud/aws/services/ec2/eips.go
+++ b/cloud/aws/services/ec2/eips.go
@@ -13,28 +13,20 @@
 
 package ec2
 
-import "sigs.k8s.io/cluster-api-provider-aws/cloud/aws/providerconfig/v1alpha1"
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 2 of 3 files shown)


cloud/aws/services/ec2/natgateway.go
```diff
@@ -0,0 +1,126 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to create NAT gateway for subnet ID %q", subnetID)
+	}
+
+	return final, nil
+}
+
+func (s *Service) waitUntilNatGatewayAvailable(gatewayID string) (*ec2.NatGateway, error) {
+
+	err := s.ec2.WaitUntilNatGatewayAvailable(&ec2.DescribeNatGatewaysInput{
```
vincepri: Do we need a helper method for this?

cloud/aws/services/ec2/natgateway.go
```diff
@@ -0,0 +1,126 @@
...
+}
+
+func (s *Service) findNatGatewayByID(gatewayID string) (*ec2.NatGateway, error) {
+	describeNatGatewayInput := ec2.DescribeNatGatewaysInput{
+		NatGatewayIds: []*string{&gatewayID},
+	}
+
+	gateway := &ec2.NatGateway{}
+
+	err := s.ec2.DescribeNatGatewaysPages(
```
vincepri: We could probably use the simple `DescribeNatGateways` method here, given that we're using an id.

cloud/aws/services/ec2/natgateway.go
```diff
@@ -0,0 +1,126 @@
...
+		})
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to describe NAT gateways with VPC ID %q", vpcID)
+	}
+	return gateways, nil
+}
+
+func (s *Service) allocateAddress() (string, error) {
+	out, err := s.ec2.AllocateAddress(&ec2.AllocateAddressInput{})
```
vincepri: By default the address is allocated in `classic` ec2 mode. The `AllocateAddressInput` needs `Domain: aws.String(ec2.DomainTypeVpc)` to make sure the EIP is available in VPC mode.

cloud/aws/services/ec2/natgateway.go
```diff
@@ -0,0 +1,126 @@
...
+			return false
+		})
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to describe NAT gateways with VPC ID %q", vpcID)
+	}
+	return gateways, nil
+}
+
+func (s *Service) allocateAddress() (string, error) {
```
vincepri: Should this method be moved somewhere else? I'm not totally sure if the gateway file is the best place for it.

cloud/aws/services/ec2/natgateway.go
```diff
@@ -0,0 +1,126 @@
...
+
+		_, err := s.createNatGateway(sn.ID)
+		if err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func (s *Service) describeNatGateways(vpcID string) (map[string]*ec2.NatGateway, error) {
```
vincepri: Little nitpick around naming: describeNATGatewaysBySubnet is more explicit for users. For example, I had to read through the code to understand that the key of the map was the subnetId.

cloud/aws/services/ec2/natgateway.go
randomvariable: Wasn't sure either, when I wrote it earlier, so it's been inlined.

cloud/aws/services/ec2/natgateway.go
randomvariable: Thanks for that. Amended

cloud/aws/services/ec2/natgateway.go
randomvariable: Have renamed. Thanks.

cloud/aws/services/ec2/natgateway.go
randomvariable: It's been moved. Was there for the convenience of not opening a new file at the time.

cloud/aws/services/ec2/natgateway.go
randomvariable: Totally wasn't thinking there. Have replaced.

cloud/aws/services/ec2/natgateways.go
```diff
@@ -0,0 +1,110 @@
...
+
+	for _, sn := range subnets {
+		if sn.ID == "" || !sn.IsPublic {
+			continue
+		}
+		if _, ok := existing[sn.ID]; ok {
+			continue
+		}
+
+		_, err := s.createNatGateway(sn.ID)
```
vincepri: Do we want to create a new nat gateway for each public subnet? Or one per AZ?

cloud/aws/services/ec2/natgateways.go
```diff
@@ -0,0 +1,110 @@
...
+	})
+
+	{
+		err := s.ec2.WaitUntilNatGatewayAvailable(&ec2.DescribeNatGatewaysInput{
+			NatGatewayIds: []*string{out.NatGateway.NatGatewayId},
+		})
+		if err != nil {
+			return nil, errors.Wrapf(err, "failed to create NAT gateway for subnet ID %q", subnetID)
+		}
+	}
```
vincepri: Do we need the nested inner block above?

cloud/aws/services/ec2/natgateways.go
```diff
@@ -0,0 +1,110 @@
...
+func (s *Service) findNatGatewayByID(gatewayID string) (*ec2.NatGateway, error) {
+	out, err := s.ec2.DescribeNatGateways(&ec2.DescribeNatGatewaysInput{
+		NatGatewayIds: []*string{&gatewayID},
+	})
+
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to describe NAT gateway with ID %q", gatewayID)
+	}
+
+	return out.NatGateways[0], nil
```
vincepri: I believe we'd need to check bounds here, and return a `NotFound` error in case the slice is empty. See https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cloud/aws/services/ec2/gateways.go#L71-L73

cloud/aws/services/ec2/natgateways.go
randomvariable: For MVP, I think we should pick sensible defaults for subnet size, and then if you want to add a subnet to an existing AZ within the same cluster, it's because of scalability - and then you probably want an additional NAT gateway on top anyway.

I imagine there to be more variations down the lines. Within our customer base, they vary from:

* Using EC2 instead of NAT Gateway for to reduce data transfer cost purposes.
* Transparent proxying, with a mix of block & allow lists, TLS intercept, ClamAV scanning etc... (try to push back on this stuff as much as possible, but don't always win out)
* DNS intercept

These EC2-based gateways can be handled with passing in an AMI ID and starting a machine appropriately, but this relates to #88 .

cloud/aws/services/ec2/natgateways.go
randomvariable: No, not really.

cloud/aws/services/ec2/natgateways.go
randomvariable: Thanks. Added.

cloud/aws/services/ec2/natgateways.go
vincepri: I share the sentiment of picking sensible defaults. What do we think about creating only a single nat gateway for now and share it across the subnets in the same AZ? This would be good to track in a separate issue for long term plans. 

MVP is single AZ, see https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/docs/proposal/mvp-architecture.md.

vincepri: Tracked here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/89

vincepri:
@randomvariable This is looking good, last thing that's missing are unit tests. We've been using gomock, you can see an example here https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cloud/aws/services/ec2/instances_test.go

detiber:
@randomvariable I believe one of the recently merged PRs may have broken this, based on the automated test errors.

randomvariable:
@detiber Two things. Interface got renamed, so need to rebase, and also #93 

https://www.youtube.com/watch?v=6g4dkBF5anU

PR #91, opened 2018-09-20
machineactuator: Initial Update implementation

### LINKED ISSUE #33: [machine actuator] Implement Update

phyber:
Closes #33 

Initial implementation of the Machine Actuator Update method.

 - [x] Implement Update method
 - [x] Basic testing

I think this is correct. I've made the assumption that the method is being passed the new `Machine` object and that we want to extract the new `Status` from that and pass that to the `updateStatus` method.

Basic testing of the happy path included.

### DIFF
--- a/cloud/aws/actuators/machine/actuator.go
+++ b/cloud/aws/actuators/machine/actuator.go
@@ -143,7 +143,24 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 // Update updates a machine and is invoked by the Machine Controller
 func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	glog.Infof("Updating machine %v for cluster %v.", machine.Name, cluster.Name)
-	return fmt.Errorf("TODO: Not yet implemented")
+
+	// Handling of machine config changes is not yet implemented.
+	// We should check which pieces of configuration have been updated, throw
+	// errors if an attempt is made to modify any immutable state, otherwise
+	// go ahead and modify what we can.
+
+	// Get the new status from the provided machine object.
+	status, err := a.machineProviderStatus(machine)
+	if err != nil {
+		return errors.Wrap(err, "failed to get machine status")
+	}
+
+	err = a.updateStatus(machine, status)
+	if err != nil {
+		return errors.Wrap(err, "failed to update machine status")
+	}
+
+	return nil
 }
 
 // Exists test for the existence of a machine and is invoked by the Machine Controller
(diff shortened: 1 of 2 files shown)


cloud/aws/actuators/machine/actuator.go
```diff
@@ -142,7 +142,19 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 // Update updates a machine and is invoked by the Machine Controller
 func (a *Actuator) Update(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
 	glog.Infof("Updating machine %v for cluster %v.", machine.Name, cluster.Name)
-	return fmt.Errorf("TODO: Not yet implemented")
+
```
detiber: I think this approach to Update is sufficient to start, but it will probably be a good idea to add a comment to state that mutating the machine based on changes to the config, or handling machine upgrades is still an outstanding todo.

phyber: Would a full update of the machine object be the thing to do here, then to update the entire thing at once? Something that eventually calls `s.Machine.Update(machine)` in the style of this? https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cloud/aws/services/cluster/service.go#L25

detiber: I believe there are few different use cases that will need to be taken into account and we would need to decide how we want to handle them as part of the implementation:

- User tries to mutate attributes of the machine that cannot be done in place
  - either fail to update, bubbling up appropriate error message to Machine object
  - or potentially delete and recreate the machine
    - would need to take into account if the machine is a control plane node or not
    - would need to be able to attempt to gracefully cordon/drain the node prior to removing the existing machine.
    - would need to take into account if/how to recover in the case of failure
- User tries to mutate attributes of the machine that can be done in place, such as instance size
  - attempt to gracefully cordon/drain the node prior to shutting the instance down and mutating the instance before restarting and uncordoning.
  - this would also need to take into account if/how to recover in the case of failure
- User attempts to perform an upgrade on the machine
  - the workflow(s) for this would need to be designed, scoped, and implemented

phyber:
Based on comments from @detiber and talking with @randomvariable, there's a bit more here than I initially realised.
Currently, I think I'd like to merge this naive implementation of `Update` before implementing some more complicated logic to mutate machine config.

Initially, that mutation will probably cover Tags and Security Groups, while throwing errors if attempts to change anything else are made.

phyber:
Great. I was working on the Tag and SecurityGroup updating earlier. Once this is merged, I'll have a new WIP PR ready.

phyber:
The flaky test here seems to be an issue resolving `github.com`, hopefully this retest fixes it.

PR #96, opened 2018-09-21
[WIP] Add kubeadm cloud-init to image generation

### LINKED ISSUE #16: Cloud-init module
todo - fill in here 
/assign @chuckha

detiber:
Adds support for a kubeadm cloud-init module to image generation.

Resolves #16 

Depends on #133 #132 #115


cloud/aws/services/ec2/instances.go
```diff
@@ -69,8 +72,57 @@ func (s *Service) InstanceIfExists(instanceID *string) (*Instance, error) {
 }
 
 // CreateInstance runs an ec2 instance.
-func (s *Service) CreateInstance(machine *clusterv1.Machine) (*Instance, error) {
-	input := &ec2.RunInstancesInput{}
+func (s *Service) CreateInstance(machine *clusterv1.Machine, providerConfig *providerconfigv1.AWSMachineProviderConfig, cluster *clusterv1.Cluster, clusterProviderConfig *providerconfigv1.AWSClusterProviderConfig) (*Instance, error) {
```
detiber: @chuckha so, it dawned on me that not only do we need the machine information here, but we also need the cluster information as well to get the config exposed through the common ClusterNetworkingConfig, but we'll also need to get common per-cluster config to expose things like common cluster tags, or the generated security groups.

PR #97, opened 2018-09-21
Add code to handle routing tables

vincepri:


cloud/aws/services/ec2/routetables.go
```diff
@@ -21,10 +21,69 @@ import (
...
+
+	for _, sn := range in.Subnets {
+		if _, ok := subnetRouteMap[sn.ID]; ok {
+			// TODO(vincepri): if the route table ids are both non-empty and they don't match, replace the association.
+			// TODO(vincepri): check that everything is in order, e.g. routes match the subnet type.
+			continue
+		}
+
+		// For each subnet that doesn't have a routing table associated with it,
+		// create a new table with the appropriate default routes and associate it to the subnet.
```
detiber: How would this work as we extend to multi-az deployments? This comments seems to indicate that we would have a route table per-subnet rather than a route-table per public/private for the VPC.

cloud/aws/services/ec2/natgateways.go
```diff
@@ -85,12 +89,43 @@ func (s *Service) createNatGateway(subnetID string) (*ec2.NatGateway, error) {
...
+	if err := s.EC2.WaitUntilNatGatewayAvailable(wReq); err != nil {
+		return nil, errors.Wrapf(err, "failed to wait for nat gateway %q in subnet %q", *out.NatGateway.NatGatewayId, subnetID)
+	}
+
 	return out.NatGateway, nil
 }
+
+func (s *Service) getNatGatewayForSubnet(subnets v1alpha1.Subnets, sn *v1alpha1.Subnet) (string, error) {
+	if sn.IsPublic {
+		return "", errors.Errorf("cannot get NAT gateway for public subnet %q", sn.ID)
```
detiber: If I'm not mistaken, isn't the NAT gateway actually associated with the public subnet here though? I would expect that we would be interfacing directly with routes only for the private subnets.

vincepri: I should probably improve the error message or the method name, the method is intended to use only with private subnets, the NAT actually lives in a public subnet. Does that make sense?

cloud/aws/services/ec2/routetables.go
vincepri: This part should support multi-az deployments just fine I believe. Even with a routing table for each subnet, on creation, we'd pick a NAT gateway to associate using `getNatGatewayForSubnet`, which should follow some rules to pick the best gateway (i.e. same AZ, number of subnets already associated, etc).

cloud/aws/providerconfig/v1alpha1/types.go
```diff
@@ -178,17 +178,28 @@ type VPC struct {
...
+	IsPublic         bool    `json:"public"`
+	RouteTableID     *string `json:"routeTableId"`
+	NatGatewayID     *string `json:"natGatewayId"`
 }
 
 // Subnets is a slice of Subnet.
 type Subnets []*Subnet
 
+// ToMap returns a map from id to subnet.
+func (s Subnets) ToMap() (res map[string]*Subnet) {
```
detiber: I might be missing something, since I'm still relatively new to Go, but wouldn't this cause issues due to the default value for a map being nil?

cloud/aws/providerconfig/v1alpha1/types.go
```diff
@@ -178,17 +178,28 @@ type VPC struct {
...
+	for _, x := range s {
+		res[x.ID] = x
+	}
+
+	return
+}
+
 // FilterPrivate returns a slice containing all subnets marked as private.
-func (s Subnets) FilterPrivate() (res []*Subnet) {
+func (s Subnets) FilterPrivate() (res Subnets) {
```
detiber: Similar question here as well.

cloud/aws/providerconfig/v1alpha1/types.go
vincepri: Thanks for catching that! That's totally right

cloud/aws/providerconfig/v1alpha1/types.go
vincepri: That's true for the map, but append() will allocate the array and re-assign it, so this is safe.

detiber: Good to know, thanks!

vincepri:
@detiber gonna add some tests before removing the WIP

vincepri:
@detiber ptal

cloud/aws/services/ec2/routetables.go
```diff
@@ -21,10 +21,69 @@ import (
 )
 
 func (s *Service) reconcileRouteTables(in *v1alpha1.Network) error {
```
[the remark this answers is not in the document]
vincepri: The errors returned from this method are already wrapped in the inner functions, did I miss some?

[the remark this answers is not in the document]
chuckha: ahhh i see gotcha, thanks

PR #98, opened 2018-09-24
Default network configuration

johanneswuerbach:
When running through the setup described [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws#running-clusterctl) I encountered a NPE as the examples don't set any network configuration.

This sets some empty values if nothing is defined, which triggers the defaults to be used.


vincepri:
Thanks for the contribution and welcome! 

Should we move the defaults to the cluster actuator reconcile method? Ideally that method should be ultimately responsible to gather / store information about the cluster status.

chuckha:
This is actually a bug with the types we've created. Instead of doing these nil checks we are going to be getting rid of the pointer types in the v1alpha1/types.go file

vincepri:
#106 removes pointers, which should remove the need for defaults.

/close

PR #100, opened 2018-09-24
Example improvements

johanneswuerbach:
Happy to extract those into separate PRs, but as they were tiny in bundled those changes.

**Fix example spec**

When running `clusterctl` two deployments will be created `clusterapi-apiserver` and `clusterapi-controllers`. Both have the label `api: clusterapi`, so add another label to the controllers to match only controller pods (the apiserver deployment has already the additional label `apiserver: true`)

**Allow to override dev image tags**

I found it a bit cumbersome to manually edit the tag when recreating the controllers, instead `export DEV_TAG` can now be used to modify the tag.

**Clarify example generation**

I tried to run `./clusterctl/examples/aws/generate-yaml.sh`, which failed as the script can only be executed inside `./clusterctl/examples/aws`

### DIFF
--- a/clusterctl/examples/aws/provider-components.yaml.template
+++ b/clusterctl/examples/aws/provider-components.yaml.template
@@ -4,12 +4,18 @@ metadata:
   name: clusterapi-controllers
   labels:
     api: clusterapi
+    controllers: "true"
 spec:
   replicas: 1
+  selector:
+    matchLabels:
+      api: clusterapi
+      controllers: "true"
   template:
     metadata:
       labels:
         api: clusterapi
+        controllers: "true"
     spec:
       nodeSelector:
         node-role.kubernetes.io/master: ""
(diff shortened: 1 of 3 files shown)


chuckha:
Thanks for the fixes! Could you explain the label issue a little bit more? I haven't run into any problems with the labels from the example yaml

johanneswuerbach:
The issue appears when you manually delete the pod as it seems to not be recreated as it matches the 1 pod of the apiserver deployment. Assumbly modifying the deployment might also have unwanted side effects

README.md
```diff
@@ -62,7 +62,10 @@ You will also need to configure minikube or your bootstrap cluster with credenti
 or another mechanism.
 
 #### Running clusterctl
-3. Generate the input files with [clusterctl/examples/aws/generate-yaml.sh](/clusterctl/examples/aws/generate-yaml.sh)
+3. Generate the input files with [clusterctl/examples/aws/README.md](/clusterctl/examples/aws/README.md)
```
chuckha: I think we have some updates that improve this workflow

PR #101, opened 2018-09-24
Fix NPE when deleting not yet created machine

johanneswuerbach:
`InstanceIfExists` can also return `nil` if the call succeeded, but no instance has been found.

This allows to delete not yet started machines.


chuckha:
oops

/lgtm

PR #104, opened 2018-09-25
machineactuator: Update Tags and SecurityGroups implementation

phyber:

Implements further mechanics of the `Update` method for the machine actuator.

Related to #33

Implements:
  - [x] Addition/Update/Deletion of Tags the machine actuator is responsible for
  - [x] Addition/Update/Deletion of Security Groups the machine actuator is responsible for
  - [x] Get/Set of specific machine annotations
  - [x] Get/Set of specific machine annotations that use unstructured JSON/maps.
  - [x] Removal of actuator tests as discussed with @randomvariable, @detiber, @chuckha 

- Does not implement:
  - Errors on attempts to update immutable state, will be a separate PR after this since this PR is already large.

Although related to #33, it doesn't close the issue yet due to the immutable state note above.

**Release note**:

```
None
```


phyber:
Currently implementing the tracking of Tags and SecurityGroups that are being handled by the machine actuator, which will allow us to properly clean up our resources if they disappear from the `Additional{SecurityGroups,Tags}` objects.

Tracking will be done via annotations on the machine object.

phyber:
I have now implemented some more complicated mechanics here, which will allow us to properly `Update` the bits of the machine object that we are responsible for.

  - Methods to get/set specific machine annotation
  - Methods to get/set specific machine annotation using unstructured JSON/maps
  - Improved handling of the AdditionalSecurityGroups in the machine config 
    - This compares the incoming `AdditionalSecurityGroups` to an annotation of what we did last time `Update` was called, allowing us to work out if any of our `AdditionalSecurityGroups` have been dropped and should be deleted. It does this without disturbing other security groups that we are not managing by ignoring any security groups that are present in AWS but were not set in our annotation or `AdditionalSecurityGroups`.

phyber:
I'm guessing @chuckha and @detiber may want to look at this (based on previous comments/reviews). It's kind of complicated. I've done my best to ensure that the comments accurately describe what's going on, but I'm kind of fried now :)

If it looks OK, I'll do similar for Tags. I'm also aware that
  - Some of these methods may be better suited to some other files, if so, just say where
  - That there is another PR implementing a bunch of stuff for dealing with tags. I'm happy to update this PR to use those methods if/when they become available.

cloud/aws/actuators/machine/actuator.go
```diff
@@ -141,21 +152,76 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+		if err != nil {
+			return errors.Wrap(err, "failed to update security group annotation")
+		}
+	}
+
+	// Check if the instance tags were changed. If they were, update them.
+	// It would be possible here to only send new/updated tags, but for the
+	// moment we send everything, even if only a single tag was created or
+	// upated.
+	tagsChanged := a.tagsChanged(config.AdditionalTags, instanceDescription.Tags)
```
detiber: Is the plan to update tag reconciliation with an annotation-based approach as well?

cloud/aws/actuators/machine/actuator.go
```diff
@@ -141,21 +152,76 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+		if err != nil {
+			return errors.Wrap(err, "failed to update network interface security groups")
+		}
+
+		// And then update the annotation to reflect our new managed state.
+		err = a.updateMachineAnnotationJSON(machine, SecurityGroupsLastAppliedAnnotation, sgAnnotation)
+		if err != nil {
+			return errors.Wrap(err, "failed to update security group annotation")
+		}
+	}
```
detiber: I'd prefer to see the security group reconciliation broken out into a separate method

cloud/aws/actuators/machine/actuator.go
```diff
@@ -141,21 +152,76 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+
+	// Now we check to see if we need to mutate any mutable state.
+	// Check if any additions have been made to the instance Security Groups.
+	sgChanged, sgIDs, sgAnnotation := a.securityGroupsChanged(
+		securityGroupAnnotation,
+		config.AdditionalSecurityGroups,
+		instanceDescription.SecurityGroups)
+	if sgChanged {
+		// Finally, update the instance with our new security groups.
+		err := a.ec2.UpdateNetworkInterfaceSecurityGroups(&instanceDescription.PrimaryInterface, sgIDs)
```
detiber: I'm curious why updating security groups on the primary interface here rather than setting the security groups on the instance itself?

cloud/aws/actuators/machine/actuator.go
```diff
@@ -188,6 +254,269 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+//    we add it to the map.
+//  - If we find a groupID in AWS that's in our state and flagged for deletion,
+//    we do not add it to the map and we allow it to drop.
+//
+// Next we compare the map we have made after comparing our desired state and
+// the AWS state against AWS again to decide if any changes actually happened.
+//
+// Finally, we return three items listed at the top of this comment.
+// The caller then acts upon this information.
+func (a *Actuator) securityGroupsChanged(annotation map[string]interface{}, src []v1alpha1.AWSResourceReference, dst map[string]string) (bool, []*string, map[string]interface{}) {
```
detiber: @chuckha @vincepri thoughts on the approach here? The lack of set theory primitives in Go drives me batty sometimes and I haven't internalized the idiomatic way of handling it yet.

cloud/aws/actuators/machine/actuator.go
phyber: Correct, however I have stopped on the tags part for the moment since there's another PR which looks like it's implementing a lot of tag helper methods. Once that is merged, I will resume working on the tags and use those new methods.

cloud/aws/actuators/machine/actuator.go
phyber: Reasonable. I can move that out into a separate private method simply enough.

cloud/aws/actuators/machine/actuator.go
phyber: I'd been referencing some of our code which turned out to be a bit legacy. I'll make the change here to apply the SecurityGroups directly to the Instance.

phyber:
That test failure is legit. I've got it fixed up enough locally that Go panics, currently working out how to satisfy the mock to prevent that.

phyber:
The `TestUpdate` is now working properly and tests the most basic happy path. I am currently implementing the management of the instance tags. Once that's complete, additional testing will be added to test the updating of security groups and tags.

phyber:
I think the `ensureTags` and `ensureSecurityGroups` methods here reflect what's needed now. Work on this tomorrow will mostly be adding new tests around the updating/deleting of security groups.

phyber:
OK, I think this is ready for a proper review. I think everything that I want is in here for now as the PR is very large at this point.

The part I'm most unsure about is the `updateMachine` method (called after annotations on the machine object have been updated). It feels like perhaps I should be setting some kind of subresource here instead of the entire object, but I don't know what that subresource would be.

cloud/aws/actuators/machine/actuator.go
vincepri: I think this is mostly personal preference. Usually I prefer using `map[string]struct{}`. The empty struct takes 0 bytes and checking if an item is in the map is as easy as `if _, ok := set["test"]`. The actual value in the map can be ignored at that point, which I find nice.

cloud/aws/actuators/machine/actuator.go
```diff
@@ -188,6 +248,401 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+
+	// We made it through the loop, and everything that was in src, was also
+	// in dst. Nothing changed.
+	return changed, created, deleted, newAnnotation
+}
+
+// Returns a map[string]interface from a JSON annotation.
+// This method gets the given `annotation` from the `machine` and unmarshalls it
+// from a JSON string into a `map[string]interface{}`.
+func (a *Actuator) machineAnnotationJSON(machine *clusterv1.Machine, annotation string) (map[string]interface{}, error) {
```
chuckha: What is the logic behind using annotations? could querying ec2 replace annotations?

randomvariable: Because we need to track the history of tags / sgs that were applied by this controller as opposed to other things. See discussion in [implementers' office hours](https://docs.google.com/document/d/1IZ2-AZhe4r3CYiJuttyciS7bGZTTx4iMppcA8_Pr3xE/edit#heading=h.fb006hyr6m70) and Slack [1](https://kubernetes.slack.com/archives/C8TSNPY4T/p1537883775000100), [2](https://kubernetes.slack.com/archives/C8TSNPY4T/p1537886836000100)

cloud/aws/actuators/machine/actuator.go
phyber: Good point, I'd forgotten about the empty struct. I'll make that change where appropriate.

phyber: Where possible, maps now use an empty struct if they don't care about the map value.

phyber:
- Rebased against recent merges
 - Tests fixed up based on changes
 - Now using empty structs where we don't care about `map` values when working out security group changes

cloud/aws/actuators/machine/actuator.go
phyber: As @randomvariable says, we can't easily query AWS for this information as we need to know:
  - The incoming new state that the machine actuator `Update` method is acting on
  - The current state of the Tags and SecurityGroups as AWS knows it (`DescribeInstances` via `InstanceIfExists`)
  - The state as we left it last time the `Update` method was called (the annotation)

From these, we work out what we have been, and what we are currently responsible for, letting us know which Tags or SecurityGroups we can delete if they disappear from the `AdditionalTags` or `AdditionalSecurityGroups` fields. 

This is due to the machine actuator not being responsible for the total state of the machine Tags and SecurityGroups (supporting comments for this are in the [`AWSMachineProviderConfig`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/8ece4cacc094348080001c4ee6791097c5080308/cloud/aws/providerconfig/types.go#L27) type.

cloud/aws/actuators/machine/actuator.go
```diff
@@ -183,6 +243,408 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+	u := unstructured.Unstructured{}
```
chuckha: I don't think we need unstructured in here. This is unmarshalling json into a map[string]interface{} and we could do without the extra dependency

```go
func (a *Actuator) machineAnnotationJSON(machine *clusterv1.Machine, annotation string) (map[string]interface{}, error) {
	out := map[string]interface{}{}
	jsonAnnotation := machine.GetAnnotations()[annotation]

	if len(jsonAnnotation) == 0 {
		return out, nil
	}

	// We have some kind of string, attempt to unmarshal it.
	err := json.Unmarshal([]byte(jsonAnnotation), &out)
	return out, err
}
```

cloud/aws/actuators/machine/actuator.go
```diff
@@ -183,6 +243,408 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+		err = a.updateMachineAnnotationJSON(machine, TagsLastAppliedAnnotation, newAnnotation)
+		if err != nil {
+			return false, err
+		}
+	}
+
+	return changed, nil
+}
+
+// Arguments:
```
chuckha: This comment is quite verbose and going to be hard to keep up to date. Go doc will do a good job with just a function signature and a few lines of explaining text. I think this could be reduced to something like:

```go
// securityGroupsChanged determines which security groups to delete and which to add.
```

The detail is nice but it's really hard to maintain comments like this

cloud/aws/actuators/machine/actuator.go
```diff
@@ -183,6 +243,408 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+//    we add it to the map.
+//  - If we find a groupID in AWS that's in our state and flagged for deletion,
+//    we do not add it to the map and we allow it to drop.
+//
+// Next we compare the map we have made after comparing our desired state and
+// the AWS state against AWS again to decide if any changes actually happened.
+//
+// Finally, we return three items listed at the top of this comment.
+// The caller then acts upon this information.
+func (a *Actuator) securityGroupsChanged(annotation map[string]interface{}, src []v1alpha1.AWSResourceReference, dst map[string]string) (bool, []*string, map[string]interface{}) {
```
chuckha: I think we will want a set package. It's ok to put this off for now, but the bulk of this function should only be a few lines doing set difference and intersection and such.

cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -249,16 +249,95 @@ func TestUpdate(t *testing.T) {
...
+					InstanceType: aws.String("m5.large"),
+					SubnetId:     aws.String("subnet-1"),
+					ImageId:      aws.String("ami-1"),
 				},
 			},
-		}).
-		Return(&clusterv1.Machine{}, nil)
+		}, nil)
+
+	gomock.InOrder(
```
chuckha: I don't see any tags trying to be updated here, please add a test to test the logic written in the actuator

cloud/aws/actuators/machine/actuator.go
```diff
@@ -183,6 +243,408 @@ func (a *Actuator) Exists(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+	}
+
+	return changed, nil
+}
+
+// Ensure that the tags of the machine are correct
+// Returns bool, error
+// Bool indicates if changes were made or not, allowing the caller to decide
+// if the machine should be updated.
+func (a *Actuator) ensureTags(machine *clusterv1.Machine, instanceID *string, additionalTags map[string]string) (bool, error) {
```
chuckha: Generally I'd like to keep the actuator small and put this kind of logic in some other object or package. Maybe a tags manager or similar

cloud/aws/actuators/machine/actuator.go
randomvariable: Am generally in favour of over commenting vs. under commenting, at least until we've properly documented our implementation decisions and assumptions that affect consumers of cluster-api-provider-aws.

randomvariable:
Would have been nice to have had these objections much earlier in the process rather than having expended engineering effort against a plan that was communicated in advance.

cloud/aws/providerconfig/v1alpha1/types.go
```diff
@@ -368,5 +368,8 @@ type Instance struct {
 	EBSOptimized *bool `json:"ebsOptimized"`
 
 	// The tags associated with the instance.
-	Tags map[string]string `json:"tag"`
+	Tags map[string]string `json:"tags"`
```
phyber: I'm not sure if this wanted to be `json:"tags"` (which I'd set in my PR) or `json:"tag"` that someone elses PR had set. Given that the rest of the fields had their JSON field name set to the same as the struct field name, I opted for `tags`.

cloud/aws/services/ec2/instances.go
```diff
@@ -199,10 +199,72 @@ func (s *Service) runInstance(i *v1alpha1.Instance) (*v1alpha1.Instance, error)
...
+			return err
+		}
+	}
+
+	return nil
+}
+
+// fromSDKTypeToInstance takes a ec2.Instance and returns our v1.alpha1.Instance
+// type. EC2 types are wrapped or converted to our own types here.
+func (s *Service) fromSDKTypeToInstance(v *ec2.Instance) *v1alpha1.Instance {
```
phyber: I moved this function to a method on `Service` since it needed to access some of the `Service` helper methods for dealing with `Tags` and `SecurityGroups`.

cloud/aws/providerconfig/v1alpha1/types.go
vincepri: +1 Thanks for catching that

cloud/aws/services/ec2/instances.go
vincepri: Could we make `tagToMap` or `mapToTag` functions instead of methods?

phyber: We could. I could leave them in `tags.go` since that seems to be a good logical place for them, move them out of the `Service` struct and move `fromSDKTypeToInstance` back out of `Service`.

phyber: (If there is a better place that standalone functions like that belong, I could also move them there)

vincepri: Yeah I like that idea! For now it seems that `tags.go` is a good place to but those functions in. Later on, we can think about putting all these conversion functions in a different place for sure, we should probably open an issue to track that.

phyber: Done, they're moved out of `Service` and are standalone functions now. `fromSDKTypeToInstance` has been restored to its former functionness.

chuckha:
I think this just needs to rm the actuator tests and then we can track refactor work with another issue, like creating a set package and extracting this logic out to keep the actuator's Update function very tiny

phyber:
Actuator tests are gone, so this should be good to go.

cloud/aws/services/ec2/tags.go
```diff
@@ -86,3 +86,34 @@ func (s *Service) buildTags(clusterName string, lifecycle ResourceLifecycle, add
 
 	return tags
 }
+
+// tagsToMap converts a []*ec2.Tag into a map[string]string.
+func tagToMap(src []*ec2.Tag) map[string]string {
```
vincepri: Should the function name match the comment with plural "tags"?

cloud/aws/services/ec2/tags.go
```diff
@@ -86,3 +86,34 @@ func (s *Service) buildTags(clusterName string, lifecycle ResourceLifecycle, add
...
+
+	for _, t := range src {
+		tags[*t.Key] = *t.Value
+	}
+
+	return tags
+}
+
+// mapToTag converts a map[string]string to a []*ec2.Tag
+func mapToTag(src map[string]string) []*ec2.Tag {
```
vincepri: Same here, should we use "tags"?

phyber:
The `mapToTags` and `tagsToMap` functions are now using the plural of tags in their names.

detiber:
/approve

+1 to creating an issue to track the cleanup/refactor work discussed.

PR #105, opened 2018-09-25
Dx

chuckha:
Improve the developer experience

* adds a make target for the example.
* adds an environment file for generating the example yaml

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,3 +64,16 @@ fmt:
 
 vet:
 	go vet ./...
+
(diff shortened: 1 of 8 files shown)


detiber:
@chuckha one of my thoughts here have been to see if we can leverage the vscode Kubernetes extension to provide a decent DX for people that are in that ecosystem. Alternatively, we might be able to provide some type of `skaffold` based workflow.

PR #119, opened 2018-09-27
Fix subnet discovery, reconcile route tables, add logs

vincepri:

- Add much needed logs throughout the reconcile loop.
- Forces to store the state in the cluster actuator with defer.
- Fixes subnet discovery which caused the first iteration to fail.
- Actually reconcile routing tables.
- Complete cluster actuator tests.
- Fix NPE in route table association.

NONE

NONE

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/aws/providerconfig/v1alpha1/types.go
+++ b/cloud/aws/providerconfig/v1alpha1/types.go
@@ -14,6 +14,8 @@
 package v1alpha1
 
 import (
+	"fmt"
+
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
@@ -174,6 +176,11 @@ type VPC struct {
 	CidrBlock string `json:"cidrBlock"`
 }
 
+// String returns a string representation of the VPC.
+func (v *VPC) String() string {
+	return fmt.Sprintf("id=%s", v.ID)
+}
+
 // Subnet defines an AWS subnet attached to a VPC.
 type Subnet struct {
 	ID string `json:"id"`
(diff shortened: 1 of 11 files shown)


cloud/aws/services/ec2/gateways.go
```diff
@@ -16,11 +16,14 @@ package ec2
 import (
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"sigs.k8s.io/cluster-api-provider-aws/cloud/aws/providerconfig/v1alpha1"
 )
 
 func (s *Service) reconcileInternetGateways(in *v1alpha1.Network) error {
+	glog.Infof("Reconciling internet gateways")
```
detiber: We should probably require an increased verbosity to output for anything below the cluster itself to avoid spamming the logs. It would also be good to give additional context as to what we are attempting to reconcile since the controller has multiple workers.

vincepri:
@detiber Comments should be all resolved, ptal :)

cloud/aws/providerconfig/v1alpha1/types.go
```diff
@@ -174,6 +176,11 @@ type VPC struct {
 	CidrBlock string `json:"cidrBlock"`
 }
 
+// String returns a string representation of the VPC.
+func (v *VPC) String() string {
+	return fmt.Sprintf("id=%s", v.ID)
```
chuckha: What do you think about `fmt.Sprintf("%#v", v)` for Stringer functions?

vincepri: Wouldn't that be a little verbose? Especially for the subnets, but I'm +1 if that's preferred, I don't have strong opinions on this.

chuckha: i also don't care so let's roll with this and fix it if we need something else.

PR #132, opened 2018-09-27
create an instance

### LINKED ISSUE #53: [aws] Idempotent create instance

chuckha:

This PR starts the work of implementing the instances

Fixes #53 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -37,17 +37,27 @@ genmocks: depend
 	hack/generate-mocks.sh "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1 MachineInterface" "cloud/aws/actuators/machine/mock_machineiface/mock.go"
 	hack/generate-mocks.sh "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1 ClusterInterface" "cloud/aws/actuators/cluster/mock_clusteriface/mock.go"
 
-build: depend
-	CGO_ENABLED=0 go install -a -ldflags '-extldflags "-static"' sigs.k8s.io/cluster-api-provider-aws/cmd/cluster-controller
-	CGO_ENABLED=0 go install -a -ldflags '-extldflags "-static"' sigs.k8s.io/cluster-api-provider-aws/cmd/machine-controller
+build: depend clusterctl-bin cluster-controller machine-controller
+
+clusterctl-bin:
 	CGO_ENABLED=0 go install -a -ldflags '-extldflags "-static"' sigs.k8s.io/cluster-api-provider-aws/clusterctl
 
 images: depend
 	$(MAKE) -C cmd/cluster-controller image
 	$(MAKE) -C cmd/machine-controller image
 
-dev_push: depend
+dev_push: depend cluster-controller-dev-push machine-controller-dev-push
+
+cluster-controller:
+	CGO_ENABLED=0 go install -a -ldflags '-extldflags "-static"' sigs.k8s.io/cluster-api-provider-aws/cmd/cluster-controller
+
+cluster-controller-dev-push: cluster-controller
 	$(MAKE) -C cmd/cluster-controller dev_push
+
+machine-controller:
+	CGO_ENABLED=0 go install -a -ldflags '-extldflags "-static"' sigs.k8s.io/cluster-api-provider-aws/cmd/machine-controller
+
+machine-controller-dev-push: machine-controller
 	$(MAKE) -C cmd/machine-controller dev_push
 
 push: depend
--- a/cloud/aws/actuators/machine/actuator.go
+++ b/cloud/aws/actuators/machine/actuator.go
@@ -30,9 +30,10 @@ import (
 // ec2Svc are the functions from the ec2 service, not the client, this actuator needs.
 // This should never need to import the ec2 sdk.
 type ec2Svc interface {
-	CreateInstance(*clusterv1.Machine) (*ec2svc.Instance, error)
+	CreateInstance(*clusterv1.Machine, *v1alpha1.AWSMachineProviderConfig, *v1alpha1.AWSClusterProviderStatus) (*ec2svc.Instance, error)
 	InstanceIfExists(*string) (*ec2svc.Instance, error)
 	TerminateInstance(*string) error
+	CreateOrGetMachine(*clusterv1.Machine, *v1alpha1.AWSMachineProviderStatus, *v1alpha1.AWSMachineProviderConfig, *v1alpha1.AWSClusterProviderStatus) (*ec2svc.Instance, error)
 }
 
 // codec are the functions off the generated codec that this actuator uses.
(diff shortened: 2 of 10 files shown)


Makefile
```diff
@@ -37,17 +37,25 @@ genmocks: depend
 	hack/generate-mocks.sh "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1 MachineInterface" "cloud/aws/actuators/machine/mock_machineiface/mock.go"
 	hack/generate-mocks.sh "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1 ClusterInterface" "cloud/aws/actuators/cluster/mock_clusteriface/mock.go"
 
-build: depend
+build: depend clusterctl-bin
```
chuckha: These split up the build since I'm only iterating on one thing a time this is faster for me

cloud/aws/actuators/machine/actuator.go
```diff
@@ -66,42 +69,47 @@ type ActuatorParams struct {
...
 func NewActuator(params ActuatorParams) (*Actuator, error) {
+	e, ok := params.EC2Service.(*ec2svc.Service)
+	if !ok {
+		return nil, errors.New("EC2Service is not a real ec2service")
+	}
 	return &Actuator{
 		codec:          params.Codec,
 		ec2:            params.EC2Service,
 		machinesGetter: params.MachinesGetter,
+		infra: &infra.Infra{
```
chuckha: This is an experiment. I am playing around with a new idea. any feedback welcome

cloud/aws/actuators/machine/actuator.go
```diff
@@ -66,42 +69,47 @@ type ActuatorParams struct {
...
+			EC2Svc: e,
+		},
 	}, nil
 }
 
 // Create creates a machine and is invoked by the machine controller.
 func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine) error {
-	// will need this machine config in a bit
-	_, err := a.machineProviderConfig(machine.Spec.ProviderConfig)
+	status, err := a.machineProviderStatus(machine)
```
chuckha: such boiler plate

cloud/aws/actuators/machine/actuator.go
```diff
@@ -30,7 +32,7 @@ import (
 // ec2Svc are the functions from the ec2 service, not the client, this actuator needs.
 // This should never need to import the ec2 sdk.
 type ec2Svc interface {
-	CreateInstance(*clusterv1.Machine) (*ec2svc.Instance, error)
+	CreateInstance(*clusterv1.Machine, *v1alpha1.AWSMachineProviderConfig, *v1alpha1.AWSClusterProviderStatus) (*ec2svc.Instance, error)
```
detiber: I've been struggling with this on my cloud-init integration work as well. In addition to info that is currently stored in the cluster providerstatus, we also need some of the info that is available within the base cluster object as well. I think it would make sense to unify the machine and cluster info needed into a consolidated struct rather than continuing to add additional disparate structs.

cloud/aws/actuators/machine/actuator.go
```diff
@@ -66,42 +69,47 @@ type ActuatorParams struct {
...
+		return errors.Wrap(err, "failed to create or get machine")
 	}
 
 	status.InstanceID = &i.ID
 	status.InstanceState = &i.State
+	machine.Status.NodeRef = &corev1.ObjectReference{
+		Kind:      machine.Kind,
+		Namespace: machine.Namespace,
+		Name:      machine.Name,
+	}
```
detiber: This seems off to me, the NodeRef should link to a Kubernetes `Node` object and not a cluster-api `Machine`. It really only applies to the case of a `Node` that exists in the same cluster as the `Machine` currently.

cloud/aws/actuators/machine/actuator.go
detiber: Is this an attempt to separate out the business logic from the AWS wrapper?

chuckha: it is. What do you think?

cloud/aws/services/ec2/instances.go
```diff
@@ -33,6 +36,8 @@ const (
 
 	// InstanceStatePending indicates the instance is pending
 	InstanceStatePending = ec2.InstanceStateNamePending
+
+	defaultAMIID = "ami-057c58c0eca3e6fe3"
```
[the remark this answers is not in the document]
chuckha: 💯

cloud/aws/actuators/machine/actuator.go
detiber: I like the approach in general, just a matter of getting the naming and code structure to be more intuitive :)

cloud/aws/actuators/machine/actuator.go
chuckha: Yeah I was hella confused here

detiber: At some point it might be nice if we could extend this in a way that the node object could also be enriched with an apiserver endpoint/credentials so that it could live in a remote cluster to be tracked/synced...  Way out of scope for this PR though :rofl:

Makefile
```diff
@@ -37,17 +37,25 @@ genmocks: depend
...
 
 images: depend
 	$(MAKE) -C cmd/cluster-controller image
 	$(MAKE) -C cmd/machine-controller image
 
-dev_push: depend
+dev_push: depend cluster-controller-dev-push machine-controller-dev-push
+
+cluster-controller-dev-push:
+	CGO_ENABLED=0 go install -a -ldflags '-extldflags "-static"' sigs.k8s.io/cluster-api-provider-aws/cmd/cluster-controller
```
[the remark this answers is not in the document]
chuckha: for sure it does

cloud/aws/services/ec2/instances.go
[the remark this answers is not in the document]
vincepri: Another note here is that AMIs are region-specific, so this might end up failing in a different region.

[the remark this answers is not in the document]
chuckha: yeah, took that into account with the new logic. It's **very** bad right now though because the AMIs don't have tags to filter on, but it's fine for now.

cloud/aws/actuators/machine/actuator.go
chuckha: decision was to defer to after MVP work

cloud/aws/services/ec2/ami.go
```diff
@@ -0,0 +1,57 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ec2
+
+// DefaultAMILookup returns the default AMI based on region
+func (s *Service) defaultAMILookup(region string) string {
+	//TODO(chuckha) tags on our built images need to be copied to the public images
```
detiber: I'd update this comment to say that we cannot use tags for lookup here and that we need to determine a better method of filtering for public AMIs.

cloud/aws/services/ec2/instances.go
```diff
@@ -69,8 +72,32 @@ func (s *Service) InstanceIfExists(instanceID *string) (*Instance, error) {
 }
 
 // CreateInstance runs an ec2 instance.
-func (s *Service) CreateInstance(machine *clusterv1.Machine) (*Instance, error) {
-	input := &ec2.RunInstancesInput{}
+func (s *Service) CreateInstance(machine *clusterv1.Machine, config *v1alpha1.AWSMachineProviderConfig, clusterStatus *v1alpha1.AWSClusterProviderStatus) (*Instance, error) {
+	id := config.AMI.ID
+	if id == nil {
+		// TODO(chuckha) region is defined by the session which is bad for HA stuff
```
detiber: Arrgh, we need to bubble this up to a tracking issue, I think. Ideally, we'd be able to define the region on a per-cluster level, though some might also want to specify it at the machine level.

cloud/aws/services/ec2/ami.go
chuckha: I'll update it, ticket exists here https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/138

cloud/aws/services/ec2/instances.go
chuckha: do you think it's ok to defer machine-level-region feature for later or is this a requirement for MVP?
Regardless, issue is being tracked here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/144

detiber: I definitely don't think machine-level override is a requirement for MVP, I think cluster-level region is borderline at this point.

PR #133, opened 2018-09-28
Update image generation

detiber:
- Add kubeadm cloud-init to image generation
- Update base centos image to undocumented public AMI
- Attempt to add packer builder for Ubuntu Bionic
- Add support for Ubuntu Bionic
- Update default k8s version to v1.12
- Add list of pre-existing base AMIs to README


images/README.md
```diff
@@ -37,10 +37,10 @@ There are additional variables that may be set that affect the behavior of speci
 
 If packer build is run without specifying which images to build, then it will attempt to build all configured images. `packer inspect packer.json` will list the configured builders. The `--only` option can be specified when running `packer build` to limit the images built.
 
-For example, to build only the Ubuntu image:
+For example, to build only the Ubuntu Bionic image:
```
vincepri: Maybe we should add the version number too in this example?

images/ansible/roles/common/tasks/redhat.yml
```diff
@@ -16,6 +16,18 @@
...
+  get_url:
+    url: "{{ common_redhat_cloud_init_copr_rpm }}"
+    dest: /etc/yum.repos.d/cloud-init-copr.repo
+
+- name: add yum priority
+  lineinfile:
+    path: /etc/yum.repos.d/cloud-init-copr.repo
+    regexp: '^priority='
+    line: 'priority=1'
+  when: ansible_distribution == "Amazon"
```
vincepri: Is this required only on Amazon?

detiber: Yes, Amazon Linux adds an optional yum plugin for "priorities", that is a horrid mess to deal with.

images/README.md
detiber: Maybe, but we do discuss it in another section. It might be good to see if we can get some copy editing on our docs to improve the overall experience.

README.md
```diff
@@ -93,6 +93,56 @@ I0926 14:05:56.173557   15312 clusterdeployer.go:124] Creating master  in namesp
 I0926 14:05:56.183134   15312 clusterclient.go:563] Waiting for Machine aws-controlplane-65vtk to become ready...
 ```
 
+#### Pre-built Kubernetes AMIs for testing
```
chuckha: 🙏

PR #135, opened 2018-09-28
Simple context implementation for cluster actuator

vincepri:

This is a basic idea on how to push forward the idea of a context based approach in the actuators.

None

None
**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/aws/actuators/cluster/context.go
+++ b/cloud/aws/actuators/cluster/context.go
@@ -1,0 +1,72 @@
+// Copyright © 2018 The Kubernetes Authors.
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package cluster
+
+import (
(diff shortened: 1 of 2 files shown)


cloud/aws/actuators/cluster/context.go
```diff
@@ -0,0 +1,72 @@
...
+// Context defines the context services accept and work within.
+type Context interface {
+	Cluster() *clusterv1.Cluster
+	Client() client.ClusterInterface
+	Config() *providerconfigv1.AWSClusterProviderConfig
+	Status() *providerconfigv1.AWSClusterProviderStatus
+}
+
+// contextFromCluster returns a new cluster context.
+func (a *Actuator) contextFromCluster(cluster *clusterv1.Cluster) (Context, error) {
```
chuckha: This should return a concrete type, not an interface. This is an antipattern that kubernetes follows and we should not follow that pattern.

Please see this paragraph https://peter.bourgon.org/go-for-industrial-programming/#how-much-interface-do-i-need

I will say defining interfaces is fine when other people have to implement an interface you define. For example, the sort package does this.

cloud/aws/actuators/cluster/context.go
```diff
@@ -0,0 +1,72 @@
...
+	}
+
+	if err := a.codec.DecodeProviderStatus(cluster.Status.ProviderStatus, res.providerStatus); err != nil {
+		return nil, errors.Errorf("failed to load cluster provider status: %v", err)
+	}
+
+	return res, nil
+}
+
+type contextImpl struct {
```
chuckha: can we call this Context and delete the interface since we're not using it anymore?

if you want to keep the interface, then this struct should at lest be exported

vincepri:
Gonna close this for now, as it's kinda early for it.

/close

Issue #62, opened 2018-08-30
[cluster actuator] Delete control plane node security group


detiber:
/kind feature
/priority important-soon

Issue #61, opened 2018-08-30
[cluster actuator] Delete NAT Gateway


detiber:
/kind feature
/priority important-soon

Issue #60, opened 2018-08-30
[cluster actuator] Delete routes for private subnet


detiber:
/kind feature
/priority important-soon

Issue #59, opened 2018-08-30
[cluster actuator] Delete private subnet


detiber:
/kind feature
/priority important-soon

PR #152, opened 2018-10-03
VPC Teardown

randomvariable:
Tears down all VPC related resources, pretty much in the reverse order that it was created.

Supports #39, #41, #43, #56, #59, #61, #60, #62

For resources that take some time to delete, like NAT gateways, was originally going
to return a retryable error for the actuator to requeue. Unfortunately, the re queuing behaviour is not tunable, and is based on constant retries within a short interval up to a maximum of 10.

This PR therefore implements a waiter based on that within API Machinery, but with an
additional helper to handle errors given by the AWS SDK.

This also provides a better UX for end-users as errors are really about something going
wrong rather than dealing with eventual consistency.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/aws/actuators/cluster/actuator.go
+++ b/cloud/aws/actuators/cluster/actuator.go
@@ -14,8 +14,6 @@
 package cluster
 
 import (
-	"fmt"
-
 	providerconfigv1 "sigs.k8s.io/cluster-api-provider-aws/cloud/aws/providerconfig/v1alpha1"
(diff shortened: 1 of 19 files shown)


randomvariable:
More or less functionally done. Cleaning up now.

randomvariable:
Need to add tests for teardown, but @detiber has in mind a slightly different approach to recursively delete resources in the VPC that we own, so probably doesn't make sense to do those right now.

randomvariable:
made a mess of rebasing against the bastion stuff

randomvariable:
This is sorted now. Properly rebased against #155

PR #153, opened 2018-10-03
Add runInstance method & refactor

vincepri:

This PR runs a `runInstance` method to the ec2 service which uses our own internal Instance type to run and describe instances.

None

None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/aws/providerconfig/v1alpha1/types.go
+++ b/cloud/aws/providerconfig/v1alpha1/types.go
@@ -341,8 +341,8 @@ type Instance struct {
 	// The name of the SSH key pair.
 	KeyName *string `json:"keyName"`
 
-	// The IAM instance profile associated with the instance, if applicable.
-	IamProfileID *string `json:"iamProfileId"`
+	// The ARN of the IAM instance profile associated with the instance, if applicable.
+	IAMProfile *AWSResourceReference `json:"iamProfile"`
 
 	// The private IPv4 address assigned to the instance.
 	PrivateIP *string `json:"privateIp"`
--- a/cloud/aws/services/ec2/instances.go
+++ b/cloud/aws/services/ec2/instances.go
@@ -29,8 +29,8 @@ func (s *Service) InstanceIfExists(instanceID *string) (*v1alpha1.Instance, erro
 	input := &ec2.DescribeInstancesInput{
 		InstanceIds: []*string{instanceID},
 	}
-	out, err := s.EC2.DescribeInstances(input)
 
+	out, err := s.EC2.DescribeInstances(input)
 	switch {
 	case IsNotFound(err):
 		return nil, nil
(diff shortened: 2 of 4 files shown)


cloud/aws/providerconfig/v1alpha1/types.go
```diff
@@ -341,8 +341,8 @@ type Instance struct {
 	// The name of the SSH key pair.
 	KeyName *string `json:"keyName"`
 
-	// The IAM instance profile associated with the instance, if applicable.
-	IamProfileID *string `json:"iamProfileId"`
+	// The ARN of the IAM instance profile associated with the instance, if applicable.
+	IamProfileARN *string `json:"iamProfileANR"`
```
chuckha: should this be a AWSResourceReference ?

cloud/aws/services/ec2/instances.go
```diff
@@ -39,47 +39,48 @@ func (s *Service) InstanceIfExists(instanceID *string) (*v1alpha1.Instance, erro
 	}
 
 	if len(out.Reservations) > 0 && len(out.Reservations[0].Instances) > 0 {
```
chuckha: should this throw error cases if those lens are zero? it's out of scope for this PR, but maybe an improvement for later

cloud/aws/services/ec2/instances.go
```diff
@@ -39,47 +39,48 @@ func (s *Service) InstanceIfExists(instanceID *string) (*v1alpha1.Instance, erro
...
 	}
 
-	reservation, err := s.EC2.RunInstances(input)
-	if err != nil {
-		return nil, errors.Wrapf(err, "failed to run instances")
-	}
-
-	if len(reservation.Instances) <= 0 {
-		return nil, errors.New("no instance was created after run was called")
+	// Pick instance profile.
```
chuckha: it might be worth noting that there is a default for some fields but not for others

cloud/aws/services/ec2/instances_test.go
```diff
@@ -208,7 +208,12 @@ metadata:
 spec:
   versions:
     kubelet: v1.11.2
-    controlPlane: v1.11.2`),
+    controlPlane: v1.11.2
+  providerConfig:
+  	value:
```
chuckha: There is a tab here, should be two spaces, but also this YAML isn't used. the expect is getting values from the call, not from the raw extension. Not terribly important, but probably good to be consistent with other tests

cloud/aws/services/ec2/instances.go
vincepri: hmm, maybe a not found error?

cloud/aws/services/ec2/instances.go
```diff
@@ -120,6 +121,35 @@ func (s *Service) CreateOrGetMachine(machine *clusterv1.Machine, status *v1alpha
 	return s.CreateInstance(machine, config, clusterStatus)
 }
 
+func (s *Service) runInstance(i *v1alpha1.Instance) (*v1alpha1.Instance, error) {
+	input := &ec2.RunInstancesInput{
+		InstanceType: aws.String(i.Type),
+		SubnetId:     aws.String(i.SubnetID),
+		ImageId:      aws.String(i.ImageID),
+		KeyName:      i.KeyName,
+		EbsOptimized: i.EBSOptimized,
```
chuckha: looks like this isn't passed through so it will always be nil

vincepri: Yeah I think that's fine for now, we might want to give the option later on

PR #156, opened 2018-10-04
Add cluster-api-provider-aws specific tag

detiber:
This PR adds a cluster-api-provider-aws specific tag to resources created and owned by cluster-api-provider-aws.

I plan on also reworking the way that we query objects based on tags and limit what operations we perform for owned vs shared objects.

### DIFF
--- a/cloud/aws/services/ec2/bastion.go
+++ b/cloud/aws/services/ec2/bastion.go
@@ -25,9 +25,6 @@ import (
 )
 
 const (
-	// TagValueBastionRole describes the value associated with the role tag.
-	TagValueBastionRole = "bastion"
-
 	bastionUserData = `#!/bin/bash
 
 BASTION_BOOTSTRAP_FILE=bastion_bootstrap.sh
@@ -103,14 +100,23 @@ func (s *Service) describeBastionInstance(clusterName string, status *v1alpha1.A
 		return nil, errors.Wrap(err, "failed to describe bastion host")
 	}
 
-	if len(out.Reservations) == 0 || len(out.Reservations[0].Instances) == 0 {
-		return nil, NewNotFound(errors.New("bastion host not found"))
+	// TODO: properly handle multiple bastions found rather than just returning
+	// the first non-terminated.
+	for _, res := range out.Reservations {
+		for _, instance := range res.Instances {
+			if aws.StringValue(instance.State.Name) != ec2.InstanceStateNameTerminated {
+				return fromSDKTypeToInstance(instance), nil
+			}
+		}
 	}
 
-	return fromSDKTypeToInstance(out.Reservations[0].Instances[0]), nil
+	return nil, NewNotFound(errors.New("bastion host not found"))
 }
 
 func (s *Service) getDefaultBastion(clusterName string, region string, network v1alpha1.Network, keyName string) *v1alpha1.Instance {
+	name := fmt.Sprintf("%s-bastion", clusterName)
+	tags := s.buildTags(clusterName, ResourceLifecycleOwned, name, TagValueBastionRole, nil)
+
 	i := &v1alpha1.Instance{
 		Type:     "t2.micro",
 		SubnetID: network.Subnets.FilterPublic()[0].ID,
@@ -120,10 +126,7 @@ func (s *Service) getDefaultBastion(clusterName string, region string, network v
 		SecurityGroupIDs: []string{
 			network.SecurityGroups[v1alpha1.SecurityGroupBastion].ID,
 		},
-		Tags: map[string]string{
-			s.clusterTagKey(clusterName): string(ResourceLifecycleOwned),
-			TagNameAWSClusterAPIRole:     TagValueBastionRole,
-		},
+		Tags: tags,
 	}
 
 	return i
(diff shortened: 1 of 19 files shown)


cloud/aws/services/ec2/bastion.go
```diff
@@ -103,14 +103,27 @@ func (s *Service) describeBastionInstance(clusterName string, status *v1alpha1.A
 		return nil, errors.Wrap(err, "failed to describe bastion host")
 	}
 
-	if len(out.Reservations) == 0 || len(out.Reservations[0].Instances) == 0 {
-		return nil, NewNotFound(errors.New("bastion host not found"))
+	// TODO: properly handle multiple bastions found rather than just returning
+	// the first non-terminated.
```
vincepri: I think we might be able to solve this by filtering on running instances only?

detiber: Maybe, we also need to consider that during reconciliation we might need to start stopped instances, etc.

vincepri: Sounds good, I think the TODO captures that.

cloud/aws/services/ec2/eips.go
```diff
@@ -14,14 +14,31 @@
...
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
 	"sigs.k8s.io/cluster-api-provider-aws/cloud/aws/services/wait"
 )
 
-func (s *Service) allocateAddress(clusterName string) (string, error) {
+func (s *Service) getOrAllocateAddress(clusterName string, role string) (string, error) {
+	addrs, err := s.describeAddresses(clusterName, role)
+	if err != nil {
+		return "", err
```
chuckha: this should get wrapped

cloud/aws/services/ec2/eips.go
```diff
@@ -30,11 +47,25 @@ func (s *Service) allocateAddress(clusterName string) (string, error) {
...
+
+func (s *Service) describeAddresses(clusterName string, role string) ([]*ec2.Address, error) {
+	out, err := s.EC2.DescribeAddresses(&ec2.DescribeAddressesInput{
+		Filters: []*ec2.Filter{
+			s.filterCluster(clusterName),
+			s.filterAWSProviderRole(role),
+		},
+	})
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to describe elastic IPs %q", err)
```
chuckha: This should just be `errors.Wrap(err, "failed to describe elastic IPs")`

chuckha: but upon further inspection, you can instead return `s.EC2.DescribeAddresses(...` directly

cloud/aws/services/ec2/routetables.go
```diff
@@ -142,12 +144,19 @@ func (s *Service) describeVpcRouteTables(clusterName string, vpcID string) ([]*e
 	return out.RouteTables, nil
 }
 
-func (s *Service) createRouteTableWithRoutes(clusterName string, vpc *v1alpha1.VPC, routes []*ec2.Route) (*v1alpha1.RouteTable, error) {
+func (s *Service) createRouteTableWithRoutes(clusterName string, vpc *v1alpha1.VPC, routes []*ec2.Route, isPublic bool) (*v1alpha1.RouteTable, error) {
```
chuckha: This and `createSubnet` look very similar. It's ok to copy pasta them for now, but one is getting isPublic passed in and the other is calculating it, might be something to clean up later

cloud/aws/services/ec2/tags.go
```diff
@@ -53,8 +61,8 @@ func (s *Service) clusterTagKey(clusterName string) string {
 }
 
 // createTags tags a resource with tags including the cluster tag
-func (s *Service) createTags(clusterName string, resourceID string, lifecycle ResourceLifecycle, additionalTags map[string]string) error {
-	tags := s.buildTags(clusterName, lifecycle, additionalTags)
+func (s *Service) createTags(clusterName string, resourceID string, lifecycle ResourceLifecycle, name string, role string, additionalTags map[string]string) error {
```
chuckha: can cluster the strings for a cleaner signature, optional, but `createTags(clusterName, resourceID, name, role string, ...)`

cloud/aws/services/ec2/tags.go
```diff
@@ -28,9 +28,17 @@ const (
...
+	// cluster-api-provider-aws owned components from other tooling that
+	// uses TagNameKubernetesClusterPrefix
+	TagNameAWSProviderManaged = "sigs.k8s.io/cluster-api-provider-aws/managed"
+
 	// TagNameAWSClusterAPIRole is the tag name we use to mark roles for resources
 	// dedicated to this cluster api provider implementation.
 	TagNameAWSClusterAPIRole = "sigs.k8s.io/cluster-api-provider-aws/role"
+
+	TagValueCommonRole    = "common"
+	TagValueAPIServerRole = "apiserver"
```
vincepri: Should we have a godoc string around these?

cloud/aws/services/ec2/routetables.go
detiber: Added a comment for later cleanup

cloud/aws/services/ec2/tags.go
detiber: Done

PR #162, opened 2018-10-05
Remove actuator tests

chuckha:

The actuator tests need a better separation between themselves and the service implementations. Until that exists these slow us down and add nothing.

**Release note**:

```release-note
NONE
```


randomvariable:
:fire: :fire: :fire: :fire:

PR #169, opened 2018-10-05
Use string for IAMInstanceProfile

detiber:
/assign @randomvariable 
/assign @chuckha


randomvariable:
Think this is fine. I'll let @chuckha do the honours.

PR #175, opened 2018-10-06
Misc bugfixes

detiber:
- Set keyname on instances

### DIFF
--- a/cloud/aws/services/certificates/certificates.go
+++ b/cloud/aws/services/certificates/certificates.go
@@ -191,12 +191,22 @@ func EncodePrivateKeyPEM(key *rsa.PrivateKey) []byte {
 	return pem.EncodeToMemory(&block)
 }
 
+// DecodeCertPEM attempts to return a decoded certificate or nil
+// if the encoded input does not contain a certificate
 func DecodeCertPEM(encoded []byte) (*x509.Certificate, error) {
 	block, _ := pem.Decode(encoded)
+	if block == nil {
+		return nil, nil
+	}
 	return x509.ParseCertificate(block.Bytes)
 }
 
+// DecodePrivateKeyPEM attempts to return a decoded key or nil
+// if the encoded input does not contain a private key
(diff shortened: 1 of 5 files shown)


clusterctl/examples/aws/machines.yaml.template
```diff
@@ -7,14 +7,17 @@ items:
...
+        controlPlane: v1.12.0
       providerConfig:
         value:
           apiVersion: awsproviderconfig/v1alpha1
           kind: AWSMachineProviderConfig
           instanceType: m5.large
-          iamInstanceProfile: "control-plane.cluster-api-provider-aws.sigs.k8s.io"
+          iamInstanceProfile: 
+            filters:
+            - "control-plane.cluster-api-provider-aws.sigs.k8s.io"
```
vincepri: I saw a change that converted the values to a single string earlier today, does that matter?

detiber: Nice catch! Nothing like committing partial unsaved changes unintentionally.

cloud/aws/services/certificates/certificates.go
```diff
@@ -193,10 +193,16 @@ func EncodePrivateKeyPEM(key *rsa.PrivateKey) []byte {
...
 func DecodeCertPEM(encoded []byte) (*x509.Certificate, error) {
 	block, _ := pem.Decode(encoded)
+	if block.Bytes == nil {
+		return nil, nil
+	}
 	return x509.ParseCertificate(block.Bytes)
 }
 
 func DecodePrivateKeyPEM(encoded []byte) (*rsa.PrivateKey, error) {
 	block, _ := pem.Decode(encoded)
```
vincepri: Should we log out the error here and above?

detiber: Potentially, that was going to be my first approach, but then I figured since we are returning a pointer to a certificate/key that it might make sense to just return nil instead to indicate no certificate/key could be decoded from input...

I can go either way with it, though.

vincepri: That sounds good, we should put that in a comment

PR #193, opened 2018-10-08
[clusterawsadm] Do not return error if stack update successful

detiber:
Currently clusterawsadm returns an error if the stack already exists even if the update is successful, this change will only return an error if the update fails.

/assign @randomvariable 
/assign @vincepri


detiber:
It looks like I need to handle the case where there are no updates to be performed still.
/hold

detiber:
Ok, edge case of update without changes handled now.
/cancel hold

Issue #78, opened 2018-09-11
Metrics/Tracing


randomvariable:
Would be good to get at least metrics added early on. 

A cluster operator may want to metrics from the controllers available in their monitoring system of choice, e.g. Prometheus. I would be interested in metrics with regards to the creation, deletion, mutation and failure rates of AWS APIs, and the state of their objects at least to help diagnose when limits have been reached.

Specifically, adding https://opencensus.io/quickstart/go/metrics/ would do this in a vendor agnostic fashion and allow the choice of tracing / metrics system, incl. CloudWatch and AWS X-Ray.

randomvariable:
if there's a strong preference for prometheus client_golang, also fine.

vincepri:
We should have some umbrella issue to track this, not sure what the standards is across k8s components.

CC @detiber

randomvariable:
Found them. https://github.com/kubernetes/community/blob/master/contributors/devel/instrumentation.md

https://docs.google.com/document/d/1U_h7osUDF6plKn_nkDF2Y-EjZA6BY8RR6LfTezrFBwM/edit?ts=5b926ad1 may also be useful.

randomvariable:
From the new KEP, this seems to be important:

> Kubernetes also makes extensive use of a global metrics registry to register metrics to be exposed. Aside from general shortcomings of global variables, Kubernetes is seeing actual effects of this, causing a number of components to use `sync.Once` or other mechanisms to ensure to not panic, when registering metrics. Instead a metrics registry should be passed to each component in order to explicitly register metrics instead of through `init` methods or other global, non-obvious executions.

> While uncertain at this point, once cleaned up, this effort may put us a step closer to having stability guarantees for Kubernetes around metrics. Currently metrics are excluded from any kind of stability requirements.

detiber:
@randomvariable I like this idea very much, should we also track this at the more generic cluster-api level as well?

detiber:
@vincepri I think we can use this issue as the umbrella issue and spawn out more specific issues to address as identified.

randomvariable:
On the last attempt at this, it seemed like one of the best ways for instrumenting AWS calls is to use the `WithContext` methods in the AWS SDK, and add a hook to get stats. It required opening up the HTTP requests to do well, so left it last time.

randomvariable:
Will sync with SIG Instrumentation to get latest advice

ncdc:
[...]

PR #194, opened 2018-10-08
[WIP] Add instrumentation and event recorder

### LINKED ISSUE #127: Events should be recorded into the events API
/kind feature

Events recorded as part of bootstrapping the cluster should be added to the k8s-wide Events API

randomvariable:
* Adds instrumentation and tracing across the controllers
  * Uses opencensus.io
* Adds an object event recorder for more structured logging and also updating
  API Server with status.

Fixes #127 and #78 

Not sure why I ended up doing this, now. Something, something, didn't know what the hell
the machine actuator was upto, something something.

**Definitely not ready**

And not for alpha1v1 in any case.

**Release note**:

```release-note

```

### DIFF
--- a/cloud/aws/events/events.go
+++ b/cloud/aws/events/events.go
@@ -1,0 +1,137 @@
+package events
+
+import (
+	"encoding/json"
+	"fmt"
+	"github.com/golang/glog"
+	"github.com/pkg/errors"
+	"k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/tools/record"
+	ref "k8s.io/client-go/tools/reference"
+	clusterapiclientsetscheme "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/scheme"
+)
+
+// ObjectRec holds a contained event recorder
+type ObjectRec struct {
+	rec *record.EventRecorder
+	obj runtime.Object
+}
+
+type key int
+
+const (
+	contextKey key = 0
+
+	// Normal represents a normal event
+	Normal = v1.EventTypeNormal
+
+	// Warning event
+	Warning = v1.EventTypeWarning
+
+	// Failure event
+	Failure = "Failure"
+
+	// Pending event
+	Pending = "Pending"
+)
+
+var (
+	globalEventRec *record.EventRecorder
+)
+
+type logEvent struct {
+	Object  *string `json:"object,omitempty"`
+	Type    string  `json:"type"`
+	Reason  string  `json:"reason"`
+	Message string  `json:"message"`
+}
+
+// Eventf sends a formatted event to the event recorder sink
+func (o *ObjectRec) Eventf(eventtype, reason, messageFmt string, args ...interface{}) {
+	rec := *o.rec
+	if o.obj == nil {
+		o.Infof(eventtype, reason, messageFmt, args...)
+	}
+	rec.Eventf(o.obj, eventtype, reason, messageFmt, args...)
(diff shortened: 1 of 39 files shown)


cloud/aws/events/events.go
```diff
@@ -0,0 +1,137 @@
...
+
+	// Failure event
+	Failure = "Failure"
+
+	// Pending event
+	Pending = "Pending"
+)
+
+var (
+	globalEventRec *record.EventRecorder
```
chuckha: I'm ok if this is a DefaultEventRecorder, similar to how http does DefaultHttpClient, but we should have a way to make a new one and not use mutable globals wherever possible.

What about if the API were like this:

```go
recorder := events.DefaultObjectRecorder
// or
recorder, err := events.NewRecorder(param1, param2)
// check error
objectRecorder, err := events.NewStdObjRecorder(recorder)
// check error
```

cloud/aws/instrumentation/instrumentation.go
```diff
@@ -0,0 +1,131 @@
...
+	"strings"
+)
+
+var (
+	AWSOperation, _ = tag.NewKey("aws.operation")
+	AWSService, _   = tag.NewKey("aws.service")
+	AWSRegion, _    = tag.NewKey("aws.region")
+)
+
+func awsRequestAction(r *http.Request) string {
```
chuckha: this reads the body, parses the query params, then I'm pretty confused on line 50, why does this put the body back into a reader that isn't closable? it then returns the query value for the Action paramter

randomvariable: Just experimenting at the moment. Was hacking around to get the body without destroying the request. It doesn't yet actually do the thing I want it to.

chuckha: ohh ok, didn't realize this was experimental code. hack on!

chuckha:
ah ha, totally missed https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/78 even though it was linked in the description

cloud/aws/events/events.go
randomvariable: Hah, yeah. Felt dirty writing that.

randomvariable:
Sorry, deliberately didn't say much on this PR so as to not distract from the MVP stuff.

Basically, I was wondering how easy it would be to monitor the AWS calls so that we could easily spot things like duplicate RunInstances when I got lost in the weeds of the machine actuator.

> Could you break this into several PRs?
* Yes. More of a POC than anything.

Think the order of importance here is:
1. Event recorder, as that'll populate the objects when people do a `kubectl describe`.
2. Instrumentation with either Prometheus client library or OpenCensus
3. Tracing with either OpenCensus, OpenTracing or other.

On the OpenCensus vs. OpenTracing stuff - don't think they're exactly the same thing. OpenTracing is pretty much client libraries + a specific protocol, whereas OpenCensus is a client library with pluggable protocols, one of which happens to be Prometheus.

The instrumentation's more important than tracing, and so the OpenCensus vs. Prometheus client library is probably more salient. Think I'll try both and see what they're like.

Advantage of the former is support for exporting metrics to CloudWatch & also getting support for tracing in a single library for those who want an all-on-AWS experience.

In terms of API stability, neither Prometheus client or the OpenCensus library for Go has reached version 1, but the latter is more in flux (hence pinning an actual commit is a constraint in Gopkg.toml.

Also, from a previous Twitter conversation, @rakyll was happy to assist.

randomvariable:
To return in a later PR...

PR #195, opened 2018-10-08
Create certificates in cluster actuator instead of machine actuator

detiber:
This should resolve the issue with missing certificates

I'm still testing this
/hold

/assign @chuckha


detiber:
/hold cancel

This appears to get things moving along further... It appears that the LB isn't properly wired up yet, though.

chuckha:
/lgtm 

makes way more sense to create the ca in the  cluster, also fixes the issue with multi-control-plane node clusters.

PR #196, opened 2018-10-08
[wip] Fixups for load balancers

randomvariable:
Fixes up ELB management. Still testing. Feel free to copy branch into your own testing,
fix and PR etc...

Fixes #

**Release note**:

```release-note

```


randomvariable:
Outstanding items:

- [ ] EC2 instances get created twice.
- [ ] here's a pesky ENI that hangs around well after the ELB is deleted causing errors on teardown.
- [ ] Not sure if instance association works.

cloud/aws/services/interfaces.go
```diff
@@ -70,4 +70,6 @@ type EC2MachineInterface interface {
 // ELBInterface encapsulates the methods exposed by the elb service.
 type ELBInterface interface {
 	ReconcileLoadbalancers(clusterName string, network *providerv1.Network) error
+	DeleteLoadbalancers(clusterName string, network *providerv1.Network) error
+	RegisterInstanceWithAPIServerELB(clusterName string, instanceID string) error
```
vincepri: Should we simplify this method by passing in a load balancer name/id instead?

randomvariable: That'd be better I think, the method names in the load balancer service did surprising things (like only ever creating an API Server load balancer).

Needs a bit of a rewrite I think.

randomvariable:
load balancer attachment verified

randomvariable:
This gets to a running cluster (at least once).

Additional fixups:
* SAN attributes for the ELB
* Setting pod CIDR range compatiable with Calico
* Getting the right URL in clusterctl

PR #199, opened 2018-10-09
Misc fixes for release

detiber:
Current working branch for testing for release

Builds upon #196


detiber:
Tests still need to be patched up for this, and after pivoting clusterctl still throws an error when trying to update the cluster endpoint, since the status is not transfered during the pivot.

There is also an issue with a second machine instance being created after pivoting, also because of the status not being transferred as part of the pivot.

Issue #165, opened 2018-10-05
Reconciling a cluster after manually deleting resources in AWS console broken


detiber:
/kind bug

**What steps did you take and what happened:**
Manually deleting the VPC (or other components) in ec2 causes the cluster reconciliation to break

**What did you expect to happen:**
The cluster actuator should either throw a non-retryable error (setting ErrorMessage/ErrorReason on the cluster status) or re-create the missing resources.

Current workaround is to edit the cluster object and cleanup the status to allow for the actuator to recreate the resources.

vincepri:
- [ ] VPC
- [ ] Subnets
- [ ] Internet Gateway
- [ ] NAT Gateway
- [ ] Routing Tables
- [ ] Routes
- [ ] Security groups

vincepri:
I went through each resource and they seem to be recreated appropriately on reconcile.
Feel free to reopen if the issue happens again.
/close

PR #216, opened 2018-10-10
bastion: Fix bastion code to always lookup via tags

phyber:
**What this PR does / why we need it**: Fixes bastion reconciliation after manually deleting in AWS console

Related to #165

**Special notes for your reviewer**: This fixes things so bastion lookup always happens via instance tags, fixing the reconciliation.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/cloud/aws/services/ec2/bastion.go
+++ b/cloud/aws/services/ec2/bastion.go
@@ -101,10 +101,6 @@ func (s *Service) DeleteBastion(clusterName string, status *v1alpha1.AWSClusterP
 }
 
 func (s *Service) describeBastionInstance(clusterName string, status *v1alpha1.AWSClusterProviderStatus) (*v1alpha1.Instance, error) {
-	if status.Bastion.ID != "" {
-		return s.InstanceIfExists(aws.String(status.Bastion.ID))
-	}
-
 	input := &ec2.DescribeInstancesInput{
 		Filters: []*ec2.Filter{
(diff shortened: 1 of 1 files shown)


vincepri:
@phyber Can you provide an example on why this is needed? Wouldn't the ID be sufficient in case we need to reconcile?

vincepri:
This was actually fixed in #219 I believe, the bastion instance describe filters now has a state filter.

CC @randomvariable

phyber:
I not sure that #219 fixes it. In the flow where Go was panicking, `describeBastionInstance` would describe the bastion via `InstanceIfExists`. This would return that it had found the instance, even if it was Terminated. And reconciliation would not happen. After an hour, the terminated instance would disappear from AWS and Go would now panic. This is probably due to `InstanceIfExists` returning `nil` for the error if it doesn't find the instance.

Removing the `InstanceIfExists` call and just getting the bastion by filtering on tags allowed the bastion to be reconciled if it was Terminated as `describeBastionInstance` knows how to deal with the terminated instance if it finds it via filtering on tags.

Arguably, I could have fixed up the `InstanceIfExists` call to check the `State` of the instance it's about to return, so that if it finds a `shutting-down` or `terminated` instance it returns that it found nothing (I actually think this should still be done).

However, due to other conversations I was under the impression that we were wanting to use the instance tags to find instances rather than relying on the instance IDs, so I opted for the solution here, to remove the check for `InstanceID` and just rely on the tags.

PR #242, opened 2018-10-12
linting: Switch to using golang-ci's metalinter

### LINKED ISSUE #239: go fmt silently errors
/kind bug

Changing from `/hack/verify-gofmt.sh` to `go fmt ./pkg/...` now just swallows errors.

``` shell
➜ make fmt
go fmt ./pkg/... ./cmd/...
pkg/cloud/aws/services/certificates/certificates.go
➜
```

Should have exited with an error code

Go 1.10 and 1.11 have different opinions on whitespace alignment, which is causing
developer issues.

**Environment:**

randomvariable:

Switch to using golang-ci's metalinter

gofmt gives different results on Go 1.11 vs 1.10, whereas this will give consistent
results.

Additionally, stops make from silently erroring.

Linting is also restricted to tests as far as Make goes

Fixes #239 

**Release note**:

```release-note

```


randomvariable:
I thought it might be a bit out of the ordinary, but turns out not so much:
controller-runtime, which we rely, on uses [gometalinter](https://github.com/kubernetes-sigs/controller-runtime/blob/master/.travis.yml#L18)

[Skaffold](https://github.com/GoogleContainerTools/skaffold) is the other major consumer of this particular linter though.

Makefile
```diff
@@ -60,24 +60,16 @@ cluster-api-dev-helper: vendor
...
-
-# Run go vet against code.
-vet:
-	go vet ./pkg/... ./cmd/...
-
-# Run go lint.
 lint:
-	golint || go get -u golang.org/x/lint/golint
-	golint -set_exit_status ./cmd/... ./pkg/... ./clusterctl/...
+	golangci-lint > /dev/null || scripts/install-golint.sh -b $$GOPATH/bin v1.10.2
```
chuckha: `$$GOPATH` should be `$(GOPATH)` since it's defined at the top of the file and not expected to be in the user's environment.

Why are is this installing into go's bin path and not the the default `./bin`?

chuckha:
Could you also add vendor as .PHONY target?

Makefile
randomvariable: copy pasta'd. fixing up

randomvariable: we don't have a bin directory at present btw

randomvariable:
Fixed GOPATH, and set up vendor as .PHONY unless FASTBUILD=true as previously.

Not changing where the lint tool is installed because we haven't yet defined a bindir, 
and that should be a different PR I think.

The current destination is taken from https://github.com/GoogleContainerTools/skaffold/blob/master/hack/linter.sh#L23 which is used in Travis builds

randomvariable:
Going to redo in Bazel. brb.

PR #244, opened 2018-10-14
clusterawsadm: improve user experience

### LINKED ISSUE #191: Improve clusterawsadm UX
/kind feature
/good-first-issue

The `clusterawsadm` command has some terrible UX:

e.g. you need to provide an account ID to this command.

``` bash
clusterawsadm alpha bootstrap generate-cloudformation
Error: accepts 1 arg(s), received 0
Usage:
clusterawsadm alpha bootstrap generate-cloudformation [flags]

Flags:
-h, --help   help for generate-cloudformation

accepts 1 arg(s), received 0

```
But it doesn't tell you what the argument is supposed to be.
Neither does it check that it's an account id.

It also reuses the glog tooling from the controllers, which is not
best for human readability, as in this example where no credentials
are present for CloudFormation to use.

``` bash
-> clusterawsadm alpha bootstrap create-stack
ERROR: logging before flag.Parse: E1008 16:38:12.671749    9045 bootstrap.go:82] unable to get caller identity: NoCredentialProviders: no valid providers in chain. Deprecated.
For verbose messaging see aws.Config.CredentialsChainVerboseErrors
```

This could be much improved, with the logging prefix removed, 
and errors handled gracefully with meaningful errors for
end users.

xmudrii:

This PR improves UX for clusterawsadm so it's easier to understand and fix errors.
The following changes are introduced:

* glog is not used anymore, as it's maybe not a great solution for CLI. Instead we've resorted to handling errors with Cobra and fmt.Printf
* Improved error messages to improve understandability and add instructions how to obtain needed resources
* Exit codes are correctly handled

Fixes #191 

Known problems:

* Link for Machine ID instructions may be incorrect one. Is there documentation for that?
* fmt.Printf approach may not be an elegant one. For example, usually messages are \n\n suffixed so output looks better, which is maybe not a great solution. If we would get rid of error codes, we could just return error and cobra would handle it. In that case error code would always be 1. 

(Potentially) breaking changes:

* Error codes are changed. By Unix/POSIX conventions error codes are in range 0-255. In some parts we were using error codes such as 500, which may cause some misbehavings.

**Release note**:

```release-note
clusterawsadm error messages are improved and correctly handled. Exit codes are changed to follow UNIX convention.
```

### DIFF
--- a/cmd/clusterawsadm/cmd/alpha/ec2/ec2.go
+++ b/cmd/clusterawsadm/cmd/alpha/ec2/ec2.go
@@ -15,8 +15,7 @@ package ec2
 
 import (
 	"fmt"
-
-	"github.com/golang/glog"
+	"os"
 
 	"github.com/aws/aws-sdk-go/aws/session"
 	awsec2 "github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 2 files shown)


cmd/clusterawsadm/cmd/alpha/ec2/ec2.go
```diff
@@ -40,30 +39,38 @@ func RootCmd() *cobra.Command {
...
 func consoleOutputCmd() *cobra.Command {
 	newCmd := &cobra.Command{
-		Use:   "console-output",
+		Use:   "console-output [EC2 Machine ID]",
 		Short: "Get console output for host",
-		Long:  "Get console output for host",
-		Args:  cobra.ExactArgs(1),
-		Run: func(cmd *cobra.Command, args []string) {
+		Long: `Get console output for host.
+You must enter a Machine ID for machine you want to get console output. Machine ID can be obtained by following https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-metadata.html
```
xmudrii: This link may not be a correct one. Is there a way to obtain Machine ID without accessing the machine?

randomvariable: I think we should kill this atm. It doesn't work very well (my fault), and that command itself only works for Generation 5 instances as it later turned out.

randomvariable:
/lgtm

will kill the ec2 in another PR.

PR #251, opened 2018-10-15
adds basic CRD to get started

chuckha:

This PR adds the CRD definitions to continue the building. These should be generated but for now they are not. This is a stop gap to keep development going.

**Release note**:

```release-note
NONE
```


randomvariable:
if these came from kubebuilder, can we also have the annotated go files in pkg/apis?

PR #254, opened 2018-10-15
Migrate to kubebuilder

### LINKED ISSUE #252: Generate CRDs with kubebuilder
/kind feature

Instead of copy/pasting CRDs from the gcp project we should generate them ourselves

**Environment:**

- OS (e.g. from `/etc/os-release`): @

chuckha:
* Spec/Status is now housed under one ProviderConfig
* Controllers and types generated with kubebuilder
* CRDs generated with kubebuilder

This PR migrates our code to be in line with kubebuilder expectations. This has a few side effects such as reducing some boiler plate code.

The controllers are responsible for updating the objects so we no longer have to manage this ourselves.

Fixes #252 

I'm happy to walk through this with you if you like

**Release note**:

```release-note
NONE
```


pkg/apis/awsproviderconfig/v1alpha1/awsclusterproviderconfig_types.go
```diff
@@ -0,0 +1,74 @@
...
+type AWSClusterProviderConfigSpec struct {
+	// The AWS Region the cluster lives in.
+	Region string `json:"region"`
+
+	// SSHKeyName is the name of the ssh key to attach to the bastion host.
+	SSHKeyName string `json:"sshKeyName,omitempty"`
+}
+
+// AWSClusterProviderConfigStatus defines the observed state of AWSClusterProviderConfig
+type AWSClusterProviderConfigStatus struct {
```
detiber: I'm not sure we should overload the providerconfig to include a persisted status here...

chuckha: oh hmm because the Cluster would have a status already?

chuckha:
/hold

so i can get some feedback

chuckha:
oops, close but this is not what we want. close though!

chuckha:
oops, close but this is not what we want. close though!

vincepri:
@chuckha can you elaborate a little please?

pkg/apis/awsproviderconfig/v1alpha1/awsclusterproviderconfig_types.go
detiber: We'd be conflating spec/status concepts and would be splitting our status between the actual Cluster/Machine Status and Spec.ProviderConfig

PR #255, opened 2018-10-15
Return a VPC only if it's available or pending

vincepri:

This should fix the reconcile in case the VPC doesn't exist anymore.

Fixes #

None

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/vpc.go
+++ b/pkg/cloud/aws/services/ec2/vpc.go
@@ -100,7 +100,11 @@ func (s *Service) deleteVPC(v *v1alpha1.VPC) error {
 }
 
 func (s *Service) describeVPC(clusterName string, id string) (*v1alpha1.VPC, error) {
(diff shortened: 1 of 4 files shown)


pkg/cloud/aws/services/ec2/vpc.go
```diff
@@ -124,6 +128,12 @@ func (s *Service) describeVPC(clusterName string, id string) (*v1alpha1.VPC, err
 		return nil, NewConflict(errors.Errorf("found more than one vpc with supplied filters. Please clean up extra VPCs: %s", out.GoString()))
 	}
 
+	switch *out.Vpcs[0].State {
+	case ec2.VpcStateAvailable, ec2.VpcStatePending:
```
detiber: Is this needed if we are filtering on the VPC states we are checking for here?

vincepri: I guess we don't just added it to be explicit in case we ever change the filters

PR #256, opened 2018-10-15
Adds kubebuilder migration types

chuckha:

This Adds the the config/status types that are generated with kubebuilder

Fixes #

This does not *use* the types, simply adds them.

**Release note**:

```release-note
NONE
```


vincepri:
@chuckha can we move the other types in the new folder instead of duplicating them?

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
```diff
@@ -0,0 +1,51 @@
...
+	// The AWS Region the cluster lives in.
+	Region string `json:"region"`
+
+	// SSHKeyName is the name of the ssh key to attach to the bastion host.
+	SSHKeyName string `json:"sshKeyName,omitempty"`
+}
+
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// AWSClusterProviderConfigList contains a list of AWSClusterProviderConfig
```
detiber: I wonder if there is a way to exclude the lists when generating the types, I can't think of a case where we would need them, since our types are embedded.

pkg/apis/awsprovider/v1alpha1/awsclusterproviderstatus_types.go
```diff
@@ -0,0 +1,55 @@
...
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// AWSClusterProviderStatus contains the status fields
+// relevant to AWS in the cluster object.
+// +k8s:openapi-gen=true
+type AWSClusterProviderStatus struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Region  string   `json:"region"`
```
detiber: Not directly related to this conversion, but we probably shouldn't be duplicating the Region in both the providerconfig and providerstatus.

pkg/controller/add_cluster_conteroller.go
```diff
@@ -0,0 +1,34 @@
+/*
```
detiber: should this file be add_cluster_controller.go instead?

config/samples/awsprovider_v1alpha1_awsclusterproviderconfig.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: awsprovider.k8s.io/v1alpha1
```
detiber: We should probably exclude these samples.

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
chuckha: no, but I can delete them. I also don't think we'll need this type.

pkg/controller/add_cluster_conteroller.go
chuckha: lmao yes

randomvariable:
Not so concerned about exact preservation of the file history atm, we're still in alpha2.

PR #258, opened 2018-10-15
Improve NAT logging & filtering

vincepri:

This PR improves NAT gateway logs and the filtering when describing NAT gateways. Also makes sure to always return a sorted list of AZs.

Related to #241

None

**Release note**:

```release-note
NONE
```


randomvariable:
Possibly some more idempotency stuff to be done here (may end up with multiple NAT gateways if filtering only on available and another reconciliation loop starts), but still an improvement.

/lgtm

Issue #247, opened 2018-10-15
UMBRELLA: CRD Migration


randomvariable:
/kind feature

**Describe the solution you'd like**
Umbrella issue for outstanding post-CRD migration steps

- [X] Update documentation (#253)
- [x] Generate and set up CRDs in bootstrap cluster (#252)
- [x] Enable status updates (#288)

PR #265, opened 2018-10-17
controller: Add RBAC annotations for kubebuilder

randomvariable:
Adds kubebuilder RBAC annotations for the manifest creation command
to automagically create relevant RBAC roles.

Part of #247 

**Release note**:

```release-note

```


randomvariable:
think i've matched the group name correctly now

detiber:
@randomvariable this looks fine, at some point we probably need to see if we can slim down the objects that we are creating, since we don't actually need to store, update, and retrieve the CRDs directly.

/approve
/lgtm

Issue #253, opened 2018-10-15
Panic: No AWSClusterProviderConfig type available [CRD]


randomvariable:
/kind bug

Part of #247 

**What steps did you take and what happened:**

After placing CRD types into the same namespace as the AWS controller and into provider components, attempted to start. Controller attempts to read object, and then fails:

``` json
{
    "level": "error",
    "ts": 1539618115.3119898,
    "logger": "kubebuilder.controller",
    "caller": "controller/controller.go:209",
    "msg": "Reconciler error",
    "Controller": "cluster-controller",
    "Request": "default/test1",
    "error": "failed to load cluster provider config: decoding failure: no kind \"AWSClusterProviderConfig\" is registered for version \"awsproviderconfig/v1alpha1\"",
[...]

PR #286, opened 2018-10-18
Generate manifests with kustomize, fix RBAC

vincepri:

This PR standardized on kustomize and makes the development process a little less tedious. It also fixes some issues with RBAC that prevented the manager from running accordingly. It also moves `clusterctl` under `cmd/`.

Fixes #281, #287, #253 

**Release note**:

```release-note

```


BUILD.bazel
```diff
@@ -16,7 +16,7 @@ gazelle(
 filegroup(
     name = "all-images",
     srcs = [
-        "//clusterctl:clusterctl-image",
+        "//cmd/clusterctl:clusterctl-image",
         "//cmd/clusterawsadm:clusterawsadm-image",
```
detiber: Am I misreading this? Is this creating a container image for clusterctl and clusterawsadm? @liztio

Makefile
```diff
@@ -81,27 +82,42 @@ vet:
...
 
 # Push the docker image
 docker-push:
-	docker push ${IMG}
+	docker push ${MANAGER_IMAGE}
 
 # Cleanup
 clean:
-	rm -rf out
+	rm -rf cmd/clusterctl/examples/aws/out/
```
detiber: I think we should consider keeping the generated yaml closer to the root of the repo to avoid the user having to reference deep nested paths when using the generated files.

config/default/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
...
+secretGenerator:
+- name: machine-controller-credential
+  type: Opaque
+  commands:
+    credentials: "cat ../../cmd/clusterctl/examples/aws/out/credentials"
+
+configMapGenerator:
+- name: machine-setup
+  files:
+  - ../../cmd/clusterctl/examples/aws/out/machines.yaml
```
detiber: Is this creating a configMap of the machines.yaml? Or am I missing something?

cmd/clusterctl/examples/aws/generate-yaml.sh
```diff
@@ -65,21 +62,13 @@ if [ $OVERWRITE -ne 1 ] && [ -f $CLUSTER_GENERATED_FILE ]; then
...
-
 mkdir -p ${OUTPUT_DIR}
 
-envsubst < $PROVIDERCOMPONENT_TEMPLATE_FILE \
-  > "${PROVIDERCOMPONENT_GENERATED_FILE}"
-echo "Done generating ${PROVIDERCOMPONENT_GENERATED_FILE}"
-
-envsubst < $CLUSTER_TEMPLATE_FILE \
-  > "${CLUSTER_GENERATED_FILE}"
+envsubst < $CLUSTER_TEMPLATE_FILE > "${CLUSTER_GENERATED_FILE}"
```
detiber: Could we accomplish this by generating a kustomize patch instead of using envsubst directly against the cluster and machine yaml? It seems odd to modify templates in two separate ways. It may also be possible to use the `kustomize edit` command to modify settings based on env vars directly.

Makefile
```diff
@@ -81,27 +82,42 @@ vet:
...
 	rm -f minikube.kubeconfig
 
-manifests:
-	go run vendor/sigs.k8s.io/controller-tools/cmd/controller-gen/main.go rbac --name aws-manager
+# Manifests.
+cmd/clusterctl/examples/aws/out/:
+	./cmd/clusterctl/examples/aws/generate-yaml.sh
+
+cmd/clusterctl/examples/aws/out/credentials: cmd/clusterctl/examples/aws/out/ clusterawsadm
+	clusterawsadm alpha bootstrap generate-aws-default-profile > cmd/clusterctl/examples/aws/out/credentials
```
ashish-amarnath: Don't think `cmd/clusterctl/examples/aws/out` is a good place to place the credentials as it is not part of an example.
Suggest creating like a `build.out` directory to store the manifests. WDYT?

cmd/clusterctl/examples/aws/generate-yaml.sh
```diff
@@ -2,30 +2,27 @@
...
+# Directories.
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
-
 OUTPUT_DIR=${DIR}/out
 
-# provider-components
-export CLUSTER_CONTROLLER_IMAGE="${CLUSTER_CONTROLLER_IMAGE:-gcr.io/k8s-cluster-api/aws-cluster-controller:0.0.1}"
-export MACHINE_CONTROLLER_IMAGE="${MACHINE_CONTROLLER_IMAGE:-gcr.io/k8s-cluster-api/aws-machine-controller:0.0.1}"
+# Manager image.
+export MANAGER_IMAGE="${MANAGER_IMAGE:-gcr.io/k8s-cluster-api/aws-cluster-controller:0.0.1}"
```
ashish-amarnath: `MANAGER_IMAGE` is supplied here and also in the makefile.
Can we supply this value in one place?

ashish-amarnath: maybe error out if `MANAGER_IMAGE` is not set when this script runs.

cmd/clusterctl/examples/aws/generate-yaml.sh
```diff
@@ -2,30 +2,27 @@
...
 
-# aws credentials
-# TODO generate a secret instead of embedding directly in the YAML
-export AWS_REGION="${AWS_REGION:-us-west-2}"
-export AWS_ACCESS_KEY_ID="${AWS_ACCESS_KEY_ID}"
-export AWS_SECRET_ACCESS_KEY="${AWS_SECRET_ACCESS_KEY}"
+# Machine settings.
+export CONTROL_PLANE_MACHINE_TYPE="${CONTROL_PLANE_MACHINE_TYPE:-t2.medium}"
```
ashish-amarnath: Suggest naming this variable `MASTER_NODE_TYPE`  and `WORKER_NODE_TYPE`.

cmd/clusterctl/examples/aws/generate-yaml.sh
```diff
@@ -2,30 +2,27 @@
...
-PROVIDERCOMPONENT_TEMPLATE_FILE=${DIR}/provider-components.yaml.template
-PROVIDERCOMPONENT_GENERATED_FILE=${OUTPUT_DIR}/provider-components.yaml
-
+# Templates.
 CLUSTER_TEMPLATE_FILE=${DIR}/cluster.yaml.template
 CLUSTER_GENERATED_FILE=${OUTPUT_DIR}/cluster.yaml
-
 MACHINES_TEMPLATE_FILE=${DIR}/machines.yaml.template
 MACHINES_GENERATED_FILE=${OUTPUT_DIR}/machines.yaml
+MANAGER_PATCH_TEMPLATE_FILE=${DIR}/aws_manager_image_patch.yaml.template
```
ashish-amarnath: Does it make sense to gather all the template yamls into a `/templates` directory?

config/manager/manager.yaml
```diff
@@ -1,3 +1,25 @@
+apiVersion: v1
+kind: Namespace
+metadata:
+  labels:
+    controller-tools.k8s.io: "1.0"
+  name: system #TODO: add the prefix here manually, next version of kustomize won't prefix namespaces.
```
ashish-amarnath: Suggest using a variable for the namespace here. It'll allow us to change the value of the variable and let it take effect in all the places instead of changing it in multiple places. Maybe in `envsubst`

config/manager/manager.yaml
```diff
@@ -43,7 +66,7 @@ spec:
             - name: certs
               mountPath: /etc/ssl/certs
             - name: credentials
-              mountPath: /etc/aws
+              mountPath: /root/.aws
```
ashish-amarnath: This assumes that the controllers run as root. Is that intentional?

config/default/kustomization.yaml
```diff
@@ -0,0 +1,25 @@
...
+- ../crds/awsprovider_v1alpha1_awsmachineproviderstatus.yaml
+- ../rbac/rbac_role.yaml
+- ../rbac/rbac_role_binding.yaml
+- ../manager/manager.yaml
+
+patches:
+- ../../cmd/clusterctl/examples/aws/out/aws_manager_image_patch.yaml
+
+secretGenerator:
+- name: machine-controller-credential
```
ashish-amarnath: Suggest naming this `bootstrap-credentials` or `infra-provisioner-creds` as these aren't the creds that will be used by the machine controllers in the internal clusters.

cmd/clusterctl/examples/aws/generate-yaml.sh
detiber: We're actively working on avoiding new uses of `master` and trying to eliminate existing uses of `master` as we can.

config/default/kustomization.yaml
vincepri: That's correct

cmd/clusterctl/examples/aws/generate-yaml.sh
vincepri: I don't think kustomize is there yet to allow for environment substitution, but I might be wrong. We should revisit this as kustomize adds in more features.

Makefile
vincepri: The clusterctl command always references the out directory in the example. We should probably revisit this later on when the example grows.

cmd/clusterctl/examples/aws/generate-yaml.sh
vincepri: `MANAGER_IMAGE` in the `Makefile` actually sets this variable when the Makefile runs.

cmd/clusterctl/examples/aws/generate-yaml.sh
vincepri: Yeah I think that'd be valuable to refactor and use a for loop instead, not sure for how long we're keeping this bash script though, given that in the future we might rely on `kustomize` more and more.

config/manager/manager.yaml
vincepri: That might complicate things a little, we'd need a template specific for the namespace.

config/manager/manager.yaml
vincepri: They do, the docker container user is root (for now).

config/default/kustomization.yaml
vincepri: +1 changed

config/default/kustomization.yaml
detiber: I don't think we have a use case for storing the machines as a config map, the machines should be applied directly from the machines.yaml on the command line from clusterctl.

randomvariable:
Fixes #253

Makefile
ashish-amarnath: We should change that. IMO, the credentials file doesn't really belong in the `examples` dir. We don't want people leaking their creds  by mistake.

vincepri: Could we track / fix that as a follow-up item or is this a blocking issue?

detiber: Lets track as a followup item.

detiber:
Lets get this merged to unblock others, and file issues for any followup items.
/lgtm

PR #301, opened 2018-10-23
misc post-crd fixes

detiber:
- clusterawsadm: consolidate controller roles/profiles, since there is only a single controller manager now
- clusterawsadm: fixup glog error
- aws provider controller manager: fixup argument error for the aws provider controller manager statefulset.

### DIFF
--- a/pkg/cloud/aws/services/cloudformation/bootstrap.go
+++ b/pkg/cloud/aws/services/cloudformation/bootstrap.go
@@ -15,9 +15,11 @@ package cloudformation
 
 import (
 	"fmt"
+
 	"github.com/awslabs/goformation/cloudformation"
 	"github.com/golang/glog"
 	"github.com/pkg/errors"
+
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/awserrors"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/iam"
 )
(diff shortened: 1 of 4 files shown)


pkg/cloud/aws/services/cloudformation/bootstrap.go
```diff
@@ -27,34 +29,21 @@ import (
...
-			cloudformation.Ref("AWSIAMRoleClusterController"),
-			cloudformation.Ref("AWSIAMRoleControlPlane"),
-		},
-	}
-
-	template.Resources["AWSIAMManagedPolicyMachineController"] = cloudformation.AWSIAMManagedPolicy{
-		ManagedPolicyName: iam.NewManagedName("machine-controller"),
-		Description:       `For the Kubernetes Cluster API Provider AWS Machine Controller`,
-		PolicyDocument:    machineControllerPolicy(accountID),
+	template.Resources["AWSIAMManagedPolicyControllers"] = cloudformation.AWSIAMManagedPolicy{
```
detiber: @randomvariable considering the post-crd deployment model of using a shared controller manager, it seems to make sense to collapse these roles/profiles down. Thoughts?

randomvariable: For now, maybe. But we might want to separate them back out at some point.

The GCP provider actually takes a flag to say which actuator your starting.

I was thinking of the following use case:

You have a central management cluster that's responsible for spinning up clusters and control plane nodes, and that's not pivoted.

The target cluster may still have a machine actuator which deals with machinesets or machines that the cluster autoscaler is creating / modifying.

randomvariable: Happy to lgtm it with whatever you think.

detiber:
@vincepri PRs are squashed on merge automatically: https://github.com/kubernetes/test-infra/blob/2005cfe3b3f0148cfcfbb7272a9dee64e4a87211/prow/config.yaml#L418

/hold cancel

vincepri:
Nice, thanks for the link :)

PR #304, opened 2018-10-23
Cluster and machine actuator updates status

ashish-amarnath:
Currently, updated status for cluster and machine don't get stored so the actuators work is not saved. This PR saves the status for both cluster and machine.

In this PR:
* add methods to encode ClusterProviderStatus and MachineProviderStatus
* add methods to store cluster and machine status
* store cluster and machine status in actuator function
 
Fixes # 288

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/apis/awsprovider/v1alpha1/register.go
+++ b/pkg/apis/awsprovider/v1alpha1/register.go
@@ -97,7 +97,6 @@ func EncodeMachineStatus(status *AWSMachineProviderStatus) (*runtime.RawExtensio
 
 	return &runtime.RawExtension{
 		Raw: rawBytes,
-		//Object: ??,
 	}, nil
 }
 
@@ -115,6 +114,5 @@ func EncodeClusterStatus(status *AWSClusterProviderStatus) (*runtime.RawExtensio
 
 	return &runtime.RawExtension{
 		Raw: rawBytes,
-		//Object: ??,
 	}, nil
 }
--- a/pkg/apis/awsprovider/v1alpha1/register.go
+++ b/pkg/apis/awsprovider/v1alpha1/register.go
@@ -81,3 +81,40 @@ func MachineStatusFromProviderStatus(extension *runtime.RawExtension) (*AWSMachi
 
 	return status, nil
 }
+
+func EncodeMachineStatus(status *AWSMachineProviderStatus) (*runtime.RawExtension, error) {
+	if status == nil {
+		return &runtime.RawExtension{}, nil
+	}
+
+	var rawBytes []byte
+	var err error
+
+	//  TODO: use apimachinery conversion https://godoc.org/k8s.io/apimachinery/pkg/runtime#Convert_runtime_Object_To_runtime_RawExtension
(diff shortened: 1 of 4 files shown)


pkg/apis/awsprovider/v1alpha1/register.go
```diff
@@ -81,3 +81,37 @@ func MachineStatusFromProviderStatus(extension *runtime.RawExtension) (*AWSMachi
...
+func EncodeClusterStatus(status *AWSClusterProviderStatus) (*runtime.RawExtension, error) {
+	if status == nil {
+		return &runtime.RawExtension{}, nil
+	}
+	var rawBytes []byte
+	var err error
+
+	if rawBytes, err = yaml.Marshal(status); err != nil {
+		return nil, err
+	}
```
vincepri: The above could be simplified a little
```
rawBytes, err := yaml.Marshal(status)
if err != nil {
		return nil, err
	}
```
same for the `EncodeMachineStatus`

pkg/apis/awsprovider/v1alpha1/register.go
```diff
@@ -81,3 +81,37 @@ func MachineStatusFromProviderStatus(extension *runtime.RawExtension) (*AWSMachi
...
+
+	var rawBytes []byte
+	var err error
+	if rawBytes, err = yaml.Marshal(status); err != nil {
+		return nil, err
+	}
+
+	return &runtime.RawExtension{
+		Raw: rawBytes,
+		//Object: ??,
```
vincepri: Do we need the Object part at all?

CC @detiber

pkg/apis/awsprovider/v1alpha1/register.go
ashish-amarnath: This will goaway when we switch to the api machinery conversion.
I will do that as a part of this PR

pkg/apis/awsprovider/v1alpha1/register.go
ashish-amarnath: This will goaway when we switch to the api machinery conversion.
I will do that as a part of this PR

ashish-amarnath: Looks like we'll need to revisit the api-machinery conversion. Will open an issue for that if this remains unsolved.

pkg/apis/awsprovider/v1alpha1/register.go
ashish-amarnath: Looks like we'll need to revisit the api-machinery conversion. Will open an issue for that if this remains unsolved.

ashish-amarnath: Created https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/306

pkg/apis/awsprovider/v1alpha1/register.go
ashish-amarnath: Created https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/306

PR #310, opened 2018-10-24
[WIP] Remove make from end-user workflow

detiber:
Removes `make` from end-user workflow and replaces the current manifest generation steps with `clusterawsadm`


detiber:
Still very much WIP, mostly some moving things around currently.

Plan is to continue to extend by moving the steps performed in `generate-yaml.sh` into clusterawsadm as well with the output being kustomize patches rather than transforming directly into the manifests.

Once the patches are created, then we can have the user run kustomize as a separate step.

randomvariable:
Superseded by #580 

/close

PR #311, opened 2018-10-25
Add a make target for generating the binaries for release

detiber:
Adds a make target for generating the binaries needed for cutting a release of cluster-api-provider-aws.


randomvariable:
/hold 
Should be getting #290 in shortly, and then we should test the Bazel images.

randomvariable:
Was superseded by #290 
/close

PR #325, opened 2018-10-26
Adding test script to test ci credentials work for us

### LINKED ISSUE #318: Periodic job that tests our amazon credentials
We want a periodic Prow job that checks out AWS credentials situation. It should call some very basic "am I logged in?" endpoint with Amazon, maybe check what our account limits currently are.

This should run once every hour or so.

rdodev:
Signed-off-by: Ruben Orduz <rubenoz@gmail.com>

Simple shell script which will be run periodically to test the aws credentials work for our purposes. 

Fixes #318 

Opted to use bash script (as opposed to Go script) for sake of quick turn-around.

### DIFF
--- a/scripts/ci-aws-cred-test.sh
+++ b/scripts/ci-aws-cred-test.sh
@@ -1,0 +1,58 @@
+#!/bin/bash
+
+# Copyright 2018 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+
+GROUP_NAME=capa-cred-test-group-$(openssl rand -hex 8)
+USERNAME=capa-cred-test-user-$(openssl rand -hex 8)
+
+echo "Group: ${GROUP_NAME}\nUser: ${USERNAME}"
+
+echo "Creating IAM group"
+aws iam create-group --group-name "${GROUP_NAME}"
+
+echo "Attaching Policies"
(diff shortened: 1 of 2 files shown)


rdodev:
@detiber I did think about that. But given how _not-idempotent_ aws is, you'd have to have some sort of state storage so that you can delete the resources that were actually created and any other entity association therein (policy attachment, for example). If you try to cleanup orphan entities wholesale that job is going to fail, a lot :/ Now, we don't really create many entities/objects. We only create a group and a user (which we echo to the tty). The policy attachments are AWS built-in. So in the worst case scenario, we'd have user and or groups, which should be easy to delete from a root account if it ever becomes an issue.

chuckha:
I'm not sure I understand how this is supposed to be used. It doesn't appear to be part of CI since I did not see a corresponding PR in kubernetes/test-infra. How would I run this? It looks like I should already have aws credentials in my environment or use something like aws-vault?

rdodev:
There's a credential-testing job PR in test-infra. I should have included a link in my PR: https://github.com/kubernetes/test-infra/pull/9936

scripts/ci-aws-cred-test.sh
```diff
@@ -0,0 +1,58 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+set -o pipefail
+
+
+GROUP_NAME=$(openssl rand -hex 8)
```
chuckha: a prefix or suffix here would make it easier to identify the potentially orphaned resources for someone manually cleaning up the account.

The alternative would be a trap with some deletes, but that can get annoying to write for sure.

rdodev: Good idea. Just pushed changes for that.

chuckha:
/lgtm
do we know who the person who has root access who can help clean this up is?

/assign @detiber

rdodev:
@chuckha according to Ben, Justin sb set these credentials up to be used for kops testing.

chuckha:
I am just a little worried about finding a human to clean this up, but if you think it's not an issue, i'm good with this strategy

rdodev:
@vincepri can you 👀 ? Apparently it needs your review

detiber:
@rdodev it looks like it's treating `--bare` as the command here: https://github.com/rdodev/test-infra/blob/ae0a10cb3fbe7b83f9d6f4a044fa5f0a306add01/config/jobs/kubernetes-sigs/cluster-api-provider-aws/cluster-api-provider-aws-presubmits.yaml#L35

rdodev:
Yup. Gonna see what's that all about.

PR #334, opened 2018-10-29
Fix issue if no gopath is set

chuckha:

Fixes an issue with go generate that requires GOPATH to be set.

Fixes #

**Release note**:
```release-note
NONE
```


chuckha:
This comes from bad defaults with gopath here: https://github.com/kubernetes/gengo/blob/master/args/args.go#L98

PR #336, opened 2018-10-29
Add state filtering for subnets

vincepri:

Adds a filter to the describe method making sure subnets are either pending or available

Related to #165 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/aws/services/ec2/subnets_test.go
+++ b/pkg/cloud/aws/services/ec2/subnets_test.go
@@ -72,6 +72,10 @@ func TestReconcileSubnets(t *testing.T) {
 								Name:   aws.String("tag-key"),
 								Values: []*string{aws.String("kubernetes.io/cluster/test-cluster")},
 							},
+							{
+								Name:   aws.String("state"),
+								Values: []*string{aws.String("pending"), aws.String("available")},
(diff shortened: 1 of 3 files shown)


detiber:
I just ungated the PR for adding the credentials test, so that should fix up the test failure.

pkg/cloud/aws/services/ec2/subnets_test.go
```diff
@@ -149,6 +153,10 @@ func TestReconcileSubnets(t *testing.T) {
 								Name:   aws.String("tag-key"),
 								Values: []*string{aws.String("kubernetes.io/cluster/test-cluster")},
 							},
+							{
```
detiber: Any reason we need to actually compare the arguments here instead of using gomock.Any()?

vincepri: We could, but in this case for example, we do want to make sure the filters are actually in place. I think we should use `Any` as some sort of last resort, otherwise it'll be hard to tell what unit tests are testing, thoughts?

detiber:
@vincepri might want to rebase and force push to trigger the right prow criteria.

vincepri:
Updated

PR #340, opened 2018-10-30
Improve release tooling

### LINKED ISSUE #321: Update cluster-api-aws-controller image tag to not use latest
/kind feature

Currently we are using latest for the built/pushed cluster-api-aws-controller image tag, which means during the next release we will break the existing v1.0.0-alpha.2 release, which also references the latest tag.

We should use more specific tags, but don't necessarily have to align with the same version as the release, since not every release will necessarily require a new controller image.

detiber:
- Update release image to use a version tag rather than latest
- Add a target for generating the release binaries

Fixes #321

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,7 +22,7 @@ FASTBUILD ?= n ## Set FASTBUILD=y (case-sensitive) to skip some slow tasks
 ## Image URL to use all building/pushing image targets
 STABLE_DOCKER_REPO ?= gcr.io/cluster-api-provider-aws
 MANAGER_IMAGE_NAME ?= cluster-api-aws-controller
-MANAGER_IMAGE_TAG ?= latest
+MANAGER_IMAGE_TAG ?= 0.0.2
 MANAGER_IMAGE ?= $(STABLE_DOCKER_REPO)/$(MANAGER_IMAGE_NAME):$(MANAGER_IMAGE_TAG)
 DEV_DOCKER_REPO ?= gcr.io/$(shell gcloud config get-value project)
 DEV_MANAGER_IMAGE ?= $(DEV_DOCKER_REPO)/$(MANAGER_IMAGE_NAME):$(MANAGER_IMAGE_TAG)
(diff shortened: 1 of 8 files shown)


cmd/clusterctl/BUILD
```diff
@@ -17,6 +16,5 @@ go_binary(
     name = "clusterctl",
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
+    pure = "on",
```
vincepri: Is this a valid option for go_library? I can't find it in the docs https://github.com/bazelbuild/rules_go/blob/master/go/core.rst#go-library

Makefile
```diff
@@ -81,6 +81,15 @@ clusterawsadm: dep-ensure ## Build clusterawsadm binary.
...
 	bazel build //hack/cluster-api-dev-helper $(BAZEL_ARGS)
 
+.PHONY: release-binaries
+release-binaries: ## Build release binaries
+	bazel build --platforms=@io_bazel_rules_go//go/toolchain:linux_amd64 //cmd/clusterctl //cmd/clusterawsadm
+	bazel build --platforms=@io_bazel_rules_go//go/toolchain:darwin_amd64 //cmd/clusterctl //cmd/clusterawsadm
+	install bazel-bin/cmd/clusterawsadm/darwin_amd64_pure_stripped/clusterawsadm clusterawsadm-darwin-amd64
+	install bazel-bin/cmd/clusterawsadm/linux_amd64_pure_stripped/clusterawsadm clusterawsadm-linux-amd64
+	install bazel-bin/cmd/clusterctl/darwin_amd64_pure_stripped/clusterctl clusterctl-darwin-amd64
+	install bazel-bin/cmd/clusterctl/linux_amd64_pure_stripped/clusterctl clusterctl-linux-amd64
```
liztio: the root seems like a weird place to put these binaries

cmd/clusterctl/BUILD
detiber: It is not supported for `go_library` afaict, but it is set on a `go_binary` here.

Makefile
detiber: fixed, thanks!

PR #353, opened 2018-10-31
Fixes machine delete

### LINKED ISSUE #312: Deleting a machine does not delete the instance
/kind bug

I created a cluster and one machine using the clusterctl workflow. I then applied a machine delete to the target cluster after the pivot. The controller showed no errors but the machine still existed in my aws account.

I expected the instance to be terminated.

chuckha:
This separates out updating of the machine and updating of the status.

Adds really basic tests for the machine actuator.

This PR allows the manager to actually delete machines when a machine delete is issued

Fixes #312 

**Release note**:
```release-note
NONE
```


pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -133,7 +133,11 @@ func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 		return errors.Wrap(err, "failed to reconcile LB attachment")
 	}
 
-	if err := a.storeMachineStatus(machine, status, true); err != nil {
+	if err := a.updateMachine(machine); err != nil {
+		glog.Errorf("failed to update machine %q in namespace %q trying to set annotation: %v", machine.Name, machine.Namespace, err)
```
detiber: If this update fails, then `clusterctl` will still error out, instead should we add a retry that fetches the latest version of the machine object and applies the status/annotation updates?

This is going to get more important over time as we move some of the items that are only in status currently into the providerconfig/providerspec in order to allow for user-driven config.

chuckha: I don't quite follow what you're saying. Maybe this is worth opening an issue and we can fix that after this merges?

detiber: That may do the trick, since exists will return `false` if the status has not been updated.

detiber: Consider the `clusterctl` worklfow:
- clusterctl applies cluster.yaml
  - Create cluster
    - Update fails (results in missing annotation that clusterctl looks for)
    - Status update succeeds
  - Reconcile triggers Update, since Exists returns true.
    - Update succeeds (does not apply clusterctl required annotation)
- `clusterctl` times out waiting for machine annotation to be present

If the update failure instead throws a retry error, then it should be:
- clusterctl applies cluster.yaml
  - Create cluster
    - Update fails (results in missing annotation that clusterctl looks for)
    - Status update succeeds
  - Reconcile triggers Create, since Exists returns true.
    - Update succeeds (applies clusterctl required annotation)
    - Status update succeeds (or fails)
- `clusterctl` finds machine annotation and continues

chuckha: ok, that makes a lot of sense. i appreciate you taking the time to write that out. I'll fix it!

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -133,7 +133,14 @@ func (a *Actuator) Create(cluster *clusterv1.Cluster, machine *clusterv1.Machine
...
+		return &controllerError.RequeueAfterError{
+			RequeueAfter: time.Minute,
+		}
+	}
+
+	if err := a.storeMachineStatus(machine, status); err != nil {
```
detiber: The machine update above should also take care of updating the status, is that not the case?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -169,8 +176,10 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster, machine *clusterv1.Machine
 		return errors.Wrap(err, "failed to get cluster provider config")
 	}
 
+	// status.InstanceID is nil, so don't do this
 	if status.InstanceID == nil {
```
detiber: Would it make sense to try and lookup by tags here? Currently we could potentially orphan an instance if create is called, the instance is successfully created, the object update fails, delete is processed before the requeued create is processed.

pkg/cloud/aws/actuators/machine/actuator.go
chuckha: That's correct, that is not the case. I suspect it has something to do with CRDs and subresources, but I'm unfamiliar with the depths of CRDs

pkg/cloud/aws/actuators/machine/actuator.go
chuckha: yes, I agree. can that be a followup issue or should it be part of this PR?

PR #356, opened 2018-11-01
Update yaml to use official sigs.k8s.io fork

detiber:
Since upstream has official forked ghodss/yaml under sigs.k8s.io, update to use the official fork.

Other changes are formatting changes that gazelle applied when `make dep-install` was run.


detiber:
/hold
to update after kubernetes-sigs/yaml#1 merges

PR #360, opened 2018-11-05
Add generate-iam-policy-docs command

### LINKED ISSUE #190: Extend clusterawsadm to show IAM policy templates for non-CloudFormation users
/kind feature

Given that the IAM policy is a code construct, it should be possible to show
the IAM policy document without the surrounding CloudFormation for users using
Terraform, the console or other tooling.

**Environment:**

ashish-amarnath:
* register generate-iam-policy-docs clusterawsadm command
* implement generate-iam-policy-docs command

Allow clusterawsadm to generate JSON representation of  PolicyDocument for all ManagedIAMPolicies 
Fixes #190 

**Release note**:

```release-note
`clusterawsadm` can now generate JSON representation of  PolicyDocument for all ManagedIAMPolicies 
```

### DIFF
--- a/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
+++ b/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
@@ -66,6 +66,7 @@ func RootCmd() *cobra.Command {
 	}
 	newCmd.AddCommand(generateCmd())
 	newCmd.AddCommand(createStackCmd())
+	newCmd.AddCommand(generateIAMPolicyDocJSON())
 	newCmd.AddCommand(encodeAWSSecret())
 	newCmd.AddCommand(generateAWSDefaultProfile())
 	return newCmd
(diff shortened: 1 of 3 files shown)


cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
```diff
@@ -142,6 +143,60 @@ func createStackCmd() *cobra.Command {
 	return newCmd
 }
 
+func generateJSON() *cobra.Command {
+	newCmd := &cobra.Command{
+		Use:   "generate-bootstrap-json [Directory for JSON]",
```
detiber: `generate-bootstrap-json` seems a bit redundant in this context, since the full command is `clusterawsadm alpha bootstrap generate-bootstrap-json`, what about using `generate-cloudformation-template` instead?

pkg/cloud/aws/services/cloudformation/bootstrap.go
```diff
@@ -281,6 +285,45 @@ func cloudProviderNodeAwsPolicy() *iam.PolicyDocument {
 	}
 }
 
+func makePath(dir string, fileName string) string {
+	if strings.HasSuffix(dir, "/") {
```
detiber: Can [path.Join](https://golang.org/pkg/path/#Join) be used here instead?

pkg/cloud/aws/services/cloudformation/bootstrap.go
```diff
@@ -281,6 +285,45 @@ func cloudProviderNodeAwsPolicy() *iam.PolicyDocument {
...
+	}
+
+	return jsonBytes, err
+}
+
+// ExportBootstrapTemplate exports bootstrap AWS resources to JSON
+func (s *Service) ExportBootstrapTemplate(resourceDir string, accountID string) error {
+
+	template := BootstrapTemplate(accountID)
+	for rn, rv := range template.Resources {
```
detiber: Any reason not to write a single template out using `template.JSON()`?

randomvariable:
Other than Jason's comments, in order to complete #190, will need to add documentation to say something along the lines of "if you don't want to use cloudformation, look at the JSON output from <command>"

pkg/cloud/aws/services/cloudformation/bootstrap.go
ashish-amarnath: Thank you! I tried searching for this API and clearly, I didn't look long enough :)

pkg/cloud/aws/services/cloudformation/bootstrap.go
ashish-amarnath: It seemed more tractable having a JSON file per resource. That said, I don't have a strong preference one way or the other.

randomvariable:
I've had a go at this branch to see what the UX is like.

The problem here is that this command doesn't create anything usable by an AWS operator.

The fragments of JSON are insufficient to properly form a CloudFormation template, and it doesn't resolve #190 in an easy-to-understand way.

A single CFN stack should be all that's required for a typical user who hasn't got an opinion on their favourite cloud deployment tool.

Someone who wants to use Terraform, Pulumi, AWS CDK, or their own custom CFN stack would want just the policy documents without the surrounding CloudFormation fragment.

In addition, this command behaves like the `create-stack` command in assuming AWS environment variables, instead of directly requiring an AWS Account ID like the existing `generate-cloudformation` script.

`generate-bootstrap-json` also doesn't properly describe that this creates CFN fragments that need to be merged into a single template.

I'll place some more information on #190 on how the UX should probably look like

cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
ashish-amarnath: Landed on `generate-iam-policy-docs`. WDYT?

PR #364, opened 2018-11-07
[DOCS] More info on minikube and aws user in the getting started doc

echarles:

More info on minikube and aws user in the getting started doc.


randomvariable:
@echarles , how do you feel about the suggested changes. Happy to lgtm otherwise

PR #384, opened 2018-11-16
Do not bail early in Machine create

chuckha:
* Fix kubeadm type to successfully join a node

This PR fixes the error when a node is joining:

```no kind "NodeConfiguration" is registered for version "kubeadm.k8s.io/v1alpha3" in scheme "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/scheme/scheme.go:31"```

Fixes #

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/services/userdata/node.go
+++ b/pkg/cloud/aws/services/userdata/node.go
@@ -37,9 +37,9 @@ HOSTNAME="$(curl http://169.254.169.254/latest/meta-data/local-hostname)"
 cat >/tmp/kubeadm-node.yaml <<EOF
 ---
 apiVersion: kubeadm.k8s.io/v1alpha3
-kind: NodeConfiguration
+kind: JoinConfiguration
 token: {{.BootstrapToken}}
-discoveryTokenAPIServers: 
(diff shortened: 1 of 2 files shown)


pkg/cloud/aws/services/userdata/node.go
```diff
@@ -37,9 +37,9 @@ HOSTNAME="$(curl http://169.254.169.254/latest/meta-data/local-hostname)"
 cat >/tmp/kubeadm-node.yaml <<EOF
 ---
 apiVersion: kubeadm.k8s.io/v1alpha3
-kind: NodeConfiguration
+kind: JoinConfiguration
```
vincepri: How is this different than before? Is there any upside to it?

chuckha: NodeConfiguration is not a valid type

vincepri: How would it work if this isn't correct? :/

vincepri: Interesting, I think NodeConfiguration might have been only available in v1alpha2

PR #425, opened 2018-11-29
Record basic events

### LINKED ISSUE #127: Events should be recorded into the events API

vincepri:

Fixes #127 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -194,7 +194,13 @@ func (s *Service) createInstance(machine *actuators.MachineScope, token string) 
 		input.KeyName = aws.String(defaultSSHKeyName)
 	}
 
-	return s.runInstance(machine.Role(), input)
+	out, err := s.runInstance(machine.Role(), input)
+	if err != nil {
+		return nil, err
+	}
+
+	record.Eventf(machine.Machine, "CreatedInstance", "Created new %s instance with id %q", machine.Role(), out.ID)
+	return out, nil
 }
 
 // TerminateInstance terminates an EC2 instance.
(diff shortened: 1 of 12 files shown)


vincepri:
/hold

Putting this on hold because seems it's throwing some errors instead of actually recording it.

vincepri:
@detiber ptal

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -309,6 +311,7 @@ func (s *Service) runInstance(i *v1alpha1.Instance) (*v1alpha1.Instance, error)
 
 	s.scope.EC2.WaitUntilInstanceRunning(&ec2.DescribeInstancesInput{InstanceIds: []*string{out.Instances[0].InstanceId}})
 
+	record.Eventf(s.scope.Cluster, "CreatedInstance", "Created new %s instance %q", role, *out.Instances[0].InstanceId)
```
detiber: This will cause Machine-based instance creation/deletion (control plane, workers) to associate their events with the Cluster instead of the Machine.

To start, it might be a good idea to limit the events to the actuator-based methods, and then we can expand usage from there.

vincepri: Sounds good, the reason this was interesting is that it actually prints when a bastion host is spun up. I can change it to be limited to only machine-based ones instead and move the bastion creation event to the reconcile loop.

pkg/record/recorder.go
```diff
@@ -65,7 +49,7 @@ func Event(object runtime.Object, reason, message string) {
 
 // Eventf is just like Event, but with Sprintf for the message field.
 func Eventf(object runtime.Object, reason, message string, args ...interface{}) {
-	defaultRecorder.Eventf(object, corev1.EventTypeNormal, strings.Title(reason), message, args)
```
chuckha: Since there is a defaultRecorder, that implies to me there could be a customized recorder. This is a very similar pattern to the http client that the net/http package exposes, but, like the net/http package, we should also export our DefaultRecorder so people can use methods on it if they don't care.

vincepri: Users can customize the internal recorder through the Init method offered by this package. I usually try to avoid exposing too much unless strictly necessary. Which seems fine in this case 😃

PR #429, opened 2018-11-30
Update Makefile for easier DX

vincepri:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -147,8 +147,13 @@ manifests: cmd/clusterctl/examples/aws/out/credentials ## Generate manifests for
 	kustomize build vendor/sigs.k8s.io/cluster-api/config/default/ >> cmd/clusterctl/examples/aws/out/provider-components.yaml
 
 .PHONY: manifests-dev
-manifests-dev: ## Push development manifest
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -147,9 +147,14 @@ manifests: cmd/clusterctl/examples/aws/out/credentials ## Generate manifests for
...
 
 .PHONY: manifests-dev
-manifests-dev: ## Push development manifest
+manifests-dev: dep-ensure dep-install binaries-dev ## Builds development manifests
 	MANAGER_IMAGE=$(DEV_MANAGER_IMAGE) $(MAKE) manifests
 
+.PHONY: binaries-dev
+binaries-dev: ## Builds and installs the binaries on the local GOPATH
+	go get -v ./...
+	go install -v ./...
```
detiber: Why not use the existing bazel build targets here?

vincepri: That's a good question, I've been using the binaries on the GOPATH for `clusterctl` and `manager` so this would help in rebuilding those.

detiber: Please add a comment that we should reconcile this with the bazel built binaries, otherwise lgtm

vincepri: Done :)

Issue #271, opened 2018-10-17
PR Blocking test(s)


detiber:
/kind feature

create and wire up pr-blocking integration tests

/assign @liztio

liztio:
what sort of things qualify as "integration tests" here? What's our test for completion?

detiber:
@liztio ideally we'd do a live test where we actually exercise the creation of a cluster using cluster-api and perform a minimal smoke test that the cluster appears to be functional. I do not think that this needs to include the full clusterctl workflow, since that introduces additional cleanup/teardown issues and also drastically increases the test time.

timothysc:
In an ideal world we would test the following on a per PR basis: 

- [MVP] Behavioral Tests 
    -  Infrastructure (vpc, etc) has been stood up according to the declarative spec 
        - Traffic inside of the vpc are routable 
        - Check the health of etcd 
        - At a minimum, we need verification of (quick-test)  

--- 

As a follow on we will open an issue to enumerate the future test blocking issues.
- [Future] Integration Tests 
   - Standup, and Teardown 
   - (TODO) Define the breakout here.
- [Future] All behavioral level conformance tests are run (sonobuoy)
- [Future] Traffic has been verified to (ingress is accessible)

liztio:
update on this: we're blocked on better aws resource management, otherwise we'd run out of resources pretty much immediately in CI

current plan is to add AWS support to test-infra's existing resource manager, boskos

ncdc:
@detiber please reassign to @akutz

chuckha:
/close

we can iterate on existing tests now!

PR #440, opened 2018-12-04
Add a single, simple ginkgo test: Make sure our set comes up

liztio:

Our existing E2E 

Part of  (but not all of) #271 

Why did I add Ginkgo?
1. It's what upstream k8s uses for E2E tests
2. I like it

**Release note**:
```release-note
NONE
```


test/e2e/util/kind/setup.go
```diff
@@ -0,0 +1,129 @@
...
+	c.applyYAML()
+}
+
+// Teardown attempts to delete the KIND cluster
+func (c *Cluster) Teardown() {
+	c.run(exec.Command(*kindBinary, "delete", "cluster"))
+	os.RemoveAll(c.tmpDir)
+}
+
+// a\pplyYAML takes the provided awsProviderYAML and clusterAPIYAML and applies them to a cluster given by the kubeconfig path kubeConfig.
```
vincepri: typo

test/e2e/metacluster_test.go
```diff
@@ -0,0 +1,61 @@
+package e2e_test
```
vincepri: Missing copyright header, also what do you think about adding `integration` build tag so these tests won't run locally in the dev environment unless we're explicit about it?

Same comment applies to `e2e_suite_test.go`.

liztio: It looks like rules_go doesn't play great with the integration build tags. I can add `size="large"` to the bazel config though, will that help?

vincepri: I'm not sure, usually dev environments run test using the `go` command. If there is no better way that's fine, maybe I can play around with it later to see how to exclude these locally.

cmd/clusterctl/examples/aws/aws_manager_image_patch.yaml.template
```diff
@@ -8,4 +8,5 @@ spec:
     spec:
       containers:
       - image: ${MANAGER_IMAGE}
+        imagePullPolicy: IfNotPresent
```
vincepri: I believe this is already set here https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/config/manager/manager.yaml#L57

liztio: @vincepri it's set to Always there. Do you want me to change it to IfNotPresent in the other location?

test/e2e/metacluster_test.go
liztio: If you run `go test` on an individual package, it won't get pulled in. If you're running the whole suite, I recommend using `bazel test` anyway! It's much faster!

cmd/clusterctl/examples/aws/aws_manager_image_patch.yaml.template
vincepri: That would break the development experience, is there a reason why `Always` isn't suitable here?

liztio: Because during integration testing there's no repository available.

detiber: @vincepri maybe we should leverage a kustomize patch here to set it to `Always` for dev workflows. `IfNotPresent` makes a lot more sense for end users.

liztio: Or just change your dev workflow to use kind :wink:

Issue #430, opened 2018-11-30
Cluster and machine status fails to update on first run


vincepri:
/kind bug

**What steps did you take and what happened:**
Create a cluster, on the first run, the status doesn't get updated properly.
```
E1130 17:44:52.063316       1 machine_scope.go:121] [machinescope] failed to store provider status for machine "aws-controlplane-s6qx9" in namespace "default": Operation cannot be fulfilled on machines.cluster.k8s.io "aws-controlplane-s6qx9": the object has been modified; please apply your changes to the latest version and try again
E1130 17:44:52.070631       1 scope.go:164] [scope] failed to store provider status for cluster "test1" in namespace "default": Operation cannot be fulfilled on clusters.cluster.k8s.io "test1": the object has been modified; please apply your changes to the latest version and try again
```

**What did you expect to happen:**
Status should get updated without an error.

vincepri:
The cluster status is fixed. Machine is next, although from yesterday's tests it seems the machine gets updated somewhere else.

PR #443, opened 2018-12-04
Fix status update error on first run

vincepri:

This PR fixes the status update error that happens when the actuator runs. It'll fetch the latest cluster object from the api and update the status.

Further investigation is needed for the other issues.

Related to #430

**Release note**:

```release-note

```


detiber:
@chuckha exactly, because we are updating the machine object itself before attempting to update the status.

PR #450, opened 2018-12-06
Add ability to configure cluster name, remove generateName

vincepri:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterctl/examples/aws/machines.yaml.template
+++ b/cmd/clusterctl/examples/aws/machines.yaml.template
@@ -4,7 +4,7 @@ items:
   - apiVersion: "cluster.k8s.io/v1alpha1"
     kind: Machine
     metadata:
-      generateName: aws-controlplane-
+      name: controlplane-0
       labels:
         set: controlplane
     spec:
@@ -21,7 +21,7 @@ items:
   - apiVersion: "cluster.k8s.io/v1alpha1"
     kind: Machine
     metadata:
-      generateName: aws-node-
+      generateName: node-
       labels:
         set: node
     spec:
--- a/cmd/clusterctl/examples/aws/cluster.yaml.template
+++ b/cmd/clusterctl/examples/aws/cluster.yaml.template
@@ -1,7 +1,7 @@
 apiVersion: "cluster.k8s.io/v1alpha1"
 kind: Cluster
 metadata:
-  name: test1
+  name: ${CLUSTER_NAME}
 spec:
   clusterNetwork:
     services:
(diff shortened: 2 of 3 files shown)


cmd/clusterctl/examples/aws/machines.yaml.template
```diff
@@ -21,7 +21,7 @@ items:
   - apiVersion: "cluster.k8s.io/v1alpha1"
     kind: Machine
     metadata:
-      generateName: aws-node-
+      name: aws-node-${CLUSTER_NAME}-1
```
detiber: I think we should still use `generateName` for nodes, that way creating a new node is simply a copy/paste.

cmd/clusterctl/examples/aws/machines.yaml.template
```diff
@@ -4,7 +4,7 @@ items:
   - apiVersion: "cluster.k8s.io/v1alpha1"
     kind: Machine
     metadata:
-      generateName: aws-controlplane-
+      name: aws-controlplane-${CLUSTER_NAME}
```
detiber: I would suggest using a static index here rather than `${CLUSTER_NAME}`.

We really should be prefixing the aws instance `Name` tags with the cluster name rather than just passing in the machine name.

vincepri:
@detiber Fixed comments, ptal

cmd/clusterctl/examples/aws/machines.yaml.template
```diff
@@ -4,7 +4,7 @@ items:
   - apiVersion: "cluster.k8s.io/v1alpha1"
     kind: Machine
     metadata:
-      generateName: aws-controlplane-
+      name: aws-controlplane-${CLUSTER_NAME}-0
```
detiber: My apologies, I meant that we should be adding the cluster name to the instance through the actuator and not the machine template (which can be handled through a separate issue/PR)

vincepri: I guess this should still be ok for HA?

I agree we should change the Name in the actuator though

detiber: > I guess this should still be ok for HA?

Yes, because for HA we will have 3 distinct machines for the control plane.

cmd/clusterctl/examples/aws/machines.yaml.template
```diff
@@ -4,7 +4,7 @@ items:
   - apiVersion: "cluster.k8s.io/v1alpha1"
     kind: Machine
     metadata:
-      generateName: aws-controlplane-
+      name: controlplane-${CLUSTER_NAME}-0
```
detiber: CLUSTER_NAME template still showing here for me

PR #452, opened 2018-12-07
Fix cluster/machine status updates

### LINKED ISSUE #430: Cluster and machine status fails to update on first run

vincepri:

Fixes #430 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/apis/awsprovider/v1alpha1/register.go
+++ b/pkg/apis/awsprovider/v1alpha1/register.go
@@ -114,7 +114,26 @@ func EncodeMachineStatus(status *AWSMachineProviderStatus) (*runtime.RawExtensio
 	}, nil
 }
 
-// EncodeClusterStatus marshals the cluster status
+// EncodeMachineSpec marshals the machine provider spec.
+func EncodeMachineSpec(spec *AWSMachineProviderSpec) (*runtime.RawExtension, error) {
+	if spec == nil {
+		return &runtime.RawExtension{}, nil
+	}
+
+	var rawBytes []byte
+	var err error
+
+	//  TODO: use apimachinery conversion https://godoc.org/k8s.io/apimachinery/pkg/runtime#Convert_runtime_Object_To_runtime_RawExtension
+	if rawBytes, err = json.Marshal(spec); err != nil {
+		return nil, err
+	}
+
+	return &runtime.RawExtension{
+		Raw: rawBytes,
+	}, nil
+}
+
+// EncodeClusterStatus marshals the cluster status.
 func EncodeClusterStatus(status *AWSClusterProviderStatus) (*runtime.RawExtension, error) {
 	if status == nil {
 		return &runtime.RawExtension{}, nil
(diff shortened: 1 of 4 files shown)


pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -94,31 +95,33 @@ func (m *MachineScope) Region() string {
...
 }
 
 func (m *MachineScope) Close() {
 	if m.MachineClient == nil {
 		return
 	}
 
-	if _, err := m.MachineClient.Update(m.Machine); err != nil {
-		klog.Errorf("[machinescope] failed to update machine: %v", err)
+	latestMachine, err := m.MachineClient.Get(m.Name(), metav1.GetOptions{})
```
detiber: Shouldn't this be done after the update? It looks like the Update below is actually a noop, and if it somehow did manage to Update the object it would cause the following UpdateStatus to fail.

pkg/cloud/aws/actuators/scope.go
```diff
@@ -166,11 +150,18 @@ func (s *Scope) Close() {
 		return
 	}
 
-	if err := s.storeClusterConfig(); err != nil {
+	latestCluster, err := s.ClusterClient.Get(s.Name(), metav1.GetOptions{})
```
detiber: same comment here as with machine update

pkg/cloud/aws/actuators/machine_scope.go
vincepri: During my tests, it actually showed that the machine somehow gets updated before this method gets called. The `Update` below returns the latest version of the Machine (if successful) with the changes.

After trying and debugging in different stages, the conclusion I came up with is that these calls have to be chained. There is definitely something weird about the object being updated somewhere else during the first run, that's why we need to get the latest version. That said, doing the call even on subsequent calls seems to be fine.

vincepri:
@detiber This is good to go I believe. It requires https://github.com/kubernetes-sigs/cluster-api/pull/627 to be merged and vendor updated to function on the first run.

pkg/apis/awsprovider/v1alpha1/register.go
```diff
@@ -133,17 +152,17 @@ func EncodeClusterStatus(status *AWSClusterProviderStatus) (*runtime.RawExtensio
 	}, nil
 }
 
-// EncodeClusterConfig marshals the cluster config.
-func EncodeClusterConfig(status *AWSClusterProviderSpec) (*runtime.RawExtension, error) {
```
ashish-amarnath: Thank you for fixing my code :)

PR #463, opened 2018-12-25
Control plane node join

### LINKED ISSUE #413: Add machine actuator support for control plane join
/kind feature

The machine actuator should support joining additional control plane instances.

Depends on #412

ashish-amarnath:
Implements support for control plane node join

Fixes #413

**Release note**:

```release-note
cluster-api-provider-aws now supports creating clusters with HA control plane using clusterctl phases.
Steps to create clusters with HA control plane are documented <Link to readme generated as a fix for #414 >
```


ashish-amarnath:
@chuckha  Thanks for the pointers on #413 
I took a stab at this to get the scaffolding and initial ideas in place.
Publishing this PR to get some feedback on the approach.

/assign @chuckha 
cc @detiber

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -78,29 +79,24 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
-		coreClient, err := corev1.NewForConfig(clientConfig)
-		if err != nil {
-			return errors.Errorf("failed to initialize new corev1 client: %+v", err)
-		}
-
-		bootstrapToken, err = tokens.NewBootstrap(coreClient, 10*time.Minute)
-		if err != nil {
-			return errors.Errorf("failed to create new bootstrap token: %+v", err)
+	case "controlplane":
+		_, err := a.GetKubeConfig(cluster, nil)
```
ashish-amarnath: This is not a correct check for a control plane node join.

pkg/apis/awsprovider/v1alpha1/awsclusterproviderstatus_types.go
```diff
@@ -30,8 +30,9 @@ type AWSClusterProviderStatus struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Network Network  `json:"network,omitempty"`
-	Bastion Instance `json:"bastion,omitempty"`
+	Network            Network                   `json:"network,omitempty"`
+	Bastion            Instance                  `json:"bastion,omitempty"`
+	ControlPlaneStatus ClusterControlPlaneStatus `json:"clusterControlPlaneStatus,omitempty"`
```
vincepri: Is the `Status` suffix necessary here?

pkg/apis/awsprovider/v1alpha1/types.go
```diff
@@ -102,6 +102,12 @@ type Network struct {
 	APIServerELB ClassicELB `json:"apiServerElb,omitempty"`
 }
 
+// ClusterControlPlaneStatus encapsulates control-plane status for the cluster
+type ClusterControlPlaneStatus struct {
+	// ControlPlaneNodeCount is the count of the number of control-plane nodes in the cluster.
+	ControlPlaneNodeCount int `json:"controlPlaneNodeCount"`
```
vincepri: Should this just be `NodeCount` given that the struct already sets the context?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -131,6 +123,34 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
 	return nil
 }
 
+func (a *Actuator) getWorkerNodeToken(cluster *clusterv1.Cluster, controlPlaneURL string) (string, error) {
+	var bootstrapToken string
```
vincepri: Can this value be declared at the bottom? Doesn't seem necessary to be declared here, but I might be wrong.

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -128,4 +129,7 @@ func (m *MachineScope) Close() {
 	if err != nil {
 		klog.Errorf("[machinescope] failed to store provider status for machine %q in namespace %q: %v", m.Machine.Name, m.Machine.Namespace, err)
 	}
+
+	// control plane node count is a status of the cluster. update that incase a new control plane node was created.
+	m.Scope.Close()
```
vincepri: This was removed because it causes races between the machine and cluster actuator.

pkg/cloud/aws/actuators/scope.go
```diff
@@ -88,22 +88,27 @@ type Scope struct {
 	ClusterStatus *v1alpha1.AWSClusterProviderStatus
 }
 
+// ClusterControlPlaneStatus returns the cluster controlplane status object.
+func (s *Scope) ClusterControlPlaneStatus() *v1alpha1.ClusterControlPlaneStatus {
```
vincepri: Could this be simplified just to `ControlPlane()`?

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -138,41 +139,55 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
...
-			CAKey:             string(s.scope.ClusterConfig.CAPrivateKey),
-			ELBAddress:        s.scope.Network().APIServerELB.DNSName,
-			ClusterName:       s.scope.Name(),
-			PodSubnet:         s.scope.Cluster.Spec.ClusterNetwork.Pods.CIDRBlocks[0],
-			ServiceSubnet:     s.scope.Cluster.Spec.ClusterNetwork.Services.CIDRBlocks[0],
-			ServiceDomain:     s.scope.Cluster.Spec.ClusterNetwork.ServiceDomain,
-			KubernetesVersion: machine.Machine.Spec.Versions.ControlPlane,
-		})
+		var userData string
+		var err error
```
vincepri: Why were these moved here?

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -138,41 +139,55 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
...
+				KubernetesVersion: machine.Machine.Spec.Versions.ControlPlane,
+			})
 
-		if err != nil {
-			return input, err
+			if err != nil {
+				return input, err
+			}
 		}
+		controlPlaneNodeCount++
```
vincepri: Should we make these a function on the ControlPlane status instead of modifying the value?

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -204,6 +224,7 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 	if err != nil {
 		return nil, err
 	}
+	machine.ClusterControlPlaneStatus().ControlPlaneNodeCount = controlPlaneNodeCount
```
vincepri: Same as the other comment, if the increments/decrements are functions we wouldn't need to do this assignment.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: This was a copy pasta :)

pkg/apis/awsprovider/v1alpha1/awsclusterproviderstatus_types.go
ashish-amarnath: Not taking this approach anymore.

pkg/apis/awsprovider/v1alpha1/types.go
ashish-amarnath: Dropping this approach

pkg/cloud/aws/actuators/machine_scope.go
ashish-amarnath: Dropping this approach

pkg/cloud/aws/services/ec2/instances.go
ashish-amarnath: To share `userData` between controlplane create and controlplane join

pkg/cloud/aws/services/ec2/instances.go
ashish-amarnath: Dropping this approach

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -131,6 +123,34 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	kubeConfig, err := a.GetKubeConfig(cluster, nil)
+	if err != nil {
+		return bootstrapToken, errors.Errorf("failed to retrieve kubeconfig during machine creation: %+v", err)
```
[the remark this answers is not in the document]
ashish-amarnath: I'll do a sweep for this once this PR is in the closer to the end.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -131,6 +123,34 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	}
+
+	coreClient, err := corev1.NewForConfig(clientConfig)
+	if err != nil {
+		return bootstrapToken, errors.Errorf("failed to initialize new corev1 client: %+v", err)
+	}
+
+	bootstrapToken, err = tokens.NewBootstrap(coreClient, 10*time.Minute)
```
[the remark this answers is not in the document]
ashish-amarnath: changed

ashish-amarnath:
Hitting this failure with `kubeadm join`
```
[   13.081046] cloud-init[1303]: kubeadm join --config /tmp/kubeadm-controlplane-join-config.yaml --experimental-control-plane
[   13.233587] cloud-init[1303]: panic: runtime error: invalid memory address or nil pointer dereference
```
ToDo:
* make `--experimental-control-plane` part of the join configuration
* get this to run successfully

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -59,6 +61,51 @@ func NewActuator(params ActuatorParams) *Actuator {
...
+	var cpm []*clusterv1.Machine
+	for _, m := range machineList.Items {
+		if m.Spec.Versions.ControlPlane != "" {
+			cpm = append(cpm, m.DeepCopy())
+		}
+	}
+	return cpm
+}
+
+func isMachinesSame(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
```
chuckha: Rename this to `machinesEqual`.

Add a documentation note that we are defining equality as name and namespace are equivalent and that we are not checking any other fields.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -59,6 +61,51 @@ func NewActuator(params ActuatorParams) *Actuator {
...
+	return m1.Name == m2.Name && m1.Namespace == m2.Namespace
+}
+
+func (a *Actuator) isNodeJoin(controlPlaneMachines []*clusterv1.Machine, newMachine *clusterv1.Machine) (join bool, err error) {
```
chuckha: don't use named return values here

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -59,6 +61,51 @@ func NewActuator(params ActuatorParams) *Actuator {
...
+	err = nil
+	switch newMachine.ObjectMeta.Labels["set"] {
+	case "node":
+		join = true
```
chuckha: `return true, nil`

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -59,6 +61,51 @@ func NewActuator(params ActuatorParams) *Actuator {
...
+	case "controlplane":
+		join = true
```
chuckha: delete this and directly return the values instead

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -59,6 +61,51 @@ func NewActuator(params ActuatorParams) *Actuator {
...
+		// join = false, if:
+		//		1. len(controlPlaneMachines) == 1 && controlPlaneMachines[0] == newMachine
+		if len(controlPlaneMachines) == 1 && isMachinesSame(controlPlaneMachines[0], newMachine) {
+			join = false
```
chuckha: directly return false and nil here. after this `if` block, we can directly return true and nil instead of setting default values and implicitly returning

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -59,6 +61,51 @@ func NewActuator(params ActuatorParams) *Actuator {
...
+		}
+		// 		TODO: ashish-amarnath 2. if none of the controlPlaneMachines exist
+
+		klog.V(2).Infof("Machine %q should join the controlplane: %t", newMachine.Name, join)
+	default:
+		errMsg := fmt.Sprintf("Unknown value %q for label \"set\" on machine %q, skipping machine creation", newMachine.ObjectMeta.Labels["set"], newMachine.Name)
+		klog.Errorf(errMsg)
+		err = errors.Errorf(errMsg)
```
chuckha: have an explicit return here instead of assigning err

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -77,33 +124,31 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+		return errors.Errorf("failed to retrieve machines for cluster %q: %v", cluster.Name, err)
+	}
+	controlPlaneMachines := a.getControlPlaneMachines(clusterMachines)
+	isNodeJoin, err := a.isNodeJoin(controlPlaneMachines, machine)
+	if err != nil {
+		return errors.Errorf("Failed to determine whther machine %q should join cluster %q: %v", machine.Name, cluster.Name, err)
+	}
 
-		coreClient, err := corev1.NewForConfig(clientConfig)
+	bootstrapToken := ""
```
chuckha: let's keep this as `var bootstrapToken string`

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -77,33 +124,31 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	if isNodeJoin {
+		bootstrapToken, err = a.getNodeJoinToken(cluster, controlPlaneURL)
 		if err != nil {
-			return errors.Errorf("failed to initialize new corev1 client: %+v", err)
+			klog.Errorf("failed to retrieve token to create machine %q: %v", machine.Name, err)
```
chuckha: I don't think we need klog here, we can instead return the error

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -138,48 +139,69 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
...
+			userData, err = userdata.JoinControlPlane(&userdata.ContolPlaneJoinInput{
+				CACert:         string(s.scope.ClusterConfig.CACertificate),
+				CAKey:          string(s.scope.ClusterConfig.CAPrivateKey),
+				CACertHash:     caCertHash,
+				BootstrapToken: bootstrapToken,
+				ELBAddress:     s.scope.Network().APIServerELB.DNSName,
+				KubeConfig:     kubeConfig,
+			})
+			if err != nil {
+				return input, err
```
chuckha: make sure this err is wrapped in userdata call

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -191,6 +213,11 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 		}
 
 		input.UserData = aws.String(userData)
+
+	default:
+		errMsg := fmt.Sprintf("Unknown node role %q", machine.Role())
+		klog.Error(errMsg)
+		return nil, fmt.Errorf(errMsg)
```
chuckha: we don't need to klog this deep, just return an error and please use the `github.com/pkg/errors` library

pkg/cloud/aws/services/userdata/controlplane.go
```diff
@@ -55,6 +55,49 @@ nodeRegistration:
 EOF
 
 kubeadm init --config /tmp/kubeadm.yaml
+
+tar -cvzf /etc/kubernetes/pki.tar.gz /etc/kubernetes/pki/*
+kubectl -n kube-system --kubeconfig /etc/kubernetes/admin.conf create secret generic kubeadm-certs --from-file=/etc/kubernetes/pki.tar.gz
```
detiber: This is an interesting approach for managing the secrets, but it would probably be better if we took a different approach. I can see a couple of different ways of solving this without adding a secret post-init:

- pre-generate the etcd ca keypair, front-proxy ca keypair, and sa(service account) keypair in the cluster controller and store in the Cluster ProviderSpec
  - inject them into the init config like we are currently doing with the CA cert/key
  - pre-write the keypairs to the filesystem before calling kubeadm join
  - if other certificates are needed prior to calling kubeadm join, pregenerate them using kubeadm phases (kubeadm join should only really require the CAs, it would be a bug otherwise)
- Instead of modifying the cluster actuator, it would be possible to fetch the required certificates using the kubernetes client from within the machine actuator, then inject the certificates using the init script on the host prior to running kubeadm join.

chuckha: i like the simplicity of this approach. can you elaborate on why you think it would be better if we took a different approach?

i'm guessing security issues? secrets aren't actually secret and it would be far too easy to get these secrets and do bad things. But it's not that different if we store the values in the cluster provider spec. That's still plaintext secret storage in etcd.

randomvariable: I would expect operators to use envelope encryption for secrets in their bootstrap / manager clusters, so it'll be justifiable to store ca certificates as secrets, less so as values in the provider spec.

We can't pre-generate the non-authoritative certs yet because we don't know the IP addresses of the nodes beforehand, and this is particularly important for etcd, but reasonable to do so for the certificate authorities and service account key-pair.

For a solution that will also support autoscaling groups, the ideal way would be use the [instance identity document](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-identity-documents.html) as a host attestation mechanism, where a verifiable CSR is made, that the controller can sign.

detiber: So, there are a couple of different reasons why I'm suggesting adding it to the provider spec:
- It automatically provides an avenue for users to override the CA and SA keypairs to use custom values
- While we are embedding the certificates in the spec today, we can later update the spec to allow for either embedding or referencing an external source at a later point

I want to make sure that we are considering UX when we start to look at more secure private key storage, by allowing direct embedding of the cert material, we avoid having to pre-generate any additional secrets as part of the workflow and we the spec is updated to support remote references for secure material (k8s secrets or leveraging something AWS native), we can have the actuator move the cert material to the proper place as part of the workflow, or even have an admission webhook that does the job to avoid ever storing the cert material in etcd).

randomvariable: Happy with this approach. So we need to change the spec a bit now, at least, to hold all the key pairs we need for the HA control plane?

ashish-amarnath: I agree with the UX and the extendability aspect. However, IMO, we have a dependency on `kubeadm` here for generating the certs on when the first controlplane node runs the init. From reading the kubeadm docs, there seems to be a way to skip cert generation and run the other "phases". By doing this we will be deviating from the standard `kubeadm init` flow. Which, IMO, is OK as long as we call that out for UX.

Also, kubeadm will start handling the cert distribution for controlplane join in v.1.14. More info https://github.com/kubernetes/kubeadm/issues/1200
With this in mind and we aren't GA  yet, I think the cert distribution fix I have should be a stop-gap till we go to 1.14. WDYT?

detiber: > I agree with the UX and the extendability aspect. However, IMO, we have a dependency on kubeadm here for generating the certs on when the first controlplane node runs the init. From reading the kubeadm docs, there seems to be a way to skip cert generation and run the other "phases". By doing this we will be deviating from the standard kubeadm init flow. Which, IMO, is OK as long as we call that out for UX.

@ashish-amarnath If the keypair files already exist on disk in the default locations that kubeadm would use for creating them and are not close to expiration they will be picked up and used by kubeadm without having to run phases.

> Also, kubeadm will start handling the cert distribution for controlplane join in v.1.14. More info kubernetes/kubeadm#1200
With this in mind and we aren't GA yet, I think the cert distribution fix I have should be a stop-gap till we go to 1.14. WDYT?

Even if kubeadm handles the certificate copying for v1.14+ we would still benefit from having the ability for users to specify custom keypairs here.

ashish-amarnath: I've decoupled the certs into their individual secrets so that they can be each updated independent of the other secrets.

chuckha: +1 thanks for the reasoning @detiber. totally agree that changing the spec is the right way to go.

ashish-amarnath: So, I propose that we keep the spec change out of this PR and get the controlplane machine join scenario worked out. I'll follow this PR with the spec change.

pkg/cloud/aws/services/ec2/instances.go
ashish-amarnath: Outdated

pkg/cloud/aws/actuators/machine/actuator.go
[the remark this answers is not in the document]
ashish-amarnath: I did a sweep to wrap errors.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -131,6 +123,34 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	if err != nil {
+		return bootstrapToken, errors.Errorf("failed to retrieve kubeconfig during machine creation: %+v", err)
+	}
+
+	clientConfig, err := clientcmd.BuildConfigFromKubeconfigGetter(controlPlaneURL, func() (*clientcmdapi.Config, error) {
+		return clientcmd.Load([]byte(kubeConfig))
+	})
+
+	if err != nil {
+		return bootstrapToken, errors.Errorf("failed to retrieve kubeconfig during machine creation: %+v", err)
```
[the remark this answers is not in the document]
ashish-amarnath: Thanks for explaining! +1
Fixed now.

pkg/cloud/aws/services/userdata/controlplane.go
ashish-amarnath: Created https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/492 to address this.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -77,33 +130,30 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
-			return errors.Errorf("failed to retrieve kubeconfig during machine creation: %+v", err)
-		}
+	clusterMachines, err := scope.MachineClient.List(v1.ListOptions{})
+	if err != nil {
+		return errors.Wrapf(err, "failed to retrieve machines in cluster %q", cluster.Name)
+	}
+	controlPlaneMachines := a.getControlPlaneMachines(clusterMachines)
+	isNodeJoin, err := a.isNodeJoin(controlPlaneMachines, machine, cluster)
+	if err != nil {
+		return errors.Wrapf(err, "failed to determine whther machine %q should join cluster %q", machine.Name, cluster.Name)
```
daniel-garcia: whether vs whther

chuckha:
let's type better

/approve

PR #473, opened 2019-01-08
Installs binaries to gopath even if gopath is unset

### LINKED ISSUE #472: Have bazel install the built binaries into GOPATH by default
/kind feature

I mostly want to undo this commit https://github.com/chuckha/cluster-api-provider-aws/commit/55e14bf240f69eb044ae617222d4777aac610e14 except instead of assuming a GOPATH exists, use `go env GOPATH`/bin/.

chuckha:
Fixes #472

This PR installs built binaries into the GOPATH even in situations when the GOPATH is not set.

This basically reverts https://github.com/chuckha/cluster-api-provider-aws/commit/55e14bf240f69eb044ae617222d4777aac610e14 with the modification of working when GOPATH is unset.

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -68,18 +68,22 @@ check-install: ## Checks that you've installed this repository correctly
 .PHONY: manager
 manager: generate  ## Build manager binary.
 	bazel build //cmd/manager $(BAZEL_ARGS)
+	install bazel-bin/cmd/manager/*/manager $(shell go env GOPATH)/bin/aws-manager
 
 .PHONY: clusterctl
 clusterctl: generate ## Build clusterctl binary.
 	bazel build --workspace_status_command=./hack/print-workspace-status.sh //cmd/clusterctl $(BAZEL_ARGS)
+	install bazel-bin/cmd/clusterctl/*/clusterctl $(shell go env GOPATH)/bin/clusterctl
 
 .PHONY: clusterawsadm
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -68,18 +68,22 @@ check-install: ## Checks that you've installed this repository correctly
 .PHONY: manager
 manager: generate  ## Build manager binary.
 	bazel build //cmd/manager $(BAZEL_ARGS)
+	cp -f bazel-bin/cmd/manager/*/manager $(shell go env GOPATH)/bin/aws-manager
```
detiber: Using `install` might be better here since the permissions of the bazel built binary lacks write permissions.

chuckha: i do like `install` more than `cp -f`.

Makefile
```diff
@@ -68,18 +68,22 @@ check-install: ## Checks that you've installed this repository correctly
 .PHONY: manager
 manager: generate  ## Build manager binary.
 	bazel build //cmd/manager $(BAZEL_ARGS)
+	install bazel-bin/cmd/manager/*/manager $(shell go env GOPATH)/bin/aws-manager
```
detiber: Just noticed the wildcard here, this will cause issues if someone has run the release-binaries target and cross compiled binaries for other targets.

chuckha: ah hmm so it will

detiber: doh, didn't beat the merge...

PR #479, opened 2019-01-08
Lookup AMIs instead of hardcoding them.

### LINKED ISSUE #138: Images should be discoverable
/kind bug

There is no way to filter for images built with the packer script from this project. I looked for a script to do this in heptiolabs/wardroom that @liztio mentioned but could not find it.

I expect our images to be tagged with something easy to filter on.

There are a bunch of hacks to work around this issue. Namely hardcoding AMIs.

chuckha:
This is a first pass at looking up AMIs instead of hardcoding
a big case statement basd on region.

Fixes #138

This PR looks up AMIs for control plane nodes instead of relying on hardcoded case statement.

This logic requires coordination with the build/ directory because it uses the name to look up the actual AMI. There may be improvements and we want to implement this logic with the bastion hosts (I think) as well.

The goal is to iterate from here, this may not be perfect but until we release new images with a different title, this will work.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/ami.go
+++ b/pkg/cloud/aws/services/ec2/ami.go
@@ -16,43 +16,53 @@ limitations under the License.
 
 package ec2
 
+import (
+	"fmt"
+	"strings"
+
+	"k8s.io/klog"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/pkg/errors"
+)
+
+func amiName(platform, platformVersion, kubernetesVersion string) string {
+	return fmt.Sprintf("ami-%s-%s-%s-00-??????????", platform, platformVersion, strings.TrimPrefix(kubernetesVersion, "v"))
+}
+
 // defaultAMILookup returns the default AMI based on region
-func (s *Service) defaultAMILookup(region string) string {
-	//TODO(chuckha) Replace this function with a method to filter public images.
-	switch region {
-	case "ap-northeast-1":
-		return "ami-06525773f0332c0ca"
-	case "ap-northeast-2":
-		return "ami-0914a816a80ab8779"
-	case "ap-south-1":
-		return "ami-04b17f6875b4d9c29"
-	case "ap-southeast-1":
-		return "ami-0ccaa193408259c82"
-	case "ap-southeast-2":
-		return "ami-0385d030f1b7df12c"
-	case "ca-central-1":
-		return "ami-031ca32942ffbe25f"
-	case "eu-central-1":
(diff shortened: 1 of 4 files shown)


pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -106,11 +106,15 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 		Role:        aws.String(machine.Role()),
 	})
 
+	var err error
 	// Pick image from the machine configuration, or use a default one.
 	if machine.MachineConfig.AMI.ID != nil {
 		input.ImageID = *machine.MachineConfig.AMI.ID
 	} else {
-		input.ImageID = s.defaultAMILookup(machine.Region())
+		input.ImageID, err = s.defaultAMILookup("ubuntu", "18.04", machine.Machine.Spec.Versions.Kubelet)
```
randomvariable: This looks fine as a first approach, but can we put a TODO here about ubuntu 18.04 being hardcoded?

randomvariable: Actually, re-read this. this is fine.

pkg/cloud/aws/services/ec2/ami.go
```diff
@@ -16,43 +16,53 @@ limitations under the License.
...
+				Values: []*string{aws.String("x86_64")},
+			},
+			{
+				Name:   aws.String("state"),
+				Values: []*string{aws.String("available")},
+			},
+			{
+				Name:   aws.String("virtualization-type"),
+				Values: []*string{aws.String("hvm")},
+			},
```
vincepri: We need to hardcode the account id as a filter, based on these settings someone could hijack the same naming and filters and publish their own default AMI. AWS returns the most recent one.  @chuckha

pkg/cloud/aws/services/ec2/ami.go
```diff
@@ -16,43 +16,53 @@ limitations under the License.
...
+	"strings"
+
+	"k8s.io/klog"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/pkg/errors"
+)
+
+func amiName(platform, platformVersion, kubernetesVersion string) string {
```
detiber: I think we should probably look at different naming here rather than `platform` and `platformVersion` as it overloads the use of `platform` in Go

detiber: This method should also probably take the container runtime as an argument as well, I suspect there will be a use case in the future to have alternative container runtimes.

PR #485, opened 2019-01-10
Include owner as an AMI filter

chuckha:

Don't let anyone hijack the AMI image name pattern !

Will address other issues in a future PR this one just seemed not great to have.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/ami.go
+++ b/pkg/cloud/aws/services/ec2/ami.go
@@ -36,6 +36,10 @@ func (s *Service) defaultAMILookup(platform, platformVersion, kubernetesVersion 
 	describeImageInput := &ec2.DescribeImagesInput{
 		Filters: []*ec2.Filter{
 			{
+				Name:   aws.String("owner-id"),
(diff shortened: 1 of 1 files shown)


pkg/cloud/aws/services/ec2/ami.go
```diff
@@ -35,6 +35,10 @@ func amiName(platform, platformVersion, kubernetesVersion string) string {
 func (s *Service) defaultAMILookup(platform, platformVersion, kubernetesVersion string) (string, error) {
 	describeImageInput := &ec2.DescribeImagesInput{
 		Filters: []*ec2.Filter{
+			{
+				Name:   aws.String("owner-id"),
+				Values: []*string{aws.String("258751437250")},
```
randomvariable: Can we constantize this with a comment to say whose account this is?

chuckha: i can include this commit in the following cleanup PR instead since I need to clean this whole file up anyway

PR #510, opened 2019-01-22
[WIP] Switch bash to cloud config

### LINKED ISSUE #205: Switch machine actuator to use cloud-config rather than startup script
/kind bug

/priority critical-urgent

sfzylad:

Fixes #205 

**Release note**:

```release-note

```


randomvariable:
Bit more to this one.

@detiber's been working on a kubeadm module for cloud-init. Probably worth a sync up.

sfzylad:
@randomvariable, thanks for heads up.
@detiber, could you point me to the module you worked on?

vincepri:
Not sure of the current status, but there is an old PR open around cloud-init https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/96

sfzylad:
Yeah, I've found this one too and I wondered about the status as this PR since the last commit there has been done in September last year.

sfzylad:
Hi @ashish-amarnath 👋 

I've got node to work with cloud-config (I'm about to switch control plane too). It still uses `text/template` but makes it a bit less magical. Please tell me your thpughts.

pkg/cloud/aws/services/userdata/userdata.go
```diff
@@ -24,6 +24,23 @@ import (
 )
 
 const (
+	cloudConfigHeader = `## template: jinja
+#cloud-config
+#
+# Copyright 2018 by the contributors
```
detiber: I don't think we need to output a license header here

pkg/cloud/aws/services/userdata/node.go
```diff
@@ -39,6 +39,30 @@ nodeRegistration:
...
+          apiServerEndpoint: "{{.ELBAddress}}:6443"
+          caCertHashes:
+            - "{{.CACertHash}}"
+      nodeRegistration:
+        name: {{ "{{ ds.meta_data.hostname }}" }}
+        criSocket: /var/run/containerd/containerd.sock
+        kubeletExtraArgs:
+          cloud-provider: aws
+runcmd:
+  - kubeadm join --config /tmp/kubeadm-node.yaml
```
detiber: We deploy a version of cloud-init that contains a kubeadm module: https://github.com/detiber/cloud-init/blob/cluster-api/cloudinit/config/cc_kubeadm.py

pkg/cloud/aws/services/userdata/node.go
```diff
@@ -39,6 +39,30 @@ nodeRegistration:
 EOF
 
 kubeadm join --config /tmp/kubeadm-node.yaml
+`
+
+	nodeCloudInit = `{{.Header}}
+write_files:
+-   path: /tmp/kubeadm-node.yaml
+    owner: root:root
+    permissions: '0644'
```
detiber: Since this file contains secrets it might be better to set permissions to `0640` instead

pkg/cloud/aws/services/userdata/node.go
```diff
@@ -39,6 +39,30 @@ nodeRegistration:
...
+    content: |
+      ---
+      apiVersion: kubeadm.k8s.io/v1beta1
+      kind: JoinConfiguration
```
detiber: Just to give some insight, I ran into some issues the last time I attempted this with the combined InitConfiguration and ClusterConfiguration for bootstrapping the initial control plane instance. For some reason at the time kubeadm didn't decode the combined config properly and trying to nest a second yaml document using `write_files` was giving me issues as well.

pkg/cloud/aws/services/userdata/node.go
randomvariable: @detiber is the module deployed into the AMIs today?

detiber: It is, it's added as part of the image baking process.

sfzylad: Great, I like this solution much more than `runcmd`, will make a change. Thanks for the feedback.

sfzylad: Looks like this `kubeadm` module is not baked into the AMIs:

```bash
root@ip-10-0-0-105:~# ls -la /usr/lib/python3/dist-packages/cloudinit/config/cc_kube*
ls: cannot access '/usr/lib/python3/dist-packages/cloudinit/config/cc_kube*': No such file or directory
root@ip-10-0-0-105:~# ls -la /usr/lib/python3/dist-packages/cloudinit/config/cc_puppet*
-rw-r--r-- 1 root root 7910 Oct  2 20:52 /usr/lib/python3/dist-packages/cloudinit/config/cc_puppet.py
root@ip-10-0-0-105:~#
```

sfzylad: It looks like the `cloud-init` is not overwritten with the PPA package, hence the plugin `kubeadm` is not present. I will test CentOS and Amazon Linux tomorrow and create PR to fix this issue.

pkg/cloud/aws/services/userdata/userdata.go
sfzylad: I removed it.

pkg/cloud/aws/services/userdata/node.go
sfzylad: Good call, I changed it.

pkg/cloud/aws/services/userdata/node.go
sfzylad: I hit problems with `write_files` and certificates and keys. Solved it by introducing helper template function that converts certs and keys into base64 strings that `cloud-init` can recognise and handle correctly.

sfzylad:
This PR assumes #538 is merged, otherwise `ami` object in `machines.yaml` manifest has to be set to custom AMI.

pkg/cloud/aws/services/userdata/controlplane.go
```diff
@@ -16,111 +16,187 @@ limitations under the License.
...
-  kubeletExtraArgs:
-    cloud-provider: aws
-controlPlane:
-  localAPIEndpoint:
-    advertiseAddress: "${PRIVATE_IP}"
-    bindPort: 6443
-EOF
-
-kubeadm join --config /tmp/kubeadm-controlplane-join-config.yaml --v 10
+	controlPlaneJoinCloudInit = `{{.Header}}
```
detiber: Have you tested joining an additional control plane instance? I don't believe this will work as is, because the `--experimental-control-plane` is needed when invoking `kubeadm`.

My initial thoughts are that we can update the cloud-init module to add a separate operation for control plane join, which could then pass the required flag. It might need some special handling (for v1.14+, since the flag will no longer be experimental, though.

sfzylad: It seems to work. I tested it by adding additional plane node (controlplane-1) to machines.yaml file and running apply-machines phase. It joins fine.

detiber: I'll take it :man_shrugging:. I'll attempt to give this (and #538) a test this afternoon.

sfzylad: From the diff it looks like the bash version was just joining the plane without any special options.

ashish-amarnath: Yeah, the `--experimental-control-plane` is required when the join params are passed on the cli. We are using the `JoinConfiguration` to invoke `kubeadm join`.

pkg/cloud/aws/services/userdata/controlplane.go
```diff
@@ -16,111 +16,187 @@ limitations under the License.
...
+      nodeRegistration:
+        name: {{ "{{ ds.meta_data.hostname }}" }}
+        criSocket: /var/run/containerd/containerd.sock
+        kubeletExtraArgs:
+          cloud-provider: aws
+      controlPlane:
+        localAPIEndpoint:
+          advertiseAddress: {{ "{{ ds.meta_data.local_ipv4 }}" }}
+          bindPort: 6443
+kubeadm:
```
ashish-amarnath: can we also keep the `--v=10`?

detiber: To add support, we'll need to extend the cloud-init module here: https://github.com/detiber/cloud-init/tree/cluster-api

pkg/cloud/aws/services/userdata/controlplane.go
```diff
@@ -16,111 +16,187 @@ limitations under the License.
...
+-   path: /tmp/kubeadm.yaml
+    owner: root:root
+    permissions: '0640'
+    content: |
+      ---
+      apiVersion: kubeadm.k8s.io/v1beta1
+      kind: ClusterConfiguration
+      apiServer:
+        certSANs:
+          - {{ "{{ ds.meta_data.local_ipv4 }}" }}
```
ashish-amarnath: @sfzylad can you please explain what this syntax means?  :)

sfzylad: This value is builtin variable that `cloud-init` uses referring to instance's metadata. I referred to it based on this document: https://cloudinit.readthedocs.io/en/latest/topics/instancedata.html

pkg/cloud/aws/services/userdata/controlplane.go
sfzylad: Is it something that we want in scope of solving #205?

detiber: I don't think it is in scope for #205 @ashish-amarnath do you mind filing an issue for followup?

detiber:
/approve
/assign @chuckha 

Leaving lgtm to @chuckha

pkg/cloud/aws/services/userdata/controlplane.go
ashish-amarnath: Thank you for the explanation :)

pkg/cloud/aws/services/userdata/controlplane.go
ashish-amarnath: will do

chuckha:
/lgtm

thank you!

PR #516, opened 2019-01-24
Add IdleTimeout to ELB

vincepri:

Fixes an issue where `kubectl logs` would exit early with EOF due to 60s default timeout on ELBs.

Fixes #

**Release note**:

```release-note

```


randomvariable:
Looks like increasing this 60s won't completely resolve the problem, just make it less frequent. We should document this somewhere as an issue with using ELBs.

NLB would fix this, but introduces the hairpinning problem.

randomvariable:
Have created #519 to unblock this PR.

It's an lgtm from me, but CRD changes should probably get a 2nd opinion.

vincepri:
It won't fix the problem, but it'll fix 99% of the user experience issue 😄 which was the goal.

Issue #496, opened 2019-01-16
Bootstrap cluster should use STS Session tokens


vincepri:
/kind feature

**Describe the solution you'd like**
In the current behavior, when a user runs `clusterawsadm` the environment credentials are used to create the CloudFormation template and to generate an aws profile configuration file which is then used as a secret. 

For a bootstrap (short-lived) cluster, we should generate a session token using [sts api](https://docs.aws.amazon.com/sdk-for-go/api/service/sts/#GetSessionTokenInput). The session token is usually short lived and should be enough to guarantee successful operations until pivot.

After pivoting, the cluster should default to use IAM roles or static credentials, giving users a way to pick what's best for their use case.

randomvariable:
This has been discussed on Slack, but we purposely do not do STS calls as this will prevent clusterawsadm being used in environments where the user has either had to already assume a role to access the AWS account using cross-account IAM roles, or is working in an environment where access is granted using federated identity.

Either a `--no-session` parameter would need to be added to avoid this behaviour, or we do nothing and recommend users use a tool such as `aws-vault` to provide temporary sessions when they are using static AWS access keys.

chuckha:
@rosskukulinski

rosskukulinski:
@chuckha / @randomvariable I don't have quite enough context to weigh in just yet.  @randomvariable can you point me to the Slack discussions you reference?  IIRC, the Heptio-alum AWS accounts are setup where people must assume roles to access regions, right?

@vincepri - What are the failure conditions should the bootstrap cluster not come up before STS token expires?  What duration were you thinking about using?

randomvariable:
I'll need to find the Slack conversations, but this is the current state as I see it:

## Current State

### CloudFormation

`clusterawsadm alpha bootstrap create-stack` creates an AWS CloudFormation stack with the following resources:

#### IAM Managed Policies
* `nodes.cluster-api-provider-aws.sigs.k8s.io`

This is a misnomer, as it should be `nodes.cluster-provider-aws.k8s.io`, and is the per-node IAM policy as used by the AWS Cloud Provider for Kubernetes (issue to follow).

* `control-plane.cluster-api-provider-aws.sigs.k8s.io`

This is a misnomer, as it should be `control-plane.cluster-provider-aws.k8s.io`, and is the per-node IAM policy as used by the master nodes for the AWS Cloud Provider for Kubernetes (issue to follow).

[...]

Issue #297, opened 2018-10-23
Exclude aws credentials when pivoting cluster


detiber:
/kind feature
/priority important-soon

Depends on: https://github.com/kubernetes-sigs/cluster-api/issues/556

Currently w hen clusterctl pivots the cluster-api components, the aws credentials secret which is used for the bootstrapping cluster is also pivoted to the target cluster. Since the target cluster requires an existing IAM profile that provides all of the required permissions for the cluster and machine actuators, we should default to using the instance profile for credentials instead of the user provided credentials (which may be time limited).

roberthbailey:
As I mentioned in the meeting today, we have three categories of "things" that clusterctl should know about:

1. Things that should exist only in the bootstrap cluster
1. Things that should exist in both the bootstrap and target clusters
1. Things that should exist only in the target cluster

Right now clusterctl assumes that everything falls into the second bucket, except for the things specified via the addon flag, which I added so that we could put things into the third bucket. This feature request is about separating things that should be in the first bucket, which was generally agreed was a good idea.

The open question is about how to specify the differences. We can do it via flags, component config entries, different files, or annotations. Or maybe something else I haven't thought of. 

Ideas welcome!

Issue #201, opened 2018-10-09
Better AWS credentials management post-pivot


randomvariable:
/kind feature

**Describe the solution you'd like**

As part of the bootstrap, users need to provide AWS credentials
in the provider components manifest so that the controllers
can spin up AWS infrastructure from minikube.

It then pivots the manifests to the new cluster, however,
`clusterawsadm` creates an IAM role and instance profile
for this purpose, so some way of removing the env vars
for the controllers post install is needed. Could be done as:

* End-user documentation on editing the deployment
or
* Get the deployer to patch the provider components 
or
* Make it a clusterawsadm command

We should also support session tokens during bootstrap, e.g.
someone is assuming role to deploy the cluster, or is otherwise
using a temporary session.

detiber:
Upstream cluster-api issue: https://github.com/kubernetes-sigs/cluster-api/issues/556

randomvariable:
This is more or less resolved by moving away from the idea of pivoting by default, which has been done in the quick start.

/close

PR #518, opened 2019-01-24
Use default credentials chain instead of environment variables

### LINKED ISSUE #494: UX: Provider Components manifest generation should use AWS credential provider chain
/kind bug

Part of #276 

The process by which to generate the provider components manifest for clusterctl, currently `make manifests` should use the default AWS credential provider.

Credentials from ~/aws/config to be used for bootstrapping the cluster

`clusterawsadm` is currently responsible for this part of the manifest generation

**Environment:**

sfzylad:

Fixes #494 

**Release note**:

```release-note

```

### DIFF
--- a/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
+++ b/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
@@ -23,6 +23,8 @@ import (
 	"os"
 	"text/template"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/defaults"
 	"github.com/aws/aws-sdk-go/aws/session"
 	cfn "github.com/aws/aws-sdk-go/service/cloudformation"
 	awssts "github.com/aws/aws-sdk-go/service/sts"
@@ -72,7 +74,8 @@ func RootCmd() *cobra.Command {
 	newCmd.AddCommand(createStackCmd())
 	newCmd.AddCommand(generateIAMPolicyDocJSON())
 	newCmd.AddCommand(encodeAWSSecret())
-	newCmd.AddCommand(generateAWSDefaultProfile())
+	newCmd.AddCommand(generateAWSDefaultProfileWithChain())
+
 	return newCmd
 }
 
@@ -216,7 +219,7 @@ func encodeAWSSecret() *cobra.Command {
 		Short: "Encode AWS credentials as a base64 encoded Kubernetes secret",
 		Long:  "Encode AWS credentials as a base64 encoded Kubernetes secret",
 		RunE: func(cmd *cobra.Command, args []string) error {
-			creds, err := getCredentialsFromEnvironment()
+			creds, err := getCredentialsFromDefaultChain()
 
 			if err != nil {
 				return err
@@ -235,21 +238,19 @@ func encodeAWSSecret() *cobra.Command {
 	return newCmd
 }
 
-func generateAWSDefaultProfile() *cobra.Command {
+func generateAWSDefaultProfileWithChain() *cobra.Command {
 	newCmd := &cobra.Command{
 		Use:   "generate-aws-default-profile",
 		Short: "Generate an AWS profile from the current environment",
 		Long:  "Generate an AWS profile from the current environment for the ephemeral bootstrap cluster",
 		RunE: func(cmd *cobra.Command, args []string) error {
 
-			creds, err := getCredentialsFromEnvironment()
-
+			creds, err := getCredentialsFromDefaultChain()
 			if err != nil {
 				return err
 			}
 
 			profile, err := renderAWSDefaultProfile(*creds)
-
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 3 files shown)


randomvariable:
Need to run `make gazelle` on this.

cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
```diff
@@ -263,33 +264,25 @@ func generateAWSDefaultProfile() *cobra.Command {
...
-
-	region, err := getEnv("AWS_REGION")
-	if err != nil {
-		return nil, err
-	}
-	creds.Region = region
-
-	accessKeyID, err := getEnv("AWS_ACCESS_KEY_ID")
+	conf := aws.NewConfig()
+	chain := defaults.CredChain(conf, defaults.Handlers())
```
vincepri: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/05c39e93d8b035096e917df1e884b1c771f80969/vendor/github.com/aws/aws-sdk-go/aws/defaults/defaults.go#L88-L92 This comment says that this method shouldn't be used directly. 

Before merging this in, we should validate a few scenarios:
1. Does this work on an AWS instance and does it use an IAM role if available?
2. Is there a way to use [STS tokens](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/496) later on?
3. If a user has a single profile/credentials, does this effectively copy the local secrets?
4. Is it still possible to specify environment variable that override the local secret?

If all the above can be validated, the approach you proposed should probably be generalized also in the actuators.

randomvariable: No, we don't need to do anything in the actuators.

Because we don't call explicit credentials, the SDK defaults to using the default credential chain anyway.

randomvariable: Using the method directly is reasonable here, because the way we want to get the credentials into a secret is a bit hacky, and we just want to pretend that the injected secret came from the same source it would normally do if it wasn't in a pod/container.

vincepri: Right, that's the current behavior. We'd still need to revisit it or find other solutions for [credential management during pivot](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/297) and [after pivot](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/201). If there is a way to specify temporary bootstrap-only credentials, the approach used above can be used there too 😄.

randomvariable: Let's not confuse these things, because we do not know the true final environments, and we should base this on how SDKs interact with IAM in practice, as well as upcoming SDK changes around OIDC.

randomvariable: the tl;dr after chatting about this:

Need a doc/KEP on:
* the dev and management cluster use cases
* how clusterawsadm, beyond the cloudformation bootstrap, is for the dev / solution architect bootstrapping only from their computer
* the IAM credential provider chain, and how we expect it to be used in practice (e.g. instance metadata via kube2iam, OIDC etc...)

vincepri:
Given that `clusterawsadm` is meant only to be used in a development environment workflow, I think this change is ok to go.

@sfzylad Can we add a log line to `clusterawsadm` which informs users that the tool isn't meant to be used in production environments and (if possible) which credential chain is being used?

sfzylad:
@vincepri I'm thinking where would be the best place to add the log information. At the moment `cmd/clusterctl/examples/aws/out/credentials` make target simply runs:

```bash
clusterawsadm alpha bootstrap generate-aws-default-profile > cmd/clusterctl/examples/aws/out/credentials
```

As you can see, make redirects the whole output from `clusterawsadm` to the credentials file. This works fine as long as the output of the command contains the valid content of credentials file. If we add a message to the output, the credentials file will become invalid. In such case I see the following options here:

- to print the message on `STDERR,` then `clusterawsadm alpha bootstrap generate-aws-default-profile 1> cmd/clusterctl/examples/aws/out/credentials` in `Makefile` will capture the valid file content, but then we will need to do some hacks in `Makefile` to capture and print this as make doesn't print what it gets on `STDERR` as long as it gets proper exit code.
- to print the message as part of `cmd/clusterctl/examples/aws/out/credentials` make target without touching any Go code. This will only work if the command is executed via the make command.
- to print the message on `STDERR` from the bootstrap code and as part of the make target, this seems a bit redundant to me.
- to consider documenting this option to be intended not to be used in production environments.

As for the credentials chain, I see no easy and non-hacky way to access this information.

Please tell me your thoughts.

randomvariable:
The documentation definitely needs updating. It currently says we take creds only from the environment, but this isn't true anymore.

The makefile workflow is going away, and won't be consumed by end users, to be replaced by Kustomize / kubectl 1.14, but may still have the same problem.

Think it's reasonable to print to stderr in `clusterawsadm` as that is intended as a directly consumable artifact in a way the Makefile isn't.

sfzylad:
Should be fine now.

randomvariable:
/lgtm

Documented security model should be blocking for #519

PR #538, opened 2019-01-31
Make sure ubuntu gets proper version of cloud-init

### LINKED ISSUE #205: Switch machine actuator to use cloud-config rather than startup script

sfzylad:

Related to #205 

This PR does not solve #205. It fixes the problem that was discovered while working on #205.

**Release note**:

```release-note

```

### DIFF
--- a/build/amis/packer/packer.json
+++ b/build/amis/packer/packer.json
@@ -98,6 +98,27 @@
         "--extra-vars",
         "kubernetes_version={{user `kubernetes_version`}} kubernetes_cni_version={{user `kubernetes_cni_version`}}"
       ]
+    },
+    {
+      "type": "goss",
+      "vars_file": "goss/redhat-vars.yaml",
+      "tests": [
+        "goss/goss.yaml"
+      ],
+      "only": [
+        "amazon-2",
+        "centos-7"
+      ]
+    },
+    {
+      "type": "goss",
+      "vars_file": "goss/ubuntu-vars.yaml",
+      "tests": [
+        "goss/goss.yaml"
+      ],
+      "only": [
+        "ubuntu-1804"
+      ]
     }
   ]
 }
--- a/docs/packer.md
+++ b/docs/packer.md
@@ -9,6 +9,7 @@ We use it to bake our AMIs.
 
 * `packer` binary
 * [packer-goss](https://github.com/YaleUniversity/packer-provisioner-goss) plugin
+* ansible
 
 ## Plugin instalation
 
--- a/docs/packer.md
+++ b/docs/packer.md
@@ -8,7 +8,7 @@ We use it to bake our AMIs.
 ## Prerequisites
 
 * `packer` binary
-* `packer-goss` plugin
+* [packer-goss](https://github.com/YaleUniversity/packer-provisioner-goss) plugin
 
 ## Plugin instalation
 
--- a/docs/packer.md
+++ b/docs/packer.md
@@ -1,0 +1,32 @@
+# Using Packer and baking AMIs
+
+## Overview
+
+[Packer](http://packer.io/) is well known tool for baking images of any kind.
+We use it to bake our AMIs.
+
+## Prerequisites
+
+* `packer` binary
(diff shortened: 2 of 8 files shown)


detiber:
/ok-to-test
/lgtm

@chuckha which account have you been using for publishing AMIs?

chuckha:
I've been using the wardroom account since we don't have one dedicated for CAPA yet

sfzylad:
This PR requires `packer-goss` packer plugin to be present. Otherwise packer will fail. On the other hand, if the plugin is present, each AMI wil be verified if the proper version of `ckoud-init` is present. This should make our AMIs more reliable in the future. New tests can be added in the future too.

docs/packer.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Overview
+
+[Packer](http://packer.io/) is well known tool for baking images of any kind.
+We use it to bake our AMIs.
+
+## Prerequisites
+
+* `packer` binary
+* `packer-goss` plugin
```
chuckha: Can you add a link to packer-goss?

randomvariable:
Yes, it's part of but not the completion of #205. Can @sfzylad or someone with powers amend the description?

sfzylad:
> Yes, it's part of but not the completion of #205. Can @sfzylad or someone with powers amend the description?

Updated the description.

docs/packer.md
sfzylad: Added, thank's for suggestion. :)

chuckha:
I updated again to remove the wording "fixes abc" because GitHub isn't very smart about this

randomvariable:
This is an lgtm from me.

build/amis/packer/packer.json
```diff
@@ -98,6 +98,27 @@
         "--extra-vars",
         "kubernetes_version={{user `kubernetes_version`}} kubernetes_cni_version={{user `kubernetes_cni_version`}}"
       ]
+    },
+    {
+      "type": "goss",
+      "vars_file": "goss/redhat-vars.yaml",
```
chuckha: Do these get cleaned up after goss runs?

docs/packer.md
```diff
@@ -0,0 +1,32 @@
...
+* `packer` binary
+* [packer-goss](https://github.com/YaleUniversity/packer-provisioner-goss) plugin
+
+## Plugin instalation
+
+To install `packer-goss` plugin the following should be executed inside of the
+`build/amis/packer` directory:
+
+```bash
+$ curl -o packer-goss  https://github.com/YaleUniversity/packer-provisioner-goss/releases/download/v0.3.0/packer-provisioner-goss-v0.3.0-linux-amd64
```
detiber: This command as written downloads an html file with a redirect notice rather than the binary itself. Adding the `-L` flag will allow curl to follow the redirect and download the binary. The resulting binary name should also be `packer-provisioner-goss`

```suggestion
$ curl -Lo packer-provisioner-goss  https://github.com/YaleUniversity/packer-provisioner-goss/releases/download/v0.3.0/packer-provisioner-goss-v0.3.0-linux-amd64
```

docs/packer.md
```diff
@@ -0,0 +1,32 @@
...
+
+## Overview
+
+[Packer](http://packer.io/) is well known tool for baking images of any kind.
+We use it to bake our AMIs.
+
+## Prerequisites
+
+* `packer` binary
+* [packer-goss](https://github.com/YaleUniversity/packer-provisioner-goss) plugin
```
detiber: Not directly related to this PR, but `ansible` should also be listed as a prerequisite

sfzylad: Good catch. Added ansible on the list too.

build/amis/packer/packer.json
sfzylad: By default the plugin puts everything into the `/tmp` directory. This is cleaned at boot time.

chuckha:
/lgtm
/assign @detiber 

thanks for this contribution!

PR #541, opened 2019-02-01
Reconfigure cloud-init to use ec2-user instead of centos on Amazon Linux

### LINKED ISSUE #536: ec2-user not used on Amazon Linux AMI
/kind bug

Once our Amazon Linux 2 AMI is used, the default user to be used with Amazon Linux is `centos`. The default user supported by AWS is `ec2-user` and is present in the system we provide. It's that `cloud-init` is comnfigured to use `centos` user:

```bash
# cat /etc/cloud/cloud.cfg
[...]
# System and/or distro specific settings
# (not accessible to handlers/transforms)
system_info:
   # This will affect which distro class gets used
   distro: centos
   # Default user name + that default users groups (if added/used)
   default_user:
     name: centos # <--- here's where the problem is
     lock_passwd: True
     gecos: centos Cloud User # <--- this should be changed too
     groups: [wheel, adm, systemd-journal]
     sudo: ["ALL=(ALL) NOPASSWD:ALL"]
     shell: /bin/bash
   # Other config here will be given to the distro class and/or path classes
   paths:
      cloud_dir: /var/lib/cloud/
      templates_dir: /etc/cloud/templates/
   ssh_svcname: sshd
```

When instance uses Amazon Linux, `ec2-user` system user should be used to loggin the instance.

**Environment:**

sfzylad:

Fixes #536 

**Release note**:

```release-note

```

### DIFF
--- a/build/amis/ansible/roles/common/files/cloud.cfg
+++ b/build/amis/ansible/roles/common/files/cloud.cfg
@@ -1,0 +1,93 @@
+# The top level settings are used as module
+# and system configuration.
+
+# A set of users which may be applied and/or used by various modules
+# when a 'default' entry is found it will reference the 'default_user'
+# from the distro configuration specified below
+users:
+   - default
+
+# If this is set, 'root' will not be able to ssh in and they
+# will get a message to login instead as the default $user
+disable_root: true
+
+mount_default_fields: [~, ~, 'auto', 'defaults,nofail', '0', '2']
+resize_rootfs_tmp: /dev
+ssh_pwauth:   0
+
+# This will cause the set+update hostname module to not operate (if true)
+preserve_hostname: false
+
+# Example datasource config
+# datasource:
+#    Ec2:
+#      metadata_urls: [ 'blah.com' ]
+#      timeout: 5 # (defaults to 50 seconds)
+#      max_wait: 10 # (defaults to 120 seconds)
+
+# The modules that run in the 'init' stage
+cloud_init_modules:
+ - migrator
+ - seed_random
+ - bootcmd
+ - write-files
+ - growpart
+ - resizefs
+ - disk_setup
+ - mounts
+ - set_hostname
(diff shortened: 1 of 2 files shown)


sfzylad:
With this changes I was able to bake an AMI and login as `ec2-user`. User `centos` is no longer present.

sfzylad:
The file belongs to the RPM and is marked as a config file:

```bash
[root@ip-10-0-1-54 ~]# rpm -qf /etc/cloud/cloud.cfg
cloud-init-18.3+52.gc5f78957-1.el7.noarch
[root@ip-10-0-1-54 ~]# rpm -qc cloud-init-18.3+52.gc5f78957-1.el7.noarch |grep cloud.cfg
/etc/cloud/cloud.cfg
```

detiber:
/approve

@randomvariable or @chuckha want to give final lgtm?

build/amis/ansible/roles/common/files/cloud.cfg
```diff
@@ -0,0 +1,93 @@
...
+
+# Example datasource config
+# datasource:
+#    Ec2:
+#      metadata_urls: [ 'blah.com' ]
+#      timeout: 5 # (defaults to 50 seconds)
+#      max_wait: 10 # (defaults to 120 seconds)
+
+# The modules that run in the 'init' stage
+cloud_init_modules:
```
chuckha: Do we need all of these modules (and config_modules and final_modules)? Like salt-minion.

detiber: In this case I think this is fine, since this file is directly from the cloud-init package and tweaked for the correct default user.

chuckha:
/lgtm

In summary, this takes the standard config file out of the package that we were already installing and customizes it slightly so that it works on amazon-linux.

The file is marked as a config file so upgrades will not overwrite the customization.

PR #556, opened 2019-02-08
Update AMIs to latest builds

detiber:

Updates AMIs to the latest builds

**Release note**:
```release-note
NONE.
```


chuckha:
already published. bug discovered,

/lgtm

chuckha:
but not with these images, with the ami finding code! there is no guaranteed ordering.

PR #560, opened 2019-02-08
More reliable public subnet detection

sethp-nr:

The subnet tracking code did not recognize our public subnets, because we've defaulted IP assignment off everywhere. Reading the docs seems to indicate that "public" or "private" subnets in AWS are an emergent property from how the traffic flows:

https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Scenario2.html

> The instances in the public subnet can send outbound traffic directly to the Internet, whereas the instances in the private subnet can't. Instead, the instances in the private subnet can access the Internet by using a network address translation (NAT) gateway that resides in the public subnet.

Which concurs with Stack Overflow:

https://stackoverflow.com/a/38691109

> A subnet is deemed to be a Public Subnet if it has a Route Table that directs traffic to the Internet Gateway.

So this PR modifies the existing logic to instead look for an internet-facing route in the subnet's route to determine whether this subnet is public or private.

Fixes #

The tests are still failing, but I wanted to get the PR up early for any feedback you have, especially around workflow. Currently, I'm running `make docker-push-dev MANAGER_IMAGE_TAG=0.0.3-2` and re-launching the pod in my local kind cluster. That's working ok, but it's a little slow and seems like it might be missing some things (is the CRD yaml generated?). Any tips would be appreciated!

Also, on an unrelated note, I was hoping to pull in the latest `cluster-api` upstream to make use of the `get-kubeconfig` phase: would it make more sense to add that to this PR, or open a second one?

**Release note**:

```release-note
find public subnets by inspecting routes
```

### DIFF
--- a/pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
+++ b/pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
@@ -122,18 +122,18 @@ type SubnetSpec struct {
 	// AvailabilityZone defines the availability zone to use for this subnet in the cluster's region.
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 
-	// IsPublic defines the subnet as a public subnet. Refer to the AWS documentation for further information.
+	// IsPublic defines the subnet as a public subnet. A subnet is considered public when it has a direct route to and from the internet via an internet gateway.
 	// +optional
-	IsPublic bool `json:"isPublic,omitempty"`
+	IsPublic bool `json:"isPublic"`
 
 	// RouteTableID is the routing table id associated with the subnet.
 	// +optional
 	RouteTableID *string `json:"routeTableId"`
 
 	// NatGatewayID is the NAT gateway id associated with the subnet.
-	// Ignored if the subnet is public.
+	// Ignored unless the subnet is managed by the provider, in which case this is set on the public subnet where the nat gateway resides. It is then used to determe routes for private subnets in the same AZ as the public subnet.
 	// +optional
-	NatGatewayID *string `json:"natGatewayId"`
+	NatGatewayID *string `json:"natGatewayId,omitempty"`
 
 	// Tags is a collection of tags describing the resource.
 	Tags tags.Map `json:"tags,omitempty"`
--- a/pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
+++ b/pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
@@ -122,7 +122,7 @@ type SubnetSpec struct {
 	// AvailabilityZone defines the availability zone to use for this subnet in the cluster's region.
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 
-	// IsPublic defines the subnet as a public subnet. A subnet is considered public when it has a direct route to and from the internet via an internet gateway.
+	// IsPublic defines the subnet as a public subnet. A subnet is public when it is associated with a route table that has a route to an internet gateway.
 	// +optional
 	IsPublic bool `json:"isPublic"`
 
--- a/pkg/cloud/aws/services/ec2/subnets.go
+++ b/pkg/cloud/aws/services/ec2/subnets.go
@@ -36,6 +36,10 @@ const (
 	defaultPublicSubnetCidr  = "10.0.1.0/24"
 )
 
+const (
+	NameSubnetRoutable = tags.NameAWSProviderPrefix + "routable"
+)
+
 func (s *Service) reconcileSubnets() error {
 	klog.V(2).Infof("Reconciling subnets")
 
@@ -184,36 +188,50 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
 	if err != nil {
 		return nil, err
 	}
+
 	internetGateways, err := s.describeVpcInternetGateways()
 	if err != nil {
 		return nil, err
 	}
 
 	subnets := make([]*v1alpha1.SubnetSpec, 0, len(out.Subnets))
+	// Besides what the AWS API tells us directly about the subnets, we also want to discover whether the subnet is "public" (i.e. directly connected to the internet) or if there are any associated NAT gateways.
+	// We also look for a tag indicating that a particular subnet should be public, to try and determine whether a managed VPC's subnet should have such a route, but does not.
 	for _, ec2sn := range out.Subnets {
+		var hasPublicTag bool
+		for _, tag := range ec2sn.Tags {
+			if tag == nil {
+				continue
+			}
+
+			if tag.Key != nil && *tag.Key == NameSubnetRoutable && tag.Value != nil && *tag.Value == "public" {
+				hasPublicTag = true
+			}
+		}
+
 		rt := routeTables[*ec2sn.SubnetId]
+
 		var hasPublicRoute bool
-		var routeTableId, natGatewayId *string
+		var routeTableId *string
 		if rt != nil {
 			routeTableId = rt.RouteTableId
 			for _, route := range rt.Routes {
-				if route == nil {
+				if route == nil || route.GatewayId == nil {
 					continue
 				}
-				if route.GatewayId != nil {
-					for _, igw := range internetGateways {
-						if igw == nil {
-							continue
-						}
-						if *route.GatewayId == *igw.InternetGatewayId {
-							hasPublicRoute = true
-						}
+				for _, igw := range internetGateways {
+					if igw == nil {
+						continue
+					}
+					if *route.GatewayId == *igw.InternetGatewayId {
+						hasPublicRoute = true
 					}
 				}
 			}
 		}
 
 		ngw := natGateways[*ec2sn.SubnetId]
+		var natGatewayId *string
 		if ngw != nil {
 			natGatewayId = ngw.NatGatewayId
 		}
@@ -221,7 +239,7 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
 			ID:               *ec2sn.SubnetId,
 			CidrBlock:        *ec2sn.CidrBlock,
 			AvailabilityZone: *ec2sn.AvailabilityZone,
-			IsPublic:         hasPublicRoute,
+			IsPublic:         hasPublicRoute || hasPublicTag,
 			Tags:             converters.TagsToMap(ec2sn.Tags),
 			RouteTableID:     routeTableId,
 			NatGatewayID:     natGatewayId,
@@ -297,15 +315,17 @@ func (s *Service) deleteSubnet(id string) error {
 }
 
 func (s *Service) getSubnetTagParams(id string, public bool) tags.BuildParams {
-	var name strings.Builder
+	var routable string
+	if public {
+		routable = "public"
+	} else {
+		routable = "private"
+	}
 
+	var name strings.Builder
 	name.WriteString(s.scope.Name())
 	name.WriteString("-subnet-")
-	if public {
-		name.WriteString("public")
-	} else {
-		name.WriteString("private")
-	}
+	name.WriteString(routable)
 
 	return tags.BuildParams{
 		ClusterName: s.scope.Name(),
@@ -313,5 +333,8 @@ func (s *Service) getSubnetTagParams(id string, public bool) tags.BuildParams {
 		Lifecycle:   tags.ResourceLifecycleOwned,
 		Name:        aws.String(name.String()),
 		Role:        aws.String(tags.ValueCommonRole),
+		Additional: tags.Map{
+			NameSubnetRoutable: routable,
+		},
 	}
 }
--- a/pkg/cloud/aws/services/ec2/subnets.go
+++ b/pkg/cloud/aws/services/ec2/subnets.go
@@ -36,10 +36,6 @@ const (
 	defaultPublicSubnetCidr  = "10.0.1.0/24"
 )
 
-const (
-	NameSubnetRoutable = tags.NameAWSProviderPrefix + "routable"
-)
-
 func (s *Service) reconcileSubnets() error {
 	klog.V(2).Infof("Reconciling subnets")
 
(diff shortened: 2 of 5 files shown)


detiber:
> Also, on an unrelated note, I was hoping to pull in the latest `cluster-api` upstream to make use of the `get-kubeconfig` phase: would it make more sense to add that to this PR, or open a second one?

I would suggest submitting that as a separate PR

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +175,56 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+	if err != nil {
+		return nil, err
+	}
+
 	subnets := make([]*v1alpha1.SubnetSpec, 0, len(out.Subnets))
 	for _, ec2sn := range out.Subnets {
+		rt := routeTables[*ec2sn.SubnetId]
+		var hasPublicRoute bool
+		var routeTableId, natGatewayId *string
+		if rt != nil {
```
vincepri: Complexity can be reduced a little:

```
rt, ok := routeTables[*ec2sn.SubnetId]
if !ok {
    continue
}
```

which removes the need for this`if` block.

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +175,56 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+			routeTableId = rt.RouteTableId
+			for _, route := range rt.Routes {
+				if route == nil {
```
vincepri: This could be `if route == nil || route.GatewayId == nil` and remove the if condition below.

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +175,56 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+	if err != nil {
+		return nil, err
+	}
+	internetGateways, err := s.describeVpcInternetGateways()
+	if err != nil {
+		return nil, err
+	}
+
 	subnets := make([]*v1alpha1.SubnetSpec, 0, len(out.Subnets))
 	for _, ec2sn := range out.Subnets {
```
vincepri: We should add a comment describing the goal of the for loop and result.

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +175,56 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+				if route == nil {
+					continue
+				}
+				if route.GatewayId != nil {
+					for _, igw := range internetGateways {
+						if igw == nil {
+							continue
+						}
+						if *route.GatewayId == *igw.InternetGatewayId {
+							hasPublicRoute = true
```
vincepri: What's the reasoning behind checking if the route's internet gateway is in the `internetGateways` slice? Can we assume that IGW associated with a subnet's routing table inside a VPC can only point to a gateway attached to the VPC?

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +175,56 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+		}
 		subnets = append(subnets, &v1alpha1.SubnetSpec{
 			ID:               *ec2sn.SubnetId,
 			CidrBlock:        *ec2sn.CidrBlock,
 			AvailabilityZone: *ec2sn.AvailabilityZone,
-			IsPublic:         *ec2sn.MapPublicIpOnLaunch,
+			IsPublic:         hasPublicRoute,
 			Tags:             converters.TagsToMap(ec2sn.Tags),
+			RouteTableID:     routeTableId,
+			NatGatewayID:     natGatewayId,
```
vincepri: Could this just be assigned to `natGateways[*ec2sn.SubnetId]`?

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
```diff
@@ -122,18 +122,18 @@ type SubnetSpec struct {
 	// AvailabilityZone defines the availability zone to use for this subnet in the cluster's region.
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 
-	// IsPublic defines the subnet as a public subnet. Refer to the AWS documentation for further information.
+	// IsPublic defines the subnet as a public subnet. A subnet is considered public when it has a direct route to and from the internet via an internet gateway.
```
vincepri: This should be reworded to just `direct route to the internet via an internet gateway`. I don't think there is a way to check the route _from_ the igw to the VPC.

pkg/cloud/aws/services/ec2/subnets.go
vincepri: Also, could a route in the slice actually be `nil`?

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +175,56 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+	internetGateways, err := s.describeVpcInternetGateways()
+	if err != nil {
+		return nil, err
+	}
+
 	subnets := make([]*v1alpha1.SubnetSpec, 0, len(out.Subnets))
 	for _, ec2sn := range out.Subnets {
+		rt := routeTables[*ec2sn.SubnetId]
+		var hasPublicRoute bool
+		var routeTableId, natGatewayId *string
```
vincepri: Seems these declarations could go away?

sethp-nr:
Thanks for the feedback! I'm working on the unit tests now. It's a bit awkward because suddenly the subnet logic depends on route tables, nat & internet gateways, but nothing that a few dozen lines of mocks can't fix 😄

I'm also finding some rough edges trying to keep the fully-managed path working concurrently with the provided VPC path: in one case the public-ness of the subnet is discovered (the internet route is in place first), and in the other it's created (the subnet is flagged public first). I'm expecting to "solve" this with a tag on the managed subnet, but other ideas welcome: moving the canonical Spec state out to the AWS tag state feels less than ideal.

pkg/cloud/aws/services/ec2/subnets.go
sethp-nr: It's a pointer, so maybe? The data there is more or less straight from the AWS APIs, so if we got a response that looked something like:

```
...
{
  "Routes": [
    {
      "DestinationCidrBlock": "0.0.0.0/0",
      "GatewayId": "igw-zzzz",
      "Origin": "CreateRoute",
      "State": "active"
    },
    null
  ]
}
...
```

Then I would assume that would show up as a nil pointer. Whether AWS is capable of responding that way, I don't know.

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
sethp-nr: I intended to differentiate an "igw" route from an "Egress-only Internet Gateway," with my assumption being that the purpose of the former is to somehow teach the AWS SDN a route from the internet into the subnet.

Maybe `when it has a route associated with a bidirectional internet gateway`?

pkg/cloud/aws/services/ec2/subnets.go
sethp-nr: We need the `NatGatewayId` field inside that structure, which I expect to be nil (absent from the map) for private subnets.

pkg/cloud/aws/services/ec2/subnets.go
sethp-nr: From what I could tell, determining the "type" of a route's destination is a string parsing/matching problem: "does this route point to an internet gateway?" can be translated into either:

- "does it match an id that's known to be an internet gateway?"
- "does it start with `igw`?"

I went the former route because I prefer when possible to treat IDs as opaque, but I guess Hyrum's law here would suggest that amazon isn't going to change their structure any time soon. Do you have a preference on which way to go?

pkg/cloud/aws/services/ec2/subnets.go
sethp-nr: I ended up reverting this change to simplify the tests (the `continue` skipped over the `append` below), and because it felt a little more natural with the tag stuff to have each block as a "fact" that we were trying to discover.

pkg/cloud/aws/services/ec2/subnets.go
vincepri: I think the go-aws-sdk and aws APIs guarantee to not return nil values in slices, so it should be ok to remove the check.

pkg/cloud/aws/services/ec2/subnets.go
vincepri: String matching on `igw` sounds good to me, for now only internet gateways are supported anyways :)

pkg/cloud/aws/services/ec2/subnets.go
vincepri: I believe you can just do `NatGatewayID:     natGateways[*ec2sn.SubnetId]` without the intermediate step.

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
vincepri: Is there any way to actually check that though?

vincepri:
@sethp-nr Shouldn't the managed scenario work with the normal check and without tags?

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
sethp-nr: I believe the purpose of an internet gateway is to be bidirectional, but maybe the thing to do is take the phrasing straight from the [AWS docs](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Internet_Gateway.html):

> If your subnet is associated with a route table that has a route to an internet gateway, it's known as a public subnet.

sethp-nr:
I don't think so. The scenario I was imagining was:

1. We `reconcileSubnets` with no managed subnets, which leads to speccing them out and creating them
2. The `reconcileRouteTables` fails, perhaps because of an exceeded rate limit or what have you
3. The next call into `ReconcileNetwork` calls `reconcileSubnets` which sees no internet routes, and so updates the spec with `IsPublic: false` on the old "public" subnet, so `reconcileRouteTables` won't ever assign an igw route (and I think we'll re-spec and re-create a new "public" subnet)

pkg/cloud/aws/services/ec2/subnets.go
sethp-nr: I'm not sure I follow. When I change that line as you suggest, the compiler reports:

```
# sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/ec2
pkg/cloud/aws/services/ec2/subnets.go:234:4: cannot use natGateways[*ec2sn.SubnetId] (type *ec2.NatGateway) as type *string in field value
```

vincepri: Nevermind, I misread the line.

vincepri:
That seems like a valid edge case that I can see it happening. I was thinking of using `tags.NameAWSClusterAPIRole` instead of a new tag and have `public` or `private` as new Values in `tags/types.go`. The managed subnets would be automatically include the tags and users can also setup tags matching the same naming. 

Moreover `tags.Map` can be extended with a new method called `GetRole() string` which would return the role if any.

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +179,59 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+		var hasPublicRoute bool
+		var routeTableId *string
+		rt := routeTables[*ec2sn.SubnetId]
+		if rt != nil {
+			routeTableId = rt.RouteTableId
+			for _, route := range rt.Routes {
+				if route.GatewayId == nil {
+					continue
+				}
+				if strings.HasPrefix(*route.GatewayId, "igw") {
```
vincepri: We can omit the `if` above using `if route.GatewayId != nil &&  strings.HasPrefix(*route.GatewayId, "igw") {`.

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -175,14 +179,59 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
...
+				}
+			}
+		}
+
+		ngw := natGateways[*ec2sn.SubnetId]
+		var natGatewayId *string
+		if ngw != nil {
+			natGatewayId = ngw.NatGatewayId
+		}
 		subnets = append(subnets, &v1alpha1.SubnetSpec{
```
vincepri: If we move the definition of SubnetSpec on line 196, when the for loop starts, we'll be able to assign directly to the struct's fields and remove the temporary vars.

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -255,21 +304,26 @@ func (s *Service) deleteSubnet(id string) error {
...
+	name.WriteString(s.scope.Name())
+	name.WriteString("-subnet-")
+	name.WriteString(routable)
+
 	return tags.BuildParams{
 		ClusterName: s.scope.Name(),
 		ResourceID:  id,
 		Lifecycle:   tags.ResourceLifecycleOwned,
 		Name:        aws.String(name.String()),
 		Role:        aws.String(tags.ValueCommonRole),
```
vincepri: As suggested, this Role could be private/public instead of "common", which I don't think it has any particular meaning referring to managed subnets.

/cc @detiber

sethp-nr:
Thanks @vincepri, I think I captured that feedback. I want to let you know I'll be offline until 2/20 (next Wednesday) – feel free to do what you will with this PR, or leave it until then.

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
```diff
@@ -122,18 +122,18 @@ type SubnetSpec struct {
...
-	IsPublic bool `json:"isPublic,omitempty"`
+	IsPublic bool `json:"isPublic"`
 
 	// RouteTableID is the routing table id associated with the subnet.
 	// +optional
 	RouteTableID *string `json:"routeTableId"`
 
 	// NatGatewayID is the NAT gateway id associated with the subnet.
-	// Ignored if the subnet is public.
+	// Ignored unless the subnet is managed by the provider, in which case this is set on the public subnet where the nat gateway resides. It is then used to determe routes for private subnets in the same AZ as the public subnet.
```
chuckha:
```suggestion
	// Ignored unless the subnet is managed by the provider, in which case this is set on the public subnet where the NAT gateway resides. It is then used to determine routes for private subnets in the same AZ as the public subnet.
```

pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -255,21 +288,23 @@ func (s *Service) deleteSubnet(id string) error {
...
+	var role string
 	if public {
-		name.WriteString("public")
+		role = tags.ValuePublicRole
 	} else {
-		name.WriteString("private")
+		role = tags.ValuePrivateRole
 	}
 
+	var name strings.Builder
```
chuckha: `fmt.Sprtintf` would be easier to read. I believe string builder was used to avoid a var declaration, but it ended up using a var declaration anyway. 

The more go-like pattern I'd expect here is

```go
role := tags.ValuePrivateRole
if public {
    role = tags.ValuePublicRole
}
name := fmt.Sprintf("%s-subnet-%s", s.scope.Name(), role)
```

This is pretty minor though.

chuckha:
/lgtm

one typo can be fixed later

PR #563, opened 2019-02-08
Sorts AMIs when they are returned

### LINKED ISSUE #558: Describe Images does not guarantee ordering
/kind bug

us-west-2 and us-east-1 have different results when 

the Amis would be ordered by timestamp

The only thing this breaks is whatever new features were added are not available on regions with bad luck ordering Amis

chuckha:
DescribeImages does not guarantee an order so we order it in code by timestamp. The newest will alwyas be assumed the best.

This PR sorts the output of DescribeImages so we always get the newest AMI.

The strategy for testing is to override the AMI id in the Machine object.

Fixes #558 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/ami.go
+++ b/pkg/cloud/aws/services/ec2/ami.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2018 The Kubernetes Authors.
+Copyright 2019 The Kubernetes Authors.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
@@ -18,7 +18,9 @@ package ec2
 
 import (
 	"fmt"
+	"sort"
 	"strings"
+	"time"
 
 	"k8s.io/klog"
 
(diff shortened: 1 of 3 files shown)


chuckha:
This returned the correct ami for us-west-2.

randomvariable:
needs a `go fmt`

pkg/cloud/aws/services/ec2/ami.go
```diff
@@ -80,8 +85,43 @@ func (s *Service) defaultAMILookup(baseOS, baseOSVersion, kubernetesVersion stri
...
+
+// Swap swaps the elements with indexes i and j.
+func (i images) Swap(k, j int) {
+	i[k], i[j] = i[j], i[k]
+}
+
+func getLatestImage(imgs []*ec2.Image) *ec2.Image {
+	// old to new (newest one is last)
+	sort.Sort(images(imgs))
+	return imgs[len(imgs)-1]
```
vincepri: This can panic if imgs is empty

chuckha: true, but the assumption of this code, which perhaps I should document, is that it won't be empty

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/563/files/6715f7002a93f2b3c590abd5d761e2f83d98eb7b#diff-26464ba5826fa06de3ee8d0a402f6efcR85

Issue #272, opened 2018-10-17
Periodic e2e testing


detiber:
/kind feature

wire up periodic full e2e tests 

/assign @liztio

timothysc:
This should be the conformance suite on a full cluster deployment.

liztio:
# Current Status

## E2E Tests

### Background

We spent a lot of time discussing what exactly the E2E tests should entail. 
The current "e2e" tests just build Docker images and test that they can boot.
This is a decent smoke test, but to validate our assumptions more detailed tests.

The proposed minimum viable test stack is as follows:

1. The tests get an AWS account
2. A [kind][kind] kubernetes cluster is booted
3. The ClusterAPI and ClusterAPIProviderAWS controllers are created in the cluster
4. A cluster object is created using the Cluster API
5. A machine object is created in said cluster
6. We wait for the cluster api aws provider controller to mark the created machine as "ready"

### Current Status

_[E2E Pull Request][e2e-pr]_

I've wrapped up [a pull request][e2e-pr] that implements the above steps. 
It makes several assumptions about the environment it's running in.

First, `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY`.
The [`checkin_account.py`][checkin] scripts will set these. 
It's intended to be used as `eval $(./hack/checkin_account.py)`
Those scripts expect a `BOSKOS_HOST` environment variable, in the form of `localhost:8080`. 
[`checkout_account.py`][checkout] also needs the name of the account from boskos.
The `eval()` command will set this.

[kind]: https://github.com/kubernetes-sigs/kind
[e2e-pr]: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/573
[checkin]: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/hack/checkin_account.py
[checkout]: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/hack/checkout_account.py

### Future

#### Deployment

The existing integration tests are run as E2E in [`scripts/ci-bazel-e2e.sh`][ci-e2e]. If we choose to make
the new tests E2E, we will only need to modify that script.

If we wish to instead use periodic E2E (as this issue requests), we will need to [create a new prow job][job].

[ci-e2e]: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/scripts/ci-bazel-e2e.sh
[...]

PR #564, opened 2019-02-08
Python scripts for checking in/out Boskos AWS accounts

liztio:

We are working on using Boskos for managing AWS credentials with 

Part of #272 

Why Python?
* Don't need to worry about proper GOPATH handling
* Prior art for Python scripts in `hack/`

I can rewrite this in Go if need be!

**Release note**:

```release-note
NONE
```


liztio:
yes, it _did_ pain me to write python2 code :cry:

detiber:
I think python is fine for this use case.

/approve
/lgtm

PR #568, opened 2019-02-11
Fix repeated error messages for clusterawsadm

### LINKED ISSUE #550: Repeated error message with clusterawsadm
/kind bug

I was trying to create a cloudformation stack with `clusterawsadm`'s `create-stack` command at which point I stumbled upon the case where an error was displayed twice.

```
➜  cluster-api-provider-aws git:(master) ✗ clusterawsadm alpha bootstrap create-stack
Attempting to create CloudFormation stack cluster-api-provider-aws-sigs-k8s-io
Error: failed to create AWS CloudFormation stack: AccessDenied: User: arn:aws:iam::XXXXXXXXXXXX:user/bootstrapper.cluster-api-provider-aws.sigs.k8s.io is not authorized to perform: cloudformation:CreateStack on resource: arn:aws:cloudformation:us-east-2:XXXXXXXXXXXX:stack/cluster-api-provider-aws-sigs-k8s-io/*
	status code: 403, request id: cb7b5896-2958-11e9-92e4-d31491798876
Usage:
  clusterawsadm alpha bootstrap create-stack [flags]

Flags:
  -h, --help   help for create-stack

failed to create AWS CloudFormation stack: AccessDenied: User: arn:aws:iam::XXXXXXXXXXXX:user/bootstrapper.cluster-api-provider-aws.sigs.k8s.io is not authorized to perform: cloudformation:CreateStack on resource: arn:aws:cloudformation:us-east-2:XXXXXXXXXXXX:stack/cluster-api-provider-aws-sigs-k8s-io/*
	status code: 403, request id: cb7b5896-2958-11e9-92e4-d31491798876
```

1. An error should be displayed only once. 
2. **If** folks are in favor of this: The `cmd.Help()` function should only be called when there is an error with the command (for example, incorrect flags being specified) instead of showing the `Usage:` instruction everytime there is any error.

1. The error is intended and correct. It's just that it is displayed twice.
2. `clusterawsadm`'s binary was built by `make clusterawsadm` and a released binary was not used for this purpose.

**Environment:**

- Cluster-api-provider-aws version: master
- Kubernetes version: (use `kubectl version`): `v1.12.0`
- OS (e.g. from `/etc/os-release`): macOS Mojave

ishantanu:
Signed-off-by: ishantanu <shantanud106@gmail.com>

1. Fixes repeated error messages for clusterawsadm CLI.
2. Adds a validation for arguments passed to `clusterawsadm`'s `create-stack` command.
3. Skips printing `Usage:` when a genuine error occurs for `create-stack`. Only prints `Usage` details when invalid arguments are passed.

Fixes #550

### DIFF
--- a/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
+++ b/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
@@ -122,18 +122,7 @@ func createStackCmd() *cobra.Command {
 		Use:   "create-stack",
 		Short: "Create a new AWS CloudFormation stack using the bootstrap template",
 		Long:  "Create a new AWS CloudFormation stack using the bootstrap template",
-
-		Args: func(cmd *cobra.Command, args []string) error {
-			if len(args) >= 1 {
-				fmt.Printf("Error: No arguments are required for create-stack. \n")
-				if err := cmd.Help(); err != nil {
-					return err
-				}
-				os.Exit(400)
-			}
-
-			return nil
-		},
+		Args:  cobra.NoArgs,
 		RunE: func(cmd *cobra.Command, args []string) error {
 			stackName := "cluster-api-provider-aws-sigs-k8s-io"
 			fmt.Printf("Attempting to create CloudFormation stack %s\n", stackName)
@@ -141,21 +130,22 @@ func createStackCmd() *cobra.Command {
 				SharedConfigState: session.SharedConfigEnable,
 			})
 			if err != nil {
-				return err
+				fmt.Printf("Error: %v", err)
+				return nil
 			}
 
 			stsSvc := sts.NewService(awssts.New(sess))
 			accountID, stsErr := stsSvc.AccountID()
 			if stsErr != nil {
 				fmt.Printf("Error: %v", stsErr)
-				os.Exit(401)
+				return nil
 			}
 
 			cfnSvc := cloudformation.NewService(cfn.New(sess))
 			err = cfnSvc.ReconcileBootstrapStack(stackName, accountID)
 			if err != nil {
 				fmt.Printf("Error: %v", err)
-				os.Exit(402)
+				return nil
 			}
 
 			return cfnSvc.ShowStackResources(stackName)
--- a/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
+++ b/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
@@ -122,6 +122,18 @@ func createStackCmd() *cobra.Command {
 		Use:   "create-stack",
 		Short: "Create a new AWS CloudFormation stack using the bootstrap template",
 		Long:  "Create a new AWS CloudFormation stack using the bootstrap template",
+
+		Args: func(cmd *cobra.Command, args []string) error {
+			if len(args) >= 1 {
+				fmt.Printf("Error: No arguments are required for create-stack. \n")
+				if err := cmd.Help(); err != nil {
+					return err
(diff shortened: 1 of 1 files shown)


cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
```diff
@@ -135,13 +147,15 @@ func createStackCmd() *cobra.Command {
 			stsSvc := sts.NewService(awssts.New(sess))
 			accountID, stsErr := stsSvc.AccountID()
 			if stsErr != nil {
-				return stsErr
+				fmt.Printf("Error: %v", stsErr)
+				os.Exit(401)
```
chuckha: We should not be manually exiting from within a `RunE` function or Args for that matter.

I believe what you're seeing is a result of this issue: https://github.com/spf13/cobra/issues/340. We should consider evaluating our usage of `RunE` instead of hacking our way around its rather opinionated conventions. Doing this will help new and existing contributors understand the flow of the code more easily since it's not obvious why we need `os.Exit` here.

ishantanu: @chuckha Would you recommend using simply `Run` instead of `RunE`? However, `Run` does not return anything by default. And, yes, we don't need `os.Exit` here (my mistake, a `return nil` along with printing error could also do it).

chuckha: I would have to look into it more. I suspect replacing `RunE` with `Run` would be sufficient, but I can't say for sure without spending some time on it as cobra has never really been intuitive for me.

ishantanu: Well, replacing `RunE` with `Run` and keeping everything else same would not work since `Run` does not return anything. That also means every `return err` will be replaced by `fmt.Printf("Error: %v", err); return nil`. And, if the convention of `Run` is to be followed over `RunE`, then subsequent `RunE`'s will be also be required to be replaced by `Run`.

Or

If we are to keep `RunE` as is for now, I can replace the `os.Exit()` calls from `Args` and other places with something like:
```
if stsErr != nil {
	fmt.Printf("Error: %v", stsErr)
	return nil
}
```
Do let me know which way would you prefer and I will make the appropriate changes.

chuckha: I think that's a reasonable suggestion, keeping RunE and replacing os.Exit from Args with return nil and printing the error. Feels like we should clean that up eventually though and not use RunE unless we want that behavior. @randomvariable what do you think about this?

ishantanu: Is there going to be more discussion around the possible solution which was discussed in the above comments? Or should I push the changes?

chuckha: please push the changes

ishantanu: Updated. Also used cobra's built-in function `cobra.NoArgs` for validation on number of arguments passed instead of a manual check.

chuckha:
/assign @randomvariable 
assigning for approval

cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
randomvariable: I think someone else changed it from Run to RunE to fix a different UI issue, so I don't think I'm any wiser.

PR #571, opened 2019-02-12
Add support for MachineClass

### LINKED ISSUE #513: Implement MachineClassProviderSpec
/kind feature

In order to make full use of the cluster-api APIs we will need to implement a MachineClass provider spec that can be used to specify machine types on AWS.

We may decided we don't need this part of cluster-api. If that is the case, close this issue.

Please see https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/machineclass_types.go#L33

**Environment:**

detiber:

Adds support for retrieving the ProviderSpec from a MachineClass

Fixes #513

**Release note**:
<!--  Write your release note:
```release-note
- cluster-api-provider-aws now supports using MachineClass
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine_scope.go
+++ b/pkg/cloud/aws/actuators/machine_scope.go
@@ -18,10 +18,13 @@ package actuators
 
 import (
 	"github.com/pkg/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/klog"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1alpha1"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
+	"sigs.k8s.io/yaml"
 )
 
 // MachineScopeParams defines the input parameters used to create a new MachineScope.
@@ -40,7 +43,7 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 		return nil, err
 	}
 
-	machineConfig, err := v1alpha1.MachineConfigFromProviderSpec(params.Machine.Spec.ProviderSpec)
+	machineConfig, err := machineConfigFromProviderSpec(params.Client, params.Machine.Spec.ProviderSpec)
 	if err != nil {
 		return nil, errors.Wrap(err, "failed to get machine config")
 	}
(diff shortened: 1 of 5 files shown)


detiber:
/assign @vincepri 

@vincepri @chuckha with these changes I've been able to spin up Machines using a MachineClass instead of an embedded provider spec.

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -131,3 +134,46 @@ func (m *MachineScope) Close() {
...
+	var config v1alpha1.AWSMachineProviderSpec
+	if providerConfig.Value != nil {
+		klog.V(4).Info("Decoding ProviderConfig from Value")
+		return unmarshalProviderSpec(providerConfig.Value)
+	}
+
+	if providerConfig.ValueFrom != nil && providerConfig.ValueFrom.MachineClass != nil {
+		ref := providerConfig.ValueFrom.MachineClass
+		klog.V(4).Info("Decoding ProviderConfig from MachineClass")
+		if ref != nil {
```
chuckha: we already know `providerConfig.ValueFrom.MachineClass` is != nil, this check can go away since it's on line 145

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -131,3 +134,46 @@ func (m *MachineScope) Close() {
...
+			klog.V(6).Infof("ref: %v", ref)
+			if ref.Provider != "" && ref.Provider != "aws" {
```
chuckha: does this imply that the default provider is "aws" when `ref.Provider` is empty?

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -131,3 +134,46 @@ func (m *MachineScope) Close() {
...
+}
+
+func unmarshalProviderSpec(spec *runtime.RawExtension) (*v1alpha1.AWSMachineProviderSpec, error) {
+	var config v1alpha1.AWSMachineProviderSpec
+	if spec != nil {
+		if err := yaml.Unmarshal(spec.Raw, &config); err != nil {
+			return nil, err
+		}
+	}
+	klog.V(6).Infof("Found ProviderSpec: %v", config)
```
chuckha: I think `%+v` might be better for this debug statement

pkg/cloud/aws/actuators/machine_scope.go
detiber: I figured we should also attempt to decode the providerSpec if the user doesn't specify a provider in the MachineClass, so instead of returning an error here the error will be returned when attempting to unmarshal the providerSpec if it is meant for a different provider.

PR #576, opened 2019-02-14
[WIP] Rev versions of containerd and Kubernetes

### LINKED ISSUE #570: Update version of containerd used in AMIs
/kind bug
/priority important-soon
/milestone v1alpha1

The AMIs generated by the packer build tooling are affected by CVE-2019-5736, we should update to a version of containerd that is not affected.

Blocked on: https://github.com/containerd/containerd/issues/3006

detiber:

- Update k8s to latest patch release
- Update containerd for CVE-2019-5736

Fixes #570 

**Release note**:
```release-note
- Update default AMI and examples for Kubernetes v1.13.1
- Update default AMI for CVE-2019-5736
```


build/amis/ansible/roles/containerd/defaults/main.yml
```diff
@@ -12,5 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 ---
-containerd_version: "1.2.1"
+containerd_version: "1.2.4"
```
sfzylad: I would suggest to add a goss test to make sure we are good here.

build/amis/packer/packer.json
```diff
@@ -3,7 +3,7 @@
     "aws_access_key": "",
     "aws_secret_key": "",
     "build_timestamp": "{{timestamp}}",
-    "kubernetes_version": "1.13.2-00",
+    "kubernetes_version": "1.13.3-00",
```
sfzylad: Same here. It may be worth testing if the packages we want to install are there.

build/amis/ansible/roles/containerd/defaults/main.yml
randomvariable: Binaries are on containerd github, but not in the GCS bucket we're using. Going to see how easy it is to switch.

detiber: I tried that, unlike the GCS bucket binaries the GitHub release does not have the systems unit file.

build/amis/packer/packer.json
detiber: I don't think a Goss test is needed here because the Ansible playbook will fail if the specified version of the package is not installed.

build/amis/ansible/roles/containerd/defaults/main.yml
detiber: The Ansible playbook will fail if the specified version of containerd isn't installed. A Goss test for this would be redundant.

randomvariable: We added a check to ensure containerd is running and that `crictl ps` succeeds.

detiber:
@chuckha this should be good to go now.

PR #594, opened 2019-02-21
Fix manifest-dev workflow after Bazel updates

vincepri:

This PR is a follow-up to #587, in details:
- Updates provider-components-base files and adds PHONY directive in Makefile to make sure they are always up to date.
- Updates the `create-cluster` Makefile directive to use `-dev` components.
-  Regenerates CRDs.

Fixes #

**Release note**:

```release-note

```


Makefile
```diff
@@ -171,7 +173,7 @@ create-cluster: ## Create a Kubernetes cluster on AWS using examples
 	--bootstrap-type kind \
 	-m ./cmd/clusterctl/examples/aws/out/machines.yaml \
 	-c ./cmd/clusterctl/examples/aws/out/cluster.yaml \
-	-p ./cmd/clusterctl/examples/aws/out/provider-components.yaml \
+	-p ./cmd/clusterctl/examples/aws/out/provider-components-dev.yaml \
```
detiber: Considering this is using the `-dev` resources here, should this target be renamed to `create-cluster-dev` instead?

cmd/clusterctl/examples/aws/generate-yaml.sh
```diff
@@ -104,16 +104,16 @@ echo "Done copying ${ADDONS_FILE}"
...
-echo "Done writing ${PROVIDER_COMPONENTS_FILE}"
-echo "WARNING: ${PROVIDER_COMPONENTS_FILE} includes credentials"
-
 if [ -f $PROVIDER_COMPONENTS_SRC_DEV ]; then
   PROVIDER_COMPONENTS_DEV=$(cat ${PROVIDER_COMPONENTS_SRC_DEV})
 
   echo -e "${PROVIDER_COMPONENTS_DEV}\n${CREDENTIALS}" > "${PROVIDER_COMPONENTS_FILE_DEV}"
   echo "Done writing ${PROVIDER_COMPONENTS_FILE_DEV}"
   echo "WARNING: ${PROVIDER_COMPONENTS_FILE_DEV} includes credentials"
+else
```
chuckha: I don't super love this bash change because of our current manual release process. I agree with it being technically correct, but it's more inconvenient and I can see this leading to a few minutes of confusion while someone works out they have to delete their dev file to do a release build.

vincepri:
Is it ok if I follow up to cleanup that part another time? I agree it's a little yucky and should be fixed, but given the focus on deletion and other broken things I gave priority elsewhere.

Makefile
chuckha: since we've now brought make out of the mix for users, I think the implication is that it will always be a dev target

PR #596, opened 2019-02-21
Update cluster-api dep to latest / ownerRef

vincepri:

This PR updates cluster-api dependency which sets the ownerRef to Machines.

Fixes #

**Release note**:

```release-note

```


Makefile
```diff
@@ -159,15 +159,18 @@ crds:
 	cp -R bazel-genfiles/config/crds/* config/crds/
 	cp -R bazel-genfiles/config/rbac/* config/rbac/
 
-# TODO(vincepri): This should move to rebuild Bazel binaries once every
-# make target uses Bazel bins to run operations.
```
detiber: Shouldn't these comments persist since `binaries-dev` is not yet updated to use bazel?

Makefile
```diff
@@ -159,15 +159,18 @@ crds:
...
+## Define local development targets here
+
 .PHONY: binaries-dev
 binaries-dev: ## Builds and installs the binaries on the local GOPATH
 	go get -v ./...
 	go install -v ./...
 
 .PHONY: create-cluster
-create-cluster: ## Create a Kubernetes cluster on AWS using examples
+create-cluster: binaries-dev ## Create a Kubernetes cluster on AWS using examples
```
randomvariable: Maybe worth considering ditching this in favour of `bazel run //test/e2e` ?

Makefile
vincepri: I removed them because I think for the time being these dev binaries will stay on the GOPATH and should be rebuilt this way. Moving those to Bazel is optional and could be done later if there is real necessity.

detiber: Actually, do we even need this step now? Or should it be updated to just call the `clusterctl` and `clusterawsadm` targets? Those targets already install the binaries into the ${GOPATH}/bin

PR #598, opened 2019-02-22
Configurable region

akutz:
This patch makes the AWS region used for e2e configurable via the environment variable `AWS_DEFAULT_REGION`, falling back to `AWS_REGION` if the former is not specified. If not specified, the region defaults to `us-east-1`.

NA

You're special, and gosh darnit, people like you.

**Release note**:
```release-note
The AWS region used for e2e may now be configured with the environment variable AWS_DEFAULT_REGION, falling back to AWS_REGION if the former is not specified. The region defaults to "us-east-1".
```

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -19,6 +19,7 @@ package e2e_test
 import (
 	"flag"
 	"io/ioutil"
+	"os"
 	"time"
 
 	. "github.com/onsi/ginkgo"
@@ -57,16 +58,24 @@ const (
 	clusterName      = "capa-test-cluster"
 	controlPlaneName = "capa-test-control-plane"
 
-	awsRegion   = "us-east-1"
 	stackName   = "cluster-api-provider-aws-sigs-k8s-io"
 	keyPairName = "cluster-api-provider-aws-sigs-k8s-io"
 )
 
 var (
+	region      string
 	credFile    = flag.String("credFile", "", "path to an AWS credentials file")
 	clusterYAML = flag.String("clusterYAML", "", "path to the YAML for the cluster we're creating")
 	machineYAML = flag.String("machineYAML", "", "path to the YAML describing the control plane we're creating")
 )
+
+func init() {
+	if region = os.Getenv("AWS_DEFAULT_REGION"); region == "" {
+		if region = os.Getenv("AWS_REGION"); region == "" {
+			region = "us-east-1"
+		}
+	}
+}
 
 var _ = Describe("AWS", func() {
 	var (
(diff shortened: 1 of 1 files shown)


pkg/cloud/aws/regions/regions.go
```diff
@@ -0,0 +1,28 @@
...
+
+package regions
+
+import "os"
+
+// Default is the region used when no region is specified.
+var Default string
+
+func init() {
+	if Default = os.Getenv("AWS_DEFAULT_REGION"); Default == "" {
```
detiber: Instead of creating a new variable here, would it make sense to use `AWS_REGION` here instead of `AWS_DEFAULT_REGION`? We don't yet have support for specifying specific regions and end-users would have to update the aws-provider-controller-manager StatefulSet definition to set this env var, which might be more intuitive if it aligns with the standard AWS environment variables.

akutz: Hi @detiber,

So `AWS_DEFAULT_REGION` is the standard value. Please see https://github.com/aws/aws-sdk-go/issues/2103 for more information.

pkg/cloud/aws/services/ec2/account.go
```diff
@@ -22,18 +22,15 @@ import (
...
 	"github.com/pkg/errors"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/filter"
-)
-
-const (
-	defaultRegion = "us-east-1"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/regions"
 )
 
 func (s *Service) getRegion() string {
```
chuckha: This function is unused and can be deleted instead

pkg/cloud/aws/regions/regions.go
akutz: Hi @detiber,

I am happy to support the fallback and support `AWS_REGION` if `AWS_DEFAULT_REGION` is not set.

randomvariable:
you need `make gazelle` here to update the Bazel build files for the new import.

pkg/cloud/aws/regions/regions.go
detiber: TIL, I'm not sure how I never came across that before.

akutz:
> you need `make gazelle` here to update the Bazel build files for the new import.

Got it, thank you @randomvariable.

akutz:
Hi @chuckha,

I've rebased the commit to remove changes to `account.go`. Please feel free to make that change yourself. This PR now simply isolates the region to a single package and the e2e test consumes that package's value.

pkg/cloud/aws/regions/regions.go
```diff
@@ -0,0 +1,30 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package regions
+
+import "os"
+
+// Default is the region used when no region is specified.
+var Default string
```
chuckha: since this isn't how cluster-api-provider-aws sets regions for sessions could this be kept to the test package? Otherwise I'm concerned this will cause confusion (much like you experienced) when people are reading through this code.

akutz: Hi @chuckha,

I think that makes sense. I'll update the PR.

akutz:
> This appears to be a change related to the e2e tests, conversation on slack here https://kubernetes.slack.com/archives/CD6U2V71N/p1550859218075700
> 
> Please file an issue explaining what the issue so we can link this PR back to an issue.

Hi @chuckha,

I'm updating the PR to isolate the changes to the e2e tests. I also recommend updating the documentation and removing the unused functions to prevent future confusion.

PR #607, opened 2019-02-25
Add cli flag to supply namespace for capa controller to watch for capi objects

### LINKED ISSUE #581: Optional ability to restrict to a single namespace
/kind feature

I'm interested in being able to manage clusters in multiple AWS accounts. I believe a natural and easy way to achieve this could be to partition by namespace, where all the clusters in a single namespace belong to the same AWS account.

I'd like to add an optional ability to restrict CAPA so it watches for changes only in a single namespace, instead of across all namespaces. Operators using this option would need to deploy 1 CAPA controller pod per namespace/AWS account if they want to support multiple accounts.

The controller-runtime `manager.Options` struct has a `Namespace` field that restricts the listing/watching to the specified namespace. I'm not sure if there have been any previous discussions about how to configure the controller, but we could add an optional flag to the command line for this.

**Environment:**

ashish-amarnath:
Add cli flag to supply namespace for capa controller to watch for capi objects

This PR adds a CLI flag to restrict the namespace that capa controllers watch for capi objects to reconcile

Fixes #581

**Release note**:

```release-note

```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -44,13 +44,13 @@ func main() {
 	cfg := config.GetConfigOrDie()
 
 	// Setup a Manager
-	var mgrOps manager.Options
+	var opts manager.Options
 	if *watchNamespace != "" {
-		mgrOps.Namespace = *watchNamespace
-		klog.Infof("Reconciling cluster-api objects in namespace %q", mgrOps.Namespace)
+		opts.Namespace = *watchNamespace
+		klog.Infof("Watching cluster-api objects in only in namespace %q for reconciliation.", opts.Namespace)
 	}
 
-	mgr, err := manager.New(cfg, mgrOps)
+	mgr, err := manager.New(cfg, opts)
 	if err != nil {
 		klog.Fatalf("Failed to set up overall controller manager: %v", err)
 	}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -37,12 +37,20 @@ import (
 func main() {
 	klog.InitFlags(nil)
 	flag.Set("logtostderr", "true")
+	watchNamespace := flag.String("namespace", "",
+		"Namespace that the controller watches to reconcile cluster-api objects. If unspecified, the controller watches for cluster-api objects across all namespaces.")
 	flag.Parse()
 
 	cfg := config.GetConfigOrDie()
 
 	// Setup a Manager
-	mgr, err := manager.New(cfg, manager.Options{})
+	var mgrOps manager.Options
+	if *watchNamespace != "" {
+		mgrOps.Namespace = *watchNamespace
+		klog.Infof("Reconciling cluster-api objects in namespace %q", mgrOps.Namespace)
+	}
+
+	mgr, err := manager.New(cfg, mgrOps)
 	if err != nil {
 		klog.Fatalf("Failed to set up overall controller manager: %v", err)
 	}
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -47,7 +47,7 @@ func main() {
 	var opts manager.Options
 	if *watchNamespace != "" {
 		opts.Namespace = *watchNamespace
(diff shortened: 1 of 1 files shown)


cmd/manager/main.go
```diff
@@ -37,12 +37,20 @@ import (
 func main() {
 	klog.InitFlags(nil)
 	flag.Set("logtostderr", "true")
+	watchNamespace := flag.String("namespace", "",
```
vincepri: Should we define this outside the main function instead?

cmd/manager/main.go
```diff
@@ -37,12 +37,20 @@ import (
...
+		"Namespace that the controller watches to reconcile cluster-api objects. If unspecified, the controller watches for cluster-api objects across all namespaces.")
 	flag.Parse()
 
 	cfg := config.GetConfigOrDie()
 
 	// Setup a Manager
-	mgr, err := manager.New(cfg, manager.Options{})
+	var mgrOps manager.Options
```
vincepri:
```suggestion
	var opts manager.Options
```

cmd/manager/main.go
```diff
@@ -37,12 +37,20 @@ import (
...
+	if *watchNamespace != "" {
+		mgrOps.Namespace = *watchNamespace
+		klog.Infof("Reconciling cluster-api objects in namespace %q", mgrOps.Namespace)
+	}
+
+	mgr, err := manager.New(cfg, mgrOps)
```
vincepri:
```suggestion
	mgr, err := manager.New(cfg, opts)
```

cmd/manager/main.go
```diff
@@ -37,12 +37,20 @@ import (
...
 	flag.Parse()
 
 	cfg := config.GetConfigOrDie()
 
 	// Setup a Manager
-	mgr, err := manager.New(cfg, manager.Options{})
+	var mgrOps manager.Options
+	if *watchNamespace != "" {
+		mgrOps.Namespace = *watchNamespace
+		klog.Infof("Reconciling cluster-api objects in namespace %q", mgrOps.Namespace)
```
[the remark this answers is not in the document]
vincepri: This message is going to be shown only once upon start. I'm not sure if `Reconciling` is actually correct given that the reconciliation step is done in a different loop. I'd change this to something more concrete that warns the user that the objects this manager can see are limited to a single namespace.

cmd/manager/main.go
ashish-amarnath: If we had more cli flags, it would make sense to create a `CliFlags` struct and parse the flags in a separate function that would return the `CliFlags` struct. Considering we have just one flag other than the `logtostderr` this should be good.

cmd/manager/main.go
ashish-amarnath: fixed

cmd/manager/main.go
ashish-amarnath: fixed

cmd/manager/main.go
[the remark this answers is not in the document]
ashish-amarnath: Agree. Reworded the log msg to read
`Watching cluster-api objects in only in namespace %q for reconciliation.`

cmd/manager/main.go
```diff
@@ -37,12 +37,20 @@ import (
...
 	flag.Parse()
 
 	cfg := config.GetConfigOrDie()
 
 	// Setup a Manager
-	mgr, err := manager.New(cfg, manager.Options{})
+	var opts manager.Options
+	if *watchNamespace != "" {
+		opts.Namespace = *watchNamespace
+		klog.Infof("Watching cluster-api objects in only in namespace %q for reconciliation.", opts.Namespace)
```
detiber:
```suggestion
		klog.Infof("Watching cluster-api objects only in namespace %q for reconciliation.", opts.Namespace)
```

detiber:
/approve

leaving lgtm to @vincepri

cmd/manager/main.go
ashish-amarnath: derp.. 😝 fixed it. sorry for not catching that myself

PR #606, opened 2019-02-23
e2e tests

akutz:
**What this PR does / why we need it**:
This patch updates documentation, the Makefile, and creates some new programs under the `hack` directory in order to make e2e testing as easy as possible, both locally and as a invocation point for Prow jobs.

This is a work-in-progress, and the PR is being opened as a point of discussion and to provide an opportunity for the community to give feedback early in the process.

Blocked by:
1. ~~https://github.com/kubernetes/test-infra/pull/11425~~
2. ~~https://github.com/kubernetes/test-infra/issues/11471~~
3. ~~https://github.com/kubernetes/test-infra/pull/11482~~
4. ~~https://github.com/kubernetes/test-infra/pull/11504~~

/hold

**Which issue(s) this PR fixes**:
Partially-Sorta-Kinda-Phixes #272 (see **Special notes**)

**Special notes for your reviewer**:
This PR alone will not phix #272 as a period Prow job will need to be created. However, the content of this PR will be how Prow initiates the e2e testing.

**Release note**:
```release-note
Execute unit tests with "make test", integration tests with "make integration", and end-to-end tests with "make e2e".
```

PR #612, opened 2019-02-26
[WIP] Tighter IAM policies

randomvariable:
**What this PR does / why we need it**:

EC2 and ELB instances are tagged upon creation. Therefore, ensuring that only EC2/ELB instances are deleted when they match the Cluster API Provider AWS tag minimises privileges.

This approach cannot be used for other resources as they are tagged after creation - backing out of a failed tagging may prevent deletion, and controller can tag a resource appropriately to delete it.

Also additionally constrain which tags are allowed to be used by the controller.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #608 

**Special notes for your reviewer**:

In order to restrict privilege escalation via instance profile, all resources are now put under a `sigs.k8s.io/cluster-api-provider-aws` path so this can be referenced in a policy condition. Unfortunately, this breaks CloudFormation being able to do an in-place update. There doesn't seem to be a way around this, so have also taken the opportunity to rename two incorrectly named resources. These are in the release notes below.

Recommend this is not merged prior to e2e tests making use of CloudFormation with a delete process.

Current example policy:
``` json
{
    "Version": "2012-10-17",
    "Statement": [
        {
            "Condition": {
                "Bool": {
                    "aws:SecureTransport": [
                        "true"
                    ]
                }
            },
            "Action": [
                "ec2:AllocateAddress",
                "ec2:AssociateRouteTable",
                "ec2:AttachInternetGateway",
                "ec2:AuthorizeSecurityGroupIngress",
                "ec2:CreateInternetGateway",
                "ec2:CreateNatGateway",
                "ec2:CreateRoute",
                "ec2:CreateRouteTable",
                "ec2:CreateSecurityGroup",
                "ec2:CreateSubnet",
                "ec2:CreateVpc",
                "ec2:DeleteInternetGateway",
                "ec2:DeleteNatGateway",
                "ec2:DeleteRouteTable",
                "ec2:DeleteSecurityGroup",
                "ec2:DeleteSubnet",
                "ec2:DeleteVpc",
                "ec2:DescribeAddresses",
                "ec2:DescribeAvailabilityZones",
                "ec2:DescribeInternetGateways",
                "ec2:DescribeImages",
                "ec2:DescribeNatGateways",
                "ec2:DescribeRouteTables",
                "ec2:DescribeSecurityGroups",
                "ec2:DescribeSubnets",
[...]

PR #617, opened 2019-03-01
[WIP] Embed example templates into clusterawsadm

### LINKED ISSUE #299: Remove requirement for GNU Make, envsubst & gettext for end-user experience
/kind feature

Part of #276 

Having gettext, envsubst and GNU Make installed is unusual for end-user
software, and it's extra things that will need to be 
`brew/apt/dnf` installed, as Make in particular is not installed
by default on all systems.

**Environment:**

randomvariable:
Removes gettext by starting to embed the templates into `clusterawsadm`.

User will run the following to create their examples:

```
clusterawsadm alpha config init --cluster-name default --ssh-key-name default --control-plane-instance-type t3.medium <etc...>
```

Which will write the same outputs as before.

Also moving the examples to a top-level `/examples` directory for better visibility. Don't think they need to live under `/cmd/clusterctl/examples`

These files are used by a Bazel rule to generate a go file with the files zipped into a Go virtual file stream which is then used by the CLI to generate the examples as rendered Go templates.

Fixes #299 

This PR will require changes to e2e, so will wait for #606 before completion.

**Release note**:

```release-note

```


randomvariable:
Actually agree on this.

Only doing CLI flags as equivalents of the environment variables as used by envsubst in `generate-yaml.sh`.

What behaviour do we want to keep, and what should be jettisoned. 

Don't see the point, for example of having the instance type env vars when we as cluster api devs can use kustomize patches for our workflows, and an end user should be expected to inspect and edit the cluster and machine yamls.

randomvariable:
How about separate commands for cluster, machine and provider components, but with the machine command, you can pass in `--cluster-config cluster.yaml` and it will generate machine.yamls with the same cluster name and ssh key?

chuckha:
so the CLI might look something like this:

```
clusterawsadm alpha config init cluster
clusterawsadm alpha config init machine --cluster-config cluster.yaml
clusterawsadm alpha config init provider-components
```

What would cluster.yaml look like?

randomvariable:
Was thinking of exposing the user to the real cluster object at this stage rather than present a clusterawsadm abstraction, but using the cluster object as a hint for some of the fields for new machines.

chuckha:
ah so, the output of `clusterawsadm alpha config init cluster` would be fed into `clusterawsadm alpha config init machine --cluster-config cluster.yaml` ?

ashish-amarnath:
That's how I thought this would work. This will be kinda guiding the user through the steps to help them understand what they are getting and what is all under their control.

randomvariable:
Yup, that's exactly it.

vincepri:
/milestone Next

Feel free to change if necessary.

randomvariable:
The underlying issue is v1alpha1

Going to finish this one before #612 .

detiber:
closing, we can reconsider down the line.

/close

PR #619, opened 2019-03-04
Automate release process

chuckha:
Using goreleaser this will automatically build binaries,
collect resources, package them into appropriate tarballs
and upload them to github. There is commented out code
for also pushing container images to gcr.io that we could
consider using.

This PR automates the release process. It may not be the way we want to go but it could be a nice first option.

This uses goreleaser to automate the release process. It however does have some ups and downs.

The pros of this are:
* Automatic GitHub release drafts with notes
* Can also push container images to gcr.io (with correct permissions)
* Wraps up binaries and files together

The cons of this:
* It's awkward to use this in conjunction with bazel as our build tool because bazel wants to do things its own way.
* There is no way to specify where files end up in the release tar. See: https://github.com/goreleaser/goreleaser/issues/679
* Building a hack binary only to overwrite it with Bazel
* cannot use docker feature of goreleaser because we use bazel to build docker images

```release-note

```

### DIFF
--- a/.goreleaser.yml
+++ b/.goreleaser.yml
@@ -1,0 +1,45 @@
+# Consider this for container image release.
+# before:
+#   hooks:
+#     - make docker-push
+builds:
+  - env:
+      - CGO_ENABLED=0
+    main: ./cmd/noop/main.go
+    binary: clusterctl
+    goarch:
+      - amd64
+    hooks:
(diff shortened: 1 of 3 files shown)


chuckha:
closing since I'm not a huge fan of this

PR #634, opened 2019-03-07
fix: repair node joins

sethp-nr:
We observed that the user-data being created for nodes had illegal
indentation, and omitted the port from the api server address.

Previously, we observed user-data being generated that looks like:

```
    permissions: '0640'
    content: |
      ---
            apiVersion: kubeadm.k8s.io/v1beta1
      caCertPath: ""
      discovery:
        bootstrapToken:
```

Removing the leading spaces produced a cloud-config that seem 

**Release note**:

```release-note
NONE
```


chuckha:
/lgtm

thanks for the fix

PR #635, opened 2019-03-08
Fix bug where additional security groups and tags are not applied

rudoi:

Currently, the AdditionalSecurityGroups and AdditionalTags fields do not function. This is due to the following:

- `a.isMachineOutdated` always returns true because the Instance struct has no IAMProfile field. This causes the Update function to return, which means `ensureSecurityGroups` and `ensureTags` are never called
- the annotations SecurityGroupsLastAppliedAnnotation and TagsLastAppliedAnnotation don't pass validation, because there are multiple `/`'s
- the bootstrap IAM policy does not have `ec2:ModifyInstanceAttribute`, which is required to add additional security groups

edited to add: bootstrap policy also requires `ec2:DeleteTags` for removing additional tags

**Release note**:

```release-note
Fixes bug where additional security groups and tags are not applied
```


pkg/cloud/aws/converters/instance.go
```diff
@@ -36,14 +38,12 @@ func SDKToInstance(v *ec2.Instance) *v1alpha1.Instance {
 		EBSOptimized: v.EbsOptimized,
 	}
 
+	i.IAMProfile = strings.Split(aws.StringValue(v.IamInstanceProfile.Arn), "instance-profile/")[1]
```
vincepri: We should make sure that the value isn't nil or empty string and check if the length of the slice returned from Split is actually greater or equal to 1. The line above can cause a panic otherwise.

pkg/cloud/aws/actuators/machine/security_groups.go
```diff
@@ -31,7 +31,7 @@ const (
 	// annotation which tracks the SecurityGroups that the machine actuator is
 	// responsible for. These are the SecurityGroups that have been handled by
 	// the AdditionalSecurityGroups in the Machine Provider Config.
-	SecurityGroupsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-aws/last-applied/security-groups"
+	SecurityGroupsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-aws-last-applied-security-groups"
```
vincepri: Could we link the [annotation rules](https://kubernetes.io/docs/concepts/overview/working-with-objects/annotations/) in the comment above?

pkg/cloud/aws/actuators/machine/tags.go
```diff
@@ -27,7 +27,7 @@ const (
 	// which tracks the SecurityGroups that the machine actuator is responsible
 	// for. These are the SecurityGroups that have been handled by the
 	// AdditionalTags in the Machine Provider Config.
-	TagsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-aws/last-applied/tags"
+	TagsLastAppliedAnnotation = "sigs.k8s.io/cluster-api-provider-aws-last-applied-tags"
```
vincepri: Same as above

pkg/cloud/aws/converters/instance.go
```diff
@@ -36,14 +38,12 @@ func SDKToInstance(v *ec2.Instance) *v1alpha1.Instance {
...
+
 	for _, sg := range v.SecurityGroups {
 		i.SecurityGroupIDs = append(i.SecurityGroupIDs, *sg.GroupId)
 	}
 
-	// TODO: Handle returned IAM instance profile, since we are currently
-	// using a string representing the name, but the InstanceProfile returned
-	// from the sdk only returns ARN and ID.
```
vincepri: We should be able to generate the IAM ARN given a name, the account-id can be queried at runtime

pkg/cloud/aws/services/cloudformation/bootstrap.go
```diff
@@ -176,6 +176,7 @@ func controllersPolicy(accountID string) *iam.PolicyDocument {
 					"ec2:DescribeVpcs",
 					"ec2:DetachInternetGateway",
 					"ec2:DisassociateRouteTable",
+					"ec2:ModifyInstanceAttribute",
```
ashish-amarnath: @rudoi can you please clarify what operation needs to perform this action?

rudoi: Sure thing. It is called when updating instance security groups:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/870d03d2d5a25a407b335d3f0b702227dfd7cd59/pkg/cloud/aws/services/ec2/instances.go#L414

pkg/cloud/aws/actuators/machine/security_groups.go
rudoi: done!

pkg/cloud/aws/actuators/machine/tags.go
rudoi: done!

pkg/cloud/aws/converters/instance.go
rudoi: Updated.

pkg/cloud/aws/converters/instance.go
rudoi: Are you suggesting that a better flow would be to look up the Account ID, generate an ARN given the name passed into the instance, and then compare that to the full ARN returned by the EC2 query?

pkg/cloud/aws/converters/instance.go
```diff
@@ -36,14 +38,18 @@ func SDKToInstance(v *ec2.Instance) *v1alpha1.Instance {
 		EBSOptimized: v.EbsOptimized,
 	}
 
+	// Extract IAM Instance Profile name from ARN
+	split := strings.Split(aws.StringValue(v.IamInstanceProfile.Arn), "instance-profile/")
```
vincepri: We should also check that `v.IamInstanceProfile` isn't nil.

pkg/cloud/aws/converters/instance.go
vincepri: We could do that instead and it'd save us the trouble of splitting the string returned from the API and have a more consistent behavior, a TODO comment should do it, we don't have to implement this here.

rudoi: Understood. Thanks for clarifying! Comment added.

pkg/cloud/aws/converters/instance.go
rudoi: done!

rudoi:
Something I discovered while troubleshooting this: It seems that some errors in the Update flow get swallowed - most notably permissions errors returned by AWS APIs. I will likely open a follow-up Issue/PR to make those more clear in the controller logs.

pkg/cloud/aws/services/cloudformation/bootstrap.go
rudoi: I also added DeleteTags, which is called here:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/870d03d2d5a25a407b335d3f0b702227dfd7cd59/pkg/cloud/aws/services/ec2/instances.go#L461

ashish-amarnath: Thanks for clarifying!

pkg/cloud/aws/converters/instance.go
randomvariable: > We should be able to generate the IAM ARN given a name, the account-id can be queried at runtime

This isn't strictly true. IAM objects can be referenced by the "Friendly Name" as they are unique per-object type-per-account, but as an ARN, an IAM object can be stored in a path. This is used for scoping down permissions. For example:

An instance profile stored at:
`arn:aws:iam::123456789012:instance-profile/application_abc/example-profile` can be referenced simply as `example-profile` or the ARN, but `arn:aws:iam::123456789012:instance-profile/example-profile` would result in a not found.

You would need to resolve the ARN of the friendly name using IAM::GetInstanceProfile

rudoi: To clarify, would `foo/bar` be valid as a friendly name? I'm curious as to why having paths would make generating an ARN from friendly name more difficult.

rudoi: So would it be best to update the TODO to reflect that the best end state would be to look up the instance profile by name and compare the ARNs?

randomvariable: Pretty much. I guess the analogues AWS are using is between a fully qualified DN including the Organisational Unit path and a plain Username in LDAP.

rudoi: Makes sense to me. Thanks for the clarification!

PR #639, opened 2019-03-11
Development doc improvements

ashish-amarnath:
Development doc improvements:
* Add instructions to build container images with custom tags
* Add instructions to push to non GCR container registry
* Add troubleshooting section for commonly encountered errors.

Impreoves development doc.
Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/docs/development.md
+++ b/docs/development.md
@@ -103,21 +103,25 @@ The dev version of the manifests can be generated with
 
 1. [Install the gcloud cli][gcloud_sdk].
 2. Set project: `gcloud config set project YOUR_PROJECT_NAME`.
-3. To build images with custom tags, edit [.bazelrc](./.bazelrc) as follows:
+3. To build images with custom tags, run the `make docker-push-dev` as follows:
 
-   `build --define=MANAGER_IMAGE_TAG=<YOUR_CUSTOM_TAG_HERE>`
-4. Build & push dev images: `make docker-push-dev`.
+```(bash)
+BAZEL_ARGS="--define=MANAGER_IMAGE_TAG=<YOUR_TAG_HERE> make docker-push-dev
+```
 
 #### Building and pushing dev images to custom (non GCR) container registry
 
 1. Login to your container registry using `docker login`.
 
    E.g. `docker login quay.io`
-2. To build images with custom tags, edit [.bazelrc](./.bazelrc) as follows:
+2. To build images with custom tags and push to your custom image registry,
+   run the `make docker-build-dev` as follows::
 
-   `build --define=MANAGER_IMAGE_TAG=<YOUR_CUSTOM_TAG_HERE>`
-3. Build dev images: `REGISTRY_DEV=<YOUR_CONTAINER_REGISTRY_HERE> make clean docker-build-dev`
-4. Push your docker images as `docker push <ContainerImage>:<YourTag>`
+```(bash)
+BAZEL_ARGS="--define=MANAGER_IMAGE_TAG=<YOUR_TAG_HERE> --define=REGISTRY_DEV=<YOUR_REGISTRY_HERE>" make docker-build-dev
+```
+
+3. Push your docker images as `docker push <ContainerImage>:<YourTag>`
 
 #### Running clusterctl
 
--- a/docs/development.md
+++ b/docs/development.md
@@ -5,6 +5,7 @@
 <!-- Below is generated using VSCode yzhang.markdown-all-in-one >
 
 <!-- TOC depthFrom:2 -->
+
 - [Setting up](#setting-up)
   - [Base requirements](#base-requirements)
   - [Get the source](#get-the-source)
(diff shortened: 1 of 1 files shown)


docs/development.md
```diff
@@ -99,7 +103,21 @@ The dev version of the manifests can be generated with
 
 1. [Install the gcloud cli][gcloud_sdk].
 2. Set project: `gcloud config set project YOUR_PROJECT_NAME`.
-3. Build & push dev images: `make docker-push-dev`.
+3. To build images with custom tags, edit [.bazelrc](./.bazelrc) as follows:
+
+   `build --define=MANAGER_IMAGE_TAG=<YOUR_CUSTOM_TAG_HERE>`
```
detiber: Is it not possible to set this through normal env vars?

docs/development.md
```diff
@@ -99,7 +103,21 @@ The dev version of the manifests can be generated with
...
+4. Build & push dev images: `make docker-push-dev`.
+
+#### Building and pushing dev images to custom (non GCR) container registry
+
+1. Login to your container registry using `docker login`.
+
+   E.g. `docker login quay.io`
+2. To build images with custom tags, edit [.bazelrc](./.bazelrc) as follows:
+
+   `build --define=MANAGER_IMAGE_TAG=<YOUR_CUSTOM_TAG_HERE>`
```
detiber: Same comment as above

docs/development.md
chuckha: the docker-* family of make targets will all work with `make docker-* BAZEL_ARGS="--define=MANAGER_IMAGE_TAG=my-manager-test-tag"`

docs/development.md
```diff
@@ -99,7 +103,21 @@ The dev version of the manifests can be generated with
 
 1. [Install the gcloud cli][gcloud_sdk].
 2. Set project: `gcloud config set project YOUR_PROJECT_NAME`.
-3. Build & push dev images: `make docker-push-dev`.
+3. To build images with custom tags, edit [.bazelrc](./.bazelrc) as follows:
```
chuckha: don't modify `.bazelrc`, use make with environment variables. If a build target is missing the ability to customize the Bazel command please file an issue

docs/development.md
```diff
@@ -99,7 +103,21 @@ The dev version of the manifests can be generated with
...
+#### Building and pushing dev images to custom (non GCR) container registry
+
+1. Login to your container registry using `docker login`.
+
+   E.g. `docker login quay.io`
+2. To build images with custom tags, edit [.bazelrc](./.bazelrc) as follows:
+
+   `build --define=MANAGER_IMAGE_TAG=<YOUR_CUSTOM_TAG_HERE>`
+3. Build dev images: `REGISTRY_DEV=<YOUR_CONTAINER_REGISTRY_HERE> make clean docker-build-dev`
+4. Push your docker images as `docker push <ContainerImage>:<YourTag>`
```
chuckha: this is all achievable with `make docker-push-dev BAZEL_ARGS="--define=MANAGER_IMAGE_TAG=my-manager-test-tag --define=REGISTRY_DEV=registry.example.com"`

Although looking through the bazel file, I'd think `REGISTRY_DEV` and `REGISTRY_STABLE` should both just be `REGISTRY`

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cmd/manager/BUILD.bazel#L111
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cmd/manager/BUILD.bazel#L119

But that's out of scope for this PR

docs/development.md
vincepri: Should we prefix optional steps with `(Optional)` or something similar to communicate these aren't required for the workflow that follows the preset conventions?

docs/development.md
ashish-amarnath: yep. You are right. I did this by editing the `bazelrc` file for my local dev. Will update the doc to take these values on the cli

docs/development.md
ashish-amarnath: I was able to pass the custom tag and the registry on the command line. Will document that approach.

docs/development.md
ashish-amarnath: yeah. Let's clean that up in a different PR.

docs/development.md
ashish-amarnath: I think we should always recommend folks to use custom tags so that they don't accidentally end up pushing wrong tags resulting in a prod deployment. Ideally, you would want to use the image SHA in all your deployments. But that's a different topic :)

docs/development.md
ashish-amarnath: Fixed.

docs/development.md
ashish-amarnath: fixed

docs/development.md
ashish-amarnath: fixed

docs/development.md
ashish-amarnath: I think keeping `REGISTRY_DEV` and `REGISTRY_STABLE` makes following the build easier. There isn't much to gain by collapsing the two into one. WDYT?

ashish-amarnath: Making the change to collapse the two variables, I kinda confused myself. 🤣

PR #644, opened 2019-03-11
fix: no longer panic when instance id is nil

yozamacs:
This PR fixes a bug where if the machine resource was deleted before the ec2 instance is created, the provider panics. The alternative to this is removing the finalizer by hand and then deleting them.

Fixes #

**Release note**:

```release-note
Provider no longer panics when instance id is nil
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine/actuator.go
+++ b/pkg/cloud/aws/actuators/machine/actuator.go
@@ -241,6 +241,11 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	ec2svc := ec2.NewService(scope.Scope)
 
+	if scope.MachineStatus.InstanceID == nil {
+		klog.Error("Instance does not have an instance id")
+		return nil
+	}
+
 	instance, err := ec2svc.InstanceIfExists(*scope.MachineStatus.InstanceID)
 	if err != nil {
 		return errors.Errorf("failed to get instance: %+v", err)
--- a/pkg/cloud/aws/actuators/machine/actuator.go
+++ b/pkg/cloud/aws/actuators/machine/actuator.go
@@ -241,12 +241,7 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	ec2svc := ec2.NewService(scope.Scope)
 
-	if scope.MachineStatus.InstanceID == nil {
-		klog.Error("Instance does not have an instance id")
-		return nil
-	}
-
-	instance, err := ec2svc.InstanceIfExists(*scope.MachineStatus.InstanceID)
+	instance, err := ec2svc.InstanceIfExists(scope.MachineStatus.InstanceID)
 	if err != nil {
 		return errors.Errorf("failed to get instance: %+v", err)
 	}
@@ -342,7 +337,7 @@ func (a *Actuator) Update(ctx context.Context, cluster *clusterv1.Cluster, machi
 	ec2svc := ec2.NewService(scope.Scope)
 
 	// Get the current instance description from AWS.
-	instanceDescription, err := ec2svc.InstanceIfExists(*scope.MachineStatus.InstanceID)
+	instanceDescription, err := ec2svc.InstanceIfExists(scope.MachineStatus.InstanceID)
 	if err != nil {
 		return errors.Errorf("failed to get instance: %+v", err)
 	}
@@ -393,7 +388,7 @@ func (a *Actuator) Exists(ctx context.Context, cluster *clusterv1.Cluster, machi
 		return false, nil
 	}
 
-	instance, err := ec2svc.InstanceIfExists(*scope.MachineStatus.InstanceID)
+	instance, err := ec2svc.InstanceIfExists(scope.MachineStatus.InstanceID)
 	if err != nil {
 		return false, errors.Errorf("failed to retrieve instance: %+v", err)
 	}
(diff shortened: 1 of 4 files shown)


yozamacs:
github connected to cncf

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -241,6 +241,11 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	ec2svc := ec2.NewService(scope.Scope)
 
+	if scope.MachineStatus.InstanceID == nil {
```
ashish-amarnath: I think it'll be better to perform this check in the `InstanceIfExists` method https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/aws/services/ec2/instances.go#L72.
WDYT?

chuckha: why would it be better in the `InstanceIfExists` method? I'm not saying I disagree, but I want to know your thinking behind it

ashish-amarnath: Adding the null check in the `InstanceIfExists` method will benefit all callers of that method instead of every caller having to do this check.

yozamacs: you can't check the instance id inside of `InstanceIfExists` because there is a panic if the instance id doesn't exist

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -241,6 +241,11 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+		klog.Error("Instance does not have an instance id")
+		return nil
```
ashish-amarnath: What does it mean when `instanceId` is nil?
If `instanceId` is nil, then that means that the machine object hasn't been reconciled yet or is in the process of being reconciled by another thread. So there could be a race. For that reason, I think we shouldn't be returning nil (no error) here. By returning an error, this delete will be re-queued and will be processed once the reconciliation is complete.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: I see. @chuckha What were you thinking?

pkg/cloud/aws/actuators/machine/actuator.go
yozamacs: This is trying to capture the case where the `instanceID` is nil because the ec2 instance never existed (wrong iam role or any other issue). The retries would happen over and over again but with no change. In that case, it seems that the reconciliation process wouldn't make any difference.

pkg/cloud/aws/actuators/machine/actuator.go
detiber: Why not update `InstanceIfExists` to take a pointer? Then you could just pass `scope.MachineStatus.InstanceID` directly and handle the null check?

pkg/cloud/aws/actuators/machine/actuator.go
detiber: There is another use case that I hit as well where the instance exists but the Status has not been fully populated yet (pivoting the cluster resources). Handling this in InstanceIfExists would allow for the filtered lookup by labels to resolve both issues.

pkg/cloud/aws/actuators/machine/actuator.go
yozamacs: I'll go ahead and update the function signature. I was trying to avoid that but it's probably the cleanest solution

chuckha: @ashish-amarnath I agree with you, but I wanted to hear your logic

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -69,11 +69,16 @@ func (s *Service) InstanceByTags(machine *actuators.MachineScope) (*v1alpha1.Ins
 }
 
 // InstanceIfExists returns the existing instance or nothing if it doesn't exist.
-func (s *Service) InstanceIfExists(id string) (*v1alpha1.Instance, error) {
-	klog.V(2).Infof("Looking for instance %q", id)
+func (s *Service) InstanceIfExists(id *string) (*v1alpha1.Instance, error) {
+	if id == nil {
+		klog.Error("Instance does not have an instance id")
+		return nil, nil
```
ashish-amarnath: I think we want to return an error here, not nil.

yozamacs: if we return an error then the function will continue to loop and wait for the instance id to exists....but if it never exists then the machine object won't get deleted

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: @yozamacs if we were to return `nil` here, what would happen if the machine is in the process of reconciliation?

pkg/cloud/aws/services/ec2/instances.go
ashish-amarnath: Sorry to fork the conversation here. I responded https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/644#discussion_r264913317

pkg/cloud/aws/actuators/machine/actuator.go
yozamacs: I'm not sure. Where does the reconciliation happen? I'm not very familiar with that process

ashish-amarnath: Answered here https://kubernetes.slack.com/archives/CD6U2V71N/p1552434137136400?thread_ts=1552423306.129400&cid=CD6U2V71N.
returning nil error should be fine :)

ashish-amarnath:
/approved
/assign @detiber 
for lgtm

PR #651, opened 2019-03-13
Allow user to use arbitraty userdata

enxebre:

This PR removes tight coupling between machine resource and kubeadm while keeping the current behaviour pristine.
This allows flexibility for the API consumers to support broader use cases by injecting arbitrary userdata into a secret optionally referenced by the spec.
While being a fairly scoped change this would favour adoption of the project extremely.

Related:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/135
https://github.com/kubernetes-sigs/cluster-api/issues/721 and https://docs.google.com/document/d/1pzXtwYWRsOzq5Ftu03O5FcFAlQE26nD3bjYBPenbhjg/edit#heading=h.etxqa0k5waej

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -158,15 +158,64 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 	}
 
 	// apply values based on the role of the machine
+	userData, err := s.getUserData(machine, bootstrapToken, caCertHash, kubeConfig)
+	if err != nil {
+		return nil, awserrors.NewFailedDependency(
+			errors.Errorf("failed to get userData for machine %q", machine.Name()),
+		)
+	}
+	input.UserData = userData
+
+	securityGroupID, err := s.getSecurityGroupID(machine)
+	if err != nil {
+		return nil, awserrors.NewFailedDependency(
+			errors.Errorf("failed to get securityGroup ID for machine %q", machine.Name()),
+		)
+	}
+	input.SecurityGroupIDs = append(input.SecurityGroupIDs, securityGroupID)
+
+	// Pick SSH key, if any.
+	if machine.MachineConfig.KeyName != "" {
+		input.KeyName = aws.String(machine.MachineConfig.KeyName)
+	} else {
+		input.KeyName = aws.String(defaultSSHKeyName)
+	}
+
+	out, err := s.runInstance(machine.Role(), input)
+	if err != nil {
+		return nil, err
+	}
+
+	record.Eventf(machine.Machine, "CreatedInstance", "Created new %s instance with id %q", machine.Role(), out.ID)
+	return out, nil
+}
+
+func (s *Service) getSecurityGroupID(machine *actuators.MachineScope) (string, error) {
 	switch machine.Role() {
 	case "controlplane":
 		if s.scope.SecurityGroups()[v1alpha1.SecurityGroupControlPlane] == nil {
-			return nil, awserrors.NewFailedDependency(
+			return "", awserrors.NewFailedDependency(
 				errors.New("failed to run controlplane, security group not available"),
 			)
 		}
+		return s.scope.SecurityGroups()[v1alpha1.SecurityGroupControlPlane].ID, nil
+	case "node":
+		return s.scope.SecurityGroups()[v1alpha1.SecurityGroupNode].ID, nil
+	default:
+		return "", errors.Errorf("Unknown node role %q", machine.Role())
+	}
+	return "", nil
+}
 
-		var userData string
+func (s *Service) getUserData(machine *actuators.MachineScope, bootstrapToken, caCertHash string, kubeConfig string) (string, error) {
+	var userData string
+	switch machine.Role() {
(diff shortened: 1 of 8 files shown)


enxebre:
cc @vincepri @derekwaynecarr @justaugustus

chuckha:
where is the branch on your fork to check this out? I can't seem to locate a branch named `allow-user-userdata`

enxebre:
@chuckha https://github.com/enxebre/cluster-api-provider-aws-2/tree/allow-user-userdata
See also https://github.com/openshift/cluster-api-provider-aws/blob/master/pkg/actuators/machine/actuator.go#L203-L214

pkg/cloud/aws/services/userdata/userdata.go
```diff
@@ -90,3 +93,21 @@ func funcMap(funcs map[string]interface{}) template.FuncMap {
 
 	return funcMap
 }
+
+func GetUserDataFromSecret(machine *actuators.MachineScope) (string, error) {
```
frobware: Is it possible to create a unit test case for this function?

chuckha: A couple of suggestions and questions:
* flip around the nil check and return early if the UserDataSecret is nil
* the `userData` var is unnecessary, directly encode and return at line 105 and remove the else statement
* why is it an error if the secret doesn't contain the `userData` key? My expectation would be to log an error and return nothing
* please add godoc and a test to this function

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -158,15 +158,73 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 	}
 
 	// apply values based on the role of the machine
```
ashish-amarnath: is this comment still valid?

pkg/cloud/aws/services/userdata/userdata.go
```diff
@@ -90,3 +93,21 @@ func funcMap(funcs map[string]interface{}) template.FuncMap {
...
+	if machine.MachineConfig.UserDataSecret != nil {
+		userData := []byte{}
+		userDataSecret, err := machine.CoreClient.Secrets(machine.Namespace()).Get(machine.MachineConfig.UserDataSecret.Name, metav1.GetOptions{})
```
ashish-amarnath: How will this secret be created?
What will happen if the secret doesn't exist?

enxebre:
> Before I get too far into code reviewing let me make sure I understand what this is doing:
> 
> This allows someone to add their own script to be run at node-boot time. It's not in addition to the default installation, but entirely replaces it.
> 
> Can you outline what your use case is so I can better understand where you're coming from?

Hey @chuckha for our use case in OpenShift we have a clearer separation between machine (hardware) and software/kubelet. The software/kubelet is managed orthogonally.
We pre-populate a secret with an [ignition](https://coreos.com/ignition/docs/latest/) config which then is fetched via the userdata API field https://github.com/openshift/cluster-api-provider-aws/blob/master/pkg/actuators/machine/actuator.go#L203-L214. 
Then the kubelet and software upgrades/lifecycle management is oblivious to this API

pkg/cloud/aws/services/userdata/userdata.go
enxebre: Correct. If you set https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/651/files#diff-7fbfb526ca988417add16e1058fcd340R85
You are assuming the secret with that name to exist.
If you don't set the spec field, current behaviour remains
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/651/files#diff-e7748766dc5475de766bbb0d40ef7fd1R98
Y

ncdc:
As discussed in today's Cluster API meeting, we'd like to see a document outlining requirements and use cases. Going to mark this on hold for now. Thanks!

/hold

enxebre:
> As discussed in today's Cluster API meeting, we'd like to see a document outlining requirements and use cases. Going to mark this on hold for now. Thanks!
Unfortunately attending the meeting is not always feasible being in CET. @ncdc I can't see any action item in the meeting notes https://docs.google.com/document/d/1Ys-DOR5UsgbMEeciuG0HOgDQc8kZsaWIWJeKJ1-UfbY/edit#heading=h.h3b651pv3rlt I'm happy to get some doc in place otherwise.

This relates to https://github.com/kubernetes-sigs/cluster-api/issues/721
The case this PR is adding is allowing the user to use arbitrary userdata in machines

ncdc:
@enxebre understood about the time zone issue. The relevant section in the meeting notes is `Ignition support - node bootstrap commonalities between all providers?`, specifically the part about doing a design doc.

ncdc:
We would like to go through the design process to come up with something that works well for everyone, and would like adequate time to review/comment/revise etc.

ncdc:
As part of the Cluster API v1alpha2 milestone that we just started working on, we're separating bootstrapping from infrastructure provisioning, and you'll be able to bring your own user data. As such, I'm going to close this in favor of that effort.

/close

PR #654, opened 2019-03-13
Fix issues with `clusterctl delete`

detiber:
Fixes issues with `clusterctl delete`
- Try to lookup machine instance using tags on delete
- Do not fail if VPC ID does not exist
- Update cluster-api dependency

**Release note**:
```release-note
NONE
```


pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -247,9 +247,15 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
...
-		// The machine hasn't been created yet
-		klog.Info("Instance is nil and therefore does not exist")
-		return nil
+		instance, err = ec2svc.InstanceByTags(scope)
+		if err != nil {
+			return errors.Errorf("failed to query instance by tags: %+v", err)
+		}
+		if instance == nil {
+			// The machine hasn't been created yet
+			klog.Info("Instance is nil and therefore does not exist")
```
chuckha: This should probably be behind a `.V(3)` or higher -- I'm trying to imagine a situation when I'm reading logs and I see this line pop up. I think it either needs more context or should not be in the Info log

pkg/cloud/aws/services/ec2/vpc.go
```diff
@@ -130,10 +130,11 @@ func (s *Service) deleteVPC() error {
 	_, err := s.scope.EC2.DeleteVpc(input)
 	if err != nil {
 		// Ignore if it's already deleted
-		if code, ok := awserrors.Code(err); code != "InvalidVpcID.NotFound" && ok {
-			return errors.Wrapf(err, "failed to delete vpc %q", vpc.ID)
+		if code, ok := awserrors.Code(err); code == "InvalidVpcID.NotFound" && ok {
```
chuckha: Nice, +1 on this change. the flow is much easier to read

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -247,9 +247,15 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 
 	if instance == nil {
-		// The machine hasn't been created yet
-		klog.Info("Instance is nil and therefore does not exist")
-		return nil
+		instance, err = ec2svc.InstanceByTags(scope)
+		if err != nil {
+			return errors.Errorf("failed to query instance by tags: %+v", err)
+		}
+		if instance == nil {
```
vincepri: This could be `else if`

pkg/cloud/aws/actuators/machine/actuator.go
vincepri: +1

detiber:
@chuckha @vincepri updated

Issue #661, opened 2019-03-14
Do not use relative tags (:latest) in release YAML


chuckha:
/kind bug
/priority critical-urgent

**What steps did you take and what happened:**
Release YAML should not have `:latest` hardcoded in the image tag. A release should always reference absolute tags (yes, I know docker tags that are not the SHA of the container image are mutable).

detiber:
I think this will be less of an issue once we are vendoring a tag from cluster-api rather than vendoring a point in time version of the master branch.

That said, we should probably cut interim releases of cluster-api to avoid this issue all together.

chuckha:
This isn't really an issue as pointed by Jason

PR #662, opened 2019-03-14
A little make & bazel cleaning

chuckha:
Benefits include

* Removes Make's direct dependency on `gcloud` for production builds
* Reduce the number of unique variables (e.g. REGISTRY_STABLE vs REGISTRY_DEV)
* Reduce the number of bazel targets to be more reusable
* Removes generated provider-components file from repo
* Removes extra bazel and make targets

These are some small clean ups that will make fixing #661 easier.

Fixes #

**Release note**:

```release-note
NONE
```


Makefile
```diff
@@ -59,19 +57,19 @@ test: generate verify ## Run tests
 
 .PHONY: docker-build
 docker-build: generate ## Build the production docker image
-	bazel run //cmd/manager:manager-image $(BAZEL_DOCKER_ARGS)
+	bazel run //cmd/manager:manager-image $(BAZEL_ARGS)
 
 .PHONY: docker-build-dev
 docker-build-dev: generate ## Build the development docker image
-	bazel run //cmd/manager:manager-image-dev $(BAZEL_DOCKER_ARGS)
+	bazel run //cmd/manager:manager-image $(BAZEL_ARGS) --define=REGISTRY=gcr.io/$(shell gcloud config get-value project)
```
detiber: We should probably also expose this out as an env var for make, otherwise overriding it requires new contributors to know how to format BAZEL_ARGS.

Makefile
```diff
@@ -108,13 +106,13 @@ manifests-dev: cmd/clusterctl/examples/aws/provider-components-base-dev.yaml ##
...
 cmd/clusterctl/examples/aws/provider-components-base.yaml:
-	bazel build //cmd/clusterctl/examples/aws:provider-components-base $(BAZEL_DOCKER_ARGS)
+	bazel build //cmd/clusterctl/examples/aws:provider-components-base $(BAZEL_ARGS)
 	install bazel-genfiles/cmd/clusterctl/examples/aws/provider-components-base.yaml cmd/clusterctl/examples/aws
 
 .PHONY: cmd/clusterctl/examples/aws/provider-components-base-dev.yaml
 cmd/clusterctl/examples/aws/provider-components-base-dev.yaml:
-	bazel build //cmd/clusterctl/examples/aws:provider-components-base-dev $(BAZEL_DOCKER_ARGS)
-	install bazel-genfiles/cmd/clusterctl/examples/aws/provider-components-base-dev.yaml cmd/clusterctl/examples/aws
+	bazel build //cmd/clusterctl/examples/aws:provider-components-base $(BAZEL_ARGS) --define=PULL_POLICY=Always --define=REGISTRY=gcr.io/$(shell gcloud config get-value project)
```
detiber: Are we overriding anything other than PULL_POLICY and REGISTRY here? Any reason this needs to be a separate target from provider-components-base.yaml?

detiber: In addition to exposing REGISTRY and IMAGE_PULL_POLICY as env vars from make, it might be helpful to plumb through the version override as well.

chuckha: I wanted to keep the Make commands the same.

Correct me if I'm wrong, but what you're suggesting is:

```
target:
  bazel build thing --define=PULL_POLICY=$PULL_POLICY
```

to combine the -dev target and this target.

TBH I agree here and would prefer development builds to be default. That would mean the development workflow would be

```
make docker-push
```

and the release workflow would be

```
make docker-push PULL_POLICY=IfNotPreset REGISTRY=gcr.io/cluster-api-provider-aws
```

Does that sound reasonable?

chuckha: Yep, I'm going to hold off on adding features though, I want to fix the workflow first. However, I absolutely agree with you and you'll see several more PRs for bazel over the course of the day where I will address exactly this

.bazelrc
```diff
@@ -1,6 +1,8 @@
 build --define=MANAGER_IMAGE_NAME=cluster-api-aws-controller
 build --define=MANAGER_IMAGE_TAG=0.0.4
-build --define=REGISTRY_STABLE=gcr.io/cluster-api-provider-aws
```
vincepri: 😍

Makefile
vincepri: +1 to using a variable instead. I think there are some folks using a different registry.

Makefile
```diff
@@ -59,19 +57,19 @@ test: generate verify ## Run tests
...
 
 .PHONY: docker-push
 docker-push: ## Push production docker image
-	bazel run //cmd/manager:manager-push $(BAZEL_DOCKER_ARGS)
+	bazel run //cmd/manager:manager-push $(BAZEL_ARGS)
 
 .PHONY: docker-push-dev
 docker-push-dev: ## Push development image
-	bazel run //cmd/manager:manager-push-dev  $(BAZEL_DOCKER_ARGS)
+	bazel run //cmd/manager:manager-push $(BAZEL_ARGS) --define=REGISTRY=gcr.io/$(shell gcloud config get-value project)
```
vincepri: Same as above

Makefile
detiber: +1 to making the dev workflow the default and overriding (in a documented way) for the release workflow :)

.gitignore
```diff
@@ -47,4 +47,4 @@ kubeconfig
 
 # Ignore output manifests
 cmd/clusterctl/examples/aws/out
-cmd/clusterctl/examples/aws/provider-components-base-dev.yaml
+cmd/clusterctl/examples/aws/provider-components-base.yaml
```
chuckha: both -dev and this one are generated files. Neither of which should have been checked in

detiber:
lgtm, leaving final lgtm to @vincepri

PR #580, opened 2019-02-14
Release artifacts MVP

randomvariable:
**What this PR does / why we need it**:
Refactors the way manifests are generated so that provider components are generated by Bazel in line with the image.

`make release-artifacts` will generate a tar file with the manifest templates, shell script and getting started guide.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #578 
Fixes #569, 
Related to #549, #276, #122

**Special notes for your reviewer**:

E2E tests were rewired to make use of the artifact creation here, so that it tests the manifests generated by the tools we provide to the user. During that, made some changes to provide more logging information during the run. Can confirm no secrets are leaked by doing this.

**Release note**:

```release-note
Sample manifests are now provided as a tar file with each release.
```

PR #665, opened 2019-03-14
Remove more bazel defaults

chuckha:

This moves all the build variables into the make file so a release can happen entirely through make without having to modify bazel definitions

Aids in the next release

**Release note**:
```release-note
NONE
```

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,6 +1,3 @@
-build --define=MANAGER_IMAGE_NAME=cluster-api-aws-controller
-build --define=MANAGER_IMAGE_TAG=0.0.4
-
 build --workspace_status_command=./hack/print-workspace-status.sh
 build --verbose_failures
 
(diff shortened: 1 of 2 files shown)


randomvariable:
#580 and friends were working towards a world where make would mostly go away. This and #606 goes back the other way. 

There's also an issue here about consuming `cluster-api-provider-aws` as a repo downstream using Bazel as the information is no longer present to get a set of manifests with the correct images without copy-pasting chunks of the Makefile.

Would prefer us to be explicit about long-term intent here.

vincepri:
I agree with @randomvariable that we should discuss the long term use of Make and Bazel. I personally would like to use Bazel as a complement to Go tooling and Make should be the glue around it only for development purposes.

chuckha:
@randomvariable and I talked for a bit, we're going to plan out a longer term build environment post v1alpha1.

PR #667, opened 2019-03-14
Release tool

chuckha:

Adds some technical debt to the releaser

```release-note
None
```

### DIFF
--- a/cmd/release/main.go
+++ b/cmd/release/main.go
@@ -38,6 +38,11 @@ github token for modifying release
 const (
 	// TODO(chuckha): figure this out based on directory name
 	repository = "cluster-api-provider-aws"
+
+	// TODO move these into config
+	registry        = "gcr.io"
+	managerImageTag = "0.0.5"
+	pullPolicy      = "IfNotPresent"
 )
 
 func main() {
@@ -73,6 +78,12 @@ func main() {
 	}
 
 	run := &runner{
+		builder: makebuilder{
+			registry:   registry,
+			imageName:  repository,
+			imageTag:   managerImageTag,
+			pullPolicy: pullPolicy,
+		},
 		releaser: gothubReleaser{
 			artifactsDir: cfg.artifactDir,
 			user:         user,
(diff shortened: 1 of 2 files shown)


cmd/release/main.go
```diff
@@ -179,17 +190,30 @@ type config struct {
...
 
+// TODO sha512 the artifacts!
+
 func (r runner) run() error {
 	fmt.Print("tagging repository ")
 	if err := r.tagger.tag(r.config.version); err != nil {
 		return err
 	}
 	fmt.Println("🐲")
+	fmt.Print("checking out tag ")
```
vincepri: Should we output the tag name?

cmd/release/main.go
```diff
@@ -179,17 +190,30 @@ type config struct {
...
+	if err := r.tagger.checkout(r.config.version); err != nil {
+		return err
+	}
+	fmt.Println("🐲")
+	fmt.Print("building artifacts ")
```
vincepri: Same as above, might be useful to know which artifacts

cmd/release/main.go
```diff
@@ -179,17 +190,30 @@ type config struct {
...
+	if err := r.builder.build(); err != nil {
+		return err
+	}
+	fmt.Println("🐲")
 	fmt.Print("pushing tag ")
```
vincepri: Same as above

cmd/release/main.go
```diff
@@ -255,12 +280,20 @@ type git struct {
 }
 
 func (g git) tag(version string) error {
-	cmd := exec.Command("git", "tag", "-s", "-m", fmt.Sprintf("A release of %q for version %q", g.repository, version), version)
-	out, err := cmd.CombinedOutput()
+	// if the tag exists then return nil; if it doesn't creat it
```
chuckha:
```suggestion
	// if the tag exists then return nil; if it doesn't create it
```

cmd/release/main.go
chuckha: it will iterate through the uploaded files later in the program...but this is a wrapper for `make release artifacts`.

would the output of the wrapped command be sufficient?

cmd/release/main.go
```diff
@@ -179,23 +190,36 @@ type config struct {
...
 		return err
 	}
 	fmt.Println("🐲")
-	fmt.Print("pushing tag ")
+	fmt.Printf("checking out tag %q ", r.config.version)
+	if err := r.tagger.checkout(r.config.version); err != nil {
+		return err
+	}
+	fmt.Println("🐲")
+	fmt.Printf("building artifacts %v ", r.config.artifacts)
```
chuckha: @vincepri this will print the list of artifacts that it expects to build. There still more to do here like checking for created files but I think that's out of scope for this PR

PR #668, opened 2019-03-15
clean up release tool a tiny bit

chuckha:

This PR cleans up the release tool to make it slightly more reusable in some ways and less reusable in others.

There will be some tricky part with regards to determining exactly what is in a release for each provider and how to build it. This may need an interface, but it's a bit early to define that right now

```release-note
NONE
```

### DIFF
--- a/cmd/release/main.go
+++ b/cmd/release/main.go
@@ -40,9 +40,10 @@ const (
 	repository = "cluster-api-provider-aws"
 
 	// TODO move these into config
-	registry        = "gcr.io"
-	managerImageTag = "0.0.5"
-	pullPolicy      = "IfNotPresent"
+	registry         = "gcr.io"
+	managerImageTag  = "0.0.5"
+	managerImageName = "cluster-api-aws-controller"
+	pullPolicy       = "IfNotPresent"
 )
 
 func main() {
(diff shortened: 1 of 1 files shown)


cmd/release/main.go
```diff
@@ -288,7 +320,7 @@ func (g git) tag(version string) error {
 		// assume this means it doesn't exist
 		cmd = exec.Command("git", "tag", "-s", "-m", fmt.Sprintf("A release of %q for version %q", g.repository, version), version)
 		out, err2 := cmd.CombinedOutput()
-		if err != nil {
+		if err2 != nil {
 			fmt.Println(string(out))
```
detiber: any reason not to use:
```go
if out, err := cmd.CombinedOutput(); err != nil {
  fmt.Println(string(out))
  return err
}
```

cmd/release/main.go
```diff
@@ -288,7 +320,7 @@ func (g git) tag(version string) error {
...
 		}
 		return err2
```
detiber: Shouldn't this return `err`, since `err2` here is not guaranteed to be non-nil?

chuckha: no, the problem here is go1.11 doesn't let you get exit codes. The assumption here is that if `err` is not nil, then it's erring because the tag exists which is a known case and we can discard the error.

Therefore err2 is the only interesting error at this point.

This will look better when we migrate to go 1.12 that allows me to check exit codes cleanly

cmd/release/main.go
chuckha: No, it's just a style thing. My reason for writing it this way is that I'd be returning `err` regardless if it's a nil or not but I only want to print when the `err2` is not nil

cmd/release/main.go
detiber: Gotcha, I think it might make sense to be more explicit about it then, otherwise it might cause others confusion down the line as well.

chuckha: there is a comment *just* out of context of what the diff shows -- does that comment look like it explains what I've described above or do you think I need more words?

PR #670, opened 2019-03-18
Update NodeRef

### LINKED ISSUE #214: Prevent deletion of machine and cluster hosting cluster-api's api-server
/kind bug

Currently it is possible to delete the Machine Object (and transitively the Instance backing that machine) that is hosting the Control Plane for the cluster that is running the Cluster API CRDs/Controllers. If this happens, then all other Machine Instances and other corresponding AWS resources attached to the Cluster Object are orphaned.

There is some checking within the upstream Cluster API code to avoid deleting a machine in this case, but it is tied to the link between the Node Object and the Machine Object, which I do not believe we are currently handling properly in our provider implementation.

/priority important-soon

detiber:
Previously we were only setting the "Name" field for the NodeRef, this updates it to also set UID, Kind, and APIVersion fields.

Fixes #214

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine/actuator.go
+++ b/pkg/cloud/aws/actuators/machine/actuator.go
@@ -459,9 +459,10 @@ func (a *Actuator) getNodeReference(scope *actuators.MachineScope) (*apicorev1.O
 			// TODO(vincepri): Improve this comparison without relying on substrings.
 			if strings.Contains(node.Spec.ProviderID, instanceID) {
 				return &apicorev1.ObjectReference{
-					Kind:       node.Kind,
-					APIVersion: node.APIVersion,
+					Kind:       "Node",
+					APIVersion: apicorev1.SchemeGroupVersion.String(),
 					Name:       node.Name,
+					UID:        node.UID,
 				}, nil
 			}
 		}


detiber:
/assign @ashish-amarnath 

I'm currently testing this change, but it should resolve the issue you were seeing with `isDeleteAllowed`

detiber:
Just finished testing, and this does indeed block deletion of the machine hosting the Cluster API components now.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -459,9 +459,10 @@ func (a *Actuator) getNodeReference(scope *actuators.MachineScope) (*apicorev1.O
 			// TODO(vincepri): Improve this comparison without relying on substrings.
 			if strings.Contains(node.Spec.ProviderID, instanceID) {
 				return &apicorev1.ObjectReference{
-					Kind:       node.Kind,
-					APIVersion: node.APIVersion,
+					Kind:       "Node",
```
vincepri: Isn't node.Kind `Node`? What about the APIVersion change?

detiber: I thought so as well, but if you look at the output of @ashish-amarnath's testing you will see that previously Kind and APIVersion were unset: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/214#issuecomment-473615702

PR #671, opened 2019-03-18
Update the resync interval to 10m instead of 10h

detiber:
Overrides the default resync interval (10h) to 10m

Related to: https://github.com/kubernetes-sigs/cluster-api/issues/748

**Release note**:
```release-note
The resync interval for reconciling Machines and Clusters is now 10m instead of the Kubebuilder default of 10h
```

### DIFF
--- a/cmd/manager/main.go
+++ b/cmd/manager/main.go
@@ -18,6 +18,7 @@ package main
 
 import (
 	"flag"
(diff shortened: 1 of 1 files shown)


vincepri:
Should we follow-up with a PR upstream as well?

PR #681, opened 2019-03-27
Add Roles for ControlPlane and Nodes

tahsinrahman:
Add roles to nodes
```
NAME                         STATUS   ROLES    AGE    VERSION
ip-10-0-0-113.ec2.internal   Ready    master   113m   v1.13.2
ip-10-0-0-138.ec2.internal   Ready    node     11m    v1.13.2
```

If we don't add this label, roles column becomes `<none>`
```
NAME                         STATUS   ROLES      AGE    VERSION
ip-10-0-0-138.ec2.internal   Ready    <none>     11m    v1.13.2
```

### DIFF
--- a/pkg/cloud/aws/services/kubeadm/aws_defaults.go
+++ b/pkg/cloud/aws/services/kubeadm/aws_defaults.go
@@ -39,6 +39,9 @@ const (
 	apiServerBindPort = 6443
 
 	cloudProvider = "aws"
+
+	controlPlaneRole = "node-role.kubernetes.io/master="
+	nodeRole         = "node-role.kubernetes.io/node="
 )
 
 // SetDefaultClusterConfiguration sets default dynamic values without overriding
@@ -126,6 +129,7 @@ func SetInitConfigurationOverrides(base *kubeadmv1beta1.InitConfiguration) {
 		klog.Infof("Overriding node's cloud-provider to the required value of %q.", cloudProvider)
 	}
 	base.NodeRegistration.KubeletExtraArgs["cloud-provider"] = cloudProvider
+	base.NodeRegistration.KubeletExtraArgs["node-labels"] = controlPlaneRole
 }
 
 // SetJoinNodeConfigurationOverrides overrides user input for certain fields of
@@ -162,6 +166,11 @@ func SetJoinNodeConfigurationOverrides(caCertHash, bootstrapToken string, machin
 		klog.Infof("Overriding node's cloud-provider to the required value of %q.", cloudProvider)
 	}
 	base.NodeRegistration.KubeletExtraArgs["cloud-provider"] = cloudProvider
+	if machine.Machine.Labels["set"] == "node" {
+		base.NodeRegistration.KubeletExtraArgs["node-labels"] = nodeRole
+	} else {
+		base.NodeRegistration.KubeletExtraArgs["node-labels"] = controlPlaneRole
+	}
 }
 
 // SetControlPlaneJoinConfigurationOverrides user input for kubeadm join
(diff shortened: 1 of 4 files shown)


pkg/cloud/aws/services/kubeadm/aws_defaults.go
```diff
@@ -162,6 +166,11 @@ func SetJoinNodeConfigurationOverrides(caCertHash, bootstrapToken string, machin
 		klog.Infof("Overriding node's cloud-provider to the required value of %q.", cloudProvider)
 	}
 	base.NodeRegistration.KubeletExtraArgs["cloud-provider"] = cloudProvider
+	if machine.Machine.Labels["set"] == "node" {
```
detiber: Instead of using labels here, this should probably filter on the existence or absence of `machine.spec.versions.controlPlane`.

pkg/cloud/aws/services/kubeadm/aws_defaults.go
```diff
@@ -162,6 +166,11 @@ func SetJoinNodeConfigurationOverrides(caCertHash, bootstrapToken string, machin
...
+		base.NodeRegistration.KubeletExtraArgs["node-labels"] = nodeRole
+	} else {
+		base.NodeRegistration.KubeletExtraArgs["node-labels"] = controlPlaneRole
```
detiber: Is setting controlPlaneRole necessary here? `kubeadm` should handle this for us.

pkg/cloud/aws/services/kubeadm/aws_defaults.go
```diff
@@ -126,6 +129,7 @@ func SetInitConfigurationOverrides(base *kubeadmv1beta1.InitConfiguration) {
 		klog.Infof("Overriding node's cloud-provider to the required value of %q.", cloudProvider)
 	}
 	base.NodeRegistration.KubeletExtraArgs["cloud-provider"] = cloudProvider
+	base.NodeRegistration.KubeletExtraArgs["node-labels"] = controlPlaneRole
```
detiber: Any reason to override this here and not let `kubeadm` handle this?

pkg/cloud/aws/services/kubeadm/aws_defaults.go
```diff
@@ -162,6 +165,9 @@ func SetJoinNodeConfigurationOverrides(caCertHash, bootstrapToken string, machin
 		klog.Infof("Overriding node's cloud-provider to the required value of %q.", cloudProvider)
 	}
 	base.NodeRegistration.KubeletExtraArgs["cloud-provider"] = cloudProvider
+	if util.IsControlPlaneMachine(machine.Machine) {
```
detiber: It looks like the logic here is reversed, shouldn't this apply the nodeRole if it is not a control plane node?

```suggestion
	if !util.IsControlPlaneMachine(machine.Machine) {
```

tahsinrahman:
@detiber i've updated it, can you check if it is okay?

detiber:
@tahsinrahman Did you push the latest changes up? It looks like my comment about the `util.IsControlPlaneMachine` logic still applies

vincepri:
@tahsinrahman Seems we should run `make generate clientset` to update Bazel files

vincepri:
One comment is that I see we're only applying the role for nodes, is the api server already applied by kubeadm?

detiber:
> One comment is that I see we're only applying the role for nodes, is the api server already applied by kubeadm?

Yes, I also think we should probably follow up with `kubeadm` to add support for setting the node role by default as well.

PR #684, opened 2019-03-30
fix: apply security groups by ENI rather than instance

### LINKED ISSUE #638: Additional security groups not applied to control plane
/kind bug

Created a control plane machine with AdditionalSecurityGroups as follows:

```yaml
apiVersion: "cluster.k8s.io/v1alpha1"
kind: Machine
metadata:
  name: controlplane-0
  namespace: my-cluster
  labels:
    cluster.k8s.io/cluster-name: my-cluster
    set: controlplane
spec:
  versions:
    kubelet: v1.13.3
    controlPlane: v1.13.3
  providerSpec:
    value:
      additionalSecurityGroups:
        - id: sg-1234567
      apiVersion: awsprovider/v1alpha1
      kind: AWSMachineProviderSpec
      instanceType: "t2.medium"
      iamInstanceProfile: "my-profile"
      keyName: "cluster-api-provider-aws.sigs.k8s.io"
```

Saw in logs that control plane reconciliation was looping after:
```
I0308 17:28:05.970023       1 instances.go:407] Attempting to update security groups on instance "i-03a829e5f06e4504f"
```

I added an additional log statement:
```go
if _, err := s.scope.EC2.ModifyInstanceAttribute(input); err != nil {
        // here
	klog.Errorf("failed to modify security groups: %v", err)
	return errors.Wrapf(err, "failed to modify instance %q security groups", instanceID)
}
```

```
E0308 17:28:06.126219       1 instances.go:415] failed to modify security groups: InvalidInstanceID: There are multiple interfaces attached to instance 'i-03a829e5f06e4504f'. Please specify an interface ID for the operation instead.
```

Looks like we need some additional handling, because IIRC, we would need ModifyNetworkInterfaceAttribute once the instance has multiple network interfaces?

Additional security groups applied to control plane

**Environment:**

- Cluster-api-provider-aws version: current master
- Kubernetes version: (use `kubectl version`): 1.13.3
- OS (e.g. from `/etc/os-release`): kind on macOS 10.14.3

rudoi:

Additional security groups are now applied via ec2.ModifyNetworkInterfaceAttribute rather than ec2.ModifyInstanceMetadata in order to support instances with multiple ENIs.

Fixes #638 

**Release note**:

```release-note
Support adding security groups to instances with multiple network interfaces
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -411,13 +411,22 @@ func (s *Service) runInstance(role string, i *v1alpha1.Instance) (*v1alpha1.Inst
 func (s *Service) UpdateInstanceSecurityGroups(instanceID string, ids []string) error {
 	klog.V(2).Infof("Attempting to update security groups on instance %q", instanceID)
 
-	input := &ec2.ModifyInstanceAttributeInput{
-		InstanceId: aws.String(instanceID),
-		Groups:     aws.StringSlice(ids),
+	enis, err := s.getInstanceENIs(instanceID)
+	if err != nil {
+		return errors.Wrapf(err, "failed to get ENIs for instance %q", instanceID)
 	}
 
-	if _, err := s.scope.EC2.ModifyInstanceAttribute(input); err != nil {
-		return errors.Wrapf(err, "failed to modify instance %q security groups", instanceID)
+	klog.V(3).Infof("Found %v ENIs on instance %q", len(enis), instanceID)
+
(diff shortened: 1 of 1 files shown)


vincepri:
@detiber thoughts on putting this into v1alpha1?

detiber:
I don't have any objections to this change as a short term solution, however longer term we'll probably want to consider the possibility of not wanting to apply the same set of security groups to all attached ENIs.

/lgtm

detiber:
@vincepri no objections for me to including in v1alpha1

PR #686, opened 2019-04-01
Update cluster-api to v0.1.0

detiber:
Updates cluster-api to v0.1.0


detiber:
Tested and verified that generated manifests use the correct image for cluster-api as well (using bazel 0.24.0).

This PR needs verification that `make manifests` still works properly under bazel 0.23.x before removing the hold.

detiber:
bazel 0.23.x has been verified as working
/hold cancel

detiber:
/hold cancel
unused and conflicting bazel config removed.

Issue #608, opened 2019-02-25
Restrict IAM policies further to only allow deletion (and possibly update) of resources


detiber:
/kind feature

**Describe the solution you'd like**
While we currently have our IAM policy scoped down with respect to Operations, we can further scope down the policy through the use of Conditions. In particular, we should scope down the Delete operations only to instances that are tagged with the tags we create. We should also investigate if we can further scope down other operations as well.

/priority important-soon
/milestone Next

randomvariable:
Have opened https://github.com/kubernetes/cloud-provider-aws/issues/33 as the cloud provider itself needs better IAM policies to make this truly effective in some scenarios.

randomvariable:
Loose consensus (n=2) is that for Cloud Provider AWS, will scope requests down to objects with the `k8s.io/cluster` tag.

Cloud Provider AWS does not currently set `k8s.io/cluster-provider-aws/managed=true`

randomvariable:
/remove-lifecycle

I've done some initial feasibility investigations in #612, which should be used as a basis for further work, but will want to have e2e for scoping these down.

ncdc:
Actually going to clear the milestone so we re-triage
/milestone clear

randomvariable:
Interest has dropped off on this one.

/close

PR #697, opened 2019-04-02
Fix: all traffic ingress rule triggers fatal nil dereference

sethp-nr:

This change does two things:

1. Handles ingress rules that don't have a "port" notion. Counterintuitively, ICMP and ICMPv6 traffic _does_ have a "port" notion in AWS-land.
2. Fix a bug related to deleting ingress rules without a description, which AWS considers distinct from those with an empty description

These patches fix the crasher, but the story isn't yet fully told. We found ourselves in this situation after spinning up a Service of `Type: LoadBalancer`, which caused the cloud provider to helpfully add a rule to the nodes' security group for "all traffic" coming from the ELB.

This change will enable the management tooling to handle that new rule, decide it's not spec'd, and clean it up properly. At which point, I expect, the cloud provider will put it back. We're still investigating what to do about that tug of war, but I wanted to open the PR for collaboration.

**Release note**:

```release-note
prevent fatal crash when handling "all traffic" rules
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/securitygroups.go
+++ b/pkg/cloud/aws/services/ec2/securitygroups.go
@@ -31,6 +31,20 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/awserrors"
 )
 
+const (
+	// IpProtocolTCP is how EC2 represents the TCP protocol in ingress rules
+	IpProtocolTCP = "tcp"
+
+	// IpProtocolUDP is how EC2 represents the UDP protocol in ingress rules
+	IpProtocolUDP = "udp"
+
+	// IpProtocolICMP is how EC2 represents the ICMP protocol in ingress rules
+	IpProtocolICMP = "icmp"
+
+	// IpProtocolICMPv6 is how EC2 represents the ICMPv6 protocol in ingress rules
+	IpProtocolICMPv6 = "58"
+)
+
 func (s *Service) reconcileSecurityGroups() error {
 	klog.V(2).Infof("Reconciling security groups")
 
@@ -333,8 +347,7 @@ func (s *Service) getSecurityGroupTagParams(name string, role v1alpha1.SecurityG
 	}
 }
 
-func ingressRuleToSDKType(i *v1alpha1.IngressRule) *ec2.IpPermission {
-	var res *ec2.IpPermission
+func ingressRuleToSDKType(i *v1alpha1.IngressRule) (res *ec2.IpPermission) {
 	switch i.Protocol {
 	case v1alpha1.SecurityGroupProtocolTCP,
 		v1alpha1.SecurityGroupProtocolUDP,
@@ -375,13 +388,19 @@ func ingressRuleToSDKType(i *v1alpha1.IngressRule) *ec2.IpPermission {
 		res.UserIdGroupPairs = append(res.UserIdGroupPairs, userIDGroupPair)
 	}
 
-	return res
+	return
 }
 
-func ingressRuleFromSDKType(v *ec2.IpPermission) *v1alpha1.IngressRule {
-	var res *v1alpha1.IngressRule
+func ingressRuleFromSDKType(v *ec2.IpPermission) (res *v1alpha1.IngressRule) {
+	// Ports are only well-defined for TCP and UDP protocols, but EC2 overloads the port range
+	// in the case of ICMP(v6) traffic to indicate which codes are allowed. For all other protocols,
+	// including the custom "-1" All Traffic protcol, FromPort and ToPort are omitted from the response.
+	// See: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IpPermission.html
 	switch *v.IpProtocol {
-	case "tcp", "udp", "icmp", "58" /* ICMPv6 */ :
+	case IpProtocolTCP,
+		IpProtocolUDP,
+		IpProtocolICMP,
+		IpProtocolICMPv6:
 		res = &v1alpha1.IngressRule{
 			Protocol: v1alpha1.SecurityGroupProtocol(*v.IpProtocol),
 			FromPort: *v.FromPort,
@@ -413,5 +432,5 @@ func ingressRuleFromSDKType(v *ec2.IpPermission) *v1alpha1.IngressRule {
 		res.SourceSecurityGroupIDs = append(res.SourceSecurityGroupIDs, *pair.GroupId)
 	}
 
-	return res
+	return
 }
--- a/pkg/cloud/aws/services/ec2/securitygroups.go
+++ b/pkg/cloud/aws/services/ec2/securitygroups.go
@@ -348,6 +348,9 @@ func (s *Service) getSecurityGroupTagParams(name string, role v1alpha1.SecurityG
 }
 
 func ingressRuleToSDKType(i *v1alpha1.IngressRule) (res *ec2.IpPermission) {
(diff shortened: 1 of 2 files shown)


pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -334,34 +334,63 @@ func (s *Service) getSecurityGroupTagParams(name string, role v1alpha1.SecurityG
 }
 
 func ingressRuleToSDKType(i *v1alpha1.IngressRule) *ec2.IpPermission {
-	res := &ec2.IpPermission{
-		IpProtocol: aws.String(string(i.Protocol)),
-		FromPort:   aws.Int64(i.FromPort),
-		ToPort:     aws.Int64(i.ToPort),
+	var res *ec2.IpPermission
+	switch i.Protocol {
+	case v1alpha1.SecurityGroupProtocolTCP,
```
vincepri: Is there some documentation that we can link to from here?

pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -334,34 +334,63 @@ func (s *Service) getSecurityGroupTagParams(name string, role v1alpha1.SecurityG
 }
 
 func ingressRuleToSDKType(i *v1alpha1.IngressRule) *ec2.IpPermission {
-	res := &ec2.IpPermission{
-		IpProtocol: aws.String(string(i.Protocol)),
-		FromPort:   aws.Int64(i.FromPort),
-		ToPort:     aws.Int64(i.ToPort),
+	var res *ec2.IpPermission
```
vincepri: This could be declared as named return

pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -334,34 +334,63 @@ func (s *Service) getSecurityGroupTagParams(name string, role v1alpha1.SecurityG
...
 }
 
 func ingressRuleFromSDKType(v *ec2.IpPermission) *v1alpha1.IngressRule {
-	res := &v1alpha1.IngressRule{
-		Protocol: v1alpha1.SecurityGroupProtocol(*v.IpProtocol),
-		FromPort: *v.FromPort,
-		ToPort:   *v.ToPort,
+	var res *v1alpha1.IngressRule
+	switch *v.IpProtocol {
+	case "tcp", "udp", "icmp", "58" /* ICMPv6 */ :
```
vincepri: Should we move all of these to be package constants with descriptions?

pkg/cloud/aws/services/ec2/securitygroups.go
sethp-nr: I'll see if I can dig some up – the current link I have goes to a multi-megabyte HTML page that makes all my fans spin up to look at.

pkg/cloud/aws/services/ec2/securitygroups.go
sethp-nr: You don't like the `"58"` hanging out there, apropos of nothing? 😆 

I'll move 'em on in

vincepri: That for sure! 😂 Main reasoning was that these variables are probably going to be reused in other places too so it might be worthwhile to declare them somewhere else.

sethp-nr:
I've been ruminating on what to do about the trouble with `type: LoadBalancer` that caused this crasher, and this fix would break. So far my ideas fall into one of two camps:

1. Be more permissive with ingress rules. In my mind, that breaks down into either: going "add-only" and not worrying about removing rules (outside of deletion), or trying to identify the LoadBalancer rules and either slurp them back into the Spec or just ignore them entirely.
2. Separate out the security group(s) owned by cluster-api and owned by the cluster itself: remove the "owned" tag from the security group spun up by cluster-api. The tricky bit here becomes the cloud provider config: either we need to create (but not update) a security group for the rules to be added to, or maybe we pre-configure something like: https://medium.com/@marciorag/kubernetes-and-aws-elb-what-to-do-when-you-reach-the-security-group-limit-in-aws-45207e423e

What do you think?

sethp-nr:
I'd also like to know your preference on trying to fix that issue in this PR vs. fixing the crasher and opening a new issue/PR for the LoadBalancer thing.

randomvariable:
> ICMP requires a port range (though I'm mystified what that could map to), and for ICMPv6 "ports" are apparently overloaded to restrict certain types of messages?

For both ICMP and ICMPv6, from/to port will actually correspond to the ICMP types allowed. 

> I've been ruminating on what to do about the trouble with type: LoadBalancer that caused this crasher, and this fix would break.

Hmm, given AWS still hasn't updated the [limit of 5 security groups per network interface](https://docs.aws.amazon.com/vpc/latest/userguide/amazon-vpc-limits.html) (raised to 10 by request), I guess we need to coalesce our rules into the smallest number of SGs possible, as to maximise the number of possible LB attachments.

So, 1 security group per node, regardless of type, and the merger of control plane & worker node rules for the control plane instance.

This only helps us in the case that the cloud provider doesn't try to inject rules into an SG managed by Cluster API. This relates to https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/608#issuecomment-467900867 where it was found that the provider doesn't set `k8s.io/cluster-provider-aws/managed=true`, and only checks for the cluster tag. Cloud provider can't tell the difference between resources Cluster API has provisioned, and which belong to it.

sethp-nr:
>  I guess we need to coalesce our rules into the smallest number of SGs possible, as to maximise the number of possible LB attachments.

Yeah, that 5 limit is a lot lower than I'd like. I guess my question is whether "smallest number" of SGs in this case is one or two.

> Cloud provider can't tell the difference between resources Cluster API has provisioned, and which belong to it.

Cluster-api tells the cloud provider it owns the resource: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/aws/tags/tags.go#L94

It seems like either CAPA should allow the provider-injected rules to exist in the shared group, or it should create a separate group for the cloud provider to inject its rules into. I'm presently leaning toward the latter for its simplicity, but that does mean the carrying cost of a CAPA cluster is 2/5 security group slots.

randomvariable:
> It seems like either CAPA should allow the provider-injected rules to exist in the shared group, or it should create a separate group for the cloud provider to inject its rules into. > I'm presently leaning toward the latter for its simplicity, but that does mean the carrying cost of a CAPA cluster is 2/5 security group slots.

Don't think that was ever our intent, though maybe @detiber can clarify. We're tagging that cluster-api-provider-aws is managing the resource (`sigs.k8s.io/cluster-api-provider-aws/managed=true`, and we are also tagging everything that is managed by the provider with the cluster ID (`kubernetes.io/cluster/blah=owned`, because we use it for reconciliation.
Unfortunately, the cloud provider only checks for cluster ID `kubernetes.io/cluster/blah=owned`.

Either we make the AWS cloud provider aware of tooling that uses `kubernetes.io/cluster/blah=owned` but it is not managing, or we may ultimately have to not use `kubernetes.io/cluster/blah=owned` if we expect further clashes with the cloud provider.

pkg/cloud/aws/services/ec2/securitygroups.go
randomvariable: @vincepri would https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_IpPermission.html be sufficient?

sethp-nr: whoops – that's what I put in below, I'll add it here

randomvariable:
lgtm from me

PR #699, opened 2019-04-03
Temporarily add a fix to use account state as "clean" if 'free' fails…

harishspqr:
It fixes the AWS E2E tests running in test-infra. 

```release-note
NONE
```

### DIFF
--- a/hack/checkout_account.py
+++ b/hack/checkout_account.py
@@ -27,18 +27,26 @@ BOSKOS_HOST=os.environ.get("BOSKOS_HOST", "boskos")
 RESOURCE_TYPE = "aws-account"
 USER = "cluster-api-provider-aws"
 
-if __name__ == "__main__":
-    conn = httplib.HTTPConnection(BOSKOS_HOST)
-
+def post_request(conn, input_state = "free"):
     conn.request("POST", "/acquire?%s" % urllib.urlencode({
         'type': RESOURCE_TYPE,
         'owner': USER,
-        'state': 'free',
+        'state': input_state,
         'dest': 'busy',
     })
 
     )
-    resp = conn.getresponse()
+    return conn.getresponse()
+
+if __name__ == "__main__":
+    conn = httplib.HTTPConnection(BOSKOS_HOST)
+
+    resp = post_request(conn)
+    #  we're working around an issue with the data in boskos.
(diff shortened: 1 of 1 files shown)


hack/checkout_account.py
```diff
@@ -27,24 +27,29 @@
...
     if resp.status != 200:
         sys.exit("Got invalid response %d: %s" % (resp.status, resp.reason))
 
     body = json.load(resp)
     conn.close()
     print 'export BOSKOS_RESOURCE_NAME="%s";' % body['name']
     print 'export AWS_ACCESS_KEY_ID="%s";' % body['userdata']['access-key-id']
-    print 'export AWS_SECRET_ACCESS_KEY="%s";' % body['userdata']['secret-access-key']
-
+    print 'export AWS_SECRET_ACCESS_KEY="%s";' % body['userdata']['secret-access-key']
```
ncdc: Does your editor remove newlines at the end of the file?

hack/checkout_account.py
```diff
@@ -27,24 +27,29 @@
 RESOURCE_TYPE = "aws-account"
 USER = "cluster-api-provider-aws"
 
-if __name__ == "__main__":
-    conn = httplib.HTTPConnection(BOSKOS_HOST)
-
+def post_request(input_state = "free"):
```
ncdc: I'd add a comment that we're working around an issue with the data in boskos and that we'll remove this code that tries both free and clean once all the data is good.

hack/checkout_account.py
```diff
@@ -27,24 +27,29 @@
...
     })
 
     )
-    resp = conn.getresponse()
+    return conn.getresponse()
+
+if __name__ == "__main__":
+    conn = httplib.HTTPConnection(BOSKOS_HOST)
+
+    resp = post_request()
```
ncdc: You'll need to pass `conn` in, won't you?

detiber:
We should probably hold off on this to see if @BenTheElder is able to fix the state for us.

hack/checkout_account.py
harishspqr: Not sure. Will add back.

hack/checkout_account.py
harishspqr: absolutely!

PR #700, opened 2019-04-04
Update checkout_account.py to not reuse connections

detiber:
Currently the automated e2e tests are having issues related reusing the httplib connection: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-cluster-api-provider-aws-bazel-e2e/1113802682641944577, this udpates checkout_account.py to not reuse connections.

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/checkout_account.py
+++ b/hack/checkout_account.py
@@ -27,7 +27,8 @@ BOSKOS_HOST=os.environ.get("BOSKOS_HOST", "boskos")
 RESOURCE_TYPE = "aws-account"
 USER = "cluster-api-provider-aws"
 
-def post_request(conn, input_state = "free"):
+def post_request(host, input_state = "clean"):
+    conn = httplib.HTTPConnection(host)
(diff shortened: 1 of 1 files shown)


hack/checkout_account.py
```diff
@@ -36,22 +37,21 @@ def post_request(conn, input_state = "free"):
     })
 
     )
-    return conn.getresponse()
+    resp = conn.getresponse()
+    conn.close()
```
ncdc: I'm fairly certain that if you close the connection before reading the response (done below in `json.load(resp)`), that you won't get any of the data.

detiber: arrgh, of course it would behave that way :joy: Let me fix up to return the json.

detiber:
@ncdc it isn't pretty, but it should work

ncdc:
/lgtm
fingers crossed

PR #701, opened 2019-04-04
Make sure Calico can talk IP-in-IP

### LINKED ISSUE #554: Validate networking rules and configuration
/kind feature

As a pre-req for cross-AZ support, we need to validate the security group ruleset to include IP in IP for cross-subnet traffic.

Additionally, Calico is deployed here with the default MTU of 1440 bytes, but we should bump this to [8981  bytes](https://docs.projectcalico.org/v3.4/usage/configuration/mtu) which is the max MTU for IP in IP inside standard jumbo ethernet frames, which is supported by AWS.

**Environment:**

sfzylad:
Allow Calico to talk IP-in-IP protocol

Fixes #554 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/aws/services/ec2/securitygroups.go
+++ b/pkg/cloud/aws/services/ec2/securitygroups.go
@@ -289,6 +289,13 @@ func (s *Service) getSecurityGroupIngressRules(role v1alpha1.SecurityGroupRole) 
 					s.scope.SecurityGroups()[v1alpha1.SecurityGroupNode].ID,
 				},
 			},
+			{
+				Description: "IP-in-IP (calico)",
+				Protocol:    v1alpha1.SecurityGroupProtocolIPinIP,
+				FromPort:    1,
+				ToPort:      65535,
+				CidrBlocks:  []string{anyIPv4CidrBlock},
+			},
 		}, nil
 
 	case v1alpha1.SecurityGroupNode:
--- a/pkg/cloud/aws/services/ec2/securitygroups.go
+++ b/pkg/cloud/aws/services/ec2/securitygroups.go
@@ -43,9 +43,6 @@ const (
 
 	// IPProtocolICMPv6 is how EC2 represents the ICMPv6 protocol in ingress rules
 	IPProtocolICMPv6 = "58"
-
-	// IPProtocolIPv4 is how EC2 represents IP in IP protocol
-	IPProtocolIPv4 = "4"
 )
 
 func (s *Service) reconcileSecurityGroups() error {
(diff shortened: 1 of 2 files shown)


sethp-nr:
This looks like it overlaps a bit with #697 in handling the From/To port gunk. I guess we're racing to see who gets the conflict? :)

Also, I think you'll want to add this rule to the ControlPlane security group as well.

sfzylad:
Let's see who'll be the first to get into master. :)

I only added it to the node SG because now it's getting attached to the control plane nodes too, so I thought it's redundant. However I'm happy to add it.

vincepri:
It might be worthwhile to touch base on both PR at the next community meeting, wdyt?

sfzylad:
@vincepri Great idea. Let's discuss it Monday. :)

sethp-nr:
That's Monday the 8th, right? That'd work for me.

pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -304,6 +304,13 @@ func (s *Service) getSecurityGroupIngressRules(role v1alpha1.SecurityGroupRole)
 					s.scope.SecurityGroups()[v1alpha1.SecurityGroupNode].ID,
 				},
 			},
+			{
+				Description: "IP-in-IP (calico)",
+				Protocol:    v1alpha1.SecurityGroupProtocolIPinIP,
+				FromPort:    1,
```
randomvariable: This should be `-1` to mean everything, if at all possible.

pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -304,6 +304,13 @@ func (s *Service) getSecurityGroupIngressRules(role v1alpha1.SecurityGroupRole)
...
+				ToPort:      65535,
```
randomvariable: Same as above

pkg/cloud/aws/services/ec2/securitygroups.go
sethp-nr: Wouldn't the AWS APIs just ignore these values?

randomvariable: It will, so it's a nit to stay in line with AWS docs rather than something concrete.

pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -304,6 +304,13 @@ func (s *Service) getSecurityGroupIngressRules(role v1alpha1.SecurityGroupRole)
...
+				CidrBlocks:  []string{anyIPv4CidrBlock},
```
randomvariable: Not a nit this time:
We shouldn't use `0.0.0.0/0` here, but instead reference the security group ID using the `SourceSecurityGroupIDs` parameter.

PR #706, opened 2019-04-05
refactor: separate CAPA resources from cluster

### LINKED ISSUE #704: Is Cluster API Provider AWS clashing with cloud provider code?
/kind bug

Drawing https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/697#issuecomment-480075174 into it's own issue:

We're tagging that cluster-api-provider-aws is managing the resource (`sigs.k8s.io/cluster-api-provider-aws/managed=true`, and we are also tagging everything that is managed by the provider with the cluster ID (`kubernetes.io/cluster/blah=owned`, because we use it for reconciliation.

Unfortunately, the AWS cloud provider only checks for cluster ID `kubernetes.io/cluster/blah=owned`, and ends up mutating resources being managed by this provider.

> Give therefore to the emperor the things that are the emperor’s, and to God the things that are God’s

Substituting the AWS Cloud Provider for the emperor, and Cluster API for God.

**Environment:**

sethp-nr:
This patch tries to separate the ownership concerns between CAPA and the
in-cluster cloud integration.

Flagging it as a WIP for now, it seems to work for me but I'd like to discuss the approach in more detail at next week's meeting.

As it stands, the in-cluster AWS integration tries to add security group rules when it reconciles a load balancer that CAPA then removes.

Fixes #704

This is fairly dependent on #697 or #701, else CAPA will crash trying to read the rules created by the LB integration. We could decouple them by adding a check before reading the rules from an "owned" security group, but I'm expecting one of those will land before this PR anyway.

**Release note**:

```release-note
BREAKING CHANGE: improve handling of resources shared with the in-cluster AWS integration. See docs/upgrade-to-0.3.0.md for migration details.
```


docs/proposal/aws-resource-handling.md
```diff
@@ -38,7 +38,11 @@ Where possible use [client tokens](https://docs.aws.amazon.com/AWSEC2/latest/API
 
 ## Tagging of resources
 
-Resources that are managed by the controllers/actuators should be tagged with: `kubernetes.io/cluster/<name or id>=owned` and `sigs.k8s.io/cluster-api-provider-aws=true`. The latter tag being used to differentiate from resources managed by other tools/components that make use of the common tag.
+Resources handled by these components fall into one of three categories:
+
+1. Fully-managed resources whose lifecycle is tied to the cluster. These resources should be tagged with `sigs.k8s.io/cluster-api-provider/<name or id>=owned` and `sigs.k8s.io/cluster-api-provider-aws/managed=true` and the actuator is expected to keep these resources as closely in sync with the spec as possible.
```
randomvariable: Typo? `sigs.k8s.io/cluster-api-provider/`
should be `sigs.k8s.io/cluster-api-provider-aws/` throughout I think?

randomvariable:
Happy with changing the tag over unless there's a good reason not to.

Technically, should we add a breaking change release note as existing resources will be missing tags?

docs/proposal/aws-resource-handling.md
```diff
@@ -38,7 +38,11 @@ Where possible use [client tokens](https://docs.aws.amazon.com/AWSEC2/latest/API
...
+2. Resources whose management is shared with the in-cluster aws cloud provider,, such as a security group for load balancer ingress rules. These resources should be tagged with `sigs.k8s.io/cluster-api-provider/<name or id>=owned` and `kubernetes.io/cluster/<name or id>=owned`, with the latter being the tag defined by the cloud provider. These resources are create/delete only: that is to say their ongoing management is "handed off" to the cloud provider.
+3. Unmanaged resources that are provided by config (such as a common VPC). These resources should be tagged with neither `sigs.k8s.io/cluster-api-provider-aws/managed=true` nor `sigs.k8s.io/cluster-api-provider/<name or id>=owned`. It is expected that the provider will avoid changing these resources as much as is possible.
```
detiber: I think we could probably support `sigs.k8s.io/cluster-api/<name or id>=shared` to provide an analog to the cloud-provider tags.

docs/proposal/aws-resource-handling.md
```diff
@@ -38,7 +38,11 @@ Where possible use [client tokens](https://docs.aws.amazon.com/AWSEC2/latest/API
 
 ## Tagging of resources
 
-Resources that are managed by the controllers/actuators should be tagged with: `kubernetes.io/cluster/<name or id>=owned` and `sigs.k8s.io/cluster-api-provider-aws=true`. The latter tag being used to differentiate from resources managed by other tools/components that make use of the common tag.
+Resources handled by these components fall into one of three categories:
+
+1. Fully-managed resources whose lifecycle is tied to the cluster. These resources should be tagged with `sigs.k8s.io/cluster-api-provider/<name or id>=owned` and `sigs.k8s.io/cluster-api-provider-aws/managed=true` and the actuator is expected to keep these resources as closely in sync with the spec as possible.
+2. Resources whose management is shared with the in-cluster aws cloud provider,, such as a security group for load balancer ingress rules. These resources should be tagged with `sigs.k8s.io/cluster-api-provider/<name or id>=owned` and `kubernetes.io/cluster/<name or id>=owned`, with the latter being the tag defined by the cloud provider. These resources are create/delete only: that is to say their ongoing management is "handed off" to the cloud provider.
```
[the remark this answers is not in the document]
sethp-nr: We've found two:

1. Persistent volume support via EBS (requires tagging the instance)
2. Security group rules for ELBs (requires tagging the security group)

This PR covers those cases by adding the "cluster-owned" tag back, but I expect there will be more cases that this change misses.

docs/proposal/aws-resource-handling.md
sethp-nr: I think that makes sense, though I'm not quite sure what the semantics would be. Currently this case covers user-supplied infrastructure (e.g. we have a VPC that's been around since the dawn of the account that we're deploying into). Whose responsibility would it be to keep the `shared` tag in sync?

detiber: We could adopt the semantics of the cloud-provider here. The shared ownership tag should indicate that it is a cluster resource, but ownership is external. Users (or higher level tooling interacting with cluster-api) would be responsible for managing this tag.

docs/proposal/aws-resource-handling.md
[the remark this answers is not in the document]
detiber: Ah, gotcha. In that case, would it make more sense to apply `kubernetes.io/cluster/<name or id>=shared` instead? Since the cloud provider would not actually own the resource? It should still be able to query and make use of the resource.

docs/proposal/aws-resource-handling.md
sethp-nr: I have not done a great job with consistency. The doc changes say one thing, and the code another 😄In code, I went with `sigs.k8s.io/cluster-api-provider-aws/cluster/<id>`, as a riff on the existing prefix: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/aws/tags/types.go#L87

Currently, that's used to define both `sigs.k8s.io/cluster-api-provider-aws/managed` (which this PR would supersede) and `sigs.k8s.io/cluster-api-provider-aws/role`, so adding a `/cluster/` felt fairly natural.

That said, I'm happy to adjust the tagging scheme to be `sigs.k8s.io/cluster-api/{role,cluster}` if that makes more sense to y'all.

detiber: I think cluster-api-provider-aws is fine. It might make sense to standardize more as we refine the v1alpha2 and beyond milestones for cluster-api proper.

pkg/apis/awsprovider/v1alpha1/types.go
```diff
@@ -237,6 +239,9 @@ var (
 
 	// SecurityGroupControlPlane defines a Kubernetes control plane node role
 	SecurityGroupControlPlane = SecurityGroupRole("controlplane")
+
+	// SecurityGroupLB defines a container for the cloud provider to inject its load balnacer ingress rules
```
detiber:
```suggestion
	// SecurityGroupLB defines a container for the cloud provider to inject its load balancer ingress rules
```

sethp-nr:
Oh, thanks for bringing up the backwards-compatibility question! I meant to raise that during today's meeting, but it slipped my mind. 

I had to manually clean up some resources to get the code in its existing state to function properly because CAPA wouldn't "see" them, try to re-create them, and get a conflict error.

Technically, it's simple enough to read both sets of tags and write just the new set, but I deferred work on that because:

a) Testing that flow was a bit of a pain, and I wasn't even sure how many CAPA-managed clusters existed in the world that would be affected by the change
b) I didn't know how or when to drop support for the legacy scheme. v1alpha2?

docs/proposal/aws-resource-handling.md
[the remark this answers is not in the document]
sethp-nr: Hmm, maybe? The comment in the cloud provider gives me pause: https://github.com/kubernetes/kubernetes/blob/d7103187a37dcfff79077c80a151e98571487628/pkg/cloudprovider/providers/aws/tags.go#L45-L52

>    // ResourceLifecycleOwned is the value we use when tagging resources to indicate
>    // that the resource is considered owned and managed by the cluster,
>    // and in particular that the lifecycle is tied to the lifecycle of the cluster.
> ...
>	// ResourceLifecycleShared is the value we use when tagging resources to indicate
>	// that the resource is shared between multiple clusters, and should not be destroyed
>	// if the cluster is destroyed.

And at least as far as I can see, the provider never reads the value of the tag (it only checks for its existence: https://github.com/kubernetes/kubernetes/blob/d7103187a37dcfff79077c80a151e98571487628/pkg/cloudprovider/providers/aws/tags.go#L133 ).

Maybe the thing to do is tag them with a new value, like `provided`, and PR the cloud provider to adopt those semantics (since it'd just be a `const` and a comment at this point)?

docs/proposal/aws-resource-handling.md
sethp-nr: The semantics of the cloud provider are a lot fuzzier than I'd like. It also plays it fairly fast and loose: e.g. if you have no properly tagged subnets it'll fall back to using the current one, and if you've got exactly one security group (tagged or not), that's the one it'll pick.

I'm also feeling a bit of tension around the workflow that a per-cluster shared tag introduces: maybe I've gotten spoiled, but I'm kind of used to just throwing new clusters into my VPC whenever I want 'em 😆I'd feel better about that if we picked a more static tag for shared infra, something like `sigs.k8s.io/cluster-api-provider-aws/shared`?

pkg/cloud/aws/services/ec2/bastion.go
```diff
@@ -38,7 +38,7 @@ const (
 
 // ReconcileBastion ensures a bastion is created for the cluster
 func (s *Service) ReconcileBastion() error {
-	if s.scope.VPC().IsProvided() {
+	if s.scope.VPC().IsProvided(s.scope.Name()) {
```
vincepri: Is there any way to avoid the repetition here?

pkg/apis/awsprovider/v1alpha1/types.go
```diff
@@ -21,6 +21,8 @@ import (
 	"sort"
 	"time"
 
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/tags"
+
```
vincepri: Space should be removed

pkg/cloud/aws/filter/types.go
```diff
@@ -15,29 +15,3 @@ limitations under the License.
...
-
-	// TagValueAPIServerRole describes the value for the apiserver role
-	TagValueAPIServerRole = "apiserver"
-
-	// TagValueBastionRole describes the value for the bastion role
-	TagValueBastionRole = "bastion"
-
-	// TagValueCommonRole describes the value for the common role
-	TagValueCommonRole = "common"
-)
```
vincepri: +1

pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -85,6 +86,10 @@ func (s *Service) reconcileSecurityGroups() error {
 	// Second iteration creates or updates all permissions on the security group to match
 	// the specified ingress rules.
 	for role, sg := range s.scope.SecurityGroups() {
+		if sg.Tags.HasAWSCloudProviderOwned(s.scope.Name()) {
+			// skip rule reconciliation
```
vincepri: Can we add a little more details on why are we skipping in the comment?

pkg/cloud/aws/services/ec2/securitygroups.go
```diff
@@ -305,6 +310,9 @@ func (s *Service) getSecurityGroupIngressRules(role v1alpha1.SecurityGroupRole)
 				},
 			},
 		}, nil
+	case v1alpha1.SecurityGroupLB:
+		// We ignore these
```
vincepri: Same as above

pkg/cloud/aws/tags/cluster.go
```diff
@@ -20,5 +20,10 @@ import "fmt"
 
 // ClusterKey generates the key for resources associated with a cluster.
 func ClusterKey(name string) string {
-	return fmt.Sprintf("%s%s", NameKubernetesClusterPrefix, name)
+	return fmt.Sprintf("%s%s", NameAWSProviderOwned, name)
+}
+
+// ClusterAWSCloudProviderKey generates the key for resources associated a cluster's AWS cloud provider
```
vincepri:
```suggestion
// ClusterAWSCloudProviderKey generates the key for resources associated a cluster's AWS cloud provider.
```
Nit

pkg/cloud/aws/services/ec2/bastion.go
sethp-nr: What did you have in mind? This happened because the "do I own this?" question now implicitly depends on the name of the cluster, and the `VPCSpec` doesn't know its own cluster name.

pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
```diff
@@ -112,8 +112,8 @@ func (v *VPCSpec) String() string {
 }
 
 // IsProvided returns true if the VPC is unmanaged.
-func (v *VPCSpec) IsProvided() bool {
-	return v.ID != "" && !v.Tags.HasManaged()
+func (v *VPCSpec) IsProvided(clusterName string) bool {
```
ashish-amarnath:
```suggestion
func (v *VPCSpec) IsUnmanaged(clusterName string) bool {
```

pkg/apis/awsprovider/v1alpha1/types.go
```diff
@@ -237,6 +238,9 @@ var (
 
 	// SecurityGroupControlPlane defines a Kubernetes control plane node role
 	SecurityGroupControlPlane = SecurityGroupRole("controlplane")
+
+	// SecurityGroupLB defines a container for the cloud provider to inject its load balancer ingress rules
+	SecurityGroupLB = SecurityGroupRole("lb")
```
ashish-amarnath: suggest creating package constants for `lb`, `controlplane`, `node` and `bastion`. I'll happy to take that as a follow-up.

sethp-nr: I started down this path, but I think it's large enough and self-contained enough to make more sense as a separate PR

sethp-nr:
Here's the steps I followed to upgrade a 0.2.0 cluster:

1. `kubectl scale statefulset -n aws-provider-system aws-provider-controller-manager --replicas=0`
2. `AWS_PROFILE=my-profile clusterawsadm migrate -n cluster-name 0.3.0`
3. Update the image for the aws-provider-controller-manager to point to something with this fix
4. `kubectl scale statefulset -n aws-provider-system aws-provider-controller-manager --replicas=1`
5. Wait ~2 minutes for the security group changes to all settle
6. `kubectl exec -n kube-system -it kube-controller-manager-ip-10-11-0-233.ec2.internal -- sh -c 'kill 1'` as a workaround for kubernetes/kubernetes#77019

Gopkg.toml
```diff
@@ -65,3 +65,7 @@ version="v1.4.7"
   [[prune.project]]
   name = "sigs.k8s.io/cluster-api"
   unused-packages = false
+
+[[constraint]]
+  name = "github.com/aws/aws-sdk-go"
+  version = "1.17.5"
```
detiber: Any reason for the specific constrain here rather than just `dep ensure -update github.com/aws/aws-sdk-go`?

sethp-nr: I'm not sure, but I believe @rudoi needed a newer version for the 

We can remove the implicit `^` constraint for something more open ended like `>=`, if that'd be your preference?

rudoi: I don't recall manually adding this constraint, but git blame doesn't lie. 🤔

detiber: @sethp-nr >= would be good

PR #707, opened 2019-04-08
Fixes a bug and adds tests for kubeadm defaults

### LINKED ISSUE #685: Add unit tests for Kubeadm defaults
/kind feature

We should add some unit tests to [kubeadm's defaulting functions](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/aws/services/kubeadm/aws_defaults.go) to ensure the output of those functions match some set of expectations.

chuckha:
The pointers were not working as expected so the API is changing
to be more functional and leverage kubernetes' DeepCopy function.

This PR fixes a bug (bad pointer work) with regards to kubeadm defaults being set properly on the JoinConfiguration option.

This bug still exists in the neighboring functions but will be addressed in following PRs.

This adds tests to that function and makes the function more functional and easier to use.

Fixes #685 

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine_scope.go
+++ b/pkg/cloud/aws/actuators/machine_scope.go
@@ -97,6 +97,16 @@ func (m *MachineScope) Region() string {
 	return m.Scope.Region()
 }
 
+// GetMachine returns the machine wrapped in the scope.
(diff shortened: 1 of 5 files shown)


chuckha:
oh yeah those failing tests are 100% vacation brain

PR #708, opened 2019-04-09
Update listed v1.14 AMIs to v1.14.1

detiber:
Update listed v1.14 AMIs to v1.14.1

**Release note**:
```release-note
NONE
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,19 @@ policy may be made to more closely align with other providers in the Cluster API
 
 ------
 
+## Kubernetes versions with published AMIs
+
+Note: These AMIs are not updated for security fixes and it is recommended to always use the latest patch version for the Kubernetes version you wish to run. For production-like environments, it is highly recommended to build and use your own custom images.
+
+| Kubernetes minor version | Kubernetes full version |
+|-|-|
+| v1.13                    | v1.13.3                 |
+|                          | v1.13.5                 |
+| v1.14                    | v1.14.0                 |
+|                          | v1.14.1                 |
+
+------
+
 ## Documentation
 
 Documentation is in the `/docs` directory, and the [index is here](docs/README.md).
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -60,6 +60,19 @@ policy may be made to more closely align with other providers in the Cluster API
...
+## Kubernetes versions with published AMIs
+
+Note: These AMIs are not updated for security fixes and it is recommended to always use the latest patch version for the Kubernetes version you wish to run. For production-like environments, it is highly recommended to build and use your own custom images.
+
+| Kubernetes minor version | Kubernetes full version |
+|-|-|
+| v1.13                    | v1.13.3                 |
+|                          | v1.13.5                 |
+| v1.14                    | v1.14.0                 |
+|                          | v1.14.1                 |
```
detiber: @chuckha ptal

README.md
```diff
@@ -60,6 +60,19 @@ policy may be made to more closely align with other providers in the Cluster API
 
 ------
 
+## Kubernetes versions with published AMIs
+
+Note: These AMIs are not updated for security fixes and it is recommended to always use the latest patch version for the Kubernetes version you wish to run. For production-like environments, it is highly recommended to build and use your own custom images.
```
vincepri:
```suggestion
Note: These AMIs are offered as a guideline implementation and aren't meant to be used in production environments. AMIs aren't guaranteed to be updated for latest security fixes and it is recommended to always use the latest patch version for the Kubernetes version you wish to run. Users meaning to run Cluster API AWS in production critical environments should fork or use custom images.
```

Reworded a little

README.md
chuckha: perf ty

README.md
chuckha: the rewording feels more complex to me -- what do you think the rewording adds that the original version missed?

vincepri: It was a little confusing at first if the security fixes were related to Kubernetes or base OS images. The suggestion I proposed is a little longer but it should make it more clear. Feel free to ignore though, no reason to hold the PR any longer.

chuckha: I don't think the rewording clarifies whether it's the base OS or k8s that is not being updated for security issues, but I think it's ok to be ambiguous in this case since security fixes aren't being made for either k8s or base os

Issue #641, opened 2019-03-11
Logging as a proper dependency


chuckha:
/kind feature

**Describe the solution you'd like**
I would like logging to not be global functions but part of the structs using proper dependency injection so we can test and control how the logger behaves if it's important to the test

This allows us to treat logging as a first-class citizen. For example, we cannot append context to log lines without modifying each and every log line. 

One solution could be to figure out what subset of functions we use in klog, make an interface for that and wrap klog up in a struct that satisfies that interface while we pass that struct around wherever it is needed. 

At this point `klog` becomes an implementation detail and we can swap it out with any other logger and implement our interface. We can also create our own loggers like `contextKlog` that defines the `Infof/Info` to always print a context before calling the underlying implementation.

This would make #323 a trivial issue.

**Anything else you would like to add:**
Since logging isn't part of the important features like in kubeadm this issue is backlogged but would be a massive refactor that should be done one piece at a time, not all at once. An all-at-once PR will not be accepted for this issue because the changes will be too big. This problem must be broken down and merged over time. If you'd like to tackle this I have ideas for how to do this work in pieces but do not want to dictate implementation here and leave it up to the issue owner.

vincepri:
[logr](https://github.com/go-logr/glogr) should be a good starting point for context based logging. Controller-runtime already supports it and `klog` can be still be used as logging output.

detiber:
We should make sure not to duplicate some of the work being done to improve klog itself with this work.

chuckha:
@detiber The only part of this that would duplicate work is writing a context-logger ourselves. That will likely be done in klog and klog will likely use the logr interface and our structs will likely require logr interfaces for logging. The example given was in case `klog` doesn't do exactly what we want, the overall architecture would be flexible enough for us to implement whatever logger we like.

The goal here is to restructure our work to treat logging like a real dependency and not a global.

PR #713, opened 2019-04-11
Logging

chuckha:
This is the start to adding logging as a proper dependency through the project.

Related to #641

This only imports `github.com/go-logr/logr` and shows how to use it in the project. This does not replace any `klog` calls yet.

The `klog` implementation was mostly lifted from `glogr` with some modifications.

**Release note**:
```release-note
NONE
```


pkg/logging/klog/logger.go
```diff
@@ -0,0 +1,160 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package klog implements logr using klog.
+// This particular implementation is mostly copy pasta'd from https://github.com/go-logr/glogr/blob/master/glogr.go
```
chuckha: making sure this line is seen

detiber: Any reason for the custom implementation here vs using [klogr](https://github.com/kubernetes/klog/blob/master/examples/klogr/main.go)?

chuckha: oops, no, failed to notice that dir in `klog`. However, I do prefer dealing with structs instead of things like `New` that return interfaces, but I can live with it.

Updating PR, thanks for pointing that out!

cmd/manager/main.go
```diff
@@ -70,7 +70,8 @@ func main() {
 
 	// Initialize cluster actuator.
 	clusterActuator := cluster.NewActuator(cluster.ActuatorParams{
-		Client: cs.ClusterV1alpha1(),
+		Client:         cs.ClusterV1alpha1(),
+		LoggingContext: "[cluster actuator]",
```
chuckha: We'll have to keep the klog stuff like `flag.Parse` alive in the manager until we have a better solution

detiber:
Looks good to me, are you planning on extending this to the machine actuator as well as part of this PR or a separate PR?

chuckha:
yes, this will look identical in the machine PR

chuckha:
oops sorry, separate PR

PR #714, opened 2019-04-11
Logging

### LINKED ISSUE #323: append some context to logs
/kind feature

I'd like to know if logs are coming from the machine actuator or the controller actuator. would be happy with [controller] or [machine] in front of the log.

This is up for discussion, but it would be useful in the effort to make this easier to debug

**Environment:**

### LINKED ISSUE #641: Logging as a proper dependency

chuckha:

Fixes #641 
Fixes #323 

I'm adding comments to the parts you should pay attention to. Generally I've tried to make the logging a bit more consistent and have changed a few things, but left a few things unchanged that could use improvement. This now logs objects as json and we should probably clean up the logging, but it is *much* easier to understand what the manager is actually doing now.

**Release note**:
```release-note
NONE
```


pkg/cloud/aws/services/certificates/certificates.go
```diff
@@ -69,73 +68,78 @@ type Config struct {
...
-		}
-		if kp == nil {
-			return v1alpha1.KeyPair{
-				Cert: EncodeCertPEM(x509Cert),
-				Key:  EncodePrivateKeyPEM(privKey),
-			}, nil
-		}
-		kp.Cert = EncodeCertPEM(x509Cert)
-		kp.Key = EncodePrivateKeyPEM(privKey)
+func generateCACert(kp *v1alpha1.KeyPair, user string) (v1alpha1.KeyPair, error) {
```
chuckha: I changed this function to only do one thing (generate certs) instead of one or the other

cmd/manager/main.go
```diff
@@ -71,12 +71,13 @@ func main() {
 	// Initialize cluster actuator.
 	clusterActuator := cluster.NewActuator(cluster.ActuatorParams{
 		Client:         cs.ClusterV1alpha1(),
-		LoggingContext: "[cluster actuator]",
+		LoggingContext: "[\033[32mcluster actuator\033[0m]",
```
chuckha: the color makes it super easy to read, but I'm ok if this isn't good for whatever reason

detiber: I do like the idea of adding color, but I'm not necessarily sure we can rely on the user's shell supporting it.

chuckha:
@detiber I made a first pass at adding namespaces, any places I've missed please add a comment to the PR

cmd/manager/main.go
vincepri: This should probably be configurable

cmd/manager/main.go
```diff
@@ -71,12 +71,13 @@ func main() {
...
 	})
 
 	// Initialize machine actuator.
 	machineActuator := machine.NewActuator(machine.ActuatorParams{
-		Client: cs.ClusterV1alpha1(),
+		Client:         cs.ClusterV1alpha1(),
+		LoggingContext: "[\033[31mmachine actuator\033[0m]",
```
vincepri: Same as above, colors should be configurable. We might also want to use a proper library that detects a terminal

pkg/cloud/aws/actuators/scope.go
```diff
@@ -76,6 +77,7 @@ func NewScope(params ScopeParams) (*Scope, error) {
 		ClusterClient: clusterClient,
 		ClusterConfig: clusterConfig,
 		ClusterStatus: clusterStatus,
+		Logger:        params.Logger,
```
vincepri: We could use the information we have in the cluster and machine Scope to prefix WithName with apigroup + /namespace + [/machine/<id> or / cluster/<id>] and avoid the repetition.

pkg/cloud/aws/services/certificates/certificates.go
```diff
@@ -69,73 +68,78 @@ type Config struct {
 
 // ReconcileCertificates generate certificates if none exists.
 func (s *Service) ReconcileCertificates() error {
-	klog.V(2).Infof("Reconciling certificates for cluster %q", s.scope.Cluster.Name)
-	clusterCAKeyPair, err := getOrGenerateCACert(&s.scope.ClusterConfig.CAKeyPair, clusterCA)
-	if err != nil {
-		return errors.Wrapf(err, "Failed to generate certs for %q", clusterCA)
-	}
-	s.scope.ClusterConfig.CAKeyPair = clusterCAKeyPair
+	s.scope.V(2).Info("Reconciling certificates", "cluster-name", s.scope.Cluster.Name, "cluster-namespace", s.scope.Cluster.Namespace)
```
vincepri: Should this be `s.scope.Logger` for consistency?

cmd/manager/main.go
chuckha: That's a feature for a different ticket. I'm removing the colors in a sec. Can you open an issue with this feature request?

detiber: @vincepri a library for detecting the terminal won't really work here, since this is running in a pod, and the user's shell will be pulling the logs through the api server.

vincepri: To be clear I don't think we should add colors given that the support depends on user configuration and might not look pretty for everyone. Given that this is going away, then it's all good!

pkg/cloud/aws/services/certificates/certificates.go
chuckha: Ooooops I think `s.scope.V(2)` is a little cleaner and what I meant to do, do you prefer `s.scope.Logger` or `s.scope.V(n)`?

pkg/cloud/aws/actuators/scope.go
detiber: Actually, I really like this idea, except I would say:
```
apigroup + /namespace + /cluster/ [ + machine/ ], since the machine will have a cluster (at least with the current code).

File: pkg/cloud/aws/services/certificates/certificates.go (hunk 5 above)
Comment (vincepri) [reply pkg/cloud/aws/services/certificates/certificates.go]: I don't mind either way 😄, I think we should pick one and suggest it as default just for consistency

File: pkg/cloud/aws/actuators/scope.go (hunk 4 above)
Comment (vincepri) [reply pkg/cloud/aws/actuators/scope.go]: Even better!

File: pkg/cloud/aws/services/certificates/certificates.go (hunk 5 above)
Comment (chuckha) [reply pkg/cloud/aws/services/certificates/certificates.go]: totally, thanks for pointing out inconsistency! This was a mind numbing task haha

File: pkg/cloud/aws/actuators/scope.go (hunk 4 above)
Comment (chuckha) [reply pkg/cloud/aws/actuators/scope.go]: yeeee that sounds great

File: pkg/cloud/aws/services/kubeadm/aws_defaults_test.go [hunk 6]
```diff
@@ -20,12 +20,13 @@ import (
...
 	"testing"
 
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloudtest"
+
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	kubeadmv1beta1 "k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1alpha1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/actuators"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/kubeadm"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloudtest"
```
vincepri: vscode reordering? Seems odd to make cloudtest a separate import group

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -115,7 +118,7 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) error {
 	}
 
 	if err := ec2svc.DeleteNetwork(); err != nil {
-		klog.Errorf("Error deleting cluster %v: %v.", cluster.Name, err)
+		a.log.Error(err, "Error deleting cluster", "cluster-name", cluster.Name, "cluster-namespace", cluster.Namespace)
```
vincepri: Could this use the scoped logger?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -112,7 +117,7 @@ func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachine
 				return false, errors.Wrapf(err, "failed to verify existence of machine %q", m.Name())
 			}
 
-			klog.V(2).Infof("Machine %q should join the controlplane: %t", scope.Machine.Name, ok)
+			a.log.V(2).Info("Machine joining control plane", "machine-name", scope.Machine.Name, "machine-namespace", scope.Machine.Name, "should-join-control-plane", ok)
```
vincepri: Same here, could this use the scoped logger?

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -136,54 +143,54 @@ func (m *MachineScope) Close() {
...
 	_, err = m.storeMachineStatus(latestMachine)
 	if err != nil {
-		klog.Errorf("[machinescope] failed to store provider status for machine %q in namespace %q: %v", m.Machine.Name, m.Machine.Namespace, err)
+		m.Error(err, "failed to store machine provider status")
 	}
 }
 
 // MachineConfigFromProviderSpec tries to decode the JSON-encoded spec, falling back on getting a MachineClass if the value is absent.
-func MachineConfigFromProviderSpec(clusterClient client.MachineClassesGetter, providerConfig clusterv1.ProviderSpec) (*v1alpha1.AWSMachineProviderSpec, error) {
+func MachineConfigFromProviderSpec(clusterClient client.MachineClassesGetter, providerConfig clusterv1.ProviderSpec, log logr.Logger) (*v1alpha1.AWSMachineProviderSpec, error) {
```
vincepri: Can this function be made a method? Is it used anywhere else? If so, I'm not sure if the log lines should actually be in here or expect the caller would log.

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -136,54 +143,54 @@ func (m *MachineScope) Close() {
...
-			return unmarshalProviderSpec(&mc.ProviderSpec)
+			return unmarshalProviderSpec(&mc.ProviderSpec, log)
 		}
 	}
 
 	return &config, nil
 }
 
-func unmarshalProviderSpec(spec *runtime.RawExtension) (*v1alpha1.AWSMachineProviderSpec, error) {
+func unmarshalProviderSpec(spec *runtime.RawExtension, log logr.Logger) (*v1alpha1.AWSMachineProviderSpec, error) {
```
vincepri: Same as above

pkg/cloud/aws/actuators/scope.go
```diff
@@ -70,12 +72,17 @@ func NewScope(params ScopeParams) (*Scope, error) {
...
+		params.Logger = klogr.New().WithName("default-logger")
+	}
+
 	return &Scope{
 		AWSClients:    params.AWSClients,
 		Cluster:       params.Cluster,
 		ClusterClient: clusterClient,
 		ClusterConfig: clusterConfig,
 		ClusterStatus: clusterStatus,
+		Logger:        params.Logger.WithName(params.Cluster.APIVersion).WithName(params.Cluster.Namespace).WithName(params.Cluster.Name),
```
vincepri: How does chaining `WithName` work here? It'd be useful to form a path here with proper prefixes like in @detiber's other comment

pkg/cloud/aws/services/certificates/certificates.go
vincepri: Might be worthwhile to add a comment / description as part of the change

pkg/cloud/aws/services/ec2/ami.go
```diff
@@ -99,15 +97,14 @@ func (i images) Len() int {
...
+// TODO(chuckha) Ignoring errors until this causes a problem
 func (i images) Less(k, j int) bool {
 	firstTime, err := time.Parse(createDateTimestampFormat, aws.StringValue(i[k].CreationDate))
 	if err != nil {
-		klog.Infof("unable to parse an AMI creation timestamp: %q", aws.StringValue(i[k].CreationDate))
 		return false
 	}
 	secondTime, err := time.Parse(createDateTimestampFormat, aws.StringValue(i[j].CreationDate))
 	if err != nil {
-		klog.Infof("unable to parse an AMI creation timestamp: %q", aws.StringValue(i[j].CreationDate))
```
vincepri: Removing these might be making it hard for users to debug, is there any way to bubble this up? Opening a related issue if there is no simple solution is completely ok

pkg/cloud/aws/services/ec2/bastion.go
```diff
@@ -72,21 +71,21 @@ func (s *Service) ReconcileBastion() error {
...
-		klog.V(4).Info("Skipping bastion deletion in unmanaged mode")
+		s.scope.V(4).Info("Skipping bastion deletion in unmanaged mode")
 		return nil
 	}
 
 	instance, err := s.describeBastionInstance()
 	if err != nil {
 		if awserrors.IsNotFound(err) {
-			klog.V(2).Info("bastion instance does not exist")
+			s.scope.V(2).Info("bastion instance does not exist")
```
vincepri: These bastion logs should probably have a reference to the id or instance name

pkg/cloud/aws/actuators/scope.go
chuckha: Chaining adds `/` between each .WithName

pkg/cloud/aws/services/ec2/bastion.go
chuckha: yeah, agreed, will add that

pkg/cloud/aws/actuators/cluster/actuator.go
chuckha: yes it could

pkg/cloud/aws/actuators/machine_scope.go
chuckha: yes, this needs refactoring. I'll add it to the list of follow on issues.

pkg/cloud/aws/actuators/machine_scope.go
chuckha: also added

pkg/cloud/aws/services/certificates/certificates.go
chuckha: I don't understand the request. I like to reserve comments for when something unexpected is happening, but that's not really the case here. Can you elaborate on what you'd like changed?

pkg/cloud/aws/services/ec2/ami.go
chuckha: Done. This is part of #716.

pkg/cloud/aws/services/ec2/bastion.go
chuckha: sure

pkg/cloud/aws/services/certificates/certificates.go
vincepri: Oh, I noticed the functions don't have any comments on what they do, just adding a small comment might be useful :)

chuckha:
@vincepri @detiber Last pass? Anything major will be tracked https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/716

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -33,17 +35,22 @@ type MachineScopeParams struct {
...
 }
 
 // NewMachineScope creates a new MachineScope from the supplied parameters.
 // This is meant to be called for each machine actuator operation.
 func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
-	scope, err := NewScope(ScopeParams{AWSClients: params.AWSClients, Client: params.Client, Cluster: params.Cluster})
+	scope, err := NewScope(ScopeParams{
+		AWSClients: params.AWSClients,
+		Client:     params.Client, Cluster: params.Cluster,
+		Logger: params.Logger,
```
detiber: Should the `WithName` context be added here instead of below?

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -57,7 +64,7 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 	if params.Client != nil {
 		machineClient = params.Client.Machines(params.Machine.Namespace)
 	}
-
+	scope.Logger = scope.Logger.WithName(params.Machine.Name)
```
detiber: Should this be: `scope.Logger.WithName(params.Cluster.APIVersion).WithName(params.Cluster.Namespace).WithName(params.Cluster.Name).WithName(params.Machine.Name)` instead to align with the cluster scope logger and also capture the cluster context?

chuckha: No, the call to NewScope above on line 44 will create that namespaced context and then this line will append the machine name to the end of it. So at this point it's already got APIVersion, Namespace, Name as part of the context

pkg/cloud/aws/actuators/machine_scope.go
chuckha: No, otherwise machine will come before the APIVersion

PR #718, opened 2019-04-12
feat: support customizing root device size

rudoi:

This PR adds `rootDeviceSize` into the AWSMachineProviderSpec, allowing customization of the provisioned EC2 instance's root volume size.

**Release note**:

```release-note
Add support for customizing instance root device size
```

### DIFF
--- a/pkg/apis/awsprovider/v1alpha1/awsmachineproviderconfig_types.go
+++ b/pkg/apis/awsprovider/v1alpha1/awsmachineproviderconfig_types.go
@@ -74,6 +74,10 @@ type AWSMachineProviderSpec struct {
 	// +optional
 	KeyName string `json:"keyName,omitempty"`
 
+	// RootDeviceSize is the size of the root volume.
+	// +optional
+	RootDeviceSize int64 `json:"rootDeviceSize,omitempty"`
+
 	// KubeadmConfiguration holds the kubeadm configuration options
 	// +optional
 	KubeadmConfiguration KubeadmConfiguration `json:"kubeadmConfiguration,omitempty"`
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -403,7 +403,7 @@ func (s *Service) runInstance(role string, i *v1alpha1.Instance) (*v1alpha1.Inst
 		}
 
 		input.BlockDeviceMappings = []*ec2.BlockDeviceMapping{
-			&ec2.BlockDeviceMapping{
+			{
 				DeviceName: rootDeviceName,
 				Ebs: &ec2.EbsBlockDevice{
 					DeleteOnTermination: aws.Bool(true),
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -576,6 +576,10 @@ func (s *Service) getInstanceRootDeviceSize(instance *ec2.Instance) (*int64, err
 	return nil, nil
 }
 
+// sdkToInstance converts an AWS EC2 SDK instance to the CAPA instance type.
+// converters.SDKToInstance populates all instance fields except for rootVolumeSize,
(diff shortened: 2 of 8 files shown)


pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -509,3 +528,55 @@ func (s *Service) getInstanceENIs(instanceID string) ([]*ec2.NetworkInterface, e
...
+				return nil, errors.Errorf("no volumes found for id %q", aws.StringValue(bdm.Ebs.VolumeId))
+			}
+
+			return out.Volumes[0].Size, nil
+		}
+	}
+	return nil, nil
+}
+
+func (s *Service) sdkToInstance(v *ec2.Instance) (*v1alpha1.Instance, error) {
```
rudoi: I'm not a huge fan of this function... but I also did not want to make converters.SDKToInstance a method of service... open to any suggestions 😢

pkg/apis/awsprovider/v1alpha1/awsmachineproviderconfig_types.go
```diff
@@ -74,6 +74,10 @@ type AWSMachineProviderSpec struct {
 	// +optional
 	KeyName string `json:"keyName,omitempty"`
 
+	// RootDeviceSize is the size of the root volume.
+	// +optional
+	RootDeviceSize int64 `json:"rootDeviceSize,omitempty"`
```
sethp-nr: I think you should duplicate this change in the CRD yaml: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/config/crds/awsprovider_v1alpha1_awsmachineproviderspec.yaml

rudoi: ah! forgot to run `make generate-crds`. thanks Seth! updated.

pkg/cloud/aws/services/ec2/instances.go
vincepri: Hmm, could we get the volume information when we do a Describe call instead of having a method for it? I think there should be only 2 places where this happens.

rudoi: As in the Describe for the instance? Unfortunately that call returns everything about the volumes EXCEPT size. :(

vincepri: Got it, the only concern I have is that now there are two functions to convert an SDK ec2 instance to our own instance type, which might be confusing. Maybe let's leave a comment somewhere?

rudoi: It is definitely confusing. I'll add some comments. The thing that bugs me the most is that the original `converters.SDKToInstance` feels dishonest now, because the "our" instance that is returned by that is now incomplete, and not really usable without also sugaring in the Volume information.

rudoi: Updated. Comment language is hard :sob:.

chuckha: One concern here is that now the SDKToInstance that will give a complete instance is private while the SDKToInstance that will give an incomplete instance is public.

Does converters.SDKToInstance have any use now? Should that method be private and this one be public? I think we should also add a comment to converters.SDKToInstance that outline why that is insufficient for a complete instance representation.

rudoi: I wonder if maybe converters.SDKToInstance's logic should just be inside service.SDKToInstance and that method should be made public. converters being private wouldn't do us any good I think?

chuckha: That makes sense since converting from sdk to instance now *requires* a call to the ec2 service

rudoi: Ok so I made the above change. I'm not sure if/how we'd deprecate converters.SDKToInstance? I left it in there with a comment about how it's insufficient.

chuckha:
/lgtm

/assign @detiber @vincepri 
for approval

PR #726, opened 2019-04-18
fix: Don't try to update root size when it's unset

sethp-nr:

This commit looks for empty RootDeviceSize in the spec and ignores it.
Otherwise, none of our control plane machines were updating with this
error:

```
E0418 23:07:48.250925       1 controller.go:214] Error updating machine "ns/controlplane-2": found attempt to change immutable state for machine "controlplane-2": ["Root volume size cannot be mutated from 8 to 0"]
```

Fixes #

**Release note**:

```release-note
fixes issue updating machines with an unspecified root device size
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine/actuator.go
+++ b/pkg/cloud/aws/actuators/machine/actuator.go
@@ -306,7 +306,7 @@ func (a *Actuator) isMachineOutdated(machineSpec *v1alpha1.AWSMachineProviderSpe
 	}
 
 	// Root Device Size
-	if machineSpec.RootDeviceSize != instance.RootDeviceSize {
+	if machineSpec.RootDeviceSize > 0 && machineSpec.RootDeviceSize != instance.RootDeviceSize {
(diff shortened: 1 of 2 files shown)


sethp-nr:
h/t to @rudoi for his help in finding/fixing this bug.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -306,7 +306,7 @@ func (a *Actuator) isMachineOutdated(machineSpec *v1alpha1.AWSMachineProviderSpe
 	}
 
 	// Root Device Size
-	if machineSpec.RootDeviceSize != instance.RootDeviceSize {
+	if machineSpec.RootDeviceSize > 0 && machineSpec.RootDeviceSize != instance.RootDeviceSize {
```
ashish-amarnath: Could you please update the unit tests for this as well?

ncdc:
I just ran into this, and it's 👍 from me, so

/lgtm

Issue #730, opened 2019-04-22
Bazel version for periodics is outdated


chuckha:
/kind bug

We've been seeing consistent failures in the prow periodic job. The [current failure](https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/periodic-cluster-api-provider-aws-bazel-e2e/1120356433536225283) is a Bazel problem. Prow is using 0.18.1. The output that I reproduced locally with that Bazel version is:

```
 chuckh-a01:cluster-api-provider-aws cha$ ./bazel build  //pkg/cloud/aws/services/mocks:mocks \
> //pkg/cloud/aws/services/ec2/mock_ec2iface:mocks \
> //pkg/cloud/aws/services/elb/mock_elbiface:mocks
WARNING: failed to raise resource limit 8 to 524288: Invalid argument
Extracting Bazel installation...
Starting local Bazel server and connecting to it...
INFO: Repository rule 'bazel_skylib' returned: {"remote": "https://github.com/bazelbuild/bazel-skylib", "commit": "6741f733227dc68137512161a5ce6fcf283e3f58", "shallow_since": "2019-02-08", "init_submodules": False, "verbose": False, "strip_prefix": "", "patches": [], "patch_tool": "patch", "patch_args": ["-p0"], "patch_cmds": [], "name": "bazel_skylib"}
ERROR: /private/var/tmp/_bazel_cha/b7ce3b297716e241b109da3b2c60846f/external/bazel_gomock/gomock.bzl:100:5: name '_gomock_prog_gen' is not defined
ERROR: Skipping '//pkg/cloud/aws/services/ec2/mock_ec2iface:mocks': error loading package 'pkg/cloud/aws/services/ec2/mock_ec2iface': Extension 'gomock.bzl' has errors
ERROR: error loading package 'pkg/cloud/aws/services/ec2/mock_ec2iface': Extension 'gomock.bzl' has errors
INFO: Elapsed time: 34.954s
INFO: 0 processes.
FAILED: Build did NOT complete successfully (0 packages loaded)
    currently loading: pkg/cloud/aws/services/ec2/mock_ec2iface ... (3 packages)
```

but with bazel 0.21.0 the build finished successfully.

We must update the kubekins version the prow job is using.

/assign
/priority important-longterm
/lifecycle active

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
[...]

PR #731, opened 2019-04-22
Ensure bazel is the correct version

chuckha:

This PR adds a step to the WORKSPACE that ensures the version of bazel being run is a supported version of this project.

related to #730 

Periodics will now fail with this error until the kubekins version is updated (coming soon)

```release-note
NONE
```

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -40,6 +40,20 @@ http_archive(
     urls = ["https://github.com/kubernetes/repo-infra/archive/df02ded38f9506e5bbcbf21702034b4fef815f2f.tar.gz"],
 )
 
+http_archive(
+    name = "bazel_skylib",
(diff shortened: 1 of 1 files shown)


WORKSPACE
```diff
@@ -40,6 +40,22 @@ http_archive(
     urls = ["https://github.com/kubernetes/repo-infra/archive/df02ded38f9506e5bbcbf21702034b4fef815f2f.tar.gz"],
 )
 
+skylib_version = "0.8.0"
```
detiber: The variable here feels a bit misleading since updating it would also require updating the sha256 in the http_archive definition.

chuckha: that's a fair point, I copy/pasta'd and should have used more discretion. I will merge this version into the URL like we do in the above archives

PR #742, opened 2019-04-24
Switch Kustomize to use bases and not resources.

### LINKED ISSUE #611: Kustomize should use bases instead of resources
/kind feature

In an effort to move to support Kustomize2, @figo pointed out that there is a nice way to gain support without breaking changes, while still maintaining backward compatibility. Kustomize supports `bases`, which is a list of folders that can be a relative path (see [cluster-api](https://github.com/kubernetes-sigs/cluster-api/tree/master/config) as reference). 

Each folder will have its own `kustomization.yaml` file referencing the files that should be exported.

/priority important-longterm
/milestone Next

sfzylad:
It will prevent problems while switching to kustomize2.

Fixes #611 

**Release note**:

```release-note

```


config/BUILD.bazel
```diff
@@ -34,22 +34,35 @@ controller_gen(
...
     name = "provider-non-stamped-base",
     srcs = [
-        "//config/default:kustomization.yaml",
-        "//config/default:rbac_role_binding_patch.yaml",
-        "//config/manager:manager.yaml",
+        "//config:default/rbac_role_binding_patch.yaml",
+        "//config:manager/manager.yaml",
         "config",
+        "//config:default/kustomization.yaml",
+        #"//config:rbac/kustomization.yaml",
```
detiber: Is the comment here intentional?

sfzylad: It suppose to be removed. Fixed in the new commit.

config/manager/BUILD.bazel
```diff
@@ -1,20 +0,0 @@
...
-# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-exports_files(
-    [
-        "manager.yaml",
-    ],
-    visibility = ["//visibility:public"],
-)
```
vincepri: Any reason why this bazel file got removed?

config/default/BUILD.bazel
```diff
@@ -1,21 +0,0 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-
-exports_files(
-    [
-        "kustomization.yaml",
-        "rbac_role_binding_patch.yaml",
-    ],
-    visibility = ["//visibility:public"],
-)
```
vincepri: Same here

config/manager/BUILD.bazel
sfzylad: As far as I remember it clashed with the entries I added to `config/BUILD.bazel` and removing it fixed it.

config/default/BUILD.bazel
sfzylad: As far as I remember it clashed with the entries I added to `config/BUILD.bazel` and removing it fixed it.

sfzylad:
There's one more small change on the way. Please don't approve it yet.

sfzylad:
It seems ready for review.

randomvariable:
Will do the review for tomorrow. Sorry about the delay.

PR #745, opened 2019-04-25
Fix control plane node join logic

### LINKED ISSUE #740: Terminating controlplane-0 results in a new cluster
/kind bug

We're running a HA (3 control plane) cluster, and in order to pick up a change to the control plane machines we terminated `controlplane-0` after making a change to its spec.

When it came back up, it had no JoinConfiguration but instead an InitConfiguration, leading to our having two independent clusters. That was more or less OK until we terminated `controlplane-1`, at which point we lost quorum and the HA cluster's data.

Controlplane-0 would re-join the existing machines.

**Environment:**

- Cluster-api-provider-aws version: master-ish
- Kubernetes version: (use `kubectl version`): 1.13.2
- OS (e.g. from `/etc/os-release`): ubuntu

ashish-amarnath:
Fixes the control plane node join logic
Fixes #740

**Release note**:

```release-note

```


pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -112,16 +113,21 @@ func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachine
...
+			controlplaneExists, err = ec2svc.MachineExists(m)
 			if err != nil {
-				return false, errors.Wrapf(err, "failed to verify existence of machine %q", m.Name())
+				a.log.V(2).Info("Could not verify existence of control plane machine, continuing", "machine-name", scope.Machine.Name, "machine-namespace", scope.Machine.Namespace)
+				continue
 			}
 
-			a.log.V(2).Info("Machine joining control plane", "machine-name", scope.Machine.Name, "machine-namespace", scope.Machine.Name, "should-join-control-plane", ok)
-			return ok, nil
+			if !controlplaneExists {
```
tahsinrahman: you should return true if `controlplaneExists == true`, and return `false` at the end of loop.

currently, it won't work if there are two controlplanes and the first one exists, `controlplaneExists` be false at the end of loop

ashish-amarnath:
This PR is untested. I am working on getting myself an AWS account to test this. 🤞

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: fixed.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -112,16 +113,21 @@ func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachine
...
+				a.log.V(2).Info("Control plane machine does not exist, continuing", "machine-name", scope.Machine.Name, "machine-namespace", scope.Machine.Namespace)
```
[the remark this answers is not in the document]
ashish-amarnath: good catch! fixed and also added some commentary.

ashish-amarnath:
I need some help testing this PR. I haven't managed to get myself an AWS account yet :(

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -107,23 +112,32 @@ func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachine
...
+				continue
 			}
 
 			ec2svc := ec2.NewService(m.Scope)
 
-			ok, err := ec2svc.MachineExists(m)
+			controlplaneExists, err = ec2svc.MachineExists(m)
 			if err != nil {
-				return false, errors.Wrapf(err, "failed to verify existence of machine %q", m.Name())
+				a.log.V(2).Info("Failed to verify existence of control plane machine", "machine-name", m.Machine.Name, "machine-namespace", m.Machine.Namespace)
```
ashish-amarnath: Thinking that this should return the error instead of `continuing`

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,51 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
 }
 
+// isNodeJoin determines if a machine, in scope, should join of the cluster.
 func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachines []*clusterv1.Machine) (bool, error) {
 	switch set := scope.Machine.ObjectMeta.Labels["set"]; set {
 	case "node":
+		// non control plane machines will always join the cluster.
 		return true, nil
 	case "controlplane":
+		// control plane machines will join the cluster if the cluster has an existing control plane.
```
vincepri:
```suggestion
		// Control plane machines will join the cluster if the cluster has an existing control plane.
```

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,51 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
 	return m1.Name == m2.Name && m1.Namespace == m2.Namespace
 }
 
+// isNodeJoin determines if a machine, in scope, should join of the cluster.
 func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachines []*clusterv1.Machine) (bool, error) {
 	switch set := scope.Machine.ObjectMeta.Labels["set"]; set {
 	case "node":
+		// non control plane machines will always join the cluster.
```
vincepri:
```suggestion
		// Worker nodes (not part of a control plane) machines will always join the cluster.
```

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,51 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
 	return m1.Name == m2.Name && m1.Namespace == m2.Namespace
 }
 
+// isNodeJoin determines if a machine, in scope, should join of the cluster.
 func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachines []*clusterv1.Machine) (bool, error) {
```
vincepri: Maybe we should rename this method to something like `shouldJoin`?

pkg/cloud/aws/actuators/machine/actuator.go
vincepri: Should we move the code within this case to its own method?

ashish-amarnath: Fixed. Also wanted to fix `Control plane` to `Controlplane`.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: Fixed. Also wanted to fix `Control plane` to `Controlplane`.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: The convention I have seen for methods returning `bool` is starting with  `Has`, `Is`, `Can` or `Allow`. `shouldJoin` kinda falls out of the convention.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: This code is still pretty manageable, esp considering we have pretty good code coverage for this method. But it's a good idea to keep that in mind with future changes.

sethp-nr:
Well, the code here seems to do the right thing in that my first `controlplane` instance does get created with a `JoinConfiguration` and tries to run `kubeadm join`.

Unfortunately, that didn't work because of https://github.com/kubernetes/kubeadm/issues/1432: I had to manually remove the etcd member with `sudo ETCDCTL_API=3 etcdctl --endpoints <FUNCTIONING_CONTROL_PLANE>:2379 --cacert /etc/kubernetes/pki/etcd/ca.crt --cert /etc/kubernetes/pki/etcd/peer.crt --key /etc/kubernetes/pki/etcd/peer.key member remove <MEMBER_ID>` and then edit the kubeadm-config configmap to remove the old controlplane from the ClusterStatus.

sethp-nr:
Also, I know this is a bigger question, but I did get to wondering if "machine existence" is the right way to decide whether to init or join. Do you think there's a way to depend on something in the cluster's status?

ashish-amarnath:
Yeah, I agree that we should make that decision on cluster status and not merely `MachineExists`. I don't think we have a status field for that today.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/apis/awsprovider/v1alpha1/awsclusterproviderstatus_types.go
https://github.com/kubernetes-sigs/cluster-api/blob/master/pkg/apis/cluster/v1alpha1/cluster_types.go

This would be something the new data model will have. @vincepri can confirm

sethp-nr:
Oh, good catch – I see the new `phase` field in the cluster status.

It sounds like whomever is responsible for that first `init` would be responsible for setting the cluster `phase` to something that indicates for new control plane machines to `join`, do I have that right? Would it be hasty to create a provider-specific field for now that we update when we reconcile the `cluster` and then drive this logic from that data, maybe as a follow-up?

sethp-nr:
And it was my pleasure to test this out – thank you for a quick turnaround!

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,50 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
 	switch set := scope.Machine.ObjectMeta.Labels["set"]; set {
```
detiber: Not directly related to this PR, but instead of switching on the `set` label, this could just check if `scope.Machine.Spec.Versions.ControlPlane != ""`

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,50 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
+				Machine:    cm,
+				Cluster:    scope.Cluster,
+				Client:     a.clusterClient,
+				Logger:     a.log,
+				AWSClients: scope.AWSClients,
 			})
 
 			if err != nil {
-				return false, errors.Wrapf(err, "failed to create machine scope for machine %q", cm.Name)
+				return false, errors.Wrapf(err, "failed to create machine scope for machine %q in namespace %q", cm.Name, cm.Namespace)
```
detiber: Is this needed with the new logging? @chuckha can you confirm?

If it is still needed, I would suggest using `return false, errors.Wrapf(err, "failed to create machine scope for machine %s/%s", cm.Namespace, cm.Name)` instead.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,50 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
 			}
 
 			ec2svc := ec2.NewService(m.Scope)
 
-			ok, err := ec2svc.MachineExists(m)
+			controlplaneExists, err = ec2svc.MachineExists(m)
 			if err != nil {
-				return false, errors.Wrapf(err, "failed to verify existence of machine %q", m.Name())
+				return false, errors.Wrapf(err, "failed to verify existence of machine %q in namespace %q", m.Machine.Name, m.Machine.Namespace)
```
detiber: Same as above, if namespace and machine is still needed with new logging this should use <namespace>/<name>

pkg/cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -498,3 +506,377 @@ func TestImmutableStateChange(t *testing.T) {
...
+				t,
+				getMockEC2APIDescribeInstancesNotFoundAndPass([]string{"master-0", "master-1"}, mockCtrl),
+			),
+			inputControlplaneMachines: getTestControlplaneMachines(),
+			expectedIsNodeJoin:        true,
+			expectedError:             nil,
+		},
+	}
+
+	for _, tc := range testCases {
```
chuckha: this needs to run each test case as a subtest using:

```
for _, tc := range testCases {
        ...everything else
    }
}
```

pkg/cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -498,3 +506,377 @@ func TestImmutableStateChange(t *testing.T) {
...
+		},
+		{
+			name: "should return error when unable to verify controlplane machine existence",
+			inputScope: getControlplaneMachineScope(
+				t,
+				getMockEC2APIDescribeInstancesFail([]string{"master-0", "master-1"}, mockCtrl),
+			),
+			inputControlplaneMachines: getTestControlplaneMachines(),
+			expectedIsNodeJoin:        false,
+			expectedError: fmt.Errorf(
```
chuckha: I think testing the exact error message is quite brittle, what do you think about simply testing nil or not for the error?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,50 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
 func (a *Actuator) isNodeJoin(scope *actuators.MachineScope, controlPlaneMachines []*clusterv1.Machine) (bool, error) {
 	switch set := scope.Machine.ObjectMeta.Labels["set"]; set {
 	case "node":
+		// Worker machines, not part of the controlplane, will always join the cluster.
 		return true, nil
 	case "controlplane":
+		// Controlplane machines will join the cluster if the cluster has an existing control plane.
+		controlplaneExists := false
+		var err error
 		for _, cm := range controlPlaneMachines {
```
detiber: Should there be some type of locking here to avoid parallel initialization between multiple control plane hosts?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,50 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
 			m, err := actuators.NewMachineScope(actuators.MachineScopeParams{
```
detiber: Do we actually need to create a new scope for each Control Plane machine? It seems odd to initialize a scope just to query the existence of an ec2 instance.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -96,37 +96,50 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
-			ok, err := ec2svc.MachineExists(m)
+			controlplaneExists, err = ec2svc.MachineExists(m)
 			if err != nil {
-				return false, errors.Wrapf(err, "failed to verify existence of machine %q", m.Name())
+				return false, errors.Wrapf(err, "failed to verify existence of machine %q in namespace %q", m.Machine.Name, m.Machine.Namespace)
 			}
 
-			a.log.V(2).Info("Machine joining control plane", "machine-name", scope.Machine.Name, "machine-namespace", scope.Machine.Name, "should-join-control-plane", ok)
-			return ok, nil
+			if !controlplaneExists {
```
chuckha: This logic could be streamlined by moving log lines out of this function

```
for ... {
    controlPlaneExists, err := ec2svc.MachineExists(m)
    if controlPlaneExists {
        return controlPlaneExists, err
    }
}
return false, nil
```

This reduces the complexity of this code since the complexity comes from wanting to add a log line. If you feel the log line is necessary it could be pulled to the caller like this:

```
willJoin, err := isNodeJoin(...)
if willJoin {
     a.log.Info("node is joining the cluster")
}
```

pkg/cloud/aws/actuators/machine/actuator.go
chuckha: depends on if we log the error or not. I'd expect us to log the error which means this could just be a .Wrap call like `errors.Wrap(err, "failed to create a machine scope")`

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: Yes! But the upstream cluster deployer doesn't support multiple controlplane nodes in the namespace.
https://github.com/kubernetes-sigs/cluster-api/blob/master/cmd/clusterctl/clusterdeployer/clusterclient/clusterclient.go#L1045
So there will be no racing to be the first `controlplane` machine.
Once the upstream master start supporting multiple controlplane machines in the cluster, we'll have to make `isNodeJoin` thread safe.

Maybe worth adding comments about this.

ashish-amarnath: I created this issue https://github.com/kubernetes-sigs/cluster-api/issues/925 and I'll link it in the code.

pkg/cloud/aws/actuators/machine/actuator_test.go
ashish-amarnath: Will do. Is the benefit here just to let the framework run the tests or are there other benefits or running it this way?

pkg/cloud/aws/actuators/machine/actuator_test.go
ashish-amarnath: Yeah, I understand your point. But for this method, I feel strict unit tests should be OK/ preferred.
If the error changes in a future change, the unit test will also need to be updated, which IMO, should be good, letting the unit test grow with the code.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: The idea is to determine whether or not this machine in scope should join the cluster or not and also be able to handle all "kinds" of machines- controlplane, worker, and any other we might add later. 

`scope.Machine.Spec.Versions.ControlPlane != ""` will just tell us whether this is a controlplane machine or not.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: This error should be logged. I'll ensure it is and it is inline w/ @detiber's suggestion

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: Fixed to use a shared scope.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: This error is wrapped in the machine actuator `Create` method so it should get logged.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: that would be better. Thanks!

pkg/cloud/aws/actuators/machine/actuator.go
detiber: So, multiple control-plane nodes should work today if using `clusterctl` it will instantiate them serially. We even have a make target for easy testing `make create-cluster-ha`

detiber: The serial instantiation is more for the benefit of `kubeadm` which will not support parallel `init` and controlplane `join` until v1.15.

pkg/cloud/aws/actuators/machine/actuator_test.go
chuckha: benefit is that a single subtest can fail while others continue to run. We get good output about which test case in the test table actually failed. Without it we just know this method failed but not which case.

pkg/cloud/aws/actuators/machine/actuator.go
detiber: Yes, I do think that we should be able to potentially special case other solutions as well, but currently we have two different ways of determining a control-plane machine within cluster-api upstream (for the purposes of clusterctl) and per-provider. If we can standardize on a single solution for controlplane or not, then it is less cognitive overhead for the end user. 

Removing the requirement of the `set` label for control plane instances would also reduce user confusion when trying to deploy a MachineSet or MachineDeployment, where it isn't overly clear that they need to include those in the labels for the template.

ashish-amarnath: I agree on not having 2 ways of doing the same. As it is not directly related to this PR, I'll follow-up with a PR for this.

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: Correct. I had forgotten about that. Thanks for reminding.

ashish-amarnath: @detiber This PR, this is a bug fix. We can make `isNodeJoin` thread safe as an enhancement and fix to this issue https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/763 
I don't think it should block this PR, do you disagree?

vincepri:
Generally LGTM, leaving the final review/lgtm to @chuckha and @sethp-nr

sethp-nr:
lgtm!

ashish-amarnath:
@chuckha needs `approve` too

chuckha:
/approve

headdesk.

PR #748, opened 2019-04-30
Fixes the infinite reconcile loop

chuckha:

* Uses patch for updating the cluster and machine specs
  - patch does not cause a re-reconcile in the capi controller
* Uses update for updating the cluster and machine status
  - update for status is ok since it does not update any of the metadata
    no re-reconcile is necessary for the capi controller

apparently there is no issue, but creating a cluster was weird.

I believe this will be the first step to fixing the long-create times we were seeing.

**Release note**:
```release-note
None
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine_scope.go
+++ b/pkg/cloud/aws/actuators/machine_scope.go
@@ -17,16 +17,18 @@ limitations under the License.
 package actuators
 
 import (
+	"encoding/json"
 	"fmt"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/client-go/util/retry"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1alpha1"
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
+	"sigs.k8s.io/controller-runtime/pkg/patch"
 	"sigs.k8s.io/yaml"
 )
 
@@ -69,6 +71,7 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 	return &MachineScope{
 		Scope:         scope,
 		Machine:       params.Machine,
+		MachineCopy:   params.Machine.DeepCopy(),
 		MachineClient: machineClient,
 		MachineConfig: machineConfig,
 		MachineStatus: machineStatus,
@@ -79,7 +82,9 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 type MachineScope struct {
 	*Scope
 
-	Machine       *clusterv1.Machine
+	Machine *clusterv1.Machine
+	// MachineCopy is used to generate a patch diff at the end of the scope's lifecycle.
+	MachineCopy   *clusterv1.Machine
 	MachineClient client.MachineInterface
 	MachineConfig *v1alpha1.AWSMachineProviderSpec
 	MachineStatus *v1alpha1.AWSMachineProviderStatus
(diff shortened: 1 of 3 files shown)


pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -117,9 +120,10 @@ func (m *MachineScope) GetScope() *Scope {
 
 // Close the MachineScope by updating the machine spec, machine status.
 func (m *MachineScope) Close() {
-	if m.MachineClient == nil {
+	if m.MachineClient == nil || m.Machine == nil {
```
detiber: What would cause `m.Machine` to be nil? Wouldn't we always have a Machine?

pkg/cloud/aws/actuators/scope.go
```diff
@@ -135,9 +138,10 @@ func (s *Scope) Region() string {
 
 // Close closes the current scope persisting the cluster configuration and status.
 func (s *Scope) Close() {
-	if s.ClusterClient == nil {
+	if s.ClusterClient == nil || s.Cluster == nil {
```
detiber: Is there a case where `s.Cluster` would be nil? Or is this more cautionary?

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -131,43 +135,40 @@ func (m *MachineScope) Close() {
...
-		// The machine may have status (nodeRef) that the latest doesn't yet
-		// have, however some timestamps may be rolled back a bit with this copy.
-		m.Machine.Status.DeepCopyInto(&latest.Status)
-		latest.Status.ProviderStatus = status
-		_, err = m.MachineClient.UpdateStatus(latest)
-		return err
-	}); err != nil {
-		m.Error(err, "error retrying on conflict")
+	m.Machine.Spec.ProviderSpec.Value = ext
+	old, err := m.MachineClient.Get(m.Machine.Name, v1.GetOptions{})
```
detiber: Instead of doing a `Get` here, should we be storing the old value when we instantiate the scope instead?

pkg/cloud/aws/actuators/scope.go
```diff
@@ -149,30 +153,40 @@ func (s *Scope) Close() {
...
-		latest.Status.ProviderStatus = status
-		_, err = s.ClusterClient.UpdateStatus(latest)
-		return err
-	}); err != nil {
-		s.Error(err, "failed to retry on conflict")
-	}
-	s.V(2).Info("Successfully updated cluster")
+	// First set the provider spec
+	s.Cluster.Spec.ProviderSpec.Value = ext
+	old, err := s.ClusterClient.Get(s.Cluster.Name, v1.GetOptions{})
```
detiber: Instead of doing a Get here, should we be storing the old value when we instantiate the scope instead?

pkg/cloud/aws/actuators/machine_scope.go
chuckha: heh, I think this snuck in. I'll remove it and cluster nil check.

pkg/cloud/aws/actuators/machine_scope.go
chuckha: could do, what's the benefit of storing it when we start? seems like that would only increase the amount of time something could update it.

detiber: I was thinking that it would ensure that we are only attempting to patch fields that we explicitly changed during the reconciliation. This would avoid us stomping a non-conflicting change that was made by a separate process. For example if another process modified annotations, labels, etc we wouldn't clobber those changes.

chuckha: So the assertion is that cluster-api will only be managing cluster-api owned fields/labels/annotations/etc and external tools will never touch the cluster-api owned fields/labels/annotations/etc. Conversely cluster-api will never update non-cluster-api owned fields/labels/annotations/etc. If these are correct then I agree with you and this should be storing an object copy at the init of scope.

detiber: So, if you consider things like Upgrade, where we are orchestration some type of end state programmatically or end users attempting to modify fields then we definitely have to take in to account external changes.

detiber: I would expect the changes mentioned above to be picked up during the next subsequent reconcile and not overwritten by the patch we are doing as part of the current reconciliation

chuckha: Ok, I think I'm on board. It still feels racy, but it's better than fetching right here, I agree.

chuckha:
it feels so good to not get stuck on Running Instance anymore.

@detiber ptal!

PR #749, opened 2019-05-01
[WIP] Node draining

### LINKED ISSUE #614: Cordon and Drain nodes when deleting machines
/kind feature

Currently when we delete Machines we just blindly delete the underlying instance. Instead we should probably Cordon and Drain the associated Node (if possible) prior to deleting the underlying instance. This will cause less disruption to workloads that are running on the Cluster and will also provide a much cleaner path during cluster upgrades.

The functionality for `drain` used to be within `kubectl` only, but has recently been refactored out into a reusable library: https://github.com/kubernetes/kubernetes/pull/72827

sfzylad:

Fixes #614 

**Release note**:

```release-note

```


pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -253,6 +256,19 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	defer scope.Close()
 
+	a.log.Info("Cordoning the node", machine.Name)
+	err, patchErr := a.cordon(ctx, cluster, machine, scope)
+	if err != nil {
+		return errors.Errorf("Faild to cordon the node: %+v", err)
```
detiber:
```suggestion
		return errors.Errorf("Failed to cordon the node: %+v", err)
```

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -253,6 +256,19 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	}
+
+	if patchErr != nil {
+		return errors.Errorf("Faild to cordon the node: %+v", err)
```
detiber:
```suggestion
		return errors.Errorf("Failed to cordon the node: %+v", patchErr)
```

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -253,6 +256,19 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	}
+
+	// sleep, so I can investigate the state of the node and pods with kubectl
+	time.Sleep(120 * time.Second)
```
detiber: Based on the comment, I'm assuming this is just for temporary debugging purposes?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -289,6 +305,66 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 	return nil
 }
 
+func (a *Actuator) cordon(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine, scope *actuators.MachineScope) (error, error) {
```
detiber: Returning two error values seems a bit odd to me, why not just a single error?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -289,6 +305,66 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+	a.log.Info("Inside cordon()")
+
+	node, err := a.getNodeObject(ctx, cluster, machine, scope)
+	if err != nil {
+		a.log.Info("ERROR: ", err.Error())
+		return nil, err
+	}
+
+	config, err := rest.InClusterConfig()
```
detiber: This looks like it will only work for self-managed clusters, since remote nodes would not be able to be cordoned using the in-cluster config.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -289,6 +305,66 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+func (a *Actuator) getNodeObject(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine, scope *actuators.MachineScope) (*apicorev1.Node, error) {
+	coreClient, err := a.coreV1Client(scope.Cluster)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to retrieve corev1 client for cluster %q", scope.Cluster.Name)
+	}
+
+	listOpt := metav1.ListOptions{}
+
+	for {
+		nodeList, err := coreClient.Nodes().List(listOpt)
```
detiber: Similar to the cordon comment, this will only work for nodes that exist in the same cluster as the machine objects (self-managed clusters).

pkg/cloud/aws/actuators/machine/actuator.go
sfzylad: Exactly, I"m making it sleep for a bit, so I can play around with kubectl to see if everything looks fine.

pkg/cloud/aws/actuators/machine/actuator.go
sfzylad: This is because I'm using these method that returns two erros. To be honest it was odd to me too:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/840be1a5ffda479425d2a4cf61d9b81be47f389c/vendor/k8s.io/kubernetes/pkg/kubectl/drain/cordon.go#L75:24

pkg/cloud/aws/actuators/machine/actuator.go
sfzylad: Right, I'll switch this bit once I'll get pods evixtion mechanism in place. Thanks for the feedback 👍

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: The 2nd error is non-nil if it was unable to create the two-way merge patch.

Honestly, the `CordonHelper` is small enough that I would probably recommend implementing its logic here. Also, I'm not sure we should be letting the helper send an "update" request for the Node in the event we got an error trying to create the merge patch. Patch >>> Update.

chuckha:
@sfzylad It would be great to get just the cordon stuff in and then get the drain in after. Do you think you could clean up this PR (mostly just the sleep) and we can get it merged?

pkg/cloud/aws/actuators/machine/actuator.go
sfzylad: This has been fixed.

pkg/cloud/aws/actuators/machine/actuator.go
sfzylad: It's been already fixed.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -269,6 +314,12 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	defer scope.Close()
 
+	a.log.Info("Cordoning the node", machine.Name)
+	err = a.cordon(ctx, cluster, machine, scope)
+	if err != nil {
+		return errors.Wrapf(err, "Faild to cordon the node")
```
ncdc:
```suggestion
		return errors.Wrapf(err, "Failed to cordon node for machine %q", machine.Name)
```

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -305,6 +356,59 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 	return nil
 }
 
+func (a *Actuator) cordon(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine, scope *actuators.MachineScope) error {
+	node, err := a.getNodeObject(ctx, cluster, machine, scope)
+	if err != nil {
+		a.log.Info("ERROR: ", err.Error())
```
ncdc: Was this for debugging? Can we remove it?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -269,6 +314,12 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	defer scope.Close()
 
+	a.log.Info("Cordoning the node", machine.Name)
```
ncdc:
```suggestion
	a.log.Info("Cordoning the node", "machine-namespace", machine.Namespace, "machine-name", machine.Name, "cluster-name", cluster.Name)
```

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -305,6 +356,59 @@ func (a *Actuator) Delete(ctx context.Context, cluster *clusterv1.Cluster, machi
 	return nil
 }
 
+func (a *Actuator) cordon(ctx context.Context, cluster *clusterv1.Cluster, machine *clusterv1.Machine, scope *actuators.MachineScope) error {
+	node, err := a.getNodeObject(ctx, cluster, machine, scope)
+	if err != nil {
+		a.log.Info("ERROR: ", err.Error())
+		return err
+	}
+	a.log.Info("cordoning node: ", node.GetName())
```
ncdc:
```suggestion
	a.log.Info("cordoning node: ", "node-name", node.Name)
```

ncdc:
@vincepri retrieving the node object will be easier once #830 is merged, won't it? Because we'll already have the node ref, so we can do a `Get()`?

ncdc:
Or rather, it'll be easier once you deploy a version of CAPI that contains the node ref changes...

vincepri:
That's correct! And once we rebase we'll be able to use the remote package upstream to access the target cluster

sfzylad:
Awesome, thank you for your input. I will do my best to find some time this week to push this issue further. It's been busy time for me recently.

ncdc:
@sfzylad please let us know if there's anything we can do to help!

ncdc:
FYI, we now have remote node references in cluster-api, so we should be able to do node cordoning and draining in cluster-api itself and have it apply to all providers. Michael Gugino is working on it and hopes to have a PR early next week.

ncdc:
@sfzylad thanks for your work on this. We are going to be able to handle cordon & drain generically in cluster-api and have it apply to all providers. We're waiting on https://github.com/kubernetes/kubernetes/pull/80045 to finalize and merge, and then we'll be able to proceed in cluster-api. Therefore, I'm going to close this and hopefully we can see some movement in the other PR soon.

/close

PR #751, opened 2019-05-02
Update Gopkg.lock and cleanup Makefile

detiber:
- Cleans up Gopkg.lock diff after running make targets
- Fixes target name for `create-cluster-ha`
- Leverages bazel for building `binaries-dev` target instead of rebuilding with `go get`

**Release note**:
```release-note
NONE
```


Makefile
```diff
@@ -211,8 +211,7 @@ release-artifacts: ## Build release artifacts
 ## --------------------------------------
 
 .PHONY: binaries-dev
-binaries-dev: ## Builds and installs all development binaries using go get
-	go get -v ./...
+binaries-dev: generate manager clusterawsadm clusterctl
```
chuckha: this installs the binaries to GOPATH?

detiber: Yes: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/970144de689f92f1b7d42c34da87b218f1f01dc4/Makefile#L178-L191

PR #752, opened 2019-05-02
Reduce the number of re-reconciles

chuckha:
This PR minimizes the number of updates performed to only update when necessary.

There were only 2 re-reconciles for each machine object with this patch.

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine_scope.go
+++ b/pkg/cloud/aws/actuators/machine_scope.go
@@ -19,6 +19,7 @@ package actuators
 import (
 	"encoding/json"
 	"fmt"
+	"reflect"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -131,9 +132,14 @@ func (m *MachineScope) Close() {
 		m.Error(err, "failed to encode machine spec")
 		return
 	}
-	status, err := v1alpha1.EncodeMachineStatus(m.MachineStatus)
+	newStatus, err := v1alpha1.EncodeMachineStatus(m.MachineStatus)
 	if err != nil {
 		m.Error(err, "failed to encode machine status")
+		return
+	}
+	oldStatus, err := v1alpha1.MachineStatusFromProviderStatus(m.MachineCopy.Status.ProviderStatus)
+	if err != nil {
+		m.Error(err, "failed to get machine status from provider status")
 		return
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -144,25 +150,33 @@ func (m *MachineScope) Close() {
...
+	// Do not update status if the statuses are the same
+	if reflect.DeepEqual(m.MachineStatus, oldStatus) {
 		return
 	}
 
-	updated.Status.ProviderStatus = status
-	if _, err := m.MachineClient.UpdateStatus(updated); err != nil {
+	m.Machine.Status.ProviderStatus = newStatus
+	if _, err := m.MachineClient.UpdateStatus(m.Machine); err != nil {
 		m.Error(err, "failed to update cluster status")
```
detiber: Not related to the diff, but this should be: `"failed to update machine status"`

pkg/cloud/aws/actuators/machine_scope.go
```diff
@@ -144,25 +150,33 @@ func (m *MachineScope) Close() {
...
 		return
 	}
-
-	m.V(1).Info("Updated machine")
```
detiber: This looks to remove the only positive logging we had for a machine being updated, do we need to log when we've patched the machine or updated the status?

chuckha: I can add this back in. It could be a no-op now though, so a more accurate message might be "scope closed" which I was thinking is not very useful.

Do you think it would be useful to log when we patch and when we update status?

detiber: +1 to logging when we patch or update status, even if it is a higher verbosity level.

I don't see any reason to log closing the scope.

chuckha: I think https://github.com/kubernetes-sigs/controller-runtime/blob/master/TMP-LOGGING.md this post makes a good argument about Verbose logging:

> While it's possible to use higher log levels, it's recommended that you stick with V(1) or V(0)(which is equivalent to not specifyingV), and then filter later based on key-value pairs or messages; different numbers tend to lose meaning easily over time, and you'll be left wondering why particular logs lines are at V(5)instead ofV(7)`.

I'll add patch & status update as a .V(1)

pkg/cloud/aws/actuators/scope.go
```diff
@@ -164,22 +170,32 @@ func (s *Scope) Close() {
...
+			return
+		}
+		s.Logger.V(1).Info("Patching cluster")
+		result, err := s.ClusterClient.Patch(s.Cluster.Name, types.JSONPatchType, pb)
+		if err != nil {
+			s.Error(err, "failed to patch cluster")
+			return
+		}
+		// Keep the resource version updated so the status update can succeed
+		s.Cluster.ResourceVersion = result.ResourceVersion
```
detiber: This seems a bit confusing to just update the resource version, would it be better to update the entire `Machine` object here, or possibly the returned object as a separate variable before doing the status update?

chuckha: This is the minimal amount of updating necessary to make sure the UpdateStatus call succeeds. 

The code gets weird if I'm saving to an alternative variable, it looks like this:

```
var updatedCluster *v1alpha1.Cluster
if len(p) > 0 {
   ... run updates
   updatedCluster, err = Patch()
}

if updatedCluster != nil  {
   if reflect.DeepEqual(updatedCluster, s.ClusterStatus) {
    return
  }
}
 UpdateStatus(updatedClusteR)

```

What is the benefit of updating the entire machine? Especially given we're just on the verge of chucking the whole thing away as the scope closes

detiber: My main concern is around future changes that might break some of the assumptions we have in place, but I'm probably just being overly paranoid.

PR #757, opened 2019-05-02
Simple cache for sessions

vincepri:

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/aws/actuators/scope.go
+++ b/pkg/cloud/aws/actuators/scope.go
@@ -19,6 +19,7 @@ package actuators
 import (
 	"encoding/json"
 	"reflect"
+	"sync"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/session"
@@ -33,6 +34,25 @@ import (
 	client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
 	"sigs.k8s.io/controller-runtime/pkg/patch"
 )
+
+var (
+	sessionCache sync.Map
+)
+
+func sessionForRegion(region string) (*session.Session, error) {
+	s, ok := sessionCache.Load(region)
+	if ok {
+		return s.(*session.Session), nil
+	}
+
+	ns, err := session.NewSession(aws.NewConfig().WithRegion(region))
+	if err != nil {
+		return nil, err
+	}
+
+	sessionCache.Store(region, ns)
+	return ns, nil
+}
 
 // ScopeParams defines the input parameters used to create a new Scope.
 type ScopeParams struct {
@@ -59,7 +79,7 @@ func NewScope(params ScopeParams) (*Scope, error) {
 		return nil, errors.Errorf("failed to load cluster provider status: %v", err)
 	}
 
-	session, err := session.NewSession(aws.NewConfig().WithRegion(clusterConfig.Region))
+	session, err := sessionForRegion(clusterConfig.Region)
 	if err != nil {
 		return nil, errors.Errorf("failed to create aws session: %v", err)
 	}


pkg/cloud/aws/actuators/scope.go
```diff
@@ -34,6 +35,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/patch"
 )
 
+var (
+	sessionLock  sync.Mutex
```
ashish-amarnath:
```suggestion
	cacheLock  sync.Mutex
```

pkg/cloud/aws/actuators/scope.go
```diff
@@ -59,17 +65,25 @@ func NewScope(params ScopeParams) (*Scope, error) {
 		return nil, errors.Errorf("failed to load cluster provider status: %v", err)
 	}
 
-	session, err := session.NewSession(aws.NewConfig().WithRegion(clusterConfig.Region))
-	if err != nil {
-		return nil, errors.Errorf("failed to create aws session: %v", err)
+	sessionLock.Lock()
+	defer sessionLock.Unlock()
```
ashish-amarnath:
```suggestion
	defer cacheLock.Unlock()
```

pkg/cloud/aws/actuators/scope.go
```diff
@@ -59,17 +65,25 @@ func NewScope(params ScopeParams) (*Scope, error) {
 		return nil, errors.Errorf("failed to load cluster provider status: %v", err)
 	}
 
-	session, err := session.NewSession(aws.NewConfig().WithRegion(clusterConfig.Region))
-	if err != nil {
-		return nil, errors.Errorf("failed to create aws session: %v", err)
+	sessionLock.Lock()
```
ashish-amarnath:
```suggestion
	cacheLock.Lock()
```

pkg/cloud/aws/actuators/scope.go
```diff
@@ -59,17 +65,25 @@ func NewScope(params ScopeParams) (*Scope, error) {
...
+	defer sessionLock.Unlock()
+
+	var localSession *session.Session
+	if cachedSession, ok := sessionCache[clusterConfig.Region]; !ok {
+		localSession, err = session.NewSession(aws.NewConfig().WithRegion(clusterConfig.Region))
+		if err != nil {
+			return nil, errors.Errorf("failed to create aws session: %v", err)
+		}
```
ashish-amarnath:
```suggestion
		}
		sessionCacheLock[clusterConfig.Region] = localSession
```

pkg/cloud/aws/actuators/scope.go
```diff
@@ -59,17 +65,25 @@ func NewScope(params ScopeParams) (*Scope, error) {
...
-	if err != nil {
-		return nil, errors.Errorf("failed to create aws session: %v", err)
+	sessionLock.Lock()
+	defer sessionLock.Unlock()
+
+	var localSession *session.Session
+	if cachedSession, ok := sessionCache[clusterConfig.Region]; !ok {
+		localSession, err = session.NewSession(aws.NewConfig().WithRegion(clusterConfig.Region))
+		if err != nil {
+			return nil, errors.Errorf("failed to create aws session: %v", err)
```
ashish-amarnath: When will we remove entries from the cache?

vincepri:
@ashish-amarnath This PR is just for tracking purposes, not meant to be merged in yet until we have a final solution / plan

pkg/cloud/aws/actuators/scope.go
ncdc: AFAIK that should not be necessary

pkg/cloud/aws/actuators/scope.go
```diff
@@ -34,6 +35,11 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/patch"
 )
 
+var (
+	sessionLock  sync.Mutex
+	sessionCache map[string]*session.Session
```
ncdc: Where do you initialize this map?

pkg/cloud/aws/actuators/scope.go
```diff
@@ -59,17 +65,25 @@ func NewScope(params ScopeParams) (*Scope, error) {
 		return nil, errors.Errorf("failed to load cluster provider status: %v", err)
 	}
 
-	session, err := session.NewSession(aws.NewConfig().WithRegion(clusterConfig.Region))
-	if err != nil {
-		return nil, errors.Errorf("failed to create aws session: %v", err)
+	sessionLock.Lock()
+	defer sessionLock.Unlock()
+
+	var localSession *session.Session
```
ncdc: It might be cleaner to move this to a function - something like `sessionForRegion(region)`.

chuckha:
this looks like it fits the use case of `sync.Map`! https://golang.org/pkg/sync/#Map a cache that only grows that has a key that's written once and read many times. Could avoid manual locking/unlocking but add type checking. If it's wrapped in a function though it's not terrible

pkg/cloud/aws/actuators/scope.go
detiber: If we do have some type of ttl, then we can avoid having to scale/redeploy the controller StatefulSet to pick up a change to the credentials (if mounted from a Secret).

ncdc: good point! no harm in having a ttl

vincepri: Do we need this for v1 of the session cache?

vincepri:
@ashish-amarnath @ncdc @chuckha ptal, updated to use sync.Map, without a TTL for now

pkg/cloud/aws/actuators/scope.go
```diff
@@ -34,6 +35,29 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/patch"
 )
 
+var (
+	sessionCache sync.Map
+)
+
+func sessionForConfig(config *aws.Config) (*session.Session, error) {
```
detiber: Should this be `sessionForRegion` and only take a `Region` parameter? The current naming and param seems misleading, since if I called it with a different Config, but it had the same Region as an already cached Session, I would get a Session that doesn't necessarily match my config.

pkg/cloud/aws/actuators/scope.go
```diff
@@ -34,6 +35,29 @@ import (
...
+	if ok {
+		return s.(*session.Session), nil
+	}
+
+	ns, err := session.NewSession(config)
+	if err != nil {
+		return nil, err
+	}
+
+	s, ok = sessionCache.LoadOrStore(config.Region, ns)
```
detiber: Why not just `sessionCache.Store(config.Region, ns)` here? In the racy case we would be discarding a session anyway, so why not just discard the older session from the cache?

pkg/cloud/aws/actuators/scope.go
detiber: No, but we should create a tracking issue.

pkg/cloud/aws/actuators/scope.go
vincepri: I went back and forth on it, I think down the line we might want to change the hashing function from other config params

ncdc: Let's start with what we're actively using now - sessionForRegion. We can change later, but only if needed.

pkg/cloud/aws/actuators/scope.go
vincepri: Yeah maybe that's fine, I was trying to avoid having multiple sessions doing different calls to AWS. Truthfully without a lock on the function, that's not guaranteed, wdyt?

ncdc: If you want the guarantee, you need the mutex, yes.

vincepri: Fixed this as well

ncdc:
LGTM. @detiber?

PR #760, opened 2019-05-03
Fix up //test/e2e:manifests

detiber:
Fixes the recently introduced e2e test failure when building the test manifests

**Release note**:
```release-note
NONE
```


chuckha:
nice! ty

/lgtm

PR #762, opened 2019-05-03
Few release tool improvements

chuckha:

This PR makes publishing releases a little bit less scary. Or more depending on your perspective.

If you mess up a release or need to add some more commits to a branch before releasing this will force update the tag and force push the tag instead of making you do it by hand.
```release-note
NONE
```

### DIFF
--- a/cmd/release/main.go
+++ b/cmd/release/main.go
@@ -17,11 +17,13 @@ limitations under the License.
 package main
 
 import (
+	"bufio"
 	"flag"
 	"fmt"
 	"os"
 	"os/exec"
 	"path"
+	"strings"
 )
 
 /*
@@ -193,7 +195,6 @@ Examples:
 		./release -remote YOUR_FORK -user YOUR_GITHUB_USER_NAME -version v1.1.1`)
 }
 
-// TODO: split up the config file?
 // config defines all configuration needed to get a release going
 type config struct {
 	// version is the version being released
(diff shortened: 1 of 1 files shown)


cmd/release/main.go
```diff
@@ -314,25 +318,33 @@ type git struct {
...
+		return "", err
+	}
+	return string(out), err
 }
 
 func (g git) pushTag(version string) error {
-	// TODO(chuckha): this shouldn't exit if it fails because the tag already
-	cmd := exec.Command("git", "push", g.remote, version)
+	g.logger.Infof("You are force pushing a tag: %v\n", version)
+	cmd := exec.Command("git", "push", "-f", g.remote, version)
```
detiber: Seems a bit scary to me to default to a force push of the tag. It might be better to try and push without force and then prompt the user before doing a force push of the tag if the first push fails.

cmd/release/main.go
```diff
@@ -314,25 +318,33 @@ type git struct {
...
 
 func (g git) tag(version string) error {
-	// if the tag exists then return nil; if it doesn't create it
-	cmd := exec.Command("git", "rev-parse", "--quiet", "--verify", version)
-	_, err := cmd.CombinedOutput()
+	//TODO: this could probably use some work
+	if oldRev, err := g.tagRevision(version); err != nil {
+		g.logger.Info("non critical error, failed to get tag revision: %v", err)
+	} else {
+		g.logger.Infof("You are forcing the update of a tag reference from %v to the current commit", oldRev)
```
detiber: We should probably prompt before attempting to force overwriting of an existing tag.

chuckha:
I still plan to fix this

chuckha:
@detiber this now asks for confirmation before any force action. it will only call force actions if necessary.

PR #768, opened 2019-05-08
Set cluster status api endpoints. (# 764)

### LINKED ISSUE #764: Ensure Cluster.Status.APIEndpoints is set regardless of how cluster is created
/kind bug

Create a cluster using any mechanism **other than** `clusterctl create cluster`. The cluster's `status.apiEndpoints` field is not filled in. (It is filled in when you do `clusterctl create cluster`.)

Regardless of how you create a cluster, its apiEndpoints must be populated.

We should add code to the cluster actuator to reconcile the endpoints.

**Environment:**

- Cluster-api-provider-aws version: commit 7e32f349f6b4bcc92c416af487e8884467edef78

kaushik229:
* Make sure cluster reconciler set the api endpoints if it is not set
or changed.

Fixes #764 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/aws/actuators/scope.go
+++ b/pkg/cloud/aws/actuators/scope.go
@@ -39,6 +39,8 @@ var (
 	sessionCache sync.Map
 )
 
+const apiEndpointPort = 6443
+
 func sessionForRegion(region string) (*session.Session, error) {
 	s, ok := sessionCache.Load(region)
 	if ok {
@@ -175,11 +177,6 @@ func (s *Scope) Close() {
 		s.Error(err, "failed encoding cluster status")
 		return
 	}
-	oldStatus, err := v1alpha1.ClusterStatusFromProviderStatus(s.ClusterCopy.Status.ProviderStatus)
-	if err != nil {
-		s.Error(err, "failed to get provider status from status")
-		return
-	}
 
 	s.Cluster.Spec.ProviderSpec.Value = ext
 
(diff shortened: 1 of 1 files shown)


pkg/cloud/aws/actuators/scope.go
```diff
@@ -207,15 +209,28 @@ func (s *Scope) Close() {
 		s.Cluster.ResourceVersion = result.ResourceVersion
 	}
 
+	// Check if API endpoints is not set or has changed.
+	if s.Cluster.Status.APIEndpoints == nil || s.Cluster.Status.APIEndpoints[0].Host != s.ClusterStatus.Network.APIServerELB.DNSName {
```
chuckha: The problem can be summarized: 

The cluster status is only being updated if the cluster *ProviderStatus* has changed, but the status is more than *ProviderStatus*.

The main changes I'd like to see here are:

1. Instead of reflect.DeepEqual on the providerstatus, do a deep equal between `s.Cluster.Status` and `s.ClusterCopy.Status`. => delete the oldStatus var
2. Don't do an additional UpdateStatus call => remove the logger and the additional UpdateStatus call this PR adds
3. move line 230 above the logic this PR adds regarding the check for apiendpoints having changed
4. change the reflect.DeepEqual to compare the s.Cluster.Status and s.ClusterCopy.Status

pkg/cloud/aws/actuators/scope.go
```diff
@@ -39,6 +39,8 @@ var (
 	sessionCache sync.Map
 )
 
+const apiEndpointPort = 443
```
detiber: I believe this should probably be 6443: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1f9a31c0a17d3637a570e43171c9020c9a06379e/pkg/cloud/aws/services/elb/loadbalancer.go#L149

detiber: On a tangential note, Instead of hardcoding the value here and in the load balancer creation we really should take into account the ClusterConfiguration: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go#L60

That is something we should treat as a followup item, though.

pkg/cloud/aws/actuators/scope.go
kaushik229: done

pkg/cloud/aws/actuators/scope.go
kaushik229: Port change done

kaushik229:
WIP. Please do not merge

detiber:
@kaushik229 I modified the PR title to prepend it with `[WIP]` this will prevent the bot from merging until it is removed.

chuckha:
the logic for reflect.DeepEquals is reversed. Before it returned when it was true, now it will only update when the two are equal which is the opposite of what we want

that one took a sec to figure out 😂

kaushik229:
I already found that and put WIP.
Will send fix in 10 mins after some testing.

chuckha:
beautiful :japanese_castle: :rat: :boar:
 
/approve

PR #774, opened 2019-05-14
Fix security group reconcile

vikaschoudhary16:
Currently security groups reconciling works only for the creation of new subnets. For existing subnets case, `ResourceID` is not being set in buildParams which results in the aws api failure. Overall it leads to network reconciliation failure.

```
I0513 13:45:02.870613       1 securitygroups.go:48] [cluster-actuator]/cluster.k8s.io/v1alpha1/default/test-vichoudh "level"=2 "msg"="Reconciling security groups"  
E0513 13:45:03.507470       1 controller.go:134] Error reconciling cluster object test-vichoudh; failed to reconcile network for cluster "test-vichoudh": failed to ensure tags on security group "sg-02e51493b84d53085": failed to tag resource "" in cluster "test-vichoudh": InvalidParameterValue: Value ( null ) for parameter resourceId is invalid. Null/empty value for resourceId is invalid
    status code: 400, request id: 2002e650-02d8-4749-8491-e71616ab0cf1
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/securitygroups.go
+++ b/pkg/cloud/aws/services/ec2/securitygroups.go
@@ -88,7 +88,7 @@ func (s *Service) reconcileSecurityGroups() error {
 		// Make sure tags are up to date.
 		err := tags.Ensure(existing.Tags, &tags.ApplyParams{
 			EC2Client:   s.scope.EC2,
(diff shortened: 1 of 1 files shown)


ncdc:
Do we want CAPA to modify preexisting resources?

ncdc:
cc @rosskukulinski

PR #779, opened 2019-05-16
Fix controlplane machine race

### LINKED ISSUE #763: isNodeJoin should be thread safe
/kind feature

`isNodeJoin` method in the machine actuator is not thread safe. It is possible for two machines to race to be the first controlplane machine and can result in 2 machines, incorrectly, trying to run `kubeadm init`. This can land the cluster in a bad state.

Also, kubedm, until `v1.15`, doesn't support parallel `init` and `join`.

For these reasons,  controlplane machines will have to be added one at a time

Follow commentary here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/745#discussion_r280096104

**Environment:**

ncdc:
Fix race condition when creating a control plane from scratch. If you
try to create more than one control plane machine at the same time, make
sure that only one machine ever runs 'kubeadm init' and the others wait
for the apiserver to be operational before trying to create bootstrap
tokens and join the cluster.

Fixes #763

This adds 2 new annotations to the cluster:
- `aws.cluster.sigs.k8s.io/infrastructure-ready`
- `aws.cluster.sigs.k8s.io/control-plane-ready`

The cluster actuator sets `aws.cluster.sigs.k8s.io/infrastructure-ready` after it has successfully reconciled certs, network, bastion, load balancer.

The machine actuator checks the cluster's `aws.cluster.sigs.k8s.io/infrastructure-ready` annotation at the very beginning of `Create`, adding the machine back to the queue if the infra is not ready.

If the cluster doesn't have `aws.cluster.sigs.k8s.io/control-plane-ready`, the cluster actuator checks all the control plane machines. If it finds one with a non-nil node ref, it sets this annotation to true.

The machine actuator defaults to assuming all new machines are node joins, because only the first control plane machine is an init. If the cluster doesn't have the `aws.cluster.sigs.k8s.io/control-plane-ready` annotation, and the new machine is a control plane machine, the machine actuator checks for the existence of a special "control plane lock" configmap. If it exists, the machine gets readded to the queue because it's an additional control plane machine and the first machine isn't done initializing. Otherwise, the machine actuator tries to create this configmap lock. If it succeeds, the machine actuator is allowed to proceed with the actual VM creation.

**Release note**:

```release-note
Fix a race condition when creating a new cluster with multiple control plane machines at the same time.
```


ncdc:
Have not tested or finished coding 100%, but wanted to publish for initial review.

ncdc:
Aware of some logic issues - working on a fix

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -167,20 +119,84 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+
+	// Except for the very first control plane machine, everything will be a join, so let's assume we're joining.
+	isNodeJoin := true
+
+	// TODO make sure this name isn't longer than the max allowed length
+	controlPlaneSyncConfigMapName := fmt.Sprintf("%s-controlplane", cluster.Name)
+
+	controlPlaneConfigMap, err := a.coreClient.ConfigMaps(cluster.Namespace).Get(controlPlaneSyncConfigMapName, metav1.GetOptions{})
+	if apierrors.IsNotFound(err) {
+		if machine.Spec.Versions.ControlPlane == "" {
```
sethp-nr: Is this the canonical way to tell if it's a control plane machine? I thought CAP* relied on `ObjectMeta.Labels["set"]` to distinguish.

ncdc: Either is probably correct. The implementation in CAPA for `GetControlPlaneMachines` uses this method.

ncdc:
@ashish-amarnath FYI, working on a fix for the control plane node join race condition. This takes a different approach to what you had started. Would love to get your feedback on it. Thanks!

sethp-nr:
I'm wondering about using a ConfigMap vs. part of the Cluster object for this problem. I know ConfigMaps are a well-established pattern for locking in kubernetes, but it feels more natural to me for there to be a field (or annotation) on the cluster that says "this cluster has been initialized, any controlplane adds are now joins".

My main concerns are:
1) will the configmap survive a pivot?
2)the CM feels "external" to the cluster to me in a way the Cluster object doesn't, which leads me to believe future-me is more likely to accidentally delete that resource and thereby get my cluster into a bad situation again

ncdc:
At this point this is mostly a WIP/POC. You bring up a very good point about the ConfigMap not necessarily surviving the pivot unless we code for it. I considered using an annotation on the Cluster, but was trying to avoid an unnecessary update to the Cluster, but that's probably a better approach.

I'll switch to annotating the Cluster.

ncdc:
I just tested these changes and I was able to `k apply` a `Cluster` and a `MachineList` containing 3 control plane machines, and after several minutes, I had a 3-node control plane!

I expect there are improvements we could make here, including the switch from a ConfigMap to an annotation on the Cluster. I plan to spend some more time on this, but I need to step away for the rest of the day. I would appreciate any additional review comments y'all might have in the meantime. Thanks!

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -167,32 +119,110 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+		// This is the first control plane machine, so it's not a join.
+		isNodeJoin = false
+	} else if err != nil {
+		return errors.Wrap(err, "error getting control plane sync config map")
+	} else {
+		// If this is the first control plane machine, there is a good chance that the initial attempt to create the
+		// machine failed because the AWS network wasn't ready yet. We need to therefore check if this is the first
+		// control plane machine, and if so, set isNodeJoin to false so it can be created correctly.
+
+		if controlPlaneConfigMap.Data["firstMachine"] == machine.Name {
```
ashish-amarnath: Do we care if the data stored in this configmap becomes stale? This can happen if we delete/ replace the first controlplane machine. Maybe we need to cleanup the config map on delete.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -167,32 +119,110 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+			return errors.Wrap(err, "error creating control plane sync config map")
+		}
+
+		// This is the first control plane machine, so it's not a join.
+		isNodeJoin = false
+	} else if err != nil {
+		return errors.Wrap(err, "error getting control plane sync config map")
+	} else {
+		// If this is the first control plane machine, there is a good chance that the initial attempt to create the
+		// machine failed because the AWS network wasn't ready yet. We need to therefore check if this is the first
```
ashish-amarnath: > because the AWS network wasn't ready yet
Will this logic also help if we delete a controlplane machine and recreate another one w/ the same name?

ncdc:
I don’t believe it matters. Once we get the first control plane node operational, the contents of the data are no longer important. What does matter is that we can check that the data exist so we know it’s a join and not the initial control plane member.

On May 16, 2019, at 2:46 PM, Ashish Amarnath <notifications@github.com<mailto:notifications@github.com>> wrote:


@ashish-amarnath commented on this pull request.

________________________________

In pkg/cloud/aws/actuators/machine/actuator.go<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fkubernetes-sigs%2Fcluster-api-provider-aws%2Fpull%2F779%23discussion_r284848218&data=02%7C01%7Cgoldsteina%40vmware.com%7C00b8801da5d4456bdac708d6da2ed7bc%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C636936292051353052&sdata=%2BUkLoUm545tncc7kwJz8Qq6YTt%2FxCtpyMLbt4%2BmnCKE%3D&reserved=0>:

> +
+               _, err = a.coreClient.ConfigMaps(cluster.Namespace).Create(controlPlaneConfigMap)
+               if err != nil && !apierrors.IsAlreadyExists(err) {
+                       return errors.Wrap(err, "error creating control plane sync config map")
+               }
+
+               // This is the first control plane machine, so it's not a join.
+               isNodeJoin = false
+       } else if err != nil {
+               return errors.Wrap(err, "error getting control plane sync config map")
+       } else {
+               // If this is the first control plane machine, there is a good chance that the initial attempt to create the
+               // machine failed because the AWS network wasn't ready yet. We need to therefore check if this is the first
+               // control plane machine, and if so, set isNodeJoin to false so it can be created correctly.
+
+               if controlPlaneConfigMap.Data["firstMachine"] == machine.Name {


Do we care if the data stored in this configmap becomes stale? This can happen if we delete/ replace the first controlplane machine. Maybe we need to cleanup the config map on delete.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -167,32 +119,110 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 
 	controlPlaneMachines := GetControlPlaneMachines(clusterMachines)
+	if len(controlPlaneMachines) == 0 {
+		return errors.New("unable to proceed with machine creation until there is at least 1 control plane machine")
+	}
 
-	isNodeJoin, err := a.isNodeJoin(scope, controlPlaneMachines)
-	if err != nil {
-		return errors.Wrapf(err, "failed to determine whether machine %q should join cluster %q", machine.Name, cluster.Name)
+	/*
```
ashish-amarnath: Can we add unit tests for this? It will protect us from future violations of today's assumptions 😊

ncdc:
That’s an edge case I hadn’t considered and one that needs to be handled. If the control plane is not ready yet and someone deletes this first machine, we need to make sure to clear out the data so a subsequent machine can be considered the initial member.

On May 16, 2019, at 2:48 PM, Ashish Amarnath <notifications@github.com<mailto:notifications@github.com>> wrote:


@ashish-amarnath commented on this pull request.

________________________________

In pkg/cloud/aws/actuators/machine/actuator.go<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fkubernetes-sigs%2Fcluster-api-provider-aws%2Fpull%2F779%23discussion_r284848729&data=02%7C01%7Cgoldsteina%40vmware.com%7C14e8d06ad6404740cdac08d6da2f0680%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C636936292834841785&sdata=ZcV%2FCGWg203yZyAfwMS%2BKzMU06mgyaOehgdEQsqYLSU%3D&reserved=0>:

> +                             "firstMachine": machine.Name,
+                       },
+               }
+
+               _, err = a.coreClient.ConfigMaps(cluster.Namespace).Create(controlPlaneConfigMap)
+               if err != nil && !apierrors.IsAlreadyExists(err) {
+                       return errors.Wrap(err, "error creating control plane sync config map")
+               }
+
+               // This is the first control plane machine, so it's not a join.
+               isNodeJoin = false
+       } else if err != nil {
+               return errors.Wrap(err, "error getting control plane sync config map")
+       } else {
+               // If this is the first control plane machine, there is a good chance that the initial attempt to create the
+               // machine failed because the AWS network wasn't ready yet. We need to therefore check if this is the first


because the AWS network wasn't ready yet
Will this logic also help if we delete a controlplane machine and recreate another one w/ the same name?

ncdc:
Of course! I only had a limited amount of time today and I wanted this up for review first.

On May 16, 2019, at 2:50 PM, Ashish Amarnath <notifications@github.com<mailto:notifications@github.com>> wrote:


@ashish-amarnath commented on this pull request.

________________________________

In pkg/cloud/aws/actuators/machine/actuator.go<https://nam04.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fkubernetes-sigs%2Fcluster-api-provider-aws%2Fpull%2F779%23discussion_r284849789&data=02%7C01%7Cgoldsteina%40vmware.com%7C47002bce323d41c67d3008d6da2f6b93%7Cb39138ca3cee4b4aa4d6cd83d9dd62f0%7C0%7C0%7C636936294531916255&sdata=JzBPaduql767gvHATZXRk%2BPW2N7QToI2u5XivG1D0Ng%3D&reserved=0>:

>
-       isNodeJoin, err := a.isNodeJoin(scope, controlPlaneMachines)
-       if err != nil {
-               return errors.Wrapf(err, "failed to determine whether machine %q should join cluster %q", machine.Name, cluster.Name)
+       /*


Can we add unit tests for this? It will protect us from future violations of today's assumptions 😊

pkg/cloud/aws/actuators/machine/actuator.go
ashish-amarnath: Maybe refactor this code into a method that we can write tests against :)

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -167,32 +119,110 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+					},
+				},
+			},
+			Data: map[string]string{
+				"firstMachine": machine.Name,
+			},
+		}
+
+		_, err = a.coreClient.ConfigMaps(cluster.Namespace).Create(controlPlaneConfigMap)
+		if err != nil && !apierrors.IsAlreadyExists(err) {
```
sethp-nr: I'm not sure, but I think a conflict indicates that we lost the race, which suggests to me we want to move into the "wait for API availability" mode?

ashish-amarnath: The error returns a retry-able error to the controller and this machine will be reconciled in the next reconciliation loop in which the condition on L153 will be false and we'll wait for the api server to be available.
@ncdc, correct me if I am wrong 😊

sethp-nr: Oh, neat! I think we just want to remove the `&& !apierrors.IsAlreadyExists(err)` half of this condition then (or maybe return a different, more specific error in that case).

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -167,32 +119,110 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 	controlPlaneMachines := GetControlPlaneMachines(clusterMachines)
+	if len(controlPlaneMachines) == 0 {
+		return errors.New("unable to proceed with machine creation until there is at least 1 control plane machine")
+	}
 
-	isNodeJoin, err := a.isNodeJoin(scope, controlPlaneMachines)
-	if err != nil {
-		return errors.Wrapf(err, "failed to determine whether machine %q should join cluster %q", machine.Name, cluster.Name)
+	/*
+	If new machine is a control plane machine, we may be in the following states:
```
chuckha: This is almost like a lease, I haven't thought it all the way through but I'm wondering if we can reuse v1coordination.Lease

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: Yes, I do think that would work

ncdc:
I am prepping for KubeCon and most likely won't be able to get back to this until the week after next, FYI.

pkg/cloud/aws/actuators/machine/actuator.go
sethp-nr: That lease stuff looks really cool! I think there's a fairly similar shape here for sure.

I'd still be concerned about pivoting a lease, and I'm not sure how to set the timeout – I think it's preferable to me to stop doing anything than it is to end up with a split brain cluster.

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: (copying my email comments that didn't slot in the right place)

Of course! I only had a limited amount of time today and I wanted this up for review first.

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: That’s an edge case I hadn’t considered and one that needs to be handled. If the control plane is not ready yet and someone deletes this first machine, we need to make sure to clear out the data so a subsequent machine can be considered the initial member.

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: (copying my email replies that didn't get slotted in the right place):

~~I don’t believe it matters. Once we get the first control plane node operational, the contents of the data are no longer important. What does matter is that we can check that the data exist so we know it’s a join and not the initial control plane member.~~

That’s an edge case I hadn’t considered and one that needs to be handled. If the control plane is not ready yet and someone deletes this first machine, we need to make sure to clear out the data so a subsequent machine can be considered the initial member.

akutz: In Cluster.Reconcile you can always check to see if the ID of the lock holder still exists as a machine, and if not, force release the lock.

ncdc:
For those following along, I am picking this back up. After talking with @akutz, we think it will be less racy if we continue to use a configmap for this. And we have a way to handle the pivot too. Stay tuned!

pkg/cloud/aws/actuators/machine/actuator.go
akutz: Re: Lease - a Slack conversation between myself and @ncdc:

> Andrew Kutz  [3 hours ago]
> FWIW @andygoldstein, `coordination.Lease` is `v1beta1` in 1.13 and GA in 1.14. It’s likely fine to use that since you can support 1.13 with the beta type.
> 
> Andy Goldstein  [3 hours ago]
> unless @detiberusj you think we can use this?
> 
> REMOVED
> 
> Andrew Kutz  [3 hours ago]
> But still, maybe you can figure out how to use the `Lease`
> 
> Andrew Kutz  [3 hours ago]
> The only examples I can find are for leader election
> 
> Andy Goldstein  [3 hours ago]
> there’s a package for it, 1 sec
> 
> Andrew Kutz  [3 hours ago]
> It doesn’t seem much more than a formalized version of what we’re doing. They just make things first-class values instead of shoving an encoded struct into a field
> 
> Andrew Kutz  [3 hours ago]
> Yeah, I dug into that pretty deep last weekend
> 
> Andrew Kutz  [3 hours ago]
> https://godoc.org/k8s.io/kubernetes/pkg/apis/coordination
> 
> Andrew Kutz  [3 hours ago]
> Again, they do have a `v1beta1` type as well that can support 1.13
> 
> Andrew Kutz  [3 hours ago]
> I like the config object because it’s not tied to the K8s version
> 
> Andrew Kutz  [3 hours ago]
> And it’s super simple to do
> 
> Andy Goldstein  [3 hours ago]
> yeah, i think that’s best (configmap). i’ll update so the cluster creates it
> 
> Andy Goldstein  [3 hours ago]
> and machines gate on it
> 
> Andrew Kutz  [3 hours ago]
> Anyway, the LeaseSpec doesn’t seem to provide the framework for using the lease. It just provides a first-class schema for defining a lease holder
> 
> Andy Goldstein  [3 hours ago]
> yea
> 
> Andrew Kutz  [3 hours ago]
> I wanted the actual framework for using the lease.
> 
> Andy Goldstein  [3 hours ago]
> https://github.com/kubernetes/client-go/blob/master/tools/leaderelection/resourcelock/interface.go
> tools/leaderelection/resourcelock/interface.go
> 
> Andy Goldstein  [3 hours ago]
> that?
> 
> Andrew Kutz  [3 hours ago]
> Yeah, but that’s tied to leader election, and I guess that’s kind of what an exclusive lock is? But I was hoping for an implementation of `sync.Locker` to keep things super simple
> 
> Andy Goldstein  [3 hours ago]
> i thought that was a generic locking interface?
> 
> Andrew Kutz  [3 hours ago]
> `sync.Locker`? It is
> 
> 
> Andrew Kutz  [3 hours ago]
> But the linked leader election package doesn’t do anything I see that maps to that. It’s more of a means of managing the leases?
> 
> Andy Goldstein  [3 hours ago]
> oh ok i see what you’re saying
>
> Andrew Kutz  [3 hours ago]
Basically it’s not much more than what we’re doing with config maps, just tied to the leases
>
> Andrew Kutz  [3 hours ago]
> It’s still just CRUD based

pkg/cloud/aws/actuators/machine/actuator.go
akutz: So had a long discussion about this with @detiber a while back - https://kubernetes.slack.com/archives/C8TSNPY4T/p1560807049390100.

Basically `set` isn't part of CAPI, and there's nowhere that CAPA *uses* `set` that couldn't be the formal way of checking via the version string. In fact, there's a CAPI function, [`IsControlPlane`](https://github.com/kubernetes-sigs/cluster-api/tree/master/pkg/util/util.go#L134-L137), that does just that.

In other words, wherever CAPA uses `set`, it should likely switch to checking the Versions information.

AndyGoldstein:
Hi guys... re

> Andrew Kutz [3 hours ago]
> FWIW @AndyGoldstein, coordination.Lease is v1beta1 in 1.13 and GA in 1.14. It’s likely fine to use that since you can support 1.13 with the beta type.
> 
> Andy Goldstein [3 hours ago]
> unless @detiberusj you think we can use this?
> 
> REMOVED
> 
> Andrew Kutz [3 hours ago]
> But still, maybe you can figure out how to use the Lease
> 
> Andrew Kutz [3 hours ago]
> The only examples I can find are for leader election
> 
> Andy Goldstein [3 hours ago]
> there’s a package for it, 1 sec

etc. This is not me. Not sure how you got my name on it - maybe there's a different Andy Goldstein on Slack, with a different user name here..

- Andy

ncdc:
@AndyGoldstein it's a mismatch between Slack and GitHub user IDs. Sorry for the noise!

AndyGoldstein:
Yeah - what I figured. No problem - thanx.

- Andy

akutz:
Hi @ncdc,

I tried to get away from annotations entirely in favor of checking backend state where possible, but I could not avoid the need for at least one annotation on the cluster thanks to the pivot (https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/387/commits/843d8b2447e7f58b1d1f7c724c865d38b8aa5efa). 

Tha's why I wonder if `aws.cluster-api.sigs.k8s.io/infrastructure-ready` provides value beyond the failed dependency infrastructure already in place in CAPA. Instead of validating the existence of the resources, this annotation asserts their presence. Since CAPA already has the concept of failed dependencies, this change even necessary? And if so, does it need to be part of this PR?

ncdc:
@akutz I don't think it's _strictly_ necessary to add the annotations in this PR. However, it makes the logic significantly easier. If we don't use these annotations, then we'd have to allow the machine actuator to attempt to create an ec2 instance when it's the first control plane, even if the infrastructure isn't ready. When that happens, the actuator would see the failed dependencies and requeue the machine. The next time the machine is reconciled, Create is called again, but this time, we already have the control plane configmap lock. However, there are 0 machines that are in flight, so we need to allow a control plane machine to proceed to ec2 instance creation, but it's much harder to determine which one we should let go through. Does that make sense?

akutz:
> Does that make sense?

Hi @ncdc,

Completely. Thank you for taking the time to explain things. I think my initial response was just more of how our providers differ ever so slightly. For example, this is CAPV's Machine actuator's `Create` call:

```golang
// Create creates a new machine.
func (a *Actuator) Create(
	parentCtx goctx.Context,
	cluster *clusterv1.Cluster,
	machine *clusterv1.Machine) (opErr error) {

	ctx, err := context.NewMachineContext(
		&context.MachineContextParams{
			ClusterContextParams: context.ClusterContextParams{
				Context:    parentCtx,
				Cluster:    cluster,
				Client:     a.client,
				CoreClient: a.coreClient,
				Logger:     klogr.New().WithName("[machine-actuator]"),
			},
			Machine: machine,
		})
	if err != nil {
		return err
	}

	if _, ok := machine.Annotations[constants.MaintenanceAnnotationLabel]; ok {
		ctx.Logger.V(4).Info("skipping operations on machine", "reason", "annotation", "annotation-key", constants.MaintenanceAnnotationLabel)
		return &clusterErr.RequeueAfterError{RequeueAfter: constants.DefaultRequeue}
	}

	defer func() {
		opErr = actuators.PatchAndHandleError(ctx, "Create", opErr)
	}()

	if err := a.reconcilePKI(ctx); err != nil {
		return err
	}

	return a.doInitOrJoin(ctx)
}
```

Note that we don't get to `doInitOrJoin` until after `reconcilePKI` is invoked. That will return a `RequeueAfter` error if the cluster object doesn't yet have the PKI toolchain initialized. 

So basically we're doing the same thing, just handling it differently.

Truth be told, I like your approach. I may steal it. My other concern is specific to the things you're no longer checking each time, but then I realized you're not removing those checks, just gating them behind this initial check. Since the initial check, the annotation, is subject to reconciliation, even if it is in error, a later check will prevent an err'd state. And as you pointed out, this is handled already through the cluster reconciliation.

So again, we're doing the same thing, but CAPV's dependency checks are early enough in the actuator code that we don't necessarily need the infrastructure ready annotation. Still, having one would allow us to move the dependency checks closer to when they matter. And if we did, because of the config map, we'd absolutely need the infrastructure annotation for the same reason you do -- we don't want to claim we're initializing the control plane before we **can**, or then we start to have to check who owns the config map to reclaim ownership of the lock.

Thank you for taking the time to respond :)

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -186,13 +201,6 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	i, err := ec2svc.CreateOrGetMachine(scope, bootstrapToken)
 	if err != nil {
-		if awserrors.IsFailedDependency(errors.Cause(err)) {
-			a.log.Error(err, "network not ready to launch instances yet")
-			return &controllerError.RequeueAfterError{
-				RequeueAfter: time.Minute,
-			}
-		}
-
```
sethp-nr: I think we want to check `isNodeJoin` here and clean up the config map if it's false.

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -122,9 +132,56 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+		machines, err := a.client.Machines(cluster.Namespace).List(actuators.ListOptionsForCluster(cluster.Name))
+		if err != nil {
+			return errors.Wrapf(err, "failed to list machines for cluster %q", cluster.Name)
+		}
+
+		controlPlaneMachines := machine.GetControlPlaneMachines(machines)
+
+		machineReady := false
+		for _, machine := range controlPlaneMachines {
+			if machine.Status.NodeRef != nil {
```
sethp-nr: I can't think of another way to do this check, but IIRC the node controller doesn't work for "management cluster" style topologies. It seems like there'd be a workaround in manually flagging the control plane as "ready", should we  document and/or tool that flow?

detiber: Since we know we have a load balancer, and the first api endpoint is that load balancer, and we know we have credentials for the cluster, could we not just query the /healthz endpoint for the api endpoint?

ncdc: The NodeRef controller works for all clusters, management or otherwise. This is new in CAPI v0.1.4 and CAPA has been updated to work with it.

pkg/apis/awsprovider/v1alpha1/types.go
```diff
@@ -442,3 +442,9 @@ type Instance struct {
 	// The tags associated with the instance.
 	Tags map[string]string `json:"tags,omitempty"`
 }
+
+// TODO move somewhere else?
```
vincepri: Usually types.go is a fine place for these annotations

pkg/apis/awsprovider/v1alpha1/types.go
```diff
@@ -442,3 +442,9 @@ type Instance struct {
...
+const (
+	AnnotationClusterInfrastructureReady = "aws.cluster-api.sigs.k8s.io/infrastructure-ready"
```
vincepri: Should we follow the api group name? `aws.cluster-api.sigs.k8s.io` -> `aws.cluster.sigs.k8s.io` ?

ncdc:
@sethp-nr thanks for your comments!

> I think all of the first cases are covered by checking the error from CreateOrGetMachine and cleaning up the lock in the "init" case. I'm not sure what to do for the second case – in the past we've usually fixed the config and then deleted the control plane machine to get the reconciler to retry. Do we have a way to plumb that flow through here?

When @akutz & I were working through the design for this, we decided not to try to build in any self-healing, for now. If the first control plane machine fails to come up for whatever reason, the remedy is for the user to manually delete the configmap lock (and probably the failed machine, too). Once the configmap is gone, the same initial "election" logic will happen again.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -93,61 +94,19 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
 		return errors.Errorf("missing cluster for machine %s/%s", machine.Namespace, machine.Name)
 	}
-	a.log.Info("Creating machine in cluster", "machine-name", machine.Name, "machine-namespace", machine.Namespace, "cluster-name", cluster.Name)
+
+	log := a.log.WithValues("machine-name", machine.Name, "namespace", machine.Namespace, "cluster-name", cluster.Name)
+	log.Info("Processing machine creation")
+
+	if cluster.Annotations[v1alpha1.AnnotationClusterInfrastructureReady] != "true" {
+		log.Info("Cluster infrastructure is not ready yet - requeuing machine")
+		return &controllerError.RequeueAfterError{RequeueAfter: 5 * time.Second}
```
vincepri: Should we wait a little longer than 5s here? Usually it might take a while until the Cluster is fully ready.

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: The cluster actuator has code to do this once the control plane is ready. Is that sufficient, or is there another case you're thinking of?

pkg/cloud/aws/actuators/machine/actuator.go
detiber: I prefer the explicit check against versions (at least for v1alpha1). If we do standardize on labels for v1alpha2+, then I would prefer we choose something namespaced and more descriptive for the key than `set`.

pkg/cloud/aws/actuators/control_plane_lock.go
```diff
@@ -0,0 +1,33 @@
...
+
+package actuators
+
+import (
+	"fmt"
+	"sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func ControlPlaneConfigMapName(cluster *v1alpha1.Cluster) string {
```
vincepri: Needs godoc

pkg/cloud/aws/actuators/control_plane_lock.go
```diff
@@ -0,0 +1,33 @@
...
+	return fmt.Sprintf("%s-controlplane", cluster.UID)
+}
+
+func ListOptionsForCluster(clusterName string) metav1.ListOptions {
```
vincepri: Needs godoc

pkg/cloud/aws/actuators/machine/actuator.go
vincepri: For the sake of consistency in CAPA, I'd use the `set` label here to check if the machine is a control plane. Later, when we move to v1alpha2 standards to determine the role of a Machine, we can quickly find and replace.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -158,26 +117,82 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+			// Need to requeue.
+			log.Info("Control plane configmap lock already exists but control plane is not ready yet - requeuing")
+			return &controllerError.RequeueAfterError{RequeueAfter: 5 * time.Second}
+		}
 	}
 
 	var bootstrapToken string
 	if isNodeJoin {
-		a.log.V(2).Info("Machine will join the cluster", "cluster", cluster.Name, "machine-name", machine.Name, "machine-namespace", machine.Namespace)
+		// First control plane machine has already been created. Now we can't proceed until its apiserver is up.
```
detiber: This comment doesn't seem accurate, since the waiting is done prior to this point.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -158,26 +117,82 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
+							Kind:       cluster.Kind,
+							Name:       cluster.Name,
+							UID:        cluster.UID,
+						},
+					},
+				},
+			}
+
+			log.Info("Attempting to create control plane configmap lock")
+			_, err = a.coreClient.ConfigMaps(cluster.Namespace).Create(controlPlaneConfigMap)
```
vincepri: Do we need the Get before this Create call?

pkg/cloud/aws/actuators/machine/actuator.go
detiber: No, I think cleaning up of the configMap in the Cluster actuator is correct, otherwise we have a race condition between the configMap being cleaned up and the controlplane ready annotation being set.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -158,26 +117,82 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 	}
 
 	var bootstrapToken string
 	if isNodeJoin {
-		a.log.V(2).Info("Machine will join the cluster", "cluster", cluster.Name, "machine-name", machine.Name, "machine-namespace", machine.Namespace)
+		// First control plane machine has already been created. Now we can't proceed until its apiserver is up.
 		coreClient, err := a.coreV1Client(cluster)
 		if err != nil {
-			return errors.Wrapf(err, "failed to retrieve corev1 client for cluster %q", cluster.Name)
+			return errors.Wrap(err, "unable to proceed until control plane is ready (error creating client)")
```
vincepri: We should probably keep the cluster name and add namespace to this error

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: - I doubt 5s is harmful, except for the fact this logs at the Info level and you'll see this every 5s.
- We can increase it, but I don't want it to be too long.
- We should probably create a constant.
- We may want multiple constants, if we decide to have different requeue times.

akutz: Related to https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/392

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: I think @vincepri might win here - we call `machine.Role()` to determine if it's a `"controlplane"` machine or not, and the implementation of `Role()` just grabs the `set` label.

I'll add a TODO

pkg/cloud/aws/actuators/machine/actuator.go
sethp-nr: It's in reference to the "invalid API request" mode above. I think that would take care of the "I asked for an IAM role that didn't exist" or "that security group is in a different account" cases. We make those kinds of mistakes fairly frequently in the inner iteration loop, and I'm not confident in future-me's ability to discover deleting the config map as the way to get myself un-stuck.

It's a little verbose, but maybe what I'm really asking for is instructions in the log message on line 181. Something to the effect of "if your control plane machine is stuck, run `kubectl delete -n NAMESPACE configmap/NAME-OF-CONFIGMAP` to try again"?

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: Actually, probably don't need the TODO since this logic ultimately will move out of CAPA for v1alpha2.

akutz: CAPA uses both `set` and the version in different places, and the CAPI function uses the version. Just saying :)

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: The Get is on line 144. If we can have Get return a successful response code, that is a good thing. We had some Velero users complaining because it always tried to create a resource, and most of the time it would fail (already exists). It shows up in the GCP/GKE error logs and they didn't like that.

pkg/apis/awsprovider/v1alpha1/types.go
ncdc: Sure

pkg/cloud/aws/actuators/machine/actuator.go
sethp-nr: Speaking of race conditions, I lost the read-modify-write cycle there 😄 

I was thinking of `GetOrCreateMachine` as create-only, but now that I've written that out I see my error. I guess we don't have the information at this point to determine whether it was a create error (could never possibly become ready) or a get error (it may yet be ready, who knows?), so it would indeed be racy.

pkg/cloud/aws/actuators/cluster/actuator.go
ncdc: @detiber we could, but that requires an extra HTTP call - do you want me to add that? I figured this was a pretty clean way to determine things.

detiber: The NodeRef controller wouldn't work for a pod-based control plane, or an externally managed controlplane. Or would we just say that those use cases need to set the annotation manually?

pkg/apis/awsprovider/v1alpha1/types.go
detiber: I like the idea of using `cluster-api`, since it is more directly tied to the project name, but I don't think `-` can be used in api groups, so maybe `cluster` is better for consistency.

pkg/cloud/aws/actuators/cluster/actuator.go
ncdc: Good points. How about a TODO for now to consider /healthz?

akutz: > The NodeRef controller wouldn't work for a pod-based control plane

I consider those edge cases. We're kind of all in on the NodeRef at this point. If there's an exception, I feel like another controller that sets some other version of a NodeRef/OwnerRef/RelatedRef should exist?

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: True, but I do think the same logic that determines if it's a control plane init/join should be used here (which is checking `set`).

pkg/cloud/aws/actuators/machine/actuator.go
vincepri: Fair enough!

pkg/apis/awsprovider/v1alpha1/types.go
vincepri: I'd say if we want to make it future-ready we can also use `aws.infrastructure.cluster.sigs.k8s.io` 😄

ncdc:
Are you all generally 👍 on the approach, and I should start fixing the broken/failing tests?

vincepri:
+1 from me

akutz:
+1

akutz:
@ncdc,

What did you end up calling your annotations? Did you rename the control-plane-ready label?

ncdc:
@akutz for now:
```go
const (
	AnnotationClusterInfrastructureReady = "aws.cluster.sigs.k8s.io/infrastructure-ready"
	AnnotationControlPlaneReady = "aws.cluster.sigs.k8s.io/control-plane-ready"
)
```

akutz:
> @akutz for now:
> 
> ```go
> const (
> 	AnnotationClusterInfrastructureReady = "aws.cluster.sigs.k8s.io/infrastructure-ready"
> 	AnnotationControlPlaneReady = "aws.cluster.sigs.k8s.io/control-plane-ready"
> )
> ```

Thanks. I didn't know if you used `infrastructure` for the control plane label or not.

ncdc:
Ready for final review

pkg/cloud/aws/actuators/cluster/actuator.go
detiber: They may very well be edge cases, but we do need to document how those cases could be handled, especially since we know there are people using cluster-api for this today.

ncdc: @detiber given that this code is in CAPA, and CAPA only supports machine-based control plane members today, do you still think there is anything actionable here?

detiber: @ncdc I think we still need to document it, since someone potentially could be using CAPA with a pre-existing external control plane. Otherwise they would be prevented from creating any Machines.

ncdc: Fair point. Maybe it makes more sense to switch to the /healthz check? Or would that have edge cases too?

detiber: I think just documenting the annotations should be sufficient, that should cover the external use cases.

ncdc: I thought we were talking about NodeRef?

detiber: We are, but setting of the annotation shortcuts the handling of NodeRef for the control plane and unblocks Machine joins.

ncdc:
/hold
Adding some docs

ncdc:
/hold cancel

Chatted in Slack - once this is lgtm + approved, we'll merge it. I'll file an issue to make sure we document this. We'll take some time to decide if this goes into 0.3.x or if master becomes 0.4. And I'll fix #869 for 0.3.x separately.

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -122,9 +133,54 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
 		return errors.Wrapf(err, "failed to get kubeconfig secret for cluster %q", cluster.Name)
 	}
 
+	if cluster.Annotations[v1alpha1.AnnotationControlPlaneReady] != v1alpha1.ValueReady {
```
vincepri: Nit: If we switch these two if conditions, we can return when ` cluster.Annotations[v1alpha1.AnnotationControlPlaneReady] == v1alpha1.ValueReady` and indent-out the longer part of the code

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -122,9 +133,54 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+				return errors.Wrapf(err, "Error deleting control plane configmap lock %q", configMapName)
+			}
+		}
+
+	}
+
 	return nil
 }
 
+const waitForControlPlaneMachineDuration = 15 * time.Second
```
vincepri: Nit: move globals at the top of the file

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -154,6 +210,7 @@ func (a *Actuator) Delete(cluster *clusterv1.Cluster) error {
 	if err := ec2svc.DeleteNetwork(); err != nil {
 		a.log.Error(err, "Error deleting cluster", "cluster-name", cluster.Name, "cluster-namespace", cluster.Namespace)
 		return &controllerError.RequeueAfterError{
+			// TODO: use a Duration expression for this
 			RequeueAfter: 5 * 1000 * 1000 * 1000,
```
vincepri: Can we actually change this as part of this PR?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -93,63 +114,27 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
-		return controlplaneExists, err
-
-	default:
-		return false, errors.Errorf("Unknown value %q for label `set` on machine %q", set, scope.Machine.Name)
-	}
-}
+const (
+	waitForClusterInfrastructureReadyDuration   = 15 * time.Second
+	waitForControlPlaneMachineExistenceDuration = 5 * time.Second
+	waitForControlPlaneReadyDuration            = 5 * time.Second
```
vincepri: Same as above, move these at the top of the file?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -158,26 +143,44 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 
 	controlPlaneMachines := GetControlPlaneMachines(clusterMachines)
+	if len(controlPlaneMachines) == 0 {
+		log.Info("No control plane machines exist yet - requeuing")
+		return &controllerError.RequeueAfterError{RequeueAfter: waitForControlPlaneMachineExistenceDuration}
+	}
+
+	// Except for the very first control plane machine, everything will be a join, so let's assume we're joining.
+	isNodeJoin := true
+	if cluster.Annotations[v1alpha1.AnnotationControlPlaneReady] != v1alpha1.ValueReady {
```
vincepri: Should we make this a method?

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -208,10 +204,75 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 	}
-	a.log.Info("Create completed", "machine-name", machine.Name)
+
+	log.Info("Create completed")
+
 	return nil
 }
 
+// controlPlaneInitLocker uses a ConfigMap to synchronize cluster initialization.
+type controlPlaneInitLocker struct {
```
vincepri: This could probably go into its own file

pkg/cloud/aws/actuators/machine/actuator_test.go
```diff
@@ -580,379 +585,115 @@ func TestImmutableStateChange(t *testing.T) {
...
-			},
-		}
-		mockEC2.EXPECT().DescribeInstances(dmi).Return(
-			nil,
-			awserrors.NewNotFound(fmt.Errorf("mocked not found"))).Times(1)
-	}
-
-	return mockEC2
+func (c *configMapClient) Update(*v1.ConfigMap) (*v1.ConfigMap, error) {
+	panic("not implemented")
```
chuckha: These panics are generally why I scope down interfaces as much as possible. Unfortunately client-go doesn't make it easy and since these objects will be namespaced at run time we'd have to scope both ConfigMapNamespacer and then the ConfigMapGetter which is kind of ugly.

What you've done is great though, this comment needs no action

pkg/cloud/aws/actuators/cluster/actuator.go
ncdc: Sure, that's 5 seconds, right?

pkg/cloud/aws/actuators/cluster/actuator.go
ncdc: I like them close to where they're accessed, but I can move them if you feel strongly.

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: I like them close to where they're accessed, but I can move them if you feel strongly.

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: Do you mean something to the effect of `determineIsNodeJoin`?

pkg/cloud/aws/actuators/cluster/actuator.go
vincepri: Correct, had to make sure though :) https://play.golang.org/p/LFjZSdONubV

pkg/cloud/aws/actuators/machine/actuator.go
vincepri: Yeah something along those lines

pkg/cloud/aws/actuators/machine/actuator.go
vincepri: Not a huge problem, was pointing out for consistency's sake

pkg/cloud/aws/actuators/cluster/actuator.go
ncdc: Do you think the `return nil` at the end of the `ValueReady` branch would be missed or confusing?:

```go
if cluster.Annotations[v1alpha1.AnnotationControlPlaneReady] == v1alpha1.ValueReady {
		configMapName := actuators.ControlPlaneConfigMapName(cluster)
		log.Info("Checking for existence of control plane configmap lock", "configmap-name", configMapName)

		_, err := a.coreClient.ConfigMaps(cluster.Namespace).Get(configMapName, metav1.GetOptions{})
		switch {
		case apierrors.IsNotFound(err):
			// It doesn't exist - no-op
		case err != nil:
			return errors.Wrapf(err, "Error retrieving control plane configmap lock %q", configMapName)
		default:
			if err := a.coreClient.ConfigMaps(cluster.Namespace).Delete(configMapName, nil); err != nil {
				return errors.Wrapf(err, "Error deleting control plane configmap lock %q", configMapName)
			}
		}

		// would this be easy to miss / be confusing?
		return nil
	}

	log.Info("Cluster does not have ready annotation - checking for ready control plane machines")

	machines, err := a.client.Machines(cluster.Namespace).List(actuators.ListOptionsForCluster(cluster.Name))
	if err != nil {
		return errors.Wrapf(err, "failed to list machines for cluster %q", cluster.Name)
	}

	controlPlaneMachines := machine.GetControlPlaneMachines(machines)

	machineReady := false
	for _, machine := range controlPlaneMachines {
		if machine.Status.NodeRef != nil {
			machineReady = true
			break
		}
	}

	if machineReady {
		log.Info("Setting cluster ready annotation")
		cluster.Annotations[v1alpha1.AnnotationControlPlaneReady] = v1alpha1.ValueReady
	} else {
		log.Info("No control plane machines are ready - requeuing cluster")
		return &controllerError.RequeueAfterError{RequeueAfter: waitForControlPlaneMachineDuration}
	}

	return nil
```

vincepri: Yeah I'd do something like:

```go
        [...]
	if !machineReady {
		log.Info("No control plane machines are ready - requeuing cluster")
		return &controllerError.RequeueAfterError{RequeueAfter: waitForControlPlaneMachineDuration}
	}

        log.Info("Setting cluster ready annotation")
	cluster.Annotations[v1alpha1.AnnotationControlPlaneReady] = v1alpha1.ValueReady
	return nil
```

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -208,10 +179,31 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
...
 		return errors.Errorf("failed to reconcile LB attachment: %+v", err)
 	}
-	a.log.Info("Create completed", "machine-name", machine.Name)
+
+	log.Info("Create completed")
+
 	return nil
 }
 
+func (a *Actuator) isNodeJoin(log logr.Logger, cluster *clusterv1.Cluster, machine *clusterv1.Machine) (bool, error) {
```
ashish-amarnath: can we have test coverage for this?

pkg/cloud/aws/actuators/cluster/actuator.go
vincepri: Maybe also add a comment before and after the if condition to make sure folks are aware what's going on?

pkg/cloud/aws/actuators/machine/actuator.go
ncdc: Yes, I can add it. This was extracted from within the Create method, which made it hard to test.

pkg/cloud/aws/actuators/cluster/actuator.go
vincepri: Specifically 

`if cluster.Annotations[v1alpha1.AnnotationControlPlaneReady] == v1alpha1.ValueReady {`

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -93,63 +105,21 @@ func machinesEqual(m1 *clusterv1.Machine, m2 *clusterv1.Machine) bool {
...
 	if cluster == nil {
 		return errors.Errorf("missing cluster for machine %s/%s", machine.Namespace, machine.Name)
 	}
-	a.log.Info("Creating machine in cluster", "machine-name", machine.Name, "machine-namespace", machine.Namespace, "cluster-name", cluster.Name)
 
-	scope, err := actuators.NewMachineScope(actuators.MachineScopeParams{Machine: machine, Cluster: cluster, Client: a.clusterClient, Logger: a.log})
+	log := a.log.WithValues("machine-name", machine.Name, "namespace", machine.Namespace, "cluster-name", cluster.Name)
+	log.Info("Processing machine creation")
+
+	if cluster.Annotations[v1alpha1.AnnotationClusterInfrastructureReady] != v1alpha1.ValueReady {
```
ashish-amarnath: So the cluster actuator, to set the cluster-infra, waits for a control-plane machine to have a nodeRef which will happen once the machine is created. However, seems like the machine actuator also waits for the cluster infra to be ready. Am I missing something?

ncdc: There are 2 different annotations: infra-ready, and control-plane-ready. The cluster actuator waits for the infra (VPC, ELB, etc) to be ready, then it sets infra-ready. It does not wait for the control plane for infra-ready.

ashish-amarnath: Yes, ofcourse I was missing something. The name of the annotations `AnnotationClusterInfrastructureReady` and `AnnotationControlPlaneReady`

ashish-amarnath: 🤦‍♂

ashish-amarnath:
@ncdc LGTM from me except for the unit tests for `isNodeJoin` I am OK if you want to do it in a different PR.

ncdc:
@vincepri updated the branching style.

@ashish-amarnath added a unit test.

ncdc:
Chatted with Vince in Slack:

/hold cancel

PR #790, opened 2019-05-23
chore: append nodeRole label if additional labels provided

### LINKED ISSUE #789: Kubernetes node labels not pass into nodes
/kind bug

We want the propagate Kubernetes node labels through machinedeployment. 

I added the following to the machinedeployment yaml

```
    spec:
      versions:
        kubelet: v1.13.6
      providerSpec:
        value:
          apiVersion: awsprovider/v1alpha1
          kind: AWSMachineProviderSpec
          instanceType: m5.large
          iamInstanceProfile: "nodes.cluster-api-provider-aws.sigs.k8s.io"
          keyName: gophoton-global-keypair
          kubeadmConfiguration:
            init:
              nodeRegistration:
                kubeletExtraArgs: 
                  node-labels: vmware.com/testlabel=testlabel1
            join:
              nodeRegistration:
                kubeletExtraArgs: 
                  node-labels: vmware.com/testlabel=testlabel2
```

Specifically I added the "kubeadmConfiguration" section

The expected behavior is that the when I do "kubectl get node", we should see the node-label that I defined there. It is not there.

I read some of the CAPA code and noticed that:

```
if !util.IsControlPlaneMachine(machine.GetMachine()) {
		out.NodeRegistration.KubeletExtraArgs["node-labels"] = nodeRole
}
```

It seems like no matter waht I pass in as the KubeletExtraArgs["node-labels"] , it will be overwritten by the CAPA code to be "nodeRole". I think this is what caused the issue. Can someone confirm and help with the issue? Thanks

rudoi:

This PR appends the `nodeRole` Node label when the `node-labels` Kubelet argument is supplied, rather than overriding it. 

Fixes #789 

Tested with worker node MachineDeployments.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/aws/services/kubeadm/aws_defaults.go
+++ b/pkg/cloud/aws/services/kubeadm/aws_defaults.go
@@ -193,7 +193,11 @@ func SetJoinNodeConfigurationOverrides(caCertHash, bootstrapToken string, machin
 	}
 	out.NodeRegistration.KubeletExtraArgs["cloud-provider"] = CloudProvider
 	if !util.IsControlPlaneMachine(machine.GetMachine()) {
-		out.NodeRegistration.KubeletExtraArgs["node-labels"] = nodeRole
+		if labels, ok := out.NodeRegistration.KubeletExtraArgs["node-labels"]; ok {
+			out.NodeRegistration.KubeletExtraArgs["node-labels"] = fmt.Sprintf("%s,%s", labels, nodeRole)
+		} else {
+			out.NodeRegistration.KubeletExtraArgs["node-labels"] = nodeRole
+		}
 	}
 	return out
 }
--- a/pkg/cloud/aws/services/kubeadm/aws_defaults.go
+++ b/pkg/cloud/aws/services/kubeadm/aws_defaults.go
@@ -18,6 +18,7 @@ package kubeadm
 
 import (
 	"fmt"
+	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/cloud/aws/services/kubeadm/aws_defaults.go
```diff
@@ -193,7 +193,11 @@ func SetJoinNodeConfigurationOverrides(caCertHash, bootstrapToken string, machin
 	}
 	out.NodeRegistration.KubeletExtraArgs["cloud-provider"] = CloudProvider
 	if !util.IsControlPlaneMachine(machine.GetMachine()) {
-		out.NodeRegistration.KubeletExtraArgs["node-labels"] = nodeRole
+		if labels, ok := out.NodeRegistration.KubeletExtraArgs["node-labels"]; ok {
+			out.NodeRegistration.KubeletExtraArgs["node-labels"] = fmt.Sprintf("%s,%s", labels, nodeRole)
```
vincepri: Should we use strings.Split and strings.Join instead of manual concatenation?

vincepri: Could we add a test case as well?

rudoi: Yep I agree with this. I believe I have done all the things. 😸

PR #796, opened 2019-05-28
Remove retries

chuckha:

This PR Removes the retries around creating a secret. Instead of retrying for a long time we should just bail and let the cluster-api controller re-reconcile. These retries are not necessary and will only slow down imminent failure.

Another added benefit here is that errors on creating secrets don't get swallowed.

**Release note**:
```release-note
NONE
```


chuckha:
@vincepri The default action when an error is returned is to re-add the request to the reconcile queue with a rate limited request. 

If we explicitly return a RequeueAfter error, how long would you suggest we tell the controller to rereconcile? I can't think of any value based on logic here and feel we should leave this logic up to the controller's rate limited request function (the default behavior).

PR #805, opened 2019-06-04
Build and update images for k8s v1.14.2

detiber:

Previously the latest version of the images for v1.14.x was v1.14.1

**Release note**:
```release-note
NONE
```

### DIFF
--- a/build/amis/ansible/roles/common/tasks/redhat.yml
+++ b/build/amis/ansible/roles/common/tasks/redhat.yml
@@ -15,6 +15,8 @@
 - name: add epel repo
   yum:
     name: "{{ common_redhat_epel_rpm }}"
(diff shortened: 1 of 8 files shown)


build/amis/ansible/roles/common/tasks/redhat.yml
```diff
@@ -15,6 +15,8 @@
 - name: add epel repo
   yum:
     name: "{{ common_redhat_epel_rpm }}"
+    state: present
+    lock_timeout: 60
```
detiber: lock_timeout is needed for Ansible version 2.8.0+, and only available in version 2.8.0+, which prompted the previous version requirement bump for Ansible in build/amis/README.md

PR #811, opened 2019-06-07
Add availaibilityZone to Machine Spec

### LINKED ISSUE #799: Support zone for machine deployment
/kind feature

For me use case, I want to be able to associate zone to machine deployment instead of using subnet. If we can achieve step 1: cluster-api team to add the list of zones as an extra field in the ```AWSClusterProviderSpec`'s NetworkSpec``` and when creating VPC, creates multiple zones specified in the spec, each associated with a subnet. Then maybe we can move to step 2, which is support for using zone(one of the zones defined as part of cluster object) to create machine deployment

Please feel free to reach out to me with alternative design choices and the viability of the approach I mentioned above.

vincepri:

This PR adds a new field to the Machine Spec called "availabilityZone". As expressed in the linked issue, users might prefer to specify an AZ to a SubnetID for simplicity. This PR is a simple implementation for that request which picks the first available private subnet within the AZ specified.

Fixes #799 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -159,10 +159,23 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 		}
 	}
 
-	// Pick subnet from the machine configuration, or default to the first private available.
+	// Pick subnet from the machine configuration, or based on the availability zone specified,
+	// or default to the first private subnet available.
+	// TODO(vincepri): Move subnet picking logic to its own function/method.
 	if machine.MachineConfig.Subnet != nil && machine.MachineConfig.Subnet.ID != nil {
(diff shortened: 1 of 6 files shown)


pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -159,10 +159,24 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
...
+					machine.Name(),
+					*machine.MachineConfig.AvailabilityZone,
+				),
+			)
+		}
+		input.SubnetID = sns[0].ID
+	}
+
+	// Default to the first private subnet available.
+	if input.SubnetID == "" {
```
detiber: Why not an additional else condition instead of a separate if statement?

vincepri: Fixed! Not sure what past-Vince was thinking last night :P

PR #829, opened 2019-06-12
Ignore deleted control plane machines

### LINKED ISSUE #828: Additional control plane machines that should join the cluster may inadvertently form new clusters instead
/kind bug

- Create a cluster with multiple control plane nodes
- Remove one of the control plane nodes from the etcd member list
- `kubectl delete` the machine for that node
- While the machine is still being deleted (deletion timestamp is set but the machine resource still exists while the AWS infra is being deleted), create a new control plane machine
- The new control plane machine ran `kubeadm init` instead of `kubeadm join`

The new control plane machine should always just the cluster

It looks like possibly `GetControlPlaneMachines` in the machine actuator code needs to ignore machines that have a non-zero deletion timestamp. Otherwise, `isNodeJoin` may ask AWS if the deleted machine's ec2 instance is still either pending or running, and this won't be the case. If this happens, `isNodeJoin` returns false, causing CAPA to consider this new machine to be the first control plane node.

**Environment:**

- Cluster-api-provider-aws version: master / commit f9051c978d278dd575f39dce071fe6234431b44a

ncdc:
Ignore deleted control plane machines when checking if a new machine should be a control plane node join or not.

Fixes #828

**Release note**:

```release-note
Fix an issue where adding extra control plane machines could accidentally create new clusters instead of joining the existing cluster.
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine/actuator.go
+++ b/pkg/cloud/aws/actuators/machine/actuator.go
@@ -80,11 +80,11 @@ func NewActuator(params ActuatorParams) *Actuator {
 	}
 }
 
-// GetControlPlaneMachines retrieves all control plane nodes from a MachineList
+// GetControlPlaneMachines retrieves all non-deleted control plane nodes from a MachineList
 func GetControlPlaneMachines(machineList *clusterv1.MachineList) []*clusterv1.Machine {
 	var cpm []*clusterv1.Machine
 	for _, m := range machineList.Items {
-		if m.Spec.Versions.ControlPlane != "" {
+		if m.DeletionTimestamp.IsZero() && m.Spec.Versions.ControlPlane != "" {
 			cpm = append(cpm, m.DeepCopy())
 		}
 	}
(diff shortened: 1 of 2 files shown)


ncdc:
I would love someone to double check my assessment of this issue - thanks!

rudoi:
@ncdc - This change feels logical to me, but I'm having a hard time understanding how finding deleted Machines would result in a new Cluster. Would you mind elaborating a little bit? 😅

ashish-amarnath:
@rudoi  https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/828

> It looks like possibly GetControlPlaneMachines in the machine actuator code needs to ignore machines that have a non-zero deletion timestamp. Otherwise, isNodeJoin may ask AWS if the deleted machine's ec2 instance is still either pending or running, and this won't be the case. If this happens, isNodeJoin returns false, causing CAPA to consider this new machine to be the first control plane node.

rudoi:
@ashish-amarnath ah, reading the issue helps! 😄 Thank you, that makes perfect sense.

PR #830, opened 2019-06-13
Add support for NodeRef controller

vincepri:

This PR adds support for the newly NodeRef controller in upstream CAPI.

Fixes #

The Gopkg.toml has been updated to consume CAPI master instead of the release. Once we backport the changes upstream, I'll switch it back. This was used to test on a development environment and shouldn't be merged as is. 

**Release note**:

```release-note

```


pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -88,6 +97,28 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+	kubeConfigSecret := &apiv1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Name: remote.KubeConfigSecretName(cluster.Name),
+		},
+		StringData: map[string]string{
+			"value": kubeConfig,
+		},
+	}
+
+	if _, err := a.coreClient.Secrets(cluster.Namespace).Create(kubeConfigSecret); err != nil {
```
ncdc: I would add a `Get()` check to see if it exists, and only try to create if it doesn't.

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -28,31 +32,36 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/pkg/apis/cluster/v1alpha1"
 	client "sigs.k8s.io/cluster-api/pkg/client/clientset_generated/clientset/typed/cluster/v1alpha1"
 	controllerError "sigs.k8s.io/cluster-api/pkg/controller/error"
+	"sigs.k8s.io/cluster-api/pkg/controller/remote"
 )
 
 //+kubebuilder:rbac:groups=awsprovider.k8s.io,resources=awsclusterproviderconfigs;awsclusterproviderstatuses,verbs=get;list;watch;create;update;patch;delete
 //+kubebuilder:rbac:groups=cluster.k8s.io,resources=clusters;clusters/status,verbs=get;list;watch;create;update;patch;delete
+//+kubebuilder:rbac:groups=,resources=secrets,verbs=create;update
```
ncdc: I'd do get & create and omit update

vincepri: Would we need the update long term? I added in case we need it at some point, so it's easier to upgrade clusters running v1alpha1, wdyt?

Gopkg.toml
```diff
@@ -50,7 +50,7 @@ required = [
 
 [[constraint]]
   name = "sigs.k8s.io/cluster-api"
-  branch = "release-0.1"
+  branch = "master"
```
detiber: This will need to be updated back to `release-0.1`

pkg/cloud/aws/actuators/cluster/actuator.go
detiber: I'd say to avoid update for now, if we need it in the future we can add it then.

ncdc: If we need to "update", the user can manually delete the secret, and it will be recreated on the next reconcile pass.

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -88,6 +97,30 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Name: remote.KubeConfigSecretName(cluster.Name),
+		},
+		StringData: map[string]string{
+			"value": kubeConfig,
+		},
+	}
+
+	secretClient := a.coreClient.Secrets(cluster.Namespace)
+	if _, err := secretClient.Get(kubeConfigSecret.Name, metav1.GetOptions{}); err != nil && apierrors.IsNotFound(err) {
```
ncdc: Please move this up to wrap all the secret logic

config/rbac/rbac_role.yaml
```diff
@@ -30,6 +30,13 @@ rules:
   - update
   - patch
   - delete
+- apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - create
+  - update
```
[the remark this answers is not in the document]
detiber: Looks like this needs to be regenerated.

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -88,6 +97,30 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+		if _, err := secretClient.Create(kubeConfigSecret); err != nil {
+			return err
+		}
+	} else if err != nil {
+		return err
```
ncdc: Need to wrap the error

pkg/cloud/aws/actuators/cluster/actuator.go
```diff
@@ -88,6 +97,30 @@ func (a *Actuator) Reconcile(cluster *clusterv1.Cluster) error {
...
+		},
+		StringData: map[string]string{
+			"value": kubeConfig,
+		},
+	}
+
+	secretClient := a.coreClient.Secrets(cluster.Namespace)
+	if _, err := secretClient.Get(kubeConfigSecret.Name, metav1.GetOptions{}); err != nil && apierrors.IsNotFound(err) {
+		if _, err := secretClient.Create(kubeConfigSecret); err != nil {
+			return err
```
ncdc: Need to wrap the error

pkg/cloud/aws/actuators/cluster/actuator.go
vincepri: What do you mean?

ncdc: We only need to call `a.Deployer.GetKubeConfig` and prepare an `apiv1.Secret` object if we don't find the secret when we call `Get`

vincepri: ptal

ncdc:
LGTM. @detiber wdyt?

PR #833, opened 2019-06-14
Improve dependency graph

chuckha:
Importing the types of CAPA no longer results in pulling in the aws sdk

This PR cleans up the dep graph. Previous to this commit if the types of CAPA were imported then we would also pull in the aws sdk. This moves tags into the types and allows importing of (some) CAPA code without needing to import the aws sdk.

```release-note
NONE
```


pkg/cloud/aws/services/mocks/services_mock.go
```diff
@@ -51,18 +51,18 @@ func (m *MockEC2Interface) EXPECT() *MockEC2InterfaceMockRecorder {
 }
 
 // CreateOrGetMachine mocks base method
-func (m *MockEC2Interface) CreateOrGetMachine(arg0 *actuators.MachineScope, arg1, arg2 string) (*v1alpha1.Instance, error) {
+func (m *MockEC2Interface) CreateOrGetMachine(arg0 *actuators.MachineScope, arg1 string) (*v1alpha1.Instance, error) {
```
detiber: oops, I guess I missed this :joy:

chuckha:
this wouldn't affect vendor for this repo, it would only affect vendor of apps that may be pulling in CAPA types

pkg/apis/awsprovider/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -509,6 +508,28 @@ func (in *KubeadmConfiguration) DeepCopy() *KubeadmConfiguration {
...
+		*out = make(Map, len(*in))
+		for key, val := range *in {
+			(*out)[key] = val
+		}
+		return
+	}
+}
+
+// DeepCopy is an autogenerated deepcopy function, copying the receiver, creating a new Map.
+func (in Map) DeepCopy() Map {
```
sethp-nr: Kind of a nice side effect bonus.

pkg/cloud/aws/tags/tags.go
```diff
@@ -81,19 +82,19 @@ type BuildParams struct {
...
 	// Any additional tags to be added to the resource.
 	// +optional
-	Additional Map
+	Additional v1alpha1.Map
 }
 
 // Build builds tags including the cluster tag and returns them in map form.
-func Build(params BuildParams) Map {
-	tags := make(Map)
+func Build(params BuildParams) v1alpha1.Map {
```
sethp-nr: I think this could safely move across to the types package too

vincepri:
+1 to @sethp-nr comment, v1alpha1.Map seems very generic

pkg/cloud/aws/services/ec2/routetables_test.go
```diff
@@ -118,8 +117,8 @@ func TestReconcileRouteTables(t *testing.T) {
 				VPC: v1alpha1.VPCSpec{
 					InternetGatewayID: aws.String("igw-01"),
 					ID:                "vpc-routetables",
-					Tags: tags.Map{
-						tags.ClusterKey("test-cluster"): "owned",
+					Tags: v1alpha1.Map{
+						v1alpha1.ClusterKey("test-cluster"): "owned",
```
vincepri: We should prefix the exposed functions as well, reading this line without going to the definition is hard to understand this is meant to be used in tags.

pkg/cloud/aws/services/ec2/bastion.go
```diff
@@ -101,7 +101,7 @@ func (s *Service) DeleteBastion() error {
 func (s *Service) describeBastionInstance() (*v1alpha1.Instance, error) {
 	input := &ec2.DescribeInstancesInput{
 		Filters: []*ec2.Filter{
-			filter.EC2.ProviderRole(tags.ValueBastionRole),
+			filter.EC2.ProviderRole(v1alpha1.ValueBastionRole),
```
vincepri: Same here, this should be prefixed somehow

pkg/cloud/aws/tags/tags.go
chuckha: That's fine with me, `BuildParams` can safely be moved to v1alpha1

pkg/cloud/aws/services/ec2/bastion.go
chuckha: Any suggestions here? I don't think a prefix on the roles makes sense, but happy to change if you feel strongly...maybe `ValueBastionRoleTagValue`?

chuckha: actually on second thought maybe a `Tag` prefix makes sense... but `Value` is already there. my first suggestion was not good

chuckha: actually on second thought maybe a `Tag` prefix makes sense... but `Value` is already there. my first suggestion was not good

chuckha: ah, similar to what I did with the key => `BastionRoleTagValue` move the prefix to a suffix and add the word `Tag`.

Thoughts?

detiber: +1 from me on `BastionRoleTagValue`

vincepri: +2 `BastionRoleTagValue`

chuckha:
this test passed locally... maybe it just needs a few more trys

/retest

PR #831, opened 2019-06-13
Remove unused parameter

detiber:
**What this PR does / why we need it**:
Removes and unused parameter

**Release note**:
```release-note
NONE
```

PR #835, opened 2019-06-17
Add logic to handle implicit route table associations to main route table

erstaples:
Currently, when specifying a pre-provisioned VPC in the cluster provider spec, the provider controller scans each subnet in the VPC and labels it as `public` if it has a tag `sigs.k8s.io/cluster-api-provider-aws/role = public`, or if it's _explicitly_ associated with a route table that has an internet gateway route. 

The problem is that a subnet may be _implicitly_ associated with a "main" route table in the VPC, which can include an internet gateway route. The current logic ignores such implicit associations, which can leave users wondering why their subnet is not marked as public. 

This PR fixes that.

Fixes #

**Release note**:

```release-note
Add logic to handle implicit route tables associations
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/routetables.go
+++ b/pkg/cloud/aws/services/ec2/routetables.go
@@ -30,6 +30,7 @@ import (
 
 const (
 	anyIPv4CidrBlock = "0.0.0.0/0"
+	mainRouteTableInVPCKey = "main"
 )
 
 func (s *Service) reconcileRouteTables() error {
@@ -109,6 +110,9 @@ func (s *Service) describeVpcRouteTablesBySubnet() (map[string]*ec2.RouteTable, 
 	res := make(map[string]*ec2.RouteTable)
 	for _, rt := range rts {
 		for _, as := range rt.Associations {
+			if *as.Main {
+				res[mainRouteTableInVPCKey] = rt
+			}
 			if as.SubnetId == nil {
 				continue
 			}
--- a/pkg/cloud/aws/services/ec2/routetables.go
+++ b/pkg/cloud/aws/services/ec2/routetables.go
@@ -110,7 +110,7 @@ func (s *Service) describeVpcRouteTablesBySubnet() (map[string]*ec2.RouteTable, 
 	res := make(map[string]*ec2.RouteTable)
 	for _, rt := range rts {
 		for _, as := range rt.Associations {
-			if *as.Main {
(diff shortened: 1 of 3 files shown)


pkg/cloud/aws/services/ec2/subnets.go
```diff
@@ -202,6 +202,12 @@ func (s *Service) describeVpcSubnets() (v1alpha1.Subnets, error) {
 
 		// ... or if it has an internet route
 		rt := routeTables[*ec2sn.SubnetId]
+		if rt == nil {
+			// If there is no explicit association, subnet defaults to main route table as implicit association
+			if mainRt, ok := routeTables[mainRouteTableInVPCKey]; ok {
+				rt = mainRt
```
sethp-nr: nit: I don't think we need the `if` here, since the zero value (what we get back for a missing key) of a pointer is nil. I think we can just do 
```
if rt == nil {
    // If there is no explicit association, subnet defaults to main route table as implicit association
    rt = routeTables[mainRouteTableInVPCKey]
}
```

+1 on the comment, that's very helpful.

erstaples: Ah, good call. Will fix.

pkg/cloud/aws/services/ec2/routetables.go
```diff
@@ -109,6 +110,9 @@ func (s *Service) describeVpcRouteTablesBySubnet() (map[string]*ec2.RouteTable,
 	res := make(map[string]*ec2.RouteTable)
 	for _, rt := range rts {
 		for _, as := range rt.Associations {
+			if *as.Main {
```
detiber: Is main guaranteed to be non-nil for all RouteTableAssociations? If not, then this should probably be: `if as.Main != nil && *as.Main {`

erstaples: This explains the breaking test. Thanks.

erstaples:
@detiber @vincepri The `CreateOrGetMachine` service mock updated after running `make test`, although I didn't change any code related to that function. I'm a little thrown off by that.

detiber:
> The CreateOrGetMachine service mock updated after running make test, although I didn't change any code related to that function. I'm a little thrown off by that.

:joy: oops that was my fault, missed that in: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/831

PR #839, opened 2019-06-18
[0.3] Update events for Machines

detiber:
Adds additional events for Machines

**Release note**:
```release-note
Emit proper events for Machine creation/termination
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -388,7 +388,6 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 		return nil, err
 	}
 
-	record.Eventf(machine.Machine, "CreatedInstance", "Created new %s instance with id %q", machine.Role(), out.ID)
 	return out, nil
 }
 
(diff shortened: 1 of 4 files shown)


pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -484,7 +482,15 @@ func (s *Service) CreateOrGetMachine(machine *actuators.MachineScope, bootstrapT
 		return instance, nil
 	}
 
-	return s.createInstance(machine, bootstrapToken)
+	instance, err = s.createInstance(machine, bootstrapToken)
+	if err != nil {
+		if !awserrors.IsFailedDependency(errors.Cause(err)) {
```
ncdc: What happens if it is a failed dependency?

detiber: At the actuator/controller level it is requeued. This is to avoid spamming failure events for those cases.

ncdc: Thanks, that is quite helpful! Would you mind adding this info as a comment?

ncdc:
Would you have time to add events for all AWS resources that we create (VPC, subnet, route, security group, etc)?

PR #823, opened 2019-06-11
fix: use provided NodeRegistrationOptions as base

rudoi:
**What this PR does / why we need it**:
This uses the provided Init/Join configurations' NodeRegistration as a base. If there is no provided NodeRegistration, this effectively functions the same as before, starting with a blank NodeRegistrationOptions and executing the funcOpts.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #822 

**Release note**:

```release-note
Fixes bug where provided NodeRegistration is not honored
```

PR #863, opened 2019-07-01
fix: respect cri socket if set

### LINKED ISSUE #856: support alternative cri implementations
/kind feature

Support for CRI implementations beyond `containerd`, namely `docker`. A good first pass may just be to avoid overwriting the `criSocket` value in `nodeRegistration` if it is set by the machine spec.

Our current container environment is dependent on the use of `docker`, which requires setting `kubeadmConfiguration.[init,join].nodeRegistration.criSocket` as well as an alternative ami with `docker` installed in order to suggest to `kubeadm` that `docker` be used over `containerd`. 

A useful comment from @detiber on slack:
> Longer term, my plan was to add the container runtime to the ami naming convention we use, and make it selectable via the providerConfig

I do think it would be useful to support multiple different CRI implementations via swapping out of amis, but I think there would always be a need to use a custom ami (for ex. we build amis for our kubernetes clusters with system dependencies specific to our use-cases).

**Environment:**

- Cluster-api-provider-aws version: master/0.3.2
- Kubernetes version: (use `kubectl version`): 1.13.7
- OS (e.g. from `/etc/os-release`): ubuntu-18.04

sethp-nr:
Allows overriding the container runtime interface on a case-by-case
basis. Useful (and likely required) if the Machine is using a custom
AMI.

Fixes #856 

I also grabbed a quick typo fix in the docs for a file I was looking at.

**Release note**:

```release-note
respect value in Machine's `kubeadmConfiguration.[init,join].nodeRegistration.criSocket`, if set
```


sethp-nr:
cc @chuckha

detiber:
lgtm, waiting for @chuckha to chime in.

pkg/cloud/aws/services/kubeadm/noderegistration.go
```diff
@@ -41,10 +41,12 @@ func WithTaints(taints []corev1.Taint) NodeRegistrationOption {
 	}
 }
 
-// WithCRISocket sets the location of the container runtime socket.
-func WithCRISocket(socket string) NodeRegistrationOption {
+// WithDefaultCRISocket sets the location of the container runtime socket, if it's not already set
+func WithDefaultCRISocket(socket string) NodeRegistrationOption {
```
chuckha: I think this gets a little confusing when this function may or may not update the socket, it would be more clear if it only does one thing and we figure out the crisocket before we instantiate the object

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -325,7 +325,7 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 				kubeadm.SetNodeRegistrationOptions(
 					&machine.MachineConfig.KubeadmConfiguration.Join.NodeRegistration,
 					kubeadm.WithNodeRegistrationName(hostnameLookup),
-					kubeadm.WithCRISocket(containerdSocket),
+					kubeadm.WithDefaultCRISocket(containerdSocket),
```
chuckha: My suggestion would be to figure out which socket to use above this instantiation logic and then use the existing method to set the CRISocket

is that possible with your use case?

chuckha:
I made one suggestion to see if we could keep the functional options as dumb as possible. Please let me know if my suggestion works for your use case @sethp-nr

pkg/cloud/aws/services/kubeadm/noderegistration.go
sethp-nr: Yeah, I didn't love the maybe-option (and the combinatorial explosion functions it represented). Your idea sounds way better, I'll do that :)

sethp-nr:
@chuckha well that turned into a little big bigger of a refactor than I was anticipating 😀 

I wanted to write some unit tests that expressed the various use-cases in "machine config in, kubeadm config out" form. I think I got close, but feedback is definitely appreciated.

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -202,6 +204,8 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 		return input, err
 	}
 
+	apiServerEndpoint := fmt.Sprintf("%s:%d", machine.Network().APIServerELB.DNSName, apiServerBindPort)
```
chuckha: does this need a nil check at all or are we ok?

pkg/cloud/aws/services/ec2/instances_test.go
```diff
@@ -553,3 +555,472 @@ vuO9LYxDXLVY9F7W4ccyCqe27Cj1xyAvdZxwhITrib8Wg5CMqoRpqTw5V3+TpA==
...
+					KubeletExtraArgs: map[string]string{
+						"cloud-provider": "aws",
+					},
+				},
+			},
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			setInitConfigurationOptions(&tt.args.initConfig, tt.args.machine)
```
chuckha: There is no logic in `SetInitConfigurationOptions` so I'm a little confused as to what this is testing. Because we've tested the builder we can assume that this function works, but what do you think?

The only function you've added where there is logic is `stringOrDefault` which we can see is both easily testable (but doesn't necessarily need tests) and is unlikely to contain a bug.

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -376,6 +344,74 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
...
+				kubeadm.WithNodeRegistrationName(hostnameLookup),
+				kubeadm.WithCRISocket(stringOrDefault(joinConfig.NodeRegistration.CRISocket, containerdSocket)),
+				kubeadm.WithKubeletExtraArgs(map[string]string{"cloud-provider": cloudProvider}),
+			),
+		),
+		kubeadm.WithLocalAPIEndpointAndPort(localIPV4Lookup, apiServerBindPort),
+	)
+}
+
+func stringOrDefault(val, def string) string {
```
chuckha: This is super generic, what do you think about something like `getCRISocketPath`  which embeds the default in the function? Maybe we can make it generic if we have more stringOrDefaulting to do in future? 

`.WithCRISocket(getCRISocketPath(joinConfig.NodeRegistration.CRISocket))` reads a little nicer to me, but what do you think?

you could even reuse stringOrDefault in the `getCRISocketPath` function if you wanted

sethp-nr: yeah, I like `WithCRISocket(getCRISocketPath(joinConfig.NodeRegistration.CRISocket))` a lot better. I'll change it.

pkg/cloud/aws/services/ec2/instances_test.go
sethp-nr: We had tested all of the individual pieces, but we hadn't tested their composition anywhere. The logic of "an Init configuration must include these pieces, but these pieces should be user overridable" that gets made up by which options end up getting passed with what arguments into the builder.

If we'd had a test like this "before" (a foolhardy assertion, I know 😆) I think it would've caught #823 .

chuckha: ahhh ok got it 👍 totally approve! thank you for explaining!

pkg/cloud/aws/services/ec2/instances.go
sethp-nr: Good question – I had to check, but a few lines before this we're doing

```
	if s.scope.Network().APIServerELB.DNSName == "" {
		return nil, awserrors.NewFailedDependency(
			errors.New("failed to run controlplane, APIServer ELB not available"),
		)
	}
```

So if the Network were nil at this point, it would've already panic'd there (everything else is a value & not a pointer).

That said, I'm going to look a little further into whether Network might be nil anyway, since a panic seems like the wrong response if so.

sethp-nr: It looks like Network() is a reference into the scope itself that'll always be valid, so we're good on that front.

chuckha:
oops,

/lgtm

PR #868, opened 2019-07-02
Prefix templated nodes with cluster names

liztio:

I often use the examples to create multiple clusters in one management cluster. This is a quality-of-life improvement so I don't have to rename machines and deployments to keep them from conflicting with each other.

**Release note**:

```release-note
NONE
```


detiber:
Should the other example files be updated for this as well?

liztio:
@detiber doh, updated

PR #870, opened 2019-07-03
[0.3] Use cluster label selector for machine list

### LINKED ISSUE #869: [0.3] Adding a new control plane machine is potentially broken if you have multiple clusters
/kind bug

- Create multiple clusters, each with its own machines
- Try to add a 2nd control plane machine to a cluster
- CAPA thinks it's not a node join, so it does `kubeadm init`

- Node join / `kubeadm join`

I will be fixing this in master with #779.

For the `release-0.3` branch, the problem is https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1e7225b91241a5a5a4be4e431999fbec5b76ad6c/pkg/cloud/aws/actuators/machine/actuator.go#L161 does not filter machines by the cluster name label selector.

**Environment:**

- Cluster-api-provider-aws version: 0.3.x

ncdc:
Make sure the machine actuator's Create() is using the cluster label
selector to list machines for the specific cluster; otherwise,
isNodeJoin might return false and incorrectly have an additional control
plane machine run `kubeadm init` instead of `kubeadm join`.

Fixes #869 

**Release note**:

```release-note
Adding additional control plane machines when you have multiple clusters should now correctly join them to the control plane, instead of creating a new, separate control plane.
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine/actuator.go
+++ b/pkg/cloud/aws/actuators/machine/actuator.go
@@ -158,7 +158,7 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
 
 	ec2svc := ec2.NewService(scope.Scope)
 
-	clusterMachines, err := scope.MachineClient.List(v1.ListOptions{})
+	clusterMachines, err := scope.MachineClient.List(v1.ListOptions{LabelSelector: fmt.Sprintf("%s=%s", clusterv1.MachineClusterLabelName, cluster.Name)})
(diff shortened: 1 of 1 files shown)


vincepri:
```
gentestmain: ParseFile("pkg/cloud/aws/actuators/machine/actuator.go"): pkg/cloud/aws/actuators/machine/actuator.go:161:149: missing ',' in argument list (and 10 more errors)
```

ncdc:
Fixed

PR #872, opened 2019-07-03
Rev minimum version of Bazel to match requirements

detiber:

Currently the docker-push target requires Bazel 0.25.0+, but the bazel minimum version check is set to 0.23.0+.

**Release note**:
```release-note
NONE
```


detiber:
This requires updating the kubekins images for the prow jobs: https://github.com/kubernetes/test-infra/pull/13319

PR #879, opened 2019-07-09
Add additionalUserDataFiles to ClusterSpec

### LINKED ISSUE #847: Support laying down arbitrary files on disk
/kind feature

Some Kubernetes API server features require additional configuration files.
Add support to the API to be able to lay down these files prior to kubeadm initialization.
cloud-init has a files section that can be utilised to achieve this, but will need to be wary of exceeding the 16KB userdata limit (see #846)

**Environment:**

vincepri:

This PR adds a new optional field to the Cluster spec called `additionalUserDataFiles`. Users might use this field to add files to each Machine upon creation.

Fixes #847

**Release note**:

```release-note

```


pkg/apis/awsprovider/v1alpha1/awsclusterproviderconfig_types.go
```diff
@@ -57,6 +58,10 @@ type AWSClusterProviderSpec struct {
 	// ClusterConfiguration holds the cluster-wide information used during a
 	// kubeadm init call.
 	ClusterConfiguration kubeadmv1beta1.ClusterConfiguration `json:"clusterConfiguration,omitempty"`
+
+	// AdditionalUserDataFiles specifies extra files to be passed to Machines upon creation.
+	// +optional
+	AdditionalUserDataFiles []userdata.Files `json:"additionalUserDataFiles,omitempty"`
```
ncdc: I know this wasn't specced out anywhere, but I wonder if we want to do this at the machine level vs cluster (or both)? I would imagine that we wouldn't need (or even want) e.g. a client certificate used to authenticate to a webhook service to be placed on all machines.

vincepri: maybe both is the way to go

vincepri: Added and updated the PR

pkg/cloud/aws/services/ec2/instances.go
```diff
@@ -228,6 +228,10 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 			}
 
 			userData, err = userdata.NewJoinControlPlane(&userdata.ControlPlaneJoinInput{
+				AdditionalFiles: append(
+					s.scope.ClusterConfig.AdditionalUserDataFiles,
```
sethp-nr: Do we write the cluster back when we close the scope? I'm wondering if this will add the machine's config to the cluster's.

vincepri: It won't, `append` creates a new slice from the one passed in

sethp-nr: It makes a new slice, but the backing array may be the same if it's large enough: https://play.golang.org/p/VWESAzrxGmB

sethp-nr: I guess that would go the other way, though – unless we change the len of the `scope.ClusterConfig.AdditionalUserDataFiles` we won't write back the changes, but future appends to that slice could clobber what we've handed off to the `ControlPlaneJoinInput`.

I guess in this case it's ok, since it immediately consumes the data and puts it into a string. Still might be safer to make a new slice from both :man_shrugging:

vincepri: I think it's safe to use with append, I don't think there is any case where this slice has cap != len

ncdc:
@vincepri I can do the 0.3 backport if you haven't started yet

vincepri:
I haven't, thank you!

PR #881, opened 2019-07-09
Avoid infinite APIEndpoints

vincepri:
Signed-off-by: Vince Prignano <vincepri@vmware.com>

**What this PR does / why we need it**:
This PR fixes an issue where the first reconciliation loop assigns APIEndpoints.Host to an empty string because the ELB might not be ready, and it'll start appending after that. The logic here changes to always have a single APIEndpoint and update if needed.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #880

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #877, opened 2019-07-08
Cleanup services/userdata package

vincepri:
Signed-off-by: Vince Prignano <vincepri@vmware.com>

**What this PR does / why we need it**:
This PR cleans up the `userdata` package and simplifies handling a large amount of files that need to be written to disk. In a future PR, this work is going to be used to store extra files on disk.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Related to #847

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #875, opened 2019-07-08
Fix ELB panic on deletion

vincepri:
Signed-off-by: Vince Prignano <vincepri@vmware.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #874 

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #855, opened 2019-06-28
Match ncdc owner alias membership with k/org

ncdc:
This matches my entries for GitHub permissions in k/org.

Signed-off-by: Andy Goldstein <goldsteina@vmware.com>

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

_Please confirm that if this PR changes any image versions, then that's the sole change this PR makes._

**Release note**:

```release-note

```

PR #887, opened 2019-07-10
0.3 backports

ncdc:
Backports of:

- #855
- #868
- #872
- #875
- #877
- #879
- #881

I think the one notable one that isn't in here now is "fix: respect cri socket if set (#863)" and it wasn't a clean cherry-pick. I can look into backporting it too if there is interest? cc @sethp-nr


sethp-nr:
Mentioned in slack, but for completeness: I'm :+1: on skipping #863

PR #888, opened 2019-07-10
Go modules

### LINKED ISSUE #837: Switch to go modules
/kind feature

Would make #659 obsolete

vincepri:

Fixes #837 

**Release note**:

```release-note

```


WORKSPACE
```diff
@@ -64,7 +64,7 @@ versions.check(
     maximum_bazel_version = "1.0.0",
 )  # fails if not within range
 
-load("@io_bazel_rules_go//go:deps.bzl", "go_rules_dependencies", "go_register_toolchains")
+load("@io_bazel_rules_go//go:deps.bzl", "go_register_toolchains", "go_rules_dependencies")
```
detiber: Why the change in ordering here? I'm not sure if the ordering is significant, but the previous ordering matches the setup instructions for rules_go: https://github.com/bazelbuild/rules_go#setup

WORKSPACE
```diff
@@ -26,8 +26,8 @@ load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 # )
 git_repository(
     name = "io_bazel_rules_go",
-    remote = "https://github.com/bazelbuild/rules_go.git",
     commit = "f2373c9fbd09586d8e591dda3c43d66445b2d7ca",
+    remote = "https://github.com/bazelbuild/rules_go.git",
```
detiber: We *might* be able to remove this hack now that rules_go v0.18.7 has been released

go.vendor
```diff
@@ -0,0 +1 @@
+sigs.k8s.io/cluster-api
```
detiber: What's the significance of this file?

WORKSPACE
vincepri: I think it’s just reordered because of the vscode bazel extention

go.vendor
vincepri: It’s to fully vendor cluster-api repository, see https://github.com/kubernetes-sigs/cluster-api/pull/1134

hack/BUILD.bazel
```diff
@@ -9,3 +11,21 @@ exports_files(glob([
...
+    deps = [
+        "//vendor/k8s.io/code-generator/cmd/client-gen:go_default_library",
+        "//vendor/k8s.io/code-generator/cmd/conversion-gen:go_default_library",
+        "//vendor/k8s.io/code-generator/cmd/deepcopy-gen:go_default_library",
+        "//vendor/k8s.io/code-generator/cmd/defaulter-gen:go_default_library",
+        "//vendor/k8s.io/code-generator/cmd/informer-gen:go_default_library",
+        "//vendor/k8s.io/code-generator/cmd/lister-gen:go_default_library",
+        "//vendor/k8s.io/code-generator/cmd/register-gen:go_default_library",
+        "//vendor/k8s.io/code-generator/cmd/set-gen:go_default_library",
+        "//vendor/sigs.k8s.io/controller-tools/cmd/controller-gen:go_default_library",
```
chuckha: This is out of scope for this PR but I'm interested in what others think.

The benefits to keeping source code in tree as I see them are: being able to run the commands across platforms and updating versions of these tools is as easy as checking in a new version to vendor/. Folks developing will pull in changes and have the new version, no effort on their part. What else am I missing?

The downsides as I see it are that go mod doesn't support vendoring code that isn't directly referenced and we end up with `hack/tools.go` + the vendor copy, which will copy in the entire project and not just the necessary files.

So my question: Why are we treating controller-tools/code-generator in a special way from any other external dependency?

detiber: One reason for keeping the vendor directory around is that we don't have to worry about dependencies being renamed (case sensitivity), moved, or referenced tags/hashes no longer being present.

detiber: regarding controller-tools and other utilities referenced from `hack/tools.go`, we ideally want to use the versions that are compatible with the versions of controller-runtime and controller-tools that we are using. If we don't leverage the vendored versions, then we have to manually keep those in sync another way. That said, with the way that gomod handles dependencies, we may not be ensuring that they are compatible versions today :(

chuckha: I recognize things change as a theoretical problem but in practice I have never found it to be a problem. It would be good to understand the decision made to keep vendor and what problems we're solving for.

chuckha: Since we have most dependencies in bazel, could we use bazel to keep these dependencies in sync?

detiber: @chuckha I think it's likely to be more of an issue at the current point in time due to changes made between controller-runtime 1.x and the 2.x betas we are currently relying on.

I'd be fine with defering to Bazel to manage the dependencies for the generation tooling, we'd just need to document keeping those dependencies updated along with controller-tools.

chuckha: ok, I'm not saying we should, I'm mostly thinking about this from the problems I'm about to slam my head into with the bootstrap controller provider. I really appreciate the discussion, thank you

WORKSPACE
vincepri: Do you want me to do it as part of this PR?

detiber: No, I can do that as a followup

PR #902, opened 2019-07-15
Document custom ssh key

dlipovetsky:
Document how to use an SSH key already in AWS to deploy a cluster.


sethp-nr:
Nice catch, thank you!
/lgtm

PR #904, opened 2019-07-16
Convert Status.Ready to plain bool

vincepri:

Fixes #

**Release note**:

```release-note

```


detiber:
lgtm, but want to give @ncdc a chance to look over.

pkg/apis/infrastructure/v1alpha2/awsmachine_types.go
```diff
@@ -83,7 +83,8 @@ type AWSMachineSpec struct {
 // AWSMachineStatus defines the observed state of AWSMachine
 type AWSMachineStatus struct {
 	// Ready is true when the provider resource is ready.
-	Ready *bool `json:"ready,omitempty"`
+	// +optional
```
ncdc: I don't think you can call this optional unless it's a pointer - remove this?

vincepri: It seems controller tools allows it, if it’s not marked as optional, it’ll be marked as required in the CRD

ncdc: Does that mean it will error if it's not in the json/yaml when you POST it?

vincepri: Yes, the validation will fail, similar to what we've seen with ObjectMeta

ncdc: k, thanks.

PR #905, opened 2019-07-17
Restructure repository / cleanup

vincepri:

This PR is an effort to cleanup the repository structure, in particular:
- Removes the `aws` portion from `pkg/cloud/aws/...` packages, which seems redundant.
- Gives machine and cluster scopes their own package under `pkg/cloud/scope`.
- Moves the control plane locker helper functions under `ClusterScope`.

Fixes #

**Release note**:

```release-note

```


detiber:
@vincepri any reason to not switch up to the kubebuilder v2 directory layout in the process as well? I don't see any particular reason we need to keep /pkg around.

vincepri:
I am not sure if we should remove pkg/..., maybe we can promote the controllers up later and match kubebuilder v2, but it seems like a much bigger change

vincepri:
@ncdc @detiber Is this good to go?

Issue #899, opened 2019-07-15
Unable to delete machine when ELB is not present


williamsandrew:
/kind bug

**What steps did you take and what happened:**

I attempted to delete a cluster but was unable to] because the cluster is being affected by this (now fixed) bug: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/893. However, I'm also seeing an error on this same cluster where a machine cannot be deleted because the controller cannot find the associated load balancer. Logs:

```
I0715 16:01:15.657019       1 machine_controller.go:115] Reconciling Machine "control-plane-andrewwi2-test-0"
I0715 16:01:15.657144       1 actuator.go:417] [machine-actuator] "level"=0 "msg"="Checking if machine exists in cluster"  "cluster-name"="andrewwi2-test" "machine-name"="control-plane-andrewwi2-test-0" "machine-namespace"="4d655f91724cacd06f950138ccaf253d33375a01"
I0715 16:01:15.665271       1 instances.go:105] [machine-actuator]/cluster.k8s.io/v1alpha1/4d655f91724cacd06f950138ccaf253d33375a01/andrewwi2-test/control-plane-andrewwi2-test-0 "level"=2 "msg"="Looking for instance by id"  "instance-id"="i-06b690c17ec9a6655"
I0715 16:01:15.824534       1 actuator.go:442] [machine-actuator] "level"=0 "msg"="Found instance for machine"  "instance"={"id":"i-06b690c17ec9a6655","instanceState":"running","type":"m4.large","subnetId":"subnet-06064113669d43b71","imageId":"ami-00c07cb7da8d91b7a","keyName":"gophoton-global-keypair","securityGroupIds":["sg-02aec52f68d96ff42","sg-022fed0f317d3b04a","sg-0ed7e27932b2f0095"],"iamProfile":"control-plane.cluster-api-provider-aws.sigs.k8s.io","privateIp":"10.0.0.61","enaSupport":true,"ebsOptimized":false,"rootDeviceSize":8,"tags":{"Name":"control-plane-andrewwi2-test-0","kubernetes.io/cluster/andrewwi2-test":"owned","sigs.k8s.io/cluster-api-provider-aws/cluster/andrewwi2-test":"owned","sigs.k8s.io/cluster-api-provider-aws/role":"controlplane"}} "machine-name"="control-plane-andrewwi2-test-0" "machine-namespace"="4d655f91724cacd06f950138ccaf253d33375a01"
[...]

PR #908, opened 2019-07-17
Do not update LB attachments or providerID in Exists

detiber:

- Moves the update of LB attachments and providerID from Exists to Update
- Always set providerID if unset.

**Release note**:
```release-note
- Moves the update of LB attachments and providerID from Exists to Update
- Always set providerID if unset.
```

### DIFF
--- a/pkg/cloud/aws/actuators/machine/actuator.go
+++ b/pkg/cloud/aws/actuators/machine/actuator.go
@@ -400,6 +400,12 @@ func (a *Actuator) Update(ctx context.Context, cluster *clusterv1.Cluster, machi
 		return errors.Errorf("failed to ensure tags: %+v", err)
 	}
 
+	scope.MachineStatus.InstanceState = &instanceDescription.State
+
+	if err := a.reconcileLBAttachment(scope, machine, instanceDescription); err != nil {
+		return errors.Errorf("failed to reconcile load balancer attachment: %+v", err)
+	}
+
 	return nil
 }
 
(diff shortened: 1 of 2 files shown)


detiber:
/assign @vincepri @ncdc 
I still need to finish testing this, but would like some initial feedback.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -181,6 +181,11 @@ func (a *Actuator) Create(ctx context.Context, cluster *clusterv1.Cluster, machi
 		return errors.Errorf("failed to reconcile LB attachment: %+v", err)
 	}
 
+	if machine.Spec.ProviderID == nil || *machine.Spec.ProviderID == "" {
+		providerID := fmt.Sprintf("aws:////%s", *scope.MachineStatus.InstanceID)
```
vincepri: Shouldn't we always reconcile the providerID if it's not there? This is to account for the case that the ProviderID didn't get stored for one reason or another, but we retrieved the instance using tags instead.

pkg/cloud/aws/actuators/machine/actuator.go
```diff
@@ -400,6 +405,17 @@ func (a *Actuator) Update(ctx context.Context, cluster *clusterv1.Cluster, machi
 		return errors.Errorf("failed to ensure tags: %+v", err)
 	}
 
+	scope.MachineStatus.InstanceState = &instanceDescription.State
+
+	if err := a.reconcileLBAttachment(scope, machine, instanceDescription); err != nil {
+		return errors.Errorf("failed to reconcile load balancer attachment: %+v", err)
+	}
+
+	if machine.Spec.ProviderID == nil || *machine.Spec.ProviderID == "" {
```
vincepri: Given that this is repeated, can we do it outside here and the Create call, what do you think about moving it outside?

detiber:
@vincepri I moved the updating of providerID to the scope, ptal

ncdc:
@detiber does this fix #899?

detiber:
@ncdc it does not, I'm not able to replicate #899 and not sure exactly how it could be possible to hit that issue per the comment I left there:

> I'm trying to diagnose this issue a bit more and from what I can tell, this shouldn't be hit if the Machine is being deleted, only if the machine is attempting to reconcile without the presence of a load balancer.

PR #909, opened 2019-07-18
Improve AWSMachine reconciliation logic

vincepri:

This PR rewrites the logic ported from the old machine actuator, in particular:
- Reduces the amount of calls to AWS to describe the instance.
- Removes the `Status.InstanceID` in favor of `Spec.ProviderID`.
- Uses `Spec.ProviderID` to determine if to create an instance or describe it.

Fixes #

**Release note**:

```release-note

```


ncdc:
@vincepri after you & I chatted yesterday, Jason informed me today that there's a chance we could have a machine without a provider id but still have an ec2 VM created for it. If the VM creation succeeds and we fail to set the provider id on the machine spec (CAPA crashes, apiserver down, 🤷‍♂), we could be in this situation. So we probably still need to check by tags...

vincepri:
Makes sense! I applied some changes to always call InstanceByTags first and provide a fallback to query by ID

pkg/cloud/scope/machine.go
```diff
@@ -126,14 +128,15 @@ func (m *MachineScope) Region() string {
...
+// GetInstanceID returns the AWSMachine instance id by parsing Spec.ProviderID.
 func (m *MachineScope) GetInstanceID() *string {
-	return m.ProviderMachine.Status.InstanceID
-}
-
-// SetInstanceID sets the AWSMachine instance id.
-func (m *MachineScope) SetInstanceID(v string) {
-	m.ProviderMachine.Status.InstanceID = pointer.StringPtr(v)
+	if pid := m.GetProviderID(); pid != nil && *pid != "" {
+		// Get the ProviderID, and if it's populated, return the last part.
```
ncdc: Please use the ProviderID code from CAPI

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +153,71 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
-		return errors.Errorf("failed to create or get machine: %+v", err)
-	}
-
-	scope.SetInstanceID(i.ID)
-	scope.SetInstanceState(i.State)
-	scope.SetAnnotation("cluster-api-provider-aws", "true")
-
-	if err := r.reconcileLBAttachment(scope, i); err != nil {
-		return errors.Errorf("failed to reconcile LB attachment: %+v", err)
+		return errors.Wrapf(err, "failed to create AWSMachine instance")
```
ncdc: I'm not sure this error message is accurate

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +153,71 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
+			if err != nil {
+				return errors.Wrapf(err, "failed to create AWSMachine instance")
+			}
+		}
 
-	if instance == nil {
-		return false, nil
+		// Set the ProviderID.
+		scope.SetProviderID(fmt.Sprintf("aws:////%s", instance.ID))
+	} else if instance == nil {
```
ncdc: I'd like to try to find a way to make the branching and conditionals easier to follow. I don't have a good idea in my head yet. I'll keep thinking about it.

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +153,71 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
-	controlPlaneMachines := util.GetControlPlaneMachinesFromList(machineList)
-	if len(controlPlaneMachines) == 0 {
-		scope.Info("No control plane machines exist yet - requeuing")
-		return &capierrors.RequeueAfterError{RequeueAfter: waitForControlPlaneMachineExistenceDuration}
-	}
-
-	// Create the Machine.
-	i, err := ec2svc.CreateOrGetMachine(scope)
+	// Look for the instance by tags first.
+	instance, err := ec2svc.InstanceByTags(scope)
```
ncdc: Would it make more sense to check by id first?

pkg/cloud/scope/machine.go
vincepri: What do you mean?

pkg/controller/awsmachine/awsmachine_controller.go
vincepri: Wouldn't that have the same effect as before in case the ID is nil?

pkg/cloud/scope/machine.go
ncdc:
```go
parsed, err := noderefutil.NewProviderID(*pid)
if err != nil {
  // log and return nil?
}
return pointer.StringPtr(parsed.ID())
```

pkg/controller/awsmachine/awsmachine_controller.go
vincepri: I could move it to its own function, but then we'd get into the `getOrCreate` scenario again

pkg/cloud/scope/machine.go
vincepri: Done and rewrote a little

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +154,72 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
+		return errors.Wrapf(err, "failed to parse Spec.ProviderID")
 	}
 
-	// Create the Machine.
-	i, err := ec2svc.CreateOrGetMachine(scope)
-	if err != nil {
-		return errors.Errorf("failed to create or get machine: %+v", err)
+	// If the Spec.ProviderID is not set and the instance is not found,
+	// create the AWSMachine.
+	if pid == nil && instance == nil {
```
detiber: This seems to rely on pid == nil in the case that noderefutil.NewProviderID returns an error, while that is likely always going to be the case it seems to go against Go convention of not relying on any particular value of a return value if an error is returned.

pkg/cloud/services/ec2/instances.go
```diff
@@ -81,10 +81,6 @@ func (s *Service) InstanceIfExists(id *string) (*v1alpha2.Instance, error) {
 
 	input := &ec2.DescribeInstancesInput{
 		InstanceIds: []*string{id},
-		Filters: []*ec2.Filter{
-			filter.EC2.VPC(s.scope.VPC().ID),
-			filter.EC2.InstanceStates(ec2.InstanceStateNamePending, ec2.InstanceStateNameRunning),
-		},
```
detiber: Removing this filter is likely to result in issues with trying to recreate a Machine with the same name after deletion if the backing instance is still in a Terminated state.

pkg/controller/awsmachine/awsmachine_controller.go
detiber: I think a `getOrCreate` scenario in this file is probably fine, but I'd like to avoid having something buried in the service code to make generating events cleaner.

pkg/controller/awsmachine/awsmachine_controller.go
vincepri: I can move this create block in the `err == noderefutil.ErrEmptyProviderID`, although the nesting of if conditions here is definitely a concern for readability

pkg/cloud/services/ec2/instances.go
vincepri: The logic of checking the state of a Machine moved up now, where we set the error state if the Machine is not running or pending.

detiber: Right, so if a user deletes a Machine and the backing instance is terminated and then the user tries to create a new Machine with the same name, this will now return the terminated instance and result in the new Machine returning an error rather than creating a new instance.

vincepri: In case a Machine is in a terminated state, after the we check the ProviderID, if we don't have one (which means we can't query by ID), we query by tags using `InstanceByTags`, which still has the filtering behavior you'd expect. We should probably rename the method to something like `GetRunningInstanceByTags`

vincepri:
@ncdc @detiber Refactored the code in a smaller `getOrCreate` function

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +155,92 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
+		return err
 	}
 
-	controlPlaneMachines := util.GetControlPlaneMachinesFromList(machineList)
-	if len(controlPlaneMachines) == 0 {
-		scope.Info("No control plane machines exist yet - requeuing")
-		return &capierrors.RequeueAfterError{RequeueAfter: waitForControlPlaneMachineExistenceDuration}
+	// Set an error message if we couldn't find the Machine.
+	if instance == nil {
+		scope.SetErrorMessage(errors.New("EC2 instance cannot be found"))
```
detiber: This should set both ErrorMessage and ErrorReason

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +155,92 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
-		return errors.Errorf("failed to create or get machine: %+v", err)
-	}
+	// Make sure Spec.ProviderID is always set.
+	scope.SetProviderID(fmt.Sprintf("aws:////%s", instance.ID))
 
-	scope.SetInstanceID(i.ID)
-	scope.SetInstanceState(i.State)
+	// Proceed to reconcile the AWSMachine state.
+	scope.SetInstanceState(instance.State)
 	scope.SetAnnotation("cluster-api-provider-aws", "true")
```
detiber: might want to add a TODO here to remove this when it is no longer needed for `clusterctl`

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +155,92 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
 }
```
detiber: As I was trying to figure out how the error messages are written out, since they are returned from `getOrCreate` and `reconcile`, it looks like we are using plain `klog` to write them out instead of using the logger from the scope above

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -153,116 +155,92 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
 	scope.SetAnnotation("cluster-api-provider-aws", "true")
 
-	if err := r.reconcileLBAttachment(scope, i); err != nil {
+	switch instance.State {
+	case infrav1.InstanceStateRunning:
+		scope.Info("Machine instance is running", "instance-id", *scope.GetInstanceID())
+	case infrav1.InstanceStatePending:
+		scope.Info("Machine instance is pending", "instance-id", *scope.GetInstanceID())
+	default:
+		scope.SetErrorMessage(errors.Errorf("EC2 instance state %q is unexpected", instance.State))
```
detiber: This should set both ErrorMessage and ErrorReason

pkg/cloud/scope/machine.go
ncdc: This still has the string checking?

pkg/controller/awsmachine/awsmachine_controller.go
vincepri: I haven't found a good ErrorReason upstream that would match this, do you think we should add new ones?

pkg/cloud/scope/machine.go
vincepri: I'd avoid to use the noderefutil here if possible given that the GetInstanceID might be called frequently

pkg/controller/awsmachine/awsmachine_controller.go
detiber: Possibly `UpdateMachineError` or we just add a new one

pkg/controller/awsmachine/awsmachine_controller.go
vincepri: Switched those up to use the scope logger

pkg/controller/awsmachine/awsmachine_controller.go
vincepri: Sounds good to me

pkg/cloud/scope/machine.go
ncdc: What is your concern?

vincepri: I guess it doesn't matter, but it seemed small enough that it wasn't worth to change

vincepri:
Addressed all the comments, should be good to go

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -136,133 +138,129 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
 		if requeueErr, ok := errors.Cause(err).(capierrors.HasRequeueAfterError); ok {
-			klog.Infof("Reconciliation for AWSMachine %q in namespace %q asked to requeue: %v", awsm.Name, awsm.Namespace, err)
+			scope.Info("Reconciliation asked to requeue", "after", requeueErr.GetRequeueAfter())
 			return reconcile.Result{Requeue: true, RequeueAfter: requeueErr.GetRequeueAfter()}, nil
 		}
+		scope.Error(err, "Failed to reconcile AWSMachine")
 		return reconcile.Result{}, err
 	}
 
+	// TODO(vincepri): Handle deletion and finalizers here.
```
detiber: Deletion should probably be handled prior to calling `r.reconcile`

pkg/controller/awsmachine/awsmachine_controller.go
```diff
@@ -136,133 +138,129 @@ func (r *ReconcileAWSMachine) Reconcile(request reconcile.Request) (reconcile.Re
...
+
 	return reconcile.Result{}, nil
 }
 
 func (r *ReconcileAWSMachine) reconcile(ctx context.Context, scope *scope.MachineScope) error {
```
detiber: Considering this method only handles create or update, should this be renamed `createOrUpdate` instead of `reconcile`?

vincepri: Isn't reconcile createOrUpdate everywhere else? :) We're using the same pattern elsewhere as well, having reconcile/delete separate

pkg/controller/awsmachine/awsmachine_controller.go
vincepri: SGTM, going to follow-up in another PR

pkg/controller/awsmachine/awsmachine_controller.go
detiber: That was the actuator interface, since we are aligning more with the `kubebuilder` model, I think it is likely to cause confusion with having a public `Reconcile` method that handles create/update/delete that calls a private `reconcile` method that only does create/update.

vincepri: Gonna open an issue to address this either in documentation or a refactor

vincepri:
@detiber ptal

PR #910, opened 2019-07-18
Additional Cluster Events

### LINKED ISSUE #816: Add failure events to the cluster object
/kind feature

Right now, when we create a cluster in CAPA, we can see the progress of the cluster being created using the events that are part of the cluster object. This shows up when we do `kubectl describe cluster`. The events which show up are only for the success cases like, a VPC got created, subnet got created, etc. It would be good if this shows failure cases as well like SSH key not found, VPC creation failed due to out of quota error, etc. This will help the users get a better understanding of the progress. Right now, the only way to see these errors is by looking at the logs which is not ideal.

**Environment:**
- Cluster-api-provider-aws version: 0.2.1
- Kubernetes version: (use `kubectl version`): 1.14.1

detiber:
- Adds additional Cluster-related Events

Fixes #816 

**Release note**:
```release-note
TODO
```

### DIFF
--- a/pkg/apis/awsprovider/v1alpha1/types.go
+++ b/pkg/apis/awsprovider/v1alpha1/types.go
@@ -364,10 +364,22 @@ func (i *IngressRule) Equals(o *IngressRule) bool {
 		}
 	}
 
-	return i.Description == o.Description &&
-		i.FromPort == o.FromPort &&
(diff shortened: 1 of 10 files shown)


ncdc:
Code changes LGTM so far!

PR #916, opened 2019-07-22
Implement backoff and retry for AWS eventual consistency

### LINKED ISSUE #836: Add a retry to resource tagging
/kind bug

It is hard to reproduce this, but if a resource fails to tag after creation we will orphan the resource and the cluster will likely encounter problems.

We have reports of security groups failing to tag after being created (thanks amazon) and then on the subsequent try, cluster-api cannot create a duplicate security group but also can't use the existing one as it is improperly tagged. We should retry tags once or twice on create.

Optionally, and I think others should weigh in here, like @detiber, if we find a resource with a name we expect but it has no tags should we try and add the tags to the existing resource if is the resource we expect? For example on security groups, we should check if the data matches what we expect, then tag it assuming the tags after create failed?

cc @Lokicity @vivgoyal

### LINKED ISSUE #851: AWS controller recreates VPC when setting attributes fails
/kind bug

Attempt to create a new Cluster using CAPA. When setting VPC attribute fails, the CAPA repeatedly created a new VPC until it hit the limits on each reconcile iteration. 

When setting the VPC attribute fails, it must delete the created VPC to retain idempotency. 

**Environment:**
This can be easily reproduced by removing permission SetVPCAttributes from the controller managed-policy

### LINKED ISSUE #890: Failure to tag security group prevents cluster reconciliation
/kind bug

1. Followed our usual process for pivoting a new cluster
2. Part way through, the EC2 API's eventual consistency whacked us with a 404 trying to tag the security group CAPA had just created (see log message below)
3. From then on out, CAPA couldn't find the security group, so it tried to create it again, getting a conflict error on the SG's "group name" (which is different from the Name tag, because AWS). That error kept us from getting past the `reconcileNetwork` step to creating the load balancer & everything later

It's sort of similar to #851, but with different consequences because security groups have a name-based identity.

I think for security groups, we'll have a better time if we don't filter by cluster tag here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/46d5f38/pkg/cloud/aws/services/ec2/securitygroups.go#L165

That said, when we added that tag by hand, we then saw a different error trying to tag the resource:

```
E0711 14:17:57.002589       1 controller.go:134] Error reconciling cluster object gallant-agnesi; failed to reconcile network for cluster "gallant-agnesi": failed to ensure tags on security group "sg-SGIDf": failed to tag resource "" in cluster "gallant-agnesi": InvalidParameterValue: Value ( null ) for parameter resourceId is invalid. Null/empty value for resourceId is invalid
```

Some log messages (I found "the security group [extremely specific ID] does not exist" pretty funny):

```
E0711 14:08:11.908350       1 controller.go:134] Error reconciling cluster object gallant-agnesi; failed to reconcile network for cluster "gallant-agnesi": failed to tag security group "gallant-agnesi-lb" in vpc "vpc-VPCID": InvalidGroup.NotFound: The security group 'sg-SGID' does not exist
	status code: 400, request id: abd9cce5-5aa6-439c-8953-040b09e3f86c
...
E0711 14:13:53.275716       1 controller.go:134] Error reconciling cluster object gallant-agnesi; failed to reconcile network for cluster "gallant-agnesi": failed to create security group "gallant-agnesi-lb" in vpc "vpc-VPCID": InvalidGroup.Duplicate: The security group 'gallant-agnesi-lb' already exists for VPC 'vpc-VPCID'
```

**Environment:**

- Cluster-api-provider-aws version: issue looks the same on master

detiber:
- Implements backoff and retry for AWS eventual consistency

Fixes #890
Fixes #851 
Fixes #836 

**Release note**:
```release-note
TODO
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/gateways.go
+++ b/pkg/cloud/aws/services/ec2/gateways.go
@@ -26,6 +26,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/converters"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/filter"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/awserrors"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/wait"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/tags"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
 )
@@ -59,12 +60,15 @@ func (s *Service) reconcileInternetGateways() error {
 	s.scope.VPC().InternetGatewayID = gateway.InternetGatewayId
 
 	// Make sure tags are up to date.
-	err = tags.Ensure(converters.TagsToMap(gateway.Tags), &tags.ApplyParams{
-		EC2Client:   s.scope.EC2,
-		BuildParams: s.getGatewayTagParams(*gateway.InternetGatewayId),
-	})
-
-	if err != nil {
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+		if err := tags.Ensure(converters.TagsToMap(gateway.Tags), &tags.ApplyParams{
+			EC2Client:   s.scope.EC2,
+			BuildParams: s.getGatewayTagParams(*gateway.InternetGatewayId),
+		}); err != nil {
+			return false, err
+		}
+		return true, nil
+	}, awserrors.InternetGatewayNotFound); err != nil {
 		record.Warnf(s.scope.Cluster, "FailedTagInternetGateway", "Failed to tag managed Internet Gateway %q: %v", gateway.InternetGatewayId, err)
 		return errors.Wrapf(err, "failed to tag internet gateway %q", *gateway.InternetGatewayId)
 	}
@@ -121,12 +125,35 @@ func (s *Service) createInternetGateway() (*ec2.InternetGateway, error) {
 	}
 
 	s.scope.Info("Created internet gateway for VPC", "vpc-id", s.scope.VPC().ID)
-	_, err = s.scope.EC2.AttachInternetGateway(&ec2.AttachInternetGatewayInput{
-		InternetGatewayId: ig.InternetGateway.InternetGatewayId,
-		VpcId:             aws.String(s.scope.VPC().ID),
-	})
 
-	if err != nil {
+	tagParams := s.getGatewayTagParams(*ig.InternetGateway.InternetGatewayId)
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+		if err := tags.Apply(&tags.ApplyParams{
+			EC2Client:   s.scope.EC2,
+			BuildParams: tagParams,
+		}); err != nil {
+			return false, err
+		}
+		return true, nil
+	}, awserrors.InternetGatewayNotFound); err != nil {
+		record.Warnf(s.scope.Cluster, "FailedTagInternetGateway", "Failed to tag managed Internet Gateway %q: %v", *ig.InternetGateway.InternetGatewayId, err)
+		return nil, errors.Wrapf(err, "failed to tag internet gateway %q", *ig.InternetGateway.InternetGatewayId)
+	}
+
+	// Update the tags, so that when ig.InternetGateway is returned it has the
+	// latest tag data rather than returning empty tags.
+	ig.InternetGateway.Tags = converters.MapToTags(v1alpha1.Build(tagParams))
+
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+		if _, err := s.scope.EC2.AttachInternetGateway(&ec2.AttachInternetGatewayInput{
+			InternetGatewayId: ig.InternetGateway.InternetGatewayId,
+			VpcId:             aws.String(s.scope.VPC().ID),
+		}); err != nil {
+			return false, err
+		}
+		return true, nil
+	}, awserrors.InternetGatewayNotFound); err != nil {
+		record.Warnf(s.scope.Cluster, "FailedAttachInternetGateway", "Failed to attach managed Internet Gateway %q to vpc %q: %v", *ig.InternetGateway.InternetGatewayId, s.scope.VPC().ID, err)
 		return nil, errors.Wrapf(err, "failed to attach internet gateway %q to vpc %q", *ig.InternetGateway.InternetGatewayId, s.scope.VPC().ID)
 	}
 
--- a/pkg/cloud/aws/services/ec2/natgateways.go
+++ b/pkg/cloud/aws/services/ec2/natgateways.go
@@ -25,6 +25,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/apis/awsprovider/v1alpha1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/converters"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/filter"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/awserrors"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/services/wait"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/aws/tags"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
@@ -58,12 +59,15 @@ func (s *Service) reconcileNatGateways() error {
 
 		if ngw, ok := existing[sn.ID]; ok {
 			// Make sure tags are up to date.
-			err := tags.Ensure(converters.TagsToMap(ngw.Tags), &tags.ApplyParams{
-				EC2Client:   s.scope.EC2,
-				BuildParams: s.getNatGatewayTagParams(*ngw.NatGatewayId),
-			})
-
-			if err != nil {
+			if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+				if err := tags.Ensure(converters.TagsToMap(ngw.Tags), &tags.ApplyParams{
+					EC2Client:   s.scope.EC2,
+					BuildParams: s.getNatGatewayTagParams(*ngw.NatGatewayId),
+				}); err != nil {
+					return false, err
+				}
+				return true, nil
+			}, awserrors.ResourceNotFound); err != nil {
 				record.Warnf(s.scope.Cluster, "FailedTagNATGateway", "Failed to tag managed NAT Gateway %q: %v", *ngw.NatGatewayId, err)
 				return errors.Wrapf(err, "failed to tag nat gateway %q", *ngw.NatGatewayId)
 			}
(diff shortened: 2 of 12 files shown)


sethp-nr:
I was imagining a slightly different approach: if we wrote down the VPC id as soon as we got it on the cluster resource & *waves hands* made sure that a re-queue at the Cluster level did the tag retry-ing, we'd be durable against a slightly broader class of failures. It's also a pattern that (I think) generalizes across resource types, as long as my handwaving is not too extreme.

Some sticky points are restructuring the reconciliation to be ok with lots of little updates and figuring out how to communicate intention across loops (here,  maybe an optional `tags` portion of the VPC spec?), but maybe it's interesting?

sethp-nr:
Also, I assume you're planning on going further, but as it currently stands this change addresses #851 but not #890.

detiber:
> I was imagining a slightly different approach: if we wrote down the VPC id as soon as we got it on the cluster resource & waves hands made sure that a re-queue at the Cluster level did the tag retry-ing, we'd be durable against a slightly broader class of failures. It's also a pattern that (I think) generalizes across resource types, as long as my handwaving is not too extreme.

I thought about that approach as well, and I think there is definitely value of going down that path as well, but I think the backoff/retries approach will be needed regardless. My worry is that if we rely too much on the update of the Cluster object, then we do not have any recourse if we are unable to write to the API server for any reason. It may also cause us to have to record more data as part of the Spec as opposed to status, since we need to be able to recreate Status for DR-type scenarios (or pivot).

> Some sticky points are restructuring the reconciliation to be ok with lots of little updates and figuring out how to communicate intention across loops (here, maybe an optional tags portion of the VPC spec?), but maybe it's interesting?

Over time I'd like to see us tackle this as a two-pronged effort, with both recording the IDs and backoffs/retries to guard against most of the failure scenarios. 

Another option longer term is that we could also create a pre-defined [client token](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/Run_Instance_Idempotency.html#client-tokens) that we use for requests, which we persist to the cluster-api object prior to performing any AWS actions that we can use for API calls that support it, which would then ensure that we would get the same result (assuming the same input params) even if tagging of the resource fails.

> Also, I assume you're planning on going further, but as it currently stands this change addresses #851 but not #890.

Indeed, I just wanted to make this work visible as I iterated on it.

sethp-nr:
Yeah, agree 100%. We've seen more trouble with writing to AWS than to the API server, but we have only begun to stress our clusters. I like your two-pronged idea a lot (& this is decidedly the shorter prong, if I can mix that metaphor in here), as that reaps benefits both in terms of correctness and efficiency (it'll always be faster to retry just the tagging that we're sure just failed vs. reassessing the state of the world relative to the cluster object).

> Another option longer term is that we could also create a pre-defined client token that we use for requests, which we persist to the cluster-api object prior to performing any AWS actions that we can use for API calls that support it, which would then ensure that we would get the same result (assuming the same input params) even if tagging of the resource fails.

I did not know about client tokens! I'm excited to learn more about them :)

pkg/cloud/aws/services/ec2/vpc.go
```diff
@@ -140,25 +158,17 @@ func (s *Service) createVPC() (*v1alpha1.VPCSpec, error) {
 	}
 	s.scope.V(2).Info("Created new VPC with cidr", "vpc-id", *out.Vpc.VpcId, "cidr-block", *out.Vpc.CidrBlock)
 
+	// Record VPC ID as soon as we possibly can, so that we persist the ID in the case of later failure
+	s.scope.VPC().ID = aws.StringValue(out.Vpc.VpcId)
```
detiber: @sethp-nr I believe this addresses your concern about not recording the VPC ID in the case of later failure.

detiber: After further testing, if we persist the vpc ID without tagging, subsequent reconciles will treat the VPC as an "unmanaged VPC", I'm going to add a TODO and notes, but will remove this for now, since this has the potential to create hopelessly broken clusters, which may just be worse than orphaning a VPC.

detiber: @vincepri as we refactor for v1alpha2, we likely want to keep this issue in mind. We should have a way to differentiate between managed and unmanaged VPCs based on more than just tagging.

ncdc: Could we add a field that indicates it was created by CAPA and set it here too?

detiber: @ncdc I was thinking we could potentially update the current Cluster.Spec.ProviderSpec.NetworkSpec to be either a ManagedNetworkSpec or a UnmanagedNetworkSpec (maybe ExternalNetworkSpec), This would also potentially simplify config for users providing an UnmanagedNetworkSpec or ExternalNetworkSpec to avoid exposing them to fields that are not applicable to them.

detiber: That said, it would definitely complicate some of the downstream logic for consuming Managed vs Unmanaged NetworkSpecs...

ncdc: We could have a helper function that gets a generic network spec from the unioned bit (e.g. pick either ...networkSpec.Managed or ...networkSpec.Unmanaged) and use that helper everywhere?

detiber:
@ncdc Could use another look at this. I plan on revisiting a bit on Monday to see if I can build more consistency into the approach vs what I have now as well.

ncdc:
@detiber I may have a few minutes today. If not, first thing Monday.

detiber:
@sethp-nr I've gotten a bit further along here, would love feedback on the current state

pkg/cloud/aws/services/ec2/gateways.go
```diff
@@ -121,12 +127,37 @@ func (s *Service) createInternetGateway() (*ec2.InternetGateway, error) {
...
+			return false, err
+		}
+		return true, nil
+	}
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), tag, awserrors.InternetGatewayNotFound); err != nil {
+		record.Warnf(s.scope.Cluster, "FailedTagInternetGateway", "Failed to tag managed Internet Gateway %q: %v", *ig.InternetGateway.InternetGatewayId, err)
+		return nil, errors.Wrapf(err, "failed to tag internet gateway %q", *ig.InternetGateway.InternetGatewayId)
+	}
+
+	ig.InternetGateway.Tags = converters.MapToTags(v1alpha1.Build(tagParams))
```
ncdc: What's this for?

pkg/cloud/aws/services/ec2/natgateways.go
```diff
@@ -165,21 +172,32 @@ func (s *Service) createNatGateway(subnetID string) (*ec2.NatGateway, error) {
...
-	if err != nil {
+	}
+	var out *ec2.CreateNatGatewayOutput
+	create := func() (bool, error) {
+		if out, err = s.scope.EC2.CreateNatGateway(createInput); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), create, awserrors.InvalidSubnet); err != nil {
```
ncdc: Is this the only valid retryable error? What about the other ones related to invalid subnets (there are 2 or 3 more)? Where did you get your list from (I had trouble finding anything more than the generic/common list of errors)?

pkg/cloud/aws/services/ec2/vpc.go
```diff
@@ -140,26 +152,36 @@ func (s *Service) createVPC() (*v1alpha1.VPCSpec, error) {
...
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), tag, awserrors.VPCNotFound); err != nil {
+		record.Warnf(s.scope.Cluster, "FailedTagVPC", "Failed to tag managed VPC %q: %v", *out.Vpc.VpcId, err)
 	}
 
 	return &v1alpha1.VPCSpec{
 		ID:        *out.Vpc.VpcId,
 		CidrBlock: *out.Vpc.CidrBlock,
 		Tags:      v1alpha1.Build(tagParams),
-	}, nil
+	}, err
```
ncdc: Walk me through what happens if we return a valid VPCSpec with an error?

pkg/cloud/aws/services/ec2/vpc.go
```diff
@@ -177,7 +199,7 @@ func (s *Service) deleteVPC() error {
 	_, err := s.scope.EC2.DeleteVpc(input)
 	if err != nil {
 		// Ignore if it's already deleted
-		if code, ok := awserrors.Code(err); code == "InvalidVpcID.NotFound" && ok {
+		if code, ok := awserrors.Code(err); code == awserrors.VPCNotFound && ok {
```
ncdc: Can you move it to check `ok` first?

pkg/cloud/aws/services/wait/wait.go
```diff
@@ -19,7 +19,8 @@ package wait
 import (
 	"time"
 
-	aMW "k8s.io/apimachinery/pkg/util/wait"
+	"github.com/pkg/errors"
+	apiMachineryWait "k8s.io/apimachinery/pkg/util/wait"
```
ncdc: Import aliases should be all lowercase. Alternatively, we don't need an alias here. A reference to `wait.Backoff` below means the apimachinery package, not the current package.

If we want to keep an alias, I usually do something like `kwait` for kubernetes wait.

pkg/cloud/aws/services/ec2/gateways.go
```diff
@@ -121,12 +127,37 @@ func (s *Service) createInternetGateway() (*ec2.InternetGateway, error) {
...
+		EC2Client:   s.scope.EC2,
+		BuildParams: tagParams,
+	}
+	tag := func() (bool, error) {
+		if err := tags.Apply(tagApply); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), tag, awserrors.InternetGatewayNotFound); err != nil {
```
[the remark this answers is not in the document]
detiber: Yeah, while strictly not required, since the tags would still be applied after the create call finishes, but this brings it in line with the other "services".

pkg/cloud/aws/services/ec2/gateways.go
detiber: Setting this here makes sure that we are updating the tags after calling ApplyTags, and the later call to EnsureTags is a noop instead of trying to reapply the tags.

pkg/cloud/aws/services/ec2/natgateways.go
detiber: For each of the errors, I tested against the AWS API to see what errors where returned, since the docs are spotty.

pkg/cloud/aws/services/ec2/gateways.go
ncdc: Could you please add a comment to that effect?

pkg/cloud/aws/services/ec2/natgateways.go
ncdc: Painful 😦

pkg/cloud/aws/services/ec2/vpc.go
detiber: - The calling function (reconcile) will treat it as an error and return an error to the Actuator ignoring the returned vpc information
- Since the returned vpc is not copied into the scope, the returned vpc data is basically dropped (as it is today).

This is essentially a noop from today just replacing the two independent returns (nil, error) and (vpc, nil) with a single return. If it would make it clearer that there are no side effects from this (which there shouldn't be if we are sticking to Go convention to ignore the return value if there is an error), then I'd be happy to revert back to two distinct returns.

ncdc: I think it's safer to have the two distinct returns

pkg/cloud/aws/services/ec2/vpc.go
```diff
@@ -140,18 +152,29 @@ func (s *Service) createVPC() (*v1alpha1.VPCSpec, error) {
...
 	}
-
-	if err := tags.Apply(tagApply); err != nil {
+	tag := func() (bool, error) {
+		if err := tags.Apply(tagApply); err != nil {
+			return false, err
+		}
+		return true, nil
+	}
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), tag, awserrors.VPCNotFound); err != nil {
```
detiber: @ncdc thoughts on the readability of this if I inlined the `func()` instead of storing it in `tag` above? Same with the `tagApply` and `tagParams`.

ncdc: I've been holding off on suggesting inlining the funcs - I assumed you liked it better that way. If you want to do it, no objections here.

detiber: I was sticking with the previous convention used with `WaitForWithRetryable`, I much prefer the inlined approach :joy:

pkg/cloud/aws/services/elb/loadbalancer.go
```diff
@@ -259,27 +267,22 @@ func (s *Service) deleteClassicELBAndWait(name string) error {
 		LoadBalancerNames: aws.StringSlice([]string{name}),
 	}
 
-	checkForELBDeletion := func() (done bool, err error) {
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (done bool, err error) {
 		out, err := s.scope.ELB.DescribeLoadBalancers(input)
+		if err != nil {
+			if code, ok := awserrors.Code(err); ok && code == "LoadBalancerNotFound" {
```
ncdc: Do you want to use a constant for this code?

ncdc:
1 minor question.
/lgtm

Feel free to remove the hold whenever. Do you want another reviewer before this merges?

pkg/cloud/aws/services/elb/loadbalancer.go
detiber: Indeed, I had overlooked updating it. Fixed now.

sethp-nr:
I'm sorry I dropped this one this week. FWIW I'm quite happy with how it worked out (big +1 on the discussion re: `ManagedNetworkSpec` etc.), and it'd definitely solve the issue we ran into in #890 (among others). Thank you for taking it on @detiber and thanks to @ncdc for the review!

PR #923, opened 2019-07-23
Update CAPI dependency, remove clusterctl and make examples top level

vincepri:

This PR updates to the latest master version of Cluster API and removes the custom `clusterctl` binary. It also refactors the repository so the generated examples directory is top level.

Fixes #

**Release note**:

```release-note

```


Makefile
```diff
@@ -168,7 +168,7 @@ manager: ## Build manager binary.
 
 .PHONY: clusterctl
 clusterctl: ## Build clusterctl binary.
```
ncdc: Can we remove this target entirely?

ncdc: And everywhere we invoke `bin/clusterctl`, change it to `clusterctl` (must be on `$PATH`; require users to install it from CAPI/brew/etc)?

vincepri: Once we have a release and means to install clusterctl/any other tool, that'd be fine. I would like to keep this target for now while we develop locally

docs/roleassumption.md
```diff
@@ -1,21 +1,21 @@
 # Creating clusters using cross account role assumption using KIAM
```
detiber: Should we add a link at the top of this file to the same doc in the `release-0.3` branch?

vincepri: That wouldn't hurt, are we serving docs from release-0.3 branch in this repo?

detiber: We don't publish the docs anywhere other than from the git repo itself. Until we do, we should likely ensure that we add links to the `release-0.3` doc to avoid user confusion when trying to consume releases.

vincepri: How would the link look like? Is that preferred to just revert the files to what they were before?

detiber: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/release-0.3/docs/roleassumption.md for this specific doc

detiber: I'd like to avoid deferring docs changes, otherwise we risk missing more things than if we update in sync with the changes.

vincepri: Added, ptal :)

ncdc:
LGTM. @detiber?

PR #927, opened 2019-07-23
[0.3] Backport Go modules (#888)

vincepri:

Fixes #

**Release note**:

```release-note

```


go.vendor
```diff
@@ -0,0 +1 @@
+sigs.k8s.io/cluster-api
```
ncdc: What is this for?

vincepri: https://github.com/kubernetes-sigs/cluster-api/pull/1134

ncdc:
/lgtm
/hold

@detiber are you interested in reviewing?

go.mod
```diff
@@ -0,0 +1,40 @@
...
+	k8s.io/klog v0.3.2
+	k8s.io/kubernetes v1.13.3
+	sigs.k8s.io/cluster-api v0.1.8
+	sigs.k8s.io/controller-runtime v0.1.12
+	sigs.k8s.io/controller-tools v0.1.11
+	sigs.k8s.io/testing_frameworks v0.1.1
+	sigs.k8s.io/yaml v1.1.0
+)
+
+replace k8s.io/apimachinery => k8s.io/apimachinery v0.0.0-20190221213512-86fb29eff628
```
detiber: Do we need to list `cluster-api` as a replace to avoid issues after we start cutting v1alpha2-related releases?

vincepri: I'd think that given the incompatibilities it'll not default to v2. We can add a replace at some point if needed though

detiber: I'm not sure how much compatibility testing is done with gomod, especially prior to major versions being set. It definitely will not attempt to compile against versions like `dep` does.

vincepri: We can follow-up in another commit if needed, I opened this mostly to backport the modules support before the release is cut.

PR #934, opened 2019-07-26
Clean bazel mocks before generating them

vincepri:

This PR fixes an issue where a stale Bazel cache could cause the `hack/copy-mocks.sh` file to error out.

Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -52,7 +52,6 @@ all: verify-install test binaries
 help:  ## Display this help
 	@awk 'BEGIN {FS = ":.*##"; printf "\nUsage:\n  make \033[36m<target>\033[0m\n"} /^[a-zA-Z_-]+:.*?##/ { printf "  \033[36m%-15s\033[0m %s\n", $$1, $$2 } /^##@/ { printf "\n\033[1m%s\033[0m\n", substr($$0, 5) } ' $(MAKEFILE_LIST)
 
-
 ## --------------------------------------
 ## Testing
 ## --------------------------------------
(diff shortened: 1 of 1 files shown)


ashish-amarnath:
Do you mean `hack/copy-bazel-mocks.sh` will fail w/o this change?

Makefile
```diff
@@ -266,6 +265,10 @@ clean: ## Remove all generated files
 clean-bazel: ## Remove all generated bazel symlinks
 	bazel clean
 
+.PHONY: clean-bazel-mocks
+clean-bazel-mocks: ## Remove all generated bazel mocks files
+	find bazel-bin/pkg -name '*_mock.go' -type f -delete
```
ashish-amarnath: Can we use `bazel clean` here instead of `find -delete`?

vincepri: The `make clean-bazel` target is too aggressive for this particular issue, this removes only the mocks in cache.

vincepri:
@ashish-amarnath The target fails if your cache is stale, or even if you changed a folder name for example

Makefile
ashish-amarnath: Sorry for not being clear. I mean can we use `bazel clean <some-target>`? not `bazel clean` everything.

vincepri: I don't think clean accepts any arguments from what I can see

ashish-amarnath: OK. I wasn't able to find examples that we can use here.
I am 👍 w/ this change.

PR #948, opened 2019-08-01
[0.3] Update clusterawsadm bootstrap policy

detiber:

Fixes a missing IAM policy action for the CAPA controllers.

**Release note**:
```release-note
- clusterawsadm now deploys the proper policy with permission for describing vpc attributes
```


detiber:
Integration test error is unrelated, fix for it is here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/949

dlipovetsky:
Thanks for fixing this @detiber! What's the right way to update the CloudFormation stack?

/lgtm

detiber:
I followed up in slack, but wanted to follow up here in case anyone runs across it in search results.

The `clusterawsadm alpha bootstrap create-stack` command will update the CloudFormation Stack if it already exists.

PR #953, opened 2019-08-02
Update e2e tests

detiber:

- Updates the e2e to:
  - create the kind cluster using a randomized name
  - create the Namespace using a randomized name
  - create the Cluster using a randomized name
  - include deletion of Cluster
- Update kind to v0.4.0 and use script to donwload rather than Bazel
- Update kubectl to v1.15.0 and use script to download rather than Bazel

**Release note**:
```release-note
NONE
```


hack/e2e.sh
```diff
@@ -29,6 +29,21 @@ set -o pipefail
...
+  GO111MODULE="on" go get sigs.k8s.io/kind@v0.4.0
+fi
+
+# If kubectl is not available on the path, get it
+# TODO: adapt for other OS and ARCH
+if ! [ -x "$(command -v kubectl)" ]; then
+  echo 'kubectl not found, installing'
+  curl -LO https://storage.googleapis.com/kubernetes-release/release/v1.15.0/bin/linux/amd64/kubectl
+  chmod +x kubectl
+  mv kubectl ~/bin
```
detiber: Need to have a better location for downloading to, since this will likely not work in prow

detiber: @chuckha @ncdc what are your thoughts on making some of the testing prerequisites like `kubectl`, `kustomize`, and `kind` development prerequisites and us building a custom image for use in prow with them already baked in? Then we could have a script that just ensures they are >= a particular version and avoid having to download/build within bazel or do this type of downloading shenanigans in the test wrappers.

detiber: We could even potentially eliminate the `kustomize` requirement all together with a new enough version of `kubectl`

ncdc: Seems reasonable to me. And it would ensure a consistent experience (locked versions of the tools) across environments.

hack/e2e.sh
```diff
@@ -29,6 +29,23 @@ set -o pipefail
 REPO_ROOT=$(dirname "${BASH_SOURCE[0]}")/..
 cd "${REPO_ROOT}" || exit 1
 
+# If kind is not available on the path, get it
+if ! [ -x "$(command -v kind)" ]; then
+  echo 'kind not found, installing'
+  pushd ~
+  GOFLAGS="" GO111MODULE="on" go get sigs.k8s.io/kind@v0.4.0GOFLAGS="" GO111MODULE="on" go get sigs.k8s.io/kind@v0.4.0
```
ncdc: This line looks like a copy/paste error (2x)?

detiber: Indeed

detiber:
@ncdc If the current changes look good, I'd like to forward port the compatible changes and then defer additional functionality for followup.

test/e2e/aws_test.go
```diff
@@ -81,53 +79,97 @@ func initRegion() error {
...
+
+			// TODO: Retrieve Cluster kubeconfig
+			// TODO: Deploy Addons
+			// TODO: Validate Node Ready
+			// TODO: Deploy additional Control Plane Nodes
+			// TODO: Deploy a MachineDeployment
+			// TODO: Scale MachineDeployment up
+			// TODO: Scale MachineDeployment down
+
+			// Nasty little hack to ensure that the Control Plane instance has come up
```
ncdc: Do you want to add an Eventually that checks for the control plane ready annotation, rather than hard coding a 5m sleep?

ncdc:
LGTM pending 1 question

test/e2e/aws_test.go
detiber: Done, still testing the latest changes, though.

detiber:
ok, local testing for this looks good. @ncdc this should be good to go now.

PR #957, opened 2019-08-06
Update E2E

detiber:

Updates the e2e and integration test framework with the latest changes from the release-0.3 branch, and removes v1alpha1 oriented testing.

**Release note**:
```release-note
NONE
```


detiber:
@ncdc this should be good to go now. It effectively removes all e2e and integration testing that was done (they were effectively testing v1alpha1-related items anyway :joy:) but brings over the framework and kind changes.

ncdc:
I guess the plan is to copy/paste back some of the code that was deleted if it makes sense in the future, as we get closer to actual testing?

detiber:
> I guess the plan is to copy/paste back some of the code that was deleted if it makes sense in the future, as we get closer to actual testing?

Potentially, but the amount that will need to be replaced is pretty trivial.

PR #963, opened 2019-08-06
packer: Add image permission and encryption options

### LINKED ISSUE #960: Instruct Packer to make EBS snapshots public
/kind feature
/assign
/lifecycle active

If the underlying EBS snapshot of an AMI is public, then a user can copy the AMI to any region or account they like, and also encrypt it upon copying, which also then theoretically allows the use of encrypted root volumes for the cluster instances with copies of the public AMIs. Organizations may want to move the Packer builds in house to achieve the same thing, but this opens up an additional use case.

**Environment:**

- Cluster-api-provider-aws version: 0.3.6

randomvariable:
Adds Packer variables to adjust the visibility of AMIs and EBS
snapshots.

By default, the underlying EBS snapshot of an AMI will also be public,
which allows external users to copy the AMI into their own accounts for
further modification or encryption.

An additional variable `encrypted` allows a user to build an encrypted
AMI directly into their account. Instructions are updated as such.

Fixes #960 

**Release note**:

```release-note
packer: Options added for AMI visibility between accounts and AMI encryption.
```


chuckha:
Is this going to make its way into https://github.com/kubernetes-sigs/image-builder ?

I remember us justifying the image builder in here (/being hosted on VMware/heptio $) as a "sample to get people started" and we expect folks to build and maintain their own images. 

Are we seeing people take these images and use them as a base for their own images?

randomvariable:
I was mainly trying to enable @aaroniscode so he didn't had to rebuild the AMIs for the project he was on, thought too late for that. I think we should merge this if we intend to build another set of AMIs from here, otherwise only in the other project. And additionally, this should be forward ported to https://github.com/kubernetes-sigs/image-builder.

chuckha:
/approve

thanks @randomvariable

Issue #968, opened 2019-08-07
Update config/


chuckha:
/kind feature

**Describe the solution you'd like**
I'd like the `config/` directory to be in line with the other providers for v1alpha2. This will allow us to have a consistent developer experience across v1alpha2 providers.

kubebuilder v0.2 offers a single place to kustomize the entire `config/` directory, namely `config/default/kustomize.yaml`. 

I think we could run `controller-gen` and then fix the build and mostly be ok.

Current version of controller-gen is v0.2.0-beta.5 (at the time this issue was created)

chuckha:
/close

Fixed by #989

chuckha:
oops, I meant #994

PR #971, opened 2019-08-07
Remove generation from v1alpha1 types

chuckha:

This PR removes CRD generation from v1alpha1 types. We no longer need to generate CRDs from the v1alpha1 types but we will need to generate deep copy and other code. We keep the types around for the upgrade tool that will be built here.

This unblocks us from keeping this repo up-to-date with a controller-gen style config directory and a single `kustomize.yaml` file to generate all necessary CRDs.

Related to #968 

The deleted code in `clusterawsadm` is unused.

**Release note**:
```release-note
NONE
```

/assign @ncdc @detiber

### DIFF
--- a/cmd/clusterawsadm/client/BUILD.bazel
+++ b/cmd/clusterawsadm/client/BUILD.bazel
@@ -6,8 +6,6 @@ go_library(
     importpath = "sigs.k8s.io/cluster-api-provider-aws/cmd/clusterawsadm/client",
     visibility = ["//visibility:public"],
     deps = [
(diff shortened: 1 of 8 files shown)


detiber:
@chuckha looks like there are at least some left over bazel references

ncdc:
I was assuming we would be an alpha1 to alpha2 conversion tool in this repo. But maybe we just keep the structs around for that, without the generated stuff?

chuckha:
@ncdc that's totally reasonable to me.

Issue #913, opened 2019-07-22
[v1alpha2] Forward ports needed


detiber:
Some fixes/updates were added to release-0.3 that will need to be forward ported to master when it is more stable:

- [x] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/910 - cluster events (PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/961)
- [x] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/908 - LB fix (@chuckha)
- [x] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/839 - machine events (PR: #974) (@chuckha)
- [x] #916 - backoff & retry (PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/975) (@chuckha)
- [x] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/948 - clusterawsadm policy update (PR: #951)
- [x] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/955 - Query VPC prior to deletion (PR: #973)

chuckha:
Working on the first item, #910

ncdc:
I've got 910 covered

chuckha:
jk doing #908

chuckha:
#908 needs no action.

It makes two changes:

1. It will update the load balancer on Update instead of Exists. The v1alpha2 code already exhibits this behavior: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/controller/awsmachine/awsmachine_controller.go#L268 This happens just after setting the provider id. This is the normal path of reconcile and there is no early-exit condition.

2. Always set the provider id. This is now always happening. in the `reconcileNormal` method: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/controller/awsmachine/awsmachine_controller.go#L250. This happens on every reconcile.

chuckha:
@detiber Can you confirm my logic for #908 above?

If that's sound then this issue can be closed

detiber:
@chuckha I agree it appears to be a noop given the new reconciliation behavior.
/close

PR #974, opened 2019-08-07
Port #839 forward

chuckha:

This PR moves the logic from #839 into v1alpha2

Related to #913 

/assign @detiber @ncdc 

**Release note**:
```release-note
Bring additional events from v1alpha1 into v1alpha2
```

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -181,10 +181,15 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*v1alpha2.Instance,
 	s.scope.V(2).Info("Running instance", "machine-role", scope.Role())
 	out, err := s.runInstance(scope.Role(), input)
 	if err != nil {
+		// Only record the failure event if the error is not related to failed dependencies.
+		// This is to avoid spamming failure events since the machine will be requeued by the actuator.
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/ec2/instances.go
```diff
@@ -181,10 +181,13 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*v1alpha2.Instance,
 	s.scope.V(2).Info("Running instance", "machine-role", scope.Role())
 	out, err := s.runInstance(scope.Role(), input)
 	if err != nil {
+		if !awserrors.IsFailedDependency(errors.Cause(err)) {
```
detiber: Do we want to also bring the comment over for why we are only conditionally emitting the event.

chuckha: yes we do. This bit was slightly tricky due to the v1a2 differences and I must have missed it, good catch

ncdc:
I think we also need to remove instances.go:233:
```go
record.Eventf(s.scope.Cluster, "DeletedInstance", "Terminated instance %q", instanceID)
```

chuckha:
oh yes, let me go through the deletes again to make sure i don't miss any other deletes

chuckha:
updated.

PR #976, opened 2019-08-07
increase size of bazel e2e test

detiber:

Increases the "size" of the e2e test in bazel, to avoid timeouts that we've been hitting in the periodics.

**Release note**:
```release-note
NONE
```


ncdc:
😄 

what does size refer to in go_test bazel-land?

detiber:
@ncdc size in bazel land refers to both memory constraints and defaults the timeout. Timeout could be set independently, but I didn't necessarily see a reason to here.

PR #982, opened 2019-08-08
Adds controller-gen as a pinned tool

chuckha:

This PR starts to make CAPA look more like the other providers by pinning controller-gen in the tools repository.

We still use the vendor version but the overall goal is to replace that with the recommended go path for pinning dev/test tooling. 

We are not using bazel to run controller-gen so this is not a major departure from existing tooling.

Related to #968 

This code is not yet being used but will be used in existing makefile targets.

**Release note**:
```release-note
NONE
```
/assign @detiber @ncdc


detiber:
@chuckha should there also be related Makefile changes?

chuckha:
@detiber not yet, that's coming. This is just adding the tooling. (also update for fixed boilerplate)

vincepri:
/hold

How is this different from https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/hack/tools.go? Is there any benefit from one vs the other?

ncdc:
@vincepri I chatted upstream and people suggested using a separate go "sub"module for tools. We should convert CAPA to this approach and migrate what's in hack/tools.go here.

chuckha:
I can move over all the tools.go imports into this submodule

vincepri:
I'm fine either way, I wonder if any other project is following the sub-module approach. AFAIK controller-runtime and controller-tools only have `hack/tools.go`. When I was working on CAPI a while back I tried both approaches, in the end I preferred to keep everything in one `go.mod` to easily track dependencies and their replacements (if necessary). 

If we use the sub-module approach, is there going to be a case where we might have to deal with  version compatibility issues?

ncdc:
Maybe? Let's cross that bridge if/when we need to?

vincepri:
Sounds good, @chuckha can you also do the same for the other imports? We should do the same for CAPI as well to standardize across all repos.

Feel free to unhold.

chuckha:
I would strongly recommend going through this commit by commit if you're willing to review.

Contentious points:

specifying clusterctl//capi dependency version in two places.

config/crd/bases/_.yaml
```diff
@@ -0,0 +1,16 @@
...
+  names:
+    kind: ""
+    plural: ""
+  scope: ""
+status:
+  acceptedNames:
+    kind: ""
+    plural: ""
+  conditions: null
+  storedVersions: null
```
detiber: Is this file correct?

WORKSPACE
```diff
@@ -109,12 +109,6 @@ go_repository(
     tag = "v1.3.1",
 )
 
-go_repository(
-    name = "io_k8s_sigs_kustomize",
-    importpath = "sigs.k8s.io/kustomize",
-    tag = "v1.0.11",
-)
-
```
detiber: Are all the bazel targets that were using this removed?

config/crd/bases/_.yaml
chuckha: how did that get in there. No, this file should not be in here.

WORKSPACE
chuckha: No, I need to fix that up but I've run out of time for the day

chuckha:
/close

I'm going for a different approach, this was too much at once

PR #984, opened 2019-08-08
Guard against missing instanceID on delete

detiber:
Guard against missing instanceID on delete

**Release note**:
```release-note
Guard against missing instanceID when deleting Machines
```


vincepri:
The change looks good to me, can we add a related unit test? I wonder though if we should fix the called function instead, given that we should return an error if the instance doesn't have an ID.

PR #988, opened 2019-08-09
[wip] Remove custom build of cloud-init

### LINKED ISSUE #983: cloud-init fails on Amazon Linux and CentOS AMI's, "Error persisting instance-data.json"
/kind bug

Create a CAPA cluster and choose an Amazon Linux AMI. I tried `ami-092d9d6a36468257d` which is for k8s v1.14.4.

I also tried with a custom built AMI for k8s 1.15.2 and both fail with the same error.

cloud-init fails very early in the process in `sources/__init__.py`.
```2019-08-08 12:31:38,951 - __init__.py[WARNING]: Error persisting instance-data.json: 'utf8' codec can't decode byte 0x8b in position 1: invalid start byte```

Because `instance-data-json` is not written, cloud-init has no user-data and the PKI assets aren't written and `kubeadm` isn't run.

cloud-init to write PKI assets and run `kubeadm init`

Spent a good deal of time troubleshooting and trying to determine root cause. Here is what I learned.
The contents of the user-data are fine. I logged into the instance, downloaded the gzip user-data and extracted it. Stopped the ec2 instance, updated the user-data by pasting the user-data as text into the EC2 console. Restarted the node and forced cloud-init to run again. PKI assets were created and kubeadm was run.

I compared the non-working Amazon Linux node to a working Ubuntu node. Ubuntu's cloud-init is based on Python3. Amazon Linux's cloud-init is based on Python 2. cloud-init versions are very close. I don't remember the exact versions.

Similar issue on Azure with cloud-init failing to persist instance-data.json: https://bugzilla.redhat.com/show_bug.cgi?id=1631167

Guesses on root cause (as far as I got at this point)
* Maybe the user-data has a non-ascii character in the dictionary
* Perhaps Python3 handles this situation better than Python2

**Environment:**

- Cluster-api-provider-aws version: v0.3.7
- Kubernetes version: (use `kubectl version`): N/A, doesn't seem dependent on k8s version
- OS (e.g. from `/etc/os-release`): Amazon Linux 2

### PR DESCRIPTION
**Pull request by randomvariable**
Removes the custom build of cloud-init from AMIs in favour of patching the distribution provided version directly.

This fixes a broad range of compatibility issues with Amazon Linux 2, and additionally fixes a gzip decoding bug on CentOS and Amazon Linux 2 that has since been fixed upstream.

Additionally, AMI lookup has been added to Packer which will automatically select the latest AMI for each distro by default.

Fixes #983 

Currently testing AMIs

**Release note**:

```release-note
* fixed gzip decoding error in cloud-init on CentOS 7 and Amazon Linux 2
```


build/amis/ansible/roles/common/defaults/main.yml
```diff
@@ -35,9 +34,9 @@ common_debs:
...
 
-common_pinned_debs:
-- "cloud-init={{ cloud_init_version }}"
 common_extra_debs: []
 common_redhat_epel_rpm: "https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm"
-common_redhat_cloud_init_copr_rpm: "https://copr.fedorainfracloud.org/coprs/jdetiber/cloud-init/repo/epel-7/jdetiber-cloud-init-epel-7.repo"
-common_deb_cloud_init_ppr_repo: "ppa:detiber/cloud-init"
\ No newline at end of file
+common_redhat_epel_rpm_checksum: "sha256:33b8b8250960874014d7c66d04ef12df3a71277b9eb6f2146bba7553daeaf910"
+common_cloud_init_kubeadm_module: "https://raw.githubusercontent.com/detiber/cloud-init/44f9570f3005ace405c4899aded97d51873c7664/cloudinit/config/cc_kubeadm.py"
```
detiber: Why not just add the file to `build/roles/common/files/`? That would remove the need for the remote call.

randomvariable: That's what I did initially, but it's also potentially GPL licensed code, so wasn't sure how to put it in the repo.

detiber: I'm good with re-licensing to Apache, would just need approval from Craig.

randomvariable: oh, didn't read fully. It's dual licensed. I'll move it over.

build/amis/ansible/roles/common/defaults/main.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 The Kubernetes Authors.
+# Copyright 2019 The Kubernetes Authors.
```
detiber: This should remain 2018

build/amis/ansible/roles/common/tasks/debian.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 The Kubernetes Authors.
+# Copyright 2019 The Kubernetes Authors.
```
detiber: Should remain 2018

build/amis/ansible/roles/common/tasks/main.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 The Kubernetes Authors.
+# Copyright 2019 The Kubernetes Authors.
```
detiber: should remain 2018

build/amis/ansible/roles/common/tasks/redhat.yml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2018 The Kubernetes Authors.
+# Copyright 2019 The Kubernetes Authors.
```
detiber: 2018

randomvariable:
Ran into an issue with AML2 rejecting CAPA written user-data with `write-files:` present, so taking a look before taking the WIP off.

randomvariable:
Stuck on this now. CentOS 7 and Amazon Linux 2 have older versions of cloud-init that do not support the Jinja templating used to pass in the local ipv4 address into the CertSANs array in the kubeadm config.

Even if this merges, AL2 and CentOS 7 would still be badly broken.

There are patches that can be done to the custom cloud-init build to fix the gzip decoding error on Python 2.7.

Not sure how to proceed.

randomvariable:
I am thinking of removing the Jinja templating in the outputted user-data for v1alpha1 and moving the logic inside of the kubeadm module itself. Does that seem right?

detiber:
> I am thinking of removing the Jinja templating in the outputted user-data for v1alpha1 and moving the logic inside of the kubeadm module itself. Does that seem right @detiber ?

I'm not sure that is the direction that folks want to go, since it seemed like the desire was to remove the need for a custom build of cloud-init. That said, I do think the better approach will likely be to bake in some local script (via cloud-init module or otherwise) that would do the necessary variable interpolation of the kubeadm config prior to invoking kubeadm. This could be as simple as a script at a pre-determined location (that we bake in via the image builder) and run as command in the cloud-config prior to running kubeadm.

randomvariable:
Me and @detiber had an offline chat. The consensus is to work towards getting AL2 fixed for v1alpha2 in github.com/kubernetes-sigs/image-builder

PR #992, opened 2019-08-09
userdata: Fix cloud-init header

randomvariable:

Amazon Linux 2 has strict validation of the cloud-init
header and rejects `##template`.

**Release note**:

```release-note
* Set correct header for cloud-init userdata
```


randomvariable:
see we're actually using the jinja features.

PR #994, opened 2019-08-09
Convert project to kubebuilder version 2

### LINKED ISSUE #989: Convert project to kubebuilder v2
/kind feature

A tracking issue for all the tasks needed to convert this to a kubebuilder v2 project to be in line with the other providers.

- [ ] #970 Convert makefile to be more kubebuilder-like
- [ ] #968 Convert conifg/ to be more kubebuilder-like
- [ ] #1004 Move cmd/manager/main.go to the top level
- [ ] #1005 Move api to expected place for kubebuilder generation to work seamlessly

chuckha:

This PR converts capa to a kubebuilder v2 project.

Fixes #989 

**Release note**:
```release-note
NONE
```


cmd/manager/main.go
```diff
@@ -19,78 +19,68 @@ package main
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/apis/infrastructure/v1alpha2"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/log/zap"
+	// +kubebuilder:scaffold:imports
 )
 
+var (
+	myscheme = runtime.NewScheme()
+	setupLog = ctrl.Log.WithName("setup")
```
vincepri: Are these coming from kubebuilder?

cmd/manager/main.go
```diff
@@ -19,78 +19,68 @@ package main
...
+			setupLog.Error(http.ListenAndServe(*profilerAddress, nil), "listen and serve error")
 		}()
 	}
 
-	// Setup a Manager
-	syncPeriod := 10 * time.Minute
 
-	// Setup controller-runtime logger.
-	log.SetLogger(klogr.New())
+	ctrl.SetLogger(zap.Logger(true))
```
vincepri: I thought we were sticking to klogr for the time being, has that changed?

hack/tools/tools.go
```diff
@@ -0,0 +1,35 @@
...
+
+import (
+	_ "k8s.io/code-generator/cmd/client-gen"
+	_ "k8s.io/code-generator/cmd/conversion-gen"
+	_ "k8s.io/code-generator/cmd/deepcopy-gen"
+	_ "k8s.io/code-generator/cmd/defaulter-gen"
+	_ "k8s.io/code-generator/cmd/informer-gen"
+	_ "k8s.io/code-generator/cmd/lister-gen"
+	_ "k8s.io/code-generator/cmd/register-gen"
+	_ "k8s.io/code-generator/cmd/set-gen"
```
vincepri: We can probably remove code-generator from here and use `controller-gen` to generate the deepcopy files

pkg/apis/infrastructure/v1alpha2/awscluster_types.go
```diff
@@ -42,77 +38,6 @@ type AWSClusterSpec struct {
 	SSHKeyName string `json:"sshKeyName,omitempty"`
 }
 
-// NetworkSpec encapsulates all things related to AWS network.
```
vincepri: Why move these? Seems they are only used within this file

pkg/apis/infrastructure/v1alpha2/zz_generated.deepcopy.go
```diff
@@ -16,20 +16,19 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-// Code generated by main. DO NOT EDIT.
+// autogenerated by controller-gen object, do not modify manually; DO NOT EDIT
```
vincepri: How were these generated with controller-gen? I don't see any changes in the Makefile / bazel

cmd/manager/main.go
```diff
@@ -19,78 +19,68 @@ package main
 import (
```
vincepri: Should this file be moved in the root directory?

cmd/manager/main.go
chuckha: yeah, this one snuck through, reverting

cmd/manager/main.go
chuckha: it's `scheme` by default but that shadows the import so I changed it to `myscheme` instead.

The setupLog is there by default in v2 main.go though.

cmd/manager/main.go
chuckha: no, I don't think so. Since we are shipping multiple binaries with this project I think it makes more sense organizationally to keep them all under `cmd/<binary name>/`

hack/tools/tools.go
chuckha: This shouldn't actually be here, I'm not making this change as it's too big. I'll remove this entirely

pkg/apis/infrastructure/v1alpha2/zz_generated.deepcopy.go
chuckha: controller-gen uses the PROJECT file to detect which version of kubebuilder to use. I updated that to version 2 so no changes to the build need to happen.

pkg/apis/infrastructure/v1alpha2/awscluster_types.go
chuckha: these got moved to keep the awscluster_types.go and awsmachine_types.go files to contain only the kubebuilder generated types. The supporting types live in types.go

pkg/apis/infrastructure/v1alpha2/zz_generated.deepcopy.go
vincepri: That makes sense, but how was this file generated? Using which command?

chuckha: oh, I see what you're saying, yes I'll update the makefile, it's using the normal kubebuilder v2 controller-gen command with a modified path

cmd/manager/main.go
```diff
@@ -19,78 +19,68 @@ package main
...
-		klog.Infof("Profiler listening for requests at %s", *profilerAddress)
+		setupLog.Info("Profiler listening for requests", "profiler-address", *profilerAddress)
 		go func() {
-			klog.Info(http.ListenAndServe(*profilerAddress, nil))
+			setupLog.Error(http.ListenAndServe(*profilerAddress, nil), "listen and serve error")
 		}()
 	}
 
-	// Setup a Manager
-	syncPeriod := 10 * time.Minute
```
detiber: This likely needs to be preserved to avoid the default going back to 10 hours.

chuckha: ok yeah that's fair

chuckha:
@vincepri @detiber this is ready for another pass

cmd/manager/main.go
```diff
@@ -19,78 +19,71 @@ package main
 import (
 	"flag"
 	"net/http"
-	_ "net/http/pprof"
```
chuckha: this should not go away

cmd/manager/main.go
```diff
@@ -19,78 +19,71 @@ package main
...
 
-	// Get a config to talk to the api-server.
-	cfg := config.GetConfigOrDie()
-	mgr, err := manager.New(cfg, manager.Options{
+	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+		Scheme:             myscheme,
+		MetricsBindAddress: metricsAddr,
+		LeaderElection:     enableLeaderElection,
 		SyncPeriod:         &syncPeriod,
-		Namespace:          *watchNamespace,
```
chuckha: should not go away

cmd/manager/main.go
```diff
@@ -19,78 +19,71 @@ package main
...
-		MetricsBindAddress: *metricsAddress,
 	})
 	if err != nil {
-		klog.Fatalf("Failed to set up overall controller manager: %v", err)
-	}
-
-	// Initialize event recorder.
-	record.InitFromRecorder(mgr.GetEventRecorderFor("aws-controller"))
```
chuckha: also still needed

detiber: Should it, or is it something that we should set in the individual controllers instead?

pkg/apis/infrastructure/v1alpha2/awsmachinetemplate_types.go
```diff
@@ -1,63 +0,0 @@
-/*
```
detiber: Doh, I think the AWSMachineTemplate type should probably have been preserved.

pkg/cloud/scope/cluster.go
```diff
@@ -127,13 +127,6 @@ func (s *ClusterScope) ControlPlaneConfigMapName() string {
 	return fmt.Sprintf("%s-controlplane", s.Cluster.UID)
 }
 
-// ListOptionsLabelSelector returns a ListOptions with a label selector for clusterName.
-func (s *ClusterScope) ListOptionsLabelSelector() client.ListOptionFunc {
-	return client.MatchingLabels(map[string]string{
-		clusterv1alpha2.MachineClusterLabelName: s.Cluster.Name,
-	})
-}
-
```
detiber: I'm guessing this wasn't used anywhere?

cmd/manager/main.go
chuckha: oh i'm not sure. will it still work initialized in main? I'd like to keep it there and then move it later

pkg/cloud/scope/cluster.go
chuckha: That's correct, wasn't used

pkg/apis/infrastructure/v1alpha2/awsmachinetemplate_types.go
chuckha: Oh yeah. Totally.

cmd/manager/main.go
vincepri: We should probably open an issue in kubebuilder

cmd/manager/main.go
vincepri: Wouldn't kubebuilder complain when generating new controllers or types if it can't find a main.go in the root?

chuckha: Yep it does, but I'm not terribly concerned by it. Generating new apis put them in `api/` instead of `/pkg/apis/`.

Moving main.go is going to require changing paths and some Bazel. I can see this task falling into the scope of this PR. I'd prefer doing it after due to the build changes, but if you'd like it done in this PR that's fine too.

vincepri: That's fine, I was mostly wondering if it was on your radar, rather than pushing it for doing it in the same PR

cmd/manager/main.go
```diff
@@ -19,78 +19,77 @@ package main
...
 )
 
+func init() {
+	_ = scheme.AddToScheme(myscheme)
+	_ = v1alpha2.AddToScheme(myscheme)
+	// +kubebuilder:scaffold:scheme
+}
+
 func main() {
-	klog.InitFlags(nil)
```
vincepri: Don't we need this?

chuckha: yes. definitely.

config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
```diff
@@ -11,8 +11,6 @@ spec:
     kind: AWSCluster
     plural: awsclusters
   scope: ""
-  subresources:
-    status: {}
```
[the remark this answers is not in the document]
chuckha: sure is, let me regen. I probably missed some of those kubebuilder markers i didn't know needed to exist. Let me add those.

chuckha:
@detiber @vincepri 

This gets us to the point of the un-implemented controller.

Take a look at config/ as well. I added some stuff to make the credentials secret (also look at the Makefile).

Otherwise the biggest changes: i deleted the examples directory in its entirety for v1alpha2.

config/manager/manager.yaml
```diff
@@ -2,85 +2,30 @@ apiVersion: v1
...
+  name: system
 ---
 apiVersion: apps/v1
-kind: StatefulSet
+kind: Deployment
 metadata:
   name: controller-manager
   namespace: system
   labels:
     control-plane: controller-manager
```
detiber: This should likely use a different value, otherwise it will conflict with other kubebuilder projects using the defaults in the same namespace.

Makefile
```diff
@@ -87,24 +96,14 @@ e2e-janitor:
 .PHONY: docker-build
 docker-build: generate lint-full ## Build the docker image for controller-manager
 	docker build --pull . -t $(MANAGER_IMAGE)
+	# TODO: sed probably needs to be gnu sed not bsd sed
+	sed -i '' -e 's@image: .*@image: '"${MANAGER_IMAGE}"'@' ./config/default/manager_image_patch.yaml
+	sed -i '' -e 's@credentials: .*@credentials: '"${B64_CREDENTIALS}"'@' ./config/default/aws_credentials_patch.yaml
```
detiber: Should this be part of the `docker-build` target? I would expect it as part of `generate-manifests` instead.

BUILD.bazel
```diff
@@ -80,3 +81,74 @@ filegroup(
...
+    visibility = ["//visibility:public"],
+)
+
+docker_push(
+    name = "manager-push",
+    bundle = "manager-image",
+    tags = ["manual"],
+)
+
+stateful_set_patch(
```
detiber: This should likely be renamed since we are no longer using a StatefulSet

BUILD.bazel
```diff
@@ -80,3 +81,74 @@ filegroup(
...
+    pure = "on",
+    visibility = ["//visibility:public"],
+)
+
+tags = [
+    "{GIT_VERSION}",
+    "$(MANAGER_IMAGE_TAG)",
+]
+
+images = {
```
detiber: Should we have both a Dockerfile and a Bazel build for images? We should probably consolidate on one or the other.

Makefile
chuckha: This is how kubebuilder v2 makefiles are.

I believe the idea behind this being part of the docker-build step is that you'd build for a release/deployment. At which point the sed is applied to the image patch and you can now deploy your app.

If it were part of the generate-manifest we would only be guessing at the name of the image. So it might look correct but it wouldn't be valid until the docker-build/push step anyway

BUILD.bazel
chuckha: I agree but every time I touch bazel my PR blows up. Do you think we can do this after this PR? I would be more than happy to make an issue for consolidating the build/release practice.

detiber: Sounds good, lets handle this later

Makefile
detiber: can't say I'm a fan, but we can go with it for now and reassess it later.

detiber:
lgtm, but would like @vincepri to take a look since he's doing similar work in cluster-api

Makefile
chuckha: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1014

BUILD.bazel
chuckha: Agreed. I think figuring out what to do with Bazel is next on the list of cleanups here.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1015

BUILD.bazel
chuckha: Added a comment https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1015#issuecomment-521278295

config/manager/manager.yaml
chuckha: Agreed.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1016

vincepri:
Taking a look now

PR #996, opened 2019-08-10
fix: base64 decode bootstrap data

rudoi:
Signed-off-by: Andrew Rudoi <arudoi@newrelic.com>

Decodes bootstrap data. The bootstrap provider sets the bootstrap data as a base64-encoded string, so we must decode it.

Also snuck in a small typo fix. 

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -204,7 +204,7 @@ func (s *Service) GetCoreSecurityGroups(scope *scope.MachineScope) ([]string, er
 	switch scope.Role() {
 	case "node":
 		// Just the common security groups above
-	case "controlplane":
+	case "control-plane":
 		sgRoles = append(sgRoles, v1alpha2.SecurityGroupControlPlane)
 	default:
 		return nil, errors.Errorf("Unknown node role %q", scope.Role())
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/ec2/instances_test.go
```diff
@@ -245,7 +245,8 @@ func TestCreateInstance(t *testing.T) {
 				},
 				Spec: clusterv1.MachineSpec{
 					Bootstrap: clusterv1.Bootstrap{
-						Data: pointer.StringPtr("user-data"),
+						// echo "user-data" | base64
+						Data: pointer.StringPtr("dXNlci1kYXRhCg=="),
```
vincepri: Shouldn't the value in Bootstrap.Data be just a plain string?

pkg/cloud/services/ec2/instances.go
```diff
@@ -204,7 +204,7 @@ func (s *Service) GetCoreSecurityGroups(scope *scope.MachineScope) ([]string, er
 	switch scope.Role() {
 	case "node":
 		// Just the common security groups above
-	case "controlplane":
+	case "control-plane":
```
vincepri: Sidenote: We should probably document these roles in CAPI at some point

pkg/cloud/services/ec2/instances_test.go
rudoi: As in not base64? I dunno if a decision was made at some point to not do that but the bootstrapper's template func encodes it:

https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/blob/c150edc4f00e11ba1cad905f18d4caff1ad5d1ae/cloudinit/utils.go#L31

vincepri:
@rudoi Let's chat in today's meeting about the user-data and base64 encoding. I'm not sure if the kubeadm provider is encoding it, the original boostrap spec didn't account for it.

detiber:
@vincepri I think base64 encoding makes sense, since for the general use case it avoids a lot of problems of embedding a yaml string in a yaml doc.

vincepri:
/approve

Let's merge this in and fix it later if needed

Issue #981, opened 2019-08-08
Control Plane Lock should record which machine has been elected


liztio:
/kind feature

**Describe the solution you'd like**
Right now, the Control Plane Lock does not document what machine has been chosen to be the first control plane (the one that gets `kubeadm init` instead of `kubeadm join`).

If we recorded this, we could have better log messages, record events, and potentially even restart a failed machine creation.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

liztio:
Proposed flow:
1. Write out a configmap that references the machine that was elected
2. Requeue immediately
3. Retrieve Machine listed in ConfigMap
4. Call GetOrCreate()
5. Update [Instance](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/release-0.3/pkg/apis/awsprovider/v1alpha1/types.go#L411)
6. Requeue if not ready

liztio:
GetOrCreate() will reacquire any orphaned EC2 instances
The arl will be populated immediately for easier debugging

I'm worried about duplicating instances during 4) though

ncdc:
> 2. Requeue immediately

Are you suggesting this instead of allowing the winner to immediately proceed so that we only have a single code path & flow to test?

liztio:
It's so we can be sure that the etcd commit was actually successful

ncdc:
commit of what resource?

liztio:
the lock

ncdc:
`ConfigMaps(foo).Create(&cm)` will only return a `nil` error if it succeeded. We don't need to requeue to ensure the ConfigMap was written to etcd.

liztio:
Fair enough, just being defensive

ncdc:
It _could_ be easier to test if we do what you're suggesting. Otherwise, we do have 2 flows:

1. Create lock, proceed because it was successful
1. Handle case where lock was created but something went wrong and we didn't get a happy VM (i.e. proceed b/c Machine name/uid matches what's in the lock)

If we do what you're suggesting, it's:

1. Iteration 1: create lock
1. Iteration 2: query lock & proceed if lock details match current Machine

right?

liztio:
[...]

PR #1002, opened 2019-08-12
[0.3] Record machine

liztio:
Previously, capa would try exactly once to boot the first control plane node, which would be responsible for `kubeadm init`. If that failed for any reason, creation would block forever. And if that ec2  instance isn't properly associated with the machine object, we could potentially leak an ec2 node. This changes things so we can always try to create a machine and fail, we will retry with that same machine again and again.

Part of #981 (only handles v1a1)

**Release note**:

```release-note
creating a new cluster's control plane is more resilient in the face of transient errors from EC2 and Cluster API
```


vincepri:
@liztio Updated the title to reflect this is targeting release-0.3 branch

liztio:
This issue also exists in master, but will need separate implementation (the bulk of it will live in cluster-api-bootstrap-provider-kubeadm)

PR #1017, opened 2019-08-14
fix: avoid duplicate nodeRole label in spec

sethp-nr:

Previously, we observed a machine that was repeatedly failing to create
(we wanted an m5.large, but we were already at our quota for m5.larges)
would get a new copy of the node role appended to its labels every time
it tried.

I don't think it would've hurt anything and so this is a low-priority
(v1alpha1-only) item to be sure.

Fixes #

**Release note**:

```release-note
fixes issue where node-labels would grow unbounded if a machine could not be created
```

### DIFF
--- a/pkg/cloud/aws/services/ec2/instances.go
+++ b/pkg/cloud/aws/services/ec2/instances.go
@@ -324,8 +324,10 @@ func (s *Service) createInstance(machine *actuators.MachineScope, bootstrapToken
 	case "node":
 		s.scope.V(2).Info("Joining a worker node to the cluster")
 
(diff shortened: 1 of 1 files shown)


sethp-nr:
@chuckha I ran into a little trouble trying to check whether this would affect v1alpha2 – I realized I didn't know where in that scheme the provider-specific kubeadm stuff would be coming from. Is that interface defined yet?

chuckha:
This will not affect v1alpha2, the kubeadm work has all moved to CABPK and this style of creating a kubeadm config has completely gone away

Issue #1019, opened 2019-08-15
Status being overwritten in some cases prior to patching status


xrmzju:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
related issue: https://github.com/kubernetes-sigs/cluster-api/issues/1259

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

PR #1020, opened 2019-08-15
use cluster api patch helper to do the scope close

xrmzju:
fix issue: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1019
Fixes #

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -27,6 +27,7 @@ import (
 	"k8s.io/klog/klogr"
 	"sigs.k8s.io/cluster-api-provider-aws/api/v1alpha2"
 	clusterv1alpha2 "sigs.k8s.io/cluster-api/api/v1alpha2"
+	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 
@@ -66,24 +67,29 @@ func NewClusterScope(params ClusterScopeParams) (*ClusterScope, error) {
 		params.AWSClients.ELB = elb.New(session)
 	}
 
+	helper, err := patch.NewHelper(params.AWSCluster, params.Client)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to init patch helper")
+	}
 	return &ClusterScope{
-		Logger:          params.Logger,
-		client:          params.Client,
-		AWSClients:      params.AWSClients,
-		Cluster:         params.Cluster,
-		AWSCluster:      params.AWSCluster,
-		awsClusterPatch: client.MergeFrom(params.AWSCluster.DeepCopy()),
+		Logger:      params.Logger,
+		client:      params.Client,
+		AWSClients:  params.AWSClients,
+		Cluster:     params.Cluster,
+		AWSCluster:  params.AWSCluster,
+		patchHelper: helper,
 	}, nil
 }
 
 // ClusterScope defines the basic context for an actuator to operate upon.
 type ClusterScope struct {
 	logr.Logger
-	client client.Client
+	client      client.Client
+	patchHelper *patch.Helper
+
 	AWSClients
-	Cluster         *clusterv1alpha2.Cluster
-	AWSCluster      *v1alpha2.AWSCluster
-	awsClusterPatch client.Patch
+	Cluster    *clusterv1alpha2.Cluster
+	AWSCluster *v1alpha2.AWSCluster
 }
 
 // Network returns the cluster network object.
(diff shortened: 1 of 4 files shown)


vincepri:
@xrmzju It seems that the issue can be fixed by issuing Status().Patch() before the normal Patch(), see https://github.com/kubernetes-sigs/cluster-api/pull/1261

xrmzju:
I signed it

xrmzju:
> @xrmzju It seems that the issue can be fixed by issuing Status().Patch() before the normal Patch(), see [kubernetes-sigs/cluster-api#1261](https://github.com/kubernetes-sigs/cluster-api/pull/1261)

this is my first solution, and i found a problem here. i  will generate certs info  in my provider controller if not specified in spec and store it in a secret, and after first reconcile,  the certs info and secret are generated normally, after status.Patch, it will trigger a new update event, and the  cert info in the provider spec is still  nil<status.Patch does not update spec>, so the second reconcile round would do the generate certs action again without update the secret info(cause i'll try to get the secret, if exist i will not create it), that causes inconsistence between the cert info in the spec and stored in the secret.                    
the above case is only to declare that if we try to do some <one-time action > in our logic, just put status.Patch before the client.Patch will cause some problem. i think the status copy action would be the most secure one.

vincepri:
@xrmzju Thanks for the detailed explanation! The change LGTM, can we add unit tests to cover the behaviors you've described? It'd be great to cover both the failing and success cases with DeepCopy, so we can track this issue.

@detiber is also working on a generic implementation of patching in Cluster API, once that's ready we should switch to that one.

ncdc:
@detiber your patch helper will handle this, won't it?

detiber:
> @detiber your patch helper will handle this, won't it?

That is correct, an example of usage can be found here: https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/pull/148

ncdc:
@xrmzju do you want to adjust this PR to be based on the new patch helper & the example Jason provided? If not, please let us know, and we can get someone else to work on it.

xrmzju:
> @xrmzju do you want to adjust this PR to be based on the new patch helper & the example Jason provided? If not, please let us know, and we can get someone else to work on it.
@ncdc i'll adjust it.

pkg/cloud/scope/machine.go
```diff
@@ -164,15 +167,5 @@ func (m *MachineScope) SetAnnotation(key, value string) {
...
-	if err := m.client.Patch(ctx, m.AWSMachine, m.patch); err != nil {
-		return errors.WithStack(err)
-	}
-
-	if err := m.client.Status().Patch(ctx, m.AWSMachine, m.patch); err != nil {
-		return errors.WithStack(err)
-	}
-
-	return nil
+	return m.patchHelper.Patch(context.Background(), m.Machine)
```
```suggestion
	return m.patchHelper.Patch(context.Background(), m.AWSMachine)
```

pkg/cloud/scope/cluster.go
```diff
@@ -66,13 +67,17 @@ func NewClusterScope(params ClusterScopeParams) (*ClusterScope, error) {
 		params.AWSClients.ELB = elb.New(session)
 	}
 
+	helper, err := patch.NewHelper(params.AWSCluster, params.Client)
+	if err != nil {
+		return nil, errors.Errorf("failed to init patch helper: %v", err)
```
ncdc: Please use `errors.Wrap`

pkg/cloud/scope/machine.go
```diff
@@ -63,25 +63,28 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
 		params.Logger = klogr.New()
 	}
 
+	helper, err := patch.NewHelper(params.AWSMachine, params.Client)
+	if err != nil {
+		return nil, errors.Errorf("failed to init patch helper: %v", err)
```
ncdc: Please use `errors.Wrap`

pkg/cloud/scope/machine.go
```diff
@@ -63,25 +63,28 @@ func NewMachineScope(params MachineScopeParams) (*MachineScope, error) {
...
+		AWSMachine:  params.AWSMachine,
+		Logger:      params.Logger,
+		patchHelper: helper,
 	}, nil
 }
 
 // MachineScope defines a scope defined around a machine and its cluster.
 type MachineScope struct {
 	logr.Logger
-	patch  client.Patch
```
ncdc: A lot of times we separate the unexported from the exported fields, as was the case here. I'd recommend you keep this pattern - move `patchHelper` back up here, and preserve the blank line in between `client` and `Cluster`.

pkg/cloud/scope/machine.go
xrmzju: my fault... done.

pkg/cloud/scope/cluster.go
xrmzju: done

xrmzju: done

pkg/cloud/scope/machine.go
xrmzju: done

pkg/cloud/scope/machine.go
xrmzju: ok, i'll keep that pattern

ncdc:
/lgtm
/approve

/assign @detiber 
Jason, can you give it one final look, then cancel the hold? Thanks!

PR #1034, opened 2019-08-19
chore: set status.ready when machine is running

### LINKED ISSUE #1032: Status.Ready is never set on AWSMachines
/kind bug

- spun up CAPI/CAPA/CABPK
- applied some YAML for controlplane
- infrastructure is provisioned, everything seems cool
- applied some YAML for a worker node

The node doesn't join because InfrastructureReady is not set on the controlplane Machine. Did some digging and found that CAPA never sets Status.Ready, which is what CAPI looks for. 

- worker node joins and everything is great

**Environment:**

rudoi:
Sets `.Status.Ready` to true when instance is running.

Fixes #1032 

Is there anywhere in here we need to set it to false or is that too spicy?

**Release note**:

```release-note

```

### DIFF
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -139,6 +139,11 @@ func (m *MachineScope) SetInstanceState(v infrav1.InstanceState) {
 	m.AWSMachine.Status.InstanceState = &v
 }
 
+// SetReady sets the AWSMachine Ready Status
+func (m *MachineScope) SetReady() {
+	m.AWSMachine.Status.Ready = true
+}
+
 // SetErrorMessage sets the AWSMachine status error message.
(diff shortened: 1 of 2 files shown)


rudoi:
@vincepri looking for some tips on adding a test for this. It doesn't look like we have any existing testing for the machinescope-related functions? Also the controller tests don't seem set up to work with mock AWS interfaces or stuff like that yet, so I'm just wondering what the most effective approach would be.

vincepri:
Let's open up an issue to tackle unit testing the controller instead?

rudoi:
> Let's open up an issue to tackle unit testing the controller instead?

https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1035

PR #1042, opened 2019-08-19
fix: add capi to scheme, small yaml fixes

rudoi:
1. adds CAPI to scheme in main.go which is required for the client to recognize CAPI types
2. adds rbac for machines, machines/status, clusters, clusters/status, and events
3. sets `AWS_SHARED_CREDENTIALS_FILE` in the pod. I believe the user `nobody` is expected to have a home of `/` rather than `/home`, so this var must be set if we want to mount the secret at `/home/.aws`. Let me know if I'm way off on this one, but I got creds errors up until I set this env var. 

**Release note**:

```release-note
add capi to scheme, rbac fixes
```


config/default/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: IMAGE_URL
+      - image: cf-registry.nr-ops.net/arudoi/manager:dev
```
ncdc: Undo 😄

controllers/awsmachine_controller.go
```diff
@@ -56,6 +56,7 @@ type AWSMachineReconciler struct {
 
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsmachines,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsmachines/status,verbs=get;update;patch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machines;machines/status,verbs=get;list;watch
```
detiber: Events rbac should likely also be added here just for documentation.

config/default/manager_image_patch.yaml
rudoi: nothing to see here

controllers/awsmachine_controller.go
rudoi: done

rudoi:
i forgot events create rbac so just added

ncdc:
LGTM. @rudoi could you please squash your commits?

rudoi:
> LGTM. @rudoi could you please squash your commits?

👍

PR #1047, opened 2019-08-20
✨Add automation examples for v1alpha2 clusters

### LINKED ISSUE #1018: Generate examples for v1alpha2
We need to be able to generate example clusters (to also include into release artifacts for each release). From v1alpha1 to v1alpha2, in v1alpha1 each provider vendored Cluster API's YAML and built on top of it. For v1alpha2 each provider is responsible for its own yaml because it does not vendor cluster-api.

Related: Examples generation scripts were removed here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/994/files

/assign

vincepri:

This PR re-introduces the `examples/` folder, which gives users a way to play around with a simple cluster.

Fixes #1018 

**Release note**:

```release-note

```


examples/addons.yaml
```diff
@@ -0,0 +1,750 @@
+# Calico Version v3.6
```
detiber: Probably should update to 3.8 while we are adding manifests back: https://docs.projectcalico.org/v3.8/manifests/calico.yaml

examples/generate.sh
```diff
@@ -0,0 +1,106 @@
...
+
+if [ $OVERWRITE -ne 1 ] && [ -d "$OUTPUT_DIR" ]; then
+  echo "ERR: Folder ${OUTPUT_DIR} already exists. Delete it manually before running this script."
+  exit 1
+fi
+
+mkdir -p "${OUTPUT_DIR}"
+
+# Generate cluster resources.
+kustomize build "${SOURCE_DIR}/cluster" | envsubst > "${CLUSTER_GENERATED_FILE}"
```
detiber: should this use `kubectl kustomize` instead?

vincepri: It fails for me

```
Error: rawResources failed to read Resources: Load from path ../crd failed: '../crd' must be a file (got d='/Users/vince/go/src/sigs.k8s.io/cluster-api-provider-aws/config/crd')
```

detiber: Do we need to enforce a minimum version of kustomize then?

examples/addons.yaml
vincepri: Updated 😊

examples/generate.sh
vincepri: Yeah, I've been using v3.1, I can add a check in this script

vincepri: Added the apiVersion and Kind to the kustomize scripts, that should fix the version

examples/generate.sh
```diff
@@ -0,0 +1,106 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -o errexit
+set -o nounset
+
+# Directories.
+SOURCE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+OUTPUT_DIR=${OUTPUT_DIR:-${SOURCE_DIR}/_out}
+ENVSUBST=${ENVSUBST:-envsubst}
```
ncdc: Add a check to make sure this exists/is in PATH?

examples/generate.sh
```diff
@@ -0,0 +1,106 @@
...
+CLUSTERAWSADM=${CLUSTERAWSADM:-clusterawsadm}
```
ncdc: Add a check to make sure this exists/is in PATH?

examples/generate.sh
ncdc: I'd rather use `kubectl kustomize` if possible. It sounds like we may need to be more creative with bases vs resources. (I am nowhere near a kustomize expert.)

examples/generate.sh
```diff
@@ -0,0 +1,106 @@
...
+# Generate cluster resources.
+kustomize build "${SOURCE_DIR}/cluster" | envsubst > "${CLUSTER_GENERATED_FILE}"
+echo "Generated ${CLUSTER_GENERATED_FILE}"
+
+# Generate controlplane resources.
+kustomize build "${SOURCE_DIR}/controlplane" | envsubst > "${CONTROLPLANE_GENERATED_FILE}"
+echo "Generated ${CONTROLPLANE_GENERATED_FILE}"
+
+# Generate Cluster API provider components file.
+kustomize build "github.com/kubernetes-sigs/cluster-api//config/default/?ref=master" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
```
ncdc: Is the extra slash in `cluster-api//config` a typo?

Also, I don't think `ref=master` is going to be viable in the long term.

examples/generate.sh
```diff
@@ -0,0 +1,106 @@
...
+echo "Generated ${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
+
+# Generate Kubeadm Bootstrap Provider components file.
+kustomize build "github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm//config/default/?ref=master" > "${COMPONENTS_KUBEADM_GENERATED_FILE}"
+echo "Generated ${COMPONENTS_KUBEADM_GENERATED_FILE}"
+
+# Generate AWS Credentials.
+AWS_B64ENCODED_CREDENTIALS="$(${CLUSTERAWSADM} alpha bootstrap encode-aws-credentials)"
```
ncdc: This fails if `AWS_REGION` is unset. Should we add a check for this at the top of this script?

examples/generate.sh
vincepri: `kubectl kustomize` is a few versions behind and it's impossible right now to understand which version it's actually vendoring (at least I couldn't find it). For now, let's use plain `kustomize` and move to kubectl kustomize if/when it checks all the boxes.

examples/generate.sh
vincepri: `clusterawsadm` prints an error if it cannot generate find credentials, that should be enough

examples/generate.sh
vincepri: Not a typo 😄 It's to differentiate between the repository and the path

vincepri: Once we release a version, we can reference the tag directly instead of `master`

examples/generate.sh
```diff
@@ -0,0 +1,111 @@
...
+set -o errexit
+set -o nounset
+
+# Directories.
+SOURCE_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null && pwd )"
+OUTPUT_DIR=${OUTPUT_DIR:-${SOURCE_DIR}/_out}
+
+# Binaries
+ENVSUBST=${ENVSUBST:-envsubst}
+command -v "${ENVSUBST}" || echo -v "Cannot find ${ENVSUBST} in path."
```
ncdc: redirect the output to `/dev/null`

examples/generate.sh
```diff
@@ -0,0 +1,111 @@
...
+
+CLUSTERAWSADM=${CLUSTERAWSADM:-clusterawsadm}
+command -v "${CLUSTERAWSADM}" || echo -v "Cannot find ${CLUSTERAWSADM} in path, build it using 'make binaries' in this repository."
```
ncdc: redirect to `/dev/null`

examples/generate.sh
ncdc: It's v2.0.3 in k/k master now (https://github.com/kubernetes/kubernetes/blob/d951d19ec4d8ec829e83d146200ea91b664b6fa6/go.mod#L175)

examples/generate.sh
ncdc: It took several seconds of running before it failed for me. I'll leave it up to you.

examples/generate.sh
vincepri: Yeah, we're using v3.1 within Cluster API repositories

examples/generate.sh
vincepri: Got it, I moved clusterawsadm as the first thing to generate. The main time is spent fetching the kustomize remote repos (CAPI, and kubeadm bootstrapper)

vincepri:
@ncdc Good to go?

ncdc:
Gah, needs rebase

vincepri:
@ncdc rebased, thanks!

PR #1081, opened 2019-08-28
:sparkles: Functions for converting v1a1 types to v1a2

### LINKED ISSUE #917: [v1alpha2] Conversion tool from v1alpha1
/kind feature

We need to be able to take existing v1alpha1 resources and convert them to v1alpha2 resources that function with v1alpha2 components. We hopefully can reuse portions of this for other kubeadm-based providers (such as CAPV), but for now, we can start building a tool that is specific to CAPA.

This should be a standalone executable that we can run against a Kubernetes apiserver. We should be able to specify working in either 1 namespace, or all namespaces.

For each v1alpha1 `Cluster`, the tool should:
- Create a v1alpha2 `KubeadmClusterBootstrapConfig` (does not exist yet, name TBD). Need to come up with a naming convention (e.g. just the name of the cluster)
- Create a v1alpha2 `Cluster`

For each v1alpha1 `Machine`, the tool should:
- Create a v1alpha2 `KubeadmBootstrapConfig` (maybe we should rename to `KubeadmMachineBootstrapConfig`)
- Create a v1alpha2 `AWSMachine`
- Create a v1alpha2 `Machine`

For each v1alpha1 `MachineSet`, the tool should:
- Create a v1alpha2 `KubeadmBootstrapConfigTemplate` (maybe we should rename to `KubeadmMachineBootstrapConfig`)
- Create a v1alpha2 `AWSMachineTemplate`
- Create a v1alpha2 `MachineSet`

For each v1alpha1 `MachineDeployment`, the tool should do the same thing it's doing for `MachineSet` but for `MachineDeployments`.

**Environment:**

liztio:
As v1alpha2 nears release, we're going to need some tooling for converting clusters from v1a1 to v1a2. This pull request _does not_ provide those tools directly, but it does do an important part: Functions that one type and return the other type. These can be integrated into CLIs or controllers later on.

Fixes #917 

**Release note**:

```release-note
NONE
```


api/conversion/cluster.go
```diff
@@ -0,0 +1,94 @@
...
+		if err := yaml.Unmarshal(c.oldCluster.Spec.ProviderSpec.Value.Raw, &oldAWSCluster); err != nil {
+			return nil, errors.Wrap(err, "couldn't decode ProviderSpec")
+		}
+
+		c.oldAWSCluster = &oldAWSCluster
+	}
+
+	return c.oldAWSCluster, nil
+}
+
```
detiber: The ClusterConverter could likely use a method `GetSecrets()` or similar that would provide the proper format for the cluster CA, etcd CA, frontproxy CA, and service account secrets, since they are discarded in `Convert_v1alpha1_AWSClusterProviderSpec_To_v1alpha2_AWSClusterSpec` and we would not want to regenerate them in the bootstrap controller after conversion.

detiber:
/lgtm
/hold
holding for review from @ncdc

ncdc:
First item in my queue for this morning

api/conversion/machine_test.go
```diff
@@ -0,0 +1,461 @@
...
+
+	assert.stringPtrEqual(oldMachine.Spec.ProviderID, newMachine.Spec.ProviderID, "provider ID")
+
+	if err := converter.GetAWSMachine(&newAWSMachine); err != nil {
+		t.Fatalf("Unexpected error converting AWSMachine: %v", err)
+	}
+
+	t.Logf("converted machine: %+v", newAWSMachine)
+
+	// Pull the provider ID from the machine (why is it in both places?)
```
ncdc: It's in both places because the infra provider is the primary component responsible for the value, but CAPI needs it for setting node refs, so CAPI copies it from the infra resource to the machine.

go.mod
```diff
@@ -21,6 +21,7 @@ require (
 	k8s.io/kubernetes v1.14.2
 	k8s.io/utils v0.0.0-20190809000727-6c36bc71fc4a
 	sigs.k8s.io/cluster-api v0.2.0
+	sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm v0.0.0-20190903162828-be5a88101a9c
```
ncdc: Can you update this to the v0.1.0 tag?

config/manager/credentials.yaml
```diff
@@ -5,4 +5,4 @@ metadata:
   namespace: system
 type: Opaque
 data:
-  credentials: ${AWS_B64ENCODED_CREDENTIALS}
+  credentials: W2RlZmF1bHRdCmF3c19hY2Nlc3Nfa2V5X2lkID0gQUtJQVFYQ0tSTUdIN0hURDJRNU4KYXdzX3NlY3JldF9hY2Nlc3Nfa2V5ID0gQlh1Qkl6WWtpUG9rTEI2Ym5UOUNmWSs2ajVsdC9DdVUwRURPbEpxMwpyZWdpb24gPSB1cy1lYXN0LTEKCg==
```
ncdc: Revert?

ncdc: And disable/delete these creds?

liztio: ugh yeah

ncdc:
Once you yank out your creds, can you squash down to as many commits as makes sense? Thanks!

PR #1088, opened 2019-08-29
🐛 Fix bug with requeuing machines from AWSCluster requests

chuckha:

This PR fixes an issue with listing machines using a machine label.

/assign @vincepri

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -413,15 +413,28 @@ func (r *AWSMachineReconciler) AWSClusterToAWSMachines(o handler.MapObject) []ct
 		r.Log.Error(errors.Errorf("expected a AWSCluster but got a %T", o.Object), "failed to get AWSMachine for AWSCluster")
 		return nil
 	}
+	log := r.Log.WithValues("AWSCluster", c.Name, "Namespace", c.Namespace)
 
-	labels := map[string]string{clusterv1.MachineClusterLabelName: c.Name}
-	machineList := &infrav1.AWSMachineList{}
+	cluster, err := util.GetOwnerCluster(context.Background(), r.Client, c.ObjectMeta)
+	switch {
+	case apierrors.IsNotFound(err) || cluster == nil:
+		return result
+	case err != nil:
+		log.Error(err, "failed to get owning cluster")
+		return result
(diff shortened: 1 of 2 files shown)


controllers/awsmachine_controller.go
```diff
@@ -415,13 +415,13 @@ func (r *AWSMachineReconciler) AWSClusterToAWSMachines(o handler.MapObject) []ct
 	}
 
 	labels := map[string]string{clusterv1.MachineClusterLabelName: c.Name}
```
vincepri: Doesn't this assume that the AWSCluster is going to have the same name as the Cluster?

detiber: This is used in the label matcher below for getting the machines for the cluster in the query below.

controllers/awsmachine_controller.go
```diff
@@ -415,13 +415,13 @@ func (r *AWSMachineReconciler) AWSClusterToAWSMachines(o handler.MapObject) []ct
...
-	machineList := &infrav1.AWSMachineList{}
+	machineList := &clusterv1.MachineList{}
 	if err := r.List(context.Background(), machineList, client.InNamespace(c.Namespace), client.MatchingLabels(labels)); err != nil {
-		r.Log.Error(err, "failed to list AWSMachines", "AWSCluster", c.Name, "Namespace", c.Namespace)
+		r.Log.Error(err, "failed to list Machines", "AWSCluster", c.Name, "Namespace", c.Namespace)
 		return nil
 	}
 	for _, m := range machineList.Items {
-		name := client.ObjectKey{Namespace: m.Namespace, Name: m.Name}
+		name := client.ObjectKey{Namespace: m.Namespace, Name: 	m.Spec.InfrastructureRef.Name}
```
detiber: I'm wondering if we need to guard against the InfrastructureRef not being set yet here.

chuckha: I thought about that yesterday after i turned my computer off. +1, will update

controllers/awsmachine_controller.go
detiber: nm, @vincepri is correct, we are using the name of the AWSCluster here as the Cluster name.

PR #1104, opened 2019-09-04
:running: Add release binaries target

vincepri:

Fixes #

**Release note**:

```release-note

```


Makefile
```diff
@@ -208,19 +208,42 @@ set-manifest-image:
...
-	mkdir -p out/
-	kustomize build config/default > out/infrastructure-components.yaml
+	$(MAKE) release-manifests
+	$(MAKE) release-binaries
+
+.PHONY: release-manifests
+release-manifests: $(RELEASE_DIR) ## Builds the manifests to publish with a release
+	kustomize build config/default > $(RELEASE_DIR)/infrastructure-components.yaml
+
+release-binaries: ## Builds the binaries to publish with a release
```
detiber: Shouldn't this be a phony target?

Makefile
```diff
@@ -208,19 +208,42 @@ set-manifest-image:
...
+	RELEASE_BINARY=./cmd/clusterawsadm GOOS=linux GOARCH=amd64 $(MAKE) release-binary
+	RELEASE_BINARY=./cmd/clusterawsadm GOOS=darwin GOARCH=amd64 $(MAKE) release-binary
+
+release-binary: $(RELEASE_DIR)
```
detiber: Same here, since it's not defining a specific output path that exists post-build, shouldn't this be a phony target?

ncdc: Let's not do phone - 1 sec

ncdc: This can be `$(RELEASE_DIR)/$(RELEASE_BINARY)-$(GOOS)-$(GOARCH)`

ncdc: See e.g. https://github.com/heptio/velero/blob/73488cdc0c649dc7ec9a692eab934c14ceb3a4b6/Makefile#L18, https://github.com/heptio/velero/blob/73488cdc0c649dc7ec9a692eab934c14ceb3a4b6/Makefile#L40, https://github.com/heptio/velero/blob/73488cdc0c649dc7ec9a692eab934c14ceb3a4b6/Makefile#L43-L45, https://github.com/heptio/velero/blob/73488cdc0c649dc7ec9a692eab934c14ceb3a4b6/Makefile#L97

detiber: Should we trust that if we aren't specifying input source(s)?

ncdc: In general, I'd prefer to avoid phony targets if we can. Given that we only produce a single binary now, we can err on the side of specificity and have `$(RELEASE_DIR)/clusterawsadm-$(GOOS)-$(GOARCH)` as the target name - does that alleviate your concern?

detiber: I more meant that we are relying on the removal of RELEASE_DIR in the `release` target to ensure that we are building a new binary, since we do not specify source files to watch for updates to build the target.

ncdc: Oh I see what you're saying now. I wish there were a good way in a Makefile to associate a binary with go source files. Ok let's do phony for the time being.

vincepri:
All fixed

PR #1116, opened 2019-09-10
✨ Add validating webhook for AWSMachineTemplate

### LINKED ISSUE #886: Add validating webhooks
/kind feature

Using validating webhook, we need to add validation logic for AWSMachineTemplate objects to achieve immutability.

tahsinrahman:

Fixes #886

### DIFF
--- a/config/certmanager/certificate.yaml
+++ b/config/certmanager/certificate.yaml
@@ -1,6 +1,6 @@
 # The following manifests contain a self-signed issuer CR and a certificate CR.
 # More document can be found at https://docs.cert-manager.io
-apiVersion: certmanager.k8s.io/v1alpha2
+apiVersion: certmanager.k8s.io/v1alpha1
 kind: Issuer
 metadata:
   name: selfsigned-issuer
@@ -8,16 +8,16 @@ metadata:
 spec:
   selfSigned: {}
 ---
-apiVersion: certmanager.k8s.io/v1alpha2
+apiVersion: certmanager.k8s.io/v1alpha1
 kind: Certificate
 metadata:
   name: serving-cert  # this name should match the one appeared in kustomizeconfig.yaml
   namespace: system
 spec:
-  # $(SERVICENAME) and $(NAMESPACE) will be substituted by kustomize
-  commonName: $(SERVICENAME).$(NAMESPACE).svc
+  # $(SERVICE_NAME) and $(SERVICE_NAMESPACE) will be substituted by kustomize
+  commonName: $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc
   dnsNames:
-  - $(SERVICENAME).$(NAMESPACE).svc.cluster.local
+  - $(SERVICE_NAME).$(SERVICE_NAMESPACE).svc.cluster.local
   issuerRef:
     kind: Issuer
     name: selfsigned-issuer
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -1,5 +1,3 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
 # This kustomization.yaml is not intended to be run by itself,
 # since it depends on service name and namespace that are out of this kustomize package.
 # It should be run by config/default
@@ -9,7 +7,7 @@ resources:
 - bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml
 # +kubebuilder:scaffold:crdkustomizeresource
 
-#patches:
+patchesStrategicMerge:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
 # patches here are for enabling the conversion webhook for each CRD
 #- patches/webhook_in_awsmachines.yaml
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -1,6 +1,3 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
-kind: Kustomization
-
 # Adds namespace to all resources.
 namespace: capa-system
 
@@ -15,13 +12,13 @@ namePrefix: capa-
 #commonLabels:
 #  someName: someValue
 
-resources:
+bases:
 - ../crd
 - ../rbac
 - ../manager
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in crd/kustomization.yaml
 #- ../webhook
-# [CERTMANAGER] To enable cert-manager, uncomment next line. 'WEBHOOK' components are required.
+# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
 #- ../certmanager
 
 patchesStrategicMerge:
@@ -42,7 +39,37 @@ patchesStrategicMerge:
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in crd/kustomization.yaml
 #- manager_webhook_patch.yaml
 
-# [CAINJECTION] Uncomment next line to enable the CA injection in the admission webhooks.
-# Uncomment 'CAINJECTION' in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
+# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
+# Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
 # 'CERTMANAGER' needs to be enabled to use ca injection
 #- webhookcainjection_patch.yaml
+
+# the following config is for teaching kustomize how to do var substitution
+vars:
+# [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER' prefix.
+#- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
+#  objref:
+#    kind: Certificate
+#    group: certmanager.k8s.io
+#    version: v1alpha1
+#    name: serving-cert # this name should match the one in certificate.yaml
+#  fieldref:
+#    fieldpath: metadata.namespace
+#- name: CERTIFICATE_NAME
+#  objref:
+#    kind: Certificate
+#    group: certmanager.k8s.io
+#    version: v1alpha1
+#    name: serving-cert # this name should match the one in certificate.yaml
+#- name: SERVICE_NAMESPACE # namespace of the service
+#  objref:
+#    kind: Service
+#    version: v1
+#    name: webhook-service
+#  fieldref:
+#    fieldpath: metadata.namespace
+#- name: SERVICE_NAME
+#  objref:
+#    kind: Service
+#    version: v1
+#    name: webhook-service
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -17,9 +17,9 @@ bases:
 - ../rbac
 - ../manager
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in crd/kustomization.yaml
-#- ../webhook
+- ../webhook
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'. 'WEBHOOK' components are required.
-#- ../certmanager
+- ../certmanager
 
 patchesStrategicMerge:
 - manager_image_patch.yaml
@@ -37,39 +37,39 @@ patchesStrategicMerge:
 #- manager_prometheus_metrics_patch.yaml
 
 # [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix including the one in crd/kustomization.yaml
-#- manager_webhook_patch.yaml
+- manager_webhook_patch.yaml
 
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER'.
 # Uncomment 'CERTMANAGER' sections in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
 # 'CERTMANAGER' needs to be enabled to use ca injection
-#- webhookcainjection_patch.yaml
+- webhookcainjection_patch.yaml
 
 # the following config is for teaching kustomize how to do var substitution
 vars:
 # [CERTMANAGER] To enable cert-manager, uncomment all sections with 'CERTMANAGER' prefix.
-#- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
-#  objref:
-#    kind: Certificate
-#    group: certmanager.k8s.io
-#    version: v1alpha1
-#    name: serving-cert # this name should match the one in certificate.yaml
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: CERTIFICATE_NAME
-#  objref:
-#    kind: Certificate
-#    group: certmanager.k8s.io
-#    version: v1alpha1
-#    name: serving-cert # this name should match the one in certificate.yaml
-#- name: SERVICE_NAMESPACE # namespace of the service
-#  objref:
-#    kind: Service
-#    version: v1
-#    name: webhook-service
-#  fieldref:
-#    fieldpath: metadata.namespace
-#- name: SERVICE_NAME
-#  objref:
-#    kind: Service
-#    version: v1
-#    name: webhook-service
+- name: CERTIFICATE_NAMESPACE # namespace of the certificate CR
+  objref:
+    kind: Certificate
+    group: certmanager.k8s.io
+    version: v1alpha1
+    name: serving-cert # this name should match the one in certificate.yaml
+  fieldref:
+    fieldpath: metadata.namespace
+- name: CERTIFICATE_NAME
+  objref:
+    kind: Certificate
+    group: certmanager.k8s.io
+    version: v1alpha1
+    name: serving-cert # this name should match the one in certificate.yaml
+- name: SERVICE_NAMESPACE # namespace of the service
+  objref:
+    kind: Service
+    version: v1
+    name: webhook-service
+  fieldref:
+    fieldpath: metadata.namespace
+- name: SERVICE_NAME
+  objref:
+    kind: Service
+    version: v1
+    name: webhook-service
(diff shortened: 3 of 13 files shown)


Dockerfile
```diff
@@ -40,5 +40,5 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} \
 FROM gcr.io/distroless/static:latest
 WORKDIR /
 COPY --from=builder /workspace/manager .
-USER nobody
+# USER nobody
```
tahsinrahman: it is removed as I was getting this error
```
E0910 19:34:39.892550       1 main.go:164] setup "msg"="problem running manager" "error"="listen tcp :443: bind: permission denied"  
```

api/v1alpha2/awsmachinetemplate_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+		Complete()
+}
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha2-awsmachinetemplate,mutating=false,failurePolicy=fail,groups=infrastructure.cluster.x-k8s.io,resources=awsmachinetemplates,versions=v1alpha2,name=vawsmachinetemplate.kb.io
+
+var _ webhook.Validator = &AWSMachineTemplate{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachineTemplate) ValidateCreate() error {
+	return r.validateAWSMachineTemplate()
```
ncdc: I would `return nil` here, and add a comment that there is no validation on create.

api/v1alpha2/awsmachinetemplate_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+	oldAWSMachineTemplate := old.(*AWSMachineTemplate)
+	return validateUpdate(&r.Spec.Template.Spec, &oldAWSMachineTemplate.Spec.Template.Spec)
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachineTemplate) ValidateDelete() error {
+	return nil
+}
+
+func (r *AWSMachineTemplate) validateAWSMachineTemplate() error {
```
ncdc: I'd remove this

api/v1alpha2/awsmachinetemplate_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+	return nil
+}
+
+// validateUpdate checks that no immutable fields have been updated
+func validateUpdate(spec *AWSMachineSpec, oldSpec *AWSMachineSpec) error {
+	// Instance Type
+	if spec.InstanceType != oldSpec.InstanceType {
+		return errors.Errorf("instance type cannot be mutated from %q to %q", oldSpec.InstanceType, spec.InstanceType)
```
ncdc: It would probably be better to aggregate these using k8s.io/apimachinery/pkg/api/errors.NewInvalid (or whatever makes the most sense). I would aggregate each field-specific error into a `field.ErrorList`, using methods such as `field.Forbidden(field.NewPath("spec", "instanceType"), "instance type is immutable")`

api/v1alpha2/awsmachinetemplate_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+	// as a *string, so do the same here.
+	if spec.Subnet != nil {
+		if *spec.Subnet.ID != *oldSpec.Subnet.ID {
+			return errors.Errorf("machine subnet ID cannot be mutated from %q to %q",
+				*oldSpec.Subnet.ID, *spec.Subnet.ID)
+		}
+	}
+
+	// Public IP
+	if *spec.PublicIP != *oldSpec.PublicIP {
```
ncdc: nil check?

api/v1alpha2/awsmachinetemplate_webhook.go
```diff
@@ -0,0 +1,96 @@
...
+	if spec.RootDeviceSize > 0 && spec.RootDeviceSize != oldSpec.RootDeviceSize {
+		return errors.Errorf("Root volume size cannot be mutated from %v to %v", oldSpec.RootDeviceSize, spec.RootDeviceSize)
+	}
+
+	// Subnet ID
+	// spec.Subnet is a *AWSResourceReference and could technically be
+	// a *string, ARN or Filter. However, elsewhere in the code it is only used
+	// as a *string, so do the same here.
+	if spec.Subnet != nil {
+		if *spec.Subnet.ID != *oldSpec.Subnet.ID {
```
ncdc: nil check on oldSpec.Subnet?

api/v1alpha2/zz_generated.deepcopy.go
```diff
@@ -22,7 +22,7 @@ package v1alpha2
 
 import (
 	"k8s.io/api/core/v1"
-	runtime "k8s.io/apimachinery/pkg/runtime"
```
ncdc: Did you regenerate this file or what led to this change?

config/certmanager/certificate.yaml
```diff
@@ -1,23 +1,23 @@
 # The following manifests contain a self-signed issuer CR and a certificate CR.
 # More document can be found at https://docs.cert-manager.io
-apiVersion: certmanager.k8s.io/v1alpha2
+apiVersion: certmanager.k8s.io/v1alpha1
```
ncdc: Why the version downgrade?

config/default/kustomization.yaml
```diff
@@ -39,9 +39,39 @@ patchesStrategicMerge:
...
-#- manager_webhook_patch.yaml
+- manager_webhook_patch.yaml
 
 # [CAINJECTION] Uncomment next line to enable the CA injection in the admission webhooks.
 # Uncomment 'CAINJECTION' in crd/kustomization.yaml to enable the CA injection in the admission webhooks.
 # 'CERTMANAGER' needs to be enabled to use ca injection
-#- webhookcainjection_patch.yaml
+- webhookcainjection_patch.yaml
+
+# the following config is for teaching kustomize how to do var substitution
```
ncdc: Why did you move this?

config/webhook/kustomizeconfig.yaml
```diff
@@ -0,0 +1,18 @@
+# the following config is for teaching kustomize where to look at when substituting vars.
+# It requires kustomize v2.1.0 or newer to work properly.
```
ncdc: We require v3.1.0 or newer

Dockerfile
ncdc: Let's undo this. We can control it using the manager Options struct, field=Port. We can create a flag for this in main.go.

config/certmanager/certificate.yaml
tahsinrahman: certmanager is actually in v1alpha1 https://github.com/jetstack/cert-manager/blob/master/deploy/manifests/00-crds.yaml

config/webhook/kustomizeconfig.yaml
tahsinrahman: I created a sample project using kubebuilder 2.0 and copied the file from there

config/default/kustomization.yaml
tahsinrahman: I was getting this error
```console
$ kustomize build config/default/
Error: var '{SERVICE_NAME ~G_v1_Service {metadata.name}}' cannot be mapped to a field in the set of known resources
```

besides, kubebuilder creates it inside this directory, so i moved it there

api/v1alpha2/zz_generated.deepcopy.go
tahsinrahman: oops, probably my IDE gofmt/goimport-ed this file

api/v1alpha2/awsmachinetemplate_webhook.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachineTemplate) ValidateDelete() error {
+	// there's no validation for AWSMachineTempalate on delete
+	return nil
+}
+
+// validateUpdate checks that no immutable fields have been updated
+func validateUpdate(spec *AWSMachineSpec, oldSpec *AWSMachineSpec) error {
```
detiber: Are there any fields that we want to be mutable in the template types? @ncdc @vincepri I thought we had wanted to avoid allowing mutability of the template types and instead wanted to require users to update the reference in the MachineDeployment/MachineSet?

If we do want to force full immutability, then we can likely create a generic webhook in CAPI that blocks any update, and just register the webhook for the AWSMachineTemplate type here in CAPA.

config/default/kustomization.yaml
ncdc: Ah, thanks. Was this maybe generated by an earlier version of kubebuilder, perhaps? I'm definitely 👍 to match how/where kubebuilder generates things.

api/v1alpha2/awsmachinetemplate_webhook.go
ncdc: We can allow updates to metadata, but probably all other updates should be rejected. I think we could probably write some generic validation helpers in CAPI, but afaik we still need to do the full webhook code generation & registration for each type.

detiber: According to the docs, we can configure which resources are subject to which admission webhooks at runtime: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#configure-admission-webhooks-on-the-fly

vincepri: We should avoid naming changes to resources given that they're references elsewhere, unless the OwnerReferences get updated automagically?

detiber: @ncdc I just want to avoid us requiring  all providers to need to re-implement the exact same thing

Dockerfile
joonas: Filed this as https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1120

api/v1alpha2/awsmachinetemplate_webhook.go
vincepri: I think that's fair, +1

ncdc: @detiber are you thinking that there would be one `ValidatingWebhookConfiguration` in CAPI and we would have documentation for how people installing providers would need to update it to include their resources in the generic template validator?

detiber: @ncdc possibly, we could also potentially give RBAC permissions to providers to register the types in the ValidatingWebhookConfiguration themselves when the controller manager starts up.

vincepri: Can we dynamically watch unstructured objects in a group? And maybe discover those groups automatically?

CAPI should already have permissions for bootstrap.* and infrastructure.* groups

ncdc: We typically advise that registering things like CRDs and webhook configurations are tasks better left to admin accounts. I would be hesitant to let the controllers do this.

ncdc: This sounds like something that clusteradm could handle long-term

vincepri: Maybe we're talking about different things, I was thinking to make this Webhook server available by default in CAPI, and give it permissions to the groups and validate those objects

ncdc: @vincepri yes, we are talking about different things. Thanks for pointing that out. So you'd have this generic webhook use API discovery and look for `*Template` types? Although that wouldn't automatically cover third party providers whose API groups are different.

vincepri: As long as it's documented, I think that should be fine. Technically, the same applies for today's default RBAC permissions in CAPI, where we assume `bootstrap.<>` and `infrastructure.<>` as default

detiber: @ncdc agreed that clusteradm would be the best place long term to handle this

detiber: @vincepri @ncdc the validating webhook requests are initiated by the API Server, I'm not sure there is a way to use API discovery to register types to watch at runtime as we do with the controllers.

ncdc: @detiber that's right. So the only way to handle this is with the details in the ValidatingWebhookConfiguration.

What do you all think about putting the generic *Template validation as a helper function in cluster-api, but at least for now, having each provider implement their own webhooks for their own templates (using the generic function from cluster-api)?

detiber: @ncdc I think that's fine as long as we track improving that experience later when clusteradm is available.

api/v1alpha2/awsmachinetemplate_webhook.go
```diff
@@ -0,0 +1,105 @@
...
+	if spec.RootDeviceSize > 0 && spec.RootDeviceSize != oldSpec.RootDeviceSize {
+		errs = append(errs, field.Forbidden(field.NewPath("spec", "rootDeviceSize"),
+			"root volume size is immutable"))
+	}
+
+	// Subnet ID
+	// spec.Subnet is a *AWSResourceReference and could technically be
+	// a *string, ARN or Filter. However, elsewhere in the code it is only used
+	// as a *string, so do the same here.
+	if aws.StringValue(spec.Subnet.ID) != aws.StringValue(oldSpec.Subnet.ID) {
```
liztio: We should probably check that `spec.Subnet` isn't nil

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,26 @@
+
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
+kind: ValidatingWebhookConfiguration
+metadata:
+  creationTimestamp: null
+  name: validating-webhook-configuration
+webhooks:
+- clientConfig:
+    caBundle: Cg==
```
liztio: What does this do?

ncdc: This is generated by kubebuilder. I guess it's a required field. Maybe cert-manager or something else updates it? It decodes to `\n`.

config/webhook/service.yaml
```diff
@@ -0,0 +1,11 @@
+apiVersion: v1
+kind: Service
+metadata:
+  name: webhook-service
```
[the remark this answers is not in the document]
tahsinrahman: yes, kustomize adds `capa-` prefix to name and namespace!

controllers/awsmachine_controller.go
```diff
@@ -282,13 +280,6 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 		return reconcile.Result{}, nil
 	}
 
-	// TODO(ncdc): move this validation logic into a validating webhook
```
chuckha: I don't see this logic moved anywhere, am i misunderstanding this Webhook?

ncdc: It was in an earlier version of this PR. Not sure why it got deleted?

tahsinrahman: as entire spec is immutable, instead of checking every field of spec, i did this
```
if !reflect.DeepEqual(newAWSMachineTemplate.Spec, oldAWSMachineTemplate.Spec) {
	return errors.New("awsMachineTemplateSpec is immutable")
}
```

detiber: This validation was being used by the AWSMachine controller, where the current webhook is targeting AWSMachineTemplate.

tahsinrahman: got it!
that means, we need to deploy another validating webhook for `AWSMachine` types, to move these logic, right?

detiber: correct, thanks :)

ncdc: @tahsinrahman are you going to add the AWSMachine validation webhook in this PR?

config/crd/kustomization.yaml
```diff
@@ -9,7 +7,7 @@ resources:
 - bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml
 # +kubebuilder:scaffold:crdkustomizeresource
 
-#patches:
+patchesStrategicMerge:
```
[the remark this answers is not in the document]
tahsinrahman: no, they're for enabling conversion webhooks

```
# [WEBHOOK] To enable webhook, uncomment all the sections with [WEBHOOK] prefix.
# patches here are for enabling the conversion webhook for each CRD
```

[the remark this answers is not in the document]
ncdc: Thanks. It's weird that conversions are enabled in config/crd, and mutating/validating are in config/default.

controllers/awsmachine_controller.go
tahsinrahman: Sure, I can add it here, or if it becomes too big for review, i can do it in separate pr, which one works best for you? :smile:

ncdc: Either way is fine. If you keep them separate, you can retitle this PR to "Add validating webhook for AWSMachineTemplate"?

config/webhook/manifests.yaml
```diff
@@ -0,0 +1,26 @@
...
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-infrastructure-cluster-x-k8s-io-v1alpha2-awsmachinetemplate
+  failurePolicy: Fail
+  name: vawsmachinetemplate.kb.io
```
vincepri: How was this name generated?

ncdc: kubebuilder does it this way

vincepri: Got it, was mostly curious as it reads funny

ncdc:
> Do we need to cut release-0.4 first?

No, we can branch from any commit at any time.

tahsinrahman:
Do we want to merge this now? It'll break clusterctl

vincepri:
Let's hold on to this until we have all the others open PR merged in 0.4 and we cut the release branch

ncdc:
We also don't have to rush this in. Let's think more about the "clusterctl on master" impact.

ncdc:
I'm thinking this will need to move into the v1alpha3 api package (#1202), right?

tahsinrahman:
I updated the pr with v1alpha3 package!

detiber:
/lgtm
I believe the hold can be lifted now that we have the release-0.4 branch as well. @ncdc?

ncdc:
Yes, master is open for v1alpha3.
/hold cancel

Issue #1057, opened 2019-08-22
add opportunity to modify listeners for ELB


dinomiteX:
/kind feature

**Describe the solution you'd like**
I would like to have the opportunity to add listeners to the ELB of the controlpanel. 
At the moment its only one listener with port 6443, and this is hardcoded at [https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/87331ba1900892bb04be5ef58ba993ee78ae8be6/pkg/cloud/services/elb/loadbalancer.go#L138](url)

and the reconciling only uses this default configuration to create ELBs [https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/87331ba1900892bb04be5ef58ba993ee78ae8be6/pkg/cloud/services/elb/loadbalancer.go#L39](url)

correct me if im wrong but it seems like there is no way to modify this, even if you set the ProviderSpecs correctly  
**Anything else you would like to add:**
For automation purpose its nice to have this feature, because some companies have restricted firewalls and proxies.

**Environment:**
- Cluster-api-provider-aws version: v0.3.7

detiber:
You are correct. This gets a bit more complicated in v1alpha2, though with the separate bootstrap provider. I'm wondering if we need to add an optional field to the `Cluster` object to allow for overriding the API port in a single location that can be referenced from the infrastructure provider cluster, infrastructure provider machine, and the bootstrapping provider?

@vincepri @ncdc thoughts?

vincepri:
Would it make sense to add the port field under `Cluster.Spec.ClusterNetwork`? I'd fine adding the field (without the related logic) for now and then address that after v1alpha2 is released.

detiber:
Seems as good a place as any to me until we have proper control plane management.

dinomiteX:
Sounds good for me, thanks

vincepri:
@dinomiteX #1126 adds support to customize the APIServer load balancer port as we discussed, is this sufficient to fix your issue?

dinomiteX:
@vincepri I will check it thank you

dinomiteX:
This looks pretty good, i will close this issue. :)

PR #1126, opened 2019-09-12
✨ Add ability to customize APIServer LB listener port

vincepri:

Related to #1057


vincepri:
@detiber @ncdc This should be good to go

ncdc:
#1057 sounds like it's asking for both not hard-coding 6443 and the ability to add additional listeners to the ELB. Do we need to get clarification if both things are needed, as this PR only addresses the hard-coding issue?

vincepri:
I dropped the `Fixes ...` so we can get at least this functionality in, I'll circle back with the person that opened the issue to see if this fits or we need more functionality as well

PR #1129, opened 2019-09-12
✨ Record events for credentials and permissions issues

### LINKED ISSUE #1051: Surface all 401/403 errors (unauthorized, invalid client id, etc) as events
/kind feature

If the user has any combination of invalid credentials, expired credentials, or credentials that don't have the appropriate IAM permissions, we should surface this information to the user so they are aware.

We could define a single event type such as `CredentialsError`, or we could have individual ones if that makes more sense. The event should target the most appropriate CAPI/CAPA object.

**Environment:**

ncdc:
Record events for credentials and permissions issues against the AWSCluster.

Example event output:
```
Warning  UnauthorizedOperation  3s (x12 over 21s)  aws-controller  Operation DescribeVpcs failed with a credentials or permission issue
```

Fixes #1051 

/kind feature


williamsandrew:
LGTM

PR #1136, opened 2019-09-13
🏃Fix log names are too verbose

### LINKED ISSUE #1121: Controller log names are too verbose
We have log "names" (`r.Log.WithName()`) that are too verbose. Here is an example for the AWSCluster controller, and we have a similar issue with AWSMachine.

```
controllers/AWSCluster/awscluster-controller/namespace=default/awsCluster=capi-quickstart/infrastructure.cluster.x-k8s.io/v1alpha2/cluster=capi-quickstart
```

If we break that apart:

1. `controllers/AWSCluster` comes from https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a745288a721d0509fe853ff271683f7973302914/main.go#L142
1. `awscluster-controller` is a repeat of the previous name, and it comes from https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a745288a721d0509fe853ff271683f7973302914/controllers/awscluster_controller.go#L57
1. `namespace=default` is ok (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a745288a721d0509fe853ff271683f7973302914/controllers/awscluster_controller.go#L58)
1. `awsCluster=capi-quickstart` is ok (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a745288a721d0509fe853ff271683f7973302914/controllers/awscluster_controller.go#L59)
1. `infrastructure.cluster.x-k8s.io/v1alpha2` we should either remove or include what this is (`apiVersion=infrastructure.cluster.x-k8s.io/v1alpha2`) (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a745288a721d0509fe853ff271683f7973302914/controllers/awscluster_controller.go#L71)
1. `cluster=capi-quickstart` is ok (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a745288a721d0509fe853ff271683f7973302914/controllers/awscluster_controller.go#L83)

One other thing that's a little confusing is that the "name separator" is `/`, and some of our names have slashes in them. So it makes it hard to know where the separation of each name segment is.

/kind cleanup
/help
/good-first-issue
/priority important-longterm
/milestone Next

tahsinrahman:

Fixes #1121

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 	"time"
 
 	"github.com/go-logr/logr"
@@ -36,10 +35,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 )
 
-const (
-	controllerName = "awscluster-controller"
-)
-
 // AWSClusterReconciler reconciles a AwsCluster object
 type AWSClusterReconciler struct {
 	client.Client
@@ -53,9 +48,7 @@ type AWSClusterReconciler struct {
 
 func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx := context.TODO()
-	log := r.Log.WithName(controllerName).
-		WithName(fmt.Sprintf("namespace=%s", req.Namespace)).
-		WithName(fmt.Sprintf("awsCluster=%s", req.Name))
+	log := r.Log.WithValues("namespace", req.Namespace, "awsCluster", req.Name)
 
 	// Fetch the AWSCluster instance
 	awsCluster := &infrav1.AWSCluster{}
@@ -67,8 +60,6 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		return reconcile.Result{}, err
 	}
 
-	log = log.WithName(awsCluster.APIVersion)
-
 	// Fetch the Cluster.
 	cluster, err := util.GetOwnerCluster(ctx, r.Client, awsCluster.ObjectMeta)
 	if err != nil {
@@ -79,7 +70,7 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		return reconcile.Result{}, nil
 	}
 
-	log = log.WithName(fmt.Sprintf("cluster=%s", cluster.Name))
+	log = log.WithValues("cluster", cluster.Name)
 
 	// Create the scope.
 	clusterScope, err := scope.NewClusterScope(scope.ClusterScopeParams{
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -58,10 +58,7 @@ type AWSMachineReconciler struct {
 
 func (r *AWSMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx := context.TODO()
-	logger := r.Log.
-		WithName(controllerName).
-		WithName(fmt.Sprintf("namespace=%s", req.Namespace)).
-		WithName(fmt.Sprintf("awsMachine=%s", req.Name))
+	logger := r.Log.WithValues("namespace", req.Namespace, "awsMachine", req.Name)
 
 	// Fetch the AWSMachine instance.
 	awsMachine := &infrav1.AWSMachine{}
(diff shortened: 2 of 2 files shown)


controllers/awscluster_controller.go
```diff
@@ -80,7 +76,7 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		return reconcile.Result{}, nil
 	}
 
-	log = log.WithName(fmt.Sprintf("cluster=%s", cluster.Name))
+	log = log.WithValues("cluster", fmt.Sprintf("%s/%s", cluster.Namespace, cluster.Name))
```
detiber: We can likely leave off namespace here, since the namespace will be the same as the related `AWSCluster`.

controllers/awsmachine_controller.go
```diff
@@ -85,7 +80,7 @@ func (r *AWSMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		return reconcile.Result{}, nil
 	}
 
-	logger = logger.WithName(fmt.Sprintf("machine=%s", machine.Name))
+	logger = logger.WithValues("machine", fmt.Sprintf("%s/%s", machine.Namespace, machine.Name))
```
detiber: From here and blow we can likely leave off the namespace, since they will be the same as the `AWSMachine` above.

detiber:
lgtm, but in need of a rebase.

controllers/awscluster_controller.go
```diff
@@ -80,7 +74,7 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		return reconcile.Result{}, nil
 	}
 
-	log = log.WithName(fmt.Sprintf("cluster=%s", cluster.Name))
+	log = log.WithValues("cluster", cluster.Name)
```
vincepri: Do we even need this given that we're working on the awsCluster field?

controllers/awsmachine_controller.go
```diff
@@ -85,7 +80,7 @@ func (r *AWSMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		return reconcile.Result{}, nil
 	}
 
-	logger = logger.WithName(fmt.Sprintf("machine=%s", machine.Name))
+	logger = logger.WithValues("machine", machine.Name)
```
vincepri: Same here, what if we remove this as well?

controllers/awscluster_controller.go
detiber: This gives context back to the related Cluster. I'm not sure we should remove it.

controllers/awsmachine_controller.go
detiber: I think we lose context here if we don't provide the associated Machine.

controllers/awscluster_controller.go
```diff
@@ -53,9 +48,7 @@ type AWSClusterReconciler struct {
 
 func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx := context.TODO()
-	log := r.Log.WithName(controllerName).
-		WithName(fmt.Sprintf("namespace=%s", req.Namespace)).
-		WithName(fmt.Sprintf("awsCluster=%s", req.Name))
+	log := r.Log.WithValues("namespace", req.Namespace).WithValues("awsCluster", req.Name)
```
ncdc:
```suggestion
	log := r.Log.WithValues("namespace", req.Namespace, "awsCluster", req.Name)
```

controllers/awscluster_controller.go
ncdc: +1 for keeping

controllers/awsmachine_controller.go
```diff
@@ -58,10 +58,7 @@ type AWSMachineReconciler struct {
 
 func (r *AWSMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reterr error) {
 	ctx := context.TODO()
-	logger := r.Log.
-		WithName(controllerName).
-		WithName(fmt.Sprintf("namespace=%s", req.Namespace)).
-		WithName(fmt.Sprintf("awsMachine=%s", req.Name))
+	logger := r.Log.WithValues("namespace", req.Namespace).WithValues("awsMachine", req.Name)
```
ncdc:
```suggestion
	logger := r.Log.WithValues("namespace", req.Namespace, "awsMachine", req.Name)
```

controllers/awsmachine_controller.go
ncdc: +1 for keeping

PR #1144, opened 2019-09-23
:bug: remove finalizer when machine is absent

sethp-nr:
Allows machines that were never created to be deleted.

When a machine is never created (in our case, because we were hitting a service quota), it can't be deleted & will block MachineDeployments from making progress.


detiber:
/lgtm

On a side note, I'm wondering if we should avoid the case where we are currently waiting on terminate to finish and instead should return early for later reconciliation?

rudoi:
@detiber I agree with that suggestion. I think it makes sense to either consistently reconcile or consistently wait for things.

ncdc:
Let's move this discussion to an issue. I can file but it'll be a few minutes (I am about to join a meeting).

ncdc:
Filed: #1145

PR #1147, opened 2019-09-24
🐛fix: use clusterScope's port in APIEndpoint

rudoi:

Currently the reconciler requeues when the apiserver DNS name is not set. If the DNS is set, the port of the first listener is grabbed. Turns out we were not actually grabbing the listeners...

So there's an `index out of range` panic :D

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -166,10 +166,11 @@ func reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error)
 	}
 
 	// Set APIEndpoints so the Cluster API Cluster Controller can pull them
+	// TODO: should we get the Port from the first listener on the ELB?
 	awsCluster.Status.APIEndpoints = []infrav1.APIEndpoint{
 		{
 			Host: awsCluster.Status.Network.APIServerELB.DNSName,
-			Port: int(awsCluster.Status.Network.APIServerELB.Listeners[0].Port),
+			Port: int(clusterScope.APIServerPort()),
 		},
 	}
 


pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -376,5 +377,14 @@ func fromSDKTypeToClassicELB(v *elb.LoadBalancerDescription, attrs *elb.LoadBala
 		res.Attributes.IdleTimeout = time.Duration(*attrs.ConnectionSettings.IdleTimeout) * time.Second
 	}
 
+	for _, listener := range v.ListenerDescriptions {
+		res.Listeners = append(res.Listeners, &infrav1.ClassicELBListener{
+			Protocol:         v1alpha2.ClassicELBProtocol(aws.StringValue(listener.Listener.Protocol)),
+			Port:             aws.Int64Value(listener.Listener.LoadBalancerPort),
+			InstanceProtocol: v1alpha2.ClassicELBProtocol(aws.StringValue(listener.Listener.InstanceProtocol)),
```
rudoi: We may want to move these types down into this package?

controllers/awscluster_controller.go
```diff
@@ -160,7 +160,7 @@ func reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error)
 		return reconcile.Result{}, errors.Wrapf(err, "failed to reconcile load balancers for AWSCluster %s/%s", awsCluster.Namespace, awsCluster.Name)
 	}
 
-	if awsCluster.Status.Network.APIServerELB.DNSName == "" {
+	if awsCluster.Status.Network.APIServerELB.DNSName == "" || len(awsCluster.Status.Network.APIServerELB.Listeners) < 1 {
```
detiber: Do we actually need to get the value from the Listener?

I'm wondering if we can just use the same value that `s.scope.APIServerPort()` is derived from when creating the Load Balancer?

vincepri: Not sure if we should assume that the APIServerPort is the LoadBalancerPort, given that users might want to change the port the ELB listens on in the future

detiber: I would expect that to be driven by some type of config that we could also reference. I wouldn't expect users to modify the Listener out of band and for us to need to introspect the value to know what it is.

vincepri: I'd prefer to use what's returned by the describe call. I don't want to give users the false hope that changing the APIServerPort on the Cluster object, changes also the ELB port

detiber: We also have to remember that adding additional calls counts against quota that we are already hitting in some cases. If we are reconciling the load balancer declaratively then we can assume that the value we are setting in the APIEndpoints will eventually be honored through reconciliation.

rudoi: Is there something we can pull into an issue here to discuss? What should we do in the interim?

Master cannot create clusters in current state.

vincepri: Don't we hardcode the port today? If so, we should just hardcode it here as well and add a TODO comment to fix it up when we have a way for users to specify it

rudoi: spec.ClusterNetwork.APIServerPort seems to be honored 

```
// APIServerPort returns the APIServerPort to use when creating the load balancer.
func (s *ClusterScope) APIServerPort() int64 {
	if s.Cluster.Spec.ClusterNetwork != nil && s.Cluster.Spec.ClusterNetwork.APIServerPort != nil {
		return int64(*s.Cluster.Spec.ClusterNetwork.APIServerPort)
	}
	return 6443
}
```

vincepri: I was referring to https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/elb/loadbalancer.go#L174

rudoi: The APIEndpoint port should reference the Load Balancer port, right, not the instance port? The instance port appears to be hardcoded but the listener port is set to whatever API server port is in the Spec.

Sorry I think I'm confused lol.

rudoi: I'll undo the change getting it from the describe output and use the scope's port for now, with a TODO.

detiber:
/hold
holding for vince to review.

ncdc:
LGTM, FWIW 😄

controllers/awscluster_controller.go
vincepri: I'd prefer to hardcode, like we do in the other place I linked above, otherwise if you change the APIServer port, the listener won't change but only the backend port will, which is going to result in a bug

vincepri:
As I mentioned above, we shouldn't be using the scope port because that's separate from the load balancer listener, which is hardcoded here https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/elb/loadbalancer.go#L174.

Using the port from the cluster scope (which is specified in Cluster.Spec.Network) can result in errors in case a user uses anything other than 6443. The Kubeconfig secret, for example, is generated using the port in the Cluster.Status.APIEndpoints[0].Port, which would be incorrect in this case.

PR #1148, opened 2019-09-25
[v1alpha1] :sparkles: Remove use of cloud-init module

detiber:

Removes the use of the custom kubeadm cloud init module

### DIFF
--- a/pkg/cloud/aws/services/userdata/controlplane_init.go
+++ b/pkg/cloud/aws/services/userdata/controlplane_init.go
@@ -31,9 +31,8 @@ const (
 {{.ClusterConfiguration | Indent 6}}
       ---
 {{.InitConfiguration | Indent 6}}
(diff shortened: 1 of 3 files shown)


detiber:
Tested against the existing CAPA v1.15.3 ubuntu image as well as with the changes in https://github.com/kubernetes-sigs/image-builder/pull/50 and this functions as expected

detiber:
@chuckha I validated against a v1alpha1 cluster created with a 3 node control plane where one Machine was using the existing public AMI, one Machine was using a CentOS 7.7 image with the default version of cloud-init, and one Machine was using an Ubuntu 18.04 image with the default version of cloud-init and everything worked as expected. I don't see any reason to do additional validation unless you specifically want to :)

chuckha:
sounds perfect to me, nice job

PR #1152, opened 2019-09-26
fix: use copy builtin correctly for security groups

### LINKED ISSUE #1151: AdditionalSecurityGroups are not applied
/kind bug

Have AWSMachine with additionalSecurityGroups specified. Were not applied.

Security groups applied.

/assign
/lifecycle active

**Environment:**

rudoi:

Turns out, the destination in the copy builtin is first!

Fixes #1151

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -622,7 +622,7 @@ func (s *Service) attachSecurityGroupsToNetworkInterface(groups []string, interf
 		return nil
 	}
 
-	s.scope.V(3).Info("Updating security groups", "groups", totalGroups)
+	s.scope.Info("Updating security groups", "groups", totalGroups)
 
 	input := &ec2.ModifyNetworkInterfaceAttributeInput{
 		NetworkInterfaceId: aws.String(interfaceID),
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/ec2/instances.go
```diff
@@ -622,6 +622,8 @@ func (s *Service) attachSecurityGroupsToNetworkInterface(groups []string, interf
 		return nil
 	}
 
+	s.scope.V(3).Info("updating security groups", "groups", totalGroups)
```
vincepri:
```suggestion
	s.scope.V(3).Info("Updating security groups", "groups", totalGroups)
```

ncdc: Is V3 appropriate or would it be worthwhile to log this all the time?

rudoi: there are 2 other relevant logging messages that wrap this, but not with this much detail. i'm open to either.

ncdc: I'd probably err on the side of always logging, and do a follow-up to increase the V level if we find it too verbose when it's always on

rudoi: i think always would be better

detiber:
Need to make sure we include the permissions updates in the release notes

ncdc:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1146 is listed as ✨ but we should change it to ⚠️, right?

ncdc:
LGTM. Can you squash this to 2 commits - 1 for the permissions update, 1 for the copy + logging fix?

PR #1153, opened 2019-09-27
🐛Fix a leak of AWS Tags from between Machine and Cluster resources

sethp-nr:

Fixes a "leak" where the load balancer security groups' tags were being applied to other security groups because maps are secret pointers.

There's a test for the security group issue specifically, but I think I fixed/prevented some other odd behaviors around merging a Machine's tags into the cluster-level Scope.

### DIFF
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -170,7 +170,7 @@ func (s *ClusterScope) AdditionalTags() infrav1.Tags {
 		s.AWSCluster.Spec.AdditionalTags = infrav1.Tags{}
 	}
 
-	return infrav1.Tags(s.AWSCluster.Spec.AdditionalTags).DeepCopy()
+	return s.AWSCluster.Spec.AdditionalTags.DeepCopy()
 }
 
 // APIServerPort returns the APIServerPort to use when creating the load balancer.
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -170,7 +170,7 @@ func (s *ClusterScope) AdditionalTags() infrav1.Tags {
 		s.AWSCluster.Spec.AdditionalTags = infrav1.Tags{}
 	}
 
-	return s.AWSCluster.Spec.AdditionalTags
+	return infrav1.Tags(s.AWSCluster.Spec.AdditionalTags).DeepCopy()
 }
 
 // APIServerPort returns the APIServerPort to use when creating the load balancer.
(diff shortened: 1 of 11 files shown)


pkg/cloud/services/ec2/securitygroups.go
```diff
@@ -62,9 +62,9 @@ func (s *Service) reconcileSecurityGroups() error {
 	// Declare all security group roles that the reconcile loop takes care of.
 	roles := []infrav1.SecurityGroupRole{
 		infrav1.SecurityGroupBastion,
+		infrav1.SecurityGroupLB,
```
sethp-nr: This change is to make the test actually catch the scope "leakage," since the LB is the only security group (or, really, the only place in CAPA) that we add a conditional tag.

pkg/cloud/scope/machine.go
```diff
@@ -180,11 +180,9 @@ func (m *MachineScope) Close() error {
 // AdditionalTags merges AdditionalTags from the scope's AWSCluster and AWSMachine. If the same key is present in both,
 // the value from AWSMachine takes precedence. The returned Tags will never be nil.
 func (m *MachineScope) AdditionalTags() infrav1.Tags {
-	var tags infrav1.Tags = m.AWSCluster.Spec.AdditionalTags
-	if tags == nil {
-		tags = infrav1.Tags{}
-	}
+	var tags infrav1.Tags = make(infrav1.Tags)
```
vincepri:
```suggestion
	tags := make(infrav1.Tags)
```

pkg/cloud/scope/machine.go
```diff
@@ -180,11 +180,9 @@ func (m *MachineScope) Close() error {
...
 
+	tags.Merge(m.AWSCluster.Spec.AdditionalTags)
 	tags.Merge(m.AWSMachine.Spec.AdditionalTags)
```
vincepri: The above lines have such a subtle difference is almost invisible 👻, can we add a comment above both?

pkg/cloud/scope/cluster.go
```diff
@@ -170,7 +170,7 @@ func (s *ClusterScope) AdditionalTags() infrav1.Tags {
 		s.AWSCluster.Spec.AdditionalTags = infrav1.Tags{}
 	}
 
-	return s.AWSCluster.Spec.AdditionalTags
+	return infrav1.Tags(s.AWSCluster.Spec.AdditionalTags).DeepCopy()
```
vincepri: Isn't AdditionalTags already of `Tags` type? If not, we should make it so, if it's not a breaking change

sethp-nr: Both the Machine and Cluster were like that, so I didn't touch it. I don't know if it's breaking or not – I'll give it a try and report back.

sethp-nr: Once I figured out what package I was in, everything seemed to work out fine.

PR #1155, opened 2019-09-29
:sparkles: Control plane load balancer scheme can be set in cluster network spec

### LINKED ISSUE #873: Possible to provision API with an internal LB ?
/kind feature

Is it possible to add configuration to the cluster network, which would permit switching from "internet-facing" scheme, to "internal" for the control plane API?  From my reading of the docs this would be a new feature request.

I went to add a field to my cluster.yaml during clusterctl create, but failed to grok that it's just a status field in the spec, not anything you can actually change or set I think.

```
status:
  apiEndpoints:
  - host: capi-test2-apiserver-xxx.region-1.elb.amazonaws.com
    port: 443
  providerStatus:
    bastion:
      id: ""
    metadata:
      creationTimestamp: null
    network:
      apiServerElb:
        attributes:
          idleTimeout: 600000000000
        dnsName: capi-test2-apiserver-xxx.region-1.elb.amazonaws.com
        name: capi-test2-apiserver
        scheme: internet-facing
        securityGroupIds:
        - sg-12345
...
```

...so obviously I can't actually set that from the template (or in kubectl edit cluster) what chaos it would be if the cluster API and kube API moved?  There are SSL certs in the cluster providence that depend on the hostname of that ELB to remain stable, I guess that's out.

So I'm assuming this would be a strictly cluster-up decision, like most of cluster-api, and not some kind of public-to-private cluster lifecycle event or vice-versa.  I don't think my organization will ever want to make Kube API public.  I tried taking down the LB manually, but that caused all kinds of havoc with my prometheus metrics and ultimately may have been the cause of death for that cluster.

With a control plane LB in the "internal" scheme mode, we can control access to the API through VPC/VPN and enforce a second factor, and maybe in this way we can safely avoid implementing any more complicated authentication schemes for a while.  Presumably I could restrict access to the load balancer with some security group changes?

I am not expert or maybe not permissioned highly enough to figure that out, after several attempts, it seems like ~perhaps some other networking equipment attached to my VPC is preventing it because of the non-standard port 6443.  I know we have a Cisco that prevents traffic on non-standard ports from leaving the VPC, and the internet-facing LB is outside of the VPC, so~ I think I really need to use an internal load balancer configuration in order to solve this in my environment.

edit: I ~sort of solved the API security, with security groups.~ Editing the security groups in-place was misguided here ~I needed to be sure that the VPC's NAT GW was added to the security group (not my VPN's NAT GW, which is separate), as well as the ELB itself.~

There is a `[CLUSTERNAME]-lb` security group created by the default workflow, which has no rules or attachments.  I added it to my ELB (guessing that my scenario or something like it is probably what this was meant to be used for) and then put a 6443/TCP rule for it in the security group with another one for my NAT GW, and subnet with the cluster nodes in it, and the bastion host for good measure too.

The `schema: internet-facing` ELB is now usable in "private mode", although my traffic must leave the VPC which is less than optimal; I can at least work with this, but it is not a garden path either.  I understand that customizing security groups is not yet in scope for CAPA, maybe internal LB is a good place to start.

Yikes, I understand now that I should not edit the security groups in-place, or I will get a failure to reconcile.  Perhaps when I try this again next time, I will add some more security groups instead.

andrewmyhre:
Currently the control plane load balancer is always created with the `Internet-facing` scheme. Private VPCs need to be able to set the scheme to `internal`. This PR adds an optional property `ControlPlaneLoadBalancerScheme` to the `NetworkSpec` type. The value defaults to `Internet-facing` for backwards compatibility but can be set to `internal` to support private VPCs.

Fixes #873

### DIFF
--- a/api/v1alpha2/types.go
+++ b/api/v1alpha2/types.go
@@ -169,6 +169,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// ControlPlaneLoadBalancerScheme (defaults to Internet-facing)
+	// +optional
+	ControlPlaneLoadBalancerScheme ClassicELBScheme `json:"controlPlaneLoadBalancerScheme,omitempty""`
 }
 
 // APIEndpoint represents a reachable Kubernetes API endpoint.
--- a/api/v1alpha2/types.go
+++ b/api/v1alpha2/types.go
@@ -172,7 +172,7 @@ type NetworkSpec struct {
 
 	// ControlPlaneLoadBalancerScheme (defaults to Internet-facing)
 	// +optional
-	ControlPlaneLoadBalancerScheme ClassicELBScheme `json:"controlPlaneLoadBalancerScheme,omitempty""`
+	ControlPlaneLoadBalancerScheme *ClassicELBScheme `json:"controlPlaneLoadBalancerScheme,omitempty""`
 }
 
 // APIEndpoint represents a reachable Kubernetes API endpoint.
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -146,6 +146,14 @@ func (s *ClusterScope) Region() string {
 	return s.AWSCluster.Spec.Region
 }
 
+// ControlPlaneLoadBalancerScheme returns the Classic ELB scheme (public or internal facing)
+func (s *ClusterScope) ControlPlaneLoadBalancerScheme() infrav1.ClassicELBScheme {
+	if s.AWSCluster.Spec.NetworkSpec.ControlPlaneLoadBalancerScheme == "" {
+		return infrav1.ClassicELBSchemeInternetFacing
+	}
+	return s.AWSCluster.Spec.NetworkSpec.ControlPlaneLoadBalancerScheme
+}
+
 // ControlPlaneConfigMapName returns the name of the ConfigMap used to
 // coordinate the bootstrapping of control plane nodes.
 func (s *ClusterScope) ControlPlaneConfigMapName() string {
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -165,7 +165,7 @@ func GenerateELBName(clusterName string, elbName string) string {
 func (s *Service) getAPIServerClassicELBSpec() *infrav1.ClassicELB {
 	res := &infrav1.ClassicELB{
 		Name:   GenerateELBName(s.scope.Name(), infrav1.APIServerRoleTagValue),
-		Scheme: infrav1.ClassicELBSchemeInternetFacing,
+		Scheme: s.scope.ControlPlaneLoadBalancerScheme(),
 		Listeners: []*infrav1.ClassicELBListener{
 			{
 				Protocol:         infrav1.ClassicELBProtocolTCP,


andrewmyhre:
updated my github email

randomvariable:
/ok-to-test

This looks reasonable to me, but need to defer to someone else on the API change front (it's an addition, so should be fine).

api/v1alpha2/types.go
```diff
@@ -169,6 +169,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// ControlPlaneLoadBalancerScheme (defaults to Internet-facing)
+	// +optional
+	ControlPlaneLoadBalancerScheme ClassicELBScheme `json:"controlPlaneLoadBalancerScheme,omitempty""`
```
vincepri: Let’s make this a pointer given that’s optional

pkg/cloud/scope/cluster.go
```diff
@@ -146,6 +146,14 @@ func (s *ClusterScope) Region() string {
 	return s.AWSCluster.Spec.Region
 }
 
+// ControlPlaneLoadBalancerScheme returns the Classic ELB scheme (public or internal facing)
+func (s *ClusterScope) ControlPlaneLoadBalancerScheme() infrav1.ClassicELBScheme {
+	if s.AWSCluster.Spec.NetworkSpec.ControlPlaneLoadBalancerScheme == "" {
+		return infrav1.ClassicELBSchemeInternetFacing
+	}
+	return s.AWSCluster.Spec.NetworkSpec.ControlPlaneLoadBalancerScheme
```
dlipovetsky: There are downsides to applying defaults here: for example, the default values are not discoverable. With v1alpha2,  `AWSCluster` is a CRD, and at some point we can use [CRD defaulting](https://kubernetes.io/docs/tasks/access-kubernetes-api/custom-resources/custom-resource-definitions/#defaulting). However, applying the default here has precedent (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1126).

vincepri: There should be a tracking issue upstream to raise the minimum Kubernetes version to 1.16, which brings support for defaulting. For now, making it explicit in a comment should be ok

andrewmyhre: Not sure what the call-out is here.. I should add `(defaults to 'public')` in a comment?

api/v1alpha2/types.go
andrewmyhre: I've made this a pointer in my fork, but I'm having trouble figuring out how to make the change reflect here

pkg/cloud/scope/cluster.go
randomvariable: Yup, which you've done. Though, rereading this, we should probably start following Kubernetes convention and CamelCase these and go with `InternetFacing` and `Internal` as the two options.

dlipovetsky: The upstream issue for 1.16 support: https://github.com/kubernetes-sigs/cluster-api/issues/1442

api/v1alpha2/types.go
```diff
@@ -169,6 +169,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// ControlPlaneLoadBalancerScheme (defaults to Internet-facing)
+	// +optional
+	ControlPlaneLoadBalancerScheme *ClassicELBScheme `json:"controlPlaneLoadBalancerScheme,omitempty""`
```
vincepri: It'd be great if we can have a struct encapsulating all the knobs we want to expose specifically for the API Server load balancer and have this optional field (and other later probably) in there

andrewmyhre: Thanks! I'd be happy for this PR to be closed and I'll provide a separate PR with a proposed implementation.

randomvariable: No probs.

/close

PR #1163, opened 2019-10-02
✨Increase AWSCluster and AWSMachine default concurrency

vincepri:

As discussed on [slack](https://kubernetes.slack.com/archives/CD6U2V71N/p1570024727095600), and seeing lots of user confusion, this PR increased the default concurrency for the reconcilers.

### DIFF
--- a/main.go
+++ b/main.go
@@ -92,13 +92,13 @@ func main() {
 
 	flag.IntVar(&awsClusterConcurrency,
 		"awscluster-concurrency",
-		1,
+		2,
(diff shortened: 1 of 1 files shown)


detiber:
I'm a bit hesitant to go this high without doing some baseline testing to see how it affects AWS quota..

vincepri:
@rudoi has had experiences with larger clusters in the wild and it seems to be working for them, I've also tested it and it doesn't seem to hit quota

PR #1176, opened 2019-10-06
✨ Add webhook port flag

### LINKED ISSUE #1120: Make the webhook server port configurable via flag
/kind feature

It was [pointed out](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1116#discussion_r322925177) that in some cases the [default value (443)](https://github.com/kubernetes-sigs/controller-runtime/blob/699340657992a5ffe4d400a9f68330420fb4a689/pkg/webhook/server.go#L42) for the [webhook server port](https://godoc.org/sigs.k8s.io/controller-runtime/pkg/manager#Options.Port) can be problematic, so we should should make this [configurable via a flag](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a745288a721d0509fe853ff271683f7973302914/main.go#L125-L131).

**Environment:**

luigi-riefolo:
Add webhook port flag (default 8443)

Fixes #1120

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,6 +60,7 @@ func main() {
 		awsClusterConcurrency int
 		awsMachineConcurrency int
 		syncPeriod            time.Duration
+		webhookPort           int
 	)
 
 	flag.StringVar(
@@ -108,6 +109,12 @@ func main() {
 		"The minimum interval at which watched resources are reconciled (e.g. 15m)",
 	)
 
+	flag.IntVar(&webhookPort,
+		"webhook-port",
+		9443,
+		"Webhook server port",
+	)
+
 	flag.Parse()
 
 	if watchNamespace != "" {
@@ -135,6 +142,7 @@ func main() {
 		SyncPeriod:         &syncPeriod,
 		Namespace:          watchNamespace,
 		EventBroadcaster:   broadcaster,
+		Port:               webhookPort,
 	})
 	if err != nil {
 		setupLog.Error(err, "unable to start manager")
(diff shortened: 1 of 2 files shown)


detiber:
This PR should probably also update the `manager_webhook_patch.yaml` here as well: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/config/default/manager_webhook_patch.yaml#L12

luigi-riefolo:
@detiber done

ncdc:
@luigi-riefolo could you please squash this down to 1 commit? Thanks!

luigi-riefolo:
@ncdc done

main.go
```diff
@@ -108,6 +109,12 @@ func main() {
 		"The minimum interval at which watched resources are reconciled (e.g. 15m)",
 	)
 
+	flag.IntVar(&webhookPort,
+		"webhook-port",
+		8443,
+		"Webhook server port (default 8443)",
```
ncdc:
```suggestion
		"Webhook server port",
```

ncdc: The default value is handled by cobra. This is what it looks like now:
```
  -webhook-port int
        Webhook server port (default 8443) (default 8443)
```

tahsinrahman:
`kubebuilder 2.0.1` switched to default port `9443` for webhook server. should we also use `9443`?
https://github.com/kubernetes-sigs/kubebuilder/pull/983

detiber:
> kubebuilder 2.0.1 switched to default port 9443 for webhook server. should we also use 9443?
kubernetes-sigs/kubebuilder#983

Thanks for catching that @tahsinrahman, I think we should likely try to stick with kubebuilder defaults here.

luigi-riefolo:
@tahsinrahman @detiber set to 9443

ncdc:
/lgtm
@detiber over to you

PR #1178, opened 2019-10-07
🐛 Support running alongside other Cluster API pods in the same namespace with leader election enabled

### LINKED ISSUE #1172: Allow setting LeaderID in controller manager options
Running CAPA in the same namespace with CAPI or CABPK results in leader election errors. More details https://github.com/kubernetes-sigs/cluster-api-bootstrap-provider-kubeadm/issues/271

/kind bug
/milestone v0.4.x
/priority important-soon
/assign
/lifecycle active
/cc @noamran

noamran:
Currently `LeaderElectionID` is set to the default `controller-leader-election-helper`. This causes an issue when several Cluster API pods are deployed in the same namespace. Adding a unique ID per service should resolve this.

Fixes #1172

### DIFF
--- a/main.go
+++ b/main.go
@@ -132,6 +132,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+		LeaderElectionID:   "controller-leader-election-capa",
 		SyncPeriod:         &syncPeriod,
 		Namespace:          watchNamespace,
 		EventBroadcaster:   broadcaster,


main.go
```diff
@@ -132,6 +132,7 @@ func main() {
 		Scheme:             scheme,
 		MetricsBindAddress: metricsAddr,
 		LeaderElection:     enableLeaderElection,
+		LeaderElectionID: "controller-leader-election-capa",
```
detiber: spacing looks a bit off, not sure this will pass the linter.

PR #1179, opened 2019-10-07
:sparkles: add optional ControlPlaneSpec to AWSClusterSpec type

### LINKED ISSUE #873: Possible to provision API with an internal LB ?

andrewmyhre:
- Allows specification of the control plane load balancer scheme (internet-facing or private). Defaults to internet-facing
- Implements a struct for control plane configuration which can be extended

Fixes #873


api/v1alpha2/awscluster_types.go
```diff
@@ -41,6 +41,16 @@ type AWSClusterSpec struct {
 	// ones added by default.
 	// +optional
 	AdditionalTags Tags `json:"additionalTags,omitempty"`
+
+	// ControlPlaneSpec is optional configuration for customizing control plane behavior
+	// +optional
+	ControlPlaneSpec *AWSControlPlaneSpec `json:controlPlaneSpec,omitempty"`
```
vincepri:
```suggestion
	ControlPlaneLoadBalancer *AWSLoadBalancerSpec `json:controlPlaneLoadBalancer,omitempty"`
```

api/v1alpha2/awscluster_types.go
```diff
@@ -41,6 +41,16 @@ type AWSClusterSpec struct {
...
+}
+
+type AWSControlPlaneSpec struct {
+	// LoadBalancerScheme sets the scheme of the control plane load balancer  (defaults to Internet-facing)
+	// +optional
+	ControlPlaneLoadBalancerScheme *ClassicELBScheme `json:"controlPlaneLoadBalancerScheme,omitempty""`
```
vincepri:
```suggestion
	Scheme *ClassicELBScheme `json:"scheme,omitempty""`
```

api/v1alpha2/awscluster_types.go
```diff
@@ -41,6 +41,16 @@ type AWSClusterSpec struct {
 	// ones added by default.
 	// +optional
 	AdditionalTags Tags `json:"additionalTags,omitempty"`
+
+	// ControlPlaneSpec is optional configuration for customizing control plane behavior
+	// +optional
+	ControlPlaneSpec *AWSControlPlaneSpec `json:controlPlaneSpec,omitempty"`
+}
+
+type AWSControlPlaneSpec struct {
```
vincepri:
```suggestion
type AWSLoadBalancerSpec struct {
```

api/v1alpha2/awscluster_types.go
andrewmyhre: Addressed

api/v1alpha2/awscluster_types.go
andrewmyhre: Addressed

api/v1alpha2/awscluster_types.go
andrewmyhre: Addressed

api/v1alpha2/awscluster_types.go
```diff
@@ -41,6 +41,16 @@ type AWSClusterSpec struct {
 	// ones added by default.
 	// +optional
 	AdditionalTags Tags `json:"additionalTags,omitempty"`
+
+	// ControlPlaneSpec is optional configuration for customizing control plane behavior
+	// +optional
+	ControlPlaneLoadBalancer *AWSLoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
+}
+
+type AWSLoadBalancerSpec struct {
```
vincepri: Add a godoc up here :)

andrewmyhre: Added!

api/v1alpha2/awscluster_types.go
```diff
@@ -41,6 +41,17 @@ type AWSClusterSpec struct {
 	// ones added by default.
 	// +optional
 	AdditionalTags Tags `json:"additionalTags,omitempty"`
+
+	// ControlPlaneSpec is optional configuration for customizing control plane behavior
```
ncdc: Is this supposed to be ControlPlaneSpec with a nested LoadBalancerSpec, or just ControlPlaneLoadBalancer?

pkg/cloud/scope/cluster.go
```diff
@@ -146,6 +146,19 @@ func (s *ClusterScope) Region() string {
 	return s.AWSCluster.Spec.Region
 }
 
+// ControlPlaneLoadBalancer returns the ControlPlaneSpec
```
ncdc: ControlPlaneSpec -> something else?

PR #1184, opened 2019-10-08
:running: Fix hack/ensure-kustomize.sh

detiber:

Fixes the script used to download kustomize in CI, which has resulted in broken e2e test signal.


ncdc:
/lgtm

Note: starting with v3.2.1, it looks like the `v` is back in the filename

detiber:
> Note: starting with v3.2.1, it looks like the v is back in the filename
:joy: of course it is

PR #1186, opened 2019-10-08
:running: Use e2e test helpers and cleanup e2e tests

detiber:
Uses the e2e test helpers from cluster-api and cleans up some of the e2e test.


api/conversion/cluster.go
```diff
@@ -17,18 +17,32 @@ limitations under the License.
...
 
 import (
-	"github.com/pkg/errors"
+	"fmt"
 
+	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	cabpkv1a2 "sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/api/v1alpha2"
-	"sigs.k8s.io/cluster-api-bootstrap-provider-kubeadm/certs"
```
detiber: :( all of the additions in this file are related to losing access to the certs helpers in newer versions of CABPK

PR #1188, opened 2019-10-09
🏃 Add a development dockerfile

chuckha:
The third_party code is necessary for tilt's live_update feature
on a containerd-based kubernetes system

This docker file allows us to iterate quickly with tilt's live_update feature

/cc @noamran @ncdc

### DIFF
--- a/dev.dockerfile
+++ b/dev.dockerfile
@@ -1,0 +1,43 @@
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


dev.dockerfile
```diff
@@ -0,0 +1,44 @@
...
+COPY api/ api/
+COPY controllers/ controllers/
+COPY pkg/ pkg/
+
+# Allow containerd to restart pods by calling /restart.sh (mostly for tilt + fast dev cycles)
+# TODO: Remove this on prod and use a multi-stage build
+COPY third_party/forked/rerun-process-wrapper/start.sh .
+COPY third_party/forked/rerun-process-wrapper/restart.sh .
+
+RUN go build -v -o manager main.go
```
ncdc:
```suggestion
RUN go build -v -o /manager main.go
```

Can you do this & avoid the `mv` line below?

chuckha: oh, interesting, i've always done relative paths in my go build line. i dunno! if it work i'm happy to adopt the suggestion

Issue #967, opened 2019-08-07
ha cluster:  subnet reconcile miss to check and create route table


Sn0rt:
/kind bug

I want to create ha cluster with three node and the three located in different AZ.

I try the POC as follow step:

1: set the ENV
2: generate the yaml of cluster
3: apply the Cluster YAML to managerCluster
4: create two subnetes and get the subnet id // this is by manually
5: modify the control-plane machine to set the subnet ID
6: apply the the control-plane MachineList 

but, now the route table NOT contain the NAT route with the two subnet which create by hand. 

the subnet list as follow 

```json
{
    "Subnets": [
        {
            "AvailabilityZone": "us-east-2b",
            "AvailabilityZoneId": "use2-az2",
            "AvailableIpAddressCount": 250,
            "CidrBlock": "10.0.2.0/24",
            "DefaultForAz": false,
            "MapPublicIpOnLaunch": false,
            "State": "available",
            "SubnetId": "subnet-0c870838a83e8819e",
            "VpcId": "vpc-080002927c8fd2b02",
            "OwnerId": "179516646050",
            "AssignIpv6AddressOnCreation": false,
            "Ipv6CidrBlockAssociationSet": [],
            "Tags": [
                {
                    "Key": "Name",
                    "Value": "ha01-subnet-private"
                },
                {
                    "Key": "sigs.k8s.io/cluster-api-provider-aws/cluster/ha01",
                    "Value": "owned"
                },
                {
                    "Key": "sigs.k8s.io/cluster-api-provider-aws/role",
                    "Value": "private"
                }
            ],
            "SubnetArn": "arn:aws:ec2:us-east-2:179516646050:subnet/subnet-0c870838a83e8819e"
        },
        {
            "AvailabilityZone": "us-east-2c",
            "AvailabilityZoneId": "use2-az3",
            "AvailableIpAddressCount": 250,
            "CidrBlock": "10.0.3.0/24",
            "DefaultForAz": false,
            "MapPublicIpOnLaunch": false,
            "State": "available",
            "SubnetId": "subnet-0b393a067c067a4e7",
            "VpcId": "vpc-080002927c8fd2b02",
            "OwnerId": "179516646050",
            "AssignIpv6AddressOnCreation": false,
            "Ipv6CidrBlockAssociationSet": [],
            "Tags": [
                {
                    "Key": "sigs.k8s.io/cluster-api-provider-aws/role",
                    "Value": "private"
                },
                {
[...]

Issue #930, opened 2019-07-26
Add documentation for how to create a control plane that spans multiple AZs


Sn0rt:
I generate the file which is controlplane-machine-ha, and find that the master machineList file no specified detail of HA.

in fact, the user want to create a cluster is high available, that means is the cluster is HA with whole region, whole VPC. not create three master with same subnet.

/kind feature

vincepri:
To accomplish this today, you'll need to have 3 different subnets in a VPC and specify the `AvailabilityZone` field manually for each Machine.

sethp-nr:
FWIW, what @vincepri suggested is (almost) what we do – we specify the subnets directly – and it works really well.

We've found that the examples in this repo are a helpful starting point, but we pretty quickly ran into the need to template out our own configs that matched our needs.

detiber:
It is also possible to define multiple subnets when creating a Cluster with a "managed VPC" by setting Cluster.Spec.ProviderSpec.NetworkConfig.Subnets[*].AvailabilityZone for each Availability zone you wanted created in the managed VPC. When paired with setting the `AvailabilityZone` on the Machine, this allows you to pre-create Cluster and Machine resources without needing to pre-define the VPC or know the Subnets in advance before creating the Machines.

Sn0rt:
I successfully created a cluster of ha, but in a semi-manual manner.
I create a three node HA-Cluster, but the subnet and route table set is manually.

Sn0rt:
> It is also possible to define multiple subnets when creating a Cluster with a "managed VPC" by setting Cluster.Spec.ProviderSpec.NetworkConfig.Subnets[*].AvailabilityZone for each Availability zone you wanted created in the managed VPC. When paired with setting the `AvailabilityZone` on the Machine, this allows you to pre-create Cluster and Machine resources without needing to pre-define the VPC or know the Subnets in advance before creating the Machines.

is it means the HA info is cluster level not control-panel machine level ?

detiber:
It is both, since the shared infrastructure is created at the "Cluster level" and each control plane Machine needs to specify which subnet/AZ it should belong in.

chuckha:
This is something that was brought up in the meeting: https://blog.scottlowe.org/2019/09/05/highly-available-kubernetes-clusters-on-aws-with-cluster-api/

We could adopt some of this for the v0.4 milestone and rework them for the v0.5 milestone

detiber:
[...]

PR #1193, opened 2019-10-10
:warning: Use multi-AZ as default example configuration 

Sn0rt:
In most scenarios with availability requirements, customers will create clusters that fully support full-region fault tolerance.

TODO: Multiple private subnets share one natgw.

#967 

Signed-off-by: guohaowang <wangguohao.2009@gmail.com>

### DIFF
--- a/examples/cluster/cluster.yaml
+++ b/examples/cluster/cluster.yaml
@@ -6,7 +6,11 @@ metadata:
 spec:
   clusterNetwork:
     pods:
-      cidrBlocks: ["192.168.0.0/16"]
+      cidrBlocks:
+        - "192.168.0.0/16"
+    services:
+      cidrBlocks:
+        - "192.168.1.0/24"
   infrastructureRef:
     apiVersion: infrastructure.cluster.x-k8s.io/v1alpha2
     kind: AWSCluster
@@ -19,3 +23,23 @@ metadata:
 spec:
   region: ${AWS_REGION}
   sshKeyName: ${SSH_KEY_NAME}
+  networkSpec:
+    subnets:
+    - availabilityZone: ${AWS_REGION_AZA}
+      cidrBlock: "10.0.100.0/24"
+      isPublic: true
+    - availabilityZone: ${AWS_REGION_AZA}
+      cidrBlock: "10.0.101.0/24"
+      isPublic: false
+    - availabilityZone: ${AWS_REGION_AZB}
+      cidrBlock: "10.0.102.0/24"
+      isPublic: true
+    - availabilityZone: ${AWS_REGION_AZB}
+      cidrBlock: "10.0.103.0/24"
+      isPublic: false
+    - availabilityZone: ${AWS_REGION_AZC}
+      cidrBlock: "10.0.104.0/24"
+      isPublic: true
+    - availabilityZone: ${AWS_REGION_AZC}
+      cidrBlock: "10.0.105.0/24"
+      isPublic: false
--- a/examples/controlplane/controlplane.yaml
+++ b/examples/controlplane/controlplane.yaml
@@ -25,6 +25,7 @@ spec:
   instanceType: ${CONTROL_PLANE_MACHINE_TYPE}
   iamInstanceProfile: "control-plane.cluster-api-provider-aws.sigs.k8s.io"
   sshKeyName: "${SSH_KEY_NAME}"
+  availabilityZone: ${AWS_REGION_AZA}
 ---
 apiVersion: bootstrap.cluster.x-k8s.io/v1alpha2
 kind: KubeadmConfig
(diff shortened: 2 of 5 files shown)


examples/cluster/cluster.yaml
```diff
@@ -6,7 +6,11 @@ metadata:
 spec:
   clusterNetwork:
     pods:
-      cidrBlocks: ["192.168.0.0/16"]
+      cidrBlocks:
+        - "10.0.0.0/8"
+    services:
+      cidrBlocks:
+        - "192.168.0.0/16"
```
vincepri: Are these changes necessary and related to this PR?

examples/cluster/cluster.yaml
```diff
@@ -19,3 +23,23 @@ metadata:
 spec:
   region: ${AWS_REGION}
   sshKeyName: ${SSH_KEY_NAME}
+  networkSpec:
+    subnets:
+    - availabilityZone: ${AWS_REGION}a
+      cidrBlock: "10.0.100.0/24"
```
vincepri: Given that we use `10.0.0.0/16` by default, can we use the first subnets in the range http://www.davidc.net/sites/default/subnets/subnets.html?network=10.0.0.0&mask=16&division=23.ff4600 ?

vincepri:
/ok-to-test

Retitled to show that this is actually a breaking change in the manifest generation

examples/cluster/cluster.yaml
detiber: We really shouldn't overlap this with the pod network.

examples/controlplane/controlplane.yaml
```diff
@@ -25,6 +25,7 @@ spec:
   instanceType: ${CONTROL_PLANE_MACHINE_TYPE}
   iamInstanceProfile: "control-plane.cluster-api-provider-aws.sigs.k8s.io"
   sshKeyName: "${SSH_KEY_NAME}"
+  availabilityZone: ${AWS_REGION}a
```
detiber: Not every region has a `a`, `b`, and `c` AZ (nor do they necessarily align between different accounts). Hardcoding these here may lead to non-functioning manifests for some users.

examples/cluster/cluster.yaml
detiber: This will break the quick start, which is based on calico today.

ncdc:
Is this related to #930?

examples/cluster/cluster.yaml
Sn0rt: > This will break the quick start, which is based on calico today.

oho, this is my mistake, this setting just for amazon-vpc-cni. if user want to use calico , the CIDR of pods are different between vpc's.

examples/cluster/cluster.yaml
Sn0rt: > See above, I think the changes to the cluster configuration aren't necessary

yep, in fact, I hope that the generate script can generate the ha cluster yaml with amazon-vpc-cni or with  calico

examples/controlplane/controlplane.yaml
Sn0rt: > Not every region has a `a`, `b`, and `c` AZ (nor do they necessarily align between different accounts). Hardcoding these here may lead to non-functioning manifests for some users.

right, I will improve this script if this region has the a, b, c AZ , the script will generate the cluster yaml as before, if not, the script will set three masters with one az. what did you think?

Sn0rt: the az info as follows

eu-north-1: eu-north-1a eu-north-1b eu-north-1c
ap-south-1: ap-south-1a ap-south-1b ap-south-1c
eu-west-3: eu-west-3a eu-west-3b eu-west-3c
eu-west-2: eu-west-2a eu-west-2b eu-west-2c
eu-west-1: eu-west-1a eu-west-1b eu-west-1c
ap-northeast-2: ap-northeast-2a ap-northeast-2b ap-northeast-2c
ap-northeast-1: ap-northeast-1a ap-northeast-1c ap-northeast-1d
sa-east-1: sa-east-1a sa-east-1b sa-east-1c
ca-central-1: ca-central-1a ca-central-1b
ap-east-1: ap-east-1a ap-east-1b ap-east-1c
ap-southeast-1: ap-southeast-1a ap-southeast-1b ap-southeast-1c
ap-southeast-2: ap-southeast-2a ap-southeast-2b ap-southeast-2c
eu-central-1: eu-central-1a eu-central-1b eu-central-1c
us-east-1: us-east-1a us-east-1b us-east-1c us-east-1d us-east-1e us-east-1f
us-east-2: us-east-2a us-east-2b us-east-2c
us-west-1: us-west-1b us-west-1c
us-west-2: us-west-2a us-west-2b us-west-2c us-west-2d

ncdc:
Thanks for putting this together. We want to reevaluate how we support different flavors of cluster configurations after we have clusterctl v2 in place. For now, we're going to close this.

/close

PR #1194, opened 2019-10-10
🏃Use released manifests for CAPI and CABPK in examples

vincepri:

Fixes #

### DIFF
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -100,12 +100,24 @@ kustomize build "${SOURCE_DIR}/machinedeployment" | envsubst >> "${MACHINEDEPLOY
 echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
 
 # Generate Cluster API provider components file.
-kustomize build "github.com/kubernetes-sigs/cluster-api//config/default/?ref=master" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
-echo "Generated ${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
+CAPI_BRANCH=${CAPI_BRANCH:-"stable"}
+if [[ ${CAPI_BRANCH} == "stable" ]]; then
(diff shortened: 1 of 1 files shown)


examples/generate.sh
```diff
@@ -100,12 +100,24 @@ kustomize build "${SOURCE_DIR}/machinedeployment" | envsubst >> "${MACHINEDEPLOY
 echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
 
 # Generate Cluster API provider components file.
-kustomize build "github.com/kubernetes-sigs/cluster-api//config/default/?ref=master" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
-echo "Generated ${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
+CAPI_BRANCH=${CAPI_BRANCH:-"v0.2.4"}
```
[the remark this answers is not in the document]
vincepri: You're right, updated to default to stable

examples/generate.sh
```diff
@@ -100,12 +100,24 @@ kustomize build "${SOURCE_DIR}/machinedeployment" | envsubst >> "${MACHINEDEPLOY
 echo "Generated ${MACHINEDEPLOYMENT_GENERATED_FILE}"
 
 # Generate Cluster API provider components file.
-kustomize build "github.com/kubernetes-sigs/cluster-api//config/default/?ref=master" > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
-echo "Generated ${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
+CAPI_BRANCH=${CAPI_BRANCH:-"stable"}
+if [[ ${CAPI_BRANCH} == "stable" ]]; then
+  curl -L https://github.com/kubernetes-sigs/cluster-api/releases/download/v0.2.4/cluster-api-components.yaml > "${COMPONENTS_CLUSTER_API_GENERATED_FILE}"
```
ncdc: It sure would be nice to have a way to avoid the need to hard code the version in here 🙁

vincepri: Yeah agreed, we can do some magic at some point to grab the latest go module version. That said, given that these examples are only used locally, it should be ok?

PR #1196, opened 2019-10-10
🐛 Add omitempty to RouteTableID

### LINKED ISSUE #1195: Unable to use a go client to create an AWSCluster with subnets
/kind bug

1. Create an `AWSCluster` in go and make sure `Spec.NetworkSpec.Subnets` has at least one item.
1. Try to use a client-go/controller-runtime client to create the `AWSCluster`
1. Get the following error:

```
... spec.networkSpec.subnets.routeTableId in body must be of type string: \"null\"
```

No error

This is because `RouteTableID` is a pointer, but its json tag does not have `omitempty`

**Environment:**

- Cluster-api-provider-aws version: v0.4.2

ncdc:
Add omitempty to RouteTableID. Without this, if you leave the field nil
and try to use a go client (client-go / controller-runtime) to create an
AWSCluster, it will fail with the error

spec.networkSpec.subnets.routeTableId in body must be of type string:
\"null\"

Fixes #1195

### DIFF
--- a/api/v1alpha2/types.go
+++ b/api/v1alpha2/types.go
@@ -224,7 +224,7 @@ type SubnetSpec struct {
 
 	// RouteTableID is the routing table id associated with the subnet.
 	// +optional
-	RouteTableID *string `json:"routeTableId"`
+	RouteTableID *string `json:"routeTableId,omitempty"`
 
 	// NatGatewayID is the NAT gateway id associated with the subnet.
 	// Ignored unless the subnet is managed by the provider, in which case this is set on the public subnet where the NAT gateway resides. It is then used to determine routes for private subnets in the same AZ as the public subnet.


ncdc:
@detiber adding `omitempty` shouldn't be a breaking API change, I don't think. It doesn't change the OpenAPI schema in any way. Right?

ncdc:
Waiting for an answer to my question

vincepri:
Whoops I didnt see the comment

ncdc:
/hold
need to check 1 more thing

detiber:
@ncdc I think it should be fine. It won't change accepted yaml, only what is returned in the case that it is nil.
/hold cancel

ncdc:
I think we need to add `+omitempty`

ncdc:
Which will change the openapi schema

ncdc:
Thanks prow 😄

Issue #1156, opened 2019-09-29
Option to launch instances without an SSH key pair


andrewmyhre:
/kind feature

**Describe the solution you'd like**
Our organization handles SSH authentication in a specific way, I'm not familiar with the internals and it ties in to our SSO / MFA and I believe it's completely proprietary, but the TL;DR is our policy is to not use SSH keys and instead execute a process during boot to enable SSH. As such we need the option to launch instances without using an SSH key pair.

**Environment:**

- Cluster-api-provider-aws version: v0.4.0
- Kubernetes version: (use `kubectl version`): v1.15.3
- OS (e.g. from `/etc/os-release`): ubuntu 16.04

vincepri:
You should be able to specify an empty string value for `sshKeyName` https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/api/v1alpha2/types.go#L493 to achieve the suggested behavior

andrewmyhre:
Here's the relevant code:

```
if scope.AWSMachine.Spec.SSHKeyName != "" {
		input.SSHKeyName = aws.String(scope.AWSMachine.Spec.SSHKeyName)
	} else {
		input.SSHKeyName = aws.String(defaultSSHKeyName)
	}
```

Providing `sshKeyName: ""` on the `AWSMachine` specs causes the controller to try to use the default SSH key pair name which is `default`, which doesn't exist in our accounts (and we don't want to use key pairs at all).

andrewmyhre:
Here's the key pair selection code for bastions:

```
	keyName := defaultSSHKeyName
	if s.scope.AWSCluster.Spec.SSHKeyName != "" {
		keyName = s.scope.AWSCluster.Spec.SSHKeyName
	}
```

Results in the same behavior.. there is no code path allowing for 'no key pair'.

andrewmyhre:
I'm thinking something like this (in `instances.go`):

```
    if !scope.AWSMachine.Spec.NoSSHKeyPair {
        // Pick SSH key, if any.
	if scope.AWSMachine.Spec.SSHKeyName != "" {
		input.SSHKeyName = aws.String(scope.AWSMachine.Spec.SSHKeyName)
	} else {
		input.SSHKeyName = aws.String(defaultSSHKeyName)
	}
    }
```

But it's not great is it.

detiber:
[...]

PR #1198, opened 2019-10-11
🐛 Use sshKeyName from awsCluster if not set in awsMachine

### LINKED ISSUE #1101: sshKeyName on awsmachine should adopt sshKeyName on awscluster
/kind bug

The sshkeyname for cluster and machine should match because the sshkeyname used for the cluster is used on the bastion while the sshkeyname used for the machine is used on the actual instance. If they are different then you will need two keys to get onto an instance. One to ssh to the bastion and then one to ssh from the bastion to the instance. 

Are we intentionally allowing this use case for some reason? I'm trying to think of a deployment model that might use this but I'm coming up short.

I suggest we copy the keyname from the cluster to all the machines associated with the cluster.

tahsinrahman:

Fixes #1101

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -177,10 +177,11 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*infrav1.Instance, 
 	input.SecurityGroupIDs = append(input.SecurityGroupIDs, ids...)
 
 	// Pick SSH key, if any.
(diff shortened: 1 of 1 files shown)


vincepri:
@ncdc Should this go in the current milestone v0.4?

ncdc:
This seems better than what we have now (although it doesn't address #1156, not that it has to). I'd say vNext for sure. We can decide separately if we want to backport/fast-forward the release-0.4 branch.

/approve
/assign @detiber

PR #1201, opened 2019-10-11
:warning: 1156 allow no key pair

### LINKED ISSUE #1156: Option to launch instances without an SSH key pair

andrewmyhre:
Allows instances to be provisioned with no SSH key pair specified. Organizations which have policies denying the use of SSH key pairs can now set the SSH key name value to `""` to launch instances without a key pair set. A `nil` value still sets the default key pair name.

Fixes #1156


ncdc:
@andrewmyhre thanks for these PRs! This will need to go against the (not yet created) v1alpha3 API version. We'll be adding tasks soon to create the new package.

ncdc:
@andrewmyhre could you please update this PR to make the changes in the new v1alpha3 api version? Thanks!

andrewmyhre:
Closing in favor of #1243 for v1alpha3

PR #1217, opened 2019-10-15
✨ Surface inner error from `WaitForWithRetryable`

### LINKED ISSUE #1215: Surface inner error from WaitForWithRetryable
/kind feature

When `wait.WaitForWithRetryable` times out, the only error it returns is `timed out waiting for condition`. It would be better to wrap/combine that with whatever the most recent inner error that occurred, if applicable.

/help
/good-first-issue

tahsinrahman:

Fixes #1215

### DIFF
--- a/pkg/cloud/services/wait/wait.go
+++ b/pkg/cloud/services/wait/wait.go
@@ -48,7 +48,8 @@ func NewBackoff() wait.Backoff {
 
 // WaitForWithRetryable repeats a condition check with exponential backoff.
 func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, retryableErrors ...string) error { //nolint
-	return wait.ExponentialBackoff(backoff, func() (bool, error) {
+	var actualErr error
+	waitErr := wait.ExponentialBackoff(backoff, func() (bool, error) {
 		ok, err := condition()
 		if ok {
 			// All done!
@@ -69,6 +70,7 @@ func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, re
 		for _, r := range retryableErrors {
 			if code == r {
 				// We should retry.
+				actualErr = err
 				return false, nil
 			}
 		}
@@ -76,4 +78,17 @@ func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, re
 		// Got an error that we can't retry, so return it.
 		return false, err
 	})
+
+	// no error, return nil
+	if waitErr == nil {
+		return nil
+	}
+
+	// error occurred, return the actual error
+	if actualErr != nil {
+		return actualErr
+	}
+
+	// no error occurred inside condition(), the error was timeout error
+	return waitErr
 }
--- a/pkg/cloud/services/wait/wait_test.go
+++ b/pkg/cloud/services/wait/wait_test.go
@@ -1,0 +1,97 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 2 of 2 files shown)


ncdc:
Ah, I guess my write-up of the issue wasn't as clear as it could have been - I was suggesting modifying the implementation of `WaitForWithRetryable` to return the inner error (if there was one), possibly combined with information that it timed out. That way, you don't have to modify all these call sites.

tahsinrahman:
@ncdc  thanks for the clarification! I've updated the pr :D

pkg/cloud/services/wait/wait_test.go
```diff
@@ -0,0 +1,106 @@
...
+				retryableErrors: nil,
+			},
+			expectedError: nil,
+		},
+		{
+			name: "returns timeout error",
+			args: args{
+				backoff: wait.Backoff{
+					Duration: time.Second,
+					Factor:   1,
```
ncdc: 0

pkg/cloud/services/wait/wait.go
```diff
@@ -58,6 +59,7 @@ func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, re
 			// Not done, but no error, so keep waiting.
 			return false, nil
 		}
+		actualErr = err
```
detiber: Should we set `actualErr` above the `if err == nil` check above? That way it should reflect the state of the last call prior to a timeout rather than any potential transient errors in between?

detiber: On second thought, it looks like we really only care about the retryableErrors below, as anything else should be returned by `wait.ExponentialBackoff` already: https://github.com/kubernetes/apimachinery/blob/master/pkg/util/wait/wait.go#L293

pkg/cloud/services/wait/wait.go
```diff
@@ -69,11 +70,25 @@ func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, re
 		for _, r := range retryableErrors {
 			if code == r {
 				// We should retry.
+				actualErr = err
```
ncdc: I'm pretty sure this needs to come back outside - it should apply to all errors, not just retryable ones

ncdc: Nevermind, just read Jason's comment

pkg/cloud/services/wait/wait.go
```diff
@@ -48,7 +48,8 @@ func NewBackoff() wait.Backoff {
 
 // WaitForWithRetryable repeats a condition check with exponential backoff.
 func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, retryableErrors ...string) error { //nolint
-	return wait.ExponentialBackoff(backoff, func() (bool, error) {
+	var actualErr error
```
ncdc: Maybe we should call this something like `errToReturn`?

pkg/cloud/services/wait/wait.go
```diff
@@ -69,11 +70,25 @@ func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, re
...
 				return false, nil
 			}
 		}
 
 		// Got an error that we can't retry, so return it.
 		return false, err
 	})
+
+	// no error, return nil
+	if waitErr == nil {
```
detiber: I'm wondering if this should also compare against the err returned for a timeout by `wait.ExponentialBackoff` as well? 
```suggestion
	if waitErr == nil || waitErr != wait.ErrWaitTimeout {
```

tahsinrahman: I fixed the errors here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1221

pkg/cloud/services/wait/wait.go
tahsinrahman: addressed at #1221

PR #1218, opened 2019-10-15
✨ Add validation webhook for AWSMachine

tahsinrahman:

Fixes #

### DIFF
--- a/api/v1alpha3/awsmachine_webhook.go
+++ b/api/v1alpha3/awsmachine_webhook.go
@@ -17,8 +17,15 @@ limitations under the License.
 package v1alpha3
 
 import (
+	"reflect"
+
+	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/validation/field"
 	ctrl "sigs.k8s.io/controller-runtime"
 	logf "sigs.k8s.io/controller-runtime/pkg/runtime/log"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
 )
 
 // log is for logging in this package.
@@ -28,4 +35,60 @@ func (r *AWSMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
 	return ctrl.NewWebhookManagedBy(mgr).
 		For(r).
 		Complete()
+}
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine,mutating=false,failurePolicy=fail,groups=infrastructure.cluster.x-k8s.io,resources=awsmachines,versions=v1alpha3,name=validation.awsmachine.infrastructure.cluster.x-k8s.io
+
+var _ webhook.Validator = &AWSMachine{}
+
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachine) ValidateCreate() error {
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachine) ValidateUpdate(old runtime.Object) error {
+	newAWSMachine, err := runtime.DefaultUnstructuredConverter.ToUnstructured(r)
+	if err != nil {
+		return apierrors.NewInvalid(GroupVersion.WithKind("AWSMachine").GroupKind(), r.Name, field.ErrorList{
+			field.InternalError(nil, errors.Wrap(err, "failed to convert new AWSMachine to unstructured object")),
+		})
+	}
+	oldAWSMachine, err := runtime.DefaultUnstructuredConverter.ToUnstructured(old)
+	if err != nil {
+		return apierrors.NewInvalid(GroupVersion.WithKind("AWSMachine").GroupKind(), r.Name, field.ErrorList{
+			field.InternalError(nil, errors.Wrap(err, "failed to convert old AWSMachine to unstructured object")),
+		})
+	}
+
+	var allErrs field.ErrorList
+
+	newAWSMachineSpec := newAWSMachine["spec"].(map[string]interface{})
+	oldAWSMachineSpec := oldAWSMachine["spec"].(map[string]interface{})
+
+	// allow changes to providerID
+	delete(oldAWSMachineSpec, "providerID")
+	delete(newAWSMachineSpec, "providerID")
+
+	// allow changes to additionalTags
+	delete(oldAWSMachineSpec, "additionalTags")
+	delete(newAWSMachineSpec, "additionalTags")
+
+	// allow changes to additionalSecurityGroups
+	delete(oldAWSMachineSpec, "additionalSecurityGroups")
+	delete(newAWSMachineSpec, "additionalSecurityGroups")
+
+	if !reflect.DeepEqual(oldAWSMachineSpec, newAWSMachineSpec) {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "cannot be modified"))
+		return apierrors.NewInvalid(
+			GroupVersion.WithKind("AWSMachine").GroupKind(),
+			r.Name, allErrs)
+	}
+
+	return nil
+}
+
+// ValidateDelete implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachine) ValidateDelete() error {
+	return nil
 }
--- a/api/v1alpha3/awsmachine_webhook_test.go
+++ b/api/v1alpha3/awsmachine_webhook_test.go
@@ -1,0 +1,90 @@
+/*
+Copyright 2019 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha3
+
+import (
+	"testing"
+
+	"k8s.io/utils/pointer"
+)
+
+func TestAWSMachine_ValidateUpdate(t *testing.T) {
+	tests := []struct {
+		name       string
+		oldMachine *AWSMachine
+		newMachine *AWSMachine
+		wantErr    bool
+	}{
+		{
+			name: "change in providerid, tags and securitygroups",
+			oldMachine: &AWSMachine{
+				Spec: AWSMachineSpec{
+					ProviderID:               nil,
+					AdditionalTags:           nil,
+					AdditionalSecurityGroups: nil,
+				},
+			},
+			newMachine: &AWSMachine{
+				Spec: AWSMachineSpec{
+					ProviderID: pointer.StringPtr("ID"),
+					AdditionalTags: Tags{
+						"key-1": "value-1",
+					},
+					AdditionalSecurityGroups: []AWSResourceReference{
+						{
+							ID: pointer.StringPtr("ID"),
+						},
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "change in fields other than providerid, tags and securitygroups",
+			oldMachine: &AWSMachine{
+				Spec: AWSMachineSpec{
+					ProviderID:               nil,
+					AdditionalTags:           nil,
+					AdditionalSecurityGroups: nil,
+				},
+			},
+			newMachine: &AWSMachine{
+				Spec: AWSMachineSpec{
+					ImageLookupOrg: "test",
+					InstanceType:   "test",
+					ProviderID:     pointer.StringPtr("ID"),
+					AdditionalTags: Tags{
+						"key-1": "value-1",
+					},
+					AdditionalSecurityGroups: []AWSResourceReference{
+						{
+							ID: pointer.StringPtr("ID"),
+						},
+					},
+				},
+			},
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			if err := tt.newMachine.ValidateUpdate(tt.oldMachine); (err != nil) != tt.wantErr {
+				t.Errorf("ValidateUpdate() error = %v, wantErr %v", err, tt.wantErr)
+			}
+		})
+	}
+}
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -11,6 +11,24 @@ webhooks:
     service:
       name: webhook-service
       namespace: system
+      path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine
+  failurePolicy: Fail
+  name: validation.awsmachine.infrastructure.cluster.x-k8s.io
+  rules:
+  - apiGroups:
+    - infrastructure.cluster.x-k8s.io
+    apiVersions:
+    - v1alpha3
+    operations:
+    - CREATE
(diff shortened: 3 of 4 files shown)


api/v1alpha3/awsmachine_webhook.go
```diff
@@ -0,0 +1,55 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+func (r *AWSMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine,mutating=false,failurePolicy=fail,groups=infrastructure.cluster.x-k8s.io,resources=awsmachines,versions=v1alpha3,name=vawsmachine.kb.io
```
detiber: It looks like we can override the odd `vawsmachine.kb.io` naming through this annotation...  Could we name it `validating.awsmachine.x-k8s.io` here instead? @ncdc @vincepri thoughts?

ncdc: Probably would be clearer. We'll have to remember to make this change for every webhook.

detiber: @ncdc we could add a quick script to validate the default naming isn't used for CI. Might be able to even use something as naive as: `if git --no-pager grep kb.io; then echo "kb.io type prefix found"; exit 1; fi`

detiber: I created a PR for renaming the validation webhook for AWSMachineTemplate here, and we can discuss more on that PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1219

detiber:
```suggestion
// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine,mutating=false,failurePolicy=fail,groups=infrastructure.cluster.x-k8s.io,resources=awsmachines,versions=v1alpha3,name=validation.awsmachine.infrastructure.x-k8s.io
```

api/v1alpha3/awsmachine_webhook.go
```diff
@@ -0,0 +1,55 @@
...
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachine) ValidateCreate() error {
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachine) ValidateUpdate(old runtime.Object) error {
+	oldAWSMachine := old.(*AWSMachine)
+	if !reflect.DeepEqual(r.Spec, oldAWSMachine.Spec) {
+		return errors.New("awsMachineSpec is immutable")
```
ncdc:
```suggestion
		return field.Forbidden(field.NewPath("spec"), "cannot be modified")
```

ncdc:
1 small comment. @detiber @vincepri how does this look to you?

api/v1alpha3/awsmachine_webhook.go
```diff
@@ -0,0 +1,55 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+func (r *AWSMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
+	return ctrl.NewWebhookManagedBy(mgr).
+		For(r).
+		Complete()
+}
+
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine,mutating=false,failurePolicy=fail,groups=infrastructure.cluster.x-k8s.io,resources=awsmachines,versions=v1alpha3,name=validation.awsmachine.infrastructure.x-k8s.io
```
detiber:
```suggestion
// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine,mutating=false,failurePolicy=fail,groups=infrastructure.cluster.x-k8s.io,resources=awsmachines,versions=v1alpha3,name=validation.awsmachine.infrastructure.cluster.x-k8s.io
```

api/v1alpha3/awsmachine_webhook.go
detiber: I suspect we'll need to allow updates to at least a subset of the spec (controller setting `awsmachine.Spec.ProviderID` for example).

We could probably allow updates to the `AdditionalTags` and `AdditionalSecurityGroups` fields safely as well (since those should be updated on each reconcile).

It might also be helpful for people that are troubleshooting issues with initial instance creation to be able to modify fields like the AMI, ImageLookupOrg, InstanceType, IAMInstanceProfile, AvailabilityZone, Subnet, SSHKeyName, NetworkInterfaces, etc assuming the instance doesn't exist.

ncdc: Would it make sense to allow all updates as long as we haven't created an ec2 instance yet? We might need to worry about race conditions such as:

1. Simultaneously:
    1. Reconciler receives AWSMachine at ResourceVersion=1
    1. User quickly gets in some updates, now ResourceVersion=2
1. Reconciler creates ec2 instance using data from ResourceVersion=1 (maybe it's slow today)
1. Reconciler sees updated AWSMachine at ResourceVersion=2
    1. But it can't do anything at this point, so the data doesn't match what happened in ec2

detiber: @ncdc as long as we don't necessarily rely on something like the providerID being set on the awsmachine.Spec, and fallback to doing a describe for the instance we should be able to avoid race conditions.

ncdc: @detiber I don't think that helps what I described above?

detiber: Ah, I'm following now, I was thinking about a race condition resulting in orphaning. I suspect that yes it might be likely to be susceptible to the scenario that you provided.

ncdc: Which is better (or worse, depending on how you look at it)?

1. forcing the user to delete & recreate the AWSMachine to make changes (guarantees what we send to ec2 matches what's in the spec), or

2. allowing spec changes as long as we can't find the corresponding ec2 instance by tags (which might result in what we send to ec2 not matching the final state of the AWSMachine spec -- the scenario I described above)

vincepri: For this iteration, I'd keep it as simple as possible and disallow all updates, we can always iterate on the user experience later, when we have a better feeling how folks are using it

ncdc: That works for me

detiber: I think my concern here is that I've seen multiple people recently hit issues with spinning up an instance and ask for help in channel. We've gotten them back on track by having them update the definition to fix the sshKey name, iam profile, etc. 

It's possible that this will not be an issue once we build out the control plane management, but it is something that people are hitting today and deletion/recreation and update of references to a new infra config would be a big regression for resolving those types of issues.

ncdc: I don't think you'd need to adjust the reference if you keep the name the same?

ncdc: We could move the logic that's now deleted in this PR into the webhook validation code.

tahsinrahman: @ncdc our current implementation is also susceptible to this race condition you mentioned, right? so if we move the logic that's now deleted in this PR into the webhook validation code, it can also face this scenario?

ncdc: @tahsinrahman you are correct. I guess this comes back to my question from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1218/files#r344289646: is it better to allow updates with the possibility of a race, or err on the side of caution and reject all updates?

tahsinrahman: I'm on the side of disallowing all updates for now and see how it affects user experience. But lets come to an agreement on what to do! :)

detiber: sgtm, seems like the general consensus is to disallow updates for now :)

ncdc:
@chuckha the examples already set up cert-manager and webhooks. However, ultimately we need more comprehensive documentation around webhook setup, including how to use something other than cert-manager. I'd like to add that to https://github.com/kubernetes-sigs/cluster-api/issues/1757.

As far as documenting what validations we're enforcing, this will disallow all updates to AWSMachine.Spec... Although we have to address @detiber's comment:

> I suspect we'll need to allow updates to at least a subset of the spec (controller setting awsmachine.Spec.ProviderID for example).

And potentially:
> We could probably allow updates to the AdditionalTags and AdditionalSecurityGroups fields safely as well (since those should be updated on each reconcile).

I'm not sure where the best place is to document this. Any ideas?

api/v1alpha3/awsmachine_webhook.go
ncdc: I think we have consensus!

detiber:
> Should we be documenting anything on this PR? Such as what validations we are enforcing and how to set up the validation web hook? Or is everything opaque to the user and no documentation is required?

We should likely add a doc similar to https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/providers/v1alpha2-to-v1alpha3.md and add an item discussing the more strict validation there.

I don't think we should have to inform users how to deploy the validation webhook, since it will be deployed as part of the published provider components yaml.

chuckha:
@ncdc Would this fit in the cluster-api book? Sort of like how we have different instructions in the quickstart for various providers, we could have the same that link to something in this repo?

docs/webhook-validations.md perhaps? Or if it's short enough we could include this directly in the CAPI book

> > Should we be documenting anything on this PR? Such as what validations we are enforcing and how to set up the validation web hook? Or is everything opaque to the user and no documentation is required?
> 
> We should likely add a doc similar to https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/providers/v1alpha2-to-v1alpha3.md and add an item discussing the more strict validation there.
> 
> I don't think we should have to inform users how to deploy the validation webhook, since it will be deployed as part of the published provider components yaml.

We could probably add it as part of that doc I think.

"Hey, we are being stricter about what we accept, here is what we care about whereas before we did not:"

Besides that I'm thinking about where I would look for documentation on what values are accepted. I'd probably look at either the godoc or, if it were available, reference documentation on types (like the k8s reference docs).

chuckha:
> > Should we be documenting anything on this PR? Such as what validations we are enforcing and how to set up the validation web hook? Or is everything opaque to the user and no documentation is required?
> 
> We should likely add a doc similar to https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/book/src/providers/v1alpha2-to-v1alpha3.md and add an item discussing the more strict validation there.
> 
> I don't think we should have to inform users how to deploy the validation webhook, since it will be deployed as part of the published provider components yaml.

I think we could have it in that doc for the current upgrade. More informational changes to be aware of. Perhaps changes to your existing YAML you may need to make.

Then, I was thinking about future uses and where I'd go to find out validation rules. I'd probably look at either godoc or reference documentation (like the k8s reference docs). So godoc seems like a good place to put this from my perspective

tahsinrahman:
Do I have to manually go through each fields in `awsMachine.spec` one by one? or is there any helper functions that'll allow me to check only providerID, AdditionalTags and AdditionalSecurityGroups is changed?

ncdc:
Oh and we'd presumably want to allow annotations and labels to change too.

I think if you want an "easy" way, you could deep-copy both old & new, clear our the fields that we allow to be changed in both of the copies, and then compare with `reflect.DeepEqual`... maybe?

main.go
```diff
@@ -207,6 +207,11 @@ func main() {
 			os.Exit(1)
 		}
 	}
+	if err = (&infrav1alpha3.AWSMachine{}).SetupWebhookWithManager(mgr); err != nil {
+		setupLog.Error(err, "unable to create webhook", "webhook", "AWSMachine")
+		os.Exit(1)
+	}
+
```
wfernandes: This is a duplicate of what is in the above `if webhookPort != 0`
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2dc2fe8e5e968b474ee1cada99609496f0bd3af0/main.go#L197-L200

tahsinrahman: thanks, updated!

api/v1alpha3/awsmachine_webhook.go
```diff
@@ -29,3 +35,50 @@ func (r *AWSMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+	// allow changes to additionalTags
+	delete(oldAWSMachineSpec, "additionalTags")
+	delete(newAWSMachineSpec, "additionalTags")
+
+	// allow changes to additionalSecurityGroups
+	delete(oldAWSMachineSpec, "additionalSecurityGroups")
+	delete(newAWSMachineSpec, "additionalSecurityGroups")
+
+	if !reflect.DeepEqual(oldAWSMachineSpec, newAWSMachineSpec) {
+		return field.Forbidden(field.NewPath("spec"), "cannot be modified")
```
ncdc: I just learned about a better way to return validation errors - please see https://book.kubebuilder.io/cronjob-tutorial/webhook-implementation.html / xref https://github.com/kubernetes-sigs/cluster-api/issues/1795 too.

tahsinrahman: @ncdc sorry for being late. I updated the pr, ptal :)

api/v1alpha3/awsmachine_webhook.go
```diff
@@ -29,3 +36,55 @@ func (r *AWSMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+// ValidateCreate implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachine) ValidateCreate() error {
+	return nil
+}
+
+// ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
+func (r *AWSMachine) ValidateUpdate(old runtime.Object) error {
+	newAWSMachine, err := runtime.DefaultUnstructuredConverter.ToUnstructured(r)
+	if err != nil {
+		return errors.Wrap(err, "failed to convert new AWSMachine to unstructured object")
```
ncdc: Should we return `apierrors.NewInvalid()` here?

api/v1alpha3/awsmachine_webhook.go
```diff
@@ -29,3 +36,55 @@ func (r *AWSMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
...
+	}
+	oldAWSMachine, err := runtime.DefaultUnstructuredConverter.ToUnstructured(old)
+	if err != nil {
+		return errors.Wrap(err, "failed to convert old AWSMachine to unstructured object")
```
ncdc: Should we return `apierrors.NewInvalid()` here?

api/v1alpha3/awsmachine_webhook_test.go
```diff
@@ -0,0 +1,90 @@
...
+	"testing"
+
+	"k8s.io/utils/pointer"
+)
+
+func TestAWSMachine_ValidateUpdate(t *testing.T) {
+	tests := []struct {
+		name    string
+		old     *AWSMachine
+		new     *AWSMachine
```
ncdc: nit: I usually try to avoid using `new` as a variable name, given that it's a go keyword. If you want to change the name, you could use something like `updated` instead.

detiber:
/approve
leaving lgtm to @ncdc

ncdc:
@tahsinrahman if you'd like, we can merge as-is and I can open a small follow-up PR to address my remaining comments. Please let me know your preference. Thanks!

tahsinrahman:
> @tahsinrahman if you'd like, we can merge as-is and I can open a small follow-up PR to address my remaining comments. Please let me know your preference. Thanks!

oops, i somehow missed your review. let me update it in this pr

ncdc:
Thanks!
/lgtm

PR #1219, opened 2019-10-15
:running: Update name of AWSMachineTemplate validating webhook

detiber:
- Rename vawsmachinetemplate.kb.io to validating.awsmachine.infrastructure.x-k8s.io

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -13,7 +13,7 @@ webhooks:
       namespace: system
       path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachinetemplate
   failurePolicy: Fail
-  name: vawsmachinetemplate.kb.io
+  name: validation.awsmachine.infrastructure.x-k8s.io
   rules:
   - apiGroups:
     - infrastructure.cluster.x-k8s.io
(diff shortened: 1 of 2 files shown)


config/webhook/manifests.yaml
```diff
@@ -13,7 +13,7 @@ webhooks:
       namespace: system
       path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachinetemplate
   failurePolicy: Fail
-  name: vawsmachinetemplate.kb.io
+  name: validating.awsmachine.infrastructure.x-k8s.io
```
ncdc: validating or validate or validation?

tahsinrahman: This file is generated

detiber: I don't really have a preference either way, happy to go with either of those.

detiber:
bumping, anyone have more thoughts on this naming?

ncdc:
I'm partial to `validation`. And when you change it in the go source file, does one of `make generate`/`make manifests` update the generated yaml (I want to make sure it wasn't a hand edit 😄)?

detiber:
@ncdc yes the webhook definition was updated through `make generate`. Updating for s/validating/validation/

PR #1221, opened 2019-10-15
🐛 WaitForWithRetryable should return latest error

tahsinrahman:
Follow up pr for #1217 

Fixes #

### DIFF
--- a/pkg/cloud/services/wait/wait.go
+++ b/pkg/cloud/services/wait/wait.go
@@ -48,8 +48,11 @@ func NewBackoff() wait.Backoff {
 
 // WaitForWithRetryable repeats a condition check with exponential backoff.
 func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, retryableErrors ...string) error { //nolint
-	var actualErr error
+	var errToReturn error
 	waitErr := wait.ExponentialBackoff(backoff, func() (bool, error) {
+		// clear errToReturn value from previous iteration
+		errToReturn = nil
+
 		ok, err := condition()
 		if ok {
 			// All done!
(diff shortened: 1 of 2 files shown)


ncdc:
/approve
/assign @detiber 
for actual lgtm 😄

pkg/cloud/services/wait/wait.go
```diff
@@ -84,9 +87,14 @@ func WaitForWithRetryable(backoff wait.Backoff, condition wait.ConditionFunc, re
...
+	if waitErr != wait.ErrWaitTimeout {
+		return waitErr
+	}
+
+	// we are getting retryable errors always, timeout occurred
+	if errToReturn != nil {
+		return errToReturn
 	}
 
 	// no error occurred inside condition(), the error was timeout error
```
detiber: I was working on a similar PR :joy:

I had removed the initial check for `waitErr == nil` and I had slightly different comments here, which I think are a bit clearer:
```go
	// If the waitError is not a timeout error (nil or a non-retryable error), return it
	if waitErr != wait.ErrWaitTimeout {
		return waitErr
	}

	// A retryable error occurred, return the actual error
	if actualErr != nil {
		return actualErr
	}

	// The error was timeout error
	return waitErr
```

pkg/cloud/services/wait/wait_test.go
```diff
@@ -55,30 +61,59 @@ func TestWaitForWithRetryable(t *testing.T) {
...
+					firstIteration = false
+					return false, retryableError
+				}
+				return false, nil
 			},
-			retryableErrors: []string{"retryable error"},
-			expectedError:   errors.New("retryable error"),
+			retryableErrors: retryableErrorCodes,
+			expectedError:   wait.ErrWaitTimeout,
 		},
```
detiber: In my changes, I had added a test case for success after returning a retryable error as well, might be good to add that.

pkg/cloud/services/wait/wait.go
tahsinrahman: yes they're lot clearer! I updated the pr, ptal :D

detiber:
/lgtm
Thanks for addressing this @tahsinrahman

PR #1231, opened 2019-10-18
:running: Additional e2e test fixes

detiber:
Additional e2e fixes, with these changes the e2e tests are running locally for me again.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -59,13 +59,8 @@ func TestE2e(t *testing.T) {
 }
 
 const (
-	CAPI_VERSION  = "v0.2.2"
-	CABPK_VERSION = "v0.1.0"
-
 	capiNamespace       = "capi-system"
 	capiDeploymentName  = "capi-controller-manager"
-	cabpkNamespace      = "cabpk-system"
-	cabpkDeploymentName = "cabpk-controller-manager"
 	capaNamespace       = "capa-system"
 	capaDeploymentName  = "capa-controller-manager"
 	setupTimeout        = 10 * 60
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -176,6 +173,26 @@ func applyManifests(kindCluster kind.Cluster, manifests *string) {
...
+		}
+	}
+	Expect(err).NotTo(HaveOccurred())
+
+	// write out the manifests
+	manifestFile := path.Join(suiteTmpDir, "cluster-api-components.yaml")
+	Expect(ioutil.WriteFile(manifestFile, capiManifests, 0644)).To(Succeed())
+
+	// apply generated manifests
+	applyManifests(kindCluster, &manifestFile)
```
ncdc: Follow-up: change `applyManifests` so it takes in a `string`, not a `*string`?

test/e2e/e2e_suite_test.go
```diff
@@ -176,6 +173,26 @@ func applyManifests(kindCluster kind.Cluster, manifests *string) {
 	kindCluster.ApplyYAML(*manifests)
 }
 
+func deployCAPIComponents(kindCluster kind.Cluster) {
+	fmt.Fprintf(GinkgoWriter, "Generating CAPI manifests\n")
+
+	// Build the manifests using kustomize
+	capiManifests, err := exec.Command(*kustomizeBinary, "build", "https://github.com/kubernetes-sigs/cluster-api//config/default").Output()
```
ncdc: Any concern about reading all of this into memory vs using `StdoutPipe` and an `io.Copy`?

ncdc:
No blocking comments - will lgtm after we wrap up discussing the exec.Command. Thanks!

test/e2e/e2e_suite_test.go
detiber: Maybe, but this is only temporary until we publish capi manifests

test/e2e/e2e_suite_test.go
detiber: hmm, now that you mention it this should probably just call `kindCluster.ApplyYAML` instead :joy:

PR #1241, opened 2019-10-22
:sparkles: Control plane load balancer subnets can be specified in spec

### LINKED ISSUE #1180: Allow specifying control plane load balancer subnets
/kind feature

In our organization subnet creation is tightly controlled. We are given a set of subnets in our VPCs which we are permitted to use. There are also subnets we are not permitted to use.
I would like to be able to specify existing subnets by ID in the AWSCluster spec, which may not be public and may not be named or tagged by any convention (relating to CAPA anyway), for the control plane load balancer to use (machine subnets can already be specified, is my understanding).

**Environment:**

- Cluster-api-provider-aws version: 0.4.1
- Kubernetes version: (use `kubectl version`): 1.15.3 / 1.15.0
- OS (e.g. from `/etc/os-release`): ubuntu 1604

andrewmyhre:
This PR allows subnets for the control plane load balancer to specified in the AWSCluster.Spec.ControlPlaneLoadBalancer struct. In environments where subnets are all private and/or cannot be created/tagged by non-privileged users it is necessary to specify the subnets to be used.

Note: in my environment I'm unable to manually test the operator behavior without specifying subnets, so I can't verify that I haven't broken those code paths.

Fixes #1180


api/v1alpha3/awscluster_types.go
```diff
@@ -52,6 +52,10 @@ type AWSLoadBalancerSpec struct {
 	// Scheme sets the scheme of the load balancer (defaults to Internet-facing)
 	// +optional
 	Scheme *ClassicELBScheme `json:"scheme,omitempty"`
+
+	// Subnets specifies the subnets that should be used by the load balancer
+	// +optional
+	Subnets Subnets `json:"subnets,omitempty"`
```
vincepri: Can you provide an example in the godoc section on how to use this? From the implementation it looks like the AvailabilityZone field is required for example, otherwise it can fail / error, is that correct?

andrewmyhre: Yes, in my usage I did need to specify availabilityZone. Would it be overreach to remove that requirement as part of this PR?

Happy to update the godoc section, but I'm new to godoc. Is it just a matter of adding information to the comments?

vincepri:
@andrewmyhre We have an upstream proposal to address this issue for v1alpha3 https://github.com/kubernetes-sigs/cluster-api/issues/1269. Do you want to port this patch change to v1alpha2 instead?

ncdc:
@vincepri did you mean to link to https://github.com/kubernetes-sigs/cluster-api/issues/1647?

vincepri:
Yes, thanks @ncdc

andrewmyhre:
We could close this PR and I could look at providing a PR to support the new proposal.

vincepri:
@andrewmyhre Sounds good to me, I'll go ahead and close this for now.

/close

PR #1245, opened 2019-10-22
:running: [e2e] Fetch controller logs when tearing down kind cluster

detiber:

- Updates the e2e test suite to fetch the controller logs when tearing down the kind cluster, this should help with troubleshooting failures.

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -20,9 +20,11 @@ package e2e_test
 
 import (
 	"bytes"
+	"context"
 	"encoding/base64"
 	"fmt"
 	"io/ioutil"
+	"io"
 	"os"
 	"os/exec"
 	"path"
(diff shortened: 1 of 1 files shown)


detiber:
There appears to be a flake with installing kind :(
/test pull-cluster-api-provider-aws-integration

test/e2e/e2e_suite_test.go
```diff
@@ -126,10 +133,53 @@ var _ = BeforeSuite(func() {
...
+		return
+	}
+
+	fmt.Fprintf(GinkgoWriter, "CAPI Logs:\n%s\n", capiLogs)
+	fmt.Fprintf(GinkgoWriter, "CAPA Logs:\n%s\n", capaLogs)
+}
+
+func retrieveLogs(namespace, deploymentName string) string {
+	deployment := &appsv1.Deployment{}
+	Expect(kindClient.Get(context.TODO(), apimachinerytypes.NamespacedName{Namespace: namespace, Name: deploymentName}, deployment)).To(Succeed())
```
ncdc: You should be able to use `crclient.ObjectKey` which is "aliased" to `NamespacedName`, if you want to avoid the extra import

detiber: updated, thanks!

PR #1246, opened 2019-10-22
:running: [e2e] retrieve controller logs when tearing down kind cluster

detiber:
- Adds fetching of the controller logs to the teardown of the kind cluster when running e2e tests.


detiber:
/assign @ncdc 
basically a backport of #1245 with needed modifications for cabpk.

PR #1250, opened 2019-10-23
✨ LOAD_IMAGE to allow injecting locally built image into kind

dims:
Similar to what we did in https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/188 we should be able to build a local image and use that in `kind`

Fixes #


dims:
@vincepri as mentioned on slack, let's get this in as is (using `LOAD_IMAGE`) as this is the pattern we already have in CAPG, if we don't like it we can change all to the same pattern as a follow up.

Also, filed a backport to release-0.4 https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1253 as the master is b0rked! and i need this somewhere that works

/hold cancel

PR #1254, opened 2019-10-24
 ✨e2e-conformance.sh to be able to run e2e tests

dims:

Fixes #

### DIFF
--- a/examples/controlplane/controlplane.yaml
+++ b/examples/controlplane/controlplane.yaml
@@ -23,6 +23,8 @@ metadata:
   name: ${CLUSTER_NAME}-controlplane-0
 spec:
   instanceType: ${CONTROL_PLANE_MACHINE_TYPE}
+  ami:
+    id: ${IMAGE_ID}
   iamInstanceProfile: "control-plane.cluster-api-provider-aws.sigs.k8s.io"
   sshKeyName: "${SSH_KEY_NAME}"
 ---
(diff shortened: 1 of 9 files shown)


examples/controlplane/kustomization.yaml
```diff
@@ -5,3 +5,5 @@ resources:
 - controlplane.yaml
 configurations:
 - kustomizeconfig.yaml
+patchesStrategicMerge:
+- kustomizeversions.yaml
```
detiber: I worry a bit about applying this to the default examples. This seems like something we should apply *only* in the case of the conformance of kubernetes tip rather than always apply.

examples/machinedeployment/kustomization.yaml
```diff
@@ -5,3 +5,5 @@ resources:
 - machinedeployment.yaml
 configurations:
 - kustomizeconfig.yaml
+patchesStrategicMerge:
+- kustomizeversions.yaml
```
detiber: Same comment her as with the controlplane kustomization.yaml

examples/controlplane/kustomization.yaml
dims: I agree. i don't know customize enough. is there a better way? 

worst case i can append those two lines and then run kustomize (when needed)?

detiber: +1 to appending as needed in the short term.

dims: fwiw, the shell script is a no-op when CI_VERSION is not set.

examples/machinedeployment/kustomization.yaml
dims: as mentioned above. ... i can open a follow up issue

examples/controlplane/kustomization.yaml
detiber: I mainly worry that this will affect people that are currently consuming these files.

PR #1256, opened 2019-10-24
:bug: Improve EC2 state handling, and set error for manually deleted EC2 instances

### LINKED ISSUE #1238: Log entry for every event
/kind bug

Debugging a machine that hasn't reconciled correctly which has an InvalidUpdate 
entry in its Events, but didn't see a corresponding log entry.

A log entry published in addition to the Event.

**Environment:**

- Cluster-api-provider-aws version: 0.4.3
- Kubernetes version: (use `kubectl version`):  N/A

### LINKED ISSUE #1239: Terminated instance not detected
/kind bug

If an EC2 instance is deleted from when the AWSMachine is in a Running state, warning InvalidUpdate events will be recorded, but the AWSMachine does not transition to Error.

AWSMachine transitions to Error for user deletion.

Missing log entries related to the publishing of events as per #1238 

Test case:
1. Deploy as per quick start guide
2. Wait for worker nodes to reach running (verify with kubectl get machines)
3. Wait 10 minutes
4. Delete the EC2 instance backing one of the worker nodes

**Environment:**

- Cluster-api-provider-aws version: 0.4.3

randomvariable:
Tweaks EC2 instance state handling in the following ways:
* Emit log event as well as publish event to resource for better traceability
* Ignore instance attributes for validation when the instance is terminated, or about to transition to terminated from "shutting-down".
* Don't error when EC2 instance is stopping or is stopped, as machine could be rebooted.

Fixes #1239 
Fixes #1238

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -189,13 +189,13 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 		// and AWSMachine
 		// 3. Issue a delete
 		// 4. Scale controller deployment to 1
-		machineScope.V(2).Info("Unable to locate instance by ID or tags")
+		machineScope.V(2).Info("Unable to locate EC2 instance by ID or tags")
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "NoInstanceFound", "Unable to find matching EC2 instance")
 		machineScope.AWSMachine.Finalizers = util.Filter(machineScope.AWSMachine.Finalizers, infrav1.MachineFinalizer)
 		return reconcile.Result{}, nil
 	}
 
-	machineScope.V(3).Info("Instance found matching deleted AWSMachine", "instanceID", instance.ID)
+	machineScope.V(3).Info("EC2 instance found matching deleted AWSMachine", "instance-id", instance.ID)
 
 	// Check the instance state. If it's already shutting down or terminated,
 	// do nothing. Otherwise attempt to delete it.
@@ -203,9 +203,9 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
 	switch instance.State {
 	case infrav1.InstanceStateShuttingDown, infrav1.InstanceStateTerminated:
-		machineScope.Info("Instance is shutting down or already terminated", "instanceID", instance.ID)
+		machineScope.Info("EC2 instance is shutting down or already terminated", "instance-id", instance.ID)
 	default:
-		machineScope.Info("Terminating instance", "instanceID", instance.ID)
+		machineScope.Info("Terminating EC2 instance", "instance-id", instance.ID)
 		if err := ec2Service.TerminateInstanceAndWait(instance.ID); err != nil {
 			r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedTerminate", "Failed to terminate instance %q: %v", instance.ID, err)
 			return reconcile.Result{}, errors.Wrap(err, "failed to terminate instance")
@@ -231,6 +231,7 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 			}
 		}
 
+		machineScope.Info("EC2 instance successfully terminated", "instance-id", instance.ID)
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeNormal, "SuccessfulTerminate", "Terminated instance %q", instance.ID)
 	}
 
@@ -276,6 +277,7 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 
 	// If the AWSMachine doesn't have our finalizer, add it.
 	if !util.Contains(machineScope.AWSMachine.Finalizers, infrav1.MachineFinalizer) {
+		machineScope.V(1).Info("Adding Cluster API Provider AWS finalizer")
 		machineScope.AWSMachine.Finalizers = append(machineScope.AWSMachine.Finalizers, infrav1.MachineFinalizer)
 	}
 
@@ -300,6 +302,7 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 
 	// Set an error message if we couldn't find the instance.
 	if instance == nil {
+		machineScope.Info("EC2 instance cannot be found")
 		machineScope.SetErrorReason(capierrors.UpdateMachineError)
 		machineScope.SetErrorMessage(errors.New("EC2 instance cannot be found"))
 		return reconcile.Result{}, nil
@@ -308,6 +311,7 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 	// TODO(ncdc): move this validation logic into a validating webhook
 	if errs := r.validateUpdate(&machineScope.AWSMachine.Spec, instance); len(errs) > 0 {
 		agg := kerrors.NewAggregate(errs)
+		machineScope.Info("Invalid update", "failedUpdates", agg.Error())
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InvalidUpdate", "Invalid update: %s", agg.Error())
 		return reconcile.Result{}, nil
 	}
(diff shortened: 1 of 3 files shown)


vincepri:
Reviewing

controllers/awsmachine_controller.go
```diff
@@ -390,14 +400,21 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 // validateUpdate checks that no immutable fields have been updated and
 // returns a slice of errors representing attempts to change immutable state.
 func (r *AWSMachineReconciler) validateUpdate(spec *infrav1.AWSMachineSpec, i *infrav1.Instance) (errs []error) {
+
+	// EC2 instance attributes start disapearing during shutdown and termination, so do not
+	// perform more checks
+	if i.State == infrav1.InstanceStateTerminated || i.State == infrav1.InstanceStateShuttingDown {
+		return errs
```
vincepri:
```suggestion
		return nil
```

controllers/awsmachine_controller.go
```diff
@@ -323,11 +327,16 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
+		machineScope.Info("EC2 instance is running", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
 		machineScope.SetReady()
 	case infrav1.InstanceStatePending:
-		machineScope.Info("Machine instance is pending", "instance-id", *machineScope.GetInstanceID())
+		machineScope.Info("EC2 instance is pending", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
+		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstancePending", "EC2 instance is %q", instance.State)
+	case infrav1.InstanceStateStopping, infrav1.InstanceStateStopped:
+		machineScope.Info("EC2 instance is hibernating", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
+		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstanceHibernation", "EC2 instance is %q", instance.State)
 	default:
```
vincepri: Can we add a comment here that states the unexpected states and that we're going to transition to failed state?

controllers/awsmachine_controller.go
```diff
@@ -323,11 +327,16 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
 	switch instance.State {
 	case infrav1.InstanceStateRunning:
-		machineScope.Info("Machine instance is running", "instance-id", *machineScope.GetInstanceID())
+		machineScope.Info("EC2 instance is running", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
 		machineScope.SetReady()
 	case infrav1.InstanceStatePending:
-		machineScope.Info("Machine instance is pending", "instance-id", *machineScope.GetInstanceID())
+		machineScope.Info("EC2 instance is pending", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
+		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstancePending", "EC2 instance is %q", instance.State)
+	case infrav1.InstanceStateStopping, infrav1.InstanceStateStopped:
```
vincepri: This needs upstream support to make the Machine to flip back to `provisioned` state. Once https://github.com/kubernetes-sigs/cluster-api/issues/1622#issuecomment-545575154 is fixed

detiber: If it is stopped, it's not guaranteed to have the same networkID when it returns up, so stopping/stopped should likely result in a permanent failure.

vincepri: Got it, so it should possibly transition to a failure state, like it did before.

@randomvariable what's the reason behind this change?

randomvariable: what do you mean network ID? The IP address remains the same (https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html):
> The instance keeps its private IPv4 address. The instance gets a new public IPv4 address, unless it has an Elastic IP address, which doesn't change during a stop/start.

vincepri: In that case, the node should be able to come back up. Would you mind testing this use case on an actual cluster?

detiber: Ah, right...  I had forgotten we aren't assigning public IPs :joy:

Do we have to worry about Node deletion with this, though? Doesn't the AWS cloud provider delete Nodes for instances that are in a stopped state?

controllers/awsmachine_controller.go
randomvariable: Done. Explicitly stating the states, and making sure the entire lifecycle is covered. Added a link to the place in the documentation describing the states as well.

controllers/awsmachine_controller.go
randomvariable: checking now.

controllers/awsmachine_controller_test.go
```diff
@@ -27,7 +27,10 @@ import (
 	. "github.com/onsi/gomega"
 	. "github.com/onsi/gomega/gstruct"
 	"github.com/pkg/errors"
+<<<<<<< HEAD
```
detiber: looks like some rebase conflicts

randomvariable:
Ready for review.
Have added descriptive unit tests based on the various transitions that can occur.

controllers/awsmachine_controller.go
```diff
@@ -321,13 +325,29 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
+
+		// We do not error yet so that the final terminated state is captured.
+	case infrav1.InstanceStateShuttingDown:
+		machineScope.V(1).Info("Unexpected EC2 instance termination", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
+		machineScope.SetUnready()
+		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstanceUnexpectedTermination", "Unexpected instance termination")
+	case infrav1.InstanceStateTerminated:
+		machineScope.V(1).Info("Unexpected EC2 instance termination", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
+		machineScope.SetUnready()
+		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstanceUnexpectedTermination", "Unexpected instance termination")
```
vincepri: These two cases could be simplified to a single one capturing both

controllers/awsmachine_controller.go
```diff
@@ -321,13 +325,29 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
 		machineScope.SetReady()
-	case infrav1.InstanceStatePending:
-		machineScope.Info("Machine instance is pending", "instance-id", *machineScope.GetInstanceID())
-	default:
+	case infrav1.InstanceStateStopping:
+		machineScope.V(1).Info("EC2 instance is stopping", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
+		machineScope.SetUnready()
+	case infrav1.InstanceStateStopped:
+		machineScope.V(1).Info("EC2 instance is stopped", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
+		machineScope.SetUnready()
```
vincepri: Same as below, these can be simplified in a single case condition capturing both

pkg/cloud/scope/machine.go
```diff
@@ -154,6 +154,11 @@ func (m *MachineScope) SetReady() {
 	m.AWSMachine.Status.Ready = true
 }
 
+// SetUnready sets the AWSMachine Ready Status to false
+func (m *MachineScope) SetUnready() {
```
vincepri:
```suggestion
func (m *MachineScope) SetNotReady() {
```

Nit for readability

controllers/awsmachine_controller.go
randomvariable: These ones can't. Not erroring in the case of InstanceStateShuttingDown, because our state machine can then no longer transition to Terminated, hence the comment.

randomvariable:
Tweaked to be less noisy in the logs.

controllers/awsmachine_controller.go
```diff
@@ -308,26 +311,36 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
 	switch instance.State {
+	case infrav1.InstanceStatePending, infrav1.InstanceStateStopping, infrav1.InstanceStateStopped:
+		machineScope.SetNotReady()
 	case infrav1.InstanceStateRunning:
-		machineScope.Info("Machine instance is running", "instance-id", *machineScope.GetInstanceID())
 		machineScope.SetReady()
-	case infrav1.InstanceStatePending:
-		machineScope.Info("Machine instance is pending", "instance-id", *machineScope.GetInstanceID())
-	default:
+	case infrav1.InstanceStateShuttingDown, infrav1.InstanceStateTerminated:
```
vincepri: Don't we need a default case at all here?

controllers/awsmachine_controller.go
```diff
@@ -189,23 +189,23 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 		// and AWSMachine
 		// 3. Issue a delete
 		// 4. Scale controller deployment to 1
-		machineScope.V(2).Info("Unable to locate instance by ID or tags")
+		machineScope.V(2).Info("Unable to locate EC2 instance by ID or tags")
```
vincepri: In the future, it'd be great if we can separate commits or PRs that aren't in the same scope

controllers/awsmachine_controller.go
randomvariable: Nope. There's literally no other options. Gone for explicitness so people know what the control flow is for each state.

vincepri: Usually it's good practice adding default cases to catch unexpected values. For example, if a future SDK version add a new state which we don't look out for, we might incur in unexpected behavior. 

Having it would give us a safeguard, wdyt?

randomvariable: It's not an SDK thing, it's the EC2 API, this has remained stable for years, and would be a major change. I'd rather we actually fail and add the appropriate handling.
I think soundness is better than completeness in this specific case.

controllers/awsmachine_controller.go
randomvariable: will do

controllers/awsmachine_controller.go
randomvariable: Blaming the brain worms. Yup, added with test.

PR #1258, opened 2019-10-24
:bug: Disassociate Elastic IPs on deletion, if still associated

### LINKED ISSUE #1185: Cluster deletion fails for failed created cluster
/kind bug

I tried create a cluster and because of low memory assignment to CAPA pod, it crashed mid-way cluster creation.
Following this, I initiated a delete for the cluster and also increased the memory limit.
This left me with the below error on every reconcilation:

```
E1008 00:03:23.622390       1 actuator.go:232] [cluster-actuator] "msg"="Error deleting cluster" "error"="failed to release elastic IP \"100.21.75.12\" with allocation ID \"eipalloc-0f24df24d49556be8\": Still associated with association ID \"eipassoc-00317523598562bb7\""  "cluster-name"="lcm-1-10-2" "cluster-namespace"="7d523c5270ff6b52d2985a7cc18896ca721c31d5"
```

The CAPA pod would just print this log forever and never finish out successfully.
On checking more, the problem appeared to be that while creation of cluster, AWS resources were successfully created and tagged but before the cluster object could be updated with subnet/vpc ids, CAPA crashed. Now during deletion CAPA would not try to delete subnets/vpc and fail on deletion of EIPs.

Suggested solution:
1. Making deletion more robust.
2. Reconcile completely before attempting to delete.

randomvariable:
Disassociate an Elastic IP if it is still associated, most commonly with an ELB
that is no longer visible in the API after deletion.

Fixes #1185 

**For release note**:
```
CloudFormation will need to be reapplied to allow Cluster API Provider AWS to disassociate IP addresses.
```

### DIFF
--- a/pkg/cloud/services/ec2/eips.go
+++ b/pkg/cloud/services/ec2/eips.go
@@ -87,6 +87,25 @@ func (s *Service) describeAddresses(role string) (*ec2.DescribeAddressesOutput, 
 	})
 }
 
+func (s *Service) disassociateAddress(ip *ec2.Address) error {
+	err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+		_, err := s.scope.EC2.DisassociateAddress(&ec2.DisassociateAddressInput{
+			AssociationId: ip.AssociationId,
+		})
+		if err != nil {
+			cause, _ := awserrors.Code(errors.Cause(err))
+			if cause != awserrors.AssociationIDNotFound {
+				return false, err
+			}
+		}
+		return true, nil
+	}, awserrors.AuthFailure)
+	if err != nil {
+		return errors.Wrapf(err, "failed to disassociate ElasticIP %q", *ip.AllocationId)
+	}
+	return nil
+}
+
 func (s *Service) releaseAddresses() error {
 	out, err := s.scope.EC2.DescribeAddresses(&ec2.DescribeAddressesInput{
 		Filters: []*ec2.Filter{filter.EC2.Cluster(s.scope.Name())},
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/ec2/eips.go
```diff
@@ -98,7 +98,12 @@ func (s *Service) releaseAddresses() error {
 
 	for _, ip := range out.Addresses {
 		if ip.AssociationId != nil {
-			return errors.Errorf("failed to release elastic IP %q with allocation ID %q: Still associated with association ID %q", *ip.PublicIp, *ip.AllocationId, *ip.AssociationId)
+			_, err := s.scope.EC2.DisassociateAddress(&ec2.DisassociateAddressInput{
```
detiber: Should this use `wait.WaitForWithRetryable`?

randomvariable: Can do, or it will error and come back on next reconcile if it fails. Disassociation seemed to be immediate from my tests though.

vincepri: Can we move this block and if check in the retryable below before ReleaseAddress?

detiber: I think that would assume they both return the same "retryable" errors, I'm not sure that is the case.

randomvariable: Resolved this by moving to its own retryable function where it handles the errors specific to it.

vincepri:
/retitle :bug: Disassociate Elastic IPs on deletion, if still associated

vincepri:
@randomvariable Can you squash down to 1 commit?

detiber:
lgtm once commits are squashed.

randomvariable:
Squasheeed

PR #1267, opened 2019-10-25
:running: [e2e] Create access key after creating the prerequisites

detiber:
- Additional e2e fixes


chuckha:
/lgtm

discussed offline

PR #1273, opened 2019-10-25
:bug: Fix Machine reconciliation of deletion

detiber:

Fixes a nil pointer dereference in Machine reconcileDelete

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -174,7 +174,7 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
 func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) (reconcile.Result, error) {
 	machineScope.Info("Handling deleted AWSMachine")
 
-	ec2Service := r.serviceFactory(clusterScope)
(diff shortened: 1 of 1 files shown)


liztio:
:sweat_smile: 

/lgtm

controllers/awsmachine_controller.go
```diff
@@ -174,7 +174,7 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
 func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope, clusterScope *scope.ClusterScope) (reconcile.Result, error) {
 	machineScope.Info("Handling deleted AWSMachine")
 
-	ec2Service := r.serviceFactory(clusterScope)
+	ec2Service := r.getEC2Service(clusterScope)
```
ncdc: I think `serviceFactory` is for unit testing (see https://github.com/kubernetes-sigs/cluster-api-provider-aws/commit/101c68eac91844ad01ba366b5f43f072888525a3). We probably need to default the variable in the constructor so it works for the normal path (non units)

ncdc:
/hold
pending unit testing comment

controllers/awsmachine_controller.go
ncdc: Nevermind, read the code fully.

PR #1279, opened 2019-10-28
✨Fix up e2e conformance tests (release 0.4)

dims:
- Use python script for envsubst
- Add / Delete keypair if not preset
- Avoid printing unnecessary stuff
- Support specifying a lookup org for images
- option for skipping creation of the AMI image

Fixes #

### DIFF
--- a/examples/machinedeployment/machinedeployment.yaml
+++ b/examples/machinedeployment/machinedeployment.yaml
@@ -6,7 +6,7 @@ metadata:
     cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
     nodepool: nodepool-0
 spec:
-  replicas: 1
+  replicas: 2
   selector:
     matchLabels:
       cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
--- a/hack/ci/e2e-conformance.sh
+++ b/hack/ci/e2e-conformance.sh
@@ -76,6 +76,8 @@ cleanup() {
   fi
   # clean up e2e.test symlink
   (cd "$(go env GOPATH)/src/k8s.io/kubernetes" && rm -f _output/bin/e2e.test) || true
+
+  delete_key_pair
 }
 
 # our exit handler (trap)
@@ -120,9 +122,10 @@ init_image() {
       rm packer.zip && \
       ln -s $PWD/packer /usr/local/bin/packer
   fi
-  (cd "$(go env GOPATH)/src/sigs.k8s.io/image-builder/images/capi" && \
-    sed -i 's/1\.15\.4/1.16.1/' packer/config/kubernetes.json && \
-    sed -i 's/1\.15/1.16/' packer/config/kubernetes.json)
+
+  tracestate="$(shopt -po xtrace)"
+  set +o xtrace
+
   if [[ $EUID -ne 0 ]]; then
     # install goss plugin
     (cd "$(go env GOPATH)/src/sigs.k8s.io/image-builder/images/capi/packer/ami" && \
@@ -136,11 +139,15 @@ init_image() {
     # assume we are running in the CI environment as root
     # Add a user for ansible to work properly
     groupadd -r packer && useradd -m -s /bin/bash -r -g packer packer
+    # Ensure go is available in PATH
+    ln -s /usr/local/go/bin/go /usr/bin/go
     # install goss plugin
     su - packer -c "bash -c 'cd /go/src/sigs.k8s.io/image-builder/images/capi/packer/ami && make plugins'"
     # use the packer user to run the build
     su - packer -c "bash -c 'cd /go/src/sigs.k8s.io/image-builder/images/capi && AWS_REGION=$AWS_REGION AWS_ACCESS_KEY_ID=${AWS_ACCESS_KEY_ID:-""} AWS_SECRET_ACCESS_KEY=${AWS_SECRET_ACCESS_KEY:-""} make build-ami-default'"
   fi
+
+  eval "$tracestate"
 }
 
 # build kubernetes / node image, e2e binaries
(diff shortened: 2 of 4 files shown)


hack/heartbeat_account.py
```diff
@@ -0,0 +1,47 @@
...
+# limitations under the License.
+
+import sys
+import httplib
+import urllib
+import os
+import time
+
+BOSKOS_HOST=os.environ.get("BOSKOS_HOST", "boskos")
+BOSKOS_RESOURCE_NAME=os.environ['BOSKOS_RESOURCE_NAME']
```
detiber: Is anything actually setting `BOSKOS_RESOURCE_NAME`?

detiber: After some digging, it looks like this also affects the `hack/checkin_account.py` script, where I think we also have the same issue.

dims: yes, it's easy to miss. we print the variable in the python script `checkout_account.py` and then source that stdout in:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/scripts/ci-conformance.sh#L52

dims: then both checkout and heartbeat scripts get it from the env var.

examples/machinedeployment/machinedeployment.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
     cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
     nodepool: nodepool-0
 spec:
-  replicas: 1
+  replicas: 2
```
aojea: is this a cluster only with 2 nodes?

detiber: 2 worker nodes (minimum required for conformance testing), there should also be 3 control plane machines as well.

dims: 1 bastion node, 3 controllers, 2 workers (with this change included)

hack/heartbeat_account.py
```diff
@@ -0,0 +1,47 @@
...
+
+USER = "cluster-api-provider-aws"
+
+if __name__ == "__main__":
+    count = 0
+    # keep sending heart beat for 3 hours
+    while count < 180:
+        conn = httplib.HTTPConnection(BOSKOS_HOST)
+        print "POST-ing heartbeat for resource %s to %s" % (BOSKOS_RESOURCE_NAME, BOSKOS_HOST)
+        conn.request("POST", "/update?%s" % urllib.urlencode({
```
aojea: maybe use it inside a `try ... except` block to handle timeouts per example or network hiccups ??

scripts/ci-conformance.sh
```diff
@@ -69,7 +83,8 @@ if grep -iqF "$(echo "${AWS_ACCESS_KEY_ID-}" | \
   exit 1
 fi
 
-hack/ci/e2e-conformance.sh --verbose
+# TODO(dims): remove "SKIP_INIT_IMAGE" once we fix problem building images in CI
```
aojea: The images are built using kubernetes master?

hack/heartbeat_account.py
dims: don't want this to be resilient, if things fail we should bail out

scripts/ci-conformance.sh
dims: by default we pick up images with k8s 1.16.2

hack/ci/e2e-conformance.sh
```diff
@@ -174,8 +181,26 @@ generate_manifests() {
...
+      --region ${AWS_REGION} --output json | jq -r '.[0][0] | select (.!=null)')
+    if [[ -z "$image_id" ]]; then
+      echo "unable to find image using : $filter $IMAGE_LOOKUP_ORG ... bailing out!"
+      exit 1
+    fi
+  else
+    image_id=$(aws ec2 describe-images --image-ids $IMAGE_ID \
+      --query 'Images[*].[ImageId,Name]' --output json | jq -r '.[0][0] | select (.!=null)')
+    echo "using specified image id : ${IMAGE_ID}"
+    if [[ -z "$image_id" ]]; then
```
aojea: nit, this block seems redundant and can be used outside the main if block
```
if [[ -z ${IMAGE_ID:-} ]];
  ...
else
  ...
fi
if [[ -z "$image_id" ]]; then
  echo "unable to find image using : $filter $IMAGE_LOOKUP_ORG ... bailing out!"
  exit 1
fi
```

aojea:
this is an impressive work.
I'd be interested in adapting it to run conformance against an IPv6 cluster, is it something feasible?

examples/machinedeployment/machinedeployment.yaml
aojea: is there a load balancer in front of the controllers for the APIs?

aojea:
las run failed because the process of provisioning the workers got stuck
```
1030 02:25:31.968] test1-controlplane-0          aws:////i-0d55666950be00782   running
I1030 02:25:31.969] test1-controlplane-1          aws:////i-09b6460ab6b13bf08   running
I1030 02:25:31.969] test1-controlplane-2          aws:////i-0b79b4553c5f26669   running
I1030 02:25:31.969] test1-md-0-6c94dcb947-j4jxg                                 provisioning
I1030 02:25:31.969] test1-md-0-6c94dcb947-jj65z                                 provisioning
I1030 02:25:31.970] cluster did not start in 30 mins ... bailing out!
I1030 02:25:31.970] === versions ===
```
/test pull-cluster-api-provider-aws-conformance

detiber:
@aojea We've discovered what we believe is an issue with the aws janitor that is being run by boskos, where it is not fully cleaning the account prior to it being issued to a new job. We believe this is the reason why the workers are failing to provision at this time.

dims:
@aojea yes it's feasible once we get this ready a bit more :)

Also @detiber i added some code to force clean and then create stack, lets see if that helps

aojea:
/test pull-cluster-api-provider-aws-conformance
progress :muscle:

PR #1247, opened 2019-10-23
🏃 Add MachineDeployment, scale up, down tests

sirao:
**What this PR does / why we need it**:
Existing E2E tests doesn't cover MachineDeployment, and scaling it. This PR covers the below:
1. MachineDeployment
2. scale up MachineDeployment
3. scale down MachineDeployment
4. scale again MachineDeployment

Verified that the added tests are running and succeeding in local run

PR #1289, opened 2019-10-29
:running: e2e k8s conformance

bhargavmad:
:running:
Runs k8s conformance tests on workload cluster as part of e2e testing
 - Uses sonobuoy library to achieve it
 - Outputs a result tarball


Makefile
```diff
@@ -87,6 +87,11 @@ test-e2e: ## Run e2e tests
 	PULL_POLICY=IfNotPresent $(MAKE) docker-build
 	go test -v -tags=e2e -timeout=1h ./test/e2e/... -args -ginkgo.v --managerImage $(CONTROLLER_IMG)-$(ARCH):$(TAG)
```
detiber: Do we also need to add a focus here as well? Otherwise wouldn't this make target also cause the conformance tests to run?

test/e2e/aws_test.go
```diff
@@ -107,7 +107,10 @@ var _ = Describe("functional tests", func() {
 			machineName = cpMachinePrefix + "-2"
 			createAdditionalControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName)
 
-			// TODO: Deploy a MachineDeployment
+			By("Deploying a MachineDeployment")
+			machineDeploymentName := clusterName + "-worker"
+			createMachineDeployment(namespace, clusterName, machineDeploymentName, 3)
+
```
detiber: Are these changes needed? It looks like the required changes are in the helper methods rather than here. 

This will also need to be reconciled with https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1247 after it is merged.

test/e2e/aws_test.go
```diff
@@ -406,7 +517,7 @@ func makeAWSMachine(namespace, name string) {
 			Namespace: namespace,
 		},
 		Spec: infrav1.AWSMachineSpec{
-			InstanceType:       "t3.large",
+			InstanceType:       "m5.large",
```
detiber: This should probably be exposed as a parameter for `makeAWSMachine` to limit costs for running the tests contained in `aws_test.go` compared to `conformance_test.go`

Makefile
bhargavmad: Oh yes. i missed that. Thanks.

go.mod
```diff
@@ -24,3 +26,5 @@ require (
 	sigs.k8s.io/controller-runtime v0.3.0
 	sigs.k8s.io/yaml v1.1.0
 )
+
+replace k8s.io/client-go => k8s.io/client-go v0.0.0-20190918200256-06eb1244587a
```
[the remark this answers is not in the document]
bhargavmad: As I understand, sonobuoy has this dependency on a particular version of k8s/client-go and getting a newer version of client-go is breaking it. I hate to do this as well but i couldnt find a better way.

test/e2e/aws_test.go
bhargavmad: Yes. I should probably wait for #1247

test/e2e/aws_test.go
bhargavmad: Sounds good. Will do.

test/e2e/conformance_test.go
```diff
@@ -0,0 +1,201 @@
...
+	}
+	eg := &errgroup.Group{}
+	eg.Go(func() error { return <-ec })
+	eg.Go(func() error {
+		filesCreated, err := client.UntarAll(reader, ".", "")
+		if err != nil {
+			return errors.Wrap(err, "couldn't untar sonobuoy results")
+		}
+		for _, name := range filesCreated {
+			fmt.Fprintf(GinkgoWriter, "Find conformance test results at: test/e2e/%s\n", name)
```
johnSchnake: Unless I'm misreading this, I believe that the `filesCreated` is the tarball itself and so the path "test/e2e/<name>" isn't right. 

E.g. when you run `sonobuoy retrieve` it prints the filenames of the tarball it downloaded. Maybe this was misleading because of the `UntarAll` but it is untarring the response from the server which in most cases results in a single tarball name in `filesCreated` but (in theory) could be more tarballs.

go.mod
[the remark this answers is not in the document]
ncdc: If we can't fix this here, I propose we make `test/e2e` its own go module with the `replace` statement there.

test/e2e/sonobuoy-config-tmpl.yaml
```diff
@@ -0,0 +1,185 @@
...
+      command:
+      - /run_e2e.sh
+      env:
+      - name: E2E_FOCUS
+        value: \[Conformance\]
+      - name: E2E_SKIP
+      - name: E2E_PARALLEL
+        value: "1"
+      - name: E2E_USE_GO_RUNNER
+        value: "true"
```
johnSchnake: As we spoke about on slack; this will only be useful if you have a v1.16+ cluster since the gorunner didnt exist before that.

test/e2e/sonobuoy-config-tmpl.yaml
```diff
@@ -0,0 +1,185 @@
...
...ents","endpoints","ingresses","jobs","leases","limitranges","mutatingwebhookconfigurations","namespaces","networkpolicies","nodes","persistentvolumeclaims","persistentvolumes","poddisruptionbudgets","pods","podlogs","podsecuritypolicies","podtemplates","priorityclasses","replicasets","replicationcontrollers","resourcequotas","rolebindings","roles","servergroups","serverversion","serviceaccounts","services","statefulsets","storageclasses","validatingwebhookconfigurations","volumeattachments"],"Filters":{"Namespaces":".*","LabelSelector":""},"Limits":{"PodLogs":{"Namespaces":"","SonobuoyNamespace":true,"FieldSelectors":[],"LabelSelector":"","Previous":false,"SinceSeconds":null,"SinceTime":null,"Timestamps":false,"TailLines":null,"LimitBytes":null,"LimitSize":"","LimitTime":""}},"QPS":30,"Burst":50,"Server":{"bindaddress":"0.0.0.0","bindport":8080,"advertiseaddress":"","timeoutseconds":10800},"Plugins":[{"name":"e2e"},{"name":"systemd-logs"}],"PluginSearchPath":["./plugins.d","/etc/sonobuoy/plugins.d","~/sonobuoy/plugins.d"],"Namespace":"sonobuoy","WorkerImage":"sonobuoy/sonobuoy:{{.SonobuoyVersion}}","ImagePullPolicy":"IfNotPresent","ImagePullSecrets":"","ProgressUpdatesPort":"8099"}
```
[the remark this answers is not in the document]
bhargavmad: Yes. Even I didnt like the way it is. But didnt know a better way. Any suggestion?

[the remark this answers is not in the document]
johnSchnake: I'm 95% sure that if you just delete the UUID then it will be dynamically generated on the server when loaded.

ncdc:
I want to avoid needing to use a `replace` directive in CAPA's main `go.mod` file... we could either make `test/e2e` a separate go module, or switch to running sonobuoy either as a binary, or as a pod.

also
/cc @chuckha

johnSchnake:
In general the approach of executing the run using Sonobuoy as a library is just fine; I agree with the approach of using the `GenFile` so that you can look at/tweak the raw YAML that will be used and not even worry about all the `GenConfig`->YAML translation.

I think that doing it this way (using Sonobuoy as a library) will be easier to maintain long term:
 - No need to build/publish/maintain the extra wrapper and logic
 - By avoiding using the CLI in a wrapper you really limit the amount of Sonobuoy you're using which means less surface area for bugs/etc. You'll only want/need to update as core server features are updated if you want them.
 - By avoiding use of the CLI you wont need to worry about us maintaining that CLI level compatibility; we may break CLI compatibility without too much thought but are more careful when it comes to the config object itself.

go.mod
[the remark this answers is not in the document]
detiber: :grimacing: I think I would rather do the replace in the main `go.mod` file if we need to override it for sonobuoy, otherwise we risk shipping something different than we are testing.

test/e2e/sonobuoy-config-tmpl.yaml
```diff
@@ -0,0 +1,185 @@
...
+    sonobuoy-config:
+      driver: Job
+      plugin-name: e2e
+      result-format: junit
+    spec:
+      command:
+      - /run_e2e.sh
+      env:
+      - name: E2E_FOCUS
+        value: \[Conformance\]
```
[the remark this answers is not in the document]
detiber: Disruptive tests should be fine for our use case, since we are looking to test conformance specifically here and these clusters will not be used for other purposes.

go.mod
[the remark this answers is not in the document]
detiber: I take that back, as long as we can isolate the system under test, ensuring that the controller manager that is built is using the main `go.mod` file and only the test suite/utilities are using a different version of client-go, then this approach should be fine. It might also help to keep other testing dependencies from creeping into the main `go.mod` file as well.

bhargavmad:
@chuckha can you please have a look at this PR

chuckha:
yep, sorry for the delay, looking now

test/e2e/conformance_test.go
```diff
@@ -0,0 +1,208 @@
...
+
+	restConfig, err := clientcmd.RESTConfigFromKubeConfig(kubeConfigData)
+	if err != nil {
+		return errors.Wrap(err, "couldn't get rest config from kubeconfig")
+	}
+
+	sonobuoyKubeCli, err := sonodynamic.NewAPIHelperFromRESTConfig(restConfig)
+	if err != nil {
+		return err
+		return errors.Wrap(err, "couldn't get rest config from kubeconfig")
```
chuckha: unreachable code

test/e2e/conformance_test.go
```diff
@@ -0,0 +1,208 @@
...
+	if err := sonobuoyClient.Run(runConfig); err != nil {
+		return errors.Wrap(err, "error attempting to run sonobuoy")
+	}
+	reader, ec, err := sonobuoyClient.RetrieveResults(&client.RetrieveConfig{Namespace: "sonobuoy"})
+	if err != nil {
+		return errors.Wrap(err, "couldn't retrieve sonobuoy results")
+	}
+	eg := &errgroup.Group{}
+	eg.Go(func() error { return <-ec })
+	eg.Go(func() error {
```
chuckha: Are we doing anything with the results or do you think it's enough that if we're able to retrieve the results then they must have passed?

chuckha:
we may also be able to move this into the e2e test framework in capi at some point and provide conformance as an importable test

test/e2e/conformance_test.go
detiber: It should likely been written out to the artifact path, similar to what we are currently doing in the e2e suite: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/test/e2e/e2e_suite_test.go#L70-L77

As long as the file is named junit.*.xml it will be picked up by test-grid.

chuckha: Oh, I see 👍 

consider adding a comment here

bhargavmad:
@chuckha @detiber I have addressed all the comments. Please have a look.
Also, i am not very sure of the rebase procedure and i am scared i might mess up things. @detiber can you please help me out.

test/e2e/e2e_suite_test.go
```diff
@@ -95,13 +95,14 @@ var (
...
-	suiteTmpDir string
-	region      string
+	kindCluster  kind.Cluster
+	kindClient   crclient.Client
+	sess         client.ConfigProvider
+	accountID    string
+	accessKey    *iam.AccessKey
+	suiteTmpDir  string
+	region       string
+	artifactPath string
```
bhargavmad: I see that `artifactPath` is not part of the `release-0.4` branch. So this doesnt do anything now. But when merged with master, it should place the conformance results in the `artifactPath`

dims:
may i suggest avoiding `conformance`? it has a specific connotation in our community.  `validation` may be a better term.

detiber:
@dims this does indeed run k8s conformance using sonobuoy

dims:
ah excellent, sorry about the noise :)

bhargavmad:
@detiber can you push the clean branch onto this and merge if no changes required. Thanks.

PR #1302, opened 2019-10-30
:running: Update generate examples to v1alpha3

vincepri:

Fixes #

### DIFF
--- a/examples/controlplane/controlplane.yaml
+++ b/examples/controlplane/controlplane.yaml
@@ -1,11 +1,12 @@
-apiVersion: cluster.x-k8s.io/v1alpha2
+---
+apiVersion: cluster.x-k8s.io/v1alpha3
 kind: Machine
 metadata:
   name: ${CLUSTER_NAME}-controlplane-0
   labels:
     cluster.x-k8s.io/control-plane: "true"
-    cluster.x-k8s.io/cluster-name: "${CLUSTER_NAME}"
 spec:
+  clusterName: "${CLUSTER_NAME}"
   version: ${KUBERNETES_VERSION}
   bootstrap:
     configRef:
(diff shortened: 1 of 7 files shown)


examples/controlplane/controlplane.yaml
```diff
@@ -88,26 +89,27 @@ spec:
...
         cloud-provider: aws
 ---
-apiVersion: cluster.x-k8s.io/v1alpha2
+apiVersion: cluster.x-k8s.io/v1alpha3
 kind: Machine
 metadata:
   name: ${CLUSTER_NAME}-controlplane-2
   labels:
     cluster.x-k8s.io/control-plane: "true"
     cluster.x-k8s.io/cluster-name: "${CLUSTER_NAME}"
```
detiber:
```suggestion
```

To align with the previous Machines

dims:
examples/generate.sh changes look good. thanks @vincepri

dims:
/lgtm
/hold

please feel free to remove hold when you are ready

examples/machinedeployment/machinedeployment.yaml
```diff
@@ -3,20 +3,19 @@ kind: MachineDeployment
...
   name: ${CLUSTER_NAME}-md-0
   labels:
-    cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
     nodepool: nodepool-0
 spec:
+  clusterName: "${CLUSTER_NAME}"
   replicas: 1
   selector:
     matchLabels:
-      cluster.x-k8s.io/cluster-name: ${CLUSTER_NAME}
```
detiber: we probably want to keep this in matchlabels to avoid accidentally matching a machines that belong to a different cluster if someone re-uses values for `nodepool` label for example.

Either that or we'll need to start appending the cluster label to the provided matchLabels implicitly in the machinedeployment/machineset controllers.

vincepri: That's a good point, what do you think of doing that automatically in CAPI?

detiber: No objections, but it will need to likely be reflected in the generated `Status.Selector` fields as well

PR #1306, opened 2019-10-31
:running: [e2e] Add boskos heartbeat

detiber:

Adds a boskos heartbeat script to the e2e tests to prevent the account from being reaped prior to tests finishing

### DIFF
--- a/hack/heartbeat_account.py
+++ b/hack/heartbeat_account.py
@@ -1,0 +1,47 @@
+#!/usr/bin/env python
+
+# Copyright 2019 The Kubernetes Authors.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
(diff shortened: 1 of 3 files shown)


hack/heartbeat_account.py
```diff
@@ -0,0 +1,44 @@
...
+import sys
+import httplib
+import urllib
+import os
+import time
+
+BOSKOS_HOST=os.environ.get("BOSKOS_HOST", "boskos")
+BOSKOS_RESOURCE_NAME=os.environ['BOSKOS_RESOURCE_NAME']
+
+USER = "cluster-api-provider-gcp"
```
dims: wrong `USER` @detiber

scripts/ci-e2e.sh
```diff
@@ -59,6 +59,9 @@ if [ -n "${BOSKOS_HOST:-}" ]; then
     echo "error getting account from boskos" 1>&2
     exit "${checkout_account_status}"
   fi
+
+  python -u hack/heartbeat_account.py >> $ARTIFACTS/boskos.log 2>&1 &
```
dims: i had made a boo boo here as well. we need `$ARTIFACTS/boskos.log` to be `$ARTIFACTS/logs/boskos.log` and we need to ensure that the directory is created

dims:
/hold

needs changes ...

hack/heartbeat_account.py
detiber: fixed

scripts/ci-e2e.sh
detiber: I've been dumping all of the other logs directly under artifacts for now, so this is probably fine for now

PR #1307, opened 2019-10-31
[e2e] Add boskos heartbeat

detiber:

Adds a boskos heartbeat script to the e2e tests to prevent the account from being reaped prior to tests finishing


scripts/ci-e2e.sh
```diff
@@ -59,6 +59,9 @@ if [ -n "${BOSKOS_HOST:-}" ]; then
     echo "error getting account from boskos" 1>&2
     exit "${checkout_account_status}"
   fi
+
+  python -u hack/heartbeat_account.py >> $ARTIFACTS/boskos.log 2>&1 &
```
dims: please ensure there is a `/logs` directory in between

```
  ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
  mkdir -p "$ARTIFACTS/logs/"
  python -u hack/heartbeat_account.py >> "$ARTIFACTS/logs/boskos.log" 2>&1 &
```

PR #1309, opened 2019-10-31
✨ AWSCluster: Allow setting ImageLookupOrg at the cluster level

### LINKED ISSUE #1292: Add ImageLookupOrg to AWSClusterSpec
/kind feature

Currently we expose awsMachine.Spec.ImageLookupOrg to override the org used for image lookups, we should also expose this under the AWSClusterSpec to allow per-cluster overriding of this value similar to the way that we handle the defaulting for `SSHKeyName`

alexbrand:

This PR adds a new field, `ImageLookupOrg`, to the `AWSClusterSpec` struct. The `ImageLookupOrg` field can be used to override the organization ID used when looking up AMIs for machines. Note that the `ImageLookupOrg` is only used when a machine does not specify an AMI ID.

With this change, the `ImageLookupOrg` can be set at the cluster-level or at the machine-level. The machine-level `ImageLookupOrg` takes precedence over the cluster-level `ImageLookupOrg`.

Fixes #1292

### DIFF
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -45,6 +45,12 @@ type AWSClusterSpec struct {
 	// ControlPlaneLoadBalancer is optional configuration for customizing control plane behavior
 	// +optional
 	ControlPlaneLoadBalancer *AWSLoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
+
+	// ImageLookupOrg is the AWS Organization ID to look up machine images when a
+	// machine does not specify an AMI. When set, this will be used for all
+	// cluster machines unless a machine specifies a different ImageLookupOrg.
+	// +optional
+	ImageLookupOrg string `json:"imageLookupOrg,omitempty"`
 }
 
 // AWSLoadBalancerSpec defines the desired state of an AWS load balancer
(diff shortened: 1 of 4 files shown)


alexbrand:
@detiber I noticed that the CRD yaml got restructured a bit in this PR. Is that expected?

detiber:
> @detiber I noticed that the CRD yaml got restructured a bit in this PR. Is that expected?

Yes, this is the first divergence that I'm aware of with v1alpha2 -> v1alpha3, so this is expected :)

alexbrand:
Got it! Thanks @detiber

api/v1alpha3/awscluster_types.go
```diff
@@ -45,6 +45,12 @@ type AWSClusterSpec struct {
 	// ControlPlaneLoadBalancer is optional configuration for customizing control plane behavior
 	// +optional
 	ControlPlaneLoadBalancer *AWSLoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
+
+	// ImageLookupOrg is the AWS Organization ID to lookup machine images when a
```
ncdc:
```suggestion
	// ImageLookupOrg is the AWS Organization ID to look up machine images when a
```

PR #1310, opened 2019-10-31
Update cert manager / fix Makefile targets

vincepri:

Fixes #

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -293,18 +293,6 @@ release-notes: $(RELEASE_NOTES)
 ## Development
 ## --------------------------------------
 
-.PHONY: create-cluster
-create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS using examples
-	$(CLUSTERCTL) \
-	create cluster -v 4 \
-	--bootstrap-flags="name=clusterapi" \
-	--bootstrap-type kind \
-	-m ./examples/_out/controlplane.yaml \
-	-c ./examples/_out/cluster.yaml \
-	-p ./examples/_out/provider-components.yaml \
-	-a ./examples/addons.yaml
-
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -293,25 +293,18 @@ release-notes: $(RELEASE_NOTES)
...
 	kind create cluster --name=clusterapi
 	@if [ ! -z "${LOAD_IMAGE}" ]; then \
 		echo "loading ${LOAD_IMAGE} into kind cluster ..." && \
 		kind --name="clusterapi" load docker-image "${LOAD_IMAGE}"; \
 	fi
+	# Install cert manager.
+	kubectl \
+		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
+		create -f https://github.com/jetstack/cert-manager/releases/download/v0.11.0/cert-manager.yaml
+	sleep 20
```
vincepri: @tahsinrahman I know you have another PR to actually wait for the cert manager targets to be available, what do you think if we merge this first and then add your waits in here before we apply the provider components?

config/certmanager/certificate.yaml
```diff
@@ -1,22 +1,22 @@
 # The following manifests contain a self-signed issuer CR and a certificate CR.
```
detiber: Any reason to keep the certmanager directory here, since it is an external requirement (either capi or a pre-req of capi)?

Makefile
tahsinrahman: sgtm! :)

config/certmanager/certificate.yaml
vincepri: These are the certs for the controller as I understood, which are built when running kustomize

PR #1314, opened 2019-10-31
✨ Add addresses to machine status

rsmitty:
This PR is pretty similar to one merged in the GCP provider a few hours ago. 

It will add the ability to fish out the internal and external IPs
associated with an awsmachine resource and populate those values in
.status.addresses.

Happy to backport to release-0.4 after merge as well.

Signed-off-by: Spencer Smith <robertspencersmith@gmail.com>

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -321,6 +321,8 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 	// TODO(vincepri): Remove this annotation when clusterctl is no longer relevant.
 	machineScope.SetAnnotation("cluster-api-provider-aws", "true")
 
+	machineScope.SetAddresses(instance.Addresses)
+
 	switch instance.State {
 	case infrav1.InstanceStateRunning:
 		machineScope.Info("Machine instance is running", "instance-id", *machineScope.GetInstanceID())
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	"k8s.io/klog/klogr"
 	"k8s.io/utils/pointer"
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
@@ -170,6 +171,11 @@ func (m *MachineScope) SetAnnotation(key, value string) {
 		m.AWSMachine.Annotations = map[string]string{}
 	}
 	m.AWSMachine.Annotations[key] = value
+}
+
+// SetAddresses sets the AWSMachine address status.
+func (m *MachineScope) SetAddresses(addrs []corev1.NodeAddress) {
+	m.AWSMachine.Status.Addresses = addrs
 }
 
 // Close the MachineScope by updating the machine spec, machine status.
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -29,6 +29,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/awserrors"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/converters"
@@ -310,16 +311,16 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 	}
 
 	if len(i.NetworkInterfaces) > 0 {
-		instances := make([]*ec2.InstanceNetworkInterfaceSpecification, 0, len(i.NetworkInterfaces))
+		netInterfaces := make([]*ec2.InstanceNetworkInterfaceSpecification, 0, len(i.NetworkInterfaces))
 
 		for index, id := range i.NetworkInterfaces {
-			instances = append(instances, &ec2.InstanceNetworkInterfaceSpecification{
+			netInterfaces = append(netInterfaces, &ec2.InstanceNetworkInterfaceSpecification{
 				NetworkInterfaceId: aws.String(id),
 				DeviceIndex:        aws.Int64(int64(index)),
 			})
 		}
 
-		input.NetworkInterfaces = instances
+		input.NetworkInterfaces = netInterfaces
 	} else {
 		input.SubnetId = aws.String(i.SubnetID)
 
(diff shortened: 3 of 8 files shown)


controllers/awsmachine_controller.go
```diff
@@ -24,6 +24,7 @@ import (
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
```
detiber: This is already imported as `corev1` above
```suggestion
```

controllers/awsmachine_controller.go
```diff
@@ -321,6 +322,22 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 	// TODO(vincepri): Remove this annotation when clusterctl is no longer relevant.
 	machineScope.SetAnnotation("cluster-api-provider-aws", "true")
 
+	// Set addresses for awsmachine status
+	addrs := []v1.NodeAddress{
```
detiber:
```suggestion
	addrs := []corev1.NodeAddress{
```

controllers/awsmachine_controller.go
```diff
@@ -321,6 +322,22 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
+		v1.NodeAddress{
```
detiber:
```suggestion
		corev1.NodeAddress{
```

controllers/awsmachine_controller.go
```diff
@@ -321,6 +322,22 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
+			Type:    v1.NodeInternalIP,
```
detiber:
```suggestion
			Type:    corev1.NodeInternalIP,
```

controllers/awsmachine_controller.go
```diff
@@ -321,6 +322,22 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
+			Address: aws.StringValue(instance.PrivateIP),
+		},
+	}
+	if instance.PublicIP != nil {
+		publicAddress := v1.NodeAddress{
```
detiber:
```suggestion
		publicAddress := corev1.NodeAddress{
```

controllers/awsmachine_controller.go
```diff
@@ -321,6 +322,22 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
+			Type:    v1.NodeExternalIP,
```
detiber:
```suggestion
			Type:    corev1.NodeExternalIP,
```

pkg/cloud/scope/machine.go
```diff
@@ -21,6 +21,7 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	v1 "k8s.io/api/core/v1"
```
```suggestion
	corev1 "k8s.io/api/core/v1"
```
For consistency with use elsewhere

pkg/cloud/scope/machine.go
```diff
@@ -149,6 +150,11 @@ func (m *MachineScope) SetInstanceState(v infrav1.InstanceState) {
 	m.AWSMachine.Status.InstanceState = &v
 }
 
+// SetAddresses sets the AWSMachine address status.
+func (m *MachineScope) SetAddresses(addrs []v1.NodeAddress) {
```
```suggestion
func (m *MachineScope) SetAddresses(addrs []corev1.NodeAddress) {
```

controllers/awsmachine_controller.go
```diff
@@ -321,6 +322,22 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
...
+			Address: aws.StringValue(instance.PublicIP),
+		}
+		addrs = append(addrs, publicAddress)
+	}
+	machineScope.SetAddresses(addrs)
```
detiber: Should this also potentially iterate over attached ENIs as well (we do allow users to specify one or more ENIs through `Machine.Spec.NetworkInterfaces` today)?

rsmitty:
Went ahead and amended the `v1 -> corev1` import issues. Thanks for the heads up there. I need a little more time to explore the ENI stuff. Will tackle that this morning.

pkg/cloud/services/ec2/instances.go
```diff
@@ -485,7 +485,8 @@ func (s *Service) UpdateResourceTags(resourceID *string, create map[string]strin
 	return nil
 }
 
-func (s *Service) getInstanceENIs(instanceID string) ([]*ec2.NetworkInterface, error) {
+// GetInstanceENIs returns a list of ec2 network interfaces for a given instance
+func (s *Service) GetInstanceENIs(instanceID string) ([]*ec2.NetworkInterface, error) {
```
rsmitty: Made this a public func so that I could call it from machine controller.

pkg/cloud/services/ec2/instances.go
```diff
@@ -310,16 +310,16 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 	}
 
 	if len(i.NetworkInterfaces) > 0 {
-		instances := make([]*ec2.InstanceNetworkInterfaceSpecification, 0, len(i.NetworkInterfaces))
+		netInterfaces := make([]*ec2.InstanceNetworkInterfaceSpecification, 0, len(i.NetworkInterfaces))
```
rsmitty: these were mistakenly called instances I think. updated the name to be more clear that they're network interfaces.

controllers/awsmachine_controller.go
```diff
@@ -373,6 +379,40 @@ func (r *AWSMachineReconciler) getOrCreate(scope *scope.MachineScope, ec2svc ser
 	return instance, nil
 }
 
+func (r *AWSMachineReconciler) getAddresses(instance *infrav1.Instance, ec2svc *ec2.Service) ([]corev1.NodeAddress, error) {
+	addresses := []corev1.NodeAddress{}
+
+	enis, err := ec2svc.GetInstanceENIs(instance.ID)
```
rsmitty: every instance has one or more ENIs attached regardless of if it's specified in the spec, as you can have a pre-created ENI or an ephemeral one for this instance. changed the logic to just range through the ENIs and fish out IPs and DNS names.

rsmitty:
@detiber I reworked the logic for this to just look for the ENIs directly. PTAL and see what you think. It results in a pretty nice status output like: 
```
...
  status:
    addresses:
    - address: ip-172-31-11-240.ec2.internal
      type: InternalDNS
    - address: 172.31.11.240
      type: InternalIP
    - address: ec2-x-x-x-x.compute-1.amazonaws.com
      type: ExternalDNS
    - address: x.x.x.x
      type: ExternalIP
...
```

controllers/awsmachine_controller.go
```diff
@@ -373,6 +379,40 @@ func (r *AWSMachineReconciler) getOrCreate(scope *scope.MachineScope, ec2svc ser
 	return instance, nil
 }
 
+func (r *AWSMachineReconciler) getAddresses(instance *infrav1.Instance, ec2svc *ec2.Service) ([]corev1.NodeAddress, error) {
```
vincepri: I wonder if this should be part of the describe instance rather than having the AWS-specific logic in the reconciler

rsmitty:
Okay, @vincepri, I think I've got a pretty decent implementation of what we talked about earlier today. Updated the logic to add an addresses field to the instance structs, as well as moving the logic to populate that field into the SDKToInstance function. It then bubbles up from there to get added into the awsmachine resource's status inside the controller.

PR #1316, opened 2019-11-01
✨Better command line processing in e2e-conformance.sh

dims:
Allow stuff to be passed in as command line params

Also switch on the flag to use latest CI artifacts

Fixes #

### DIFF
--- a/scripts/ci-conformance.sh
+++ b/scripts/ci-conformance.sh
@@ -84,7 +84,7 @@ if grep -iqF "$(echo "${AWS_ACCESS_KEY_ID-}" | \
 fi
 
 # TODO(dims): remove "SKIP_INIT_IMAGE" once we fix problem building images in CI
-SKIP_INIT_IMAGE=yes hack/ci/e2e-conformance.sh --verbose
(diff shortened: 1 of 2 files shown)


scripts/ci-conformance.sh
```diff
@@ -84,7 +84,7 @@ if grep -iqF "$(echo "${AWS_ACCESS_KEY_ID-}" | \
 fi
 
 # TODO(dims): remove "SKIP_INIT_IMAGE" once we fix problem building images in CI
-SKIP_INIT_IMAGE=yes hack/ci/e2e-conformance.sh --verbose
+hack/ci/e2e-conformance.sh --verbose --skip-init-image --use-ci-artifacts
```
detiber: Wouldn't we want to limit the use of `--use-ci-artifacts` to just the "stable" job?

dims: @detiber y, there's more tweaking to do here. i want to be able to specify this in the CI job definition itself as well. For now, i want to enable this across all the jobs to see how it holds up over the weekend

PR #1319, opened 2019-11-01
✨Expose ImageLookupBaseOS field in AWSCluster and AWSMachine CRDs

### LINKED ISSUE #1291: Expose ability to filter the image lookup based on base OS
/kind feature

Currently we hardcode the naming convention of the Ubuntu images built by the image-builder for our image lookups, instead we should expose selection of the underlying base OS to the users to allow them to consume any of the images that we publish (Ubuntu, CentOS, Amazon Linux 2).

We should add a field to the AWSMachineSpec for `ImageLookupBaseOS` and provide defaulting to the existing images. I don't think we should provide rigid validation of this field, since it is something that could potentially be customized by users who fork image-builder.

We should probably also expose this as a field on AWSClusterSpec to provide cluster-wide defaulting for the base OS to use in filtering images.

alexbrand:

This PR enables CAPA users to specify the base operating system they want to use for cluster machines.

Before this PR, the only way to use a different base OS was by specifying the AMI ID. With this change, users can now specify (for example) `imageLookupBaseOS: "centos-7"` in their `AWSCluster` or `AWSMachine` resources.

The `AWSMachine` `imageLookupBaseOS` overrides the `AWSCluster` `imageLookupBaseOS`. If neither are specified, CAPA uses `ubuntu-18.04` as the default.

Fixes #1291 

**Note for reviewers**
Before this change, the base OS name and base OS version were treated as separate parameters. In this PR, I have refactored the code to treat base OS name and version as a single parameter. The reason for this is that we want to expose a single knob (`imageLookupBaseOS`) via the CRDs to specify base OS name and version (e.g. `centos-7`, `ubuntu-18.04`, `amazon-2`)


api/v1alpha3/awscluster_types.go
```diff
@@ -51,6 +51,12 @@ type AWSClusterSpec struct {
 	// cluster machines unless a machine specifies a different ImageLookupOrg.
 	// +optional
 	ImageLookupOrg string `json:"imageLookupOrg,omitempty"`
+
+	// ImageLookupBaseOS is the name of the base operating system used to look
+	// up machine images when a machine does not specify an AMI. When set, this
+	// will be used for all cluster machines unless a machine specifies a
+	// different ImageLookupBaseOS.
+	ImageLookupBaseOS string `json:"imageLookupBaseOS,omitempty"`
```
vincepri: This looks good for now, in the future it seems that we might benefit from having a struct that incorporates these fields, instead of having multiple fields with the same prefix

alexbrand: Ah yeah - good call!

alexbrand:
ping @detiber @vincepri

vincepri:
@alexbrand Might need to rebase, and regenerate the manifests

detiber:
It looks like a conversion function is needed for the newly added field:
```
E1108 16:13:52.819882    4439 conversion.go:667] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.AWSMachineSpec -> sigs.k8s.io/cluster-api-provider-aws/api/v1alpha2.AWSMachineSpec
E1108 16:13:52.820520    4439 conversion.go:668]   the following fields need manual conversion:
E1108 16:13:52.820762    4439 conversion.go:670]       - ImageLookupBaseOS
```

ncdc:
Running an e2e test to make sure the lookups still work

ncdc:
So far so good. Will lgtm when it finishes.

PR #1320, opened 2019-11-01
Update CI Kubebuilder / k8s 1.15.5

vincepri:


detiber:
@vincepri should the PR title reference that this is for testing and not actually modifying dependencies?

vincepri:
I omitted the emoji so it won't show up in the release notes

vincepri:
@detiber ptal

PR #1321, opened 2019-11-01
:running: [e2e] Improve logging and add event logging

detiber:
- Improves the e2e suite logging handling
- Adds event logging for the e2e suite

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -35,7 +35,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachinerytypes "k8s.io/apimachinery/pkg/types"
-	"k8s.io/apimachinery/pkg/util/wait"
 	"k8s.io/client-go/informers"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/clientcmd"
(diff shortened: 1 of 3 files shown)


detiber:
/assign @chuckha 
This should be good to go now.

chuckha:
Is this a flaky test?

detiber:
> Is this a flaky test?

Yes, it is a tad bit flaky and it is currently running against k8s master for conformance.

chuckha:
Do you have any concerns about this code in general?

detiber:
> Do you have any concerns about this code in general?

Mainly looking for feedback on the ways that I'm doing the asynchronous handling the logging and events. I mostly arrived at the current solutions through brute force.

chuckha:
> > Do you have any concerns about this code in general?
> 
> Mainly looking for feedback on the ways that I'm doing the asynchronous handling the logging and events. I mostly arrived at the current solutions through brute force.

got it, will focus on that, thanks

test/e2e/aws_test.go
```diff
@@ -134,11 +151,50 @@ var _ = Describe("functional tests", func() {
...
+		DeleteFunc: func(obj interface{}) {},
+	})
+
+	stopInformer := make(chan struct{})
+	defer close(stopInformer)
+	informerFactory.Start(stopInformer)
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
chuckha: I think the only other thing you could do here is to send a `struct{}{}` onto the `stopInformer` channel here before returning

test/e2e/aws_test.go
```diff
@@ -134,11 +151,50 @@ var _ = Describe("functional tests", func() {
...
+			f.WriteString(fmt.Sprintf("[Updated Event] %s/%s\n\tresource: %s/%s/%s\n\treason: %s\n\tmessage: %s\n\tfull: %#v\n",
+				e.Namespace, e.Name, e.InvolvedObject.APIVersion, e.InvolvedObject.Kind, e.InvolvedObject.Name, e.Reason, e.Message, e))
+		},
+		DeleteFunc: func(obj interface{}) {},
+	})
+
+	stopInformer := make(chan struct{})
+	defer close(stopInformer)
+	informerFactory.Start(stopInformer)
+	for {
```
chuckha: This shouldn't need a for loop, the select should block on the one case before it returns. Actually i'm not even sure this needs a select.

I think this could just be

```go
<-ctx.Done()
stopInformer<-struct{}{}
```

PR #1322, opened 2019-11-01
[WIP] :sparkles: add IPv6 support

aojea:

Allow creating IPv6 clusters in AWS

AWS IPv6 has several limitations: it assigns a fixed ipv6 /56 prefix, can't use NAT gateways, subnets has to be dual-stack, ...

Fixes #


aojea:
/cc @dims 

This is WIP, I couldn't test it yet, but would love some guidance and suggestions to understand if I'm in the right path

pkg/cloud/services/ec2/subnets.go
```diff
@@ -235,11 +256,21 @@ func (s *Service) describeVpcSubnets() (infrav1.Subnets, error) {
...
 		VpcId:            aws.String(s.scope.VPC().ID),
 		CidrBlock:        aws.String(sn.CidrBlock),
 		AvailabilityZone: aws.String(sn.AvailabilityZone),
-	})
+	}
+
+	if s.scope.VPC().EnableIPv6 {
+		// The CIDR block has to have a /64 prefix size
+		// and use the prefix assigned to the VPC
+		ipv6CidrBlock := fmt.Sprintf("%s:%s::/64", s.scope.VPC().Ipv6CidrBlock, strconv.Itoa(int(sn.Ipv6CidrBlock)))
```
aojea: this is wrong, have to figure out how to deal better with this conversion

pkg/cloud/services/ec2/subnets.go
```diff
@@ -235,11 +256,21 @@ func (s *Service) describeVpcSubnets() (infrav1.Subnets, error) {
...
 		VpcId:            aws.String(s.scope.VPC().ID),
 		CidrBlock:        aws.String(sn.CidrBlock),
 		AvailabilityZone: aws.String(sn.AvailabilityZone),
-	})
+	}
+
+	if s.scope.VPC().EnableIPv6 {
+		// The CIDR block has to have a /64 prefix size
+		// and use the prefix assigned to the VPC
+		ipv6CidrBlock := fmt.Sprintf("%s:%0x::/64", s.scope.VPC().Ipv6CidrBlock, sn.Ipv6CidrBlock)
```
aojea: TODO: have to check what value is returning aws in  s.scope.VPC().Ipv6CidrBlock to create the right ipv6CidrBlack for the subnet

dims:
cc @detiber @vincepri

examples/cluster/cluster.yaml
```diff
@@ -6,7 +6,7 @@ metadata:
 spec:
   clusterNetwork:
     pods:
-      cidrBlocks: ["192.168.0.0/16"]
+      cidrBlocks: ["${CLUSTER_CIDR}"]
```
detiber: Should this potentially support setting both ipv4 and ipv6 CIDRs?

aojea: yeah, but that's dual stack and is pretty alpha status, I'd prefer to take one step at a time.
Have an IPv6 cluster (the hosts will be dual stack but the cluster single stack) and once it works we can move forward

detiber: sgtm, thanks for clarifying

aojea:
I think that master e2e is not working 🤔 , I'm trying to work with 0.4 branch 
https://prow.k8s.io/job-history/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-aws-make-conformance

aojea:
/close
we'll reopen if I have time, but is polluting your backlog right now

PR #1338, opened 2019-11-08
✨Fetch logs from nodes

dims:
Let's collect some logs from AWS nodes, will be useful for debugging

Fixes #

### DIFF
--- a/hack/ci/e2e-conformance.sh
+++ b/hack/ci/e2e-conformance.sh
@@ -69,6 +69,33 @@ dump-logs() {
 
   # export all logs from kind
   kind "export" logs --name="clusterapi" "${ARTIFACTS}/logs" || true
+
+  jump_node=$(aws ec2 describe-instances --region $AWS_REGION --query "Reservations[*].Instances[*].PublicIpAddress" --output text | head -1)
+  for node in $(aws ec2 describe-instances --region $AWS_REGION --query "Reservations[*].Instances[*].PrivateIpAddress" --output text | tail -n +2)
+  do
+    echo "collecting logs from ${node} using jump host ${jump_node}"
+    dir="${ARTIFACTS}/logs/${node}"
+    mkdir -p ${dir}
+    ssh-to-node "${node}" "${jump_node}" "sudo journalctl --output=short-precise -k" > "${dir}/kern.log" || true
+    ssh-to-node "${node}" "${jump_node}" "sudo journalctl --output=short-precise" > "${dir}/systemd.log" || true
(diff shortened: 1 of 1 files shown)


hack/ci/e2e-conformance.sh
```diff
@@ -69,6 +71,34 @@ dump-logs() {
...
+function ssh-to-node() {
+  local node="$1"
+  local jump="$2"
+  local cmd="$3"
+
+  ssh_key_pem="/tmp/${SSH_KEY_NAME}.pem"
+  scp -i $ssh_key_pem $ssh_key_pem "ubuntu@${jump}:$ssh_key_pem"
+  ssh_params="-o LogLevel=quiet -o ConnectTimeout=30 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
+  ssh $ssh_params -i $ssh_key_pem \
+    -o "ProxyCommand ssh $ssh_params -W %h:%p -i $ssh_key_pem ubuntu@${jump}" \
```
ncdc: Can you use `ssh -i $ssh_key_pem -J ubuntu@${jump} ubuntu@${node}` instead? I think you can avoid having to copy the pem over, or if not, you could use a temporary `ssh-agent` to do it.

dims: the `-J ubuntu@${jump} ubuntu@${node}` didn't work out of the box. this worked when i tried by hand. will have to look into `ssh-agent` as well.

ncdc: Sounds good. I use `-J` all the time, albeit with my key loaded into `ssh-agent`.

dims: right, i need to find a balance since this script can be run by a developer as well. this path of creating a new key and wiping it off seemed a better approach.

dims:
we have logs! https://gcsweb.k8s.io/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/1338/pull-cluster-api-provider-aws-make-conformance/1192848052994445312/artifacts/logs/10.0.0.125/

ncdc:
@dims you're the best!

hack/ci/e2e-conformance.sh
detiber: Would it be worth it to leverage the aws systems manager run command instead? https://docs.aws.amazon.com/systems-manager/latest/userguide/execute-remote-commands.html

That said, I'm good with that as a followup/longerterm approach.

dims: @detiber yes, we could explore other options

PR #1340, opened 2019-11-08
:running: Update project maintainers and reviewers

vincepri:

Waiting for a few other folks to see if they want to step up.
/hold

### DIFF
--- a/OWNERS
+++ b/OWNERS
@@ -2,7 +2,6 @@
 
 approvers:
   - sig-cluster-lifecycle-leads
-  - sig-aws-leads
   - cluster-api-admins
   - cluster-api-maintainers
   - cluster-api-aws-maintainers
@@ -11,3 +10,8 @@ reviewers:
   - cluster-api-maintainers
   - cluster-api-aws-maintainers
   - cluster-api-aws-reviewers
+
+emeritus_approvers:
+  - davidewatson
+  - ingvagabund
+  - enxebre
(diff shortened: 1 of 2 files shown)


vincepri:
Got confirmation from all parties that this change is ok

ncdc:
Do you want to merge this now, or keep waiting?

detiber:
> Do you want to merge this now, or keep waiting?

Have we given this sufficient time/visibility for community review? I don't believe it was discussed at the weekly meeting last week, and I don't think we made an announcement on the mailing list.

ncdc:
I don't know (hence the question). It looks like it's just about promoting @rudoi from reviewer to maintainer, and adding @randomvariable as a maintainer (the removals have all been acked, according to @vincepri).

vincepri:
I've confirmed with all the parties, I'm ok waiting but next week is also KubeCon. I'd rather just merge it if we have consensus.

detiber:
> I've confirmed with all the parties, I'm ok waiting but next week is also KubeCon. I'd rather just merge it if we have consensus.

Confirming with all parties is fine, but without broader awareness in the community can we truly say we have consensus (or rather lazy consensus, since that is our standard)?

We can't expect all community members to watch all issues for the repo.

vincepri:
@detiber I'm ok either way, I'll leave it up to you.

rudoi:
I think we should give it the full week. I can't think of any real benefit to merging early, even with the conf next week.

rudoi:
what about now? :D

ncdc:
This has been open for 18 days. I'm saying we have lazy consensus.
/lgtm

PR #1341, opened 2019-11-08
:bug: check for subnet nil pointer when register instance to apiserver elb

jzhoucliqr:
release:  v0.4.4

Check for subnet nil pointer when register instance to apiserver elb. 

This fix the case that if AWSCluster.Spec.NetworkSpec somehow loses subnetID information, machine controller will wait for cluster controller to finish reconcileNetwork before trying to attach control plane to ELB.  

fix: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1342

### DIFF
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -67,7 +67,8 @@ func (s *Service) ReconcileLoadbalancers() error {
 	// Reconciliate the subnets from the spec and the ones currently attached to the load balancer.
 	if len(apiELB.SubnetIDs) != len(spec.SubnetIDs) {
 		_, err := s.scope.ELB.AttachLoadBalancerToSubnets(&elb.AttachLoadBalancerToSubnetsInput{
-			Subnets: aws.StringSlice(spec.SubnetIDs),
+			LoadBalancerName: &apiELB.Name,
+			Subnets:          aws.StringSlice(spec.SubnetIDs),
 		})
 		if err != nil {
 			return errors.Wrapf(err, "failed to attach apiserver load balancer %q to subnets", apiELB.Name)
@@ -144,7 +145,11 @@ func (s *Service) RegisterInstanceWithAPIServerELB(i *infrav1.Instance) error {
 	}
 
 	// Validate that the subnets associated with the load balancer has the instance AZ.
-	instanceAZ := s.scope.Subnets().FindByID(i.SubnetID).AvailabilityZone
+	subnet := s.scope.Subnets().FindByID(i.SubnetID)
+	if subnet == nil {
+		return errors.Errorf("failed to attach load balancer subnets, could not find subnet %q description in AWSCluster", i.SubnetID)
+	}
+	instanceAZ := subnet.AvailabilityZone
 	found := false
 	for _, subnetID := range out.SubnetIDs {
 		if subnet := s.scope.Subnets().FindByID(subnetID); subnet != nil && instanceAZ == subnet.AvailabilityZone {


ncdc:
@jzhoucliqr thanks for the PR. Would you please file an issue describing the problem and mark this PR as fixing it? Thanks!

jzhoucliqr:
sure @ncdc . Actually although this fix is anyway needed to check for nil pointer, I'm still trying to figure out the root cause why it goes into such condition. 

Based on tests I have done so far, the subnetID info that's populated into AWSCluster.Spec.NetworkSpec is somehow overwritten during the process, then when control plane comes up, machine controller try to attach ELB, then AWSCluster.Spec.NetworkSpec.Subnets missing all the subnetIDs, which cause issue. 

Will update once I figure out why it's got overwritten .

jzhoucliqr:
Figured out that the subnetID overwritten is because of a bug in my app. 

But I think this check for nil pointer is still useful, wdyt ?

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -144,7 +145,11 @@ func (s *Service) RegisterInstanceWithAPIServerELB(i *infrav1.Instance) error {
 	}
 
 	// Validate that the subnets associated with the load balancer has the instance AZ.
-	instanceAZ := s.scope.Subnets().FindByID(i.SubnetID).AvailabilityZone
+	isubnet := s.scope.Subnets().FindByID(i.SubnetID)
```
vincepri:
```suggestion
	subnet := s.scope.Subnets().FindByID(i.SubnetID)
```

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -144,7 +145,11 @@ func (s *Service) RegisterInstanceWithAPIServerELB(i *infrav1.Instance) error {
...
+	if isubnet == nil {
```
vincepri: Consider simplifying this code with

```
if subnet := s.scope.Subnets().FindByID(i.SubnetID); subnet == nil {
```

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -144,7 +145,11 @@ func (s *Service) RegisterInstanceWithAPIServerELB(i *infrav1.Instance) error {
...
+		return errors.Errorf("subnet FindByID failed, maybe because cluster controller haven't populate subnetID info, wait and try, subnetID: %s", i.SubnetID)
```
vincepri:
```suggestion
		return errors.Errorf("failed to attach load balancer subnets, could not find subnet %q description in AWSCluster", i.SubnetID)
```

pkg/cloud/services/elb/loadbalancer.go
jzhoucliqr: Need to get AZ from subnet, so keep it out of if condition. Changed 'isubnet' to 'subnet'.

detiber:
This change seems fine to me, can you squash the commits and Update the commit and PR descriptions to match the current changeset?

jzhoucliqr:
@detiber Done. Thanks

detiber:
Milestone says v0.5.0, but this was merged against release-0.4, does this need to be forward ported back to the master branch?

ncdc:
@jzhoucliqr can you open a PR against master for this?

jzhoucliqr:
Sorry guys, should've done this earlier.  See https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1386

PR #1348, opened 2019-11-14
🏃 Verify LoadBalancer dyanamically provisioned on service creation and deleted after cluster deletion

sirao:
Existing E2E tests, doesn't cover verification of dynamically provisioned Load Balancer deletion after cluster removal
This PR covers the 
1. Deploy a Load balancer service on infra cluster
2. Verify Load balancer provisioned
3. Verify Load balancer deleted on deleting the cluster
4. Deploy statefulset on infra cluster
5. Verify volumes are dynamically provisioned
6. Verify dynamically provisioned volumes retained on cluster deletion
Executed the tests locally and all the tests succeeded.

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -24,13 +24,21 @@ import (
 	"io/ioutil"
 	"os/exec"
 	"path"
+	"strings"
 	"time"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/elb"
+	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
+	storagev1 "k8s.io/api/storage/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	apimachinerytypes "k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
@@ -54,7 +62,25 @@ const (
 	scaleDownReplicas = 0
 )
 
+type statefulSetInfo struct {
+	name                      string
+	namespace                 string
+	replicas                  int32
+	selector                  map[string]string
+	storageClassName          string
+	volumeName                string
+	svcName                   string
+	svcPort                   int32
+	svcPortName               string
+	containerName             string
+	containerImage            string
+	containerPort             int32
+	podTerminationGracePeriod int64
+	volMountPath              string
+}
+
 var _ = Describe("functional tests", func() {
+
 	var (
 		namespace               string
 		clusterName             string
@@ -66,8 +92,10 @@ var _ = Describe("functional tests", func() {
 		machineDeploymentName   string
 		awsMachineTemplateName  string
 		testTmpDir              string
+		instanceType            string
 		initialReplicas         int32
 	)
+
 	BeforeEach(func() {
 		var err error
 		testTmpDir, err = ioutil.TempDir(suiteTmpDir, "aws-test")
@@ -85,34 +113,18 @@ var _ = Describe("functional tests", func() {
 		machineDeploymentName = "test-capa-md" + util.RandomString(6)
 		awsMachineTemplateName = "test-infra-capa-mt" + util.RandomString(6)
 		initialReplicas = 2
+		instanceType = "t3.large"
 	})
 
 	Describe("workload cluster lifecycle", func() {
 		It("It should be creatable and deletable", func() {
-			By("Creating an AWSCluster")
-			makeAWSCluster(namespace, awsClusterName)
-
-			By("Creating a Cluster")
-			makeCluster(namespace, clusterName, awsClusterName)
-
-			By("Ensuring Cluster Infrastructure Reports as Ready")
-			waitForClusterInfrastructureReady(namespace, clusterName)
-
-			By("Creating the initial Control Plane Machine")
-			awsMachineName := cpAWSMachinePrefix + "-0"
-			bootstrapConfigName := cpBootstrapConfigPrefix + "-0"
-			machineName := cpMachinePrefix + "-0"
-			instanceType := "t3.large"
-			createInitialControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName, instanceType)
-
-			By("Deploying CNI to created Cluster")
-			deployCNI(testTmpDir, namespace, clusterName, *cniManifests)
-			waitForMachineNodeReady(namespace, machineName)
+			By("Creating a cluster with single control plane")
+			makeSingleControlPlaneCluster(namespace, clusterName, awsClusterName, cpAWSMachinePrefix, cpBootstrapConfigPrefix, cpMachinePrefix, instanceType, testTmpDir)
 
 			By("Creating the second Control Plane Machine")
-			awsMachineName = cpAWSMachinePrefix + "-1"
-			bootstrapConfigName = cpBootstrapConfigPrefix + "-1"
-			machineName = cpMachinePrefix + "-1"
+			awsMachineName := cpAWSMachinePrefix + "-1"
+			bootstrapConfigName := cpBootstrapConfigPrefix + "-1"
+			machineName := cpMachinePrefix + "-1"
 			createAdditionalControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName, instanceType)
 
 			By("Creating the third Control Plane Machine")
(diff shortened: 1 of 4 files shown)


test/e2e/aws_test.go
```diff
@@ -304,18 +411,7 @@ func makeAWSMachineTemplate(namespace, name string) {
...
-			Namespace: namespace,
-			Name:      clusterName,
-		},
-	}
-	kubeConfigData, err := kubeconfig.FromSecret(kindClient, cluster)
-	Expect(err).NotTo(HaveOccurred())
-	kubeConfigPath := path.Join(tmpDir, clusterName+".kubeconfig")
-	Expect(ioutil.WriteFile(kubeConfigPath, kubeConfigData, 0640)).To(Succeed())
-
+func deployCNI(kubeConfigPath, manifestPath string) {
```
bhargavmad: instead of sending `kubeConfigPath` as an arg, call the `createClusterKubeConfig` from here.

Something like:
```
func deployCNI(tmpDir, namespace, clusterName, manifestPath string) {
     createClusterKubeConfig(tmpDir, namespace, clusterName)
     .
     .

```

test/e2e/aws_test.go
```diff
@@ -132,13 +139,113 @@ var _ = Describe("functional tests", func() {
 			By("Scale the MachineDeployment up again")
 			scaleMachineDeployment(namespace, machineDeploymentName, scaleDownReplicas, initialReplicas)
 
+			By("Creating the LB service")
+			sess := getSession()
+			elbClient := elb.New(sess)
```
bhargavmad: can be moved into `verifyElbExists`

test/e2e/aws_test.go
```diff
@@ -132,13 +139,113 @@ var _ = Describe("functional tests", func() {
...
+				return false, err
+			}
+			if svcCreated.Name == svcName {
+				fmt.Fprintf(GinkgoWriter, "Found the service: %s\n", svcName)
+				return true, nil
+			}
+			return false, nil
+		},
+		5*time.Minute, 15*time.Second,
+	).Should(BeTrue())
```
bhargavmad: Not required. If the `Create` fails, it will exit at L191

test/e2e/aws_test.go
```diff
@@ -132,13 +139,113 @@ var _ = Describe("functional tests", func() {
...
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace: namespace,
+			Name:      clusterName,
+		},
+	}
+	kubeConfigData, err := kubeconfig.FromSecret(kindClient, cluster)
+	Expect(err).NotTo(HaveOccurred())
+
+	restConfig, err := clientcmd.RESTConfigFromKubeConfig(kubeConfigData)
+	Expect(err).NotTo(HaveOccurred())
```
bhargavmad: using the `kubeConfigPath` here will remove the effort of getting the secret again

test/e2e/aws_test.go
```diff
@@ -132,13 +139,113 @@ var _ = Describe("functional tests", func() {
...
+
+	svcClient, err := crclient.New(restConfig, crclient.Options{})
```
bhargavmad: can use a better naming here. Since it is actually the k8s client.

test/e2e/aws_test.go
```diff
@@ -132,13 +139,113 @@ var _ = Describe("functional tests", func() {
...
 			By("Deleting the Cluster")
 			deleteCluster(namespace, clusterName)
 
+			By("Checking provisioned LB deleted")
+			verifyElbExists(elbClient, lbServiceName, false)
 		})
 	})
 })
 
+func createLBService(namespace, clusterName, svcName string) {
```
bhargavmad: Suggestion: maybe at a later stage it will useful to create a generic `createService` function which takes a `Service.Spec`

test/e2e/aws_test.go
```diff
@@ -132,13 +139,113 @@ var _ = Describe("functional tests", func() {
...
+				fmt.Fprintf(GinkgoWriter, "Found the service: %s\n", svcName)
+				return true, nil
+			}
+			return false, nil
+		},
+		5*time.Minute, 15*time.Second,
+	).Should(BeTrue())
+}
+
+func verifyElbExists(elbClient *elb.ELB, svcName string, exists bool) {
```
bhargavmad: Suggestion: It is better to get the external ip of the service created and then use it to verify.

sirao: ELBs can be describe using ELB name (in case of elb) or ELB ARN name in case of elbv2). It is not possible to list them using IP. Hence, need to describe all the ELBs and iterate over them to find the required one created.

bhargavmad:
```
"LoadBalancerName": "a067b07bbec5f4339a403992f01e5df1",
"DNSName": "a067b07bbec5f4339a403992f01e5df1-1901108057.us-east-2.elb.amazonaws.com",
```
ELB name can be extracted by removing the last `xxxxxxxxxx.us-east-2.elb.amazonaws.com` from the DNS name

test/e2e/aws_test.go
```diff
@@ -42,6 +43,7 @@ import (
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/kubeconfig"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"strings"
```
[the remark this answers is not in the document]
sirao: Moved up

test/e2e/aws_test.go
sirao: Moved inside verifyElbExists method

test/e2e/aws_test.go
sirao: Removed it

test/e2e/aws_test.go
sirao: Renamed it

test/e2e/aws_test.go
sirao: Rather than getting the external IP,  added code to fetch the ingress host name from the service and then extracted ELB name out of it which later used to check the existence.

test/e2e/aws_test.go
sirao: Added a new method which returns both configPath and cluster k8s client which were used across.

test/e2e/aws_test.go
sirao: Even though creating the service Spec is the major code addition than creating the service, added a generic method which accepts spec and labels

test/e2e/aws_test.go
sirao: I have added method to fetch kubeConfig and k8sclient which were can be used across. That will address this.

test/e2e/aws_test.go
```diff
@@ -102,11 +153,11 @@ var _ = Describe("functional tests", func() {
 			awsMachineName := cpAWSMachinePrefix + "-0"
 			bootstrapConfigName := cpBootstrapConfigPrefix + "-0"
 			machineName := cpMachinePrefix + "-0"
-			instanceType := "t3.large"
-			createInitialControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName, instanceType)
+			createInitialControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName)
```
bhargavmad: keep the `instanceType` as an argument

test/e2e/aws_test.go
```diff
@@ -133,13 +184,275 @@ var _ = Describe("functional tests", func() {
 			By("Scale the MachineDeployment up again")
 			scaleMachineDeployment(namespace, machineDeploymentName, scaleDownReplicas, initialReplicas)
 
+			By("Creating the LB service")
+			elbName := createLBService(metav1.NamespaceDefault, lbServiceName, clusterK8sClient)
+			verifyElbExists(elbName, true)
+
+			By("Deploying StatefulSet on infra")
```
bhargavmad: rephrase

test/e2e/aws_test.go
```diff
@@ -133,13 +184,275 @@ var _ = Describe("functional tests", func() {
 			By("Scale the MachineDeployment up again")
 			scaleMachineDeployment(namespace, machineDeploymentName, scaleDownReplicas, initialReplicas)
 
+			By("Creating the LB service")
```
bhargavmad: Not sure if this should be a part of the `workload cluster lifecycle`

test/e2e/aws_test.go
bhargavmad: Not sure if this should be a part of the `workload cluster lifecycle`

test/e2e/aws_test.go
```diff
@@ -133,13 +184,275 @@ var _ = Describe("functional tests", func() {
...
+	elbName := ""
+	if lbs := len(svcCreated.Status.LoadBalancer.Ingress); lbs > 0 {
+		ingressHostname := svcCreated.Status.LoadBalancer.Ingress[0].Hostname
+		elbName = strings.Split(ingressHostname, "-")[0]
+	}
+	fmt.Fprintf(GinkgoWriter, "Created Load Balancer service and ELB name is: %s\n", elbName)
+	return elbName
+}
+
+func verifyElbExists(elbName string, exists bool) {
```
bhargavmad: Suggestion:
It can be modified as `ifElbExist(elbName string) bool` which returns true when elb exists otherwise false

sirao: It can be done. Then we need to do the assertion of the value returned in the "It". Even though it can be done inside "It", to follow the previously added tests convention, used this approach where assertion will happen inside the called function.

test/e2e/aws_test.go
sirao: The verify, dynamically provisioned ELB deletion and EBS Volumes retention after cluster deletion the ELB and EBS should be created after cluster creation and before cluster deletion. Hence, did it here before cluster deletion.

test/e2e/aws_test.go
sirao: The verify, dynamically provisioned ELB deletion and EBS Volumes retention after cluster deletion the ELB and EBS should be created after cluster creation and before cluster deletion. Hence, did it here before cluster deletion.

test/e2e/aws_test.go
```diff
@@ -54,6 +62,24 @@ const (
...
+	svcName                   string
+	svcPort                   int32
+	svcPortName               string
+	containerName             string
+	containerImage            string
+	containerPort             int32
+	podTerminationGracePeriod int64
+	volMountPath              string
+}
+
```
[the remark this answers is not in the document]
sirao: Debatable. We can either hardcode magic values while creating the PodTemplateSpec/VolumeClaimTemplate/Service or pass the values as arguments to the functions creating the config objects. To avoid too many arguments to function call and to repetitively use the values created the struct and passed it across function calls.

test/e2e/aws_test.go
```diff
@@ -85,6 +116,26 @@ var _ = Describe("functional tests", func() {
...
+			volumeSize:                "1Gi",
+			svcName:                   "nginx-svc",
+			svcPort:                   int32(80),
+			svcPortName:               "nginx-web",
+			containerName:             "nginx",
+			containerImage:            "k8s.gcr.io/nginx-slim:0.8",
+			containerPort:             int32(80),
+			podTerminationGracePeriod: int64(30),
+			volMountPath:              "/usr/share/nginx/html",
+		}
```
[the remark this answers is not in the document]
sirao: May I know the reason?

test/e2e/aws_test.go
```diff
@@ -67,6 +93,11 @@ var _ = Describe("functional tests", func() {
 		awsMachineTemplateName  string
 		testTmpDir              string
 		initialReplicas         int32
+		clusterKubeConfigPath   string
+		lbServiceName           string
+		clusterK8sClient        crclient.Client
+		nginxSelector           map[string]string
+		nginxStatefulsetInfo    statefulSetInfo
```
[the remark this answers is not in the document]
sirao: As  the values are used in multiple places, these are required.

test/e2e/aws_test.go
sirao: Will do the change.

test/e2e/aws_test.go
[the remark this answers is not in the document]
bhargavmad: `BeforeEach` executes before every spec. When more specs gets added, it may not be suited to initialize all the variables again.

test/e2e/aws_test.go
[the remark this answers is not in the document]
bhargavmad: i believe name and namespace are the only fields we would require to pass around and other fields can be hardcoded as they will not be of much use for other test cases.

test/e2e/aws_test.go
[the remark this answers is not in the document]
bhargavmad: If agreed upon the above comment of not having a struct, this can be avoided.

randomvariable:
Can we have a status update - there's a lots of unresolved comments, makes it hard to figure out if this is ready or not.

test/e2e/aws_test.go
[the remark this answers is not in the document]
sirao: Some of the fields are used it more than one place. Hence, passing them as struct which is readable.

test/e2e/aws_test.go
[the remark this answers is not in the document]
sirao: Removed them as part of Describe variables.

test/e2e/aws_test.go
sirao: made it as argument

test/e2e/aws_test.go
sirao: Made this part of another scenario

test/e2e/aws_test.go
sirao: Made it part of another scenario

sirao: Did change

test/e2e/aws_test.go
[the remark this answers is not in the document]
sirao: Removed it from BeforeEach and made it part of the scenario

test/e2e/aws_test.go
```diff
@@ -135,11 +147,393 @@ var _ = Describe("functional tests", func() {
...
+	})
+
+	Describe("MachineDeployment with invalid subnet ID", func(){
+		It("It should be creatable and deletable", func(){
+			By("Creating a cluster with single control plane")
+			makeSingleControlPlaneCluster(namespace, clusterName, awsClusterName, cpAWSMachinePrefix, cpBootstrapConfigPrefix, cpMachinePrefix, instanceType, testTmpDir)
+
+			By("Creating Machine Deployment with invalid subnet ID")
+			subnetId := "notcreated"
+			createPendingMachineDeployment(namespace, clusterName, machineDeploymentName, awsMachineTemplateName, mdBootstrapConfig, instanceType,nil, &subnetId, initialReplicas)
```
bhargavmad: Use `createMachineDeployment` func

test/e2e/aws_test.go
```diff
@@ -135,11 +147,393 @@ var _ = Describe("functional tests", func() {
...
+
+			By("Ensuring MachineDeployment is not in running state")
+			capaLogs := retrieveCapaLogs()
+			Expect(waitForMachineDeploymentRunning(namespace, machineDeploymentName)).To(BeFalse())
+			Expect(strings.Contains(capaLogs, fmt.Sprintf("InvalidSubnetID.NotFound: The subnet ID '%s' does not exist", subnetId)))
```
bhargavmad: Not sure if this is the right way to verify the error

test/e2e/aws_test.go
```diff
@@ -135,11 +147,393 @@ var _ = Describe("functional tests", func() {
...
+
+			By("Deleting the cluster")
+			deleteCluster(namespace, clusterName)
+		})
+	})
+
+	Describe("MachineDeployment with AZ with no existing subnet", func(){
```
bhargavmad: Can be clubbed with the above spec

test/e2e/aws_test.go
detiber: It might be better to watch for events and find the event that is directly related to the MD created instead.

Might be good to move this test case into a separate PR until that is worked out.

test/e2e/aws_test.go
detiber: +1 to combining with previous spec

sirao: clubbed the two scnearios

test/e2e/aws_test.go
sirao: Added code to query the events and verify the error event exists.
Found issue https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1425

test/e2e/aws_test.go
sirao: Here i am creating machine deployments with invalid configuration where the MachineDeployment will not come to running stat. Hence, can't use `createMachineDeployment` which will check MachineDeployment running or not and fail if not running.

bhargavmad: Ok. My bad. 👍

detiber:
/lgtm
/approve
/hold
hold for e2e validation

PR #1349, opened 2019-11-14
[wip] :sparkles: Add tests for authentication, make local recorders for scopes

### LINKED ISSUE #1343: Must raise a warning event on machine/cluster when the AWS session creation fails due to missing credentials.
/kind bug

Try to delete the cluster with an invalid AWS credentials (after the cluster is successfully created). 

The cluster deletion fail. There must be a warning event on the machine/cluster that failed to delete

The AWS session creation itself fail in this case and no events were thrown. We must generate an appropriate event corresponding to the object that failed to create. 

**Environment:**

randomvariable:
Scope structs modified to accept replacement AWS sessions, to allow a fake HTTP server to be used for the AWS client, and exercise the credential provider chain code.

Fixes #1343

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -78,6 +78,7 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		Logger:     log,
 		Cluster:    cluster,
 		AWSCluster: awsCluster,
+		Recorder:   r.Recorder,
 	})
 	if err != nil {
 		return reconcile.Result{}, errors.Errorf("failed to create scope: %+v", err)
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -120,6 +120,7 @@ func (r *AWSMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		Logger:     logger,
 		Cluster:    cluster,
 		AWSCluster: awsCluster,
+		Recorder:   r.Recorder,
 	})
 	if err != nil {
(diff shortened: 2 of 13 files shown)


controllers/awscluster_controller.go
```diff
@@ -78,6 +78,7 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		Logger:     log,
 		Cluster:    cluster,
 		AWSCluster: awsCluster,
+		Recorder:   r.Recorder,
```
vincepri: We've usually suggested to use the global recorder instead of a local one, https://github.com/kubernetes-sigs/cluster-api/blob/master/util/record/recorder.go

ncdc: Let's open an issue if we don't have one to get rid of the local one?

vincepri: I pointed it out because it seems this PR is adding the local one, which we probably don't want to

ncdc: Roger... multitasking fail

randomvariable: I do however need to a fresh recorder for each test case. In fact, I was extending what was already there, threading the recorder from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1349/files#diff-a4c89adfffbff9126a6d9d4d96531799L101 into where it needs to go to be able to test the receive of the credential provider chain error.

If there's another refactor that works better, let me know.

randomvariable:
/hold
pending discussion about testing

randomvariable:
No resolution found.
/close

PR #1353, opened 2019-11-15
:sparkles: Improves e2e tests by using version flags

bhargavmad:
**What this PR does**:
* Adds flags to set the kubernetes version of workload clusters to test up on.
* Adds flags to set sonobuoy version to run conformance

**Why we need it**
Helps to run conformance/functional tests on latest kuberentes versions easily


detiber:
/approve
/lgtm
/hold
holding for completion of e2e test run

We'll want to forward port this to the master branch as well.

PR #1354, opened 2019-11-15
:sparkles: Adds more e2e functional tests 

bhargavmad:

1. Adds Create/Delete multiple workload clusters test scenario

- clusters in different namespaces
- clusters within same namespace

2. Adds e2e tests for delete machine lifecycle
- delete a machine using kindclient
- wait for the MachineDeployment to be reconciled

3. Adds e2e tests for multi AZ deployment
- Deploy ControlPlane and MachineDeployment in new AZ
- Deploy ControlPlane and MachineDeployment in new AZ using subnet-id

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -53,8 +53,11 @@ import (
 )
 
 var (
-	cniManifests  = capiFlag.DefineOrLookupStringFlag("cniManifests", "https://docs.projectcalico.org/v3.9/manifests/calico.yaml", "URL to CNI manifests to load")
-	kubectlBinary = capiFlag.DefineOrLookupStringFlag("kubectlBinary", "kubectl", "path to the kubectl binary")
+	cniManifests      = capiFlag.DefineOrLookupStringFlag("cniManifests", "https://docs.projectcalico.org/v3.9/manifests/calico.yaml", "URL to CNI manifests to load")
+	kubectlBinary     = capiFlag.DefineOrLookupStringFlag("kubectlBinary", "kubectl", "path to the kubectl binary")
+	availabilityZones []*string
+	privateCIDRs      = [3]string{"10.0.0.0/24", "10.0.2.0/24", "10.0.4.0/24"}
+	publicCIDRs       = [3]string{"10.0.1.0/24", "10.0.3.0/24", "10.0.5.0/24"}
 )
 
 const (
@@ -119,22 +122,22 @@ var _ = Describe("functional tests", func() {
 	Describe("workload cluster lifecycle", func() {
 		It("It should be creatable and deletable", func() {
 			By("Creating a cluster with single control plane")
-			makeSingleControlPlaneCluster(namespace, clusterName, awsClusterName, cpAWSMachinePrefix, cpBootstrapConfigPrefix, cpMachinePrefix, instanceType, testTmpDir)
+			makeSingleControlPlaneCluster(namespace, clusterName, awsClusterName, cpAWSMachinePrefix, cpBootstrapConfigPrefix, cpMachinePrefix, instanceType, testTmpDir, false)
 
 			By("Creating the second Control Plane Machine")
 			awsMachineName := cpAWSMachinePrefix + "-1"
 			bootstrapConfigName := cpBootstrapConfigPrefix + "-1"
 			machineName := cpMachinePrefix + "-1"
-			createAdditionalControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName, instanceType)
+			createAdditionalControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName, instanceType, nil, nil)
 
 			By("Creating the third Control Plane Machine")
 			awsMachineName = cpAWSMachinePrefix + "-2"
 			bootstrapConfigName = cpBootstrapConfigPrefix + "-2"
 			machineName = cpMachinePrefix + "-2"
-			createAdditionalControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName, instanceType)
+			createAdditionalControlPlaneMachine(namespace, clusterName, machineName, awsMachineName, bootstrapConfigName, instanceType, nil, nil)
 
 			By("Creating the MachineDeployment")
-			createMachineDeployment(namespace, clusterName, machineDeploymentName, awsMachineTemplateName, mdBootstrapConfig, instanceType, initialReplicas)
+			createMachineDeployment(namespace, clusterName, machineDeploymentName, awsMachineTemplateName, mdBootstrapConfig, instanceType, initialReplicas, nil, nil)
 
 			By("Scale the MachineDeployment up")
 			scaleMachineDeployment(namespace, machineDeploymentName, initialReplicas, scaleUpReplicas)
@@ -154,10 +157,10 @@ var _ = Describe("functional tests", func() {
 		lbServiceName := "test-svc-" + util.RandomString(6)
 		It("It should create and delete Load Balancer", func() {
 			By("Creating a cluster with single control plane")
-			clusterK8sClient := makeSingleControlPlaneCluster(namespace, clusterName, awsClusterName, cpAWSMachinePrefix, cpBootstrapConfigPrefix, cpMachinePrefix, instanceType, testTmpDir)
+			clusterK8sClient := makeSingleControlPlaneCluster(namespace, clusterName, awsClusterName, cpAWSMachinePrefix, cpBootstrapConfigPrefix, cpMachinePrefix, instanceType, testTmpDir, false)
 
 			By("Creating the MachineDeployment")
-			createMachineDeployment(namespace, clusterName, machineDeploymentName, awsMachineTemplateName, mdBootstrapConfig, instanceType, initialReplicas)
+			createMachineDeployment(namespace, clusterName, machineDeploymentName, awsMachineTemplateName, mdBootstrapConfig, instanceType, initialReplicas, nil, nil)
 
 			By("Creating the LB service")
 			elbName := createLBService(metav1.NamespaceDefault, lbServiceName, clusterK8sClient)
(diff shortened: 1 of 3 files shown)


test/e2e/aws_test.go
```diff
@@ -138,8 +123,87 @@ var _ = Describe("functional tests", func() {
...
+			It("should create first cluster", func() {
+				ns1 = namespace
+				clName1 = clusterName
+				By("Creating first cluster with single control plane")
+				makeSingleControlPlaneCluster(namespace, clusterName, awsClusterName, cpAWSMachinePrefix, cpBootstrapConfigPrefix, cpMachinePrefix, instanceType, testTmpDir)
+
+				By("Creating the MachineDeployment")
+				createMachineDeployment(namespace, clusterName, machineDeploymentName, awsMachineTemplateName, mdBootstrapConfig, instanceType, initialReplicas)
+			})
+			It("should create second cluster in a different namespace", func() {
```
bhargavmad: Making a new spec keeps the code clean but on the down side, increases the num of specs which may not be acceptable. Will be happy to hear any suggestions.

test/e2e/aws_test.go
```diff
@@ -138,8 +127,297 @@ var _ = Describe("functional tests", func() {
...
+	}
+	result, err := ec2c.DescribeInstances(instanceInput)
+	Expect(err).NotTo(HaveOccurred())
+	Expect(len(result.Reservations)).To(Equal(int(numOfInstances)))
+}
+
+func getAvailabilityZone(ec2c *ec2.EC2) *string {
+	azs, err := ec2c.DescribeAvailabilityZones(nil)
+	Expect(err).NotTo(HaveOccurred())
+	return azs.AvailabilityZones[1].ZoneName
```
gsadhani: Why always the second AZ? There can be some regions with only one AZ, better check the size of the AZ list before trying to access the second one.

test/e2e/aws_test.go
```diff
@@ -138,8 +127,297 @@ var _ = Describe("functional tests", func() {
...
+}
+
+func configureAWSresources(ec2c *ec2.EC2, clusterName string, azName *string) {
```
gsadhani: Will the resources created here get cleaned up with cluster delete?

test/e2e/aws_test.go
bhargavmad: yeah i did think of this before. But the tests are usually run on `us-west-2` so i just took the second one.

test/e2e/aws_test.go
bhargavmad: yes. they will be cleaned.

test/e2e/aws_test.go
detiber: I think the additional specs are fine here, it will give us room to optimize runtime by running them in parallel in the future.

test/e2e/aws_test.go
detiber: As discussed in the meeting the other day, we should leverage the ability to create multiple managed subnets with Cluster API here rather than manually adding additional subnets.

`AWSCluster.Spec.NetworkSpec.Subnets = [{AvailabilityZone: AZ1_NAME},{AvailabilityZone: AZ2_NAME},...]` should do the trick

bhargavmad: Will do.

randomvariable:
Please resolve questions if they have been addressed, thanks.

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
-		clusterName             string
-		awsClusterName          string
-		cpMachinePrefix         string
-		cpAWSMachinePrefix      string
-		cpBootstrapConfigPrefix string
-		mdBootstrapConfig       string
-		machineDeploymentName   string
-		awsMachineTemplateName  string
-		testTmpDir              string
-		initialReplicas         int32
```
detiber: I'd avoid this change for now to avoid potential merge conflicts with other PRs for now, it might also be good to separate this type of change as an independent PR.

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
 	})
+
+	Describe("multiple workload clusters", func() {
+		Context("in different namespaces", func() {
+			var ns1, clName1, ns2, clName2 string
+			It("should create first cluster", func() {
+				ns1 = setup.namespace
+				clName1 = setup.clusterName
+				By("Creating first cluster with single control plane")
+				makeSingleControlPlaneCluster(setup)
```
detiber: We seem to be doing this step in all of the different jobs now. I'm wondering if this should be moved to a BeforeEach/AfterEach instead, if we have any jobs that would require a different setup, they could be moved to a separate test file.

Not blocking for this PR, though.

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+
+				By("Creating the MachineDeployment")
+				createMachineDeployment(setup)
```
detiber: Are the MachineDeployments necessary here?

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+			})
+			It("should create second cluster in a different namespace", func() {
+				ns2 = setup.namespace
+				clName2 = setup.clusterName
+				By("Creating second cluster with single control plane")
+				makeSingleControlPlaneCluster(setup)
+
+				By("Creating the MachineDeployment")
+				createMachineDeployment(setup)
```
detiber: Are the MachineDeployments necessary here?

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+		Context("in same namespace", func() {
+			var ns, clName1, clName2 string
+			It("should create first cluster", func() {
+				ns = setup.namespace
+				clName1 = setup.clusterName
+				By("Creating first cluster with single control plane")
+				makeSingleControlPlaneCluster(setup)
+
+				By("Creating the MachineDeployment")
+				createMachineDeployment(setup)
```
detiber: Are the MachineDeployments necessary here?

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+			})
+			It("should create second cluster in the same namespace", func() {
+				setup.namespace = ns
+				clName2 = setup.clusterName
+				By("Creating second cluster with single control plane")
+				makeSingleControlPlaneCluster(setup)
+
+				By("Creating the MachineDeployment")
+				createMachineDeployment(setup)
```
detiber: Are the MachineDeployments necessary here?

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+
+			It("should delete both clusters", func() {
+				By("Deleting the Clusters")
+				deleteCluster(ns, clName1)
+				deleteCluster(ns, clName2)
+			})
+		})
+	})
+
+	Describe("Delete machine lifecycle", func() {
```
detiber:
```suggestion
	Describe("MachineDeployment will replace a deleted Machine", func() {
```

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+			setup.availabilityZone = availabilityZones[1]
+			createAdditionalControlPlaneMachine(setup, machineName, awsMachineName, bootstrapConfigName)
+
+			By("Creating the MachineDeployment in second AZ")
+			setup.availabilityZone = availabilityZones[1]
+			createMachineDeployment(setup)
+
+			setup.subnetId = getSubnetId(privateCIDR)
+			By("Checking if the nodes are deployed in second AZ")
+			verfiyInstancesInSubnet(setup.initialReplicas+1, setup.subnetId)
```
detiber: It would probably be better to hold off on creating any MachineDeployments until all control plane members are brought up.

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+
+			By("Creating the third Control Plane Machine using SubnetID in second AZ")
```
detiber: The third control plane machine should likely be brought up in a third AZ.

test/e2e/aws_test.go
```diff
@@ -45,101 +47,285 @@ import (
...
+			awsMachineName = setup.cpAWSMachinePrefix + "-2"
+			bootstrapConfigName = setup.cpBootstrapConfigPrefix + "-2"
+			machineName = setup.cpMachinePrefix + "-2"
+			createAdditionalControlPlaneMachine(setup, machineName, awsMachineName, bootstrapConfigName)
+
+			By("Checking if the node is deployed in second AZ")
+			verfiyInstancesInSubnet(setup.initialReplicas+2, setup.subnetId)
+
+			By("Creating the MachineDeployment using SubnetID in second AZ")
```
detiber: Should this MachineDeployment be in a AZ that is different than the first MachineDeployment created?

test/e2e/aws_test.go
bhargavmad: Will do.

test/e2e/aws_test.go
bhargavmad: Sounds good. Will do it in a different PR

test/e2e/aws_test.go
bhargavmad: Not really. Will remove.

test/e2e/aws_test.go
bhargavmad: I was in an impression that there are commonly 2 AZs in AWS regions. Will make it 3. 👍

detiber: Most regions have 3 azs, there are a few exceptions (some with more, some with less): https://aws.amazon.com/about-aws/global-infrastructure/regions_az/

bhargavmad:
Rebased into a single commit
Opened a new PR to increase the timeout: https://github.com/kubernetes/test-infra/pull/15581

detiber:
/lgtm
/approve
/hold
hold for another e2e test run

detiber:
It looks like the e2e failures align with flakes in the post-submit jobs since #1348 merged, let's get this in and then we can continue to tweak the tests as needed to address any flakiness in the test signal.
/hold cancel
/lgtm

PR #1363, opened 2019-11-15
🐛 POC same cluster name different namespaces

chuckha:

/hold

This isn't the actual change we want, but it's proof enough to show that we need a way to identify a cluster beyond using just the name. With this one line change I was able to fix kubernetes-sigs/cluster-api#1554 bug.

Screen Shot 2019-11-15 at 3 14 16 PM

I think this raises a bigger point in how we identify clusters. I'm wondering if we should expose an `Identifier` method on a CAPI Cluster and then mandate that anytime you need to identify a cluster you use that method in order to account for same name different namespaces.

### DIFF
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -140,7 +140,7 @@ func (s *ClusterScope) SecurityGroups() map[infrav1.SecurityGroupRole]infrav1.Se
 
 // Name returns the cluster name.
 func (s *ClusterScope) Name() string {
-	return s.Cluster.Name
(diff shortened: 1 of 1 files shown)


detiber:
@chuckha we likely would also want to impact the ClusterName used in the KubeadmConfig for the Cluster as well as ensure that this flows down into the tags that we set for the cloud-provider integration as well.

randomvariable:
Is this still a thing?

chuckha:
@randomvariable yes, this is still a problem but we can close this.

/close

PR #1374, opened 2019-11-25
 🏃Add tests for testing cluster naming validation

sirao:
This PR adds tests to validate the cluster names

Added tests to verify cluster creation having different names
    1. name having capital letters
    2. name having underscore
    3. name having period
    4. name having length more than 32 characters
    5. name having only digits

### DIFF
--- a/scripts/ci-integration.sh
+++ b/scripts/ci-integration.sh
@@ -14,6 +14,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+################################################################################
+# usage: ci-integration.sh
+#  This program runs the integration tests.
+#
+# ENVIRONMENT VARIABLES
+#  JANITOR_ENABLED
+#    Set to 1 to run the aws-janitor command after running the integration tests.
+################################################################################
+
 set -o nounset
 set -o pipefail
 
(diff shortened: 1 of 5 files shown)


test/integration/integration_suite_test.go
```diff
@@ -16,16 +16,133 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package integration_test
+package integration
```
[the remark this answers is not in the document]
sirao: Renamed

test/integration/integration_suite_test.go
```diff
@@ -16,16 +16,133 @@ See the License for the specific language governing permissions and
...
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
+	"github.com/onsi/ginkgo/config"
+	"github.com/onsi/ginkgo/reporters"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/client"
+	"github.com/aws/aws-sdk-go/service/iam"
+	testcommons "sigs.k8s.io/cluster-api-provider-aws/test/commons"
```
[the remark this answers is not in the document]
sirao: Renamed as testutil

test/integration/integration_suite_test.go
```diff
@@ -16,16 +16,133 @@ See the License for the specific language governing permissions and
...
 import (
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path"
 	"testing"
 
 	. "github.com/onsi/ginkgo"
 	. "github.com/onsi/gomega"
+
```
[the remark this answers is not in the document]
sirao: removed new line

test/integration/integration_test.go
```diff
@@ -0,0 +1,72 @@
...
+	})
+
+	Describe("cluster name validation tests", func() {
+		It("cluster creation with name having capital letters should fail", func() {
+			awsClusterName := "capa-cluster-CAPS"
+			err := testcommons.MakeAWSCluster(namespace, awsClusterName, KeyPairName, kindClient)
+			Expect(err).Should(HaveOccurred())
+			fmt.Fprintf(GinkgoWriter, "Error received is : %s \n", err.Error())
+			err = testcommons.MakeCluster(namespace, awsClusterName, awsClusterName, kindClient)
+			Expect(err).Should(HaveOccurred())
```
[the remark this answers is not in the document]
sirao: Refactored this piece of code

test/integration/integration_test.go
```diff
@@ -0,0 +1,72 @@
...
+
+	BeforeEach(func() {
+		namespace = "test-namespace-" + util.RandomString(6)
+		testcommons.CreateNamespace(namespace, kindClient)
+	})
+
+	Describe("cluster name validation tests", func() {
+		It("cluster creation with name having capital letters should fail", func() {
+			awsClusterName := "capa-cluster-CAPS"
+			err := testcommons.MakeAWSCluster(namespace, awsClusterName, KeyPairName, kindClient)
```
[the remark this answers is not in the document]
sirao: This is not global variable. It is package variable. As test_util.go is in commons package, had to pass it.

sirao:
I have added test_util.go under test/commons with intention to avoid duplicate code between E2E & Integration tests. This test_util.go contains functions which can be used in both E2E and Integration.

scripts/ci-integration.sh
```diff
@@ -26,5 +35,63 @@ source "${REPO_ROOT}/hack/ensure-go.sh"
 source "${REPO_ROOT}/hack/ensure-kind.sh"
 # shellcheck source=../hack/ensure-kubectl.sh
 source "${REPO_ROOT}/hack/ensure-kubectl.sh"
+# shellcheck source=../hack/ensure-kustomize.sh
+source "${REPO_ROOT}/hack/ensure-kustomize.sh"
+
+# If BOSKOS_HOST is set then acquire an AWS account from Boskos.
+if [ -n "${BOSKOS_HOST:-}" ]; then
```
detiber: I'd like to avoid the introduction of BOSKOS and real AWS accounts with the integration tests. Can the tests be reworked to mock out the required functionality and call into the Reconcile methods directly rather than running against a fully built and deployed CAPA deployment?

test/integration/integration_test.go
```diff
@@ -0,0 +1,90 @@
...
+		namespace   string
+		clusterName string
+	)
+
+	BeforeEach(func() {
+		namespace = "test-namespace-" + util.RandomString(6)
+		testutil.CreateNamespace(namespace, kindClient)
+	})
+
+	Describe("cluster name validation tests", func() {
```
chuckha: It looks like we are testing something that should be part of a validation webhook. The way the validation webhook is structured we should be able to write a unit test for this logic without need an entire CAPA deployment.

ncdc: Do you think it's worthwhile to have automated testing for webhook validation someplace though?

vincepri: I thought we didn't have validation webhooks for release-0.4

ncdc: We don't, ha, I wasn't paying attention to the branch.

chuckha: If we write the actual logic for testing this stuff and it's complex then yes, a test would be nice. A unit test would be ideal for a function that is doing validation. Spinning up an entire kubernetes cluster and CAPA to see if a validation function passes is a lot and we should minimize overhead where possible.

If we don't have webhooks this logic will have to exist ... I guess in the aws cluster reconciler?

If we can use kubebuilder annotations for it that would also be fine with me. We wouldn't have to write tests for those, we can assume those work.

ncdc: I don't think we need a test to verify the name is a DNS-1123 subdomain. This is handled by kube-apiserver code.

ncdc:
/hold

I'm not sure we need any of these tests for names because name validation is handled by kube-apiserver.

ncdc:
@sirao are you testing anything that is unique to `Cluster` or `AWSCluster` in terms of naming, or are you just testing for valid Kubernetes resource names?

sirao:
@ncdc I was testing to make sure with all possible name combinations by which cluster and awscluster get created. Earlier we have seen an issue that when the name is beyond 22 characters cluster creation succeeded where the awscluster was not provisioned as Load Balancer wasn't created because of lengthy name.

ncdc:
@sirao thanks for the clarification. In that case, I would recommend that at a minimum we test creating a cluster with a name > 22 characters. I'm not sure you need anything else.

detiber:
> @sirao thanks for the clarification. In that case, I would recommend that at a minimum we test creating a cluster with a name > 22 characters. I'm not sure you need anything else.

+1, I think this is the only test that we truly need. Like @ncdc mentioned the character types are enforced by the kubernetes api server.

sirao:
> > @sirao thanks for the clarification. In that case, I would recommend that at a minimum we test creating a cluster with a name > 22 characters. I'm not sure you need anything else.
> 
> +1, I think this is the only test that we truly need. Like @ncdc mentioned the character types are enforced by the kubernetes api server.

Still should it be part of integration tests? Here we not only need to verify the cluster name validated? But we need to validate the LoadBalancer created and cluster comes to provisioned state or not?

detiber:
> Still should it be part of integration tests? Here we not only need to verify the cluster name validated? > But we need to validate the LoadBalancer created and cluster comes to provisioned state or not?

It should likely be run as an e2e rather than integration

ncdc:
Superseded by #1497 
/close

PR #1385, opened 2019-12-03
🐛 make create-cluster: Improve cert-manager wait

### LINKED ISSUE #1383: Test failure: ci-cluster-api-provider-aws-make-conformance-master
As seen in e.g. https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-cluster-api-provider-aws-make-conformance-master/1201824548337487872:

```
error: error validating "examples/_out/machinedeployment.yaml": error validating data: [ValidationError(MachineDeployment.spec): unknown field "clusterName" in io.x-k8s.cluster.v1alpha2.MachineDeployment.spec, ValidationError(MachineDeployment.spec.template.spec): unknown field "clusterName" in io.x-k8s.cluster.v1alpha2.MachineDeployment.spec.template.spec]; if you choose to ignore these errors, turn validation off with --validate=false
```

Need to update examples/machinedeployment/kustomizeconfig.yaml and machinedeployment.yaml to change the apiVersion to cluster.x-k8s.io/v1alpha3

ncdc:
Instead of sleeping 10 seconds and hoping that cert-manager is ready,
use 'kubectl wait' to wait for the cert-manager webhook apiservice.

Also fix a conformance issue where the apiVersion of the MachineDeployment was wrong (v1a1 instead of v1a2).

Fixes #1383

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,7 +313,9 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS 
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
 		wait --for=condition=Ready --namespace=cert-manager --timeout=15m pods --all
 	# Wait for webhook servers to be ready to take requests
-	sleep 10
+	kubectl \
+		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
+		wait --for=condition=Available --timeout=5m apiservice v1beta1.webhook.cert-manager.io
 	# Apply provider-components.
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
--- a/Makefile
+++ b/Makefile
@@ -306,12 +306,6 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS 
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
 		create -f examples/_out/cert-manager.yaml
-	# Wait for cert-manager pods to be created
-	sleep 20
-	# Wait for cert-manager pods to be ready.
-	kubectl \
-		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
-		wait --for=condition=Ready --namespace=cert-manager --timeout=15m pods --all
 	# Wait for webhook servers to be ready to take requests
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
--- a/Makefile
+++ b/Makefile
@@ -314,6 +314,14 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS 
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
 		create -f examples/_out/provider-components.yaml
+	# Wait for CAPI pod
+	kubectl \
+		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
+		wait --for=condition=Ready --timeout=5m -n capi-system pod -l control-plane=cluster-api-controller-manager
+    # Wait for CAPA pod
+	kubectl \
+		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
+		wait --for=condition=Ready --timeout=5m -n capa-system pod -l control-plane=capa-controller-manager
 	# Create Cluster.
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
--- a/examples/machinedeployment/kustomizeconfig.yaml
+++ b/examples/machinedeployment/kustomizeconfig.yaml
@@ -1,11 +1,11 @@
 namespace:
 - kind: MachineDeployment
   group: cluster.x-k8s.io
-  version: v1alpha2
+  version: v1alpha3
   path: spec/template/spec/infrastructureRef/namespace
   create: true
 - kind: MachineDeployment
   group: cluster.x-k8s.io
-  version: v1alpha2
+  version: v1alpha3
   path: spec/template/spec/bootstrap/configRef/namespace
   create: true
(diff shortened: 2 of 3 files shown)


Makefile
```diff
@@ -313,7 +313,9 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
 		wait --for=condition=Ready --namespace=cert-manager --timeout=15m pods --all
 	# Wait for webhook servers to be ready to take requests
-	sleep 10
+	kubectl \
+		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
+		wait --for=condition=Available --timeout=5m apiservice v1beta1.webhook.cert-manager.io
```
detiber: Can we remove the sleep on line 310 and the wait on lines 312-314 as well with the addition of this wait?

ncdc: Probably. I don't have any of these other sleeps/waits in my Tiltfile and I've never had issues deploying.

detiber:
/hold
holding for conformance test verification

Makefile
```diff
@@ -305,27 +305,31 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
...
 	# Apply provider-components.
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
-		create -f examples/_out/provider-components.yaml
+		apply -f examples/_out/provider-components.yaml
+	# Wait for CAPI pod
+	kubectl \
+		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
+		wait --for=condition=Ready --timeout=5m -n capi-system pod -l control-plane=cluster-api-controller-manager
```
detiber: Would it be better to wait for the api types or the webhooks instead?

Makefile
```diff
@@ -305,27 +305,31 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
...
+    # Wait for CAPA pod
+	kubectl \
+		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
+		wait --for=condition=Ready --timeout=5m -n capa-system pod -l control-plane=capa-controller-manager
```
detiber: Would it be better to wait for the api types or the webhooks instead?

Makefile
```diff
@@ -305,27 +305,31 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
...
 	# Create Cluster.
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
-		create -f examples/_out/cluster.yaml
+		apply -f examples/_out/cluster.yaml
```
detiber: Why apply here? I'd worry that we could potentially mutate something existing vs creating from scratch here.

Makefile
```diff
@@ -305,27 +305,31 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
...
 	# Create control plane machine.
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
-		create -f examples/_out/controlplane.yaml
+		apply -f examples/_out/controlplane.yaml
```
detiber: Why apply here? I'd worry that we could potentially mutate something existing vs creating from scratch here.

Makefile
```diff
@@ -340,7 +344,7 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
 	# Create a worker node with MachineDeployment.
 	kubectl \
 		--kubeconfig=$$(kind get kubeconfig-path --name="clusterapi") \
-		create -f examples/_out/machinedeployment.yaml
+		apply -f examples/_out/machinedeployment.yaml
```
detiber: Why apply here? I'd worry that we could potentially mutate something existing vs creating from scratch here.

Makefile
ncdc: We could add waits on the CRDs too. I couldn't find a way to wait for a webhook to be ready - it's served by the pod, via the service. So if the pod is ready, there's maybe a slim chance that the webhook service isn't quite there yet, but if we start seeing flakes, we can add more waits here.

detiber: @ncdc isn't waiting for the webhook service to be available precisely what the wait we are doing in the cert manager wait above?

Makefile
ncdc: I was manually rerunning selective commands from here, and you can't do that with create.

As the code stands, this is operating against a brand new, clean kind cluster. There is nothing to overwrite. Going forward, if we proceed with requiring a valid kubeconfig (and therefore a cluster) instead of having this make target create one, I think apply is still good. This target is really "install capi + capa into a management cluster and then create a workload cluster".

Makefile
detiber: ah, nm, it's just checking the status of the CRD being registered...  I'm wondering if we could check on the status of the `Service` we deploy for the webhook instead.

ncdc: Not exactly. The cert-manager wait is waiting on an apiservice that comes from an aggregated api server, which means it's waiting to be able to talk to the service->pod for it.

ncdc: If we waited on the v1alpha3.cluster.x-k8s.io apiservice, you'd see this:
```
    "status": {
        "conditions": [
            {
                "lastTransitionTime": "2019-12-04T15:11:02Z",
                "message": "Local APIServices are always available",
                "reason": "Local",
                "status": "True",
                "type": "Available"
            }
        ]
    }
```

ncdc: The most extensive check would probably be to have liveness and readiness probes in the capi/capa pods, and then check that the 2 services have at least 1 endpoint.

ncdc: I think this is "good enough" for now. We can tweak later if needed.

Makefile
detiber: My worry with apply is that the state may not necessarily be what we expect it to be (the equivalent of a new clean deployment of capi + capa components), or that we could introduce changes that would break deployment from a clean cluster inadvertently to the manifests (requiring some new additional pre-requisite in the management cluster) that we previously didn't. The use of apply could potentially mask these until release time (or a new developer tries to replicate).

ncdc: Reverted

ncdc:
Woot, conformance is now running. https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/1385/pull-cluster-api-provider-aws-make-conformance/1202249997781831680

```
Will run 280 of 4814 specs
```

/hold cancel

PR #1388, opened 2019-12-04
🏃 Make conformance cluster names unique by default

ncdc:
**What this PR does / why we need it**:
Make conformance cluster names unique by default, so we have a better chance of being able to retrieve logs from the VMs (it's broken currently because there's already a keypair for `test1`, and because each run is _not_ generating a new keypair, it's unable to retrieve).

/hold

PR #1391, opened 2019-12-05
🏃 [0.4] Make conformance cluster names unique by default

ncdc:
*What this PR does / why we need it**:
Make conformance cluster names unique by default, so we have a better chance of being able to retrieve logs from the VMs (it's broken currently because there's already a keypair for test1, and because each run is not generating a new keypair, it's unable to retrieve).

This is a manual backport of #1388 to release-0.4.

/hold


ncdc:
Ok, the changes in this PR worked and we have logs.
/hold cancel

PR #1393, opened 2019-12-05
🏃 [0.4] conformance: Fix gsutil usage, try capturing all VM logs

ncdc:
**Change 1**
When generating examples for the conformance tests, the
/tmp/kubeadm-bootstraph.sh preKubeadmCommand downloads gsutil if it does
not exist and adjusts $PATH accordingly. However, the script that
generates examples substitutes the values for environment variable
placeholders in the file, which means it will substitute whatever the
current value of $PATH is in your local environment.

Instead of trying to adjust $PATH, use a variable for gsutil to invoke
it.

**Change 2**
We've had an issue where we aren't always capturing the logs for the
control plane VMs, especially if the cluster fails to initialize;
instead, we've managed to capture just the bastion. This change removes
the code that filters out the first private IP address when grabbing
logs. We'll still end up getting the bastion logs, but hopefully we'll
also get the first control plane instance.

**Change 3**
Filter bastion and cluster VMs using tags

Fixes #

/hold

### DIFF
--- a/examples/controlplane/kustomizeversions.yaml
+++ b/examples/controlplane/kustomizeversions.yaml
@@ -20,11 +20,12 @@ spec:
 
       [[ $(id -u) != 0 ]] && SUDO="sudo" || SUDO=""
 
-      if ! gsutil version > /dev/null; then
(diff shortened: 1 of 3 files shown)


ncdc:
This appears to have worked the way I wanted it to.
/hold cancel
/assign @vincepri

PR #1396, opened 2019-12-05
🏃 conformance: Fix gsutil usage, try capturing all VM logs

ncdc:
**Change 1**
When generating examples for the conformance tests, the
/tmp/kubeadm-bootstraph.sh preKubeadmCommand downloads gsutil if it does
not exist and adjusts $PATH accordingly. However, the script that
generates examples substitutes the values for environment variable
placeholders in the file, which means it will substitute whatever the
current value of $PATH is in your local environment.

Instead of trying to adjust $PATH, use a variable for gsutil to invoke
it.

**Change 2**
We've had an issue where we aren't always capturing the logs for the
control plane VMs, especially if the cluster fails to initialize;
instead, we've managed to capture just the bastion. This change removes
the code that filters out the first private IP address when grabbing
logs. We'll still end up getting the bastion logs, but hopefully we'll
also get the first control plane instance.

**Change 3**
Filter bastion and cluster VMs using tags

Forward port of #1393 

Fixes #

/hold

### DIFF
--- a/examples/controlplane/kustomizeversions.yaml
+++ b/examples/controlplane/kustomizeversions.yaml
@@ -20,11 +20,12 @@ spec:
 
       [[ $(id -u) != 0 ]] && SUDO="sudo" || SUDO=""
 
-      if ! gsutil version > /dev/null; then
-      curl -sSL https://sdk.cloud.google.com > /tmp/gcl && bash /tmp/gcl --install-dir=~/gcloud --disable-prompts > /dev/null 2>&1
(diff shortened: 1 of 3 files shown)


ncdc:
In the conformance test it failed to bring up the cluster, and it's finishing up grabbing logs and tearing things down now. I don't think anything in here broke cluster startup but will confirm before removing the hold.

ncdc:
Weird test failure. It took ~30 minutes for the bootstrap data to get generated, at which point, we timed out.

/hold cancel

PR #1397, opened 2019-12-05
🏃 [0.4] conformance: use the right bucket for released vs ci

ncdc:
**What this PR does / why we need it**:
If CI_VERSION is a released version such as v1.16.3, we need to use a
different GCS bucket to download the artifacts.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
This might f-i-x #1389

/cc @dims @vincepri 
/assign @dims @vincepri 
/milestone v0.4.x

PR #1398, opened 2019-12-05
:running: conformance: use the right bucket for released vs ci

ncdc:
If CI_VERSION is a released version such as v1.16.3, we need to use a
different GCS bucket to download the artifacts.

Signed-off-by: Andy Goldstein <goldsteina@vmware.com>
(cherry picked from commit 34c6e32fe5504bb8b515feb0575e748022efa4bc)

Forward port of #1397 

Fixes #


ncdc:
It turns out this doesn't actually work.
/close

vincepri:
Saw the PR :P

PR #1404, opened 2019-12-09
:running: Remove clusterctl calls from create-cluster / update CAPI

vincepri:

Removes all clusterctl calls from `create-cluster` so we can update CAPI dependency to the latest master. I'm working on getting the bootstrap data secret reference in place in a future PR, and this is needed to rev the CAPI master version.


Makefile
```diff
@@ -328,17 +328,13 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
...
 	kubectl \
 		create -f examples/_out/controlplane.yaml
-	# Get KubeConfig using clusterctl.
-	$(CLUSTERCTL) \
-		alpha phases get-kubeconfig -v=3 \
-		--kubeconfig=$${HOME}/.kube/config \
-		--namespace=default \
-		--cluster-name=$(CLUSTER_NAME)
+	# Wait for the kubeconfig to become available.
+	timeout 300 bash -c "while ! kubectl get secrets | grep test1-kubeconfig; do sleep 1; done"
```
ncdc: You can't hardcode the cluster name (`test1`)

Makefile
```diff
@@ -328,17 +328,13 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
...
+	# Get kubeconfig and store it locally.
+	kubectl get secrets test1-kubeconfig -o json | jq -r .data.value | base64 -D > ./kubeconfig
```
ncdc: Can't hardcode cluster name

ncdc: Does `./kubeconfig` end up being an untracked file for git?

Makefile
```diff
@@ -328,17 +328,13 @@ create-cluster: $(CLUSTERCTL) ## Create a development Kubernetes cluster on AWS
...
 	# Apply addons on the target cluster, waiting for the control-plane to become available.
-	$(CLUSTERCTL) \
-		alpha phases apply-addons -v=3 \
-		--kubeconfig=./kubeconfig \
-		-a examples/addons.yaml
+	timeout 300 bash -c "while ! kubectl --kubeconfig=./kubeconfig get nodes | grep master; do sleep 1; done"
```
ncdc: Can we do a query using labels or something rather than grepping for the master node role?

ncdc:
@vincepri can you do a follow-up to fix the hard coded cluster names?

vincepri:
Yeah

PR #1407, opened 2019-12-10
⚠️ Support bootstrap data in secret

vincepri:
This PR is a follow-up requirement to an upstream change that now stores the bootstrap data in a secret rather than inline plaintext field.


pkg/cloud/scope/machine.go
```diff
@@ -183,6 +184,21 @@ func (m *MachineScope) SetAddresses(addrs []corev1.NodeAddress) {
...
+// GetBootstrapData returns the bootstrap data from the secret in the Machine's bootstrap.dataSecretName.
+func (m *MachineScope) GetBootstrapData() (string, error) {
+	if m.Machine.Spec.Bootstrap.DataSecretName == nil {
+		return "", errors.New("error retrieving bootstrap data: linked Machine's bootstrap.dataSecretName is nil")
+	}
+
+	secret := &corev1.Secret{}
+	key := types.NamespacedName{Namespace: m.Namespace(), Name: *m.Machine.Spec.Bootstrap.DataSecretName}
+	if err := m.client.Get(context.TODO(), key, secret); err != nil {
+		return "", errors.Wrapf(err, "failed to retrieve bootstrapd data secret for AWSMachine %s/%s", m.Namespace(), m.Name())
```
ncdc:
```suggestion
		return "", errors.Wrapf(err, "failed to retrieve bootstrap data secret for AWSMachine %s/%s", m.Namespace(), m.Name())
```

vincepri: Fixed

pkg/cloud/services/ec2/instances.go
```diff
@@ -178,7 +179,11 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*infrav1.Instance,
 	}
 
 	// Set userdata.
-	input.UserData = aws.String(*scope.Machine.Spec.Bootstrap.Data)
+	userData, err := scope.GetBootstrapData()
+	if err != nil {
+		return nil, err
```
ncdc: Sorry for the additional comment, but I think this error will get "swallowed" in that it's only written to the pod log, and it's never visible to the user anywhere. We generate events in this method (CreateInstance), but only for certain failures. Maybe we should merge this as-is, and have a follow-up issue to adjust the success/fail event creation in CreateInstance so it covers all returnable errors?

pkg/cloud/scope/machine.go
```diff
@@ -183,6 +184,21 @@ func (m *MachineScope) SetAddresses(addrs []corev1.NodeAddress) {
 	m.AWSMachine.Status.Addresses = addrs
 }
 
+// GetBootstrapData returns the bootstrap data from the secret in the Machine's bootstrap.dataSecretName.
+func (m *MachineScope) GetBootstrapData() (string, error) {
+	if m.Machine.Spec.Bootstrap.DataSecretName == nil {
+		return "", errors.New("error retrieving bootstrap data: linked Machine's bootstrap.dataSecretName is nil")
```
ncdc: This error should probably include the machine's ns & name?

pkg/cloud/services/ec2/instances.go
vincepri: Added a error and success event for creation

pkg/cloud/scope/machine.go
vincepri: I added the AWSMachine namespace and name from the caller, so all the errors will get it

PR #1408, opened 2019-12-10
:bug: Update RBAC to add getting, listing, and watching Secrets for bootstrap data

detiber:

Adds RBAC support for getting bootstrap data from secret.

Followup to #1407

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -18,6 +18,14 @@ rules:
   - update
   - watch
 - apiGroups:
+  - ""
+  resources:
+  - secrets
+  verbs:
+  - get
+  - list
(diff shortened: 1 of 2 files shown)


detiber:
/assign @vincepri @ncdc 

I'm still seeing an error after this, but this at least gets me to the point that CAPA is retrieving the secret from the api server.

Error seen with this PR:
```
E1210 20:07:06.387295       1 controller.go:258] controller-runtime/controller "msg"="Reconciler error" "error"="failed to create EC2 instance for AWSMachine default/kcp-infra-template-flxzs: failed to decode bootstrapData: illegal base64 data at input byte 0"
```

PR #1409, opened 2019-12-10
:bug: Base64 the User Data retrieved from the bootstrap secret

detiber:

With the bootstrap data moved to a secret, we were attempting to pass in a decoded string where the ec2 api expects the user data to be base64 encoded (client-go graciously decodes the contents of the secret).

### DIFF
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -18,6 +18,7 @@ package scope
 
 import (
 	"context"
+	"encoding/base64"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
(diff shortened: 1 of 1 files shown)


ncdc:
This is probably worthy of a unit test to avoid regressions?

detiber:
> This is probably worthy of a unit test to avoid regressions?

Probably, I can look into it after I've finished cleaning up the control plane PR I have in progress, this was blocking my testing/validation of that.

ncdc:
/lgtm
+ follow-up issue for unit test

ncdc:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1410

PR #1420, opened 2019-12-13
✨ Add support for failure domains

vincepri:
Related to upstream failure domain work.


api/v1alpha3/awsmachine_types.go
```diff
@@ -70,8 +70,15 @@ type AWSMachineSpec struct {
 	// +optional
 	AdditionalSecurityGroups []AWSResourceReference `json:"additionalSecurityGroups,omitempty"`
 
+	// FailureDomain is the failure domain unique identifier this Machine should be attached to, as defined in Cluster API.
+	// For this infrastructure provider, the ID is equivalent to an AWS Availability Zone.
+	// If multiple subnets are matched for the availability zone, the first one return is picked.
+	FailureDomain *string `json:"failureDomainID,omitempty"`
```
detiber: If we are going generic with the naming here, should we add it to the Machine resource instead?

vincepri:
Force-pushed this in my branch but it's not updating

pkg/cloud/services/ec2/instances.go
```diff
@@ -151,17 +151,25 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*infrav1.Instance,
 	// TODO(vincepri): Move subnet picking logic to its own function/method.
 	if scope.AWSMachine.Spec.Subnet != nil && scope.AWSMachine.Spec.Subnet.ID != nil {
 		input.SubnetID = *scope.AWSMachine.Spec.Subnet.ID
-	} else if scope.AWSMachine.Spec.AvailabilityZone != nil {
-		sns := s.scope.Subnets().FilterPrivate().FilterByZone(*scope.AWSMachine.Spec.AvailabilityZone)
+	} else if scope.AWSMachine.Spec.AvailabilityZone != nil || scope.AWSMachine.Spec.FailureDomain != nil {
+		availabilityZone := scope.AWSMachine.Spec.AvailabilityZone
```
detiber: Since `AvailabilityZone` is the deprecated field, should this be the fallback value instead of using it as the primary value here?

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -73,7 +73,8 @@ func (s *Service) ReconcileLoadbalancers() error {
 		return errors.Wrapf(err, "failed to reconcile tags for apiserver load balancer %q", apiELB.Name)
 	}
 
-	// Reconciliate the subnets from the spec and the ones currently attached to the load balancer.
+	// Reconciliate the subnets and availability zones from the spec
```
detiber:
```suggestion
	// Reconcile the subnets and availability zones from the spec
```

pkg/cloud/services/ec2/instances.go
vincepri: Sure I can switch them up

api/v1alpha3/awsmachine_types.go
vincepri: That's another solution, in that case we'd lose the comment and description specific to the infrastructure provider though

api/v1alpha2/zz_generated.conversion.go
```diff
@@ -607,6 +613,7 @@ func autoConvert_v1alpha3_AWSMachineSpec_To_v1alpha2_AWSMachineSpec(in *v1alpha3
 	out.IAMInstanceProfile = in.IAMInstanceProfile
 	out.PublicIP = (*bool)(unsafe.Pointer(in.PublicIP))
 	out.AdditionalSecurityGroups = *(*[]AWSResourceReference)(unsafe.Pointer(&in.AdditionalSecurityGroups))
+	// WARNING: in.FailureDomain requires manual conversion: does not exist in peer-type
```
detiber: How should downconversion be handled due to this? Should we just fail downconversion if FailureDomain is set? Or should we override AvailabilityZone with the value from FailureDomain?

vincepri: That's a good point, lemme check

vincepri: Added the manual conversion from FD -> AZ, if AZ is empty

PR #1422, opened 2019-12-13
[WIP] ✨ Add tilt support & rework manager credentials

ncdc:
Add tilt support based on the new tilt workflow in cluster-api.

Remove tilt bits from Dockerfile, as they are now contained in
cluster-api's Tiltfile.

Switch to a new kustomize secretGenerator approach for managing AWS
credentials. There is now a hack script to generate
config/manager/credentials.env containing just the credentials, and this
file is in .gitignore so it won't be accidentally checked in.

This goes along with https://github.com/kubernetes-sigs/cluster-api/pull/1894

Fixes #


ncdc:
PTAL at the kustomize changes for manager credentials

rudoi:
lgtm as well, i'm so glad this stuff is coming together.

ncdc:
@fabriziopandini PTAL

ncdc:
This breaks the quick start. HOLD

tilt-provider.json
```diff
@@ -0,0 +1,12 @@
+{
+    "name": "aws",
+    "config": {
+        "image": "gcr.io/k8s-staging-cluster-api-aws/cluster-api-aws-controller",
+        "ignore": [
+            "config/manager/credentials.env"
```
detiber: Should this be `config/manager/credentials` instead?

ncdc: It's actually not needed anymore, because I'm using `only` in the Tiltfile. Will remove.

ncdc:
Re-enabling WIP as I need to not break the quickstart and clusterctl v2

ncdc:
Closing in favor of #1432

PR #1426, opened 2019-12-16
:sparkles: Add support for extra AWS policies in clusterawsadm

### LINKED ISSUE #1390: Clusterawsadm: Enable adding of additional policies to control plane & node roles
/kind feature

A situation I've come across a few times is having a common policy / set of policies that I want to add to my control plane and / or node instances. Right now that's an activity after the fact to attach those policies to the relevant roles (or the alternative of generating the capa policies and manually creating my roles ahead of time). 

Wondered what folks thought about adding something like `--additional-cp-policies` and `--additional-worker-policies` flags to `clusterawsadm alpha bootstrap create-stack` so those policies could be attached at create time through the CloudFormation stack?

I'm happy to work on this if folks think it's a good idea.

**Environment:**

- Cluster-api-provider-aws version: 0.4.7
- Kubernetes version: (use `kubectl version`): 1.16.2
- OS (e.g. from `/etc/os-release`): Ubuntu 18.04

johnharris85:
Convenience feature to allow users to specify extra policies to the control plane and node roles that `clusterawsadm` creates.

Fixes #1390 

Signed-off-by: John Harris <joharris@vmware.com>

### DIFF
--- a/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
+++ b/cmd/clusterawsadm/cmd/alpha/bootstrap/bootstrap.go
@@ -44,6 +44,11 @@ aws_session_token = {{ .SessionToken }}
 {{end}}
 `
 
+var (
+	extraControlPlanePolicies []string
+	extraNodePolicies         []string
+)
+
 // RootCmd is the root of the `alpha bootstrap command`
 func RootCmd() *cobra.Command {
 	newCmd := &cobra.Command{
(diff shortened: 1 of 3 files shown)


ncdc:
@johnharris85 this change broke our e2e code: https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-cluster-api-provider-aws-e2e-master/1206646250376531968. Would you be able to fix it? I opened https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1428

johnharris85:
Yup will look in ~1hr when I get home.

Issue #1389, opened 2019-12-04
Test failure: capa-conformance-stable*


ncdc:
Recent failures:
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-cluster-api-provider-aws-make-conformance-stable/1202280071990087680
- https://prow.k8s.io/view/gcs/kubernetes-jenkins/logs/ci-cluster-api-provider-aws-make-conformance-stable-k8s-ci-artifacts/1202280072006864896

Both seem to have similar errors:

```
I1204 17:58:17.890103   14265 clusterclient.go:242] Waiting to acquire client... server not yet available: Get https://test1-apiserver-724810617.us-west-2.elb.amazonaws.com:6443/api?timeout=32s: EOF
F1204 17:58:17.890447   14265 alpha_phase_apply_addons.go:51] unable to create cluster client: failed to acquire new client: timed out waiting for the condition
make: *** [Makefile:328: create-cluster-management] Error 1
```

We have an issue where we are currently unable to retrieve log files from the VMs. I have opened #1388 to try to address this, to assist with debugging.

/assign
/lifecycle active
/kind failing-test
/priority critical-urgent
/milestone v0.5.0

ncdc:
I didn't realize these jobs were against release-0.4, and #1388 is for master 😦. I've cherry-picked it in #1391 for 0.4.

ncdc:
Failure with logs: https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/1391/pull-cluster-api-provider-aws-make-conformance/1202559795685494786

It appears that cloud-init isn't running our bootstrap at all - see https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/1391/pull-cluster-api-provider-aws-make-conformance/1202559795685494786/artifacts/logs/10.0.1.24/cloud-final.log

ncdc:
Unless that's the bastion?

ncdc:
Yeah that's the bastion. Nevermind

ncdc:
I think I'm going to have to try to run this locally to see if I can reproduce. For some reason gathering logs is not seeing the first control plane VM, which is what we need to determine why the apiserver isn't coming online.

ncdc:
[...]

PR #1436, opened 2019-12-17
🏃 [0.4] Fix loading released artifacts in conformance tests

ncdc:

Fix loading released artifacts in conformance tests. I made the following changes:

- Removed examples/{controlplane,machinedeployment}/kustomizeversions.yaml and replaced it with more targeted patches
- Consolidated the patch to the KubeadmConfig (control plane) and KubeadmConfigTemplate (machine deployment) to add the `/tmp/bootstrap.sh` pre-kubeadm command into a single file, `hack/ci/ci-bootstrap.yaml`
- Updated the above patch to support both CI and released artifacts
- Fixed the `fix_manifests()` function so it works on macOS
- Updated `hack/ci/e2e-conformance.sh` so it invokes `fix_manifests` if either `USE_CI_ARTIFACTS` or `CI_VERSION` is set
- Fixed `hack/ensure-kind.sh` to work with kind 0.6+ (which is what I'm testing with locally)
- Adjusted `scripts/ci-conformance.sh` to remove the hard-coded `--use-ci-artifacts` flag (matching what's in master)

Hopefully #1389


ncdc:
@vincepri @dims FYI the PR conformance job is running, so I at least haven't broken that entirely 😄 (https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/1436/pull-cluster-api-provider-aws-make-conformance/1207386645469335552, although it is showing some test failures).

I still want to hold this to check out the artifacts before we merge, but it's ready for 👀.

PR #1442, opened 2019-12-20
:sparkles: Support cross zone load balancing option

### LINKED ISSUE #1416: Cross-Zone Load Balancing is disabled for Load Balancer when cluster crated with Multi AZ configuration
It is observed that when Created the cluster with Multi AZ configuration, the load balancer which was provisioned have Cross-Zone Load Balancing Disabled.

For better Load Balancing and HA, it would be good to enable Cross-Zone Load Balancing

vincepri:

Fixes #1416


api/v1alpha3/awscluster_types.go
```diff
@@ -69,6 +69,15 @@ type AWSLoadBalancerSpec struct {
...
 	Scheme *ClassicELBScheme `json:"scheme,omitempty"`
+
+	// CrossZoneLoadBalancing enables the classic load balancer load balancing.
+	// With cross-zone load balancing, each load balancer node for your Classic Load Balancer
+	// distributes requests evenly across the registered instances in all enabled Availability Zones.
+	// If cross-zone load balancing is disabled, each load balancer node distributes requests evenly across
+	// the registered instances in its Availability Zone only.
+	// Defaults to false.
+	// +optional
+	CrossZoneLoadBalancing *bool `json:"crossZoneLoadBalancing,omitempty"`
```
ncdc: I don't see any reason to make this a pointer?

api/v1alpha3/types.go
```diff
@@ -146,6 +146,10 @@ type ClassicELBAttributes struct {
 	// IdleTimeout is time that the connection is allowed to be idle (no data
 	// has been sent over the connection) before it is closed by the load balancer.
 	IdleTimeout time.Duration `json:"idleTimeout,omitempty"`
+
+	// CrossZoneLoadBalancing enables the classic load balancer load balancing.
+	// +optional
+	CrossZoneLoadBalancing *bool `json:"crossZoneLoadBalancing,omitempty"`
```
ncdc: Non-pointer?

api/v1alpha2/awscluster_conversion.go
```diff
@@ -100,3 +102,26 @@ func Convert_v1alpha3_AWSClusterStatus_To_v1alpha2_AWSClusterStatus(in *infrav1a
...
 }
+
+// Convert_v1alpha3_AWSLoadBalancerSpec_To_v1alpha2_AWSLoadBalancerSpec.
+func Convert_v1alpha3_AWSLoadBalancerSpec_To_v1alpha2_AWSLoadBalancerSpec(in *infrav1alpha3.AWSLoadBalancerSpec, out *AWSLoadBalancerSpec, s apiconversion.Scope) error { //nolint
+	if err := autoConvert_v1alpha3_AWSLoadBalancerSpec_To_v1alpha2_AWSLoadBalancerSpec(in, out, s); err != nil {
+		return err
+	}
+
+	if in.CrossZoneLoadBalancing != nil && *in.CrossZoneLoadBalancing {
+		return errors.New("cannot convert AWSLoadBalancerSpec crossZoneLoadBalancing field")
```
ncdc: Checked with apimachinery. We can't fail when downconverting because the older version doesn't have the newer field. We can do one of:

1. Silently drop the field
1. Add the new field to all API versions so that it's convertible (it would be present but unreconciled/unused in older versions)
1. Shuttle the field through metadata (annotation) so that a3 -> a2 -> a3 would preserve the value.

ncdc: Option 1 is probably a nonstarter because it would mean we'd lose data if we did a3 -> a2 -> update apiserver.

ncdc:
@vincepri you still looking to get this in?

vincepri:
Sure, I can rebase in a bit

vincepri:
@ncdc done!

api/v1alpha3/awscluster_types.go
```diff
@@ -80,6 +80,15 @@ type AWSLoadBalancerSpec struct {
 	// Scheme sets the scheme of the load balancer (defaults to Internet-facing)
 	// +optional
 	Scheme *ClassicELBScheme `json:"scheme,omitempty"`
+
+	// CrossZoneLoadBalancing enables the classic load balancer load balancing.
```
ncdc: Did you mean to say that it enables classic, or cross-zone?

ncdc:
Also add a unit test for round trip conversions?

PR #1447, opened 2019-12-23
🐛 Create events for Elastic IP allocate/disassociate/release

### LINKED ISSUE #1346: Events not generated when Create/Release/Disassociate elastic IP fails
/kind bug

Tried the create a new cluster on an AWS account where elastic IP limit is already reached 

An event generated that the Elastic IP could not be created. 

**Environment:**

aaroniscode:
also make related errors more consistent

Calls to EC2 for Elastic IP's didn't generate events if there were errors. This PR creates those events.

Fixes #1346 

Noticed some inconsistency around the related errors and made some very minor changes so they are consistent. Let me know if that's OK or if I should revert those changes.

### DIFF
--- a/pkg/cloud/services/ec2/eips.go
+++ b/pkg/cloud/services/ec2/eips.go
@@ -27,6 +27,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/filter"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/wait"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/tags"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
 )
 
 func (s *Service) getOrAllocateAddress(role string) (string, error) {
@@ -51,7 +52,8 @@ func (s *Service) allocateAddress(role string) (string, error) {
 	})
 
 	if err != nil {
-		return "", errors.Wrap(err, "failed to create Elastic IP address")
+		record.Warnf(s.scope.AWSCluster, "FailedAllocateEIP", "Failed to allocate Elastic IP for %q: %v", role, err)
+		return "", errors.Wrap(err, "failed to allocate Elastic IP")
 	}
 
 	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/ec2/eips.go
```diff
@@ -121,7 +124,8 @@ func (s *Service) releaseAddresses() error {
 				AssociationId: ip.AssociationId,
 			})
 			if err != nil {
-				return errors.Errorf("failed to release elastic IP %q with allocation ID %q: Still associated with association ID %q", *ip.PublicIp, *ip.AllocationId, *ip.AssociationId)
+				record.Warnf(s.scope.AWSCluster, "FailedDisassociateEIP", "Failed to disassociate Elastic IP %q: %v", *ip.AllocationId, err)
+				return errors.Errorf("failed to disassociate Elastic IP %q with allocation ID %q: Still associated with association ID %q", *ip.PublicIp, *ip.AllocationId, *ip.AssociationId)
```
detiber: Not directly related to this PR, but it seems really odd that we are disassociating here, but also calling the `disassociateAddress` method below in the error block for the call to `s.scope.EC2.ReleaseAddress` below...  I'm wondering if we need to replace the call to `s.scope.EC2.DisassociateAddress` here to a call to `disassociateAddress` here, or just remove it and let it be handled below with the error handling when trying to release the address?

ncdc: I mentioned that it looked odd to @vincepri and he said there was a reason for it (which I've not forgotten)... help?

detiber: I could see issues related to the eventual consistency of the API, but that could be handled by just retrying the call to release. There is also the case for the EIP associated with the NAT gateway, which we cannot disassociate or release until some background processes complete, but we could handle that by a call w/ retries to `disassociateAddress` above the call to `s.scope.EC2.ReleaseAddress`

aaroniscode: should we open a new issue to track what @detiber raised as a potential issue?  Or would you like to resolve the issues mentioned above in this PR?

detiber: Followup issue sounds good.

aaroniscode: Follow up issue #1459 opened.

PR #1448, opened 2019-12-26
🐛Prevent NPE if Machine's spec.Version is not set by failing the AWSMachine

### LINKED ISSUE #1092: NPE if MachineVersion is not set.
/kind bug

I tried to create a machine and got an NPE because I forgot to set the Machine version.

I expected a default value

This is the line throwing the NPE https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/89e8ab62a6ee0de8f6644ae44cb7f224fc202f38/pkg/cloud/services/ec2/instances.go#L126

aaroniscode:
If spec.Version is not set on Machine, the controller will crash with NPE

Fixes #1092

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -37,6 +37,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/filter"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
+	capierrors "sigs.k8s.io/cluster-api/errors"
 	"sigs.k8s.io/cluster-api/util"
 )
 
@@ -130,6 +131,13 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*infrav1.Instance, 
 	if scope.AWSMachine.Spec.AMI.ID != nil {
 		input.ImageID = *scope.AWSMachine.Spec.AMI.ID
 	} else {
+		if scope.Machine.Spec.Version == nil {
+			err := errors.New("Either AWSMachine's spec.ami.id or Machine's spec.version must be defined")
+			scope.SetFailureReason(capierrors.CreateMachineError)
+			scope.SetFailureMessage(err)
+			return nil, err
+		}
+
 		imageLookupOrg := scope.AWSMachine.Spec.ImageLookupOrg
 		if imageLookupOrg == "" {
 			imageLookupOrg = scope.AWSCluster.Spec.ImageLookupOrg


controllers/machine_webhook.go
```diff
@@ -0,0 +1,67 @@
...
+	if err != nil {
+		return admission.Errored(http.StatusBadRequest, err)
+	}
+
+	if m.Spec.InfrastructureRef.Kind == "AWSMachine" && m.Spec.Version == nil {
+		err := apierrors.NewInvalid(clusterv1.GroupVersion.WithKind("Machine").GroupKind(), m.Name, field.ErrorList{
+			field.Required(field.NewPath("spec", "version"), "is required"),
+		})
+		return admission.Denied(err.Error())
+	}
```
detiber: This should probably be added to the existing Machine Webhook here: https://github.com/kubernetes-sigs/cluster-api/blob/83c7edf23e727d89880f3adc697a9db41488e98c/api/v1alpha3/machine_webhook.go#L65-L102

ncdc: Version is optional for CAPI but currently required for CAPA to do the ami lookup (if you omit the image id). I’m not sure it’s appropriate right now to require it in the CAPI webhook but it’s ok to put it in for CAPA. Wdyt?

controllers/machine_webhook.go
```diff
@@ -0,0 +1,67 @@
...
+	client  client.Client
+	decoder *admission.Decoder
+}
+
+// Handle validates Machine admission requests.
+func (v *MachineValidator) Handle(ctx context.Context, req admission.Request) admission.Response {
+	m := &clusterv1.Machine{}
+
+	err := v.decoder.Decode(req, m)
+	if err != nil {
```
vincepri: Nit, prefer using `if err := ...; err != nil` to scope the `err` variable just to the if block

controllers/machine_webhook.go
vincepri: +1, it should be kept in infrastructure providers

aaroniscode: Note: I think I need to update the webhook to only require version if `spec.ami.id` is not set.

controllers/machine_webhook.go
aaroniscode: thanks for  the feedback @vincepri , updated the `if` block.

aaroniscode:
The update I pushed was only a minor change to scope the `err` variable to the `if` block.

But reviewing this I realized it's a bit more complicated than I initially thought. The current iteration of the PR requires the `Machines`'s version regardless if the `AWSMachine` specifies an AMI. This would prevent a valid configuration like the snipped below:

```
apiVersion: cluster.x-k8s.io/v1alpha3
kind: Machine
metadata:
  name: test
  namespace: default
spec:
  bootstrap:
    configRef:
      <snipped>
  clusterName: test
  infrastructureRef:
    kind: AWSMachine
    name: test
    namespace: default
---
kind: AWSMachine
metadata:
  name: test
  namespace: default
spec:
  instanceType: t2.medium
  sshKeyName: default
  ami:
    id: ami-01234567890abcdef
```

It's completely valid to omit the `Machine`'s version if the `AWSMachine` has a `spec.ami.id`.  But checking that in the Webhook would reject a valid `Machine` if the `AWSMachine` object hasn't been persisted. In the example above if we were to check the `AWSMachine`'s `spec.ami.id` the Webhook would reject the `Machine` because the `AWSMachine` doesn't exist yet. We could solve that by changing the order of the resources in the YAML but I'm not sure we want to require that.

I'm starting to wonder if a Webhook is the best way to address this. Right now I see two webhook options and neither seem like a great idea:

Option 1: Require `Machine` version for CAPA in all cases. 
Option 2: Require `Machine` version for CAPA if `AWSMachine` spec.ami.id isn't set.

Thoughts?

ncdc:
I'm good with option 2

ncdc:
@detiber @vincepri I chatted with @aaroniscode a few weeks ago on Slack about this, and given we can't control the order of creation (nor do we want to), and we can't make the link from AWSMachine to Machine until the MachineReconciler has set the owner ref on AWSMachine, we cannot easily validate this.

I want to propose that our short-term solution be to modify the code here:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/24b2057904692380d419db6e8ab56ec35779993c/pkg/cloud/services/ec2/instances.go#L143

to check if scope.Machine.Spec.Version is nil. If it's nil, set scope.AWSMachine.Status.FailureReason/Message and return an error. This should result in the failure being propagated to the Machine, and both will be considered in terminal failure.

WDYT?

aaroniscode:
@ncdc I've been a little busy but I have some code I just need to finish testing and push that does what we discussed.

The outstanding question is whether we still want a Webhook to reject Machine without `spec.version` if the AWSMachine is already persisted and we can verify it's missing a `spec.ami`

I've got that webhook coded and working. So wasn't sure if we want to prevent this when it's possible, or say NO, we don't want a webhook because we can't prevent all the scenarios with it.

Thoughts?

ncdc:
@aaroniscode sorry, we've gone back and forth on this a few times - what is your soon-to-be-pushed code going to do?

I don't have any objections to a webhook that rejects a machine with a nil version when we are able to trace over to a preexisting AWSMachine.

aaroniscode:
> @aaroniscode sorry, we've gone back and forth on this a few times - what is your soon-to-be-pushed code going to do?

sorry, could have phrased it better. I think it does exactly what you describe above. Fail the AWSMachine if we get to the point of looking up the AMI and it's empty.

The plan, assuming I don't get pulled into anything, is to carve out time tomorrow (Wednesday) to test and update the PR.  Really want to close this out as it's been open for a month. Thanks for all your help!

ncdc:
@aaroniscode sgtm, any updates? 😄

aaroniscode:
> @aaroniscode sgtm, any updates? 😄

yes!  I was a bit optimistic on my timing, but got it completed.

controllers/machine_webhook.go
```diff
@@ -0,0 +1,101 @@
...
+	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/cluster-api/controllers/external"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// +kubebuilder:webhook:path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-machine,mutating=false,failurePolicy=fail,groups=cluster.x-k8s.io,resources=machines,verbs=create,versions=v1alpha3,name=providervalidation.machine.infrastructure.cluster.x-k8s.io
+
+// MachineWebhook handles Machine admission webhook requests.
```
vincepri: This webhook looks quite different from the other webhooks we have around.

I'm wondering if there is a way to use the nice building blocks offered by kubebuilder by using a type alias like
```golang
// +kubebuilder:object:root=true
type CoreMachine struct {
	clusterv1.Machine
}
```

and have the registration happen as usual with:

```golang
func (r *CoreMachine) SetupWebhookWithManager(mgr ctrl.Manager) error {
	return ctrl.NewWebhookManagedBy(mgr).
		For(r).
		Complete()
}
```

aaroniscode: Hi Vince!  I'm 100% in agreement and started the implementation using the method above. The biggest challenge is that the Validating webhook methods don't allow for client injection and we need to use the client to retrieve the AWSMachine to see if it has `spec.ami.id` set. If so, we can approve a Machine without a `spec.version`

Here are some related controller runtime issues:
https://github.com/kubernetes-sigs/controller-runtime/issues/711
https://github.com/kubernetes-sigs/controller-runtime/issues/451

randomvariable: This is intriguing. I'm not sure about using a client to fetch other objects in webhook validation. There's two side effects:
a) Someone doing a `kubectl apply -f` would need to order their yaml files, and that's weird, and affects the ability to use third-party tools to apply manifests.
b) Performance impacts. Validation webhooks should be for fairly lightweight checks.

May be better to defer checking the version to the controller, and just error the AWSMachine created to fix the NPE.

aaroniscode: Yeah, we've gone back and forth whether a webhook is the right solution here. Tried to solve it completely using webhooks and it's not possible.

So the question is whether we want a Webhook to catch the issue when possible. No one needs to order their YAML files ... if the order is a certain way, the Webhook can catch it. If not it fails.

Not sure of the performance impact.  I've posed it as an outstanding question (earlier in the PR comments) if we want a webhook or if we should just let the controller catch it.  I'm good with either ... just looking for feedback. Thanks for reviewing.

ncdc: It should be reading from the cache instead of doing a live lookup, so I wouldn't expect any network latency to impact things here.

However, to keep things simple, how about we just merge the code you have that fixes the NPE? You can hold on to a copy of the webhook code and we can reevaluate it later. WDYT?

aaroniscode: Absolutely!  since the webhook can't really prevent this in all cases, maybe it's just not the right solution

pkg/cloud/services/ec2/instances.go
```diff
@@ -130,6 +131,13 @@ func (s *Service) CreateInstance(scope *scope.MachineScope) (*infrav1.Instance,
 	if scope.AWSMachine.Spec.AMI.ID != nil {
 		input.ImageID = *scope.AWSMachine.Spec.AMI.ID
 	} else {
+		if scope.Machine.Spec.Version == nil {
+			err := errors.New("Either AWSMachine's spec.ami.id or Machine's spec.version must be defined")
+			scope.SetFailureReason(capierrors.CreateMachineError)
+			scope.SetFailureMessage(err)
+			return nil, err
```
detiber: It doesn't look like returning an error here will generate an event. Errors returned from `CreateInstance` here are wrapped and returned in the MachineControllers `getOrCreate` method, which are then returned from the MachineControllers `reconcileNormal` method, which is directly returned from the primary `Reconcile` method.

detiber:
Question around the lack of Event in this workflow, otherwise lgtm

pkg/cloud/services/ec2/instances.go
ncdc: We don't strictly need an event because it will be visible in status.failure*

aaroniscode: probably would benefit from an event. I'll add one unless there is an objection

ncdc: Already merged, but let's not 😄

detiber: I figured we can take the conversation asynchronous :joy:

PR #1455, opened 2019-12-28
🐛 Persist AWSCluster/AWSMachine finalizers immediately to prevent orphaned AWS resources

### LINKED ISSUE #1454: Creating and deleting Cluster/AWSCluster quickly results orphaned AWS resources
/kind bug

```
kubectl apply -f examples/_out/cluster.yaml
sleep 1
kubectl delete -f examples/_out/cluster.yaml
```

The Cluster/AWSCluster is created and deleted but the VPC still exists with Subnets, Routes, InternetGateway, NAT, Bastion host, etc.

The VPC to be deleted with the AWSCluster object.

This appears to occur because the finalizer is not persisted until the end of the reconciliation loop and we wait for some AWS resources like the NAT gateway which take a while.

**Environment:**

- Cluster-api-provider-aws version: `master`
- Kubernetes version: (use `kubectl version`): any

aaroniscode:
If Cluster/AWSCluster resources are created and deleted relatively quickly (seconds up to a minute or so) the AWS resources in the VPC are orphaned as the AWSCluster is deleted without destroying the resources.

Fixes #1454

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -84,7 +84,7 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 		return reconcile.Result{}, errors.Errorf("failed to create scope: %+v", err)
 	}
 
-	// Always close the scope when exiting this function so we can persist any AWSMachine changes.
+	// Always close the scope when exiting this function so we can persist any AWSCluster changes.
 	defer func() {
 		if err := clusterScope.Close(); err != nil && reterr == nil {
 			reterr = err
@@ -135,6 +135,10 @@ func reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error)
 	// If the AWSCluster doesn't have our finalizer, add it.
 	if !util.Contains(awsCluster.Finalizers, infrav1.ClusterFinalizer) {
 		awsCluster.Finalizers = append(awsCluster.Finalizers, infrav1.ClusterFinalizer)
+		// Register the finalizer immediately to avoid orphaning AWS resources on delete
+		if err := clusterScope.PatchObject(); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
 	ec2Service := ec2.NewService(clusterScope)
--- a/controllers/awsmachine_controller_test.go
+++ b/controllers/awsmachine_controller_test.go
@@ -28,6 +28,7 @@ import (
 	. "github.com/onsi/gomega/gstruct"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog"
 	"k8s.io/utils/pointer"
@@ -60,9 +61,16 @@ var _ = Describe("AWSMachineReconciler", func() {
 		klog.SetOutput(GinkgoWriter)
 		var err error
 
+		awsMachine := &infrav1.AWSMachine{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "test",
+			},
+			Spec: infrav1.AWSMachineSpec{},
+		}
+
 		ms, err = scope.NewMachineScope(
 			scope.MachineScopeParams{
-				Client: fake.NewFakeClient(),
+				Client: fake.NewFakeClient([]runtime.Object{awsMachine}...),
 				Cluster: &clusterv1.Cluster{
 					Status: clusterv1.ClusterStatus{
 						InfrastructureReady: true,
@@ -76,7 +84,7 @@ var _ = Describe("AWSMachineReconciler", func() {
 					},
 				},
 				AWSCluster: &infrav1.AWSCluster{},
-				AWSMachine: &infrav1.AWSMachine{},
+				AWSMachine: awsMachine,
 			},
 		)
 		Expect(err).To(BeNil())
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -188,9 +188,14 @@ func (s *ClusterScope) ListOptionsLabelSelector() client.ListOption {
 	})
 }
 
+// PatchObject persists the cluster configuration and status.
+func (s *ClusterScope) PatchObject() error {
+	return s.patchHelper.Patch(context.TODO(), s.AWSCluster)
+}
+
 // Close closes the current scope persisting the cluster configuration and status.
 func (s *ClusterScope) Close() error {
-	return s.patchHelper.Patch(context.TODO(), s.AWSCluster)
+	return s.PatchObject()
 }
 
 // AdditionalTags returns AdditionalTags from the scope's AWSCluster. The returned value will never be nil.
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -278,6 +278,10 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 	if !util.Contains(machineScope.AWSMachine.Finalizers, infrav1.MachineFinalizer) {
 		machineScope.V(1).Info("Adding Cluster API Provider AWS finalizer")
 		machineScope.AWSMachine.Finalizers = append(machineScope.AWSMachine.Finalizers, infrav1.MachineFinalizer)
+		// Register the finalizer immediately to avoid orphaning AWS resources on delete
+		if err := machineScope.PatchObject(); err != nil {
+			return reconcile.Result{}, err
+		}
 	}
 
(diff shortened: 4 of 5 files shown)


pkg/cloud/scope/cluster.go
```diff
@@ -188,8 +188,8 @@ func (s *ClusterScope) ListOptionsLabelSelector() client.ListOption {
 	})
 }
 
-// Close closes the current scope persisting the cluster configuration and status.
-func (s *ClusterScope) Close() error {
+// Update persists the cluster configuration and status.
+func (s *ClusterScope) Update() error {
```
vincepri: Is the change to `Update()` necessary? Maybe add another method called `PatchObject` or similar?

controllers/awscluster_controller.go
```diff
@@ -135,6 +135,9 @@ func reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error)
 	// If the AWSCluster doesn't have our finalizer, add it.
 	if !util.Contains(awsCluster.Finalizers, infrav1.ClusterFinalizer) {
 		awsCluster.Finalizers = append(awsCluster.Finalizers, infrav1.ClusterFinalizer)
+		if err := clusterScope.Update(); err != nil {
```
vincepri: Why do we need to call this here? When this function returns, the caller calls `Close()` on defer as well

aaroniscode: Unless I'm mistaken, because the AWS calls take a long time (seconds to minutes) the finalizer isn't yet persisted and you can delete the AWS cluster during this time. This immediately persists the finalizer before the longer running AWS calls.

pkg/cloud/scope/cluster.go
aaroniscode: It's not a necessary change, but since I was using the `Close` function to update the AWSCluster with the finalizer, it didn't seem like the right name or the right description of what it was achieving.

Can revert if that makes the most sense. But probably makes sense to determine if the call to clusterScope.Close() is correct and necessary in `awscluster_controller.go` first.

controllers/awscluster_controller.go
aaroniscode: @vincepri also I opened #1454 that describes the issue

vincepri: Makes sense, let's add a comment to that effect

pkg/cloud/scope/cluster.go
vincepri: Close makes sense to be the last call, let's move the patching to a `PatchObject()` method, which you can call directly, and keep `Close` to be the last call. Close calls PatchObject.

aaroniscode: @vincepri I pushed an update. Let me know if I was able to capture what you were looking for in terms of creating a `PatchObject` method and having `Close` call `PatchObject`

controllers/awscluster_controller.go
aaroniscode: @vincepri comment added. Feel free to wordsmith it differently if it's not clear.

aaroniscode:
I believe all review comments are resolved.

controllers/awscluster_controller.go
```diff
@@ -135,6 +135,10 @@ func reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error)
 	// If the AWSCluster doesn't have our finalizer, add it.
 	if !util.Contains(awsCluster.Finalizers, infrav1.ClusterFinalizer) {
 		awsCluster.Finalizers = append(awsCluster.Finalizers, infrav1.ClusterFinalizer)
+		// Register the finalizer immediately to avoid orphaning AWS resources on delete
+		if err := clusterScope.PatchObject(); err != nil {
+			return reconcile.Result{}, err
+		}
```
vincepri: Do we want/need to return after here and requeue? 

/cc @ncdc

ncdc: The PatchHelper does not mutate the object being patched in memory, so any changes such as to resourceVersion don't come through. We could patch to add the finalizer and return immediately (no requeue). Because it's a change, the object would be added to the queue again fairly quickly.

On the other hand, it's _probably_ safe not to do that, and to just continue with what we already have in memory, as the deferred patch on scope closure will only be patching things we changed in this reconcile iteration.

ncdc: @aaroniscode have you tested these changes?

aaroniscode: @ncdc yes, I've tested quite a bit and re-tested today. Creating and deleting the cluster no longer orphans AWS resources. And there doesn't appear to be any issue with calling PatchHelper and then continuing with the reconcile.

But I am not an expect on controller runtime and kubebuilder and if there is a potential issue with by not immediately returning and requeuing like @vincepri suggests. Let me know if you think I should consider that route. Thanks!

ncdc: If you haven't run into issues, we can proceed. If we encounter conflict errors in the future, it potentially may come from here (but I think we're ok).

ncdc:
Should we do the same thing for AWSMachine?

aaroniscode:
> Should we do the same thing for AWSMachine?

`AWSMachine` does not have the same issue. Right after the finalizer is added, it checks for the Bootstrap data, which doesn't yet exist as the it needs to wait for the bootstrap controller and it returns and persists the finalizer.

ncdc:
Better to be safe, I think

aaroniscode:
> Better to be safe, I think

ok, thanks @ncdc ... I'll add it to AWSMachine as well. Assuming we want it in this PR. I'll add it and force-push. Just let me know if you want it in a separate PR

ncdc:
This one's good, thanks.

randomvariable:
/held
Awaiting one more addition.

@aaroniscode remove hold when ready.

ncdc:
/hold
for realz 😛

aaroniscode:
updated the PR to include AWSMachine, so both now immediately persist the finalizer after adding

/unhold

aaroniscode:
not sure why this test failed ... i don't see an obvious reason in the code going to try to re-run

/retest

aaroniscode:
> not sure why this test failed ... i don't see an obvious reason in the code going to try to re-run

so it turned out that having the `AWSMachine` controller call the patchHelper after adding the finalizer caused about half of the controller tests to fail. This was because the fake client couldn't find the `AWSMachine` to patch it. So I added a fake `AWSMachine` to the fake client and the tests are working again.

aaroniscode:
ready for review again

controllers/awsmachine_controller_test.go
```diff
@@ -60,9 +61,16 @@ var _ = Describe("AWSMachineReconciler", func() {
 		klog.SetOutput(GinkgoWriter)
 		var err error
 
+		awsmachine := &infrav1.AWSMachine{
```
ncdc:
```suggestion
		awsMachine := &infrav1.AWSMachine{
```

controllers/awsmachine_controller_test.go
```diff
@@ -60,9 +61,16 @@ var _ = Describe("AWSMachineReconciler", func() {
...
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "",
+			},
+			Spec: infrav1.AWSMachineSpec{},
+		}
+
```
ncdc: I can't comment on line 79/87, but it would probably be worthwhile to use this `awsMachine` as the value for `AWSMachine` in `MachineScopeParams`.

controllers/awsmachine_controller_test.go
```diff
@@ -60,9 +61,16 @@ var _ = Describe("AWSMachineReconciler", func() {
 		klog.SetOutput(GinkgoWriter)
```
vincepri: These shouldn't be in BeforeEach, they're applied globally

controllers/awsmachine_controller_test.go
```diff
@@ -60,9 +61,16 @@ var _ = Describe("AWSMachineReconciler", func() {
...
 		var err error
 
+		awsmachine := &infrav1.AWSMachine{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "",
```
vincepri: Why empty name?

controllers/awsmachine_controller_test.go
aaroniscode: good catch, fixed.

controllers/awsmachine_controller_test.go
aaroniscode: agreed ... updated.

controllers/awsmachine_controller_test.go
aaroniscode: no good reason ... I think I saw another example test that used this. Updated it to "test"

controllers/awsmachine_controller_test.go
vincepri: This can be cleaned up later

PR #1456, opened 2019-12-29
⚠️ ELB uses separate security group

### LINKED ISSUE #1452: Wrong security group assigned to load balancer
/kind bug

I modified the ELB listening port to 443 which by default is 6443. The control plane came up but I was not able to connect to the cluster using the kubeconfig. I verified that kubeconfig had the correct port value.
Further investigation showed that wrong security group is assigned to the load balancer. Today it reuses the control plane security group and it works in general because it uses port 6443.

I expected the kubeconfig connection to cluster work seamlessly.

Probably we should add a separate security group for load balancer itself.

**Environment:**

- Cluster-api-provider-aws version: 0.4.6

aaroniscode:
Currently the ELB provisioned for the Kubernetes API server shares the security group with the control plane nodes. This leverages the existing LB security group that was created but never used. It also exposes the external port as set in `Clusters`'s `spec.clusterNetwork.apiServerPort`

Fixes #1452

### DIFF
--- a/pkg/cloud/services/ec2/securitygroups.go
+++ b/pkg/cloud/services/ec2/securitygroups.go
@@ -62,6 +62,7 @@ func (s *Service) reconcileSecurityGroups() error {
 	// Declare all security group roles that the reconcile loop takes care of.
 	roles := []infrav1.SecurityGroupRole{
 		infrav1.SecurityGroupBastion,
+		infrav1.SecurityGroupAPIServerLB,
 		infrav1.SecurityGroupLB,
 		infrav1.SecurityGroupControlPlane,
 		infrav1.SecurityGroupNode,
@@ -456,6 +457,16 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
 					s.scope.SecurityGroups()[infrav1.SecurityGroupNode].ID,
 					s.scope.SecurityGroups()[infrav1.SecurityGroupControlPlane].ID,
 				},
+			},
+		}, nil
+	case infrav1.SecurityGroupAPIServerLB:
+		return infrav1.IngressRules{
+			{
+				Description: "Kubernetes API",
+				Protocol:    infrav1.SecurityGroupProtocolTCP,
+				FromPort:    int64(s.scope.APIServerPort()),
+				ToPort:      int64(s.scope.APIServerPort()),
+				CidrBlocks:  []string{anyIPv4CidrBlock},
 			},
 		}, nil
 	case infrav1.SecurityGroupLB:
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -253,7 +253,7 @@ func (s *Service) getAPIServerClassicELBSpec() (*infrav1.ClassicELB, error) {
 			HealthyThreshold:   5,
 			UnhealthyThreshold: 3,
 		},
-		SecurityGroupIDs: []string{s.scope.SecurityGroups()[infrav1.SecurityGroupControlPlane].ID},
+		SecurityGroupIDs: []string{s.scope.SecurityGroups()[infrav1.SecurityGroupAPIServerLB].ID},
 		Attributes: infrav1.ClassicELBAttributes{
 			IdleTimeout: 10 * time.Minute,
 		},
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -307,6 +307,9 @@ var (
 	// SecurityGroupControlPlane defines a Kubernetes control plane node role
 	SecurityGroupControlPlane = SecurityGroupRole("controlplane")
 
+	// SecurityGroupAPIServerLB defines a Kubernetes API Server Load Balancer role
+	SecurityGroupAPIServerLB = SecurityGroupRole("apiserver-lb")
+
 	// SecurityGroupLB defines a container for the cloud provider to inject its load balancer ingress rules
 	SecurityGroupLB = SecurityGroupRole("lb")
 )
--- a/pkg/cloud/services/ec2/securitygroups_test.go
+++ b/pkg/cloud/services/ec2/securitygroups_test.go
@@ -106,6 +106,41 @@ func TestReconcileSecurityGroups(t *testing.T) {
 
 				////////////////////////
 
+				securityGroupAPIServerLb := m.CreateSecurityGroup(gomock.Eq(&ec2.CreateSecurityGroupInput{
+					VpcId:       aws.String("vpc-securitygroups"),
+					GroupName:   aws.String("test-cluster-apiserver-lb"),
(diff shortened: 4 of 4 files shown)


aaroniscode:
Took a first stab at this in response to #1452.  Interested in feedback if this is the preferred approach.

detiber:
@aaroniscode we'll need an additional new Security Group for this change, the existing LB that you are re-purposing here is meant for use by the AWS cloud-provider integration (hence the different ownership tag used).

aaroniscode:
> @aaroniscode we'll need an additional new Security Group for this change, the existing LB that you are re-purposing here is meant for use by the AWS cloud-provider integration (hence the different ownership tag used).

@detiber I noticed the comment that it was meant for use by the AWS cloud-provider integration but it didn't appear to be used. When testing the AWS cloud-provider created new security groups for each load balancer. So I can learn can you point to me where I missed how this is used by the AWS cloud provider.

And I can look at creating a new k8s API load balancer security group.

detiber:
> @detiber I noticed the comment that it was meant for use by the AWS cloud-provider integration but it didn't appear to be used. When testing the AWS cloud-provider it created new security groups for each load balancer. So I can learn, can you point to me where I missed how this is used by the AWS cloud provider?

It was introduced in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/706, there might be additional context as to why in the comments on that PR

aaroniscode:
@detiber I updated the PR based on your feedback. I left the LB that we create for the AWS cloud-provider integration alone and created a new LB just for the K8s API Server.

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -253,7 +253,7 @@ func (s *Service) getAPIServerClassicELBSpec() (*infrav1.ClassicELB, error) {
 			HealthyThreshold:   5,
 			UnhealthyThreshold: 3,
 		},
-		SecurityGroupIDs: []string{s.scope.SecurityGroups()[infrav1.SecurityGroupControlPlane].ID},
+		SecurityGroupIDs: []string{s.scope.SecurityGroups()[infrav1.SecurityGroupAPIServerLB].ID},
```
vincepri: Shouldn't this have both security groups?

pkg/cloud/services/ec2/securitygroups.go
```diff
@@ -458,6 +459,23 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
...
+				ToPort:      int64(s.scope.APIServerPort()),
+				CidrBlocks:  []string{anyIPv4CidrBlock},
+			},
+			{
+				Description: "Path MTU Discovery",
+				Protocol:    infrav1.SecurityGroupProtocolICMP,
+				FromPort:    3,
+				ToPort:      4,
+				CidrBlocks:  []string{anyIPv4CidrBlock},
+			},
```
detiber: This makes me wonder if we should have a rule to allow PMTU for the other security groups (other than `SecurityGroupLB`) as well.

pkg/cloud/services/elb/loadbalancer.go
detiber: No, Traffic is allowed to the Control Plane members through an open egress policy on `SecurityGroupAPIServerLB` and the ingress policy allowing traffic to `SecurityGroupControlPlane` from this security group.

pkg/cloud/services/ec2/securitygroups.go
```diff
@@ -374,11 +375,11 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
...
-				Description: "Kubernetes API",
-				Protocol:    infrav1.SecurityGroupProtocolTCP,
-				FromPort:    6443,
-				ToPort:      6443,
-				CidrBlocks:  []string{anyIPv4CidrBlock},
+				Description:            "Kubernetes API",
+				Protocol:               infrav1.SecurityGroupProtocolTCP,
+				FromPort:               6443,
+				ToPort:                 6443,
+				SourceSecurityGroupIDs: []string{s.scope.SecurityGroups()[infrav1.SecurityGroupAPIServerLB].ID},
```
detiber: This might effect how the Talos folks are doing bootstrapping. tagging in @rsmitty

aaroniscode: Line 382 is a mistake. It should continue to be open to `anyIPv4CidrBlock`.  Not sure what I was thinking at the time. Will change it back.

pkg/cloud/services/elb/loadbalancer.go
randomvariable: In addition, because the ELB is externally facing with a public IP, they will receive NAT'd traffic from the control plane instances, which although technically continue to be in the datacentre appear to the ELB as internet originating traffic.

Different case for internal facing ELBs.

pkg/cloud/services/ec2/securitygroups.go
aaroniscode: reverted the mistaken change on line 382.

pkg/cloud/services/ec2/securitygroups.go
aaroniscode: I think you are correct @detiber  and adding PTMU in other security groups starts to widen the scope of this PR. I've removed PMTU for now and will include that in a separate PR.

pkg/cloud/services/elb/loadbalancer.go
aaroniscode: Tested both external and internal ELB. With the current inbound rule 0.0.0.0/0 it doesn't matter if the traffic comes from internet originating traffic or internal, the security group rule allows access to the API server.

aaroniscode:
I believe all review comments have been resolved. Thanks all!  Removed the Path MTU for now as it probably belongs in more than just the k8s LB security group and wanted to keep this PR focused. Will open a separate PR for Path MTU shortly.

Also tested both Internet-facing and internal ELB.

ncdc:
@aaroniscode what port did you test for the ELBs?

aaroniscode:
> @aaroniscode what port did you test for the ELBs?

6443

ncdc:
Ok. Think there'd be any issue with 443?

aaroniscode:
> Ok. Think there'd be any issue with 443?

I tested with 443 a week or so back by setting in `Cluster` object:
```
spec:
  clusterNetwork:
    apiServerPort: 443
```

vincepri:
Note for later: given that we're now adding a new security group for the ELB, we should probably consider changing this line https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1456/files#diff-b0ffc0997fd2bffcad9bd812a56adccbR382 that allows everyone to access the control plane port, regardless of network. We can probably make sure that the control plane security group is allowed to talk to the hardcoded port 6443, and the new ELB security group.

/approve

ncdc:
@vincepri we'll never come back to your note unless we file an issue - will you do that?

vincepri:
Will do yeah

PR #1462, opened 2020-01-03
🏃 Add automation to test the deletion of EC2 instance referenced by Machine object

sirao:
added test which deletes a EC2 instance which was referenced in machine object of
machine deployment.

After deletion, test verifies that the machine is in Failed state


test/e2e/aws_test.go
```diff
@@ -368,8 +368,79 @@ var _ = Describe("functional tests", func() {
...
+		func() bool {
+			if err := kindClient.Get(context.TODO(), apimachinerytypes.NamespacedName{Namespace: namespace, Name: machineName}, machine); err != nil {
+				return false
+			}
+			return machine.Status.Phase == string(phase)
+		}, 10*time.Minute, 15*time.Second,
+	).Should(BeTrue())
+}
+
+func terminateEc2Instance(instanceId string) {
```
vincepri:
```suggestion
func terminateInstance(instanceId string) {
```

Nit, given that we're in AWS land

sirao: renamed the function name

PR #1468, opened 2020-01-08
:sparkles: Adds test case to verify High Availability

bhargavmad:
Steps followed:
 - Create 2 ControlPlane machines in different AZs
 - Create a MachineDeployment in az2 and deploy nginx service
 - Now, Delete the ControlPlane machine in az2
 - Verify if the nginx service is still accessible. This proves
that the ControlPlane machine in az1 is still able to communicate
with the machines in az2

**Remarks**
_Please let me know if this approach really tests HA_

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -113,7 +113,7 @@ var _ = Describe("functional tests", func() {
 		Expect(err).NotTo(HaveOccurred())
 
 		setup.namespace = "test-" + util.RandomString(6)
-		createNamespace(setup.namespace)
+		createNamespace(setup.namespace, kindClient)
 
 		setup.clusterName = "test-" + util.RandomString(6)
 		setup.awsClusterName = "test-infra-" + util.RandomString(6)
@@ -124,7 +124,7 @@ var _ = Describe("functional tests", func() {
 		setup.machineDeploymentName = "test-capa-md" + util.RandomString(6)
 		setup.awsMachineTemplateName = "test-infra-capa-mt" + util.RandomString(6)
 		setup.initialReplicas = 2
-		setup.instanceType = "t3.large"
+		setup.instanceType = "m5.large"
 		setup.multipleAZ = false
 	})
 
@@ -256,8 +256,8 @@ var _ = Describe("functional tests", func() {
 			Expect(waitForMachineDeploymentRunning(setup.namespace, deployment2)).To(BeFalse())
 			eventList := getEvents(setup.namespace)
 			subnetError := "Failed to create instance: failed to run instance: InvalidSubnetID.NotFound: " +
-				"The subnet ID 'notcreated' does not exist"
-			Expect(isErrorEventExists(setup.namespace, deployment1, "FailedCreate", fmt.Sprintf(subnetError, setup.subnetId), eventList)).To(BeTrue())
+				"The subnet ID '%s' does not exist"
+			Expect(isErrorEventExists(setup.namespace, deployment1, "FailedCreate", fmt.Sprintf(subnetError, *setup.subnetId), eventList)).To(BeTrue())
 
 			By("Create new AwsMachineTemplate with correct Subnet ID and update its name in MachineDeployment")
 			sess = getSession()
@@ -328,8 +328,7 @@ var _ = Describe("functional tests", func() {
 			createMachineDeployment(setup)
 
 			By("Deleting a worker node machine")
-			deleteMachine(setup.namespace, setup.machineDeploymentName)
-			time.Sleep(10 * time.Second)
+			deleteMachineFromDeployment(setup.namespace, setup.machineDeploymentName)
 
 			waitForMachineDeploymentRunning(setup.namespace, setup.machineDeploymentName)
 
@@ -426,6 +425,39 @@ var _ = Describe("functional tests", func() {
 			Expect(len(machines.Items)).Should(BeNumerically(">", 0))
 			terminateInstance(*machines.Items[0].Spec.ProviderID)
 			verifyMachinePhase(setup.namespace, machines.Items[0].Name, clusterv1.MachinePhaseFailed)
+
+			By("Deleting the Cluster")
+			deleteCluster(setup.namespace, setup.clusterName)
+		})
+	})
+
+	Describe("Cluster should be highly available", func() {
+		It("Control plane should be functional after failure of one master node", func() {
+			By("Creating a workload cluster with single control plane")
+			setup.multipleAZ = true
+			clusterK8sClient := makeSingleControlPlaneCluster(setup)
+
+			By("Creating the second Control Plane Machine in second AZ")
+			awsMachineName := setup.cpAWSMachinePrefix + "-1"
+			bootstrapConfigName := setup.cpBootstrapConfigPrefix + "-1"
+			machineName := setup.cpMachinePrefix + "-1"
+			setup.availabilityZone = availabilityZones[1]
+			createAdditionalControlPlaneMachine(setup, machineName, awsMachineName, bootstrapConfigName)
+
+			By("Creating the third Control Plane Machine in third AZ")
+			awsMachineName = setup.cpAWSMachinePrefix + "-2"
+			bootstrapConfigName = setup.cpBootstrapConfigPrefix + "-2"
+			machineName = setup.cpMachinePrefix + "-2"
+			setup.availabilityZone = availabilityZones[2]
+			createAdditionalControlPlaneMachine(setup, machineName, awsMachineName, bootstrapConfigName)
+
+			By("Deleting Control Plane Machine in second AZ")
+			deleteMachine(machineName, setup.namespace)
+
+			// Create a kubernetes resource on the workload cluster to
+			// test the ability to write to the API server
+			By("Creating a kubernetes resource")
+			createNamespace("dummy", clusterK8sClient)
 
 			By("Deleting the Cluster")
 			deleteCluster(setup.namespace, setup.clusterName)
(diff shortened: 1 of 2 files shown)


randomvariable:
Can you rebase this PR?

test/e2e/aws_test.go
```diff
@@ -368,8 +370,57 @@ var _ = Describe("functional tests", func() {
...
+			machineName := setup.cpMachinePrefix + "-1"
+			setup.availabilityZone = availabilityZones[1]
+			createAdditionalControlPlaneMachine(setup, machineName, awsMachineName, bootstrapConfigName)
+
+			By("Creating the MachineDeployment in second AZ")
+			setup.availabilityZone = availabilityZones[1]
+			setup.subnetId = nil
+			createMachineDeployment(setup)
+
+			By("Creating the LB service")
```
vincepri: This is a little confusing, we should explain that this is a kubernetes service and not an AWS ELB

bhargavmad:
Sorry for the delay. I was out of office. 
Rebased
Addressed initial comment/s

vincepri:
@bhargavmad Are you able to look into the CI failures?

bhargavmad:
> @bhargavmad Are you able to look into the CI failures?

Hi @vincepri, these failures are around for a while now. The [periodic e2e job](https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-cluster-api-provider-aws#periodic-e2e-release-0-4) also fails intermittently. Not sure exactly whats causing them. Need some help here. 
cc: @ncdc

vincepri:
I mentioned it because it seems that the test added as part of this PR is failing.

`functional tests Cluster should be highly available`
```
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e/aws_test.go:412
Timed out after 900.000s.
Expected
    <bool>: false
to be true
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e/aws_test.go:1293
```

bhargavmad:
The tests passed on my local setup. This particular case `functional tests Cluster should be highly available` failed couple of times, passed once and the last attempt failed because prow timed out. I created a PR to increase the timeout. I will rerun the test again, once the timeout increases.

test/e2e/aws_test.go
```diff
@@ -406,6 +408,42 @@ var _ = Describe("functional tests", func() {
...
+
+			By("Creating the nginx service exposed as LB")
+			_, ingressHostname := createLBService(metav1.NamespaceDefault, "test-nginx-svc", clusterK8sClient)
+
+			By("Checking if the service is accessible")
+			pingService(ingressHostname)
+
+			By("Deleting Control Plane Machine in second AZ")
+			deleteMachine(machineName, setup.namespace)
+			time.Sleep(30 * time.Second)
```
detiber: Why the 30 second sleep here, would it be better to update `deleteMachine` to contain an `Eventually` that waits until the Machine is fully deleted?

test/e2e/aws_test.go
```diff
@@ -406,6 +408,42 @@ var _ = Describe("functional tests", func() {
...
+			setup.availabilityZone = availabilityZones[1]
+			createAdditionalControlPlaneMachine(setup, machineName, awsMachineName, bootstrapConfigName)
+
+			By("Creating the MachineDeployment in second AZ")
+			setup.availabilityZone = availabilityZones[1]
+			setup.subnetId = nil
+			createMachineDeployment(setup)
+
+			By("Creating the nginx service exposed as LB")
+			_, ingressHostname := createLBService(metav1.NamespaceDefault, "test-nginx-svc", clusterK8sClient)
```
detiber: This service is going to be deployed on one of the Nodes that are backing the MachineDeployment, so testing against this service will not give us signal as to the resiliency of the Control Plane as much as the MachineDeployment.

test/e2e/aws_test.go
```diff
@@ -406,6 +408,42 @@ var _ = Describe("functional tests", func() {
...
+
+			By("Checking if the service is accessible")
+			pingService(ingressHostname)
```
detiber: Since we are testing Control Plane availability, it might be better to just test the ability to write to the API Server. The type of resource should not matter much, but as long as we can create a resource we know that the API Server (and etcd) are in a good state.

test/e2e/aws_test.go
```diff
@@ -406,6 +408,42 @@ var _ = Describe("functional tests", func() {
...
+
+			By("Deleting Control Plane Machine in second AZ")
+			deleteMachine(machineName, setup.namespace)
```
detiber: Since there are only 2 Machines in this Control Plane, deleting one of the Machines will lead to loss of quorum for etcd. Instead we should test against a 3 Machine Control Plane, where the etcd cluster would be degraded but still have quorum.

bhargavmad:
This is really weird. All the tests passed after increasing timeouts on some of the functions. But git reports it as failure. Please look into [this](https://prow.k8s.io/view/gcs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/1468/pull-cluster-api-provider-aws-e2e/1220653537466257408)

ncdc:
@bhargavmad I looked at the e2e runs from this PR, as well as the periodic e2e runs, and I'm not seeing any real consistency in the test failures across the jobs. I'm guessing your most recent failure was the error in the log from ci-e2e.sh line 80, where it failed to kill the heartbeat process. We can keep trying to see how things go.

bhargavmad:
The tests are very flaky, one or the other test keeps failing every time and i cant seem to find the solution. I thought increasing timeouts will resolve the issue, but it is still failing. Will really appreciate some help here.

test/e2e/aws_test.go
```diff
@@ -431,8 +430,64 @@ var _ = Describe("functional tests", func() {
...
+			By("Creating a kubernetes resource")
+			createDummyResource("dummy-pod", metav1.NamespaceDefault, clusterK8sClient)
+
+			By("Deleting the Cluster")
+			deleteCluster(setup.namespace, setup.clusterName)
+		})
+	})
 })
 
+func createDummyResource(name, namespace string, k8sClient crclient.Client) {
```
detiber: Instead of creating a Pod here, I would probably just create a Namespace to test the functionality of the API server.

test/e2e/aws_test.go
```diff
@@ -431,8 +430,64 @@ var _ = Describe("functional tests", func() {
...
+			awsMachineName = setup.cpAWSMachinePrefix + "-2"
+			bootstrapConfigName = setup.cpBootstrapConfigPrefix + "-2"
+			machineName = setup.cpMachinePrefix + "-2"
+			setup.availabilityZone = availabilityZones[2]
+			createAdditionalControlPlaneMachine(setup, machineName, awsMachineName, bootstrapConfigName)
+
+			By("Creating the MachineDeployment in second AZ")
+			setup.availabilityZone = availabilityZones[1]
+			setup.subnetId = nil
+			createMachineDeployment(setup)
```
detiber: Since you only need to test the availability of the API server here, you can avoid creating a MachineDeployment here.

test/e2e/aws_test.go
```diff
@@ -790,8 +845,15 @@ func createClusterKubeConfigs(tmpDir, namespace, clusterName string) (string, cr
...
-	Expect(err).NotTo(HaveOccurred())
+	var k8sclient crclient.Client
+	Eventually(
+		func() error {
+			var err error
+			k8sclient, err = crclient.New(restConfig, crclient.Options{})
+			return err
+		},
+		5*time.Minute, 10*time.Second,
+	).Should(BeNil())
```
detiber: Since you are testing the result of an error, I would say it is better to use `gomega.Succeed()` here.
```suggestion
	).Should(Succeed())
```

test/e2e/aws_test.go
```diff
@@ -1130,12 +1210,17 @@ func makeAWSMachineTemplate(namespace, name, instanceType string, az, subnetId *
...
+		func() error {
+			return exec.Command(
+				*kubectlBinary,
+				"create",
+				"--kubeconfig="+kubeConfigPath,
+				"-f", manifestPath,
+			).Run()
+		},
+		5*time.Minute, 10*time.Second,
+	).Should(BeNil())
```
detiber:
```suggestion
	).Should(Succeed())
```

detiber: I'm a bit curious as the why we would need to wrap this with an Eventually.

detiber:
It looks like the most recent test failure was related to attempting to delete the Cluster following the HA tests added here.

I'll attempt to dig into the logs a bit later to see if I can tell why it is taking longer than 20 minutes to delete.

test/e2e/aws_test.go
bhargavmad: I have observed many failures in the test-grid while trying to deploy CNI. It could be because the DNS for the API server may not be available.

bhargavmad:
The latest run fails with log
```
I0130 12:03:33.938758       1 awsmachine_controller.go:264] controllers/AWSMachine "msg"="Error state detected, skipping reconciliation" "awsCluster"="test-infra-l49ppd" "awsMachine"="test-infra-capa-mtft3bkv-6kd27" "cluster"="test-fm0hip" "machine"="test-capa-mdyo14ub-787b7568f7-7qv87" "namespace"="test-qrl6my" 
I0130 12:03:59.958947       1 leaderelection.go:281] failed to renew lease capa-system/controller-leader-election-capa: failed to tryAcquireOrRenew context deadline exceeded
E0130 12:03:59.959119       1 main.go:176] setup "msg"="problem running manager" "error"="leader election lost"  
```
It is observed in both capa and capi log files. It failed during Delete cluster step of one of the tests and the subsequent specs fail at `BeforeEach` with the same error ([logs](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/1468/pull-cluster-api-provider-aws-e2e/1222802920391577600/artifacts))

cc: @ncdc @detiber

bhargavmad:
@detiber I have addressed the review comments and the latest run succeeded except for the error in the `ci-e2e.sh` script. Please do have a look and merge if everything looks good.

test/e2e/aws_test.go
```diff
@@ -975,13 +1013,31 @@ func deleteMachine(namespace, machineDeploymentName string) {
...
 	Expect(kindClient.Delete(context.TODO(), machine)).To(Succeed())
+
+	Eventually(
+		func() bool {
+			machine := &clusterv1.Machine{}
+			if err := kindClient.Get(context.TODO(), apimachinerytypes.NamespacedName{Namespace: namespace, Name: name}, machine); err != nil {
+				if apierrors.IsNotFound(err) {
+					return true
+				}
+				return false
```
detiber: This can be simplified to `return apierrors.IsNotFound(err)`

PR #1472, opened 2020-01-10
🏃 Automate cluster creation when vpc limit was already hit

sirao:
Added automation to test creating a cluster when VPC limit was already reached.

This will verify cluster is in provisioning state till the VPC resources
were available and will provisioned during reconciliation after freeing
current VPCs

Executed the test locally and passed

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -24,6 +24,7 @@ import (
 	"io/ioutil"
 	"os/exec"
 	"path"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


test/e2e/aws_test.go
```diff
@@ -369,6 +370,32 @@ var _ = Describe("functional tests", func() {
 		})
 	})
 
+	Describe("Creating cluster after reaching vpc maximum limit", func() {
+		It("Cluster created after reaching vpc limit should be in provisioning", func() {
+			By("Create clusters till vpc limit")
+			sess = getSession()
+			limit := getElasticIPsLimit()
+			var vpcsCreated []string
+			sess = getSession()
```
bhargavmad: Is it required?

sirao: Removed line 379.

PR #1482, opened 2020-01-14
⚠️ Update control plane security group rule to remove 0.0.0.0/0 ingress CIDR

### LINKED ISSUE #1478: Revisit control plane security group rule
/kind cleanup

Given that we're now adding a new security group for the ELB, we should probably consider changing this line https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1456/files#diff-b0ffc0997fd2bffcad9bd812a56adccbR382 that allows everyone to access the control plane port, regardless of network. We can probably make sure that the control plane security group is allowed to talk to the hardcoded port 6443, and the new ELB security group.

aaroniscode:
With the ELB  now using a separate security group, it makes sense to further lock down the ingress rules for the control plane security group and remove the 0.0.0.0/0 ingress CIDR rule. Inbound access to allowed on port 6443 for the 3 security groups:
* Control Plane
* Nodes
* API Server Load Balancer

Fixes #1478

### DIFF
--- a/pkg/cloud/services/ec2/securitygroups.go
+++ b/pkg/cloud/services/ec2/securitygroups.go
@@ -379,7 +379,11 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
 				Protocol:    infrav1.SecurityGroupProtocolTCP,
 				FromPort:    6443,
 				ToPort:      6443,
-				CidrBlocks:  []string{anyIPv4CidrBlock},
+				SourceSecurityGroupIDs: []string{
+					s.scope.SecurityGroups()[infrav1.SecurityGroupAPIServerLB].ID,
+					s.scope.SecurityGroups()[infrav1.SecurityGroupControlPlane].ID,
+					s.scope.SecurityGroups()[infrav1.SecurityGroupNode].ID,
+				},
 			},
 			{
 				Description:            "etcd",
--- a/pkg/cloud/services/ec2/securitygroups_test.go
+++ b/pkg/cloud/services/ec2/securitygroups_test.go
@@ -26,6 +26,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/golang/mock/gomock"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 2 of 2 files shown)


pkg/cloud/services/ec2/securitygroups.go
```diff
@@ -379,7 +379,11 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
 				Protocol:    infrav1.SecurityGroupProtocolTCP,
 				FromPort:    6443,
 				ToPort:      6443,
-				CidrBlocks:  []string{anyIPv4CidrBlock},
+				SourceSecurityGroupIDs: []string{
+					s.scope.SecurityGroups()[infrav1.SecurityGroupAPIServerLB].ID,
+					s.scope.SecurityGroups()[infrav1.SecurityGroupControlPlane].ID,
+					s.scope.SecurityGroups()[infrav1.SecurityGroupNode].ID,
```
vincepri: We'll need `infrav1.SecurityGroupBastion` as well

aaroniscode: > We'll need `infrav1.SecurityGroupBastion` as well

I thought about that when creating the list of Security Groups. While we could add the bastion, I'm not we need it. The bastion can access the Load Balancer if you want to run `kubectl` from it. And you can SSH into the nodes if you want to try to connect directly to a local port.

This is really a question of whether the bastion is just there for outside access or if we want the bastion to serve a greater purpose like for troubleshooting. Thoughts?

vincepri: We can go with this for now and see if there is a need later

vincepri:
LGTM pending a unit test

randomvariable:
/hold
pending unit test

pkg/cloud/services/ec2/securitygroups_test.go
```diff
@@ -280,6 +281,30 @@ func TestReconcileSecurityGroups(t *testing.T) {
 	}
 }
 
+func TestControlPlaneSecurityGroupNotOpenToAnyCIDR(t *testing.T) {
```
chuckha: Nice! test looks great to me.

PR #1488, opened 2020-01-17
✨Add clusterctl labels to CAPA components

wfernandes:
This PR adds clusterctl labels to all CAPA components generated when running `make release-manifests`. It removes the labels from selectors within the controller-manager deployment and services.

Ref https://github.com/kubernetes-sigs/cluster-api/issues/1989

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -9,8 +9,8 @@ namespace: capa-system
 namePrefix: capa-
 
 # Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+commonLabels:
+  cluster.x-k8s.io/provider: "infrastructure-aws"
 
 bases:
 - ../crd


config/default/kustomization.yaml
```diff
@@ -9,8 +9,8 @@ namespace: capa-system
 namePrefix: capa-
 
 # Labels to add to all resources and selectors.
-#commonLabels:
-#  someName: someValue
+commonLabels:
+  cluster.x-k8s.io/provider: "aws"
```
vincepri: either we should standardize to something like `infrastructure-aws` or `aws-infrastructure`

cc @fabriziopandini

wfernandes: Fine with whatever we pick as long as we are consistent here as well: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/725

vincepri: I prefer <type>-<name>, it reads better, and it has a meaning

detiber: One other potential solution we may have is to have two labels, one for a provider name, and one for a provider type. That said, I'm not overly sold that we need that.

fabriziopandini: I prefer a single label/two labels with the same name across all the providers because this makes the discovery simple, but I will adapt to whatever is decided

wfernandes: Ok...Let's do a single label. 
`cluster.x-k8s.io/provider: infrastructure-aws`.

This is consistent formatting with @vincepri suggestions. That is, `controlplane-kubeadm` and `bootstrap-kubeadm` for those components.
The only one that diverges is `cluster-api`. 🤷‍♂

fabriziopandini:
I'm -1 to add infrastructure to the label value
Clustectl uses aws because it is simple and read well when doing `clusterctl init --infrastructure aws`, and I would like to avoid to implement mapping between the two values

wfernandes:
@vincepri Let me know if this looks good. I have a corresponding PR in CAPV that I'd like to unhold. I just wanted to confirm the label format is good.

PR #1494, opened 2020-01-22
✨Adding an auth-proxy sidecar container

noamran:
Enables auth proxy which is deployed as sidecar container in the CAPA manager pod. This is a follow up to https://github.com/kubernetes-sigs/cluster-api/pull/2035 to make sure we are consistent.

### DIFF
--- a/config/default/kustomization.yaml
+++ b/config/default/kustomization.yaml
@@ -28,7 +28,7 @@ patchesStrategicMerge:
   # Protect the /metrics endpoint by putting it behind auth.
   # Only one of manager_auth_proxy_patch.yaml and
   # manager_prometheus_metrics_patch.yaml should be enabled.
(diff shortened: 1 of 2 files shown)


detiber:
@noamran not directly related to this PR, but I think we'll need to do similar with the kubeadm control plane controller deployment as well.

noamran:
> @noamran not directly related to this PR, but I think we'll need to do similar with the kubeadm control plane controller deployment as well.

@detiber I think it's already enabled [here](https://github.com/kubernetes-sigs/cluster-api/blob/master/controlplane/kubeadm/config/default/kustomization.yaml#L33).  Please LMK if I am missing something!

detiber:
> @detiber I think it's already enabled here. Please LMK if I am missing something!

Indeed, nevermind me. Vacation brain still in effect a bit it seems :joy:

PR #1495, opened 2020-01-23
:running: Increase e2e test timeout

detiber:
Increases the e2e test timeout to resolve the timeout errors we are seeing in the periodic jobs.


ncdc:
/lgtm

The prow job config will also need its timeout increased. Would you be able to submit a PR for that too?

detiber:
> The prow job config will also need its timeout increased. Would you be able to submit a PR for that too?

That change was merged yesterday.

PR #1496, opened 2020-01-23
:warning: machines can be provisioned without SSH key pairs

### LINKED ISSUE #1156: Option to launch instances without an SSH key pair

andrewmyhre:
Allows instances to be provisioned with no SSH key pair specified. Organizations which have policies denying the use of SSH key pairs can now set the SSH key name value to "" to launch instances without a key pair set. A nil value still sets the default key pair name.

Fixes #1156

### DIFF
--- a/api/v1alpha2/awscluster_types.go
+++ b/api/v1alpha2/awscluster_types.go
@@ -35,7 +35,7 @@ type AWSClusterSpec struct {
 	Region string `json:"region,omitempty"`
 
 	// SSHKeyName is the name of the ssh key to attach to the bastion host.
-	SSHKeyName string `json:"sshKeyName,omitempty"`
+	SSHKeyName *string `json:"sshKeyName,omitempty"`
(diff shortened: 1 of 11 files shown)


vincepri:
We probably need to have some conversion in place to handle the new feature

detiber:
> We probably need to have some conversion in place to handle the new feature

Ah, good point. We should probably be using the conversion utilities for it, similar to what we are doing here for Machine: https://github.com/kubernetes-sigs/cluster-api/blob/3117d1830b2a1e5f721fa784323ef36c352e360a/api/v1alpha2/conversion.go#L87-L129

andrewmyhre:
Are those files generated?

ncdc:
@andrewmyhre the one Jason linked to is hand-written.

andrewmyhre:
Closed in favor of #1585

PR #1498, opened 2020-01-25
:sparkles: Add extra root volume options

### LINKED ISSUE #1403: Ability to configure node disks
/kind feature

**User Story**
When creating nodes in a cluster, it would be useful to be able to request disk characteristics to control performance, space, and cost. Specifically, it would be good to be able to configure:

* Root disk size (I think this is already possible)
* Disk type (gp2, etc)
* IOPS

johnharris85:
Signed-off-by: John Harris <joharris@vmware.com>

Allows users to specify additional options for a machine's root volume.

Fixes #1403 

**Notes**:
- If a user wants to specify a KMS key, they need to ensure the relevant user / role has use permissions on the key. Should this be documented somewhere? (if so, where?)
- If EncryptionKey is set, we set Encrypted to `true` (even if it was omitted / set to `false`). Thought this was a reasonable assumption, but open to changes there (error if key set but encrypted missing / `false`).


api/v1alpha3/awsmachine_webhook.go
```diff
@@ -43,6 +43,15 @@ var _ webhook.Validator = &AWSMachine{}
 
 // ValidateCreate implements webhook.Validator so a webhook will be registered for the type
 func (r *AWSMachine) ValidateCreate() error {
+	var allErrs field.ErrorList
+
+	if r.Spec.RootVolumeOptions != nil && r.Spec.RootVolumeOptions.Type == "io1" && r.Spec.RootVolumeOptions.IOPS == 0 {
+		allErrs = append(allErrs, field.Required(field.NewPath("spec.rootVolumeOptions.iops"), "iops required if type is 'io1'"))
```
ncdc:
```suggestion
		allErrs = append(allErrs, field.Required(field.NewPath("spec", "rootVolumeOptions", "iops"), "iops required if type is 'io1'"))
```

api/v1alpha3/types.go
```diff
@@ -516,12 +516,35 @@ type Instance struct {
...
+
+	// Type is the type of the root volume (e.g. gp2, io1, etc...).
+	// +optional
+	Type string `json:"type,omitempty"`
+
+	// IOPS is the number of IOPS requested for the disk. Not applicable to all types.
+	// +optional
+	IOPS int64 `json:"iops,omitempty"`
+
+	// Encrypted is whether the volume should be encrypted or not. If EncryptionKey is set it will be used, if not the default AWS key will be used.
```
ncdc: I would move the details about EncryptionKey down to that field

johnharris85:
@vincepri pointed out on Slack this should have some restore logic built in to make sure we don't miss these fields from annotations. Will add that and address Andy's comments later today.

vincepri:
Sounds good! Holding the PR until we have those in place

/hold

pkg/cloud/services/ec2/instances.go
```diff
@@ -358,19 +358,35 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 		}
 	}
 
-	if i.RootDeviceSize != 0 {
+	if i.RootVolumeOptions != nil && i.RootVolumeOptions.Size != 0 {
```
jzhoucliqr: so we'll ignore other options if Size==0 ?

johnharris85: If the size is 0 we don't create the disk so there's nothing for the options to apply to / be relevant to afaik?

jzhoucliqr: so there is no default size? If so, shall we mark size as non-optional?

rudoi: The default size from AWS is actually 8, if that info helps. This 0 check probably originally should have been a nil check with Size as a nil-able field. I forget if there was a reason why I didn't implement it that way the first time.

jzhoucliqr: From ec2 sdk:  	
        // The size of the volume, in GiB.
	//
	// Default: If you're creating the volume from a snapshot and don't specify
	// a volume size, the default is the snapshot size.
	//
	// Constraints: 1-16384 for General Purpose SSD (gp2), 4-16384 for Provisioned
	// IOPS SSD (io1), 500-16384 for Throughput Optimized HDD (st1), 500-16384 for
	// Cold HDD (sc1), and 1-1024 for Magnetic (standard) volumes. If you specify
	// a snapshot, the volume size must be equal to or larger than the snapshot
	// size.
	VolumeSize *int64 `locationName:"volumeSize" type:"integer"`

johnharris85: Fine however we want to implement this, but if you're always going to get a disk maybe we should make it mandatory so there's not ambiguous behavior with the other options etc?

api/v1alpha2/awsmachine_conversion.go
```diff
@@ -19,19 +19,46 @@ package v1alpha2
...
+	}
+	restoreAWSMachineSpec(&restored.Spec, &dst.Spec)
+
+	return nil
+}
+
+func restoreAWSMachineSpec(restored *infrav1alpha3.AWSMachineSpec, dst *infrav1alpha3.AWSMachineSpec) {
+	if restored.RootVolumeOptions != nil {
+		dst.RootVolumeOptions = restored.RootVolumeOptions
+	}
```
vincepri: We can probably use `dst.RootVolumeOptions.DeepCopyInto(...)`

vincepri: Also note that this would overwrite the Manual conversion for `RootDeviceSize` that we have in `Convert_v1alpha2_AWSMachineSpec_To_v1alpha3_AWSMachineSpec`

api/v1alpha3/types.go
```diff
@@ -516,12 +516,35 @@ type Instance struct {
 	// Indicates whether the instance is optimized for Amazon EBS I/O.
 	EBSOptimized *bool `json:"ebsOptimized,omitempty"`
 
-	// Specifies size (in Gi) of the root storage device
-	RootDeviceSize int64 `json:"rootDeviceSize,omitempty"`
+	// Configuration options for the root storage volume
+	RootVolumeOptions *RootVolumeOptions `json:"rootVolumeOptions,omitempty"`
```
vincepri:
```suggestion
	RootVolume *RootVolumeOptions `json:"rootVolume,omitempty"`
```

api/v1alpha3/types.go
```diff
@@ -516,12 +516,35 @@ type Instance struct {
...
 
 	// Specifies ENIs attached to instance
 	NetworkInterfaces []string `json:"networkInterfaces,omitempty"`
 
 	// The tags associated with the instance.
 	Tags map[string]string `json:"tags,omitempty"`
 }
+
+// RootVolumeOptions encapsulates the configuration options for the root volume
+type RootVolumeOptions struct {
```
vincepri:
```suggestion
type RootVolume struct {
```

api/v1alpha3/types.go
vincepri: This also need `//+optional`

pkg/cloud/services/ec2/instances.go
vincepri: I'm fine having the struct and Size default to nil. Let's make sure to copy the default behavior from the SDK to our Size field.

api/v1alpha3/types.go
johnharris85: You want this renaming all over? s/RootVolumeOptions/RootVolume/g

pkg/cloud/services/ec2/instances.go
johnharris85: Still not 100% clear on what our behavior should be though. If I specify 0 (currently, or nil after these changes) for the size, I'm going to get a volume with 8 (according to AWS default? Where does this come from? Want to avoid random hard-coded values or copying too much AWS logic). Should we still apply the other options (encryption etc? assuming yes). What's the difference (from a user POV) between size = 0 and size = nil? Does allowing 0 even make sense (what am I expecting if I put 0 for a size?)

api/v1alpha3/types.go
vincepri: Yeah, `Options` is redundant given that this field is already within `spec:`

pkg/cloud/services/ec2/instances.go
jzhoucliqr: these are good questions. just my understanding, please correct me if i'm wrong: size=0 and size=nil won't make much difference from user's PoV, both will create a volume with the size of the snapshot that backs the AMI. 

provide option to use default size is valuable, because the non-zero size specified must be >= snapshot size. it's quite common that we don't know the underline AMI snapshot size, so use default size is a safe option.

vincepri: Actually, I'll defer to @detiber and @randomvariable for this, the most important part for me is to be consistent with the experience we're providing to our users

johnharris85:
Addressed most of @vincepri comments (except the [size question](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1498#pullrequestreview-351211540)) and rebased.

vincepri:
@johnharris85 Regarding https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1498, I think the only thing I'd add is a kubebuilder validation tag that checks that Size is greater than 0, and remove all the checks where we do `.Size != 0`. 

This makes sure that when a user sets the `rootVolume` struct, the `size` field should never be zero, if the struct is not-nil.

pkg/cloud/services/ec2/instances.go
```diff
@@ -564,6 +573,23 @@ func (s *Service) getImageRootDevice(imageID string) (*string, error) {
...
+		return nil, err
+	}
+
+	if len(output.Images) == 0 {
+		return nil, errors.Errorf("no images returned when looking up ID %q", imageID)
+	}
+
+	return output.Images[0].BlockDeviceMappings[0].Ebs.VolumeSize, nil
+}
+
```
johnharris85: This func is basically identical to the one above (to get the root device name), just returning a different attribute. Thoughts on adjusting the above one (which is only used once) to return size and name to save an AWS call?

config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml
```diff
@@ -397,13 +398,17 @@ spec:
                             type: integer
                           size:
                             description: Size specifies size (in Gi) of the root storage
-                              device
+                              device. Must be greater than the image root snapshot
+                              size or 8 (whichever is greater).
                             format: int64
+                            minimum: 8
```
johnharris85: I know we talked about 10 here @vincepri but 8 seemed like a more sensible guard given that it's the default.

johnharris85:
Discussion w/ @vincepri on Slack around how to handle the size wrt defaults / validation -> https://app.slack.com/client/T09NY5SBT/CD6U2V71N/thread/CD6U2V71N-1580776492.150700.

Addressed all comments I think but left a couple questions in for feedback, can squash when ready.

johnharris85:
Failure looks like generation out of date, but `make verify` succeeds on my machine...? (and `make generate` doesn't generate any changes?) Am I missing something?

ncdc:
@johnharris85 try to rebase on top of master

vincepri:
```
api/v1alpha3/awsmachine_webhook_test.go:27: File is not `gofmt`-ed with `-s` (gofmt)
		name       string
api/v1alpha2/awsmachine_conversion_test.go:154: File is not `gofmt`-ed with `-s` (gofmt)
}
```

@johnharris85 seems linting errors now

randomvariable:
Should probably queue this one after #1525 so we get the object fuzzer, and then this will then need rebasing I think.

johnharris85:
OK rebased now #1525 landed. Fuzzer found a couple bugs which I also fixed.
- `AWSClusterStatus` conversion funcs were using the auto conversions for `Instance`, changed to using the manual ones (with the additional volume options)
- `AWSCluster` conversion func needed to restore the bastion if present (from annotation)

randomvariable:
/unhold
as we have the restore logic in place.

randomvariable:
Should merge cleanly via 3-way, but merge after #1535

ncdc:
@johnharris85 could you please rebase & resolve conflicts - thanks!

pkg/cloud/services/ec2/instances.go
detiber: +1 to consolidating the calls. This can also be handled as a followup.

PR #1504, opened 2020-01-30
Fix to avoid updating tags when there is no change in them and add unit tests

### LINKED ISSUE #1499: Tags on resources are being updated during reconciliation even if those tags are not changed.
/kind bug

In our environment, we are noticing that tags on some AWS resources are being updated even if there is no change in them. It happens when there are external tags on those AWS resources created by some external entities dynamically. 

Created external tags on some AWS resources like security groups for example. There was no change in cluster created/managed tags but during reconciliation, we are seeing that those tags are still being updated even if there is no change.

Tags should not be updated if there is no change.

The code where it is happening is here:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/tags/tags.go#L58-#L60

And here is the patch we are using to handle this:
https://github.com/aveshagarwal/cluster-api-provider-aws/commit/358028f9688a94939876bfe862cef4bc2e3fefc0

**Environment:**
AWS

latest

1.15, 1.16

aveshagarwal:

This PR fixes an issue where tags were being updated even there was no change in them. It happens when there are external tags on resources in addition to cluster managed tags.

Fixes #1499 

@ncdc @detiber @vincepri @randomvariable @rudoi

### DIFF
--- a/pkg/cloud/tags/tags.go
+++ b/pkg/cloud/tags/tags.go
@@ -54,9 +54,21 @@ func Apply(params *ApplyParams) error {
 
 // Ensure applies the tags if the current tags differ from the params.
 func Ensure(current infrav1.Tags, params *ApplyParams) error {
-	want := infrav1.Build(params.BuildParams)
-	if !current.Equals(want) {
+	diff := computeDiff(current, params.BuildParams)
+	if len(diff) > 0 {
 		return Apply(params)
 	}
 	return nil
+}
+
+func computeDiff(current infrav1.Tags, buildParams infrav1.BuildParams) infrav1.Tags {
+	want := infrav1.Build(buildParams)
+
+	// Some tags could be external set by some external entities
+	// and that means even if there is no change in cluster
+	// managed tags, tags would be updated as "current" and
+	// "want" would be different due to external tags.
+	// This fix makes sure that tags are updated only if
+	// there is a change in cluster managed tags.
+	return want.Difference(current)
 }
--- a/pkg/cloud/tags/tags_test.go
+++ b/pkg/cloud/tags/tags_test.go
@@ -1,0 +1,85 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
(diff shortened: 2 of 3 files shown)


detiber:
lgtm
/assign @rudoi 
/assign @ncdc

pkg/cloud/tags/tags_test.go
```diff
@@ -0,0 +1,85 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
ncdc:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

pkg/cloud/tags/tags.go
```diff
@@ -54,9 +54,21 @@ func Apply(params *ApplyParams) error {
...
+	want := infrav1.Build(buildParams)
+
+	// Some tags could be external set by some external entities
+	// and that means even if there is no change in cluster
+	// managed tags, tags would be updated as "current" and
+	// "want" would be different due to external tags.
+	// This fix makes sure that tags are updated only if
+	// there is a change in cluster managed tags.
+	return want.Difference(current)
+}
```
vincepri: Can we just move these 2 lines in the `Ensure` call instead?

ncdc: I'm guessing the reason he did it this way is to be able to unit test computeDiff. If you want to unit test the other func, you have to either mock out the ec2 CreateTags call, or we need to use something like a function pointer with test/default configurations to avoid ec2.

vincepri: Yeah makes sense, it's probably fine to keep it this way, although it's definitely harder to read/follow than before

pkg/cloud/tags/tags_test.go
aveshagarwal: done.

randomvariable:
lgtm for me

PR #1507, opened 2020-01-30
🏃 Update Machine FailureDomain field

chuckha:

Make the field name consistent with the json tag name and other documentation.

related to kubernetes-sigs/cluster-api#2185

### DIFF
--- a/api/v1alpha3/awsmachine_types.go
+++ b/api/v1alpha3/awsmachine_types.go
@@ -72,13 +72,13 @@ type AWSMachineSpec struct {
 
 	// FailureDomain is the failure domain unique identifier this Machine should be attached to, as defined in Cluster API.
 	// For this infrastructure provider, the ID is equivalent to an AWS Availability Zone.
-	// If multiple subnets are matched for the availability zone, the first one return is picked.
-	FailureDomain *string `json:"failureDomainID,omitempty"`
+	// If multiple subnets are matched for the availability zone, the first one returned is picked.
+	FailureDomain *string `json:"failureDomain,omitempty"`
 
 	// AvailabilityZone is references the AWS availability zone to use for this instance.
 	// If multiple subnets are matched for the availability zone, the first one return is picked.
 	//
-	// DEPRECATED: Switch to FailureDomainID.
+	// DEPRECATED: Switch to FailureDomain.
 	// +optional
 	AvailabilityZone *string `json:"availabilityZone,omitempty"`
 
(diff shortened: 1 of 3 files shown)


chuckha:
actually... i'm just realizing this is defined as `FailureDomain` on the Machine. I want this to be consistent with Machine so I'm going to go the other direction with this.

/hold

api/v1alpha3/awsmachine_types.go
```diff
@@ -73,12 +73,12 @@ type AWSMachineSpec struct {
...
 	// For this infrastructure provider, the ID is equivalent to an AWS Availability Zone.
 	// If multiple subnets are matched for the availability zone, the first one return is picked.
-	FailureDomain *string `json:"failureDomainID,omitempty"`
+	FailureDomain *string `json:"failureDomain,omitempty"`
 
 	// AvailabilityZone is references the AWS availability zone to use for this instance.
 	// If multiple subnets are matched for the availability zone, the first one return is picked.
 	//
-	// DEPRECATED: Switch to FailureDomainID.
+	// DEPRECATED: Switch to FailureDomain.
```
detiber: Not directly related to this PR, but since we are deprecating this field, should we be handling some type of automatic migration to the FailureDomain field as part of the conversion webhook? It looks like we are doing the reverse (populating the AvailabilityZone in v1alpha2 if FailureDomain is set in v1alpha3).

api/v1alpha3/awsmachine_types.go
```diff
@@ -73,12 +73,12 @@ type AWSMachineSpec struct {
 	// FailureDomain is the failure domain unique identifier this Machine should be attached to, as defined in Cluster API.
 	// For this infrastructure provider, the ID is equivalent to an AWS Availability Zone.
 	// If multiple subnets are matched for the availability zone, the first one return is picked.
```
vincepri:
```suggestion
	// If multiple subnets are matched for the availability zone, the first one returned is picked.
```

If you could fix this as well as part of this PR, thank you!

api/v1alpha3/awsmachine_types.go
vincepri: Yeah we can, we might want to remove the field altogether though, this is only marked as a deprecation and we have a fallback in place that looks at the `AvailabilityZone` field if FailureDomain is nil.

chuckha:
/hold cancel

As stated above, held because I wanted to swap directions the PR was going. That is done.

api/v1alpha3/awsmachine_types.go
detiber: I'm good with removing the field altogether as well :)

chuckha: I'll file an issue so we don't forget as this is out of scope for the original intent of this PR

PR #1509, opened 2020-01-31
:running: Remove deprecated AvailabilityZone field on machine spec

### LINKED ISSUE #1508: Remove deprecated AvailabilityZone field on machine spec
Please see this comment for context: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1507#discussion_r373213811

I believe it's safe to simply remove the `AvailabilityZone` field, but @vincepri or @detiber can correct me if I'm wrong. No deprecation policy needs to be maintained here as these are alpha types.

/milestone v0.5.0
/priority important-soon
/kind cleanup

I'm marking this important soon so we get rid of it before the next release so we don't carry a deprecated field along.

vincepri:

Fixes #1508


detiber:
/assign
:eyes:

.golangci.yml
```diff
@@ -28,4 +28,5 @@ run:
   deadline: 4m
   skip-files:
     - "zz_generated.*\\.go$"
+    - ".*conversion.*\\.go$"
```
detiber: I'd prefer if we didn't avoid all linting on these file, since it does contain non-generated code. I'm wondering if we can use an exclude string similar to line 26 of `.golangci.yml` to handle the common error(s) in a way that doesn't affect other files and/or stick with using the comment based linting and exclude the specific linter that is an issue (i.e `//nolint:prealloc`) like we do in other cases.

api/v1alpha2/awsmachine_conversion.go
```diff
@@ -23,48 +23,50 @@ import (
...
 // ConvertTo converts this AWSMachine to the Hub version (v1alpha3).
-func (src *AWSMachine) ConvertTo(dstRaw conversion.Hub) error { // nolint
+func (src *AWSMachine) ConvertTo(dstRaw conversion.Hub) error {
 	dst := dstRaw.(*infrav1alpha3.AWSMachine)
-	return Convert_v1alpha2_AWSMachine_To_v1alpha3_AWSMachine(src, dst, nil)
+	if err := Convert_v1alpha2_AWSMachine_To_v1alpha3_AWSMachine(src, dst, nil); err != nil {
+		return err
+	}
+
+	return nil
```
detiber: Any particular reason for this change? It is definitely a bit more explicit about the error handling, but it doesn't actually change any behavior and doesn't match with the other common usage in this file.

api/v1alpha2/awsmachine_conversion.go
```diff
@@ -23,48 +23,50 @@ import (
...
 		return err
 	}
 
 	// Manually convert FailureDomain to AvailabilityZone.
-	if in.FailureDomain != nil && out.AvailabilityZone == nil {
-		out.AvailabilityZone = in.FailureDomain
-	}
+	out.AvailabilityZone = in.FailureDomain
 
 	// Discards ImageLookupBaseOS
```
detiber: Tangential, but it looks like we'll need to add backup/restore functionality similar to what was done in CAPI for this field (not sure about others).

api/v1alpha2/awsmachine_conversion.go
vincepri: I was initially going to add the restore logic, then I took it away. I can revert it back if you prefer, this is just first step

api/v1alpha2/awsmachine_conversion.go
vincepri: Yeah see comment above

.golangci.yml
vincepri: Yeah I agree, we should probably use the other block. We're doing the same upstream in CAPI, and maybe when we find an actual solution we should suggest it to all providers as well.

Do you want me to revert this change for this PR?

detiber: Yeah, I think revert and converge on a better solution down the line is the best path forward.

api/v1alpha2/awsmachine_conversion.go
detiber: I saw another PR that was adding the restore logic, I'll create an issue to track handling this field.

api/v1alpha2/awsmachine_conversion.go
detiber: Let's revert to avoid conflicts with the #1498

api/v1alpha2/awsmachine_conversion.go
randomvariable: happy to follow up after both this and mine are merged.

detiber: Created #1510 to track

PR #1513, opened 2020-02-03
🏃 Automate creating a cluster with name having period symbol

sirao:
Added automation to test cluster creation where cluster name
contains symbol '.'

Executed tests locally and tests succeeded.


vincepri:
LGTM

/assign @detiber

ncdc:
I'm curious what is special about the `.` that we need to test it in a full e2e test?

sirao:
> I'm curious what is special about the `.` that we need to test it in a full e2e test?

Earlier, when special characters given in cluster name, LoadBalancer was not provisioned.

ncdc:
Thanks!

/approve

PR #1518, opened 2020-02-04
:running: Remove hardcoded 3h timeout from hack/heartbeat_account.py

detiber:

Removes the hardcoded 3h timeout from hack/heartbeat_account.py, which was causing issues with e2e and conformance tests that were running longer than 3h

### DIFF
--- a/hack/heartbeat_account.py
+++ b/hack/heartbeat_account.py
@@ -26,9 +26,7 @@ BOSKOS_RESOURCE_NAME=os.environ['BOSKOS_RESOURCE_NAME']
 USER = "cluster-api-provider-aws"
 
 if __name__ == "__main__":
-    count = 0
-    # keep sending heart beat for 3 hours
-    while count < 180:
+    while True:
         conn = httplib.HTTPConnection(BOSKOS_HOST)
         print "POST-ing heartbeat for resource %s to %s" % (BOSKOS_RESOURCE_NAME, BOSKOS_HOST)
         conn.request("POST", "/update?%s" % urllib.urlencode({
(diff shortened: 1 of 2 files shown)


hack/heartbeat_account.py
```diff
@@ -26,9 +26,7 @@
 USER = "cluster-api-provider-aws"
 
 if __name__ == "__main__":
-    count = 0
-    # keep sending heart beat for 3 hours
-    while count < 180:
+    while True:
         conn = httplib.HTTPConnection(BOSKOS_HOST)
         print "POST-ing heartbeat for resource %s to %s" % (BOSKOS_RESOURCE_NAME, BOSKOS_HOST)
         conn.request("POST", "/update?%s" % urllib.urlencode({
```
ncdc: There's another ref to count at the bottom

detiber: lol, yeah caught that as I was backporting it to the `release-0.4` branch.

scripts/ci-e2e.sh
```diff
@@ -41,6 +41,13 @@ source "${REPO_ROOT}/hack/ensure-kustomize.sh"
 ARTIFACTS="${ARTIFACTS:-${PWD}/_artifacts}"
 mkdir -p "$ARTIFACTS/logs/"
 
+# our exit handler (trap)
+cleanup() {
+  # stop boskos heartbeat
```
ncdc: Want to move anything else in here?

detiber: I thought about moving the account checkin code there, but the worst that happens is that we have to wait for the reaper to reclaim if it isn't called. I think killing the heartbeat is good enough for now, since it would prevent issues for someone manually running it where the background process may live on.

PR #1535, opened 2020-02-07
:bug: Support chunking AWS secrets

### LINKED ISSUE #1531: Validation exception because secret crosses 10k limit on creation of a machine
/kind bug
I am seeing validation exception because secret crosses 10k limit on creation of a machine.
"msg"="Reconciler error" "error"="ValidationException: 1 validation error detected: Value at 'secretBinary' failed to satisfy constraint: Member must have length less than or equal to 10240\n\tstatus code: 400

Used release 0.4 staging image to create a cluster
Probably we should add it into multiple secrets and bind them together.

**Environment:**

- Cluster-api-provider-aws version: 0.4 staging -> 0.4.9

randomvariable:
AWS has a 10kb limit on binary secret data (base64 encoded). This chunks
secrets to stay under the 10kb limit and stitches them back together in
the VM during cloud-init.

Additionally, forces e2e to exercise the "large userdata" path.

Fixes #1531


randomvariable:
/hold
Still cleaning up some bits.

api/v1alpha3/awsmachine_types.go
```diff
@@ -108,11 +108,15 @@ type CloudInit struct {
 	// the userdata from Secrets Manager and additionally delete the secret.
 	InsecureSkipSecretsManager bool `json:"insecureSkipSecretsManager,omitempty"`
 
-	// SecretARN is the Amazon Resource Name of the secret. This is stored
+	// SecretCount is the number of secrets used to form the complete secret
+	// +optional
+	SecretCount int32 `json:"secretCount,omitempty"`
```
randomvariable: @detiber Were you suggesting this should be a pointer?

randomvariable: original comment:
> I'm wondering if we need to add some type of validation here that the SecretCount != 0... Or do we think it would be valid to have a SecretCount of 0?

detiber: I don't think we can consider a SecretCount of 0 as valid, and treating it as a "default" or "unset" value should be fine :)

controllers/awsmachine_controller.go
```diff
@@ -411,23 +411,26 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 }
 
 func (r *AWSMachineReconciler) deleteEncryptedBootstrapDataSecret(machineScope *scope.MachineScope, secretSvc services.SecretsManagerInterface) error {
-	// do nothing if there isn't asecret
-	if machineScope.GetSecretARN() == "" {
+	// do nothing if there isn't a secret
+	if machineScope.GetSecretPrefix() == "" {
 		return nil
 	}
```
detiber: We probably should shortcut here if SecretCount == 0, and likely consider it an error condition.

pkg/cloud/scope/machine.go
```diff
@@ -186,22 +186,30 @@ func (m *MachineScope) UseSecretsManager() bool {
...
+// SetSecretPrefix sets the prefix for the secrets belonging
 // to the AWSMachine in AWS Secrets Manager
-func (m *MachineScope) SetSecretARN(value string) {
-	m.AWSMachine.Spec.CloudInit.SecretARN = value
+func (m *MachineScope) SetSecretPrefix(value string) {
+	m.AWSMachine.Spec.CloudInit.SecretPrefix = value
 }
 
-// DeleteSecretARN deletes the AMazon Resource Name for the secret belonging
+// DeleteSecretARNs deletes the prefix for the secret belonging
```
```suggestion
// DeleteSecretPrefix deletes the prefix for the secret belonging
```

pkg/cloud/scope/machine.go
```diff
@@ -186,22 +186,30 @@ func (m *MachineScope) UseSecretsManager() bool {
...
 // to the AWSMachine in AWS Secrets Manager
-func (m *MachineScope) DeleteSecretARN() {
-	m.AWSMachine.Spec.CloudInit.SecretARN = ""
+func (m *MachineScope) DeleteSecretPrefix() {
+	m.AWSMachine.Spec.CloudInit.SecretPrefix = ""
+}
+
+func (m *MachineScope) GetSecretCount() int32 {
```
detiber: godoc?

pkg/cloud/scope/machine.go
```diff
@@ -186,22 +186,30 @@ func (m *MachineScope) UseSecretsManager() bool {
...
+	return m.AWSMachine.Spec.CloudInit.SecretCount
+}
+
+func (m *MachineScope) SetSecretCount(i int32) {
```
detiber: godoc?

pkg/cloud/services/secretsmanager/secret.go
```diff
@@ -50,40 +58,57 @@ func (s *Service) Create(m *scope.MachineScope, data []byte) (string, error) {
...
+			return "", 0, err
+		}
 
-	if err != nil {
-		return "", err
+		if len(chunk) < maxSecretSizeBytes {
+			break
+		}
+
+		chunks++
```
detiber: Do we need to add some type of attempt to cleanup in the event that we get a failure part way through the creation of the secrets?

api/v1alpha3/awsmachine_types.go
randomvariable: Added validations as such.

pkg/cloud/services/secretsmanager/secret.go
randomvariable: Added some mop up in the controller part

pkg/cloud/services/secretsmanager/secret.go
```diff
@@ -50,40 +58,57 @@ func (s *Service) Create(m *scope.MachineScope, data []byte) (string, error) {
...
+	buf := bytes.NewBuffer(data)
+	prefix := s.secretNamePrefix(m)
 
-	resp, err := s.scope.SecretsManager.CreateSecret(&secretsmanager.CreateSecretInput{
-		Name:         aws.String(name),
-		SecretBinary: data,
-		Tags:         converters.MapToSecretsManagerTags(tags),
-	})
+	for {
+		name := fmt.Sprintf("%s-%d", prefix, chunks-1)
```
detiber: Would it be better to 0 index `chunks`, and do the increment after calling `buf.Next` if `len(chunk) > 0`? Could also return early in the case that `len(chunk) > 0`, since we wouldn't need to call `CreateSecret` in that case.

detiber: The logic here is a currently a bit difficult to follow with chunks starting at 1 and the break below if `len(chunk) < maxSecretSizeBytes`

detiber: I think the current logic would attempt to create a 0 byte secret in the case that the final chunk is exactly maxSecretSizeBytes

pkg/cloud/services/secretsmanager/secret.go
```diff
@@ -33,10 +36,15 @@ const (
...
+	// we set the max secret size to well below the 10240 byte limit, because this is limit after base64 encoding,
+	// but the aws sdk handles encoding for us, so we can't send a full 10240.
+	maxSecretSizeBytes = 7000
 )
 
-// Create stores a secret in AWS Secrets Manager for a given machine
-func (s *Service) Create(m *scope.MachineScope, data []byte) (string, error) {
+// Create stores data in AWS Secrets Manager for a given machine, chunking at 10kb per secret. The prefix of the secret
+// ARN and the number of chunks are returned.
+func (s *Service) Create(m *scope.MachineScope, data []byte) (string, int32, error) {
```
vincepri: This needs more extensive tests, especially around the expected chunks and so on.

pkg/cloud/services/secretsmanager/secret.go
ncdc: sgtm

vincepri:
Screen Shot 2020-02-07 at 3 00 23 PM

```
$ kubectl get awsmachine test1-controlplane-0 -o json | jq .spec
{
  "cloudInit": {
    "secretCount": 2,
    "secretPrefix": "aws.cluster.x-k8s.io/namespaces/default/clusters/test1/instances/test1-controlplane-0-hb98dkcfbm6vvvs82vncjfs4t27xz2xzcnbbz5clxzct6rn7dj4mzl"
  },
  "iamInstanceProfile": "control-plane.cluster-api-provider-aws.sigs.k8s.io",
  "instanceType": "t2.medium",
  "sshKeyName": "default"
}
```

Confirmed to be working

vincepri:
Same goes if there is no splitting, or rather it's only 1 secret

```
  "cloudInit": {
    "secretCount": 1,
    "secretPrefix": "aws.cluster.x-k8s.io/namespaces/default/clusters/test1/instances/test1-md-0-b66lq-rb5mlskfkgrwmt8v22m6v8jgnk8p25lf7ngb4zqpdtc8glm84lv8dm"
  },
```

vincepri:
Cluster is confirmed to be up and running:

```
NAME                                       STATUS   ROLES    AGE     VERSION
ip-10-0-0-106.us-west-2.compute.internal   Ready    master   2m37s   v1.16.1
ip-10-0-0-182.us-west-2.compute.internal   Ready    master   3m29s   v1.16.1
ip-10-0-0-219.us-west-2.compute.internal   Ready    <none>   2m30s   v1.16.1
ip-10-0-0-225.us-west-2.compute.internal   Ready    master   4m24s   v1.16.1
ip-10-0-0-86.us-west-2.compute.internal    Ready    <none>   2m40s   v1.16.1
```

Now proceeding to backport this change to `release-0.4`

pkg/cloud/services/secretsmanager/helpers.go
```diff
@@ -0,0 +1,28 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package secretsmanager
+
+func splitBytes(data []byte, maxSize int, iterFunc func([]byte)) {
+	var split []byte
+	for len(data) >= maxSize {
+		split, data = data[:maxSize], data[maxSize:]
```
detiber: Any specific reason to go this route vs using a buffer?

vincepri: No, but this is easier to read imo. We can change it up later if necessary, seems using a buffer is mostly a performance improvement?

detiber: Personally, I find the following a bit easier to read and potentially a bit more approachable for new contributors:

```go
func splitBytes(data []byte, maxSize int, iterFunc func([]byte)) {
	buff := NewBuffer(data)
	for {
		chunk := buff.Next(maxSize)
		if len(chunk) == 0 {
			return
		}
		iterFunc(chunk)
	}
}
```

It's clear that we are returning once we have processed all the data and there is less chance of confusion around if we are handling boundaries properly.

vincepri: Sure, that is better! I’ll change it up later

pkg/cloud/services/secretsmanager/secret.go
randomvariable: Resolved by @vincepri breaking out the chunking with its own tests.

pkg/cloud/services/secretsmanager/helpers.go
randomvariable: Swapped it in this PR.

.golangci.yml
```diff
@@ -1,31 +1,35 @@
...
+    - gocognit
+    - gomnd
+    - unused
+    - stylecheck
+    - gosimple
+    - staticcheck
+    - errcheck
+    - scopelint
+    - gocritic
+    - unparam
```
detiber: If we allow #1545 to merge first, we should remove `unparam` from this list.

detiber:
lgtm, but would like to give @vincepri and @ncdc another chance to take a look.

pkg/cloud/services/secretsmanager/secret_fetch_script.go
```diff
@@ -33,123 +33,163 @@ const secretFetchScript = `#!/bin/bash
...
+  check_aws_command "SecretsManager::DeleteSecret" "${delete_return}" "${out}"
+  if [ ${delete_return} -ne 0 ]; then
+    log::error_exit "Could not delete secret value" 2
+  fi
+}
+
+delete_secrets() {
+  for i in $(seq 0 ${FINAL_INDEX}); do
+    delete_secret_value "$i"
+    true
```
ncdc: ?

ncdc:
Just had that 1 question about `true` and waiting on the secret prefix name change as discussed on Slack. Everything else LGTM!

pkg/cloud/services/secretsmanager/secret_fetch_script.go
randomvariable: Leftover from prior testing, deleting.

pkg/cloud/services/secretsmanager/secret.go
```diff
@@ -50,55 +51,46 @@ func (s *Service) Create(m *scope.MachineScope, data []byte) (string, error) {
...
 	})
 
-	name := s.secretName(m)
-
-	resp, err := s.scope.SecretsManager.CreateSecret(&secretsmanager.CreateSecretInput{
-		Name:         aws.String(name),
-		SecretBinary: data,
-		Tags:         converters.MapToSecretsManagerTags(tags),
+	// Build the prefix. apirand uses 27 runes, 27^54 is closest to 2^256 for 256-bits of entropy.
+	prefix := parameterPrefix + "/" + apirand.String(54)
```
randomvariable: @ncdc changed secret prefix here.

detiber:
lol gosec:

```
level=info msg="[runner] linters took 16.362484296s with stages: goanalysis_metalinter: 16.339003999s"
pkg/cloud/services/secretsmanager/secret.go:35:2: G101: Potential hardcoded credentials (gosec)
	secretTypePrefix = "aws.cluster.x-k8s.io"
	^
level=info msg="File cache stats: 126 entries of total size 1.8MiB"
level=info msg="Memory: 252 samples, avg is 427.3MB, max is 946.1MB"
level=info msg="Execution took 26.301011463s"
make: *** [Makefile:136: lint] Error 1
```

pkg/cloud/services/secretsmanager/secret.go
```diff
@@ -17,31 +17,33 @@ limitations under the License.
 package secretsmanager
 
 import (
-	"strings"
+	"fmt"
+
+	"k8s.io/apimachinery/pkg/util/uuid"
```
detiber:
```suggestion
```

pkg/cloud/services/secretsmanager/secret.go
```diff
@@ -17,31 +17,33 @@ limitations under the License.
...
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/converters"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
 )
 
 const (
-	parameterPrefix  = "aws.cluster.x-k8s.io"
-	namespacesPrefix = "namespaces"
-	clustersPrefix   = "clusters"
-	instancesPrefix  = "instances"
+	secretTypePrefix = "aws.cluster.x-k8s.io" // nolint
```
detiber: Would it be better to rename this, or alternatively scope down the `// nolint` to `// nolint:gosec`?

randomvariable: Changed the var

pkg/cloud/services/secretsmanager/secret.go
```diff
@@ -50,55 +51,45 @@ func (s *Service) Create(m *scope.MachineScope, data []byte) (string, error) {
 		Additional:  additionalTags,
 	})
 
-	name := s.secretName(m)
-
-	resp, err := s.scope.SecretsManager.CreateSecret(&secretsmanager.CreateSecretInput{
-		Name:         aws.String(name),
-		SecretBinary: data,
-		Tags:         converters.MapToSecretsManagerTags(tags),
+	prefix := entryPrefix + "/" + string(uuid.NewUUID())
```
vincepri: I'd suggest to either use path.Join, or fmt.Sprintf

PR #1546, opened 2020-02-10
🐛fix-failing-test

fabriziopandini:
Fixes the problem introduced by https://github.com/kubernetes/kubernetes/pull/87585 (no more .deb generated on CI)

Fixes https://github.com/kubernetes/kubernetes/issues/87961

### DIFF
--- a/hack/ci/ci-bootstrap.yaml
+++ b/hack/ci/ci-bootstrap.yaml
@@ -22,7 +22,7 @@ files:
       fi
       ${GSUTIL} version
 
-      # This test installs debian packages that are a result of the CI and release builds.
+      # This test installs release packages or binaries that are a result of the CI and release builds.
       # It runs '... --version' commands to verify that the binaries are correctly installed
       # and finally uninstalls the packages.
       # For the release packages it tests all versions in the support skew.
@@ -36,7 +36,6 @@ files:
         mkdir -p $CI_DIR
 
         declare -a PACKAGES_TO_TEST=("kubectl" "kubelet" "kubeadm")
-        PACKAGE_EXT="deb"
         declare -a CONTAINERS_TO_TEST=("kube-apiserver" "kube-controller-manager" "kube-proxy" "kube-scheduler")
         CONTAINER_EXT="tar"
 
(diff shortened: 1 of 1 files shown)


hack/ci/ci-bootstrap.yaml
```diff
@@ -65,10 +64,59 @@ files:
...
+EOF
+
+          mkdir -p /etc/sysctl.d/ 
+          cat <<EOF >> /etc/sysctl.d/50-kubeadm.conf 
+# The file is provided as part of the kubeadm package	
+net.ipv4.ip_forward = 1
+EOF
+
+          systemctl enable kubelet
+          systemctl start kubelet
```
detiber: The systemd unit files, modules, sysctl, and service configuration should already be configured in the base image that is used.

fabriziopandini: Just to double-check, does the base image come with an existing kubeadm package already installed? then, is it right to assume the goal of this change is only to override the binaries?

detiber: Correct, the default image id that is used is pulled from the published CAPA images: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/release-0.4/hack/ci/e2e-conformance.sh#L220-L230, so only overriding the binaries *should* be needed.

fabriziopandini: fixed. then it is simpler than expected 😄

fabriziopandini:
@detiber comment addressed, PTAL again

detiber:
This looks good to me, it doesn't look like we have a pre-submit job that is wired up to use the ci artifacts, so let's get this merged and see how it does :)

/lgtm
/approve

hack/ci/ci-bootstrap.yaml
```diff
@@ -65,10 +64,13 @@ files:
...
-            echo "* downloading package: $CI_URL/$CI_PACKAGE.$PACKAGE_EXT"
-            ${GSUTIL} cp "$CI_URL/$CI_PACKAGE.$PACKAGE_EXT" "$CI_DIR/$CI_PACKAGE.$PACKAGE_EXT"
-            ${SUDO} dpkg -i "$CI_DIR/$CI_PACKAGE.$PACKAGE_EXT" || echo "* ignoring expected 'dpkg -i' result"
+            echo "* downloading binary: $CI_URL/$CI_PACKAGE"
+            ${GSUTIL} cp "$CI_URL/$CI_PACKAGE" "$CI_DIR/$CI_PACKAGE"
+            chmod +x "$CI_DIR/$CI_PACKAGE"
+            mv "$CI_DIR/$CI_PACKAGE" "/usr/bin/$CI_PACKAGE"
           done
+
+          systemctl restart kubelet
```
neolit123: as mentioned here:
https://github.com/kubernetes-sigs/cluster-api-provider-gcp/pull/284/files#r377658995

this might not be complete fix, in case the base OS is missing some deps.

PR #1550, opened 2020-02-11
✨Add support to specify static credentials in a Secret

vincepri:

/hold

### DIFF
--- a/api/v1alpha2/awscluster_conversion.go
+++ b/api/v1alpha2/awscluster_conversion.go
@@ -47,6 +47,7 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 		return err
 	}
 
+	dst.Spec.CredentialsSecretName = restored.Spec.CredentialsSecretName
 	dst.Spec.ImageLookupOrg = restored.Spec.ImageLookupOrg
 	dst.Spec.ImageLookupBaseOS = restored.Spec.ImageLookupBaseOS
 	if restored.Spec.ControlPlaneLoadBalancer != nil {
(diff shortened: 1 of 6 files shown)


detiber:
@vincepri Is the goal for this some type of multi-tenancy, or just alternative credentials to be used?

Let me know when this is ready for review.

vincepri:
It's an alternative to the webhook refactoring I posted upstream, it's mostly to show that there is a viable alternative to the multi-tenancy approach we have taken today with namespaces

detiber:
> It's an alternative to the webhook refactoring I posted upstream, it's mostly to show that there is a viable alternative to the multi-tenancy approach we have taken today with namespaces

There is a potential problem here in that we would need some mechanism in place to prevent users from one tenant from using the credentials owned by another tenant.

vincepri:
@detiber The namespace separation still applies, given that we're not using an ObjectReference, the secret must live in the Cluster's own namespace

randomvariable:
Don't think this is quite right.

For multi-tenant AWS accounts, role assumption would be the primary practice, and this method can't support that (lack of rotation and support of session tokens).

We should start with the user stories and AWS best practice and work from there.

vincepri:
@randomvariable As I mentioned this isn't supposed to go in, but rather show that there is an easy path forward. What you mentioned about role assumption or token is also a problem with today's file credentials.

randomvariable:
> What you mentioned about role assumption or token is also a problem with today's file credentials.

Role assumption can be set today using https://docs.aws.amazon.com/cli/latest/topic/config-vars.html, and is in use.

In addition, we have the default credential provider chain best described in the Java docs at https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/credentials.html

as has been discussed previously in #201

vincepri:
Nice, forgot we had those. 

Did you want to open an issue outlining the use cases and possible solutions or want me to?

randomvariable:
> Did you want to open an issue outlining the use cases and possible solutions or want me to?
Yup, I can do that.

randomvariable:
Created https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1552

vincepri:
Closing now that we decided to split webhooks out

/close

PR #1558, opened 2020-02-13
:book: userdata-privacy: Additional troubleshooting info

randomvariable:
Adding some more troubleshooting information based on conversations with @sethp-nr


chuckha:
/lgtm

👍 

also I am surprised you are not in the approvers list? that has to be an oversight

randomvariable:
> also I am surprised you are not in the approvers list? that has to be an oversight

I went emeritus, just haven't bothered to revoke.

PR #1559, opened 2020-02-13
🏃 Enabled scopelint and fixed reported errors with suggestions

### LINKED ISSUE #1542: cleanup: enable `scopelint` linter
After #1535, we updated the golangci-lint to the latest version and matched the upstream Cluster API format for the linter, it seems there were a bunch of linters missing.

- Head over to [.golanci-lint.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/.golangci.yml) in the root of this repository.
- Remove the linter from the `linters.disable` list.
- Run `make lint-full`.
- Fix the reported errors with the suggestions.
- Submit a PR with the changes.

Feel free to reach out at any time if you need guidance or help.

/kind cleanup
/help
/good-first-issue
/milestone Next
/priority backlog

teaguecole:
This enables scopelint.  We need it so we can follow best practices with go and write cleaner code.

Fixes #1542

I am not sure which of the following labels to add. Other and Minor seemed like the most reasonable for this.  I am also open to enabling more linters and cleaning up code, but I also understand the value of having good-first-issues to bring in new contributors.  I kind of enjoy doing this as it makes me better at writing Go but I don't want to hog all the good-first-issues.

### DIFF
--- a/pkg/cloud/services/ec2/eips.go
+++ b/pkg/cloud/services/ec2/eips.go
@@ -118,8 +118,8 @@ func (s *Service) releaseAddresses() error {
 		return errors.Wrapf(err, "failed to describe elastic IPs %q", err)
 	}
 
-	for _, ip := range out.Addresses {
-		ip := ip
+	for i := range out.Addresses {
+		ip := out.Addresses[i]
 		if ip.AssociationId != nil {
 			_, err := s.scope.EC2.DisassociateAddress(&ec2.DisassociateAddressInput{
 				AssociationId: ip.AssociationId,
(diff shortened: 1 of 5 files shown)


pkg/cloud/services/ec2/eips.go
```diff
@@ -119,6 +119,7 @@ func (s *Service) releaseAddresses() error {
 	}
 
 	for _, ip := range out.Addresses {
```
detiber: nit: Instead of shadowing the variable here, what about:
```go
for i := range out.Addresses {
	ip := out.Addresses[i]
}
```

teaguecole: Yeah, I agree. Will make that change.

detiber: Thanks, do you mind updating the other instances as well?

teaguecole: All done. I think I got everything.

PR #1560, opened 2020-02-14
Add instructions for existing AWS infrastructure

### LINKED ISSUE #1511: Add docs for consuming existing AWS infrastructure
The current quickstart / docs assume CAPA will be creating all infrastructure (VPC etc).

We should add a section describing the requirements (resources / tags / etc...) for consuming existing infrastructure. 

cc/ @scottslowe

/kind feature

scottslowe:
This PR adds instructions for re-using existing AWS infrastructure. This includes information on configuring existing AWS resources as well as instructions for properly configuring CAPI manifests.

This PR adds a document explaining how users can use or consume existing AWS infrastructure with Cluster API, and updates the `docs/README.md` document with a link to the new document.

Fixes #1511

### DIFF
--- a/docs/existing-aws-infrastructure.md
+++ b/docs/existing-aws-infrastructure.md
@@ -27,22 +27,25 @@ In order for Cluster API to properly recognize and consume existing AWS resource
     sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>
     sigs.k8s.io/cluster-api-provider-aws/role
 
-The values of these tags are immaterial. For the first tag, Cluster API itself uses values of "owned" to indicate a resource that was created by Cluster API and should be destroyed by Cluster API. In the case of consuming existing infrastructure, any value other than "owned" is acceptable. Using the value "unmanaged" is not uncommon.
-
-For the second tag, Cluster API uses the value "common"; you can use that value in this use case as well. **[NEED TO VERIFY]**
-
-Additionally, for proper operation of the AWS cloud provider, your resources should also have the `kubernetes.io/cluster/<cluster-name>` tag present as well.
-
-The tags should be present on the following AWS resources:
+These two tags should be present on the following AWS resources:
 
 * VPC
-* All subnets that Cluster API
+* All subnets that Cluster API will utilize (private and public)
 * Route tables
 * Internet gateway
 * NAT gateways
 * Elastic IPs assigned to NAT gateways
 
+For the `sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>` tag, a value of "owned" is used by Cluster API itself to indicate resources are owned by a specific cluster, and that the lifecycle of the resource is tied to the lifecycle of the cluster. Cluster API recognizes a value of "shared" to indicate a resource may be shared between multiple clusters, and the lifecycle of said resource is not tied to any particular cluster. In this situation---where AWS infrastructure is _not_ being managed by Cluster API---the value should be something **other** than "owned". Using "shared" is acceptable.
 
+For the `sigs.k8s.io/cluster-api-provider-aws/role` tag, Cluster API uses a variety of values to indicate resources are associated with a particular role. Here are the tag values that should be used:
+
+* VPC, route tables, Internet gateway, and NAT gateways should use the value "common".
+* Public subnets should use the value "public".
+* Private subnets should use the value "private".
+* Elastic IPs (for use by the NAT gateways) should use the value "apiserver".
+
+Additionally, for proper operation of the AWS cloud provider, all subnets where Kubernetes nodes reside should also have the `kubernetes.io/cluster/<cluster-name>` tag present.
 
 ## Configuring the AWSCluster Specification
 
@@ -66,8 +69,6 @@ When you use `kubectl apply` to apply the Cluster and AWSCluster specifications 
 
 ## Configuring the AWSMachine Specification (Optional)
 
-**[NEED TO DOUBLE-CHECK THIS WHOLE SECTION]**
-
 To distribute EC2 instances across different subnets or different AZs, you can add information to the AWSMachine specification. This is optional; without it, Cluster API will deploy to the first private subnet it discovers.
 
 To specify that an EC2 instance should be placed in a specific subnet, add this to the AWSMachine specification:
@@ -89,5 +90,5 @@ These two settings are mutually exclusive. You may use only one or the other, no
 
 ## Caveats/Notes
 
-* If both public and private subnets are available in an AZ, CAPI will choose the private subnet in the AZ over the public subnet.
+* When both public and private subnets are available in an AZ, CAPI will choose the private subnet in the AZ over the public subnet for placing EC2 instances.
 * If you configure CAPI to use existing infrastructure as outlined above, CAPI will _not_ create an SSH bastion host. Combined with the previous bullet, this means you must make sure you have established some form of connectivity to the instances that CAPI will create.
--- a/docs/existing-aws-infrastructure.md
+++ b/docs/existing-aws-infrastructure.md
@@ -7,45 +7,27 @@ Normally, Cluster API will create infrastructure on AWS when standing up a new w
 In order to have Cluster API consume existing AWS infrastructure, you will need to have already created the following resources:
 
 * A VPC
-* One or more private subnets (subnets that do not automatically assign a public IP address to EC2 instances)
+* One or more private subnets (subnets that do not have a route to an Internet gateway)
 * A public subnet in the same Availability Zone (AZ) for each private subnet (this is required for NAT gateways to function properly)
 * A NAT gateway for each private subnet, along with associated Elastic IP addresses
 * An Internet gateway for all public subnets
 * Route table associations that provide connectivity to the Internet through a NAT gateway (for private subnets) or the Internet gateway (for public subnets)
 
-You will need the following information, which can be obtained either via the AWS Management Console or using the AWS CLI:
+Note that a public subnet (and associated Internet gateway) are required for the control plane of the workload cluster.
 
-1. The ID of the VPC
-2. The IDs of the subnets in the VPC that you want Cluster API to use
+You will need the ID of the VPC that Cluster API should use. This information is available via the AWS Management Console or the AWS CLI. It is not necessary to have a list of the subnet IDs; Cluster API will discover the subnets if the VPC ID is provided.
 
 Note that there is no need to create an Elastic Load Balancer (ELB), security groups, or EC2 instances; Cluster API will take care of these items.
 
 ## Tagging AWS Resources
 
-In order for Cluster API to properly recognize and consume existing AWS resources, these resources must be tagged with a specific set of tags. The list below provides the tags that must be present; in this list, `<cluster-name>` refers to the name of cluster as specified in the `metadata.name` field of the Cluster object (or manifest). Here are the tags that are required:
+Cluster API itself does tag AWS resources it creates. The `sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>` (where `<cluster-name>` matches the `metadata.name` field of the Cluster object) tag, with a value of `owned`, tells Cluster API that it has ownership of the resource and is able to modify the resource and manage the lifecycle of the resource.
 
-    sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>
-    sigs.k8s.io/cluster-api-provider-aws/role
+When consuming existing AWS infrastructure, Cluster API does not require any tags to be present. The absence of the tags on an AWS resource indicates to Cluster API that it should not modify the resource or attempt to manage the lifecycle of the resource.
 
-These two tags should be present on the following AWS resources:
+However, the AWS cloud provider _does_ require certain tags in order to function properly. Specifically, all subnets where Kubernetes nodes reside should have the `kubernetes.io/cluster/<cluster-name>` tag present. Private subnets should also have the `kubernetes.io/role/internal-elb` tag with a value of 1, and public subnets should have the `kubernetes.io/role/elb` tag with a value of 1. These latter two tags help the cloud provider understand which subnets to use when creating load balancers.
 
-* VPC
-* All subnets that Cluster API will utilize (private and public)
-* Route tables
-* Internet gateway
-* NAT gateways
-* Elastic IPs assigned to NAT gateways
-
-For the `sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>` tag, a value of "owned" is used by Cluster API itself to indicate resources are owned by a specific cluster, and that the lifecycle of the resource is tied to the lifecycle of the cluster. Cluster API recognizes a value of "shared" to indicate a resource may be shared between multiple clusters, and the lifecycle of said resource is not tied to any particular cluster. In this situation---where AWS infrastructure is _not_ being managed by Cluster API---the value should be something **other** than "owned". Using "shared" is acceptable.
-
-For the `sigs.k8s.io/cluster-api-provider-aws/role` tag, Cluster API uses a variety of values to indicate resources are associated with a particular role. Here are the tag values that should be used:
-
-* VPC, route tables, Internet gateway, and NAT gateways should use the value "common".
-* Public subnets should use the value "public".
-* Private subnets should use the value "private".
-* Elastic IPs (for use by the NAT gateways) should use the value "apiserver".
-
-Additionally, for proper operation of the AWS cloud provider, all subnets where Kubernetes nodes reside should also have the `kubernetes.io/cluster/<cluster-name>` tag present.
+Finally, if the controller manager isn't started with the `--configure-cloud-routes: "false"` parameter, the route table(s) will also need the `kubernetes.io/cluster/<cluster-name>` tag. (This parameter can be added by customizing the `KubeadmConfigSpec` object of the `KubeadmControlPlane` object.)
 
 ## Configuring the AWSCluster Specification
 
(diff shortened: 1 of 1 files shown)


scottslowe:
@detiber I think this is ready for review now. Mind taking a look?

hasheddan:
@scottslowe Hi! I have been running through a few tests with using existing subnets and VPC and it seems relevant to your documentation here. I was wondering if there are a certain number of subnets required? I provided a VPC and 3 subnets and see `no subnets available` in the pod logs when creating an AWS control plane machine.

hasheddan:
I believe this is related to missing tags on the infrastructure, apologies for cluttering this PR!

scottslowe:
@hasheddan Mind moving this to the #cluster-api-aws channel on the K8s Slack? That may be easier, and I can update the discussion thread here with appropriate findings/information for the benefit of future web searchers.

docs/existing-aws-infrastructure.md
```diff
@@ -0,0 +1,94 @@
...
+spec:
+  subnet:
+    id: subnet-0a3507a5ad2c5c8c3
+```
+
+To tell Cluster API that an EC2 instance should be placed in a particular AZ but allow Cluster API to select which subnet in that AZ can be used, add this to the AWSMachine specification:
+
+```yaml
+spec:
+  availabilityZone: "us-west-2a"
```
detiber: This should be failureDomain for v1alpha3
```suggestion
  failureDomain: "us-west-2a"
```

docs/existing-aws-infrastructure.md
```diff
@@ -0,0 +1,94 @@
...
+* Route table associations that provide connectivity to the Internet through a NAT gateway (for private subnets) or the Internet gateway (for public subnets)
+
+You will need the following information, which can be obtained either via the AWS Management Console or using the AWS CLI:
+
+1. The ID of the VPC
+2. The IDs of the subnets in the VPC that you want Cluster API to use
+
+Note that there is no need to create an Elastic Load Balancer (ELB), security groups, or EC2 instances; Cluster API will take care of these items.
+
+## Tagging AWS Resources
```
detiber: I wouldn't expect the tags to be required if specifying vpc ids and subnet ids, but I'll try to verify.

docs/existing-aws-infrastructure.md
```diff
@@ -0,0 +1,94 @@
+# Consuming Existing AWS Infrastructure
+
+Normally, Cluster API will create infrastructure on AWS when standing up a new workload cluster. However, it is possible to have Cluster API re-use existing AWS infrastructure instead of creating its own infrastructure. Follow the instructions below to configure Cluster API to consume existing AWS infrastructure.
+
+## Prerequisites
+
+In order to have Cluster API consume existing AWS infrastructure, you will need to have already created the following resources:
+
+* A VPC
+* One or more private subnets (subnets that do not automatically assign a public IP address to EC2 instances)
```
sethp-nr: What CAPA considers public is mainly whether the subnet in question has a route to an internet gateway – if yes, it's public, otherwise it's private.

docs/existing-aws-infrastructure.md
```diff
@@ -0,0 +1,94 @@
...
+* A public subnet in the same Availability Zone (AZ) for each private subnet (this is required for NAT gateways to function properly)
+* A NAT gateway for each private subnet, along with associated Elastic IP addresses
+* An Internet gateway for all public subnets
+* Route table associations that provide connectivity to the Internet through a NAT gateway (for private subnets) or the Internet gateway (for public subnets)
```
sethp-nr: It might be worth mentioning that that the public subnet / IGW bits are still necessary, even if you also specify an `internal` scheme for your [apiserver's load balancer](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/api/v1alpha3/types.go#L121). It looks like we're [unwilling to proceed without one](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0b83f06314410a9a3d628f006a647086f51736a8/pkg/cloud/services/ec2/subnets.go#L67).

docs/existing-aws-infrastructure.md
```diff
@@ -0,0 +1,94 @@
...
+
+You will need the following information, which can be obtained either via the AWS Management Console or using the AWS CLI:
+
+1. The ID of the VPC
+2. The IDs of the subnets in the VPC that you want Cluster API to use
```
sethp-nr: The subnet IDs are actually ignored; given a VPC we'll always look up all the subnets and populate them on the cluster object via [describe](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0b83f06314410a9a3d628f006a647086f51736a8/pkg/cloud/services/ec2/subnets.go#L170).

Providing the subnet IDs wouldn't bother CAPA, but it might lead the humans to believe they can ask CAPA to use a subset of IDs by providing a reduced list. At least at present, CAPA would ignore that intent.

docs/existing-aws-infrastructure.md
sethp-nr: Actually, we do [defer to the public role tag](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0b83f06314410a9a3d628f006a647086f51736a8/pkg/cloud/services/ec2/subnets.go#L210-L212), so if someone _really_ wants to they could tag an actually-private subnet as role "public" to get past that linked error.

Probably best to leave that out of the documentation, though, as I've no idea what might fail down the line if someone did such a thing.

docs/existing-aws-infrastructure.md
sethp-nr: Agreed – they're not required, and in fact not a good idea. The way CAPA knows whether it can modify the VPC/subnets/route tables/nat gateways/IGWs is by whether the VPC has the "owned" tag. If we mention the tags at all in this page, I would suggest saying something to the effect of "setting this tag will tell CAPA to take ownership over and modify the existing resources" and leaving it at that.

docs/existing-aws-infrastructure.md
```diff
@@ -0,0 +1,94 @@
...
+
+For the `sigs.k8s.io/cluster-api-provider-aws/role` tag, Cluster API uses a variety of values to indicate resources are associated with a particular role. Here are the tag values that should be used:
+
+* VPC, route tables, Internet gateway, and NAT gateways should use the value "common".
+* Public subnets should use the value "public".
+* Private subnets should use the value "private".
+* Elastic IPs (for use by the NAT gateways) should use the value "apiserver".
+
+Additionally, for proper operation of the AWS cloud provider, all subnets where Kubernetes nodes reside should also have the `kubernetes.io/cluster/<cluster-name>` tag present.
```
sethp-nr: Yes, and additionally private subnets should be tagged with `kubernetes.io/role/internal-elb=1` and public with `kubernetes.io/role/elb=1`, else the cloud provider can't find any subnets to create load balancers in.

docs/existing-aws-infrastructure.md
```diff
@@ -0,0 +1,94 @@
...
+  networkSpec:
+    vpc:
+      id: vpc-0425c335226437144
+    subnets:
+      - id: subnet-07758a9bc904d06af
+      - id: subnet-0a3507a5ad2c5c8c3
+      - id: subnet-02ad6429dd0532452
+      - id: subnet-02b300779e9d895cf
+      - id: subnet-03d8f353b289b025f
+      - id: subnet-0a2fe03b0d88fa078
```
sethp-nr: As mentioned above, it's not necessary to include the subnets here. However, we do find it useful to include a default `sshKeyName` and `region` at this point, like so:

```
spec:
  networkSpec:
    vpc:
      id: vpc-0425c335226437144
  region: us-west-2
  sshKeyName: some-ssh-key
```

Though neither are required.

detiber:
Bumping from the v0.5.0 milestone and moving to v0.5.x
/milestone v0.5.x

scottslowe:
@sethp-nr @detiber Thank you for the feedback! I will work on incorporating your feedback and update the PR within the next couple of days.

docs/existing-aws-infrastructure.md
scottslowe: Updated via 38ce860. I did not add information specific to the requirement of a public subnet and IGW with regard to the API server load balancer, as those items are already needed anyway. If desired, I can add those details.

docs/existing-aws-infrastructure.md
scottslowe: References to CAPI-specific tags removed via 38ce860.

docs/existing-aws-infrastructure.md
scottslowe: Fixed via 38ce860.

docs/existing-aws-infrastructure.md
scottslowe: Resolved via 38ce860, where I removed references to CAPI-specific tags being required.

docs/existing-aws-infrastructure.md
scottslowe: Added some text to that effect via 38ce860

docs/existing-aws-infrastructure.md
scottslowe: I didn't specifically address this comment (the text already noted that the public subnet and IGW are required), but I can add some additional text if desired.

docs/existing-aws-infrastructure.md
scottslowe: Fixed via 38ce860.

docs/existing-aws-infrastructure.md
scottslowe: Added information about these tags via 38ce860.

docs/existing-aws-infrastructure.md
scottslowe: Removed references to the list of subnets via 38ce860.

scottslowe:
@detiber @sethp-nr I think I've addressed all your feedback. PTAL, thanks!

docs/existing-aws-infrastructure.md
```diff
@@ -7,45 +7,27 @@ Normally, Cluster API will create infrastructure on AWS when standing up a new w
...
 
-    sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>
-    sigs.k8s.io/cluster-api-provider-aws/role
+When consuming existing AWS infrastructure, Cluster API does not require any tags to be present. The absence of the tags on an AWS resource indicates to Cluster API that it should not modify the resource or attempt to manage the lifecycle of the resource.
 
-These two tags should be present on the following AWS resources:
+However, the AWS cloud provider _does_ require certain tags in order to function properly. Specifically, all subnets where Kubernetes nodes reside should have the `kubernetes.io/cluster/<cluster-name>` tag present. Private subnets should also have the `kubernetes.io/role/internal-elb` tag with a value of 1, and public subnets should have the `kubernetes.io/role/elb` tag with a value of 1. These latter two tags help the cloud provider understand which subnets to use when creating load balancers.
```
sethp-nr:
```suggestion
However, the built-in Kubernetes AWS cloud provider _does_ require certain tags in order to function properly. Specifically, all subnets where Kubernetes nodes reside should have the `kubernetes.io/cluster/<cluster-name>` tag present. Private subnets should also have the `kubernetes.io/role/internal-elb` tag with a value of 1, and public subnets should have the `kubernetes.io/role/elb` tag with a value of 1. These latter two tags help the cloud provider understand which subnets to use when creating load balancers.
```

docs/existing-aws-infrastructure.md
```diff
@@ -7,45 +7,27 @@ Normally, Cluster API will create infrastructure on AWS when standing up a new w
...
 
 ## Tagging AWS Resources
 
-In order for Cluster API to properly recognize and consume existing AWS resources, these resources must be tagged with a specific set of tags. The list below provides the tags that must be present; in this list, `<cluster-name>` refers to the name of cluster as specified in the `metadata.name` field of the Cluster object (or manifest). Here are the tags that are required:
+Cluster API itself does tag AWS resources it creates. The `sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>` (where `<cluster-name>` matches the `metadata.name` field of the Cluster object) tag, with a value of `owned`, tells Cluster API that it has ownership of the resource and is able to modify the resource and manage the lifecycle of the resource.
 
-    sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>
-    sigs.k8s.io/cluster-api-provider-aws/role
+When consuming existing AWS infrastructure, Cluster API does not require any tags to be present. The absence of the tags on an AWS resource indicates to Cluster API that it should not modify the resource or attempt to manage the lifecycle of the resource.
```
sethp-nr:
```suggestion
When consuming existing AWS infrastructure, the Cluster API AWS provider does not require any tags to be present. The absence of the tags on an AWS resource indicates to Cluster API that it should not modify the resource or attempt to manage the lifecycle of the resource.
```

docs/existing-aws-infrastructure.md
```diff
@@ -79,14 +76,17 @@ spec:
...
 spec:
-  availabilityZone: "us-west-2a"
+  template:
+    spec:
+      subnet:
+        id: subnet-0a3507a5ad2c5c8c3
 ```
 
-These two settings are mutually exclusive. You may use only one or the other, not both.
+Users may either specify `failureDomain` on the Machine or MachineDeployment objects, _or_ users may explicitly specify subnet IDs on the AWSMachine or AWSMachineTemplate objects. The two settings are mutually exclusive.
```
sethp-nr:
```suggestion
Users may either specify `failureDomain` on the Machine or MachineDeployment objects, _or_ users may explicitly specify subnet IDs on the AWSMachine or AWSMachineTemplate objects. If both are specified, the subnet ID is used and the `failureDomain` is ignored.
```

We aren't able to validate cross-object, so we can't prevent a determined user from setting up the scenario where both are specified. Better to be explicit about what happens in that case IMO.

docs/existing-aws-infrastructure.md
```diff
@@ -7,45 +7,27 @@ Normally, Cluster API will create infrastructure on AWS when standing up a new w
...
+You will need the ID of the VPC that Cluster API should use. This information is available via the AWS Management Console or the AWS CLI. It is not necessary to have a list of the subnet IDs; Cluster API will discover the subnets if the VPC ID is provided.
 
 Note that there is no need to create an Elastic Load Balancer (ELB), security groups, or EC2 instances; Cluster API will take care of these items.
 
 ## Tagging AWS Resources
 
-In order for Cluster API to properly recognize and consume existing AWS resources, these resources must be tagged with a specific set of tags. The list below provides the tags that must be present; in this list, `<cluster-name>` refers to the name of cluster as specified in the `metadata.name` field of the Cluster object (or manifest). Here are the tags that are required:
+Cluster API itself does tag AWS resources it creates. The `sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>` (where `<cluster-name>` matches the `metadata.name` field of the Cluster object) tag, with a value of `owned`, tells Cluster API that it has ownership of the resource and is able to modify the resource and manage the lifecycle of the resource.
```
sethp-nr:
```suggestion
Cluster API itself does tag AWS resources it creates. The tag `sigs.k8s.io/cluster-api-provider-aws/cluster/<cluster-name>` (where `<cluster-name>` matches the `metadata.name` field of the Cluster object), with a value of `owned`, tells Cluster API that it has ownership of the resource. In this case, Cluster API will modify and manage the lifecycle of the resource.
```

docs/existing-aws-infrastructure.md
```diff
@@ -7,45 +7,27 @@ Normally, Cluster API will create infrastructure on AWS when standing up a new w
...
 * A VPC
-* One or more private subnets (subnets that do not automatically assign a public IP address to EC2 instances)
+* One or more private subnets (subnets that do not have a route to an Internet gateway)
 * A public subnet in the same Availability Zone (AZ) for each private subnet (this is required for NAT gateways to function properly)
 * A NAT gateway for each private subnet, along with associated Elastic IP addresses
 * An Internet gateway for all public subnets
 * Route table associations that provide connectivity to the Internet through a NAT gateway (for private subnets) or the Internet gateway (for public subnets)
 
-You will need the following information, which can be obtained either via the AWS Management Console or using the AWS CLI:
+Note that a public subnet (and associated Internet gateway) are required for the control plane of the workload cluster.
```
sethp-nr: I think I've mislead you – we insist that there _be_ public subnets, but it looks like we [know not to use them for the control plane load balancer ](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/9913ac8dbb96ed9bcffa6eba75b886cd4ef8c20b/pkg/cloud/services/elb/loadbalancer.go#L288) if the scheme is set to `internal`. Maybe:

```suggestion
Note that a public subnet (and associated Internet gateway) are required even if the control plane of the workload cluster is set to use an `internal` load balancer.
```

?

docs/existing-aws-infrastructure.md
scottslowe: I've adopted your change (included in 1667e3ab).

docs/existing-aws-infrastructure.md
scottslowe: I've adopted your change (included in 1667e3ab).

docs/existing-aws-infrastructure.md
scottslowe: This suggested change is included in 1667e3a.

docs/existing-aws-infrastructure.md
scottslowe: 1667e3a includes this suggested wording change.

docs/existing-aws-infrastructure.md
scottslowe: I've adopted your change (included in 1667e3ab). Is it worthwhile including information on the mechanism by which an internal load balancer is configured, or is that information best suited to be elsewhere?

scottslowe:
Hey @sethp-nr, thanks for the awesome feedback. I've adopted your suggested wording changes. Let me know if there are any other changes you'd like to see. Thank you!

scottslowe:
Ping @detiber @sethp-nr for any additional feedback on this PR, thanks!

docs/existing-aws-infrastructure.md
johnharris85: TIL, and this seems...not ideal?

sethp-nr:
Sorry @scottslowe, I've been taking some time off since the end of week before last. I'll be sure to review this today!

docs/existing-aws-infrastructure.md
scottslowe: I’ve tested it, and the subnet IDs are not required. I believe that text has been removed in the most recent commit against this PR.

sethp-nr:
Thanks so much for your help, Scott!

/lgtm

detiber:
Since this has been verified by multiple people already, I think it's good to go.
/approve

PR #1574, opened 2020-02-17
:running: Do not set xtrace in hack/ensure-kind.sh

detiber:

Do not set xtrace in hack/ensure-kind.sh


kris-nova:
/lgtm

Thanks @detiber

PR #1577, opened 2020-02-18
Update and require CRDv1

h0tbird:
**What this PR does / why we need it**:

- Update all dependencies to v0.17.2, which is Kubernetes v1.17.2
- Update controller-runtime to v0.5.0
- Updates the CRDs to v1.
- Updates kubebuilder-tools to 1.16.4

PR #1575, opened 2020-02-17
Refactor kustomize config/ folder

h0tbird:
**What this PR does / why we need it**:
Refactors kustomize `config/` to support multi tenancy as described in https://master.cluster-api.sigs.k8s.io/developer/providers/v1alpha2-to-v1alpha3.html#refactor-kustomize-config-folder-to-support-multi-tenancy-when-using-webhooks

**Which issue(s) this PR fixes**:
Fixes #1572

I am following the documentation and I am also using CAPI as a guide. But something is missing and I am not sure what it is yet:
```
 cluster-api-provider-aws git:(refactor-kustomize) kustomize build config
Error: failed to apply json patch '[{"op":"replace","path":"/spec/conversion/webhook/clientConfig/service/namespace","value":"capi-webhook-system"}]': replace operation does not apply: doc is missing path: /spec/conversion/webhook/clientConfig/service/namespace: missing value
```

PR #1581, opened 2020-02-19
Adds the capa-aggregated-manager-role

h0tbird:
Fixes the error below in CAPA logs:
```
E0219 09:46:31.795354       7 reflector.go:153] pkg/mod/k8s.io/client-go@v0.17.2/tools/cache/reflector.go:105: Failed to list *v1alpha3.AWSCluster: awsclusters.infrastructure.cluster.x-k8s.io is forbidden: User "system:serviceaccount:capa-system:default" cannot list resource "awsclusters" in API group "infrastructure.cluster.x-k8s.io" at the cluster scope: RBAC: clusterrole.rbac.authorization.k8s.io "aggregated-manager-role" not found
```

I got this error after merging the PRs below:
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1577
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1575

### DIFF
--- a/config/rbac/aggregated_role.yaml
+++ b/config/rbac/aggregated_role.yaml
@@ -1,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: aggregated-manager-role
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      cluster.x-k8s.io/aggregate-to-manager: "true"
+rules: []
--- a/config/rbac/kustomization.yaml
+++ b/config/rbac/kustomization.yaml
@@ -5,6 +5,7 @@ resources:
 - role_binding.yaml
 - leader_election_role.yaml
 - leader_election_role_binding.yaml
+- aggregated_role.yaml
 - auth_proxy_service.yaml
 - auth_proxy_role.yaml
 - auth_proxy_role_binding.yaml


config/rbac/aggregated_role.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: aggregated-manager-role
+aggregationRule:
+  clusterRoleSelectors:
+  - matchLabels:
+      cluster.x-k8s.io/aggregate-to-manager: "true"
```
detiber: I know we added this to core cluster-api for allowing providers to add roles to the controller-manager for the purposes of reconciling referenced resources. I'm not sure that we need it in this provider, though.

If there are errors caused by something referencing this, then it might be better to update those references instead of adding this here.

detiber: If there is a reason that we do need it for this provider, then we likely need to chose a different label name, otherwise we'd also aggregate roles that are intended only to be aggregated for the purpose of core cluster-api and not our provider.

randomvariable: Just looking at #1575 which seems to want to consume this PR. Yes, it doesn't seem necessary here. Suggest removing this and the reference thereof in #1575.

h0tbird: Thank you all! Removed here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1575/commits/44b7898a696a172d345d0b67cff50004b0580085

h0tbird:
Closing as per comments above. Fixed in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1575/commits/44b7898a696a172d345d0b67cff50004b0580085.

PR #1585, opened 2020-02-20
:warning: instances do not require an SSH keypair

### LINKED ISSUE #1156: Option to launch instances without an SSH key pair

andrewmyhre:
**What this PR does / why we need it**: Allows instances to be provisioned with no SSH key pair specified. Organizations which have policies denying the use of SSH key pairs can now set the SSH key name value to "" to launch instances without a key pair set. A nil value still sets the default key pair name.

Fixes #1156

### DIFF
--- a/api/v1alpha2/awscluster_conversion.go
+++ b/api/v1alpha2/awscluster_conversion.go
@@ -127,6 +127,9 @@ func Convert_v1alpha2_AWSClusterSpec_To_v1alpha3_AWSClusterSpec(in *AWSClusterSp
 	// Manually convert Bastion.
 	out.Bastion.Enabled = !in.DisableBastionHost
 
+	// Manually convert SSHKeyName
+	out.SSHKeyName = &in.SSHKeyName
+
 	return nil
 }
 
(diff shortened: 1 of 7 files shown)


api/v1alpha2/awscluster_conversion.go
```diff
@@ -127,6 +127,9 @@ func Convert_v1alpha2_AWSClusterSpec_To_v1alpha3_AWSClusterSpec(in *AWSClusterSp
 	// Manually convert Bastion.
 	out.Bastion.Enabled = !in.DisableBastionHost
 
+	// Manually convert SSHKeyName
+	out.SSHKeyName = &in.SSHKeyName
+
```
detiber: Do we need to handle the reverse conversion as well?

It might be good to add a test to verify that the down conversion works as expected.

ncdc:
@andrewmyhre 👋 just checking in to see if you are able to address the code review comments?

andrewmyhre:
Closing in favor of #1621

PR #1590, opened 2020-02-21
✨ Ability to specify AWS Config Alongside Credentials

bagnaram:
 This allows the user to specify an AWS config file in the CAPA controller manager. This is needed to set up custom profiles, and to assume roles via the aws-sdk-go.


bagnaram:
Need to determine the strategy for providing config file content. Currently the credentials are passed inline base64 and I'm not sure if we want to go the same route with env vars.

detiber:
/hold
holding until we have a chance to determine what the UX experience will look like with this change

bagnaram:
Modified session creation to include additional options like credentials region, and profile loaded from the environment and shared config automatically.

config/default/manager_credentials_patch.yaml
```diff
@@ -9,12 +9,25 @@ spec:
...
-          mountPath: /home/.aws
+          mountPath: /home/.aws/credentials
+          subPath: credentials 
       volumes:
       - name: credentials
         secret:
           secretName: manager-bootstrap-credentials
+      - name: config
+        configMap:
+          name: manager-bootstrap-config
```
michaelgugino: What is this part for?

bagnaram: This specifies the AWS shared session configuration. See https://docs.aws.amazon.com/sdk-for-go/api/aws/session/

michaelgugino: Why do we need this?  Credentials file isn't enough?

bagnaram: correct. Credentials file only allows to specify default creds. It does not allow for creation of alternate profiles or ARN role assumption

michaelgugino: I haven't had any problems with using just credentials specified via the ENV vars, just that the controller wasn't picking up the default creds for one reason or another.  It seems preferrable that it should always use what's in that credentials file vs something else.

bagnaram: I thought the same but the AWS provider doesn't allow many of the SDK options to be set in the credentials file. The Issue this is linked to is the main reason for this PR. There is currently no way to specify `role_arn` or related options with just a default credentials file. Also, noticed the session is created from a new config each time in `session.go` it wouldn't use all the ENV variables being set.

michaelgugino: I would say this is a related but distinct issue from the report I originally filed.  I'm just trying to follow the quick-start exactly as-is and having a problem with the credentials file.  This looks to be more along adding additional functionality to support a variety of credentials types.

bagnaram: I would say that in order to incorporate this PR, we would need to determine the best way of passing in these parameters because the manifests will bomb if we don't at least set something inside the `AWS_B64ENCODED_CREDENTIALS` variable the way it is being substituted. I have had to give it a garbage value and override the profile with the `AWS_DEFAULT_PROFILE` to one I specify in the `config` file.

bagnaram: Perhaps instead of performing `envsubst` in the installation steps, there is a step to create the credentials secret from a file. The same can be done for the `config` secret creation. The same manifest can be used for each case, even if the secret (file) is empty.

bagnaram:
Because `infrastructure-components.yaml` gets published separately with each release, this makes it quite difficult to substitute a file in the secret. It probably cannot run kustomize every time CAPA gets installed. Does anybody have suggestions how to best handle this besides using env variable substitution? It seems that is the current option.

detiber:
@bagnaram can you not use the existing `${AWS_B64ENCODED_CREDENTIALS}` variable?

bagnaram:
Unfortunately, no as discussed above

pkg/cloud/scope/session.go
```diff
@@ -33,11 +33,21 @@ func sessionForRegion(region string) (*session.Session, error) {
 		return s.(*session.Session), nil
 	}
 
-	ns, err := session.NewSession(aws.NewConfig().WithRegion(region))
+	ns, err := session.NewSessionWithOptions(session.Options{
```
bagnaram, line 36: Would it be possible to consider the following modification below as an option to support credentials as an option? This would allow the user to specify session options in the bootstrap template until the full UX is developed.

randomvariable:
Should we rally on resolving this as part of #1713 ?
Would avoid having to fiddle too much with different CAPA pod configurations.

randomvariable:
Superseded by work for #1713 
The escape hatch which should have always existed will be provided in #1717

PR #1592, opened 2020-02-25
use a default value for AWS_REGION when creating examples

boddumanohar:
When I run `make generate-examples` I get the following error. 

```
./examples/generate.sh
Generated /home/manohar/kubernetes/cluster-api-provider-aws/examples/_out/cert-manager.yaml
Error: environment variable "AWS_REGION" not found
Usage:
  clusterawsadm alpha bootstrap encode-aws-credentials [flags]
Flags:
  -h, --help   help for encode-aws-credentials
Global Flags:
      --partition string   AWS partition, for AWS GovCloud (US) it is aws-us-gov (default "aws")
environment variable "AWS_REGION" not found
Makefile:182: recipe for target 'generate-examples' failed
make: *** [generate-examples] Error 1
```
This means that the environment variable `AWS_REGION` is not set. 

Since this is only for the sake of generating examples, we can use any default for this variable. If variable is already set, we can continue with already existing value. 

So made changes to check if the variable exists and set to `us-east-1` if the variables doesn't exist.

### DIFF
--- a/examples/generate.sh
+++ b/examples/generate.sh
@@ -37,6 +37,15 @@ export CONTROL_PLANE_MACHINE_TYPE="${CONTROL_PLANE_MACHINE_TYPE:-t2.medium}"
 export NODE_MACHINE_TYPE="${CONTROL_PLANE_MACHINE_TYPE:-t2.medium}"
 export SSH_KEY_NAME="${SSH_KEY_NAME:-default}"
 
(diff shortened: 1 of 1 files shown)


boddumanohar:
@detiber thanks for reviewing this.

detiber:
We no longer have example generation in favor of using the new `clusterctl` workflow with templates.
/close

PR #1612, opened 2020-03-02
:sparkles: Bring AWS SDK in-line with Kubernetes master and add new regions

voor:
**What this PR does / why we need it**: Adds support for additional AWS region endpoints (AWS ISO regions) allowing for support in those areas without explicitly overriding the AWS endpoint URLs.

Based on discovery in slack thread https://kubernetes.slack.com/archives/CD6U2V71N/p1583175642128100?thread_ts=1583161863.102600&cid=CD6U2V71N


detiber:
/ok-to-test

@voor Are there any additional changes needed for the region support, or is bumping the aws sdk enough?

PR #1617, opened 2020-03-04
controllers: Implement pause mechanism

### LINKED ISSUE #1603: Implement pause mechanism where required
/kind feature

Port https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/785 to support clusterctl move

randomvariable:
Implements the pause mechanism.

Fixes #1603

Testing.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -87,11 +87,11 @@ help:  ## Display this help
 
 .PHONY: test
 test: ## Run tests
-	go test -v ./...
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v ./...
 
 .PHONY: test-integration
 test-integration: ## Run integration tests
-	go test -v -tags=integration ./test/integration/...
+	source ./scripts/fetch_ext_bins.sh; fetch_tools; setup_envs; go test -v -tags=integration ./test/integration/...
 
 .PHONY: test-e2e
 test-e2e: ## Run e2e tests
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -26,13 +26,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
-	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/ec2"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/elb"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/secretsmanager"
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -314,9 +314,9 @@ create-cluster: $(CLUSTERCTL) $(ENVSUBST) ## Create a development Kubernetes clu
 	kubectl wait --for=condition=Available --timeout=5m apiservice v1beta1.webhook.cert-manager.io
 
 	# Deploy CAPI
-	kustomize build github.com/kubernetes-sigs/cluster-api//config?ref=v0.3.0-rc.3 | kubectl apply -f -
-	kustomize build github.com/kubernetes-sigs/cluster-api//bootstrap/kubeadm/config?ref=v0.3.0-rc.3 | kubectl apply -f -
-	kustomize build github.com/kubernetes-sigs/cluster-api//controlplane/kubeadm/config?ref=v0.3.0-rc.3 | kubectl apply -f -
+	kustomize build github.com/kubernetes-sigs/cluster-api//config?ref=v0.3.0-rc.2 | kubectl apply -f -
+	kustomize build github.com/kubernetes-sigs/cluster-api//bootstrap/kubeadm/config?ref=v0.3.0-rc.2 | kubectl apply -f -
+	kustomize build github.com/kubernetes-sigs/cluster-api//controlplane/kubeadm/config?ref=v0.3.0-rc.2 | kubectl apply -f -
```
detiber: This bit should be reverted back to rc.3

controllers/awsmachine_controller.go
```diff
@@ -179,7 +186,32 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+		&handler.EnqueueRequestsFromMapFunc{
+			ToRequests: handler.ToRequestsFunc(r.requeueAWSMachinesForUnpausedCluster),
+		},
+		predicate.Funcs{
+			UpdateFunc: func(e event.UpdateEvent) bool {
+				oldCluster := e.ObjectOld.(*clusterv1.Cluster)
+				newCluster := e.ObjectNew.(*clusterv1.Cluster)
+				return oldCluster.Spec.Paused && !newCluster.Spec.Paused
+			},
+			CreateFunc: func(e event.CreateEvent) bool {
```
detiber: Should this return false in that case that `cluster.Spec.Paused == true`?

controllers/awsmachine_controller.go
```diff
@@ -179,7 +186,32 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+				if _, ok := e.Meta.GetAnnotations()[clusterv1.PausedAnnotation]; !ok {
+					return false
```
detiber: Should this return `true` instead? I think we'd only want to return `false` here if the annotation is present rather than absent.

That does make me wonder if we should even really care about the annotation on the Cluster resource here, since the annotations aren't expected to have a cascading effect in the same way that the `cluster.Spec.Paused` field should.

controllers/awsmachine_controller.go
randomvariable: Does seem that way. Maaaay have to open an issue in CAPV...

randomvariable:
ah, i mean
assign @vincepri 
@detiber 's updated as appropriate.

detiber:
Added lgtm after verifying changes looked good to @randomvariable over slack

PR #1621, opened 2020-03-05
:warning: 1156 allow no key pair v1alpha3 2020

### LINKED ISSUE #1156: Option to launch instances without an SSH key pair

andrewmyhre:
**What this PR does / why we need it**: Allows instances to be provisioned with no SSH key pair specified. Organizations which have policies denying the use of SSH key pairs can now set the SSH key name value to "" to launch instances without a key pair set. A nil value still sets the default key pair name.

Fixes #1156

### DIFF
--- a/api/v1alpha2/awscluster_conversion.go
+++ b/api/v1alpha2/awscluster_conversion.go
@@ -20,6 +20,7 @@ import (
 	"reflect"
 
 	apiconversion "k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/utils/pointer"
 	infrav1alpha3 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
 	v1alpha3 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
 	utilconversion "sigs.k8s.io/cluster-api/util/conversion"
@@ -45,6 +46,11 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 	restored := &infrav1alpha3.AWSCluster{}
 	if ok, err := utilconversion.UnmarshalData(src, restored); err != nil || !ok {
 		return err
+	}
+
+	// override the SSHKeyName conversion if we are roundtripping from v1alpha3 and the v1alpha3 value is nil
+	if src.Spec.SSHKeyName == "" && restored.Spec.SSHKeyName == nil {
+		dst.Spec.SSHKeyName = nil
 	}
 
 	dst.Spec.ImageLookupOrg = restored.Spec.ImageLookupOrg
@@ -127,6 +133,9 @@ func Convert_v1alpha2_AWSClusterSpec_To_v1alpha3_AWSClusterSpec(in *AWSClusterSp
 	// Manually convert Bastion.
 	out.Bastion.Enabled = !in.DisableBastionHost
 
+	// Manually convert SSHKeyName
+	out.SSHKeyName = pointer.StringPtr(in.SSHKeyName)
+
 	return nil
 }
 
@@ -139,6 +148,11 @@ func Convert_v1alpha3_AWSClusterSpec_To_v1alpha2_AWSClusterSpec(in *infrav1alpha
 
 	// Manually convert DisableBastionHost.
 	out.DisableBastionHost = !in.Bastion.Enabled
+
+	// Manually convert SSHKeyName
+	if in.SSHKeyName != nil {
+		out.SSHKeyName = *in.SSHKeyName
+	}
 
 	return nil
 }
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -35,8 +35,9 @@ type AWSClusterSpec struct {
 	// The AWS Region the cluster lives in.
 	Region string `json:"region,omitempty"`
 
-	// SSHKeyName is the name of the ssh key to attach to the bastion host.
-	SSHKeyName string `json:"sshKeyName,omitempty"`
+	// SSHKeyName is the name of the ssh key to attach to the bastion host. Valid values are empty string (do not use SSH keys), a valid SSH key name, or omitted (use the default SSH key name)
+	// +optional
+	SSHKeyName *string `json:"sshKeyName,omitempty"`
 
 	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
 	// +optional
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
@@ -596,7 +596,8 @@ spec:
                 type: string
               sshKeyName:
                 description: SSHKeyName is the name of the ssh key to attach to the
-                  bastion host.
+                  bastion host. Valid values are empty string (do not use SSH keys),
+                  a valid SSH key name, or omitted (use the default SSH key name)
                 type: string
             type: object
           status:
--- a/pkg/cloud/services/ec2/bastion.go
+++ b/pkg/cloud/services/ec2/bastion.go
@@ -139,16 +139,17 @@ func (s *Service) getDefaultBastion() *infrav1.Instance {
 	name := fmt.Sprintf("%s-bastion", s.scope.Name())
 	userData, _ := userdata.NewBastion(&userdata.BastionInput{})
 
-	keyName := defaultSSHKeyName
-	if s.scope.AWSCluster.Spec.SSHKeyName != "" {
-		keyName = s.scope.AWSCluster.Spec.SSHKeyName
+	// If SSHKeyName WAS NOT provided, use the defaultSSHKeyName
+	keyName := s.scope.AWSCluster.Spec.SSHKeyName
+	if keyName == nil {
+		keyName = aws.String(defaultSSHKeyName)
 	}
 
 	i := &infrav1.Instance{
 		Type:       "t2.micro",
 		SubnetID:   s.scope.Subnets().FilterPublic()[0].ID,
 		ImageID:    s.defaultBastionAMILookup(s.scope.AWSCluster.Spec.Region),
-		SSHKeyName: aws.String(keyName),
+		SSHKeyName: keyName,
 		UserData:   aws.String(base64.StdEncoding.EncodeToString([]byte(userData))),
 		SecurityGroupIDs: []string{
 			s.scope.Network().SecurityGroups[infrav1.SecurityGroupBastion].ID,
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -208,12 +208,15 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
 	}
 	input.SecurityGroupIDs = append(input.SecurityGroupIDs, ids...)
 
-	// Pick SSH key, if any.
-	input.SSHKeyName = aws.String(defaultSSHKeyName)
-	if scope.AWSMachine.Spec.SSHKeyName != "" {
-		input.SSHKeyName = aws.String(scope.AWSMachine.Spec.SSHKeyName)
-	} else if scope.AWSCluster.Spec.SSHKeyName != "" {
-		input.SSHKeyName = aws.String(scope.AWSCluster.Spec.SSHKeyName)
+	// If SSHKeyName WAS NOT provided in the AWSMachine Spec, fallback to the value provided in the AWSCluster Spec.
+	// If a value was not provided in the AWSCluster Spec, then use the defaultSSHKeyName
+	input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
+	if input.SSHKeyName == nil {
+		if scope.AWSCluster.Spec.SSHKeyName != nil {
+			input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
+		} else {
+			input.SSHKeyName = aws.String(defaultSSHKeyName)
+		}
 	}
 
 	s.scope.V(2).Info("Running instance", "machine-role", scope.Role())
--- a/api/v1alpha2/awscluster_conversion_test.go
+++ b/api/v1alpha2/awscluster_conversion_test.go
@@ -1,0 +1,46 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha2
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	infrav1alpha3 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
+)
+
+func TestConvertAWSCluster(t *testing.T) {
+	g := NewWithT(t)
+
+	t.Run("from hub", func(t *testing.T) {
+		t.Run("should restore SSHKeyName, retaining a nil value", func(t *testing.T) {
+			src := &infrav1alpha3.AWSCluster{
+				ObjectMeta: metav1.ObjectMeta{},
+				Spec: infrav1alpha3.AWSClusterSpec{
+					SSHKeyName: nil,
+				},
+			}
+			dst := &AWSCluster{}
+			g.Expect(dst.ConvertFrom(src)).To(Succeed())
+			restored := &infrav1alpha3.AWSCluster{}
+			g.Expect(dst.ConvertTo(restored)).To(Succeed())
+			g.Expect(restored.Spec.SSHKeyName).To(BeNil())
+		})
+	})
+
+}
--- a/api/v1alpha2/awsmachine_conversion.go
+++ b/api/v1alpha2/awsmachine_conversion.go
@@ -18,6 +18,7 @@ package v1alpha2
 
 import (
 	apiconversion "k8s.io/apimachinery/pkg/conversion"
+	"k8s.io/utils/pointer"
 	infrav1alpha3 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
 	utilconversion "sigs.k8s.io/cluster-api/util/conversion"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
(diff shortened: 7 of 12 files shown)


andrewmyhre:
@detiber @ncdc I've added the down-conversion for SSHKeyName, but I'm having trouble figuring out how to prevent conversion-gen from adding a conversion line into zz_generated.conversion.go. I've added `// +k8s:conversion-gen=false` to the SSHKeyName fields in both v1alpha2 and v1alpha3, and I'm expecting it to remove those auto conversion lines. Can anyone give me some guidance there? Once that's resolved and the compiler error goes away I can finish implementing the test.

api/v1alpha2/awscluster_conversion.go
```diff
@@ -140,6 +143,9 @@ func Convert_v1alpha3_AWSClusterSpec_To_v1alpha2_AWSClusterSpec(in *infrav1alpha
 	// Manually convert DisableBastionHost.
 	out.DisableBastionHost = !in.Bastion.Enabled
 
+	// Manually convert SSHKeyName
+	out.SSHKeyName = *in.SSHKeyName
```
ncdc: You'll want to do a nil check and only set if non-nil

ncdc:
@andrewmyhre I don't see any generated conversion funcs in this PR, but I do see generated deepcopy. Is that what you meant?

api/v1alpha2/awscluster_types.go
```diff
@@ -27,6 +27,7 @@ const (
 )
 
 // AWSClusterSpec defines the desired state of AWSCluster
+// +k8s:conversion-gen=false
```
ncdc: Please remove 😄

api/v1alpha2/awscluster_types.go
```diff
@@ -35,6 +36,7 @@ type AWSClusterSpec struct {
 	Region string `json:"region,omitempty"`
 
 	// SSHKeyName is the name of the ssh key to attach to the bastion host.
+	// +k8s:conversion-gen=false
```
ncdc: Please remove 😄

api/v1alpha2/awsmachine_types.go
```diff
@@ -77,7 +77,7 @@ type AWSMachineSpec struct {
 	Subnet *AWSResourceReference `json:"subnet,omitempty"`
 
 	// SSHKeyName is the name of the ssh key to attach to the instance.
-	SSHKeyName string `json:"sshKeyName,omitempty"`
+	SSHKeyName *string `json:"sshKeyName,omitempty"`
```
ncdc: This needs to stay a non-pointer

api/v1alpha3/awscluster_types.go
```diff
@@ -36,7 +36,12 @@ type AWSClusterSpec struct {
 	Region string `json:"region,omitempty"`
 
 	// SSHKeyName is the name of the ssh key to attach to the bastion host.
-	SSHKeyName string `json:"sshKeyName,omitempty"`
+	// If nil, will use a default SSH key pair name
+	// If empty string, will NOT set an SSH key pair
+	// Otherwise, will set the SSH key pair name
+	// +optional
+	// +k8s:conversion-gen=false
```
ncdc: Please remove 😄

andrewmyhre:
Fixed the rookie mistakes :)

The reason you're not seeing it in the PR is the SSHKeyName had already been added as a literal in both versions, so there was already an auto-conversion line for it. I need to figure out how to get it removed. Here it is in current master:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/api/v1alpha2/zz_generated.conversion.go#L432

config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
```diff
@@ -596,7 +596,9 @@ spec:
                 type: string
               sshKeyName:
                 description: SSHKeyName is the name of the ssh key to attach to the
-                  bastion host.
+                  bastion host. If nil, will use a default SSH key pair name If empty
```
ncdc: You may want to adjust your comments in the types you've adjusted in this PR so these display correctly (the generator ignores newlines)

ncdc:
Ok, you don't need to handle conversion of SSHKeyName. The generated conversions will do it (ignore the irrelevant diffs - my setup is not 100% right):

```diff
diff --git a/api/v1alpha2/zz_generated.conversion.go b/api/v1alpha2/zz_generated.conversion.go
index 02f0f0f3..c0c2e6f0 100644
--- a/api/v1alpha2/zz_generated.conversion.go
+++ b/api/v1alpha2/zz_generated.conversion.go
@@ -24,7 +24,8 @@ import (
 	time "time"
 	unsafe "unsafe"
 
-	v1 "k8s.io/api/core/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	conversion "k8s.io/apimachinery/pkg/conversion"
 	runtime "k8s.io/apimachinery/pkg/runtime"
 	v1alpha3 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
@@ -409,7 +410,9 @@ func autoConvert_v1alpha2_AWSClusterSpec_To_v1alpha3_AWSClusterSpec(in *AWSClust
 		return err
 	}
 	out.Region = in.Region
-	out.SSHKeyName = in.SSHKeyName
+	if err := v1.Convert_string_To_Pointer_string(&in.SSHKeyName, &out.SSHKeyName, s); err != nil {
+		return err
+	}
 	out.AdditionalTags = *(*v1alpha3.Tags)(unsafe.Pointer(&in.AdditionalTags))
 	if in.ControlPlaneLoadBalancer != nil {
 		in, out := &in.ControlPlaneLoadBalancer, &out.ControlPlaneLoadBalancer
@@ -429,7 +432,9 @@ func autoConvert_v1alpha3_AWSClusterSpec_To_v1alpha2_AWSClusterSpec(in *v1alpha3
 		return err
 	}
 	out.Region = in.Region
-	out.SSHKeyName = in.SSHKeyName
+	if err := v1.Convert_Pointer_string_To_string(&in.SSHKeyName, &out.SSHKeyName, s); err != nil {
+		return err
+	}
 	// WARNING: in.ControlPlaneEndpoint requires manual conversion: does not exist in peer-type
 	out.AdditionalTags = *(*Tags)(unsafe.Pointer(&in.AdditionalTags))
 	if in.ControlPlaneLoadBalancer != nil {
@@ -451,7 +456,7 @@ func autoConvert_v1alpha2_AWSClusterStatus_To_v1alpha3_AWSClusterStatus(in *AWSC
 	if err := Convert_v1alpha2_Network_To_v1alpha3_Network(&in.Network, &out.Network, s); err != nil {
 		return err
 	}
-	// WARNING: in.Bastion requires manual conversion: inconvertible types (sigs.k8s.io/cluster-api-provider-aws/api/v1alpha2.Instance vs *sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.Instance)
+	// WARNING: in.Bastion requires manual conversion: inconvertible types (./api/v1alpha2.Instance vs *sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.Instance)
 	out.Ready = in.Ready
 	// WARNING: in.APIEndpoints requires manual conversion: does not exist in peer-type
 	return nil
@@ -463,7 +468,7 @@ func autoConvert_v1alpha3_AWSClusterStatus_To_v1alpha2_AWSClusterStatus(in *v1al
 		return err
 	}
 	// WARNING: in.FailureDomains requires manual conversion: does not exist in peer-type
-	// WARNING: in.Bastion requires manual conversion: inconvertible types (*sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.Instance vs sigs.k8s.io/cluster-api-provider-aws/api/v1alpha2.Instance)
+	// WARNING: in.Bastion requires manual conversion: inconvertible types (*sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.Instance vs ./api/v1alpha2.Instance)
 	return nil
 }
 
@@ -570,10 +575,12 @@ func autoConvert_v1alpha2_AWSMachineSpec_To_v1alpha3_AWSMachineSpec(in *AWSMachi
 	out.AdditionalSecurityGroups = *(*[]v1alpha3.AWSResourceReference)(unsafe.Pointer(&in.AdditionalSecurityGroups))
 	// WARNING: in.AvailabilityZone requires manual conversion: does not exist in peer-type
 	out.Subnet = (*v1alpha3.AWSResourceReference)(unsafe.Pointer(in.Subnet))
-	out.SSHKeyName = in.SSHKeyName
+	if err := v1.Convert_string_To_Pointer_string(&in.SSHKeyName, &out.SSHKeyName, s); err != nil {
+		return err
+	}
 	// WARNING: in.RootDeviceSize requires manual conversion: does not exist in peer-type
 	out.NetworkInterfaces = *(*[]string)(unsafe.Pointer(&in.NetworkInterfaces))
-	// WARNING: in.CloudInit requires manual conversion: inconvertible types (*sigs.k8s.io/cluster-api-provider-aws/api/v1alpha2.CloudInit vs sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.CloudInit)
+	// WARNING: in.CloudInit requires manual conversion: inconvertible types (*./api/v1alpha2.CloudInit vs sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.CloudInit)
 	return nil
 }
 
@@ -591,16 +598,18 @@ func autoConvert_v1alpha3_AWSMachineSpec_To_v1alpha2_AWSMachineSpec(in *v1alpha3
 	out.AdditionalSecurityGroups = *(*[]AWSResourceReference)(unsafe.Pointer(&in.AdditionalSecurityGroups))
 	// WARNING: in.FailureDomain requires manual conversion: does not exist in peer-type
 	out.Subnet = (*AWSResourceReference)(unsafe.Pointer(in.Subnet))
-	out.SSHKeyName = in.SSHKeyName
+	if err := v1.Convert_Pointer_string_To_string(&in.SSHKeyName, &out.SSHKeyName, s); err != nil {
+		return err
+	}
 	// WARNING: in.RootVolume requires manual conversion: does not exist in peer-type
 	out.NetworkInterfaces = *(*[]string)(unsafe.Pointer(&in.NetworkInterfaces))
-	// WARNING: in.CloudInit requires manual conversion: inconvertible types (sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.CloudInit vs *sigs.k8s.io/cluster-api-provider-aws/api/v1alpha2.CloudInit)
+	// WARNING: in.CloudInit requires manual conversion: inconvertible types (sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.CloudInit vs *./api/v1alpha2.CloudInit)
 	return nil
 }
 
 func autoConvert_v1alpha2_AWSMachineStatus_To_v1alpha3_AWSMachineStatus(in *AWSMachineStatus, out *v1alpha3.AWSMachineStatus, s conversion.Scope) error {
 	out.Ready = in.Ready
-	out.Addresses = *(*[]v1.NodeAddress)(unsafe.Pointer(&in.Addresses))
+	out.Addresses = *(*[]corev1.NodeAddress)(unsafe.Pointer(&in.Addresses))
 	out.InstanceState = (*v1alpha3.InstanceState)(unsafe.Pointer(in.InstanceState))
 	// WARNING: in.ErrorReason requires manual conversion: does not exist in peer-type
 	// WARNING: in.ErrorMessage requires manual conversion: does not exist in peer-type
@@ -609,7 +618,7 @@ func autoConvert_v1alpha2_AWSMachineStatus_To_v1alpha3_AWSMachineStatus(in *AWSM
 
 func autoConvert_v1alpha3_AWSMachineStatus_To_v1alpha2_AWSMachineStatus(in *v1alpha3.AWSMachineStatus, out *AWSMachineStatus, s conversion.Scope) error {
 	out.Ready = in.Ready
-	out.Addresses = *(*[]v1.NodeAddress)(unsafe.Pointer(&in.Addresses))
+	out.Addresses = *(*[]corev1.NodeAddress)(unsafe.Pointer(&in.Addresses))
 	out.InstanceState = (*InstanceState)(unsafe.Pointer(in.InstanceState))
 	// WARNING: in.FailureReason requires manual conversion: does not exist in peer-type
 	// WARNING: in.FailureMessage requires manual conversion: does not exist in peer-type
@@ -968,7 +977,7 @@ func autoConvert_v1alpha2_Instance_To_v1alpha3_Instance(in *Instance, out *v1alp
 	out.SecurityGroupIDs = *(*[]string)(unsafe.Pointer(&in.SecurityGroupIDs))
 	out.UserData = (*string)(unsafe.Pointer(in.UserData))
 	out.IAMProfile = in.IAMProfile
-	out.Addresses = *(*[]v1.NodeAddress)(unsafe.Pointer(&in.Addresses))
+	out.Addresses = *(*[]corev1.NodeAddress)(unsafe.Pointer(&in.Addresses))
 	out.PrivateIP = (*string)(unsafe.Pointer(in.PrivateIP))
 	out.PublicIP = (*string)(unsafe.Pointer(in.PublicIP))
 	out.ENASupport = (*bool)(unsafe.Pointer(in.ENASupport))
@@ -989,7 +998,7 @@ func autoConvert_v1alpha3_Instance_To_v1alpha2_Instance(in *v1alpha3.Instance, o
 	out.SecurityGroupIDs = *(*[]string)(unsafe.Pointer(&in.SecurityGroupIDs))
 	out.UserData = (*string)(unsafe.Pointer(in.UserData))
 	out.IAMProfile = in.IAMProfile
-	out.Addresses = *(*[]v1.NodeAddress)(unsafe.Pointer(&in.Addresses))
+	out.Addresses = *(*[]corev1.NodeAddress)(unsafe.Pointer(&in.Addresses))
 	out.PrivateIP = (*string)(unsafe.Pointer(in.PrivateIP))
 	out.PublicIP = (*string)(unsafe.Pointer(in.PublicIP))
 	out.ENASupport = (*bool)(unsafe.Pointer(in.ENASupport))
```

pkg/cloud/services/ec2/bastion.go
```diff
@@ -139,9 +139,16 @@ func (s *Service) getDefaultBastion() *infrav1.Instance {
...
+	// If SSHKeyName WAS provided _but is an empty string_, do not set a key pair
+	// Otherwise (SSHKeyName WAS provided and is NOT an empty string), use the provided key pair name
+	var keyName string
+	if s.scope.AWSCluster.Spec.SSHKeyName != nil {
+		if *s.scope.AWSCluster.Spec.SSHKeyName != "" {
+			keyName = *s.scope.AWSCluster.Spec.SSHKeyName
+		}
+	} else {
+		keyName = defaultSSHKeyName
 	}
```
detiber: I think it might be a bit easier to read if we adjust the logic a bit here.
```suggestion
	keyName := s.scope.AWSCluster.Spec.SSHKeyName
	if keyName == nil {
		keyName = defaultSSHKeyName
	}
```

pkg/cloud/services/ec2/instances.go
```diff
@@ -209,11 +209,19 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
...
+		if *scope.AWSMachine.Spec.SSHKeyName != "" {
+			input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
+		}
+	} else if scope.AWSCluster.Spec.SSHKeyName != nil {
+		if *scope.AWSCluster.Spec.SSHKeyName != "" {
+			input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
+		}
+	} else {
+		input.SSHKeyName = aws.String(defaultSSHKeyName)
 	}
```
detiber: I tried to make this a bit cleaner looking, but I'm not entirely convinced I did :joy:
```suggestion
	input.SSHKeyName := scope.AWSMachine.Spec.SSHKeyName
	if input.SSHKeyName == nil {
		if scope.AWSCluster.Spec.SSHKeyName != nil {
			input.SSHKeyName := scope.AWSCluster.Spec.SSHKeyName
		} else {
			input.SSHKeyName = defaultSSHKeyName
		}
	} 
```

ncdc:
/hold
Need to get the conversion bits regenerated too

pkg/cloud/services/ec2/bastion.go
andrewmyhre: I think that wouldn't handle the case where `SSHKeyName == ""` (which specifies _not_ to use an SSH key).

How about this?
```
	keyName := ""
	if s.scope.AWSCluster.Spec.SSHKeyName == nil {
		keyName = defaultSSHKeyName
	} else {
		keyName = *s.scope.AWSCluster.Spec.SSHKeyName
	}
```

Or should I consider adding a new bool `UseSSHKeys` instead?

andrewmyhre:
@ncdc Yes I was expecting `make generate` to modify the zz_generated.conversion.go file on my system, but it's not. Am I missing something?

ncdc:
@andrewmyhre do you have your GOPATH set and/or is your repo cloned into ~/go/src/...? We've had a lot of issues related to that. @detiber @vincepri and/or @randomvariable may be able to provide more insight. I would recommend deleting the `zz_generated.conversion.go` file, running `make generate`, and checking to see if it regenerated it. If not, it may be in `~/go` somewhere, and we need to get your environment updated (env vars or whatnot) so generation puts it in the right place.

pkg/cloud/services/ec2/bastion.go
detiber: Oops, I missed something with the type change, the below should work:

```suggestion
	keyName := s.scope.AWSCluster.Spec.SSHKeyName
	if keyName == nil {
		keyName = aws.String(defaultSSHKeyName)
	}
```

The use of keyName below would need to be changed to remove the wrapping `aws.String()` call.

This should work, because if s.scope.AWSCluster.Spec.SSHKeyName == "" (or any non-nil value), then we wouldn't set keyName to a pointer to defaultSSHKeyName, so a user-specified "" would take precedence over the default key as we would expect.

andrewmyhre: This is nice, I've changed the code

detiber:
I was finally able to get conversion-gen to do the right thing by symlinking my repo under $GOPATH and running `make generate` from there. Can't say I like that as a solution, but at least it got things moving again.

I do worry a bit about the behavior for round-tripping, since the round-trip from v1alpha3 -> v1alpha2 -> v1alpha3 would essentially convert a nil SSHKeyName into an empty string.

I think we can likely work around this and I'll work on seeing if I can work up something to work around this.

detiber:
Ok, I think the commits I added resolve the outstanding issues. @andrewmyhre @randomvariable ptal.

pkg/cloud/services/ec2/bastion.go
```diff
@@ -139,16 +139,19 @@ func (s *Service) getDefaultBastion() *infrav1.Instance {
 	name := fmt.Sprintf("%s-bastion", s.scope.Name())
 	userData, _ := userdata.NewBastion(&userdata.BastionInput{})
 
-	keyName := defaultSSHKeyName
-	if s.scope.AWSCluster.Spec.SSHKeyName != "" {
-		keyName = s.scope.AWSCluster.Spec.SSHKeyName
+	// If SSHKeyName WAS NOT provided (nil) then use the defaultSSHKeyName
```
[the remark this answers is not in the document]
randomvariable: I have pretty much the same comment as @chuckha - it's not clear how the logic in the comments are realised.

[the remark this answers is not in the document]
randomvariable: Otherwise
/lgtm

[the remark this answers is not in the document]
detiber: Thanks! Updated the comments, rebased and squashed. Should be ready for another look :)

PR #1626, opened 2020-03-06
:running: Update the CAPI components used for conformance testing

detiber:

Updates the CAPI components used for conformance testing to released artifacts rather than using kustomize


ncdc:
/lgtm
Thanks! Do we have an issue to switch this make target to use `clusterctl` in the future? Or does it maybe go away as we can just tell people to run `clusterctl` on their own?

detiber:
> Thanks! Do we have an issue to switch this make target to use clusterctl in the future? Or does it maybe go away as we can just tell people to run clusterctl on their own?

Not yet, afaict this target is only really used by the conformance testing and I would expect most developers are using tilt at this point, but I was planning on trying to rationalize some of that after we get through the v1alpha3 releases.

PR #1646, opened 2020-03-14
:sparkles: Enable overriding security group roles with existing groups

### LINKED ISSUE #1137: BYO security groups
/kind feature

In our organization security group management is tightly controlled - it is not possible to grant processes or users the permission to manage security groups. They are pre-provisioned through a partially manual process. There are also specific security groups our organization requires attached to all instances, so we need to bring those into the mix as well. I would like to be able to specify the particular security groups existing in our VPCs which should be attached to the instances and load balancers created by CAPA.

`AWSMachineSpec` already has an `AdditionalSecurityGroups` property, so my proposal is to add an optional `CreateSecurityGroups` flag to the `AWSCluster.Spec.NetworkSpec` struct. The flag would default to `true`, and when set to `false` the controller would not attempt to create security groups and instead only use the groups defined in `AdditionalSecurityGroups`.

**Environment:**

- OS (e.g. from `/etc/os-release`): darwin

andrewmyhre:
**What this PR does / why we need it**: In highly regulated environments it may not be possible to run a service which has permission to create security groups except through an out-of-band review process. This PR enables Cluster API to be used in such environments by providing a way to specify the security groups that should be used for each role. The security group creation process is skipped when these are provided. 

The `NetworkSpec` type has a new field:

```golang
	// SecurityGroupOverrides is an optional set of security groups to use for cluster instances
	// +optional
	SecurityGroupOverrides map[SecurityGroupRole]AWSResourceReference `json:"securityGroupOverrides,omitempty"`}

```

The `AWSLoadBalancerSpec` type also has a new field:

```golang
	// SecurityGroups sets the security groups used by the load balancer
	// +optional
	SecurityGroups []SecurityGroup `json:"securityGroups,omitempty"`
```

Security groups can be overridden like so:

```yaml
kind: AWSCluster
metadata:
  name: my-cluster
spec:
  networkSpec:
    securityGroupOverrides:
      bastion:
        id: sg-bastion
      controlplane:
        id: sg-controlplane
      apiserver-lb:
        id: sg-apiserver-lb
      node:
        id: sg-node
      lb:
        id: sg-lb
  controlPlaneLoadBalancer:
    securityGroups:
    - id: sg-load-balancer-1
      name: "Load balancer security group 1"
    - id: sg-load-balancer-2
      name: "Load balancer security group 2"
```

When security groups have been overridden in this way, it's probably not desirable for the groups to be tagged for each cluster being provisioned. For this case the `NetworkSpec` has an additional new field:

```golang
	// TagSecurityGroups optionally disables tagging the security groups used by the cluster
	// +optional
	TagSecurityGroups *bool `json:"tagSecurityGroups,omitempty"`
```

This can be enabled in the following way:
```yaml
kind: AWSCluster
metadata:
  name: my-cluster
spec:
  networkSpec:
    tagSecurityGroups: false
````

When these fields are omitted the behavior matches the existing behavior.

Notes:
- When security groups are being created, they will be tagged regardless of the `TagSecurityGroups` setting.
- Because these fields don't exist in the v1alpha2 types, downgrading to v1alpha2 and then back to v1alpha3 loses data. Do I need to handle this? If so, how?

Fixes #1137

### CODE REVIEW

File: api/v1alpha2/zz_generated.deepcopy.go [hunk 1]
```diff
@@ -158,6 +158,13 @@ func (in *AWSLoadBalancerSpec) DeepCopyInto(out *AWSLoadBalancerSpec) {
 		*out = new(ClassicELBScheme)
 		**out = **in
 	}
+	if in.SecurityGroups != nil {
+		in, out := &in.SecurityGroups, &out.SecurityGroups
+		*out = make([]SecurityGroup, len(*in))
+		for i := range *in {
+			(*in)[i].DeepCopyInto(&(*out)[i])
+		}
+	}
```
detiber: Is this change intentional? I wouldn't expect the v1alpha2 generated deep copy to change given the rest of the changes in this PR.

api/v1alpha3/types.go
```diff
@@ -179,7 +179,14 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
-}
+
+	// TagSecurityGroups optionally disables tagging the security groups used by the cluster
+	// +optional
+	TagSecurityGroups *bool `json:"tagSecurityGroups,omitempty"`
```
detiber: Given that we are going to tag security groups that we create, and I would assume provided security groups would have separate lifecycle management, do we actually need to worry about exposing the ability to disable tagging here?

cluster-api-provider-aws-examples.tar
```diff
@@ -0,0 +1 @@
+Not Found
```
detiber: was this accidentally included?

api/v1alpha3/awscluster_types.go
```diff
@@ -92,6 +92,10 @@ type AWSLoadBalancerSpec struct {
 	// Defaults to false.
 	// +optional
 	CrossZoneLoadBalancing bool `json:"crossZoneLoadBalancing,omitempty"`
+
+	// SecurityGroups sets the security groups used by the load balancer
+	// +optional
+	SecurityGroups []SecurityGroup `json:"securityGroups,omitempty"`
```
detiber: Do we actually expect users to provide more information here than possibly a SecurityGroup ID? If not, then this should probably use a new type, otherwise users might be confused by what options would be respected/ignored when used in this context.

api/v1alpha3/types.go
```diff
@@ -179,7 +179,14 @@ type NetworkSpec struct {
...
+
+	// SecurityGroupOverrides is an optional set of security groups to use for cluster instances
+	// +optional
+	SecurityGroupOverrides map[SecurityGroupRole]AWSResourceReference `json:"securityGroupOverrides,omitempty"`}
```
detiber: Do we actually expect users to provide more information here than possibly a SecurityGroup ID? If not, then this should probably use a new type, otherwise users might be confused by what options would be respected/ignored when used in this context.

This has already caused confusion in the existing usage of the AWSResourceReference type.

config/default/manager_pull_policy.yaml-e
```diff
@@ -0,0 +1,11 @@
+apiVersion: apps/v1
```
detiber: probably need to add `config/default/manager_pull_policy.yaml-e` to .gitignore

config/manager/manager_image_patch.yaml
```diff
@@ -8,5 +8,5 @@ spec:
     spec:
       containers:
       # Change the value of image field below to your controller image URL
-      - image: gcr.io/k8s-staging-cluster-api-aws/cluster-api-aws-controller:latest
+      - image: artifactory.cloud.capitalone.com/k8s-multicluster/cluster-api-aws-controller-amd64:dev
```
detiber: Please revert this change

api/v1alpha3/types.go
andrewmyhre: Sounds good to me! Will update.

api/v1alpha2/zz_generated.deepcopy.go
andrewmyhre: You're right, I think I forgot to run `make generate` before pushing

api/v1alpha3/awscluster_types.go
andrewmyhre: Great call out, how about I make this a `[]string` instead?

api/v1alpha3/types.go
andrewmyhre: Cool, I'll make this `map[SecurityGroupRole]string`

detiber: sgtm, thanks!

api/v1alpha3/awscluster_types.go
detiber: sgtm, thanks!

api/v1alpha3/types.go
detiber: sgtm, thanks!

config/manager/manager_image_patch.yaml
andrewmyhre: Thanks, reverted

andrewmyhre:
Updated to address the comments, the PR is showing out of date code though. Maybe I need to close and create a new PR?

andrewmyhre:
I'm confused about the failing tests. The output says:
```
aws/api/v1alpha2/zz_generated.conversion.go:431:12: undeclared name: Convert_v1alpha3_NetworkSpec_To_v1alpha2_NetworkSpec
```

There is a function in my `api/v1alpha2/zz_generated.conversion.go` called Convert_v1alpha3_NetworkSpec_To_v1alpha2_NetworkSpec so I'm guessing this is a PR staleness issue?

andrewmyhre:
Is this being held or are there updates I need to make?

detiber:
/assign @randomvariable @benmoss 
@andrewmyhre Naadir and Ben are planning on taking another look at this, sorry for the delay.

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -249,6 +249,17 @@ func (s *Service) getAPIServerClassicELBSpec() (*infrav1.ClassicELB, error) {
 	if err != nil {
 		return nil, err
 	}
+
+	securityGroupIDs := []string{}
+	controlPlaneLoadBalancer := s.scope.ControlPlaneLoadBalancer()
+	if controlPlaneLoadBalancer != nil && len(controlPlaneLoadBalancer.SecurityGroups) != 0 {
+		for _, sg := range controlPlaneLoadBalancer.SecurityGroups {
+			securityGroupIDs = append(securityGroupIDs, sg)
+		}
```
benmoss, line 258:
```suggestion
		securityGroupIDs = append(securityGroupIDs, controlPlaneLoadBalancer.SecurityGroups...)
```

vincepri:
@andrewmyhre do you have time to address the above comments?

api/v1alpha3/types.go
```diff
@@ -179,6 +179,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// SecurityGroupOverrides is an optional set of security groups to use for cluster instances
```
sethp-nr, line 183: Ditto documentation.

sethp-nr:
@andrewmyhre when I run `make generate` I see this in the output:

>> E0521 10:14:00.509236   54956 conversion.go:755] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.NetworkSpec -> sigs.k8s.io/cluster-api/api/v1alpha2.NetworkSpec
E0521 10:14:00.509411   54956 conversion.go:756]   the following fields need manual conversion:
E0521 10:14:00.509417   54956 conversion.go:758]       - SecurityGroupOverrides

I'm not sure exactly what "needing manual conversion" means in this context, but I infer that it has something to do with implementing that `Convert_v1alpha3_NetworkSpec_To_v1alpha2_NetworkSpec` function in the v1alpha2 package. Maybe something similar to [the serialization/deserialization dance that CAPI does to stuff some MachineSpec pieces into an annotation](https://github.com/kubernetes-sigs/cluster-api/blob/08e90608fcb2f5e7b41bbfc1eb3bde074ad01a00/api/v1alpha2/conversion.go#L112-L117)?

api/v1alpha2/awscluster_conversion.go
```diff
@@ -75,6 +75,9 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 		dst.Spec.NetworkSpec.VPC.AvailabilityZoneSelection = restored.Spec.NetworkSpec.VPC.AvailabilityZoneSelection
 	}
 
+	dst.Spec.NetworkSpec.SecurityGroupOverrides = restored.Spec.NetworkSpec.SecurityGroupOverrides
+	dst.Spec.NetworkSpec.TagSecurityGroups = restored.Spec.NetworkSpec.TagSecurityGroups
```
detiber, line 79: I think we might need some special handling for `Spec.ControlPlaneLoadBalancer.SecurityGroups`, it looks like we are currently only handling the case that `Spec.ControlPlaneLoadBalancer` is nil.

api/v1alpha3/types.go
```diff
@@ -194,6 +194,15 @@ type NetworkSpec struct {
 	// CNI configuration
 	// +optional
 	CNI *CNISpec `json:"cni,omitempty"`
+
+	// TagSecurityGroups optionally disables tagging the security groups used by the cluster
+	// +optional
+	TagSecurityGroups *bool `json:"tagSecurityGroups,omitempty"`
```
detiber, line 200: Should this be removed as unused now?

andrewmyhre:
Closing in favor of #1933 which is up to date with master

PR #1648, opened 2020-03-16
:warning: :bug:Add flag option for uncompressed userdata.

### LINKED ISSUE #1639: userdata should not be compressed
/kind bug

I wrote a bootstrap provider for coreos ignition. But I failed to create ec2 instance because user data is gzip compressed and ignition failed to read it.

It's bootstraper's responsibility to decide whether it should compress userdata or not 

**Environment:**

- Cluster-api-provider-aws version: v0.4.10
- Kubernetes version: (use `kubectl version`): 1.15.3
- OS (e.g. from `/etc/os-release`): container linux

MartinForReal:
Add flag option for uncompressed userdata.

Fixes #1639

### DIFF
--- a/api/v1alpha2/zz_generated.conversion.go
+++ b/api/v1alpha2/zz_generated.conversion.go
@@ -603,6 +603,7 @@ func autoConvert_v1alpha3_AWSMachineSpec_To_v1alpha2_AWSMachineSpec(in *v1alpha3
 	}
 	// WARNING: in.RootVolume requires manual conversion: does not exist in peer-type
 	out.NetworkInterfaces = *(*[]string)(unsafe.Pointer(&in.NetworkInterfaces))
+	// WARNING: in.UncompressedUserData requires manual conversion: does not exist in peer-type
 	// WARNING: in.CloudInit requires manual conversion: inconvertible types (sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3.CloudInit vs *sigs.k8s.io/cluster-api-provider-aws/api/v1alpha2.CloudInit)
 	return nil
 }
--- a/api/v1alpha3/awsmachine_types.go
+++ b/api/v1alpha3/awsmachine_types.go
@@ -94,6 +94,13 @@ type AWSMachineSpec struct {
 	// +kubebuilder:validation:MaxItems=2
 	NetworkInterfaces []string `json:"networkInterfaces,omitempty"`
 
+	// UncompressedUserData specify whether the user data is gzip-compressed before it is sent to ec2 instance.
+	// cloud-init has built-in support for gzip-compressed user data
+	// user data stored in aws secret manager is always gzip-compressed.
+	//
+	// +optional
+	UncompressedUserData *bool `json:"uncompressedUserData,omitempty"`
+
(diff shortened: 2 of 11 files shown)


MartinForReal:
flag option is not added to alphav2 because it will introduce a schema change in alphav2.

api/v1alpha2/zz_generated.conversion.go
```diff
@@ -603,6 +603,7 @@ func autoConvert_v1alpha3_AWSMachineSpec_To_v1alpha2_AWSMachineSpec(in *v1alpha3
 	}
 	// WARNING: in.RootVolume requires manual conversion: does not exist in peer-type
 	out.NetworkInterfaces = *(*[]string)(unsafe.Pointer(&in.NetworkInterfaces))
+	// WARNING: in.UncompressedUserData requires manual conversion: does not exist in peer-type
```
detiber: This will require a modification to `api/v1alpha2/awscluster_conversion.go` to attempt to restore the value from annotation if it is set, similar to how `Status.Bastion` is handled here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7240d095ea189681c6f92c79e14dd37853af2512/api/v1alpha2/awscluster_conversion.go#L65-L67

api/v1alpha3/awsmachine_types.go
```diff
@@ -94,6 +94,13 @@ type AWSMachineSpec struct {
 	// +kubebuilder:validation:MaxItems=2
 	NetworkInterfaces []string `json:"networkInterfaces,omitempty"`
 
+	// UncompressedUserData specify whether the user data is gzip-compressed before it is sent to ec2 instance.
+	// cloud-init has built-in support for gzip-compressed user data
+	// user data stored in aws secret manager is always gzip-compressed.
+	//
+	// +optional
+	UncompressedUserData *bool `json:"uncompressedUserData,omitempty"`
```
detiber: What are your thoughts on the long term solution for the naming of this variable? Do you think we should revert the logic when we are ready to make the next set of breaking API changes?

api/v1alpha2/zz_generated.conversion.go
MartinForReal: I added a unit test in awsmachine_conversion_test.go. 

https://github.com/minsheng-fintech-corp-ltd/cluster-api-provider-aws/blob/35ec7d2a453c10ab1ce5ddca85ccaa7744f4eb2e/api/v1alpha2/awsmachine_conversion_test.go#L154-L166

and added manual conversion in restoreAWSMachineSpec
 
https://github.com/minsheng-fintech-corp-ltd/cluster-api-provider-aws/blob/master/api/v1alpha2/awsmachine_conversion.go

Thanks for the info!

api/v1alpha3/awsmachine_types.go
MartinForReal: I think this works well for cloud-init based images.  Since most of the Linux distribution has support for cloud-init, I think there is no need to rever the logic.

MartinForReal:
@detiber I've changed the code accordingly. Any advice is appreciated. Thanks!

PR #1650, opened 2020-03-16
🐛 Fix docu

breyer:
aws ec2 import-key-pair needs base64 encoded material.
Needed to run without an error.


breyer:
Is unnecessary if you make sure that the awscli json outputs.

PR #1653, opened 2020-03-19
:sparkles: Loosen the filter used for image lookups

detiber:

Currently image lookups are pretty strict about format, this loosens the image lookup filter while still retaining backward compatiblity for existing images produced.

### DIFF
--- a/pkg/cloud/services/ec2/ami.go
+++ b/pkg/cloud/services/ec2/ami.go
@@ -40,15 +40,16 @@ const (
 	// The pattern is:
 	// 1. the string value `capa-ami-`
 	// 2. the baseOS of the AMI, for example: ubuntu-18.04, centos-7, amazon-2
-	// 3. the kubernetes version as defined by the packages produced by kubernetes/release, for example: 1.13.0-00, 1.12.5-01
-	// 4. the timestamp that the AMI was built
-	amiNameFormat = "capa-ami-%s-%s-??-??????????"
+	// 3. the kubernetes version as defined by the packages produced by kubernetes/release with or without v as a prefix, for example: 1.13.0, 1.12.5-mybuild.1, v1.17.3
+	// 4. a `-` followed by any additional characters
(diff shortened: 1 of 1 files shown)


detiber:
e2e test failure was a flake in 1 out of 13 tests

pkg/cloud/services/ec2/ami.go
```diff
@@ -40,15 +40,16 @@ const (
 	// The pattern is:
 	// 1. the string value `capa-ami-`
 	// 2. the baseOS of the AMI, for example: ubuntu-18.04, centos-7, amazon-2
-	// 3. the kubernetes version as defined by the packages produced by kubernetes/release, for example: 1.13.0-00, 1.12.5-01
-	// 4. the timestamp that the AMI was built
-	amiNameFormat = "capa-ami-%s-%s-??-??????????"
+	// 3. the kubernetes version as defined by the packages produced by kubernetes/release with or without v as a prefix, for example: 1.13.0, 1.12.5-mybuild.1, v1.17.3
+	// 4. a `-` followed by any additional characters
+	amiNameFormat = "capa-ami-?%s-?%s-*"
```
ncdc: I'm curious about the first `?` - what is that for? The second `?` is for the optional `v` in the version, right?

detiber: Fat finger mistake, thanks for catching :)

detiber: Fixed, thanks!

detiber:

want to test just long enough to verify that the lookups still work

pkg/cloud/services/ec2/ami.go
detiber: @ncdc multi-tasking has me scattered a bit, but the remaining `?` is indeed for the optional `v` in version

detiber:
At least one of the e2e tests has successfully passed, which means it successfully did the lookup for the existing image :)
/hold cancel

PR #1656, opened 2020-03-23
:running: elb: Skip additional control plane registration if node not yet registered

randomvariable:

Mirrors change made in CAPV: https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/pull/860


detiber:
I'm not sure this is needed due to the default behavior of the ELB when adding instances (defaulting to OutOfService rather than InService by default)

Issue #1657, opened 2020-03-24
Use naive TCP instead of SSL for ELB health checks


jayunit100:
/kind bug

**What steps did you take and what happened:**

I ran ELB's with apiserver arguments which include a small set of cipher's not in the AWS 2016 defaults

What happened --- the ELBs never send traffic to apiservers, which means kubelets never register their static pods, and scheduler/kcm never get to start doing anything - so you just have an orphan API Server talking to ETCD, and cluster is effectively dead.

**What did you expect to happen:**

ELBs to happily handshake with my apiserver and forward traffic along.

**Environment:**

- Cluster-api-provider-aws version: v1alpha3 (.4)
- Kubernetes version: (use `kubectl version`): 1.17 
- OS (e.g. from `/etc/os-release`): any

detiber:
/close
Closing, since this would cause log spamming related to the health checks. Longer term we want to explore migrating over to using an NLB for the load balancer, but there are other challenges that we need to solve related to handling in hairpinning before we can make that switch.

seh:
> Longer term we want to explore migrating over to using an NLB for the load balancer, but there are other challenges that we need to solve related to handling in hairpinning before we can make that switch.

@detiber, is there an issue tracking this potential move to using an NLB?

detiber:
@seh https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1484 is probably the closest issue we have for supporting NLBs

Ankitasw:
@sedefsavas just to clarify, we have not changed the scope of this issue right, we want classic ELB health check to use TCP instead of SSL, that's all?

PR #1658, opened 2020-03-24
Use TCP so that we can support different ciphers easily for v1alpha3.

jayunit100:
🐛 (:bug:, patch and bugfixes)

USes TCP rather then SSL, so that we can support any arbitrary APISErver Cipher.  Otherwise, we need to alignt the ELB Ciphers with those in the APIServer, Scheduler, KCM, Kubelet, and so on.

**What issue this fixes**

https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1657

(undos the work in #1658, which i think was valuable, but we can live without for the timebeing, correct me if im missing something)

### DIFF
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -255,7 +255,7 @@ func (s *Service) getAPIServerClassicELBSpec() (*infrav1.ClassicELB, error) {
 			},
 		},
 		HealthCheck: &infrav1.ClassicELBHealthCheck{
-			Target:             fmt.Sprintf("%v:%d", infrav1.ClassicELBProtocolSSL, 6443),
+			Target:             fmt.Sprintf("%v:%d", infrav1.ClassicELBProtocolTCP, 6443),
 			Interval:           10 * time.Second,
 			Timeout:            5 * time.Second,
 			HealthyThreshold:   5,


pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -255,7 +255,7 @@ func (s *Service) getAPIServerClassicELBSpec() (*infrav1.ClassicELB, error) {
 			},
 		},
 		HealthCheck: &infrav1.ClassicELBHealthCheck{
-			Target:             fmt.Sprintf("%v:%d", infrav1.ClassicELBProtocolSSL, 6443),
```
yastij: I’d rather update the listeners/health checks to use custom order/set of SSL ciphers

randomvariable: As far as ELB goes, listeners != health checks in this regards. It is TCP passthrough for the listener. The healthcheck spec just ensures we do a TLS handshake.

randomvariable: Verified that AWS only supports non-elliptic curve handshake tests for ELBv1.

yastij: sgtm

jayunit100:
Oops bug !
```
     Expected   <*v1.ObjectReference | 0x0> to not be  nil
```

detiber:
/close
Closing after discussing with @jayunit100 and @randomvariable

PR #1667, opened 2020-03-26
:sparkles: Update predicate handling

detiber:

- Improve logging for predicate handling
- Improve logging for objectMapper functions
- Add triggering reconciliation on Cluster.Status.InfrastructureReady becoming available

Fixes #

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -186,7 +186,7 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
 			&source.Kind{Type: &infrav1.AWSCluster{}},
 			&handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(r.AWSClusterToAWSMachines)},
 		).
-		WithEventFilter(pausePredicates).
+		WithEventFilter(pausedPredicates(r.Log)).
 		Build(r)
 
 	if err != nil {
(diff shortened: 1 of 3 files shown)


controllers/awsmachine_controller.go
```diff
@@ -202,13 +244,48 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+
+				switch {
+				// never return true for a paused Cluster
+				case newCluster.Spec.Paused:
+					log.V(4).Info("Cluster is paused, will not reconcile associated AWSMachine.")
+					return false
+				// return true if Cluster.Status.InfrastructureReady has changed from false to true
+				case !oldCluster.Status.InfrastructureReady && newCluster.Status.InfrastructureReady:
+					log.V(4).Info("Cluster InfrastructureReady became ready, will reconcile associated AWSMachine.")
+					return true
```
detiber: @ncdc I'm having trouble reproducing the delayed reconciliation on pivot, I've tried with the current version of this as well as replicating the same predicate logic from before with additional logging. That said, I am testing with the updates to controller-runtime and capi, but I wouldn't expect either of those to have a meaningful effect.

Even if it doesn't solve the delayed resync, I think it'll still be valuable to also trigger on InfrastructureReady changes here.

controllers/awsmachine_controller.go
```diff
@@ -186,7 +187,48 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
 			&source.Kind{Type: &infrav1.AWSCluster{}},
 			&handler.EnqueueRequestsFromMapFunc{ToRequests: handler.ToRequestsFunc(r.AWSClusterToAWSMachines)},
 		).
-		WithEventFilter(pausePredicates).
+		WithEventFilter(predicate.Funcs{
```
detiber: These common predicates are still identical (other than logging messages), planning on refactoring them back out to helpers.go again.

controllers/awsmachine_controller.go
ncdc: Yeah, I've had a lot of trouble too. I think I've managed to see the delay twice in total.

The trigger on infraReady sgtm

detiber:
@ncdc this should be ready for another review now

controllers/awsmachine_controller.go
```diff
@@ -202,13 +202,48 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+				switch {
+				// never return true for a paused Cluster
+				case newCluster.Spec.Paused:
+					log.V(4).Info("Cluster is paused, will not attempt to map associated AWSMachine.")
+					return false
+				// return true if Cluster.Status.InfrastructureReady has changed from false to true
+				case !oldCluster.Status.InfrastructureReady && newCluster.Status.InfrastructureReady:
+					log.V(4).Info("Cluster InfrastructureReady became ready, will attempt to map associated AWSMachine.")
+					return true
+				// return true if Cluster.Spec.Paused has changed from false to true
```
ncdc:
```suggestion
				// return true if Cluster.Spec.Paused has changed from true to false
```

detiber:
@ncdc resolved the comment issues and only issues with the e2e run are typical flakes

PR #1671, opened 2020-03-27
:running: Update cluster-api release used for e2e/conformance tests

detiber:
- Updates the CAPI version used for conformance and e2e tests


ncdc:
Do you want to wait on the e2e before merging?

detiber:
> Do you want to wait on the e2e before merging?

It looks like at least one of the previously failed tests has already succeeded (was failing on validation of KubeadmConfig.Spec.ClusterConfiguration.Networking.PodSubnet being required).

ncdc:
Ok, let's merge it. Will probably want to adjust to 0.3.3 soon after it's out too.
/lgtm

PR #1672, opened 2020-03-30
:sparkles: Update to latest cluster-api version

detiber:
- Updates cluster-api dependencies to v0.3.3
- Replaces custom isPaused function with sigs.k8s.io/cluster-api/util/HasPausedAnnotation


controllers/awscluster_controller.go
```diff
@@ -73,13 +73,13 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
...
-		log.Info("AWSCluster or linked Cluster is marked as paused. Won't reconcile")
+	if cluster == nil {
+		log.Info("Cluster Controller has not yet set OwnerRef")
 		return reconcile.Result{}, nil
 	}
 
-	if cluster == nil {
-		log.Info("Cluster Controller has not yet set OwnerRef")
+	if util.IsPaused(cluster, awsCluster) {
+		log.Info("AWSCluster or linked Cluster is marked as paused. Won't reconcile")
```
detiber: Swapped these to ensure that we aren't passing a nil cluster in to util.IsPaused

controllers/helpers.go
```diff
@@ -48,20 +47,10 @@ func pausedPredicates(logger logr.Logger) predicate.Funcs {
 func processIfUnpaused(logger logr.Logger, obj runtime.Object, meta metav1.Object) bool {
 	kind := strings.ToLower(obj.GetObjectKind().GroupVersionKind().Kind)
 	log := logger.WithValues("namespace", meta.GetNamespace(), kind, meta.GetName())
-	if isPaused(nil, meta) {
+	if clusterutil.HasPausedAnnotation(meta) {
```
randomvariable: shouldn't this be clusterutil.IsPaused(... ?

randomvariable: Ah, nevermind. It's not for the cluster object.

detiber: `clusterutil.IsPaused` does not handle a nil cluster, so this basically just checks the paused annotation which is what using a nil cluster with the local `isPaused` would fall back to.

ncdc:
/lgtm

/hold
waiting on e2e results

PR #1674, opened 2020-03-30
:book: add first draft of AWS multitenancy proposal (#1552)

### LINKED ISSUE #1552: AWS account multitenancy using a single instance of the controller
/kind feature

At present, CAPA relies on the default authentication providers within the AWS SDK.
The principle is described in the Java SDK at https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/credentials.html, with the Go SDK supporting the following chain:

1. If `AWS_WEB_IDENTITY_TOKEN_FILE` and `AWS_IAM_ROLE_ARN` env vars are both set, use refreshable role assumption via OIDC federation (v1alpha3 onwards).
2. Environment Variables (`AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, etc...)
3. Shared credentials file specified by `AWS_SHARED_CREDENTIALS_FILE` environment variable (defaulting to ~/.aws/config)
4. ECS refreshable credentials from the URL specified in the `AWS_CONTAINER_CREDENTIALS_FULL_URI` env var, using the env var `AWS_CONTAINER_AUTHORIZATION_TOKEN` as the authorisation header
5. ECS refreshable credentials from `http://169.254.170.2/${AWS_CONTAINER_CREDENTIALS_RELATIVE_URI}`
6. EC2 metadata

In addition, the [shared credentials file supports a wide variety of combinations](https://docs.aws.amazon.com/cli/latest/userguide/cli-configure-files.html).

This provides maximum flexibility in the provision of credentials. In practice, this is implemented in CAPA by not specifying any credential parameters.

However, this means that a separate instance of CAPA must exist for each account/role/user that CAPA is meant to manage a cluster under, which may affect the deployment of conversion & defaulting webhooks as well as placing a scalability limit on running CAPA.

If possible, CAPA could support multi-tenancy within a single instance whilst preserving flexibility in terms of the authentication methods on offer.

We can break the above authentication methods into the following categories:

1. Forms of static credential provisioning
     * Env vars and static credentials encoded in the shared configuration file
2. Dynamic retrieval of credentials from a URL
     * ECS & EC2 IMDS metadata
3. Role assumption
     * Faked EC2 IMDS metadata provided by a system like Kiam
     * OIDC federation
     * Shared configuration file specifying role assumption in a profile with a static or dynamic EC2 metadata source

2 is specific to processes running on EC2 hosts or on ECS, and is not relevant, leaving 1 & 3 as methods which should be supportable for multi-tenancy within a single process.

From a Kubernetes perspective, any supply of credentials should be secured - if it comes from an API server resource, that API server resource must have appropriate RBAC (use of secret references, restricted CRD etc...).

andrewmyhre:
**What this PR does / why we need it**: Enable CAPA to reconcile Kubernetes clusters in a multi-tenant/VPC sharing architecture

Partial resolution of #1552

### DIFF
--- a/docs/proposal/aws-multitenancy.md
+++ b/docs/proposal/aws-multitenancy.md
@@ -96,7 +96,7 @@ It follows that an organization may wish to provision control planes and worker 
 
 ## Proposal
 
-(Put diagram here)
+![Diagram](https://user-images.githubusercontent.com/208667/78036771-b299ad00-7338-11ea-90b2-1a88702fb1dd.png)
 
 The current implementation of CAPA requests a new AWS EC2 and ELB service per cluster and per machine and stores these in fields on the ClusterScope struct. ClusterScopes are reference values which are created per-reconciliation:
 
--- a/docs/proposal/aws-multitenancy.md
+++ b/docs/proposal/aws-multitenancy.md
@@ -1,0 +1,254 @@
+
+---
+title: AWS Account Multi-Tenancy
+authors:
+  - "@andrewmyhre"
+reviewers:
+  - ""
+creation-date: 2020-03-30
+last-updated: 2020-03-30
+status: provisional
+see-also:
+  - "[https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1552](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1552)"
+replaces:
+  - ""
+superseded-by:
+  - ""
+---
+
+# Federated AWS Account Support
+
+<!-- BEGIN Remove before PR -->
+To get started with this template:
+1. **Make a copy of this template.**
+  Copy this template into `docs/enhacements` and name it `YYYYMMDD-my-title.md`, where `YYYYMMDD` is the date the proposal was first drafted.
+1. **Fill out the required sections.**
+1. **Create a PR.**
+  Aim for single topic PRs to keep discussions focused.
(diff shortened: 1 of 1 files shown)


docs/proposal/aws-multitenancy.md
```diff
@@ -0,0 +1,254 @@
...
+    name: my-cluster
+```
+awscluster.yaml
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AWSCluster
+metadata:
+  name: my-cluster
+spec:
+  iamRoleArn: arn:aws:iam::123456789:role/CAPARole
```
randomvariable, line 171: Only adding IAM role ARN introduces a confused deputy risk, which is important for some of our users. Would need to support external ID as well.

randomvariable:
Just an update on this. I'm just going through some docs, and ensuring we can cover the various permutations. Will update tomorrow.

randomvariable:
Hi @andrewmyhre :

I've created a Google Doc with proposed changes here:
https://docs.google.com/document/d/1vwjvWc-RIZfwDXb-3tW4gKBOaPCKhripfhXshOmB5z4/edit# 

This covers additional use cases important for some of our users, especially around hiding confidential aspects of credentials and caching.

randomvariable:
Let's rally in #1713 if that's ok?
Closing for now.

PR #1675, opened 2020-03-30
Add documentation for multi-AZ control plane 📖

### LINKED ISSUE #930: Add documentation for how to create a control plane that spans multiple AZs

scottslowe:
This PR provides information on configuring CAPA to use multiple availability zones (AZs) so that KubeadmControlPlane will distribute control plane nodes across multiple AZs (for a workload cluster). This PR also provides some caveats of this design.

Signed-off-by: Scott Lowe <lowes@vmware.com>

This PR documents the use of the `networkSpec` object to configure CAPA to create subnets across multiple AZs within a region. Once CAPA creates subnets in multiple AZs, then control plane nodes created by the KubeadmControlPlane object will automatically be distributed across multiple AZs.

Fixes #930

### DIFF
--- a/docs/multi-az-control-plane.md
+++ b/docs/multi-az-control-plane.md
@@ -33,7 +33,7 @@ spec:
 
 Specifying the CIDR block alone for the VPC is not enough; users must also supply a list of subnets that provides the desired AZ, the CIDR for the subnet, and whether the subnet is public (has a route to an Internet gateway) or is private (does not have a route to an Internet gateway).
 
-Note that CAPA insists that there must be a public subnet (and associated Internet gateway) in order to place a control plane node in that AZ, even if the control plane load balancer is configured to be an internal load balancer. Therefore, for every AZ where a control plane node should be placed, the `networkSpec` object must define both a public and private subnet.
+Note that CAPA insists that there must be a public subnet (and associated Internet gateway), even if no public load balancer is requested for the control plane. Therefore, for every AZ where a control plane node should be placed, the `networkSpec` object must define both a public and private subnet.
 
 Once CAPA is provided with a `networkSpec` that spans multiple AZs, the KubeadmControlPlane controller will automatically distribute control plane nodes across multiple AZs. No further configuration from the user is required.
 
--- a/docs/multi-az-control-plane.md
+++ b/docs/multi-az-control-plane.md
@@ -1,0 +1,42 @@
+# Multi-AZ Control Plane
+
+## Overview
+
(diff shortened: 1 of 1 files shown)


docs/multi-az-control-plane.md
```diff
@@ -0,0 +1,42 @@
...
+    - availabilityZone: us-west-2c
+      cidrBlock: 10.50.64.0/20
+      isPublic: true
+    - availabilityZone: us-west-2c
+      cidrBlock: 10.50.80.0/20
+```
+
+Specifying the CIDR block alone for the VPC is not enough; users must also supply a list of subnets that provides the desired AZ, the CIDR for the subnet, and whether the subnet is public (has a route to an Internet gateway) or is private (does not have a route to an Internet gateway).
+
+Note that CAPA insists that there must be a public subnet (and associated Internet gateway) in order to place a control plane node in that AZ, even if the control plane load balancer is configured to be an internal load balancer. Therefore, for every AZ where a control plane node should be placed, the `networkSpec` object must define both a public and private subnet.
```
sethp-nr, line 36: We should probably just fix CAPA rather than having to document around this every time, but IIRC we check whether there _are_ public subnets, but then place the control plane machines in private subnets if the configured load balancer is `internal`.

sethp-nr:
LGTM overall Scott, just that one note. 

/assign

docs/multi-az-control-plane.md
randomvariable: We should create a new issue if we don't have this already, and is ok to mark as a caveat in the doc at the present time?.

sethp-nr: Good call on opening the issue: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1696

For now, I'm thinking something like this:

```suggestion
Note that CAPA insists that there must be a public subnet (and associated Internet gateway), even if no public load balancer is requested for the control plane. Therefore, for every AZ where a control plane node should be placed, the `networkSpec` object must define both a public and private subnet.
```

As I understand it, CAPA will always place the nodes in private subnets, but depending on the load balancer's scheme may look for a public subnet to associate with the ELB itself.

scottslowe: I've updated the wording as suggested in the latest commit (6ae70d0).

scottslowe:
@sethp-nr @randomvariable I've made the requested change. PTAL when you have time. Thanks!

sethp-nr:
Thanks, Scott! I really appreciate these docs, it's so good to write these things down. And it helps me check my assumptions about the system!

/lgtm

PR #1680, opened 2020-03-31
Do not rely on status to determine if we should delete a bastion that was previously created

### LINKED ISSUE #1678: Do not rely on status to determine if we should delete a bastion that was previously created
/kind bug

Currently we are relying on AWSCluster.Status.Bastion to determine if we should delete a previously created bastion instance. Instead we should query by tags to determine if we should delete.

More context in this PR comment thread: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1677#discussion_r401118877

/priority important-soon
/milestone v0.5.x
/help

evalsocket:
Signed-off-by: Yuvraj <yuvraj.yad001@gmail.com>

Fixes #1678

### DIFF
--- a/pkg/cloud/services/ec2/bastion.go
+++ b/pkg/cloud/services/ec2/bastion.go
@@ -42,10 +42,14 @@ func (s *Service) ReconcileBastion() error {
 	}
 
 	if !s.scope.AWSCluster.Spec.Bastion.Enabled {
-		if s.scope.AWSCluster.Status.Bastion != nil {
-			return s.DeleteBastion()
+		_, err := s.describeBastionInstance()
+		if err != nil {
+			if awserrors.IsNotFound(err) {
+				return nil
+			}
+			return err
 		}
-		return nil
+		return s.DeleteBastion()
 	}
 
 	s.scope.V(2).Info("Reconciling bastion host")
@@ -80,6 +84,7 @@ func (s *Service) ReconcileBastion() error {
 
 	s.scope.AWSCluster.Status.Bastion = instance.DeepCopy()
 	s.scope.V(2).Info("Reconcile bastion completed successfully")
+
 	return nil
 }
 


pkg/cloud/services/ec2/bastion.go
```diff
@@ -36,75 +36,78 @@ const (
...
+		if s.scope.VPC().IsUnmanaged(s.scope.Name()) {
+			s.scope.V(4).Info("Skipping bastion reconcile in unmanaged mode")
+			return nil
+		}
 
-	if !s.scope.AWSCluster.Spec.Bastion.Enabled {
-		if s.scope.AWSCluster.Status.Bastion != nil {
-			return s.DeleteBastion()
+		if !s.scope.AWSCluster.Spec.Bastion.Enabled {
+			if s.scope.AWSCluster.Status.Bastion != nil {
```
detiber, line 46: Rather than using the existing Status here, this should use the output of `s.describeBastionInstance()` to determine if we should delete a previously created bastion instance.

vincepri:
@evalsocket are you still interested in getting these changes merged in?

pkg/cloud/services/ec2/bastion.go
detiber, line 46:
```suggestion
			_ , err := s.describeBastionInstance()
			if err != nil {
				if awserrors.IsNotFound(err) {
					return nil
				}
				return err
			}
			return s.DeleteBastion()
```

pkg/cloud/services/ec2/bastion.go
```diff
@@ -36,75 +36,78 @@ const (
...
+				return s.DeleteBastion()
+			}
```
detiber, line 48:
```suggestion
			_ , err := s.describeBastionInstance()
			if err != nil {
				if awserrors.IsNotFound(err) {
					return nil
				}
				return err
			}
			return s.DeleteBastion()
```

randomvariable:
@evalsocket can you squash your commits?

vincepri:
Can you please retitle with the correct format?

evalsocket:
@vincepri It's done

PR #1686, opened 2020-04-03
:running: [ci-conformance] Additional conformance test fixes

detiber:
- Fix quote handling for ssh connection tunneled through Session Manager
- Make sure KUBECONFIG is not set prior to trying to fetch logs from kind cluster during cleanup

### DIFF
--- a/hack/ci/e2e-conformance.sh
+++ b/hack/ci/e2e-conformance.sh
@@ -91,11 +91,16 @@ function ssh-to-node() {
   local node="$1"
   local cmd="$2"
 
+  user=ubuntu
   ssh_key_pem="/tmp/${AWS_SSH_KEY_NAME}.pem"
-  ssh_params="-o LogLevel=quiet -o ConnectTimeout=30 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
-  ssh "$ssh_params" -i "$ssh_key_pem" \
-    -o "ProxyCommand sh -c \"aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'\"" \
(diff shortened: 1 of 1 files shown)


hack/ci/e2e-conformance.sh
```diff
@@ -91,11 +91,16 @@ function ssh-to-node() {
   local node="$1"
   local cmd="$2"
 
+  user=ubuntu
   ssh_key_pem="/tmp/${AWS_SSH_KEY_NAME}.pem"
-  ssh_params="-o LogLevel=quiet -o ConnectTimeout=30 -o UserKnownHostsFile=/dev/null -o StrictHostKeyChecking=no"
-  ssh "$ssh_params" -i "$ssh_key_pem" \
-    -o "ProxyCommand sh -c \"aws ssm start-session --target %h --document-name AWS-StartSSHSession --parameters 'portNumber=%p'\"" \
-    ubuntu@"${node}" "${cmd}"
```
detiber, line 98: This was leading to issues related to nested quotes

hack/ci/e2e-conformance.sh
```diff
@@ -116,6 +121,7 @@ cleanup() {
 
 # our exit handler (trap)
 exit-handler() {
+  unset KUBECONFIG
```
detiber, line 124: KUBECONFIG is being set during `run_tests`, so this is to ensure that we aren't attempting to connect to the workload cluster when we should be connecting to the kind cluster.

detiber:
/assign @ncdc 
This doesn't address the test failure directly, but it fixes issues with the artifacts we are generating to further troubleshoot the cause of failures.

ncdc:
👀

PR #1687, opened 2020-04-06
:running: [ci-conformance] Fix kustomize template to better handle envsubst

detiber:

- Update templates to better handle kustomize
- Add region command line arg to proxy command
- Fix handling of Kubernetes Version, CI Version, and using imageID in tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -38,6 +38,7 @@ TEST_E2E_DIR := test/e2e
 
 # Binaries.
 CLUSTERCTL := $(BIN_DIR)/clusterctl
+KUSTOMIZE := $(TOOLS_BIN_DIR)/kustomize
 CONTROLLER_GEN := $(TOOLS_BIN_DIR)/controller-gen
 ENVSUBST := $(TOOLS_BIN_DIR)/envsubst
 GOLANGCI_LINT := $(TOOLS_BIN_DIR)/golangci-lint
@@ -150,6 +151,9 @@ $(MOCKGEN): $(TOOLS_DIR)/go.mod # Build mockgen from tools folder.
 
 $(CONVERSION_GEN): $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/conversion-gen k8s.io/code-generator/cmd/conversion-gen
+
+$(KUSTOMIZE): $(TOOLS_DIR)/go.mod # Build kustomize from tools folder.
+	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/kustomize sigs.k8s.io/kustomize/kustomize/v3
 
 $(RELEASE_NOTES) : $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR) && go build -tags tools -o $(BIN_DIR)/release-notes sigs.k8s.io/cluster-api/hack/tools/release
@@ -276,7 +280,7 @@ release: clean-release  ## Builds and push container images using the latest git
 
 .PHONY: release-manifests
 release-manifests: $(RELEASE_DIR) ## Builds the manifests to publish with a release
-	kustomize build config > $(RELEASE_DIR)/infrastructure-components.yaml
+	$(KUSTOMIZE) build config > $(RELEASE_DIR)/infrastructure-components.yaml
 
 .PHONY: release-binaries
 release-binaries: ## Builds the binaries to publish with a release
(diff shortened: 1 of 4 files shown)


detiber:
Looks like it might be a flake waiting on cluster infrastructure, trying again:
/test pull-cluster-api-provider-aws-make-conformance

detiber:
uggh, all that work to get ssm working and the boskos-managed accounts do not have the necessary permissions :joy:

ncdc:
Oh no! 😢 

Can we get the permissions added?

CecileRobertMichon:
@detiber you can use `$$` to tell envsubst to not replace a var, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/templates/kustomizeversions.yaml

detiber:
Last test hit 2hr timeout, but was working, trying again:
/test pull-cluster-api-provider-aws-make-conformance

detiber:
> Can we get the permissions added?

Yes, I created https://github.com/kubernetes/test-infra/issues/17190 to track getting the account permissions updated.

detiber:
> @detiber you can use $$ to tell envsubst to not replace a var, see https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/master/templates/kustomizeversions.yaml

@CecileRobertMichon ah, nice I went ahead and copied what you are doing in CAPZ here as well, thanks!

detiber:
The last test actually passed conformance, but failed to complete the teardown prior to the 2 hour timeout. Created https://github.com/kubernetes/test-infra/pull/17191 to fix the job to have a 3 hour timeout.

Makefile
```diff
@@ -151,6 +152,9 @@ $(MOCKGEN): $(TOOLS_DIR)/go.mod # Build mockgen from tools folder.
 $(CONVERSION_GEN): $(TOOLS_DIR)/go.mod
 	cd $(TOOLS_DIR); go build -tags=tools -o $(BIN_DIR)/conversion-gen k8s.io/code-generator/cmd/conversion-gen
 
+$(KUSTOMIZE): $(TOOLS_DIR)/go.mod # Build kustomize from tools folder.
```
CecileRobertMichon, line 155: release-manifests also uses kustomize

detiber: indeed, thanks for catching that. Fixed.

detiber:
remaining conformance test failures are related to the job timeout (which is still running off the old job config). Closing/re-opening didn't help with getting the job to run with the new job config (for the increased 3 hr timeout).

Makefile
```diff
@@ -38,6 +38,7 @@ TEST_E2E_DIR := test/e2e
 
 # Binaries.
 CLUSTERCTL := $(BIN_DIR)/clusterctl
+KUSTOMIZE := $(abspath $(TOOLS_BIN_DIR)/kustomize)
```
ncdc, line 41: macOS doesn't have `abspath` out of the box. I only see us using `abspath` for `GINKGO`. I wonder if we can remove it, or, if we do need an absolute path, we can do `cd <path> && pwd`.

Makefile
gab-satchi, line 155: If we're building kustomize here, do we still need to ensure and install [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/scripts/ci-conformance.sh#L46)?

hack/ci/e2e-conformance.sh
```diff
@@ -281,14 +241,30 @@ create_cluster() {
   # actually create the cluster
   KIND_IS_UP=true
 
+  if [[ -z ${IMAGE_ID:-} ]]; then
+    # default lookup org hould be the same as defaultMachineAMIOwnerID
+    IMAGE_LOOKUP_ORG=${IMAGE_LOOKUP_ORG:-"258751437250"}
+    filter="capa-ami-ubuntu-18.04-1.17.*"
```
gab-satchi, line 247: Would this need to be updated when the minor version of k8s is updated?

Makefile
detiber: Good catch, we do not.

hack/ci/e2e-conformance.sh
detiber: Not necessarily. The only time this would need to be updated is if there are some packaging or dependency related changes that are no longer compatible with the newer packages we are overriding at bootstrapping time.

detiber:
@gab-satchi ptal, I believe this should be ready to go now.

gab-satchi:
changes lgtm. It seems the timeout was increased to 3 hours but why are the tests still hitting a 2 hour timeout?

detiber:
> changes lgtm. It seems the timeout was increased to 3 hours but why are the tests still hitting a 2 hour timeout?

Prow is a bit finicky on picking up new job configs for already open PRs. It will usually pick up net-new jobs, but getting it to pick up changes for already triggered PRs is fickle. Sometimes Opening/Closing the PR resolves the issue, but it didn't in this case.

PR #1688, opened 2020-04-08
:book: Update roadmap for delayed v1alpha4

detiber:

Updating roadmap to account for delaying any breaking API changes in favor of adding additional features without breaking changes.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -6,34 +6,22 @@ This roadmap is a constant work in progress, subject to frequent revision. Dates
 
 - Documentation improvements
 
-## v0.5 (v1alpha3) ~ March 2020
-
-- Updates for Cluster API v1alpha3
-- Add support for `clusterctl` v2
-- [Disable creation of bastion host by default](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/947)
-- [Ability to configure node disks](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1403)
-- [SSM integration (alternative to bastion host)](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1553)
-- [Encrypted user-data](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1387)
-- [Support for cross-zone load balancing](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1416)
-- [Private API-server load balancer](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/873)
-
-## v0.6 (v1alpha4) ~ July 2020
-
-- Consume Cluster API e2e testing framework
+## v0.5.x (v1alpha3+) - June/July 2020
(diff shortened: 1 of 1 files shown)


detiber:
Initial rough cut to start gathering feedback, will be tweaking as I go through the backlog and verify relative priorities with the community.

docs/roadmap.md
```diff
@@ -6,34 +6,22 @@ This roadmap is a constant work in progress, subject to frequent revision. Dates
...
 
-- ?
+- [Multiple topologies](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1484)
+- [Machine load balancer implementation](https://github.com/kubernetes-sigs/cluster-api/issues/1250)
+- SNS/SQS-based updates from CloudWatch Events so we don’t have to poll AWS for updates
 
 ## TBD
 
-- Implement MachinePools - Autoscaling groups and spot instances
+- Implement MachinePools - Autoscaling groups and instances
```
sethp-nr, line 24: Curious about this one – where would a MachinePool <-> Spot Fleet integration land on this roadmap, do you think?

sethp-nr: Oh, might it be the place where it's explicitly called out as "spot instances support"? 🤦

sethp-nr:
This looks great to me so far, thanks Jason!

docs/roadmap.md
detiber: Not sure, the current proposal for Spot instances discusses using individual spot instances. I also expect the initial MachinePool support to likely target ASGs directly.

I'm not sure if there would be a way that we could target both ASGs and Spot Fleet with the same MachinePool resource, or if we'd need a separate resource/implementation.

Happy to hear your thoughts on how to reconcile the support with the existing planned work.

randomvariable:
looks good

docs/roadmap.md
sethp-nr: Ah, maybe it's worth splitting Spot Fleet out explicitly then in the roadmap. My understanding was that the spot work deferred targeting Spot Fleet initially so as to not take a dependency on the `MachinePool` proposal, but that it was "in scope" once the MachinePool stuff was ready. 

Then again, I haven't been tracking the MachinePool proposal as closely as I should've, so perhaps it's not an API fit for Spot Fleet after all.

detiber: Happy to split it out. I think my main concern around Spot Fleet support at the moment is that I'm not sure the delta between what type of config we'd need to expose vs an ASG to make an informed decision about how easy it would be to share a type/implementation.

sethp-nr: Yeah, from the CAPA side they'd be entirely different entities, I think. ASGs have a confusing level of overlapping support around "fleet composition" and spot markets and such, but at the end of the day they still seem to speak in terms of a (possibly weighted) quantity of _instances_ whereas spot fleet's primary scaling vehicle is a quantity of _resources_.

detiber: Added a couple of additional items below to indicate spot instance/spot fleet support around MachinePools, hopefully that helps.

vincepri:
@detiber Is this good to go?

detiber:
@vincepri should be

PR #1694, opened 2020-04-20
✨ Customizable Image Lookup Format

### LINKED ISSUE #1644: Customizable image lookup format
/kind feature

As a Kubernetes distribution maintainer, I would like to be able to customize the image lookup format so that I can still use the image lookup mechanism even if my image names don't match the current hard-coded format.

We should ideally come up with a way to have this configured globally for a single CAPA deployment.

bagnaram:
Allow the user to specify a custom format string for AMI lookup based on OS & Kubernetes version inside the AWSMachine and AWSCluster

Fixes #1644

### DIFF
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -55,7 +55,7 @@ type AWSClusterSpec struct {
 	// ImageLookupFormat is the AMI naming format to look up machine images when
 	// a machine does not specify an AMI. When set, this will be used for all
 	// cluster machines unless a machine specifies a different ImageLookupOrg.
-	// Supports substitutions for ${BASE_OS} and ${K8S_VERSION}.
+	// Supports substitutions for $BASE_OS and $K8S_VERSION.
 	// +optional
 	ImageLookupFormat string `json:"imageLookupFormat,omitempty"`
 
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -55,11 +55,15 @@ type AWSClusterSpec struct {
 	// ImageLookupFormat is the AMI naming format to look up machine images when
 	// a machine does not specify an AMI. When set, this will be used for all
 	// cluster machines unless a machine specifies a different ImageLookupOrg.
-	// Supports substitutions for {{.BaseOS}} and {{.K8sVersion}} of the base OS
-	// of the AMI and the kubernetes version as defined by the packages produced
-	// by kubernetes/release with or without v as a prefix, for example: 1.13.0,
-	// 1.12.5-mybuild.1, v1.17.3 respectively. See golang templates
-	// https://golang.org/pkg/text/template/.
+	// Supports substitutions for {{.BaseOS}} and {{.K8sVersion}} with the base
+	// OS and kubernetes version, respectively. The BaseOS will be the value in
+	// ImageLookupBaseOS or ubuntu (the default), and the kubernetes version as
+	// defined by the packages produced by kubernetes/release without v as a
+	// prefix: 1.13.0, 1.12.5-mybuild.1, or 1.17.3. For example, the default
+	// image format of capa-ami-{{.BaseOS}}-?{{.K8sVersion}}-* will end up
+	// searching for AMIs that match the pattern capa-ami-ubuntu-?1.18.0-* for a
+	// Machine that is targeting kubernetes v1.18.0 and the ubuntu base OS. See
+	// also: https://golang.org/pkg/text/template/
 	// +optional
 	ImageLookupFormat string `json:"imageLookupFormat,omitempty"`
 
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -52,6 +52,13 @@ type AWSClusterSpec struct {
 	// +optional
 	ControlPlaneLoadBalancer *AWSLoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
 
+	// ImageLookupFormat is the AMI naming format to look up machine images when
+	// a machine does not specify an AMI. When set, this will be used for all
+	// cluster machines unless a machine specifies a different ImageLookupOrg.
+	// Supports substitutions for ${BASE_OS} and ${K8S_VERSION}.
+	// +optional
+	ImageLookupFormat string `json:"imageLookupFormat,omitempty"`
+
 	// ImageLookupOrg is the AWS Organization ID to look up machine images when a
 	// machine does not specify an AMI. When set, this will be used for all
 	// cluster machines unless a machine specifies a different ImageLookupOrg.
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -55,7 +55,11 @@ type AWSClusterSpec struct {
 	// ImageLookupFormat is the AMI naming format to look up machine images when
 	// a machine does not specify an AMI. When set, this will be used for all
 	// cluster machines unless a machine specifies a different ImageLookupOrg.
-	// Supports substitutions for $BASE_OS and $K8S_VERSION.
+	// Supports substitutions for {{.BaseOS}} and {{.K8sVersion}} of the base OS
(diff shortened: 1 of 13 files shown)


api/v1alpha3/awsmachine_types.go
```diff
@@ -36,6 +36,10 @@ type AWSMachineSpec struct {
 	// AMI is the reference to the AMI from which to create the machine instance.
 	AMI AWSResourceReference `json:"ami,omitempty"`
 
+	// ImageLookupFormat is the format string used for image lookup if AMI is
+	// not set. Supports substitutions for ${BASE_OS} and ${K8S_VERSION}.
```
detiber, line 40:
```suggestion
	// not set. Supports substitutions for ${BASE_OS} and ${K8S_VERSION}.
	// +optional
```

bagnaram: Thank you. Testing locally now...

detiber:
It looks like the verify failure is:
```
make[1]: Leaving directory '/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws'
diff --git a/api/v1alpha2/zz_generated.conversion.go b/api/v1alpha2/zz_generated.conversion.go
index a2195724..72f53e09 100644
--- a/api/v1alpha2/zz_generated.conversion.go
+++ b/api/v1alpha2/zz_generated.conversion.go
@@ -446,6 +446,7 @@ func autoConvert_v1alpha3_AWSClusterSpec_To_v1alpha2_AWSClusterSpec(in *v1alpha3
 	} else {
 		out.ControlPlaneLoadBalancer = nil
 	}
+	// WARNING: in.ImageLookupFormat requires manual conversion: does not exist in peer-type
 	// WARNING: in.ImageLookupOrg requires manual conversion: does not exist in peer-type
 	// WARNING: in.ImageLookupBaseOS requires manual conversion: does not exist in peer-type
 	// WARNING: in.Bastion requires manual conversion: does not exist in peer-type
@@ -589,6 +590,7 @@ func autoConvert_v1alpha3_AWSMachineSpec_To_v1alpha2_AWSMachineSpec(in *v1alpha3
 	if err := Convert_v1alpha3_AWSResourceReference_To_v1alpha2_AWSResourceReference(&in.AMI, &out.AMI, s); err != nil {
 		return err
 	}
+	// WARNING: in.ImageLookupFormat requires manual conversion: does not exist in peer-type
 	out.ImageLookupOrg = in.ImageLookupOrg
 	// WARNING: in.ImageLookupBaseOS requires manual conversion: does not exist in peer-type
 	out.InstanceType = in.InstanceType
generated files are out of date, run make generate
make: *** [Makefile:402: verify-gen] Error 1
```

This also reminds me that we should likely have code in `api/v1alpha2/aws{cluster,machine}_conversion.go` to handle backing up/restoring this field as part of conversion to support round-tripping of the resource.

bagnaram:
Thank you for the reminder on conversions. Good lesson for me on API conversions. Will do.

api/v1alpha3/awscluster_types.go
```diff
@@ -52,6 +52,13 @@ type AWSClusterSpec struct {
 	// +optional
 	ControlPlaneLoadBalancer *AWSLoadBalancerSpec `json:"controlPlaneLoadBalancer,omitempty"`
 
+	// ImageLookupFormat is the AMI naming format to look up machine images when
+	// a machine does not specify an AMI. When set, this will be used for all
+	// cluster machines unless a machine specifies a different ImageLookupOrg.
+	// Supports substitutions for $BASE_OS and $K8S_VERSION.
```
sethp-nr, line 58:
```suggestion
	// Supports substitutions for ${BASE_OS} and ${K8S_VERSION}.
```

api/v1alpha3/awsmachine_types.go
```diff
@@ -36,6 +36,11 @@ type AWSMachineSpec struct {
 	// AMI is the reference to the AMI from which to create the machine instance.
 	AMI AWSResourceReference `json:"ami,omitempty"`
 
+	// ImageLookupFormat is the format string used for image lookup if AMI is
+	// not set. Supports substitutions for $BASE_OS and $K8S_VERSION.
```
sethp-nr, line 40:
```suggestion
	// not set. Supports substitutions for ${BASE_OS} and ${K8S_VERSION}.
```

bagnaram: I originally substituted the variables in this way, however it causes the `clusterawsadm` installation to bomb, because it expects an environment variable to be passed in to the manifest. This is only a token for string substitution.

sethp-nr: These two comments are that the actual string substitution doesn't match the docs. I think if a user supplies `imageLookupFormat: '$BASE_OS-$K8S_VERSION'` they'll have a much worse time than if they'd said `imageLookupFormat: '${BASE_OS}-${K8S_VERSION}'`

bagnaram: Perhaps I need to use a different type of substitution? '{{ BASE_OS }}-{{ K8S_VERSION }}'? That way the docs can match without overhauling the token parser in `clusterctl`

sethp-nr: Oh, I see what you mean now – yeah, I think a different token would be preferable then anyway. Maybe consider a simple https://golang.org/pkg/text/template/ ?

bagnaram: Thank you for the suggestion. Using golang templates.

bagnaram:
I have ensured to run `make verify-gen` and that I am including the latest copy of `api/v1alpha2/zz_generated.conversion.go` however the the job pull-cluster-api-provider-aws-verify isn't seeing it for some reason?

bagnaram:
It appears that `conversion-gen` outputs the `zz_generated.conversions.go` to `$GOPATH/src/api/v1alpha2/zz_generated.conversion.go` instead of `api/v1alpha2` which was causing the generated file to never update or commit as it should. Manual copying of the file seems to fix the problem

detiber:
> It appears that conversion-gen outputs the zz_generated.conversions.go to $GOPATH/src/api/v1alpha2/zz_generated.conversion.go instead of api/v1alpha2 which was causing the generated file to never update or commit as it should. Manual copying of the file seems to fix the problem

Doh, yeah conversion-gen has a bit of weird edge case behavior around GOPATH handling (set vs unset) and whether or not the repo is checked out under the GOPATH.

api/v1alpha3/awsmachine_types.go
```diff
@@ -36,6 +36,17 @@ type AWSMachineSpec struct {
 	// AMI is the reference to the AMI from which to create the machine instance.
 	AMI AWSResourceReference `json:"ami,omitempty"`
 
+	// ImageLookupFormat is the AMI naming format to look up machine images when
+	// a machine does not specify an AMI. When set, this will be used for all
+	// cluster machines unless a machine specifies a different ImageLookupOrg.
```
sethp-nr, line 41: This looks like copy-pasta from the cluster type to me.

```suggestion
	// ImageLookupFormat is the AMI naming format to look up the image for this machine
	// It will be ignored if an explicit AMI is set.
```

bagnaram: Good idea. It makes more sense what is going on with the vars. Both items have been changed

detiber:
/approve
e2e test made it past the point where a workload cluster is bootstrapped, so AMI lookups against existing AMIs still works.

/assign @sethp-nr 
for lgtm

sethp-nr:
Hmm, it does look like an AWS-side error of some kind – the instance is never assigned an internal DNS name. Do you know a way to get visibility into the AWS errors for the e2e runs?

detiber:
> Hmm, it does look like an AWS-side error of some kind – the instance is never assigned an internal DNS name. Do you know a way to get visibility into the AWS errors for the e2e runs?

There are controller logs available within the Artifacts for the jobs, but we're currently missing per-instance logs from the individual instances due to lack of required IAM permissions needed to initiate a Session Manager session currently (https://github.com/kubernetes/test-infra/issues/17190).

sethp-nr:
Apologies for letting this drop, @bagnaram . 

/lgtm

PR #1707, opened 2020-05-01
WIP: ✨ Etcd EBS Volumes

### LINKED ISSUE #1698: Consider moving etcd to it's own volume
/kind feature

Currently by default we only provision a single storage volume for instances. We should likely investigate creating a dedicated volume for etcd storage and verifying that the default configuration used is one that ensures both adequate and consistent performance characteristics.

bagnaram:
Preliminary support for attaching etcd EBS volumes
Fixes #1698


bagnaram:
This is preliminary support for etcd on additional EBS volumes. Currently, this PR requires manual modification of the `KubeadmControlPlane` resource to mitigate a kubeadm bug:
```
apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
kind: KubeadmControlPlane
spec:
  kubeadmConfigSpec:
    preKubeadmCommands:
    - if [ -d "/var/lib/etcd/lost+found" ]; then rm -Rf /var/lib/etcd/lost+found; fi
```

Also it is worth noting that the way of programmatically creating the EBS volume and attaching is a bit brittle, as it relies on hijacking the UserData creation by the AWSMachineController

vincepri:
@bagnaram There is a proposal that would allow cloud-init to get disks setup https://github.com/kubernetes-sigs/cluster-api/issues/2994

bagnaram:
I am re-working the PR to include the changes made in https://github.com/kubernetes-sigs/cluster-api/compare/master...CecileRobertMichon:etcd-mount to include an updated `cloudinit` with file-system params

bagnaram:
Now requires https://github.com/kubernetes-sigs/cluster-api/compare/master...CecileRobertMichon:etcd-mount

The UserData is provided by the bootstrap controller, therefore it is necessary to identify the device type that is being attached via EBS. This is hardcoded in the `KubeadmConfigTemplate` and the `KubeadmControlPlane` manifest. The UX needs to be determined for automating this process between the bootstrap provider and the infrastructure because currently there isn't a way to pass this type of information. Possible options include `nvme-tools` to detect inside the instance.

CecileRobertMichon:
opened https://github.com/kubernetes-sigs/cluster-api/pull/3066

Issue #1705, opened 2020-04-29
Reconcile ELB attached machines when a Machine is being deleted


vincepri:
**Describe the solution you'd like**
The `AWSMachine` controller [already watches](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/controllers/awsmachine_controller.go#L185-L188) for notifications coming from Cluster API `Machine` objects.

This provider has an ELB that is reconciled [during normal operations](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/controllers/awsmachine_controller.go#L460-L462), although the `reconcileLBAttachment` never removes a Machine from the load balancer, instead it only ever register new ones to it as shown here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/32eefe16ddea72d4727e0db4ec38e68902bd99bc/controllers/awsmachine_controller.go#L561-L573

We could call `reconcileLBAttachment` during both `reconcileNormal` and `reconcileDelete`. It should have a check that if either `AWSMachine` has a `DeletionTimestamp`, we should deregister the instance from the ELB.

/priority important-soon
/kind feature

fabriziopandini:
/assign

@vincepri @detiber it seems that in order to get this fixed a new IAM permission is required

controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile LB attachment: AccessDenied: User: arn:aws:iam::xxxx:user/bootstrapper.cluster-api-provider-aws.sigs.k8s.io is not authorized to perform: elasticloadbalancing:DeregisterInstancesFromLoadBalancer on resource: arn:aws:elasticloadbalancing:us-west-xxxx:loadbalancer/test-apiserver\n\tstatus code: 403, request id: e31c92ce-1bad-4461-aaf0-e4e83e03ece3\ncould not deregister control plane instance \"i-0796396af2ae985c1\" from load balancer\nsigs.k8s.io/cluster-api-provider-aws/controllers.

any suggestion to get this sorted out?

fabriziopandini:
Ohh... it seems this is already there
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/04921198635ba0881299da26dd28789cd36bc080/pkg/cloud/services/cloudformation/bootstrap.go#L336
Might be my cloud formation is old. re-creating

fabriziopandini:
The grant is there but the bind to resources not 😞

Issue #1689, opened 2020-04-08
Instances should be removed from LB when they're being shut down


benmoss:
/kind bug

**What steps did you take and what happened:**
While doing a control plane upgrade workflow I was continuously doing `kubectl get nodes` on the new cluster. I saw that I got periodic timeouts/failures and I'm pretty sure it's because the instance is being included in the cluster load balancer before it's ready to handle requests. You also can see in the KCP logs these kinds of errors as well: 

```
[capi-kubeadm-control-plane-controller-manager-596d455d75-vz552 manager] E0408 21:11:25.275888      98 controller.go:258] controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile the remote kubelet RBAC role: failed to determine if kubelet config rbac role \"kubeadm:kubelet-config-1.17\" already exists: Get \"https://bmo-apiserver-1404960449.us-east-1.elb.amazonaws.com:6443/apis/rbac.authorization.k8s.io/v1/namespaces/kube-system/roles/kubeadm:kubelet-config-1.17?timeout=30s\": http2: server sent GOAWAY and closed the connection; LastStreamID=9369, ErrCode=NO_ERROR, debug=\"\""  "controller"="kubeadmcontrolplane" "request"={"Namespace":"default","Name":"bmo-control-plane"}
[capi-kubeadm-control-plane-controller-manager-596d455d75-vz552 manager] W0408 21:11:44.858497      98 http.go:392] Error reading backend response: unexpected EOF
```

**What did you expect to happen:**
It should stay out of the pool until it's ready to serve

**Environment:**

- Cluster-api-provider-aws version: e2c994b
- Kubernetes version: (use `kubectl version`): 1.17.4

benmoss:
I think this might just be actually an inevitable jitter as a result of there being timeouts / intervals involved in the LB healthcheck. I think it most commonly happens as a control plane node is being removed from the cluster, it is first removed from etcd which causes it to become unresponsive and the load balancer isn't able to react fast enough

The API server has an option `--shutdown-delay-duration` that should help us here, but until https://github.com/kubernetes/kubernetes/pull/88911 was merged it was buggy and the `/readyz` endpoint didn't start returning non-200 codes until *after* the shutdown delay elapsed. Something to look at in the future maybe.

benmoss:
Note that that would also require us to actually shut down the API server, right I think we just shutdown the machine

detiber:
[...]

PR #1710, opened 2020-05-04
🏃Reconcile ELB attached machines when a Machine is being deleted

fabriziopandini:
This PR makes CAPA remove control plane machine attached to ELB as soon as the machine is deleted or detected in some not running states, so the ELB immediately stops to forward requests without waiting for the health check timeouts

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1705

```release-note
⚠️ Action Required
The IAM policy for the controllers requires new permissions:

elasticloadbalancing:DeregisterInstancesFromLoadBalancer
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -281,6 +281,14 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 
 	machineScope.V(3).Info("EC2 instance found matching deleted AWSMachine", "instance-id", instance.ID)
 
+	if err := r.reconcileLBAttachment(machineScope, clusterScope, instance); err != nil {
+		// We are tolerating AccessDenied error, so this won't block for users with older version of IAM;
+		// all the other errors are blocking.
+		if !elb.IsAccessDenied(err) {
+			return ctrl.Result{}, errors.Errorf("failed to reconcile LB attachment: %+v", err)
+		}
+	}
+
 	// Check the instance state. If it's already shutting down or terminated,
 	// do nothing. Otherwise attempt to delete it.
 	// This decision is based on the ec2-instance-lifecycle graph at
@@ -451,15 +459,15 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		if err != nil {
 			return ctrl.Result{}, errors.Errorf("failed to ensure tags: %+v", err)
 		}
+
+		if err := r.reconcileLBAttachment(machineScope, clusterScope, instance); err != nil {
+			return ctrl.Result{}, errors.Errorf("failed to reconcile LB attachment: %+v", err)
+		}
 	}
 
 	// tasks that can only take place during operational instance states
 	if machineScope.InstanceIsOperational() {
 		machineScope.SetAddresses(instance.Addresses)
-
-		if err := r.reconcileLBAttachment(machineScope, clusterScope, instance); err != nil {
-			return ctrl.Result{}, errors.Errorf("failed to reconcile LB attachment: %+v", err)
-		}
 
 		existingSecurityGroups, err := ec2svc.GetInstanceSecurityGroups(*machineScope.GetInstanceID())
 		if err != nil {
@@ -564,11 +572,27 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 	}
 
 	elbsvc := elb.NewService(clusterScope)
+
+	// In order to prevent sending request to a "not-ready" control plane machines, it is required to remove the machine
+	// from the ELB as soon as the machine gets deleted or when the machine is in a not running state.
+	if !machineScope.AWSMachine.DeletionTimestamp.IsZero() || !machineScope.InstanceIsRunning() {
+		if err := elbsvc.DeregisterInstanceFromAPIServerELB(i); err != nil {
+			r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedDetachControlPlaneELB",
+				"Failed to deregister control plane instance %q from load balancer: %v", i.ID, err)
+			return errors.Wrapf(err, "could not deregister control plane instance %q from load balancer", i.ID)
+		}
+		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeNormal, "SuccessfulDetachControlPlaneELB",
+			"Control plane instance %q is de-registered from load balancer", i.ID)
+		return nil
+	}
+
 	if err := elbsvc.RegisterInstanceWithAPIServerELB(i); err != nil {
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedAttachControlPlaneELB",
 			"Failed to register control plane instance %q with load balancer: %v", i.ID, err)
 		return errors.Wrapf(err, "could not register control plane instance %q with load balancer", i.ID)
 	}
+	r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeNormal, "SuccessfulAttachControlPlaneELB",
+		"Control plane instance %q is registered with load balancer", i.ID)
 	return nil
 }
 
--- a/pkg/cloud/services/cloudformation/bootstrap.go
+++ b/pkg/cloud/services/cloudformation/bootstrap.go
@@ -204,6 +204,7 @@ func controllersPolicy(accountID, partition string) *iam.PolicyDocument {
 					"elasticloadbalancing:DescribeTags",
 					"elasticloadbalancing:ModifyLoadBalancerAttributes",
 					"elasticloadbalancing:RegisterInstancesWithLoadBalancer",
+					"elasticloadbalancing:DeregisterInstancesFromLoadBalancer",
 					"elasticloadbalancing:RemoveTags",
 				},
 			},
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -201,11 +201,23 @@ func (s *Service) RegisterInstanceWithAPIServerELB(i *infrav1.Instance) error {
 	}
 
 	_, err = s.scope.ELB.RegisterInstancesWithLoadBalancer(input)
+	return err
+}
+
+// DeregisterInstanceFromAPIServerELB de-registers an instance from a classic ELB
+func (s *Service) DeregisterInstanceFromAPIServerELB(i *infrav1.Instance) error {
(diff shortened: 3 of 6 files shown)


pkg/cloud/services/cloudformation/bootstrap.go
```diff
@@ -204,6 +204,7 @@ func controllersPolicy(accountID, partition string) *iam.PolicyDocument {
 					"elasticloadbalancing:DescribeTags",
 					"elasticloadbalancing:ModifyLoadBalancerAttributes",
 					"elasticloadbalancing:RegisterInstancesWithLoadBalancer",
+					"elasticloadbalancing:DeregisterInstancesFromLoadBalancer",
```
ncdc, line 207: IAM change will need a release note

controllers/awsmachine_controller.go
```diff
@@ -564,6 +568,20 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 	}
 
 	elbsvc := elb.NewService(clusterScope)
+
+	// In order o prevent sending request to a "not-ready" control plane machines, it is required to remove the machine
```
vincepri, line 572:
```suggestion
	// In order to prevent sending request to a "not-ready" control plane machines, it is required to remove the machine
```

controllers/awsmachine_controller.go
```diff
@@ -564,6 +568,20 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
...
+	// from the ELB as soon as the machine gets delete or when the machine is detected or not running
+	// Nb. this is a reinforcement to the ELB health check, which blocks traffic only after a timeout is reached.
```
vincepri, line 574: I'd remove this last bit, it's more of a separate use case. I see health check to be useful when a machine goes unhealthy for reasons unrelated to intended operations on our objects. The points above explain the reasoning very well. 

Basically, I don't want to suggest there is something wrong with the current health checks, because those are meant to cover a different set of use cases

controllers/awsmachine_controller.go
```diff
@@ -564,6 +568,20 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
...
+	if !machineScope.AWSMachine.DeletionTimestamp.IsZero() || !machineScope.InstanceIsRunning() {
+		if err := elbsvc.DeregisterInstanceFromAPIServerELB(i); err != nil {
+			r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedDetachControlPlaneELB",
+				"Failed to deregister control plane instance %q from load balancer: %v", i.ID, err)
+			return errors.Wrapf(err, "could not deregister control plane instance %q from load balancer", i.ID)
+		}
+		return nil
+	}
+
+	// reconcile normal
```
vincepri, line 584:
```suggestion
```

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -208,6 +208,26 @@ func (s *Service) RegisterInstanceWithAPIServerELB(i *infrav1.Instance) error {
...
+		Instances:        []*elb.Instance{{InstanceId: aws.String(i.ID)}},
+		LoadBalancerName: aws.String(name),
+	}
+
+	_, err = s.scope.ELB.DeregisterInstancesFromLoadBalancer(input)
+	if err != nil {
+		return err
+	}
+
+	return nil
```
vincepri, line 228:
```suggestion
	return err
```

pkg/cloud/services/cloudformation/bootstrap.go
ncdc: cc @ashoksekar07

benmoss:
I think this also fixes #1689

controllers/awsmachine_controller.go
```diff
@@ -564,6 +568,20 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 	}
 
 	elbsvc := elb.NewService(clusterScope)
+
+	// In order o prevent sending request to a "not-ready" control plane machines, it is required to remove the machine
+	// from the ELB as soon as the machine gets delete or when the machine is detected or not running
+	// Nb. this is a reinforcement to the ELB health check, which blocks traffic only after a timeout is reached.
+	if !machineScope.AWSMachine.DeletionTimestamp.IsZero() || !machineScope.InstanceIsRunning() {
```
detiber, line 575: Does this also need to filter out "Pending" as well, otherwise it looks like we might potentially de-register an instance from the ELB prior to it becoming ready on initial creation.

controllers/awsmachine_controller.go
```diff
@@ -564,6 +568,20 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
...
+		if err := elbsvc.DeregisterInstanceFromAPIServerELB(i); err != nil {
+			r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedDetachControlPlaneELB",
+				"Failed to deregister control plane instance %q from load balancer: %v", i.ID, err)
+			return errors.Wrapf(err, "could not deregister control plane instance %q from load balancer", i.ID)
+		}
```
detiber, line 580: Might be good to add an event in the successful case here as well.

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -208,6 +208,26 @@ func (s *Service) RegisterInstanceWithAPIServerELB(i *infrav1.Instance) error {
...
+		return err
+	}
+
+	input := &elb.DeregisterInstancesFromLoadBalancerInput{
+		Instances:        []*elb.Instance{{InstanceId: aws.String(i.ID)}},
+		LoadBalancerName: aws.String(name),
+	}
+
+	_, err = s.scope.ELB.DeregisterInstancesFromLoadBalancer(input)
+	if err != nil {
```
detiber, line 224: This should probably differentiate a "NotFound" error from other errors, since this is one of the first things called in `reconcileDelete` and should be re-entrant.

pkg/cloud/services/cloudformation/bootstrap.go
vincepri: It was pointed out in today's community meeting to allow for a graceful fallback in case the permission is denied, which I think it's a valid point given that this is mostly a nice to have

vincepri:
@fabriziopandini It was pointed out by @ncdc  in today's community meeting to allow for a graceful fallback in case the permission is denied, which I think it's a valid point given that this is mostly a nice to have. Might be worthwhile to inspect the SDK error and not return an error if we can't de-register an instance

controllers/awsmachine_controller.go
fabriziopandini: InstanceIsRunning already includes AWS machine states "Pending" and "Running"

controllers/awsmachine_controller.go
fabriziopandini: Currently there is no success event for register, so I did not add success event for de-register. Should I add a success event in both cases?

pkg/cloud/services/elb/loadbalancer.go
fabriziopandini: The API is not throwing error in case we are registering an instance that is already de-registered, so I think we are already ok.

fabriziopandini:
@vincepri @ncdc @detiber thanks for feedback!
Comment are addressed, test passing, so I'm removing WIP

controllers/awsmachine_controller.go
```diff
@@ -473,6 +477,10 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		}
 	}
 
+	if err := r.reconcileLBAttachment(machineScope, clusterScope, instance); err != nil {
```
ncdc, line 480: The way @randomvariable recently restructured this above is that we have two sections of operations: those applicable to all known states, and those only applicable to operational states. This moves LB attachment reconciliation outside of and after those two sections. Is that appropriate? Would it still work if we moved it inside the "all known states" section, or would that be too early?

controllers/awsmachine_controller.go
ncdc: A success event would be nice, but we can do it in a follow-up PR if you don't want to include it here.

controllers/awsmachine_controller.go
vincepri: Yeah we talked about it, if the Machine is being stopped it's still an operational state, but the ELB should remove the Machine until it gets into a running state again

randomvariable: The reasoning makes sense to me.

ncdc: If we move it into the "all known states" section, wouldn't that cover it (unless it's too early i.e. we need the security groups ensured first)?

controllers/awsmachine_controller.go
randomvariable: Is elbsvc.DeregisterInstanceFromAPIServerELB(i) allowed for stopping or stopped instances?
If so, that's fine.
Otherwise might need to filter those out given this is now not called within the operational states.

controllers/awsmachine_controller.go
randomvariable: Rereading, I think it will depend on https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1710/files/82e0a992239c3f2ccba1038ea2d52c585c054a19#r420397258

ncdc: Which file/line? (It's not easy to tell what the link is pointing to.)

randomvariable: I meant the comment above, for line 575 of `controllers/awsmachine_controller.go`:

> Is elbsvc.DeregisterInstanceFromAPIServerELB(i) allowed for stopping or stopped instances?
> If so, that's fine.
> Otherwise might need to filter those out given this is now not called within the operational states.

fabriziopandini: I moved the code under "all known states"

controllers/awsmachine_controller.go
fabriziopandini: event added

controllers/awsmachine_controller.go
fabriziopandini: This is not clear from the API documentation, but from my tests it is allowed to call elbsvc.DeregisterInstanceFromAPIServerELB(i) for stopping or stopped instances

fabriziopandini:
All the comments should be addressed

vincepri:
LGTM pending squash

fabriziopandini:
@vincepri squashed

Issue #1701, opened 2020-04-27
Move e2e tests to clusterctl framework


vincepri:
With the introduction of the new clusterctl test framework, we should draw a plan of action and publish a gap-analysis of the current e2e tests CAPA has today, and how we can port them to the new framework.

/kind cleanup
/assign @gab-satchi 
/cc @fabriziopandini @detiber

fabriziopandini:
Happy to help if required!
FYI https://github.com/kubernetes-sigs/cluster-api/pull/2943 has an example of E2E config file for AWS

vincepri:
/milestone v0.5.4

/lifecycle active
for @gab-satchi

gab-satchi:
- [x] Gap analysis - [gdoc](https://docs.google.com/spreadsheets/d/1PD6qFljm0G46h4OQKIii2plQEUhBJ9xfm40l6T7Uixw/edit#gid=0)
- [x] Port CAPA e2e tests over to new e2e flow (1 to start)- PR #1716 
- [ ]  CAPA CI script and CI config
- [ ]  test-infra prow job for new e2e tests
- [ ]  Move all e2e tests to use new framework
- [ ]  Remove old CAPA e2e package
- [ ]  Switch prow to use new job as main e2e

PR #1716, opened 2020-05-06
🏃 Refactor e2e tests to use clusterctl framework from CAPI

gab-satchi:
This is a first pass at moving one of the tests over to using the clusterctl e2e framework. 
The changes here are for a development environment. I will be making another PR for CI. In order to test it out locally will need to do the following:
- From the CAPA repository root: `make docker-build REGISTRY=gcr.io/k8s-staging-cluster-api`
- `export AWS_REGION=<aws_region>`
- `export AWS_SSH_KEY_NAME=<aws_ssh_key>`
- `export AWS_B64ENCODED_CREDENTIALS=$(clusterawsadm alpha bootstrap encode-aws-credentials)`
- From CAPA root: `make test-e2e-new`

Package name and make targets have `new` added to them. That's temporary while we support both the old e2e flow and this one. Once all the tests are switched over and the old flow is cleaned up, we can remove the 'new' from everything. A lot of the changes are similar to the [new CAPI e2e tests](https://github.com/kubernetes-sigs/cluster-api/blob/master/test/e2e/quick_start.go)

#1701


test/e2e_new/aws_test.go
```diff
@@ -0,0 +1,80 @@
...
+	. "github.com/onsi/gomega"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/utils/pointer"
+	"path/filepath"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/cluster-api/test/framework/clusterctl"
+	"sigs.k8s.io/cluster-api/util"
+)
+
+var _ = Describe("functional tests", func() {
```
fabriziopandini, line 34: @gab-satchi if the goad of this test is simply to spin un a workload cluster, you can consider re-using the test spec from [test/e2e](https://github.com/kubernetes-sigs/cluster-api/blob/master/test/e2e/quick_start.go). You can find an example of how the entire test spec can be re-used [here](https://github.com/kubernetes-sigs/cluster-api/blob/c3ef9222bb61524f5d3705ca1bf1202f3afeb720/test/e2e/quick_start_test.go#L27-L39).

Also other specs are available: self-hosting, upgrades, machine health checks...

test/e2e_new/common.go
```diff
@@ -0,0 +1,70 @@
...
+	"fmt"
+	. "github.com/onsi/ginkgo"
+	corev1 "k8s.io/api/core/v1"
+	"path/filepath"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/cluster-api/test/framework"
+	"sigs.k8s.io/cluster-api/util"
+)
+
+func setupSpecNamespace(ctx context.Context, specName string, clusterProxy framework.ClusterProxy, artifactFolder string) (*corev1.Namespace, context.CancelFunc) {
```
fabriziopandini, line 32: Wondering if we should add those methods to the framework, it seems those are the same of what we have in CAPI...

test/e2e_new/e2e_suite_test.go
```diff
@@ -0,0 +1,195 @@
...
+
+	By("Tearing down the management cluster")
+	if !skipCleanup {
+		tearDown(bootstrapClusterProvider, bootstrapClusterProxy)
+	}
+})
+
+func initScheme() *runtime.Scheme {
+	sc := runtime.NewScheme()
+	framework.TryAddDefaultSchemes(sc)
```
fabriziopandini, line 139: I think that it will be useful for you adding CAPA types to the scheme

test/e2e_new/e2e_suite_test.go
```diff
@@ -0,0 +1,195 @@
...
+	flag.BoolVar(&skipCleanup, "skip-cleanup", false, "if true, the resource cleanup after tests will be skipped")
+}
+
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "capa e2e")
+}
+
+var _ = SynchronizedBeforeSuite(func() []byte {
+	Expect(configPath).To(BeAnExistingFile(), "Invalid test suite argument. configPath should be an existing file.")
```
fabriziopandini, line 83: Same consideration here about the opportunity for centralizing some code...
As far as I can see there are no changes from what exists in CAPI...

test/e2e_new/aws_test.go
gab-satchi: The thought of reusing quickstart crossed my mind for this along with another test in CAPI but I dismissed it because I couldn't configure any of the cluster parameters such as name in this case. One way to configure these parameters would be for e2e_config to expose a setter to change variables once the configuration has been loaded. This will make it easier for multiple similar but slightly different 'quick-start' tests to reuse the spec without breaking the simplicity of the quick-start spec. WDYT?

test/e2e_new/common.go
gab-satchi: I certainly agree. setupNamespace can become a helper method. dumpSpecResourcesAndCleanup does a couple of things so unsure if that should be a single helper. I understand the helper methods in the framework as helpers to interact with the cluster (getting, setting and waiting for state)

test/e2e_new/e2e_suite_test.go
gab-satchi: If CAPA only used specs from the CAPI e2e then this suite wouldn't be necessary. Some of the other tests in CAPA do things a bit more AWS specific which would require a dedicated test suite. I'm now wondering if those tests should be ported over to use the framework or just continue to exist in the current form.

test/e2e_new/aws_test.go
gab-satchi: Name is a poor example for this case but ability to modify other template variables could be useful

fabriziopandini: @gab-satchi if you can point out what you want to configure, I will be happy to follow up.
WRT to template variables, you can drive them via config file or via is variables

gab-satchi: For the AWS tests in question, name is the only one that needs to be configurable during cluster creation. The rest are all about modifying resources afterwards which can be done with helpers. I think I'm missing something conceptually here WRT reusing the existing capi specs. 

My understanding of how CAPA can use the framework for a test like "testing load balancer service type": 
- Use framework to setup the workload cluster
- Create a service of type LoadBalancer (can be done with update to helpers)
- Use framework to delete workload cluster
- Assert on infra resources (LB) getting deleted

My understanding with using the existing capi spec: quick-start will create a cluster, wait for it to be created then delete the cluster immediately after. I'm not sure how CAPA can interject in the middle to make any assertions. Leaving the cleanup of the cluster to the caller is an option.

gab-satchi: Resolved offline. CAPI e2e specs with AWS can be used as a separate prow job if necessary. Changes to CAPA e2e can use helpers in the framework and not the full specs as CAPA would need some intermediate AWS calls.

test/e2e_new/common.go
gab-satchi: Will revisit what should be exported as common functions once we have more providers (CAPV) using the framework.

fabriziopandini:
@gab-satchi looks great!
/lgtm

detiber:
/assign
reviewing now

test/e2e_new/go.mod
```diff
@@ -0,0 +1,17 @@
+module sigs.k8s.io/cluster-api-provider-aws/test/e2e_new
```
detiber, line 1: Is there a need currently for a separate go module, the main reason the existing e2e test has one is to avoid leaking the sonobuoy dependency (and any potential conflicts it brings) into the main go module.

test/e2e_new/data/infrastructure-aws/cluster-template.yaml
```diff
@@ -0,0 +1,113 @@
+---
```
detiber, line 1: Is there a need for a separate template here? it might be nice if we could re-use the existing template instead of having to maintain a separate template for the e2e tests.

test/e2e_new/data/cni/calico.yaml
```diff
@@ -0,0 +1,792 @@
+---
```
detiber, line 1: How are we currently handling this in capi? I would like to see if we can avoid having a copied version of the calico manifest here if at all possible.

test/e2e_new/e2e_suite_test.go
```diff
@@ -0,0 +1,197 @@
...
+	flag.BoolVar(&useExistingCluster, "use-existing-cluster", false, "if true, the test uses the current cluster instead of creating a new one (default discovery rules apply)")
+	flag.BoolVar(&skipCleanup, "skip-cleanup", false, "if true, the resource cleanup after tests will be skipped")
+}
+
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "capa e2e")
+}
+
+var _ = SynchronizedBeforeSuite(func() []byte {
```
detiber, line 83: Any reason to not run `clusterawsadm` here to configure the AWS pre-requisites? Otherwise we are likely to fail if they are not already present and/or fully up to date.

test/e2e_new/config/aws-dev.yaml
```diff
@@ -0,0 +1,104 @@
...
+# export AWS_B64ENCODED_CREDENTIALS=
+
+images:
+  # Use local dev images built source tree;
+  - name: gcr.io/k8s-staging-cluster-api/cluster-api-controller-amd64:dev
+    loadBehavior: mustLoad
+  - name: gcr.io/k8s-staging-cluster-api/kubeadm-bootstrap-controller-amd64:dev
+    loadBehavior: mustLoad
+  - name: gcr.io/k8s-staging-cluster-api/kubeadm-control-plane-controller-amd64:dev
+    loadBehavior: mustLoad
```
detiber, line 25: Do we actually want to load the images here from the latest upstream, or should we test against a known good release?

test/e2e_new/config/aws-dev.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+# For creating local dev images built from the source tree;
+# - from the CAPI repository root, `make docker-build REGISTRY=gcr.io/k8s-staging-cluster-api` to build the cluster-api, bootstrap kubeadm, control-plane kubeadm provider images.
+# - from the CAPA repository root, `make docker-build REGISTRY=gcr.io/k8s-staging-cluster-api` to build the aws provider image.
+
+# Following env variables should be set before running the provider
+
+# export AWS_REGION=
+# export AWS_SSH_KEY_NAME=
+# export AWS_B64ENCODED_CREDENTIALS=
```
detiber, line 16: Can we create AWS_B64ENCODED_CREDENTIALS internally using `clusterawsadm` and just rely on the aws env vars for the test instead? That will translate much easier to the prow environment and remove a manual step for developers.

test/e2e_new/config/aws-dev.yaml
```diff
@@ -0,0 +1,104 @@
...
+    type: ControlPlaneProvider
+    versions:
+      - name: v0.3.0
+        # Use manifest from source files
+        value: ../../../../cluster-api/controlplane/kubeadm/config
+        replacements:
+          - old: "imagePullPolicy: Always"
+            new: "imagePullPolicy: IfNotPresent"
+          - old: "--enable-leader-election"
+            new: "--enable-leader-election=false"
```
detiber, line 71: Does the e2e test framework allow for the same defaulting to the latest "released" version that `clusterctl` does? If so, then we could rely on that here to avoid requiring most developers from having to keep their source trees in sync to test this provider.

randomvariable: I think the value here should be along the lines of
`https://github.com/kubernetes-sigs/cluster-api//config?ref=master`, as we have in CAPV https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/master/test/e2e/e2e.conf

test/e2e_new/config/aws-dev.yaml
randomvariable: will need to be refactored on clusterawsadm v2, which i have done a fair amount for #1734 . I could fix this up post-merge?

test/e2e_new/go.mod
gab-satchi: while adding more of the tests, I had to use CAPI@master. That's the only reason I could see and that's just temporary until a new release of capi is released and more of the test framework is available

test/e2e_new/e2e_suite_test.go
gab-satchi: will resolve with clusterawsadm v2

test/e2e_new/config/aws-dev.yaml
gab-satchi: latest upstream SGTM along with using the latest released version for providers. That will enable devs to get started with the tests with minimal setup.

test/e2e_new/data/cni/calico.yaml
gab-satchi: CAPI is also using a duplicate copy. Using URLs would be cleaner. Created [CAPI 3117](https://github.com/kubernetes-sigs/cluster-api/issues/3117)

vincepri:
cc @fabriziopandini @sedefsavas

test/e2e_new/config/aws-dev.yaml
gab-satchi: Updated to use the latest released versions

test/e2e_new/e2e_suite_test.go
```diff
@@ -0,0 +1,202 @@
...
+func init() {
+	flag.StringVar(&configPath, "config-path", "", "path to the e2e config file")
+	flag.StringVar(&artifactFolder, "artifacts-folder", "", "folder where e2e test artifact should be stored")
+	flag.BoolVar(&useExistingCluster, "use-existing-cluster", false, "if true, the test uses the current cluster instead of creating a new one (default discovery rules apply)")
+	flag.BoolVar(&skipCleanup, "skip-cleanup", false, "if true, the resource cleanup after tests will be skipped")
+}
+
+func TestE2E(t *testing.T) {
+	RegisterFailHandler(Fail)
+	RunSpecs(t, "capa e2e")
```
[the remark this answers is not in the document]
gab-satchi: thanks. updated

test/e2e_new/config/aws-dev.yaml
```diff
@@ -0,0 +1,107 @@
...
+        # Use manifest from source files
+        value: ../../../config
+        replacements:
+          - old: "imagePullPolicy: Always"
+            new: "imagePullPolicy: IfNotPresent"
+          - old: "--enable-leader-election"
+            new: "--enable-leader-election=false"
+    files:
+      # Add a cluster template
+      - sourcePath: "../../../templates/cluster-template.yaml"
```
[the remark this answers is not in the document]
gab-satchi: We opted to use the cluster-template in the repo itself rather than duplicating it. I've removed the duplicate template now.

test/e2e_new/aws_test.go
```diff
@@ -0,0 +1,191 @@
...
+				Protocol: corev1.ProtocolTCP,
+			},
+		},
+		Selector: map[string]string{
+			"app": "nginx",
+		},
+	}
+	createService(svcName, svcNamespace, nil, svcSpec, k8sclient)
+	//this sleep is required for the service to get updated with ingress details
+	time.Sleep(15 * time.Second)
```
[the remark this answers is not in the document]
gab-satchi: It shouldn't add flakiness. It's taken from the existing [tests](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/test/e2e/aws_test.go#L875-L877)

PR #1713, opened 2020-05-06
:book: Single controller multitenancy proposal v3

randomvariable:
**What this PR does / why we need it**:
Succeeds #1674

Updated multi-tenancy proposal after discussion on 2020-05-01. Principals have been moved to being cluster scoped, and follow the model of GatewayClass within https://github.com/kubernetes-sigs/service-apis . Seems to be the best compromise given current Kubernetes RBAC.

CC:
  - "@andrewmyhre"
  - "@ashoksekar07"
  - "@bagnaram"
  - "@CecileRobertMichon"
  - "@detiber"
  - "@devigned"
  - "@nader-ziada"
  - "@ncdc"
  - "@richardcase"
  - "@rudoi"
  - @sethp-nr 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

Additionally relabel all the old proposals with dates.

PR #1717, opened 2020-05-07
🐛Create Session with Options

bagnaram:
Previously, a new session with a new config gets created every time. This overrides any other options the user may provide using the go-aws-sdk. This fix creates a session using the ambient config rather. 

Fixes # N/A

### DIFF
--- a/pkg/cloud/scope/session.go
+++ b/pkg/cloud/scope/session.go
@@ -33,7 +33,13 @@ func sessionForRegion(region string) (*session.Session, error) {
 		return s.(*session.Session), nil
 	}
 
-	ns, err := session.NewSession(aws.NewConfig().WithRegion(region))
+	ns, err := session.NewSessionWithOptions(session.Options{
+		// Provide SDK Config options, such as Region.
+		Config: aws.Config{
+			Region: aws.String(region),
+		},
+	})
+
 	if err != nil {
 		return nil, err
 	}


bagnaram:
This will be superseded by #1713 but will fix this desired functionality in the meantime

randomvariable:
Could you squash the commits, and then we'll get the tests running?

pkg/cloud/scope/session.go
```diff
@@ -33,7 +33,13 @@ func sessionForRegion(region string) (*session.Session, error) {
 		return s.(*session.Session), nil
 	}
 
-	ns, err := session.NewSession(aws.NewConfig().WithRegion(region))
+	ns, err := session.NewSessionWithOptions(session.Options{
```
randomvariable, line 36: Just noticed, are we specifically missing the addition of
`SharedConfigState: session.SharedConfigEnable` ?

bagnaram: I was planning to rely on user specifying `AWS_SDK_LOAD_CONFIG`  if they need that changed from the default. https://docs.aws.amazon.com/sdk-for-go/v1/developer-guide/sessions.html#sessions-with-shared-config

randomvariable: Understood

randomvariable:
Ah, my weird combination of permissions on this repo strike again.

ncdc:
I'm having trouble seeing how this change is different from what we currently have. Could you please help educate me? Thanks!

bagnaram:
> I'm having trouble seeing how this change is different from what we currently have. Could you please help educate me? Thanks!

`aws.NewConfig` creates a blank configuration object each time instead overriding any ambient settings. The change switches to using an ambient config while still passing the region like before.

ncdc:
Both the current code and the code in this PR create new config objects though?

bagnaram:
> Both the current code and the code in this PR create new config objects though?

The key difference is the behavior. Without the modification, any configuration related options such as `AWS_DEFAULT_PROFILE`, etc get overwritten/ignored because a fresh config is generated each time. The intended behavior is that the ambient configuration would also be read

ncdc:
/hold

Zoomed with Matt. This appears to be working without any changes needed.

randomvariable:
Was under the impression that this PR was coming in because the env var was not being respected without going through that particular code path. If this is actually not the case, then don't see a reason to proceed.

bagnaram:
A couple of months ago this was a problem, and it appears to be fixed on master. Likely because of either pkg updates or go-aws-sdk fixed the bug itself.

ncdc:
/close

Closing based on previous comments. If you encounter problems again, let's hop on Zoom and we can see what's up.

Issue #1323, opened 2019-11-04
only create one NatGateWay in a vpc, instead of one NatGW in every public subnet


Kevin-fqh:
/kind feature

**Describe the solution you'd like**

we are charged for creating and using a NAT gateway in our account , and it takes too long to create/delete a NAT gateway.

so can we only create one NatGateWay in a vpc, instead of one NatGW in every public subnet ?

Correspondingly, all private subnet 's RouteTable , will has a route lead to the only one NatGateWay. 

The only one NatGateWay is shared in a vpc, but not only one AvailableZone.

**Environment:**

- Cluster-api-provider-aws version:  master

detiber:
@Kevin-fqh I think this is one of those scenarios that present challenges either way. If there is only a single NAT gateway, then all NAT traffic will be subject to cross-AZ data transfer costs.

@randomvariable do you have thoughts here?

randomvariable:
Believe we should return to the original feature list and solve the topology support issue cohesively. My fear is we're getting a bit too fragmented by adding things in a piecemeal fashion. Happy to come up with a proposal for the next meeting, which is likely to be 2019/12/2 because of kubecon.

Kevin-fqh:
follow `https://docs.aws.amazon.com/zh_cn/eks/latest/userguide/getting-started-eksctl.html`， we use eksctl to create an eks,  it has three public and there private subnet, but only has a single NATGateWay in a vpc as above.

```bash
[root@142node ~]#  eksctl create cluster --name prod --version 1.14 --nodegroup-name standard-workers --node-type t3.medium --nodes 2 --nodes-min 1 --nodes-max 3 --node-ami auto
[ℹ]  eksctl version 0.8.0
[ℹ]  using region ap-southeast-1
[ℹ]  setting availability zones to [ap-southeast-1b ap-southeast-1a ap-southeast-1c]
[ℹ]  subnets for ap-southeast-1b - public:192.168.0.0/19 private:192.168.96.0/19
[ℹ]  subnets for ap-southeast-1a - public:192.168.32.0/19 private:192.168.128.0/19
[ℹ]  subnets for ap-southeast-1c - public:192.168.64.0/19 private:192.168.160.0/19
[ℹ]  nodegroup "standard-workers" will use "ami-03a2cce9abe958c6c" [AmazonLinux2/1.14]
[ℹ]  using Kubernetes version 1.14
[...]

PR #1721, opened 2020-05-12
✨ Changes to the default subnet logic

### LINKED ISSUE #1643: Support multi-AZ with default AWSCluster configuration
/kind feature

When a user creates an AWSCluster with the default configuration, we should determine the number of AvailabilityZones available in the chosen region and create a public/private Subnet pair for each available AZ.

This might require modifications to the way that we are currently handling NAT gateways ala #1323 , since us-east-1 has 6 availability zones and will exhaust the default elastic IP quota.

In order to facilitate this, we will also need to add some subnetting logic to the controller to equitably split the vpc cidrBlock. Public subnets should get a much smaller address space than private subnets, possibly a /31, /30, or /29 to give 2, 4, or 8 available addresses for public subnets. Where private subnets should be fitted to the largest cidr block possible for the remaining address space in the vpc cidrBlock

This would greatly increase UX around deploying HA clusters.

richardcase:
Changes the logic around subnets.

Default subnet strategy changed so it will create 1 public and 1 private subnet for each AZ in the region (up to a maximum of 3 AZs by default). It splits the VPC CIDR block along the lines that @benmoss suggested in the issue. The VPC CIDR is split into numAZs+1 subnets. The first subnet is further sub-divided into numAZs and used for the public subnets and the remaining original subnets are used for the private subnets. 

If a region has more than 3 AZs then the first 3 will be selected based on alphabetical order. This change also introduces 2 configuration options to allow you to customize this behavior and specify the maximum number of AZs to use  and also whether to select AZs based on alphabetical order or randomly.

> Note: by changing the number of subnets created by default it will cause more nat gateways to be created by default. See #1323

For a "managed" VPC:

- If there are no subnets and no existing subnets then it will create them using the default strategy above. 
- If you specify subnets to use in the spec there must be at least 1 private and 1 public subnet

For a "unmanaged" VPC:

- You must specify subnets in the spec
- The subnets in the spec must exist already in the VPC
- There must be at least 1 private and 1 public subnet specified
- This is very much a case of "bring-your-own-subnets"

> Note on upgrades - if you have a cluster that was created using the previous subnet stratgey (i.e. 1 public & 1 private subnet) this will update the the `name` tag of the existing subnets to use the new naming which includes the az in the name. For example, an existing subnet named 
> `capi-managed-test-subnet-private` will have its name tag updated to `capi-managed-test-subnet-private-euwest2a`.

Fixes #1643


randomvariable:
Will this affect already deployed clusters in any way on upgrade?

richardcase:
> Will this affect already deployed clusters in any way on upgrade?

@randomvariable - a very good question. Thats not something i have tested. I will test that and let you know.

richardcase:
@randomvariable - i tested the effect this has on an already deployed cluster. The only effect i could see is that the name of the existing subnets are changed to use the naming convention introduced in this change (i.e. it includes the az in the name)

pkg/cloud/services/ec2/subnets.go
```diff
@@ -342,6 +399,8 @@ func (s *Service) getSubnetTagParams(id string, public bool, manualTags infrav1.
 	name.WriteString(s.scope.Name())
 	name.WriteString("-subnet-")
 	name.WriteString(role)
+	name.WriteString("-")
+	name.WriteString(strings.Replace(zone, "-", "", -1))
```
sethp-nr, line 403: I'm curious about this one – do we bump up against length limits, otherwise?

pkg/cloud/services/ec2/subnets.go
```diff
@@ -54,36 +54,49 @@ func (s *Service) reconcileSubnets() error {
...
-	// If the subnets are empty, populate the slice with the default configuration.
-	// Adds a single private and public subnet in the first available zone.
-	if len(existing) < 2 && len(subnets) < 2 {
-		zones, err := s.getAvailableZones()
-		if err != nil {
-			return err
+	// If we have an unmanaged VPC then we expect the following to be true:
+	// 1) Subnets must be specified
+	// 2) The subnets specified must already exist
+	// Otherwise this is an error
```
sethp-nr, line 60: This seems like a pretty significant change to me – previously CAPA would "discover" the subnets attached to a VPC and whether they were private/public. With this change, it looks to me like it still will, as long as the spec is "seeded" with a single public or private subnet (see https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1721/files#diff-b636b6a430a5b4f76ee6d31bc48bf46aR112-R116 ).

I'm not sure I follow how this change is important for EKS support, would you mind explaining?

pkg/cloud/services/ec2/subnets.go
```diff
@@ -137,15 +150,59 @@ LoopExisting:
 			if err != nil {
 				return err
 			}
-
-			nsn.DeepCopyInto(subnet)
+			subnet.ID = nsn.ID
```
sethp-nr, line 153: Is there a reason to discard the extra metadata about the new subnet until the next reconcile?

pkg/cloud/services/ec2/subnets.go
richardcase: I haven't hit any limits, it was really just to reduce the the number of -.
Looking at the docs the limit on the value is 256 characters
https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html

pkg/cloud/services/ec2/subnets.go
richardcase: This part of the change isn't important to EKS support. This part of the change was more around making the 'unmanaged vpc' so that you need to explicitly create and specify your subnets....properly bring your own subnets.

The important part for the EKS support is that the default subnetting strategy for a managed vpc is multi-az. Although its not just for EKS, it feels better that the default subnet strategy is multi-az by default across the board.

An additional check will need to be added in this area in the future for unmanaged vpc's to check that there are subnets in 2 azs but i haven't added this yet as its only a requirment for EKS.

pkg/cloud/services/ec2/subnets.go
richardcase: The only thing we need from this is the ID...`subnet` already has AvailabilityZone, CidrBlock, IsPublic. Although, i could've misread the code.

pkg/cloud/services/ec2/subnets.go
sethp-nr: Oh, in that case I have a mild preference for leaving the `-`s in, to be consistent with what we do elsewhere in the failureDomains and network.availabilityZones portions of the AWSCluster status.

pkg/cloud/services/ec2/subnets.go
sethp-nr: Oh, I see. I think this belongs much further down in the function, then, after CAPA has discovered the existing subnets and populated metadata (critically: whether they're public or private). We're going to overwrite what the user provided anyway with what we discovered, so it would be better to validate against the observed state than what they specified.

Also, would you mind [adding an event to the AWSCluster object](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/6999f6dd5d41f63e7b783d3676bd6b484a31e391/pkg/cloud/services/ec2/gateways.go#L123) so we have a description of the problem when someone does a `k describe awscluster FOO`?

pkg/cloud/services/ec2/subnets.go
sethp-nr: This is how we're populating those extra fields: by using the generated `DeepCopyInto` to copy it from the `nsn` (which gets thrown away) to the `subnet` which, per the defer at the top of the function, gets saved in the AWS cluster's status.

pkg/cloud/services/ec2/subnets.go
richardcase: Thanks @sethp-nr , i had missed that part about updating the subnets in the spec with the details of the existing subnets. As you pointed out, critical for the public/private test. I have updated this and also refactored the code further. And i've also added events as requested.

pkg/cloud/services/ec2/subnets.go
richardcase: The only thing we need to update in `subnet` from `nsn` is the ID: 
image
But as suggested i have changed it back to use the `DeepCopyInto`.

pkg/cloud/services/ec2/subnets.go
richardcase: Changed so the `-` are left in on the subnet and route table names.

pkg/cloud/services/ec2/subnets.go
sethp-nr: Oh, my mistake – I assumed createSubnet was populating those other fields, but it looks like the controller was already waiting for the next reconcile to "re-discover" the tags that it just set. So it goes.

sethp-nr:
Changes LGTM – thanks for adding the extra visibility! It looks like we were already passing in the subnet IDs (despite them being ignored, prior), but if anyone wasn't that will hopefully help them identify why their cluster provisioning stopped working.

/assign @detiber for final review

pkg/cloud/services/ec2/routetables.go
```diff
@@ -317,6 +317,8 @@ func (s *Service) getRouteTableTagParams(id string, public bool) infrav1.BuildPa
 	} else {
 		name.WriteString("private")
 	}
+	name.WriteString("-")
+	name.WriteString(zone)
```
vincepri, line 321: Do we need to worry about tag maximum length?

docs/existing-aws-infrastructure.md
```diff
@@ -38,10 +38,11 @@ spec:
   networkSpec:
     vpc:
       id: vpc-0425c335226437144
+    subnets:
+    - id: subnet-0261219d564bb0dc5
+    - id: subnet-0fdcccba78668e013
```
vincepri, line 43: These are optional though right?

pkg/cloud/services/ec2/subnets.go
```diff
@@ -54,80 +54,66 @@ func (s *Service) reconcileSubnets() error {
...
-				CidrBlock:        defaultPrivateSubnetCidr,
-				AvailabilityZone: zones[0],
-				IsPublic:         false,
-			})
+	if len(subnets) == 0 {
+		if unmanagedVPC {
+			// If we have a unmanaged VPC then subnets must be specified
+			errMsg := "no subnets specified, you must specify the subnets when using an umanaged vpc"
+			record.Warnf(s.scope.AWSCluster, "FailedNoSubnets", errMsg)
+			return errors.New(errMsg)
```
vincepri, line 64: Isn't this a breaking behavioral change?

pkg/cloud/services/ec2/subnets.go
```diff
@@ -146,6 +131,51 @@ LoopExisting:
...
+	}
+
+	//TODO: should this be configurable or are we ok with 3 AZs max for the default?
+	if len(zones) > 3 {
+		s.scope.V(2).Info("region has more than 3 availability zones, picking 3 at random", "region", s.scope.Region())
+		rand.Shuffle(len(zones), func(i, j int) {
+			zones[i], zones[j] = zones[j], zones[i]
+		})
+		zones = append(zones, zones[0], zones[1], zones[2])
+	}
```
vincepri, line 147: We should probably either expose this as an option, or sort it and pick the first N.

pkg/cloud/services/ec2/subnets.go
```diff
@@ -146,6 +131,51 @@ LoopExisting:
...
+	subnetCIDRs, err := cidr.SplitIntoSubnetsIPv4(s.scope.VPC().CidrBlock, numSubnets)
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed splitting VPC CIDR %s into subnets", s.scope.VPC().CidrBlock)
+	}
+
+	publicSubnetCIDRs, err := cidr.SplitIntoSubnetsIPv4(subnetCIDRs[0].String(), len(zones))
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed splitting CIDR %s into public subnets", subnetCIDRs[0].String())
+	}
+	privateSubnetCIDRs := append(subnetCIDRs[:0], subnetCIDRs[1:]...)
```
vincepri, line 160: Should we validate that the CidrBlock has at least a minimum number of IP addresses to avoid splitting it too much and have very few IPs per subnet.

What's the default CIDR here for the VPC? Can we also document it somewhere?

pkg/cloud/services/ec2/routetables.go
richardcase: The maximum length for the value is 256 Unicode characters (https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html).

I could add a check and fail if it exceeds this length?

docs/existing-aws-infrastructure.md
richardcase: Not any more with this change if you have an unmanaged vpc.  If you have an umanaged vpc then you need to supply the subnets to use instead of it trying to determine / use all subnets in the vpc.....so it becomes bring your own subnets and explicitly tell capa about them.

pkg/cloud/services/ec2/subnets.go
richardcase: Yes it is

pkg/cloud/services/ec2/subnets.go
richardcase: The reason i didn't sort and pick the first n is that if you have `us-east-1` with 6 AZs then it will always pick a,b,c and there is an uneven distribution across the AZs in the region. I thought it might be better to distribute across all the AZs to help in some small way with resource consumption.

I could make the maximum number of AZs configurable via a option and set the default to 3. I don't think we should create subnets in every AZ for a region as thats not required for a HA solution, 3 would be a good sensible default.  What do you think?

pkg/cloud/services/ec2/subnets.go
richardcase: The default CIDR block hasn't been changed and is `10.0.0.0/16`. I can add this to the documents.

Thats a very good point about not having enough IPs per subnet.  From the AWS docs CIDR blocks must be in the range: ` /16 netmask (65,536 IP addresses) and /28 netmask (16 IP addresses)` , so if a /28 is used and then we try and split into 3 subnets then each subnet is only going to have less than a handful of IPs which isn't enough. I can add a check for this and report an error

pkg/cloud/services/ec2/subnets.go
vincepri: @detiber @randomvariable This should be considered a breaking API change, are we planning to bump to a minor release?

pkg/cloud/services/ec2/subnets.go
vincepri: Even 16 IPs is very low, @randomvariable what do you think should be the bare minimum here?

randomvariable: /24 is still fine. Don't think you want to go too far below that number of IP addresses. In particular, that gives enough "breathing space" for ELBv1 to consume IP space (need ~30 free IP addresses minimum otherwise it starts to have issues).

Have previously used /17 for the VPC size as a compromise for users who may have on-prem networks they need to connect up.

richardcase: I will add a check for the VPC CIDR block. 

If its above /24 should i error or allow with a warning?

pkg/cloud/services/ec2/subnets.go
richardcase: I've added options to expose the maximum number of AZs to use and also whether to select randomly or by alphabetical order if there is more AZs than the max to use.

pkg/cloud/services/ec2/routetables.go
richardcase: @vincepri - i thought we could perhaps create a new issue for checking length of labels as it applies to quite a few places....not just the routetable and subnets that i have added to. It could be a good first issue for someone?

pkg/cloud/services/ec2/subnets.go
richardcase: These are the 2 options i have added:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/26f2a95af8888bc2a936a322a5f15ee549517206/api/v1alpha3/types.go#L217

pkg/cloud/services/ec2/subnets.go
richardcase: @vincepri @randomvariable - should i add a check for the CIDR block as part of this change? Or shall we create another issue, which could be a 'good first issue'?

pkg/cloud/services/ec2/routetables.go
randomvariable: Yeah, let's create an issue for this.

pkg/cloud/services/ec2/subnets.go
randomvariable: I think we might need to for this PR in general.

pkg/cloud/services/ec2/subnets.go
randomvariable: Don't do any validation, was more speaking about defaults. I think people will know what they're doing if they go > /24 and will have a specific case for it.

pkg/cloud/services/ec2/routetables.go
richardcase: Created #1746

pkg/cloud/services/ec2/subnets.go
richardcase: I was thinking about this. I don't think this change is a breaking change to the API, as its only added 2 new optional fields.

But the behaviour has changed in 2 ways for new cluster creation:

- For a managed VPC, default subnet creation has changed to be multi-az
- For a unmanaged VPC, it requires you to explicitly specify the subnet id's you want to use and they must exists. This brings it in line with 'bring-your-own-networking'

docs/existing-aws-infrastructure.md
randomvariable: Marking resolved as per Slack and CAPA meeting of 2020/06/15.

randomvariable:
Based on discussions on Slack and in the CAPA meeting of 2020/06/15:

There is an acknowledged API change in that the specification of subnets is now mandatory. However, this is more in line with expected behaviour for people bringing their own subnets, and those fields were already in the spec, and part of CAPA's internal consumption.

The secondary change is defaulting to multi-AZ deployment, but this is desired behaviour too.

Finally, other API changes are additive.

Based on these considerations, moving to /lgtm.

Assign @sethp-nr for final approval.

rudoi:
/approve

awesome!

pkg/cloud/services/ec2/subnets.go
```diff
@@ -146,6 +133,65 @@ LoopExisting:
...
+	}
+
+	maxZones := defaultMaxNumAZs
+	if s.scope.VPC().AvailabilityZoneUsageLimit != nil {
+		maxZones = *s.scope.VPC().AvailabilityZoneUsageLimit
+	}
+	selectionScheme := infrav1.AZSelectionSchemeOrdered
+	if s.scope.VPC().AvailabilityZoneSelection != nil {
+		selectionScheme = *s.scope.VPC().AvailabilityZoneSelection
+	}
```
detiber, line 149: Since we are defaulting these values via OpenAPI/kubebuilder annotations, I'm wondering if it would make more sense to error if these are not set here rather than potentially having to worry about the openapi defaults and the defaults encoded here diverging

pkg/cloud/services/ec2/subnets.go
```diff
@@ -54,80 +56,66 @@ func (s *Service) reconcileSubnets() error {
...
-				return errors.New("expected at least one private subnet available for use, got 0")
-			}
+	unmanagedVPC := s.scope.VPC().IsUnmanaged(s.scope.Name())
 
-			subnets = append(subnets, &infrav1.SubnetSpec{
-				CidrBlock:        defaultPrivateSubnetCidr,
-				AvailabilityZone: zones[0],
-				IsPublic:         false,
-			})
+	if len(subnets) == 0 {
```
detiber, line 61: What happens in the case that we hit a transient error during initial reconciliation before we are able to create all of the subnets returned by `s.getDefaultSubnets()`? Would we only detect any subnets that were already created and persisted into the AWSCluster.Spec from the first reconciliation?

What about if no subnets get persisted back to the AWSClusterSpec, do we need to worry about potential subnet conflicts when trying to reconcile a second time?

Or if there is an error during initial reconciliation and no subnets are persisted back to the AWSClusterSpec for some reason (say a controller crash or deployment update happens at an inopportune time) could we possibly end up with multiple subnets being attached to the same AZ (when `AZSelectionSchemeRandom` is selected)?

richardcase: Thats a good point @detiber. 

Short term the easiest way to guard against this is to exit from the subnet/network reconcilation look after `subnets, err = s.getDefaultSubnets()` 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/ec2/subnets.go#L70
so that the subnets are are saved into the NetworkSpec. Then on the next reconciliation loop it will create the subnets in the spec (like if you have an unmanaged VPC).

Longer term perhaps it would be better to move the `s.getDefaultSubnets()` into the defaulting webhook.

Shall i do the small shorter term change and then create a issue for moving the defaulting into the webhook?

pkg/cloud/services/ec2/subnets.go
richardcase: I think anything that guards against divergence would be good

pkg/cloud/services/ec2/subnets.go
detiber: +1 to the change to returning early after the call to getDefaultSubnets and relying on a subsequent reconciliation to create the subnets.

I do like the idea of handling it through a webhook, however I don't think we'd be able to do that with the current multi-tenancy model, since the webhook is running without ec2 credentials. Perhaps that is something that we could do after we support native multi-tenancy. @randomvariable do you have any thoughts about that?

PR #1724, opened 2020-05-14
✨ Add EKS control plane

### LINKED ISSUE #1787: EKS Control Plane Implementation
/kind feature

Implement a new EKS control plane to help with the EKS implementation (#1534). This will handle the provisioning and management of a AWS EKS Cluster. Nodegroup support will be added via other issues.

It needs to support the following:
- EKS Cluster Creation
- EKS Cluster update
- EKS version update

The parent tracking issue for full EKS support is #1534

### Tasks

- [x] IAM Roles Creation
- [x] IAM Roles Update (including trust profiles)
- [x] IAM Roles Deletion
- [x] EKS Cluster Creation
- [x] EKS Cluster Update (for example enabling logging)
- [x] EKS Cluster version update (this may be handled by the item above)
- [x] Security groups need investigating and scope updating
- [x] EKS Cluster Deletion
- [x] AWSManagedCluster Support
- [x] Kubeconfig creation
- [ ] Unit tests
- [ ] Start implementation of e2e tests
- [x] Tags only update to AWSManagedControlPlane (will use change from issue #1843)

richardcase:
Add support for EKS. This is WIP.

Fixes #1787


dthorsen:
@richardcase We will be interested in using this functionality soon. Is there anything we can do to help get this merged?

richardcase:
@dthorsen - i have another PR that i want to get merged first and then i will focis on this and push the additional changes i have made.

rudoi:
hi @richardcase 👋, any updates for this PR? are there any TODO pieces of this that could perhaps be split into separate issues?

we (at New Relic) are looking to start work on the worker node aspect next week and would love to help get this controlplane support merged in any way we can.

richardcase:
@rudoi - been distracted by something else. Will pick it up again today and try and get it into a place where it can be used.

pkg/internal/tristate/tristate.go
```diff
@@ -0,0 +1,14 @@
+package tristate
```
richardcase, line 1: @michaelbeaumont - could you add the licence header comment to this?

config/crd/bases/exp.infrastructure.cluster.x-k8s.io_awsmanagedclusters.yaml
```diff
@@ -0,0 +1,86 @@
...
+---
+apiVersion: apiextensions.k8s.io/v1
+kind: CustomResourceDefinition
+metadata:
+  annotations:
+    controller-gen.kubebuilder.io/version: v0.2.9
+  creationTimestamp: null
+  name: awsmanagedclusters.exp.infrastructure.cluster.x-k8s.io
+spec:
+  group: exp.infrastructure.cluster.x-k8s.io
```
vincepri, line 11: I'd suggest to use a different group, and avoid using `exp` in the package name. In a future release, we might want to promote these API types outside the experimental package and that would be much harder to do if the group differs.

One suggestion I have is to use either start from `v1alpha1` and use `eks.aws.infrastructure.cluster.x-k8s.io`, or `v1alpha3` and use `aws.infrastructure.cluster.x-k8s.io`.

rudoi: i'm more a fan of `v1alpha3` and `aws.infrastructure.cluster.x-k8s.io` personally. if we can agree to that I'll also follow a similar pattern in our upcoming AWSMachinePool PR.

mytunguyen: awsmachine’s group is `infrastructure.cluster.x-k8s.io` but
awsmachinepool will be `aws.infrastructure.cluster.x-k8s.io`?

rudoi: hmm great point... AWSCluster and AWSManagedCluster should probably be in the same API group 🤔

vincepri: I am fine keeping it within the same API group

richardcase: Yes, very interesting and a good point.

Shall i will change the group to `infrastructure.cluster.x-k8s.io` to match the existing resource kinds. Now its been pointed out that makes more sense to me as we will want to promote these out of `exp` into the main `api` at some point in the future.

vincepri: Thank you!

richardcase: I've changed this to `infrastructure.cluster.x-k8s.io`

pkg/cloud/services/eks/cluster.go
```diff
@@ -0,0 +1,502 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to describe eks clusters")
+	}
+
+	if cluster == nil {
+		cluster, err = s.createCluster()
+		if err != nil {
+			return errors.Wrap(err, "failed to create cluster")
+		}
+		s.scope.Info("Created EKS control plane: %s", *cluster.Name)
```
rudoi, line 56: these loggers are ("message", ...keysAndValues)

so this prints `"msg"="Created EKS control plane: %s"` and then has the cluster name as a key with an empty value.

richardcase: Ah, good spot. Thanks @rudoi

pkg/cloud/services/eks/config.go
```diff
@@ -0,0 +1,131 @@
...
+	clusterName := *cluster.Name
+	userName := fmt.Sprintf("%s-admin", clusterName)
+	contextName := fmt.Sprintf("%s@%s", userName, clusterName)
+
+	cfg := &api.Config{
+		APIVersion: api.SchemeGroupVersion.Version,
+		Clusters: map[string]*api.Cluster{
+			clusterName: {
+				Server:                   *cluster.Endpoint,
+				CertificateAuthorityData: []byte(*cluster.CertificateAuthority.Data),
```
rudoi, line 81: I noticed that in the generated kubeconfig, this value was base64 encoded twice. 

`aws eks describe-cluster` outputs the CA data correctly (base64 encoded once), so I'm actually not sure how it ends up in the secret encoded twice.

exp/api/v1alpha3/awsmanagedcontrolplane_types.go
```diff
@@ -0,0 +1,184 @@
...
+	// +optional
+	Bastion infrav1.Bastion `json:"bastion"`
+
+	// TokenMethod is used to specify the method for obtaining a client token for communicating with EKS
+	// iam-authenticator - obtains a client token using iam-authentictor
+	// aws-cli - obtains a client token using the AWS CLI
+	// Defaults to iam-authenticator
+	// +kubebuilder:default=iam-authenticator
+	// +kubebuilder:validation:Enum=iam-authenticator;aws-cli
+	TokenMethod *EKSTokenMethod `json:"tokenMethod,omitempty"`
```
rudoi, line 96: This totally slipped my mind when we were initially discussing the kubeconfig generation - CAPI uses the kubeconfig, not CAPA, so in order for this to work either the aws CLI or the iam authenticator needs to be bundled in the CAPI image and then it also needs to mount AWS credentials. 

It may be worth exploring creating a CAPI serviceaccount and generating the kubeconfig from that. Right now, CAPI is caught in a loop: `getting credentials: exec: exec: \"aws\": executable file not found in $PATH" `

rudoi: The generated kubeconfig works fine for local use, though, save for the base64 issue i mentioned in another comment.

exp/api/v1alpha3/awsmanagedcontrolplane_types.go
```diff
@@ -0,0 +1,184 @@
...
+// EncryptionConfig specifies the encryption configuration for the EKS clsuter
+type EncryptionConfig struct {
+	// Provider specifies the ARN or alias of the CMK (in AWS KMS)
+	Provider *string `json:"provider,omitempty"`
+	//Resources specifies the resources to be encrypted
+	Resources []*string `json:"resources,omitempty"`
+}
+
+// AWSManagedControlPlaneStatus defines the observed state of AWSManagedControlPlane
+type AWSManagedControlPlaneStatus struct {
```
rudoi, line 121: I think we also need an `Initialized` in here:

https://github.com/kubernetes-sigs/cluster-api/blob/e955160f6f30f61eabce65231899a1fe9d513046/controlplane/kubeadm/api/v1alpha3/kubeadm_control_plane_types.go#L103-L106

https://github.com/kubernetes-sigs/cluster-api/blob/f245e8ac13677b8ad2e7079924d0552cc944b5a7/controllers/cluster_controller_phases.go#L230-L238

I don't know if this actually gives us any value though. Not sure if the distinction between ready and initialized means anything to us in EKS land.

rudoi: If there's no value, I'll update the EKSConfig PR to not check for it.

pkg/cloud/services/eks/config.go
richardcase: I'll get that fixed, thanks.

exp/api/v1alpha3/awsmanagedcontrolplane_types.go
richardcase: I've had a nagging doubt around the kubeconfig for a while and thats great @rudoi that you've been taking things for a spin. 

@rudoi @detiber - so the CAPI controllers would have a service account that they can use when using the generated kubeconfig?

exp/api/v1alpha3/awsmanagedcontrolplane_types.go
richardcase: Thats an interesting one. I did have `Initialized` in the status originally but took it out as there was no distinction from `Ready` for the managed control plane. I can add it it back in.

Having a look it seems i have also lost somewhere along the way setting `Ready`.

@rudoi i'll add Initialized in (and restore the setting of `Ready`

exp/api/v1alpha3/awsmanagedcontrolplane_types.go
rudoi: I think this can work if the generated kubeconfig secret contains a kubeconfig created from a serviceaccount token. At that point CAPI would be able to reconcile Cluster objects properly. then we slap a big warning on the docs like "hey, for AWSManagedClusters CAPI uses a serviceaccount, use the `aws` CLI to get credentials.  

A separate issue here (and I think this is fine for a follow-up PR) - By default in EKS, the only IAM entity that can access the Kubernetes cluster is the entity that created it. It would be awesome if the spec for AWSManagedControlPlane had a way to specify additional IAM roles (or at least one for an admin role).

richardcase: Yes that right, by default on creation only the user that created the cluster has access.  That would be great if we add `AdditionalRoles` or something in the control plane spec.  That should be too hard at all, especially with: https://github.com/kubernetes-sigs/aws-iam-authenticator#full-configuration-format.

And also agreed on the docs update. Shall we create a issues to cover these and link to the parent EKS issue?

rudoi: Let's do that - i'll create issues for the AdditionalRoles and docs updates.

rudoi: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1850

exp/controllers/awsmanagedcontrolplane_controller.go
```diff
@@ -0,0 +1,292 @@
...
+	controllerutil.AddFinalizer(managedScope.ControlPlane, infrav1exp.ManagedControlPlaneFinalizer)
+	if err := managedScope.PatchObject(); err != nil {
+		return ctrl.Result{}, err
+	}
+
+	ec2Service := ec2.NewService(managedScope)
+	networkSvc := network.NewService(managedScope)
+	ekssvc := eks.NewService(managedScope)
+
+	if err := networkSvc.ReconcileNetwork(); err != nil {
```
rudoi, line 188: I wonder if ReconcileSecurityGroups needs to be pulled out of this method - each EKS cluster comes with a security group that gets applied to the underlying controlplane infrastructure, and that SG needs to be applied to all nodes. Because of that I'd expect ReconcileNetwork to assign the "controlplane" security group to THAT security group, as the one that ReconcileNetwork currently creates doesn't add any value as there aren't any control plane "instances" in the EKS case.

With a hacky workaround for #1852 and the EKS-created SG added to my AWSMachine's `additionalSecurityGroups`, I was actually able to use the EKS bootstrap provider and join a node. So we're really getting there!

richardcase: Thats a good point @rudoi . I will pull the `ReconcileSecurityGroups` out.

I will also add the id for the control plane security group to the status of the `AWSManagedControlPlane`. I can look this up by looking at the value of the `aws:eks:cluster-name` tag on the security group.

pkg/cloud/services/eks/cluster.go
richardcase: This has been changes by @michaelbeaumont

pkg/cloud/services/eks/config.go
richardcase: Latest commit should fix this.

exp/controllers/awsmanagedcontrolplane_controller.go
richardcase: I've extracted out the securitygroup reconciliation from the network service as part of #1857. Additional the latest commit will lookup the automatically created security group and adds the details to the status section

exp/api/v1alpha3/awsmanagedcontrolplane_types.go
richardcase: Added Initialize

randomvariable:
Follow up question from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1815/files#r462116316, is the correct status being set?

richardcase:
> Follow up question from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1815/files#r462116316, is the correct status being set?

Both `Ready` and `Initialized` are being set now. Previously it was only `Ready`

exp/api/v1alpha3/awsmanagedcontrolplane_types.go
randomvariable: there's sample python code here https://github.com/kubernetes-sigs/aws-iam-authenticator#api-authorization-from-outside-a-cluster .

I think it should be possible to use the AWS SDK for Go to generate the presigned URL against the STS get caller identity API call, and pass that in as the initial bearer token to then generate a service account and store the kubeconfig back on the management server.

richardcase: Thats a great idea @randomvariable. I will try that out now.

pkg/cloud/services/eks/roles.go
```diff
@@ -0,0 +1,362 @@
+/*
```
randomvariable, line 1: Generally unsure of doing a bunch of IAM work inside the controller as it then has to be super-privileged over AWS accounts, vs the current model where clusterawsadm is run administratively as a one-time operation and then the controller can be run with a much smaller set of permissions.

pkg/cloud/services/eks/tags.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+
+	return nil
+}
+
+func (s *Service) getEKSTagParams(id string) *infrav1.BuildParams {
+	name := fmt.Sprintf("%s-eks-cp", s.scope.Name())
+
+	return &infrav1.BuildParams{
+		ClusterName: s.scope.Name(),
```
randomvariable, line 45: Same comment as https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1815#discussion_r469903727 wrt to converting kubernetes resource name to the eks resource name

pkg/cloud/services/eks/roles.go
richardcase: I know what you mean. The reason i added this was that with EKS its not necessarily a one-time operation and that it could be per cluster (but the roles could be reused as well)

pkg/cloud/services/eks/tags.go
rudoi: If the object name isn't going to equal the EKS cluster name, wdyt about also adding a status field with the resolved name in it? That way I can consume that from the EKS bootstrap provider.

pkg/cloud/services/eks/roles.go
richardcase: Still thinking about this one.  This code is creating the cluster service iam role: https://docs.aws.amazon.com/eks/latest/userguide/service_IAM_role.html.

Previously i have always created a role per cluster (and this is also what eksctl does) as this allows the role to be tailored to the role of the cluster. For example by adding extra policies:
https://github.com/richardcase/cluster-api-provider-aws/blob/ekscontrolplane/exp/api/v1alpha3/awsmanagedcontrolplane_types.go#L62

Like you say we could try and put it into `clusterawsadm` and perhaps create a default role when creating the initial stack. And then have a subcommand where a new role could be created with additional policies if a cluster needed something unique. Although the role then is managed via a separate mechanism and isn't attached to the cluster reconciliation.

If we keep it in the controller then the changes to the permissions can be scoped to this:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1780#issuecomment-667997298

@randomvariable @rudoi @detiber - happy to go with what you think is best here.

pkg/cloud/services/eks/tags.go
richardcase: Ha, we where thinking along exactly the same lines. I have added this to the status:
https://github.com/richardcase/cluster-api-provider-aws/blob/ekscontrolplane/exp/api/v1alpha3/awsmanagedcontrolplane_types.go#L124

pkg/cloud/services/eks/roles.go
rudoi: You can specify a role in the Spec right?

I'd be happy to leave this as-is for now and then add something optional to `clusterawsadm` in a follow-up.

randomvariable: The difference is that `eksctl` is running on the command line as the end user. In a more permanent management cluster, there's a privilege escalation issues that Kubernetes RBAC can't protect against.

In the EC2 based controller, we've been pretty strict in that the default cloudformation template only allows instance profiles based on IAM roles using the `cluster-api-provider-aws.sigs.k8s.io` suffix, allowing AWS admins to restrict what roles EC2 instances can be created with.

In particular, allowing `RoleAdditionalPolicies` means anyone who can create a cluster can create a role with administrative access.

Would it be possible to have this behaviour opt-in as a flag?

richardcase: Thats a good point about it being a different use case.

We could have `RoleAdditionalPolicies` as an opt in feature flag, no problem at all. If we did that would you be more comfortable having the IAM role creation in the controller?

randomvariable: Can we have at least the ability to disable write access to IAM entirely? Can also optionally include one for RoleAdditionalPolicies

richardcase: Sounds like a plan

pkg/cloud/services/eks/tags.go
richardcase: I have changed the code to adopt the naming convention

pkg/cloud/services/eks/roles.go
richardcase: I'm going to add 2 additional feature flags to cover this.

feature/feature.go
```diff
@@ -28,9 +28,20 @@ const (
...
 
+	// EKS is used to enable EKS support
 	// owner: @richardcase
 	// alpha: v0.4
 	EKS featuregate.Feature = "EKS"
+
+	// EKSDisableIAM will disable all IAM resource creation/modification if set to true
+	// owner: @richardcase
+	// alpha: v0.4
+	EKSDisableIAM featuregate.Feature = "EKSDisableIAM"
```
randomvariable, line 39: As this is security impacting, this should be an opt-in feature, not an opt out.

richardcase: Very good point @randomvariable - will get that changed now......

pkg/cloud/services/eks/cluster.go
```diff
@@ -0,0 +1,522 @@
...
+	if err != nil {
+		return errors.Wrap(err, "failed to describe eks clusters")
+	}
+
+	if cluster == nil {
+		cluster, err = s.createCluster(eksClusterName)
+		if err != nil {
+			return errors.Wrap(err, "failed to create cluster")
+		}
+		s.scope.Info("Created EKS ccluster in AWS", "cluster-name", eksClusterName)
```
rudoi, line 58:
```suggestion
		s.scope.Info("Created EKS cluster in AWS", "cluster-name", eksClusterName)
```

exp/controllers/awsmanagedcluster_controller.go
```diff
@@ -0,0 +1,192 @@
...
+		return fmt.Errorf("failed adding a watch for ready clusters: %w", err)
+	}
+
+	// Add a watch for AWSManagedControlPlane
+	if err = controller.Watch(
+		&source.Kind{Type: &infrav1exp.AWSManagedControlPlane{}},
+		&handler.EnqueueRequestsFromMapFunc{
+			ToRequests: handler.ToRequestsFunc(r.managedControlPlaneToManagedCluster),
+		},
+		predicates.ClusterUnpaused(r.Log),
```
rudoi, line 149: This predicate will try to cast the object into a `Cluster` and throw an error:

```
[manager] I0824 16:03:56.866549       7 cluster_predicates.go:131] controllers/AWSManagedCluster "msg"="Expected Cluster" "eventType"="update" "predicate"="ClusterUpdateUnpaused" "type"="infrastructure.cluster.x-k8s.io/v1alpha3, Kind=AWSManagedControlPlane"
```

I think this is why we're seeing AWSManagedCluster not reconcile immediately when AWSManagedControlPlane status is set to Ready.

richardcase: Thats it, thanks.

pkg/cloud/services/eks/cluster.go
richardcase: Thanks, fixed the typo

exp/controllers/awsmanagedcluster_controller.go
richardcase: Removed the ClusterUnpaused

pkg/cloud/services/eks/errors.go
```diff
@@ -0,0 +1,36 @@
...
+package eks
+
+import "github.com/pkg/errors"
+
+var (
+	// ErrClusterExists is an error if a EKS cluster already exists with
+	// the same name in the spec but that isn't owned by the CAPI cluster
+	ErrClusterExists = errors.New("eks cluster already exists with same name but isn't owned by cluster")
+	// ErrUnknowTokenMethod defines an error if a unsupported token generation method is supplied
+	ErrUnknowTokenMethod = errors.New("unknown token method")
```
randomvariable, line 26:
```suggestion
	// ErrUnknownTokenMethod defines an error if a unsupported token generation method is supplied
	ErrUnknownTokenMethod = errors.New("unknown token method")
```

And fix up the consumers.

randomvariable (commented):
One typo, but otherwise lgtm

pkg/cloud/services/eks/errors.go
richardcase: Thanks @michaelbeaumont for fixing. I also just fixed a type with one of the error messages.

randomvariable:
ok, i think we're good to go, can we clean up the commit history (either squashed into one, or 'tell the story'), and then can lgtm.

michaelbeaumont:
@randomvariable Cleaned up now to where it "tells the story" pretty well

randomvariable:
/approve
based on conversation https://kubernetes.slack.com/archives/CD6U2V71N/p1598515539065200

PR #1726, opened 2020-05-16
✨ if subnets are provided in network spec then filter to those

### LINKED ISSUE #1659: Support selection/limiting of private subnets to use for control plane replicas
/kind feature

The ability to select or limit the acceptable private subnets and/or failure domains where kubeadm control plane instances may be placed.

The current behavior results in _any_ private subnet within a vpc being selected for control plane instances. In my vpc environment there are 5 availability zones with private subnets in each, but it is preferred to only place control plane instances in 3 of those 5. This is particularly difficult since my account's `us-east-1e` availability zone does not support newer instance types, and so setting my control plane instance type to `c5.xlarge` (or any c5-family instance) results in randomly being unable to start a 3rd instance if a `us-east-1e` subnet is selected for a control plane instance placement.

**Environment:**

- Cluster-api-provider-aws version: `v0.5.1`
- Kubernetes version: (use `kubectl version`): `1.15`
- OS (e.g. from `/etc/os-release`): `ubuntu 18.04`

andrewmyhre:
**What this PR does / why we need it**: For AWS accounts with several subnets but where only a subset are preferred for tenant use, and where there are strict controls around subnet management prohibiting the addition of tags, it should to be possible to specify the set of subnets to use in the AWSCluster spec.  

Fixes #1659


vincepri:
/hold

If I'm reading this change correctly, this would be a breaking change, because subnets will only be populated once. Can you confirm if this is the case or not?

sethp-nr:
It looks like that to me as well – coincidentally, there's another PR open now that changes the same behavior: #1721

andrewmyhre:
Closing as #1721 has already been merged

PR #1729, opened 2020-05-19
 🐛 AWSMachine ssh key should defer to AWSCluster when nil

### LINKED ISSUE #1728: SSH key name is not correctly set in some cases
/kind bug

Follow up to #1621
When an AWSCluster has an `SSHKeyName: ""` and an AWSMachine has an `SSHKeyName: nil`, the launched instances are given a `KeyName=""`.

In this case the instance KeyName should not be specified, because the AWSMachine value is nil, which we should take to mean defer to the AWSCluster setting. When both the AWSCluster and the AWSMachine have an unspecified SSHKeyName value, that's when we use the default value.

The expected behavior for setting an SSH key name on an instance is as follows:
- the AWSMachine value takes precedence over the AWSCluster value
- if the AWSMachine SSHKeyName is an empty string, do not set an SSH key name value
- if the AWSMachine SSHKeyName is a string, set the SSH key name value to SSHKeyName
- if the AWSMachine SSHKeyName value is nil, then:
  - if the AWSCluster SSHKeyName value is an empty string, do not set and SSH key name value
  - if the AWSCluster SSHKeyName value is a string, set the SSH key name value to SSHKeyName
  - if the AWSCluster SSHKeyName is nil, use the default ssh key value

This is deceptively complex, so maybe there's an argument to just add a boolean to disable SSH Keys instead?

**Environment:**

- Cluster-api-provider-aws version: v0.5.0+
- Kubernetes version: (use `kubectl version`): v1.16.8
- OS (e.g. from `/etc/os-release`): ubuntu 18

andrewmyhre:
**What this PR does / why we need it**: Unexpected behavior when the AWSCluster.Spec.SSHKeyName is set to `""` and the AWSMachine.Spec.SSHKeyName is set to `nil` - results in the instance.KeyName being set to `""`, instead of being unset.

Fixes #1728

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -226,13 +226,21 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
 
 	// If SSHKeyName WAS NOT provided in the AWSMachine Spec, fallback to the value provided in the AWSCluster Spec.
 	// If a value was not provided in the AWSCluster Spec, then use the defaultSSHKeyName
-	input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
-	if input.SSHKeyName == nil {
-		if scope.AWSCluster.Spec.SSHKeyName != nil {
-			input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
-		} else {
-			input.SSHKeyName = aws.String(defaultSSHKeyName)
+	// Note that:
+	// - a nil AWSMachine.Spec.SSHKeyName value means use the AWSCluster.Spec.SSHKeyName SSH key name value
+	// - nil values for both AWSCluster.Spec.SSHKeyName and AWSMachine.Spec.SSHKeyName means use the default SSH key name value
+	// - an empty string means do not set an SSH key name at all
+	// - otherwise use the value specified in either AWSMachine or AWSCluster
+	if scope.AWSMachine.Spec.SSHKeyName != nil {
+		if *scope.AWSMachine.Spec.SSHKeyName != "" {
+			input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
 		}
+	} else if scope.AWSCluster.Spec.SSHKeyName != nil {
+		if *scope.AWSCluster.Spec.SSHKeyName != "" {
+			input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
+		}
+	} else {
+		input.SSHKeyName = aws.String(defaultSSHKeyName)
 	}
 
 	s.scope.V(2).Info("Running instance", "machine-role", scope.Role())
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -231,16 +231,21 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
 	// - nil values for both AWSCluster.Spec.SSHKeyName and AWSMachine.Spec.SSHKeyName means use the default SSH key name value
 	// - an empty string means do not set an SSH key name at all
 	// - otherwise use the value specified in either AWSMachine or AWSCluster
-	if scope.AWSMachine.Spec.SSHKeyName != nil {
-		if *scope.AWSMachine.Spec.SSHKeyName != "" {
-			input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
-		}
-	} else if scope.AWSCluster.Spec.SSHKeyName != nil {
-		if *scope.AWSCluster.Spec.SSHKeyName != "" {
-			input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
-		}
-	} else {
-		input.SSHKeyName = aws.String(defaultSSHKeyName)
+	var prioritizedSSHKeyName string
+	switch {
+	case scope.AWSMachine.Spec.SSHKeyName != nil:
+		// prefer AWSMachine.Spec.SSHKeyName if it is defined
+		prioritizedSSHKeyName = *scope.AWSMachine.Spec.SSHKeyName
+	case scope.AWSCluster.Spec.SSHKeyName != nil:
+		// fallback to AWSCluster.Spec.SSHKeyName if it is defined
+		prioritizedSSHKeyName = *scope.AWSCluster.Spec.SSHKeyName
+	default:
+		prioritizedSSHKeyName = defaultSSHKeyName
+	}
+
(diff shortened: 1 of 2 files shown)


randomvariable:
need a 2nd opinion on this. ETOOMANYIFS

pkg/cloud/services/ec2/instances.go
```diff
@@ -222,13 +222,21 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
...
+	// - nil values for both AWSCluster.Spec.SSHKeyName and AWSMachine.Spec.SSHKeyName means use the default SSH key name value
+	// - an empty string means do not set an SSH key name at all
+	// - otherwise use the value specified in either AWSMachine or AWSCluster
+	if scope.AWSMachine.Spec.SSHKeyName != nil {
+		if *scope.AWSMachine.Spec.SSHKeyName != "" {
+			input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
+		}
+	} else if scope.AWSCluster.Spec.SSHKeyName != nil {
+		if *scope.AWSCluster.Spec.SSHKeyName != "" {
 			input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
```
sethp-nr, line 236: I think it's possible to fall through this `if` without setting a value for `input.SSHKeyName` – is that intentional?

If not, consider:
```suggestion
	if scope.AWSMachine.Spec.SSHKeyName != nil && *scope.AWSMachine.Spec.SSHKeyName != "" {
		input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
	} else if scope.AWSCluster.Spec.SSHKeyName != nil && *scope.AWSCluster.Spec.SSHKeyName != "" {
		input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
```

sethp-nr: Oh, now I see it: 

>  an empty string means do not set an SSH key name at all

My mistake, sorry. I believe the code implements the spec in the comment, but it might be worth an issue around cleaning up this list of options + fallbacks for v1alpha4?

sethp-nr:
It looks like the test error is the linter disagreeing with a few things: if it helps, running `make lint-full` is usually how I iterate on fixing those.

pkg/cloud/services/ec2/instances.go
detiber: I'm wondering if we can simplify this logic a bit with something like this instead:
```go
	switch {
	case scope.AWSMachine.Spec.SSHKeyName != nil:
		// prefer AWSMachine.Spec.SSHKeyName if it is defined
		input.SSHKeyName = scope.AWSMachine.Spec.SSHKeyName
	case scope.AWSCluster.Spec.SSHKeyName != nil:
		// fallback to AWSCluster.Spec.SSHKeyName if it is defined
		input.SSHKeyName = scope.AWSCluster.Spec.SSHKeyName
	default:
		input.SSHKeyName = aws.String(defaultSSHKeyName)
	}
```

As far as I can tell, there really isn't any reason to treat an empty string as any different than an explicit value as far as the fallback logic and using a switch potentially helps remove some of the complexity related to trying to grok the nested conditionals.

andrewmyhre: > As far as I can tell, there really isn't any reason to treat an empty string as any different than an explicit value as far as the fallback logic and using a switch potentially helps remove some of the complexity related to trying to grok the nested conditionals.

There is a meaningful distinction, which is not setting a key name vs setting the default key name. Maybe adding a `dontUseSSHKeys` boolean would simplify things?

detiber: > There is a meaningful distinction, which is not setting a key name vs setting the default key name. Maybe adding a dontUseSSHKeys boolean would simplify things?

Ah, I think I'm following now. Apologies about the previous confusion.

What if we separated the fallback handling from the handling of empty string? I think it might help clear up some of the confusion around trying to handle both fallback logic and the impedance mismatch between our api and the aws api a bit better.

```go
	// handle fallback prioritization
	var prioritizedSSHKeyName string
	switch {
	case scope.AWSMachine.Spec.SSHKeyName != nil:
		// prefer AWSMachine.Spec.SSHKeyName if it is defined
		prioritizedSSHKeyName = *scope.AWSMachine.Spec.SSHKeyName
	case scope.AWSCluster.Spec.SSHKeyName != nil:
		// fallback to AWSCluster.Spec.SSHKeyName if it is defined
		prioritizedSSHKeyName = *scope.AWSCluster.Spec.SSHKeyName
	default:
		prioritizedSSHKeyName = defaultSSHKeyName
	}

	// Only set input.SSHKeyName if the user did not explicitly request no ssh key be set (explicitly setting "" on either the Machine or related Cluster)
	if prioritizedSSHKeyName != "" {
		input.SSHKeyName = aws.String(prioritiziedSSHKeyName)
	}
```

rothgar: Is it possible to not have an ssh key at all? I would prefer to not use ssh and instead connect with session manager but it doesn't appear possible because AWS_SSH_KEY_NAME is required.

detiber: @rothgar that is a good point, the template handling through clusterctl does seem to make consuming this a bit difficult today.

detiber: The idea is that if one does not specify a key to use then we would use the configured default, but if one explicitly specifies no key with an empty string, then we would explicitly not use a key when creating the instance.

rothgar: I'm fine with `AWS_SSH_KEY_NAME=""` hard to make it explicit while also not letting someone lock themselves out.

rothgar: If ssh is explicitly set to "" there should also be logic to not create the basiton security group. Are there other resources created to allow SSH?

detiber: > If ssh is explicitly set to "" there should also be logic to not create the basiton security group. Are there other resources created to allow SSH?

I believe longer term we are planning on allowing more customization here. It gets a bit tricky because this value can be overridden through a similar setting on Machine* resources, so we can't necessarily assume that if it is set to "" on the Cluster that it isn't set to a different value on one or more Machines that belong to the Cluster.

andrewmyhre:
@detiber finally got around to updating this

detiber:
@andrewmyhre thanks, this is looking good to me, just need to squash the commits

ncdc:
@andrewmyhre will you have time to squash the commits soon?

andrewmyhre:
Closing in favor of #1932 which is up to date with master

PR #1730, opened 2020-05-20
✨Make machine's providerID consistent with node providerID

alexander-demichev:
Closes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1693

This PR makes `providerID` of the machine more consistent with the node's `providerID`. See the attached issue for more details

### DIFF
--- a/api/v1alpha2/zz_generated.conversion.go
+++ b/api/v1alpha2/zz_generated.conversion.go
@@ -1009,6 +1009,7 @@ func autoConvert_v1alpha3_Instance_To_v1alpha2_Instance(in *v1alpha3.Instance, o
 	// WARNING: in.RootVolume requires manual conversion: does not exist in peer-type
 	out.NetworkInterfaces = *(*[]string)(unsafe.Pointer(&in.NetworkInterfaces))
 	out.Tags = *(*map[string]string)(unsafe.Pointer(&in.Tags))
+	// WARNING: in.AvailabilityZone requires manual conversion: does not exist in peer-type
 	return nil
 }
 
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -554,6 +554,9 @@ type Instance struct {
 
 	// The tags associated with the instance.
 	Tags map[string]string `json:"tags,omitempty"`
+
+	// Availability zone of instance
+	AvailabilityZone string `json:"availabilityZone,omitempty"`
 }
 
 // RootVolume encapsulates the configuration options for the root volume
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -18,7 +18,6 @@ package controllers
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -411,7 +410,7 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 	}
 
 	// Make sure Spec.ProviderID is always set.
-	machineScope.SetProviderID(fmt.Sprintf("aws:////%s", instance.ID))
+	machineScope.SetProviderID(instance.ID, instance.AvailabilityZone)
 
 	// See https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
 
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -19,6 +19,7 @@ package scope
 import (
 	"context"
 	"encoding/base64"
+	"fmt"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
@@ -138,8 +139,9 @@ func (m *MachineScope) GetProviderID() string {
 }
 
 // SetProviderID sets the AWSMachine providerID in spec.
-func (m *MachineScope) SetProviderID(v string) {
-	m.AWSMachine.Spec.ProviderID = pointer.StringPtr(v)
+func (m *MachineScope) SetProviderID(instanceID, availabilityZone string) {
+	providerID := fmt.Sprintf("aws:///%s/%s", availabilityZone, instanceID)
+	m.AWSMachine.Spec.ProviderID = pointer.StringPtr(providerID)
 }
 
 // GetInstanceID returns the AWSMachine instance state from the status.
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
@@ -638,6 +638,9 @@ spec:
                       - type
                       type: object
                     type: array
+                  availabilityZone:
+                    description: Availability zone of instance
+                    type: string
                   ebsOptimized:
                     description: Indicates whether the instance is optimized for Amazon
                       EBS I/O.
--- a/pkg/cloud/scope/machine_test.go
+++ b/pkg/cloud/scope/machine_test.go
@@ -201,3 +201,24 @@ func TestSetSecretARN(t *testing.T) {
 		t.Fatalf("prefix does not equal %s: %s", prefix, val)
 	}
 }
+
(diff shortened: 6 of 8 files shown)


api/v1alpha3/types.go
```diff
@@ -554,6 +554,9 @@ type Instance struct {
 
 	// The tags associated with the instance.
 	Tags map[string]string `json:"tags,omitempty"`
+
```
enxebre, line 557: The core machine and the provider API already have semantic for this -> `.failureDomain`. See https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/425e1bb337cfb76ceb38c1cce97345c26fafe08d/pkg/cloud/services/ec2/instances.go#L162-L165

controllers/awsmachine_controller.go
```diff
@@ -411,7 +411,7 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 	}
 
 	// Make sure Spec.ProviderID is always set.
-	machineScope.SetProviderID(fmt.Sprintf("aws:////%s", instance.ID))
+	machineScope.SetProviderID(fmt.Sprintf("aws:////%s/%s", instance.AvailabilityZone, instance.ID))
```
enxebre, line 414: format does not seems right here, should be:
```
aws:///<zone>/<awsInstanceId>
aws:////<awsInstanceId>
```

controllers/awsmachine_controller.go
```diff
@@ -411,7 +411,12 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 	}
 
 	// Make sure Spec.ProviderID is always set.
-	machineScope.SetProviderID(fmt.Sprintf("aws:////%s", instance.ID))
+	failureDomain := machineScope.AWSMachine.Spec.FailureDomain
```
enxebre, line 414: The particular provider failureDomain field is to be deprecated in favour of the core machine FailureDomain. So that one needs to be favoured as in
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/425e1bb337cfb76ceb38c1cce97345c26fafe08d/pkg/cloud/services/ec2/instances.go#L162-L165

controllers/awsmachine_controller.go
```diff
@@ -411,7 +411,12 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
+	if failureDomain == nil {
```
enxebre, line 415: we should use a `deref` helper function,
```
if failureDomain is nil it returns ""
else return * failureDomain
```
Then we can just
`machineScope.SetProviderID(fmt.Sprintf("aws:///%s/%s", failureDomain, instance.ID))`

alexander-demichev: 👍

controllers/awsmachine_controller.go
```diff
@@ -411,7 +412,7 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 	}
 
 	// Make sure Spec.ProviderID is always set.
-	machineScope.SetProviderID(fmt.Sprintf("aws:////%s", instance.ID))
+	machineScope.SetProviderID(fmt.Sprintf("aws:///%s/%s", aws.StringValue(machineScope.AWSMachine.Spec.FailureDomain), instance.ID))
```
enxebre, line 415: This still needs to be addressed https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1730#discussion_r428048160

controllers/awsmachine_controller.go
```diff
@@ -410,8 +411,13 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		return ctrl.Result{}, nil
 	}
 
+	failureDomain := machineScope.AWSMachine.Spec.FailureDomain
+	if failureDomain == nil {
+		failureDomain = machineScope.Machine.Spec.FailureDomain
+	}
```
enxebre, line 417: Now this logic and the format `fmt.Sprintf("aws:///%s/%s", aws.StringValue(failureDomain)`  are invariant. If we move them inside the `SetProviderID()` func and we pass only the instance.ID we can unit test the multiple scenarios for failureDomain, e.g both empty, one empty, both filled. wdyt

alexander-demichev: Good idea, that's definitely easier to test

controllers/awsmachine_controller.go
```diff
@@ -411,7 +410,7 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 	}
 
 	// Make sure Spec.ProviderID is always set.
-	machineScope.SetProviderID(fmt.Sprintf("aws:////%s", instance.ID))
+	machineScope.SetProviderID(instance.ID)
```
detiber, line 413: Would it be better to update `getOrCreate` to also return the current AZ based on the returned AWS calls instead of trying to rely on the AWSMachine or Machine fields having FailureDomain set, since it's an optional field on both resources?

alexander-demichev: Good idea. `getOrCreate` will end up converting `aws.instance` to `infrav1.instance` using `SDKToInstance`, maybe we can add AZ to `infrav1.Instance` and have conversion logic in `SDKToInstance` ? Is this an acceptable API change?

detiber: Sounds good to me, I don't expect there to be any downstream consumers of that part of the API, so changes should not be an issue.

pkg/cloud/scope/machine.go
```diff
@@ -138,8 +139,9 @@ func (m *MachineScope) GetProviderID() string {
 }
 
 // SetProviderID sets the AWSMachine providerID in spec.
-func (m *MachineScope) SetProviderID(v string) {
-	m.AWSMachine.Spec.ProviderID = pointer.StringPtr(v)
+func (m *MachineScope) SetProviderID(instanceID, availabilityZone string) {
```
enxebre, line 142: can we put some units for this?

enxebre: it might seem trivial enough as to not deserve a test, but the func is actually relying in a handwritten fmt format and an external given implementation for `StringPtr`. It’s very cheap fo us to put a unit and failry valuable if some one considers to change the impl details later on. Also it makes way easier to visualise at a glange what provider should look like for the different cases.

enxebre:
thanks Alex!
can we please make sure there's an appropriate release note for this?
/lgtm

api/v1alpha2/zz_generated.conversion.go
```diff
@@ -1009,6 +1009,7 @@ func autoConvert_v1alpha3_Instance_To_v1alpha2_Instance(in *v1alpha3.Instance, o
 	// WARNING: in.RootVolume requires manual conversion: does not exist in peer-type
 	out.NetworkInterfaces = *(*[]string)(unsafe.Pointer(&in.NetworkInterfaces))
 	out.Tags = *(*map[string]string)(unsafe.Pointer(&in.Tags))
+	// WARNING: in.AvailabilityZone requires manual conversion: does not exist in peer-type
```
detiber, line 1012: To preserve this value through a conversion that round-trips through v1alpha2, this will also need to restore the value from annotation (if present) similar to what is done around FailureDomain handling here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d3d8d2f2b505e090d0bfe517c378defb8601cffc/api/v1alpha2/awscluster_conversion.go#L62

alexander-demichev: I might be wrong but it seems to me that this case is already handled here https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d3d8d2f2b505e090d0bfe517c378defb8601cffc/api/v1alpha2/awscluster_conversion.go#L66 since `Instance` is wrapped inside `AwsCluster` .

detiber: Ah, indeed it is.

PR #1731, opened 2020-05-20
:running: Switch to using CAPI's MachineAddress instead of corev1.NodeAddress types

randomvariable:
[KEP-1664](https://github.com/kubernetes/enhancements/pull/1665) is proposing to deprecate and replace Node.Status.Address, so probably best to avoid
using those types. We have a copy of these types in CAPI, so this PR switches them over.

This doesn't really change the CRD API types other than changing the field
descriptions.

Fixes #


vincepri:
Thanks for doing this!

PR #1737, opened 2020-05-28
:bug: Ignore "not found" errors when deregistering ELB instances

### LINKED ISSUE #1736: Unable to delete machine when ELB is not present (v2)
/kind bug

This is the same issue as previously closed https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/899

I deleted my resources (`Cluster`, `AWSCluster`, `KCP`, ...) and the controller restarted right after this log line:

```
I0528 13:45:40.416475   79702 loadbalancer.go:45] controllers/AWSCluster "msg"="Reconciling load balancers" "awsCluster"="samdev-dev-a-us-east-2b-aws" "cluster"="samdev-dev-a-us-east-2b-aws" "namespace"="pipeline2"
I0528 13:45:40.682433   79702 loadbalancer.go:108] controllers/AWSCluster "msg"="Reconcile load balancers completed successfully" "awsCluster"="samdev-dev-a-us-east-2b-aws" "cluster"="samdev-dev-a-us-east-2b-aws" "namespace"="pipeline2"
I0528 13:45:40.766772   79702 controller.go:282] controller-runtime/controller "msg"="Successfully Reconciled"  "controller"="awscluster" "request"={"Namespace":"pipeline2","Name":"samdev-dev-a-us-east-2b-aws"
```

On restart, it attempted to reconcile the `AWSMachine` delete for the control plane nodes but that now fails because the ELB is already gone.

```
E0528 13:45:51.023517   79702 controller.go:258] controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile LB attachment: LoadBalancerNotFound: There is no ACTIVE Load Balancer named '2ttec84skgbqho1qo9qmyk7wz9me-k8s'\n\tstatus code: 400, request id: b9a9aaf4-bc83-48d2-b5e6-828cefe5ab4a\ncould not register control plane instance \"i-0c4b44c7dae22c788\" with load balancer\nsigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).reconcileLBAttachment\n\t/Users/ssuter/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_controller.go:591\nsigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).reconcileNormal\n\t/Users/ssuter/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_controller.go:462\nsigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).Reconcile\n\t/Users/ssuter/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_controller.go:171\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler\n\t/Users/ssuter/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.2/pkg/internal/controller/controller.go:256\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem\n\t/Users/ssuter/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.2/pkg/internal/controller/controller.go:232\nsigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).worker\n\t/Users/ssuter/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.5.2/pkg/internal/controller/controller.go:211\nk8s.io/apimachinery/pkg/util/wait.JitterUntil.func1\n\t/Users/ssuter/go/pkg/mod/k8s.io/apimachinery@v0.17.2/pkg/util/wait/wait.go:152\nk8s.io/apimachinery/pkg/util/wait.JitterUntil\n\t/Users/ssuter/go/pkg/mod/k8s.io/apimachinery@v0.17.2/pkg/util/wait/wait.go:153\nk8s.io/apimachinery/pkg/util/wait.Until\n\t/Users/ssuter/go/pkg/mod/k8s.io/apimachinery@v0.17.2/pkg/util/wait/wait.go:88\nruntime.goexit\n\t/usr/local/Cellar/go/1.14.2_1/libexec/src/runtime/asm_amd64.s:1373"  "controller"="awsmachine" "request"={"Namespace":"pipeline2","Name":"samdev-dev-a-us-east-2b-aws-control-plane-bwgpn"}
```

AWSMachine should be deleted if the ELB is already gone.

It seems like we just need some extra defenses in `reconcileLBAttachment()`. If we're in "deregister" mode and the ELB is not found, it should return success.

**Environment:**

- Cluster-api-provider-aws version: HEAD
- Kubernetes version: (use `kubectl version`): `Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-30T20:19:45Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}`

moensch:

When de-registering instances from an ELB, we ignore errors relating to "instance not found" or "loadbalancer not found". This avoids issues when the controller got interrupted **after** deleting the ELB but **before** deregistering the instances.
Also should handle the case where a controller is interrupted between the instance deregistration and the instance delete.

Fixes #1736

### DIFF
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -246,6 +246,17 @@ func (s *Service) DeregisterInstanceFromAPIServerELB(i *infrav1.Instance) error 
 	}
 
 	_, err = s.scope.ELB.DeregisterInstancesFromLoadBalancer(input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case elb.ErrCodeAccessPointNotFoundException, elb.ErrCodeInvalidEndPointException:
+				// Ignoring LoadBalancerNotFound and InvalidInstance when deregistering
+				return nil
+			default:
+				return err
+			}
+		}
+	}
 	return err
 }
 


pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -19,10 +19,11 @@ package elb
 import (
 	"fmt"
 	"reflect"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
 	"strings"
 	"time"
 
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
+
```
detiber, line 26: nit: remove space here and resort to group with the imports below

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -218,6 +219,17 @@ func (s *Service) DeregisterInstanceFromAPIServerELB(i *infrav1.Instance) error
...
+			switch aerr.Code() {
+			case elb.ErrCodeAccessPointNotFoundException, elb.ErrCodeInvalidEndPointException:
+				// Ignoring LoadBalancerNotFound and InvalidInstance when deregistering
+				return nil
+			default:
+				return err
+			}
+		}
+	}
 	return err
```
detiber, line 233: Since we are now handling the error above, this should probably explicitly return nil here now.
```suggestion
	return nil
```

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -218,6 +219,17 @@ func (s *Service) DeregisterInstanceFromAPIServerELB(i *infrav1.Instance) error
 	}
 
 	_, err = s.scope.ELB.DeregisterInstancesFromLoadBalancer(input)
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case elb.ErrCodeAccessPointNotFoundException, elb.ErrCodeInvalidEndPointException:
```
detiber, line 225: It looks like we are already doing a subset of this checking in [`IsNotFound`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/212eae6835285a130aa1688a7ab565feb529b636/pkg/cloud/services/elb/errors.go#L58-L69), would it make sense to extend that function to also check for `elb.ErrCodeAccessPointNotFoundException` and use that here?

randomvariable:
@moensch could you review the comments here?

pkg/cloud/services/elb/loadbalancer.go
moensch: @detiber The existing `IsNotFound` error does already check for `ErrCodeAccessPointNotFoundException`. I am additionally checking for `ErrCodeInvalidEndPointException` which would be an error thrown if the access point exists, but the instance you're trying to de-register was not.

So the change would be to also catch the `ErrCodeInvalidEndPointException` error in `IsNotFound` (which, during its other use in [`ReconcileLoadbalancers`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d4ade5922e04a2327f4043effd16252126dea223/pkg/cloud/services/elb/loadbalancer.go#L56) would probably never get returned). Happy to make that change though (catch that extra error code in `IsNotFound` and then simply call `IsNotFound` in `DeregisterInstanceFromAPIServerELB`.

Thinking about this more, I am realizing that this logic won't quite work and I am opening up a potential other issue with this change.
Just because we're de-registering an `AWSMachine` from the Load Balancer doesn't mean it is acceptable in **all** cases for the ELB to not exist. That condition is only acceptable if the `AWSCluster` itself is in a "delete reconcile" phase.
I'd hate to introduce some cross-check between different controllers, but do you also foresee a potential issue here that I should address? Or should I continue with the top level logic as is?

pkg/cloud/services/elb/loadbalancer.go
moensch: changed

pkg/cloud/services/elb/loadbalancer.go
moensch: Totally agree - `err` will always be `nil` in this case, so it's easier on the reader to just return `nil`.

pkg/cloud/services/elb/loadbalancer.go
randomvariable: I agree, `ErrCodeInvalidEndPointException` is semantically different, however the use case for `IsNotFound` is likely the same. It's unlikely that you need to do any cross-cluster reconciliation.

Might be better to keep this logic explicit here as you've done, in case there's a change to cluster reconciliation that would necessitate having to distinguish between the two cases. Probably long-term we should refactor the function names a bit better? WDYT @detiber?

detiber:
Current changes look good to me pending squash

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -246,7 +246,18 @@ func (s *Service) DeregisterInstanceFromAPIServerELB(i *infrav1.Instance) error
...
+				// Ignoring LoadBalancerNotFound and InvalidInstance when deregistering
+				return nil
+			default:
+				return err
+			}
+		}
```
detiber, line 258: Apologies for the late catch, but I just noticed this would swallow the error if for some reason the error returned is not an `awserr.Error`

```suggestion
		}
		return err
```

moensch: gah - missed it too. Switching back to returning `err` there.

PR #1740, opened 2020-06-04
✨  Add conditions to AWSMachine

### LINKED ISSUE #1739: Add conditions to AWSMachine
As a developer, I would like to see Conditions on an AWSMachine expressing the current state so that I could use that info for higher level objects (AWSCluster). 

AWSMachine controller does the following at the infrastructure level: 
- Create EC2 instance (with encrypted user data)
- Attach (control plane) instance to ELB 
- Ensure security groups are up-to-date
- Delete encrypted user data once it's no longer needed

Proposing to create the following conditions on the `AWSMachine` object
```
// InstanceReady reports on current status of the EC2 instance. Ready indicates the instance is in a Running state. 
// When set to false, severity can be a Warning if the instance state is pending, stopping, stopped
// When set to false, severity can be an Error if the instance state is shutting down, terminated or unknown. 
InstanceReady ConditionType = "InstanceReady"

// SecurityGroupsReady indicates the security groups are up to date on the AWSMachine. 
SecurityGroupsReady ConditionType = "SecurityGroupsReady"

// Only applicable to control plane machines. ELBAttached will report true when a control plane is successfully registered with an ELB
// When set to false, severity can be an Error if the subnet is not found or unavailable in the instance's AZ
ELBAttached ConditionType = "ELBAttached"
```
An aggregate `Ready` condition is also required to report the overall status of the AWSMachine. 

Original Conditions [CAEP](https://github.com/kubernetes-sigs/cluster-api/blob/master/docs/proposals/20200506-conditions.md)

/kind feature

gab-satchi:
This PR adds the following conditions to AWSMachine along with a summary Ready condition to report overall operational status.
- InstanceReadyCondition
- ~~BootstrapInfoReady~~
- SecurityGroupsReady
- ELBAttached

The changes focused on provisioning and operating a cluster and omitted the delete cluster flow. Unit tests added but e2e test will be added once AWSCluster has conditions that incorporates machine conditions. 

Adding hold as this uses CAPI@master in order to get the conditions work. Will un-hold once there's a release.
/hold

cc: @fabriziopandini 

Fixes #1739


controllers/awsmachine_controller.go
```diff
@@ -428,18 +434,22 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
 		machineScope.Info("Unexpected EC2 instance termination", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstanceUnexpectedTermination", "Unexpected EC2 instance termination")
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceTerminated, clusterv1.ConditionSeverityError, "")
 	default:
 		machineScope.SetNotReady()
 		machineScope.Info("EC2 instance state is undefined", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstanceUnhandledState", "EC2 instance state is undefined")
 		machineScope.SetFailureReason(capierrors.UpdateMachineError)
 		machineScope.SetFailureMessage(errors.Errorf("EC2 instance state %q is undefined", instance.State))
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceStateUnknown, clusterv1.ConditionSeverityError, "")
```
detiber, line 452: I'm wondering if it would make sense to leverage `conditions.MarkUnknown` here rather than iintruducing an unknown instance state?

controllers/awsmachine_controller.go
```diff
@@ -428,18 +434,22 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
 	}
 
 	// reconcile the deletion of the bootstrap data secret now that we have updated instance state
```
detiber, line 455: I'm wondering if there should be conditions around the creation/deletion of encrypted bootstrap data. Not necessarily something that needs to be addressed in this PR, but this just caught my eye and got me wondering :)

controllers/awsmachine_controller.go
```diff
@@ -588,10 +600,12 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 	if err := elbsvc.RegisterInstanceWithAPIServerELB(i); err != nil {
```
detiber, line 600: Would it also makes sense to update the condition above after calling `elbsvc.DeregisterInstanceFromAPIServerELB()`?

api/v1alpha3/awsmachine_types.go
```diff
@@ -28,6 +28,46 @@ const (
...
+
+	// InstanceReady False reasons
+	// (Severity=Error)
+	InstanceNotFound = "InstanceNotFound"
+	// (Severity=Error)
+	InstanceTerminated = "InstanceTerminated"
+	// (Severity=Error)
+	InstanceStateUnknown = "InstanceStateUnknown"
+	// (Severity=Warning)
+	InstanceNotReady = "InstanceNotReady"
```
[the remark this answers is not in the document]
gab-satchi: Adding Reason and Condition suffix to match the CAPI conventions. I think having severity constants for each of the condition can still lead to implementation drift if someone chooses to use the generic severity constant. I'm removing the comments but I do have some unit test coverage for expected conditions so hopefully that would help with future changes.

controllers/awsmachine_controller.go
```diff
@@ -428,18 +434,22 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 	switch instance.State {
 	case infrav1.InstanceStatePending, infrav1.InstanceStateStopping, infrav1.InstanceStateStopped:
 		machineScope.SetNotReady()
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotReady, clusterv1.ConditionSeverityWarning, "")
```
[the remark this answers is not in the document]
gab-satchi: +1 makes sense. Pending is certainly expected. I'm curious if CAPA ever does anything to lead an instance into a Stopped state. I couldn't find anything so I'm assuming Stopping / Stopped should be considered an error to prompt operator intervention.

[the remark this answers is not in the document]
detiber: It doesn't, this was added to avoid treating stopped instances as a terminal error (setting failureMessage/failureReason) at the request of users.

controllers/awsmachine_controller.go
```diff
@@ -428,18 +434,22 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
 	case infrav1.InstanceStateRunning:
 		machineScope.SetReady()
+		conditions.MarkTrue(machineScope.AWSMachine, infrav1.InstanceReadyCondition)
 	case infrav1.InstanceStateShuttingDown, infrav1.InstanceStateTerminated:
 		machineScope.SetNotReady()
 		machineScope.Info("Unexpected EC2 instance termination", "state", instance.State, "instance-id", *machineScope.GetInstanceID())
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "InstanceUnexpectedTermination", "Unexpected EC2 instance termination")
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceTerminated, clusterv1.ConditionSeverityError, "")
```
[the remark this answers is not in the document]
gab-satchi: Shutting down can only lead to termination and it looks like we make that same assumption when recording the event as well

controllers/awsmachine_controller.go
gab-satchi: ah yes. it certainly does

controllers/awsmachine_controller.go
gab-satchi: Thought a bit about that as well. I think creating encrypted bootstrap data would definitely be useful to track. In case it fails, it can report something more valuable than the current `Info` level `WaitingBootstrapInfo` condition. Deleting the data after consuming it could be valuable but might report `Ready=False` condition on the overall AWSMachine if the delete fails even if the machine is still functioning.

controllers/awsmachine_controller.go
gab-satchi: I think it makes sense for when the Deregister fails. With deletion / clean up flows, I see conditions being useful when things go wrong or get stuck. Will update

api/v1alpha3/awsmachine_types.go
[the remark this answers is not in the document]
fabriziopandini: FYI, in other PRs I'm keeping both conditions and reasons in a `condition_consts.go` file under the API; I'm also using a `const` statement to group a condition and its own related reason
see https://github.com/kubernetes-sigs/cluster-api/blob/71a233e1ea77170f297c54637a6960ed23516f85/api/v1alpha3/condition_consts.go#L29-L58

api/v1alpha3/awsmachine_types.go
```diff
@@ -28,6 +28,46 @@ const (
...
+)
+
+const (
+	// BootstrapInfoReady reports whether the bootstrap data secret is available on the Machine
+	BootstrapInfoReady clusterv1.ConditionType = "BootstrapInfoReady"
```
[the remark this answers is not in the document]
fabriziopandini: After re-reading the code, my suggestion is to remove this condition and treat `WaitingBootstrapInfo` as a reason for `InstanceReadyCondition` being false. The main motivation for this change is that generating the bootstrap info is not a responsibility of this controller/this object, and so IMO it does not make sense to expose this as a separated condition

controllers/awsmachine_controller.go
```diff
@@ -157,6 +158,8 @@ func (r *AWSMachineReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 
 	// Always close the scope when exiting this function so we can persist any AWSMachine changes.
 	defer func() {
+		// set Ready condition before AWSMachine is patched
+		conditions.SetSummary(machineScope.AWSMachine)
```
fabriziopandini, line 162: Wondering if it make sense to express the Summary with the "1 of x task completed" notation, e.g. https://github.com/kubernetes-sigs/cluster-api/blob/c437644da00406b515201a48b5b93f29b09d3878/test/infrastructure/docker/controllers/dockermachine_controller.go#L120

api/v1alpha3/awsmachine_types.go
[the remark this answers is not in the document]
gab-satchi: The controller doesn't create the BootstrapData but it does encrypt it and store as a secret. I'm not sure how volatile that action is so I'll remove this and we can track the secret generation separately if it ends up becoming a gap

controllers/awsmachine_controller.go
gab-satchi: SGTM will update

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,50 @@
...
+limitations under the License.
+*/
+
+package v1alpha3
+
+import clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+
+const (
+	WorkerConditionCount       = 2
+	ControlPlaneConditionCount = 3
```
detiber, line 23: I don't believe these will be needed with the changes made in https://github.com/kubernetes-sigs/cluster-api/pull/3152

controllers/awsmachine_controller.go
```diff
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"sigs.k8s.io/cluster-api/util/conditions"
```
detiber, line 21: nit: should be grouped with the imports group below

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+const (
+	// InstanceReadyCondition reports on current status of the EC2 instance. Ready indicates the instance is in a Running state.
+	InstanceReadyCondition clusterv1.ConditionType = "InstanceReady"
+
+	InstanceNotFoundReason   = "InstanceNotFound"
```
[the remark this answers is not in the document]
gab-satchi: Thanks for the suggestions. I've added some new comments around the reasons and when they're likely to get used. I removed the severity from the comments after [this discussion](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1740#discussion_r435400046) 

- Added wait for cluster/bootstrap data
- For instance provision started, the controller waits until the instance is in a running state so it won't be observable on the machine object. 
- Added a reason for instance provisioning failed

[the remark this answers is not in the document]
fabriziopandini: > For instance provision started, the controller waits until the instance is in a running state so it won't be observable on the machine object.

This is unfortunate, but I prefer to get the first set of conditions merged and eventually re-iterate

controllers/awsmachine_controller.go
```diff
@@ -384,28 +412,40 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
-	// Get or create the instance.
-	instance, err := r.getOrCreate(machineScope, ec2svc, secretSvc)
+	// Find existing instance
+	instance, err := r.findInstance(machineScope, ec2svc)
 	if err != nil {
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, err.Error())
 		return ctrl.Result{}, err
 	}
+	// Create new instance
+	if instance == nil {
```
detiber, line 435: I think I found a potential pre-existing bug...  It looks like we are swallowing up errors when we attempt to do a lookup for a specific instance ID (`ec2svc.InstanceIfExists()`), when the AWS api returns a NotFound error. I'm not sure that is the behavior that we want, since:
- It introduces a potential re-entrancy bug related to the eventual consistency of the AWS APIs, where we could potentially create more than one instance for a given machine
- It could potentially lead to the replacement of an instance that was terminated and purged, when it shouldn't.

controllers/awsmachine_controller.go
```diff
@@ -384,28 +412,40 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
+	}
 
 	// Set an failure message if we couldn't find the instance.
 	if instance == nil {
 		machineScope.Info("EC2 instance cannot be found")
 		machineScope.SetFailureReason(capierrors.UpdateMachineError)
 		machineScope.SetFailureMessage(errors.New("EC2 instance cannot be found"))
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, "")
 		return ctrl.Result{}, nil
 	}
```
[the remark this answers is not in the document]
gab-satchi: yeah as long as none of the calls fail silently, this is unreachable. I left it in assuming it was must've been put in for a reason but looking through the history I can't find any. Will remove

controllers/awsmachine_controller.go
```diff
@@ -384,28 +412,40 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
 
 	ec2svc := r.getEC2Service(clusterScope)
 
-	// Get or create the instance.
-	instance, err := r.getOrCreate(machineScope, ec2svc, secretSvc)
+	// Find existing instance
+	instance, err := r.findInstance(machineScope, ec2svc)
 	if err != nil {
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, clusterv1.ConditionSeverityError, err.Error())
```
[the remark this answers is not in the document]
gab-satchi: I think unknown is more appropriate. If we find specific AWS errors that could be resolved with user intervention, False would be applicable for those cases.

api/v1alpha3/conditions_consts.go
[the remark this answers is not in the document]
gab-satchi: @fabriziopandini updated it to set a reason when provisioning has started

controllers/awsmachine_controller.go
```diff
@@ -18,17 +18,18 @@ package controllers
 
 import (
 	"context"
-
```
detiber, line 21: nit: this space should remain to separate stdlib imports from other imports, `reflect` should also be in the above grouping with `context`

controllers/awsmachine_controller.go
```diff
@@ -186,6 +214,20 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+						return true
+					}
+
+					oldMachine := e.ObjectOld.(*infrav1.AWSMachine)
+					newMachine := e.ObjectNew.(*infrav1.AWSMachine)
+
+					return !reflect.DeepEqual(oldMachine.Spec, newMachine.Spec)
+				},
+			},
+		).
```
detiber, line 230: I'm trying to think if there are any cases where we'd want to do an Update when the specs do not match...  The only case I can currently think of is to force a reconciliation when the resource is no longer paused through use of the annotation.

I don't think this would block requeues that are initiated through full resyncs, since I believe those would fall under `Create` rather than `Update`.

I'm not sure if there are other legitimate times where we would want to reconcile on Update regardless of the lack of differences in spec...

fabriziopandini: > I don't think this would block requeues that are initiated through full resyncs since I believe those would fall under Create rather than Update.

Where I should look to double-check this?

controllers/awsmachine_controller.go
```diff
@@ -186,6 +214,20 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+			predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					if e.ObjectOld.GetObjectKind().GroupVersionKind().Kind != "AWSMachine" {
+						return true
+					}
+
+					oldMachine := e.ObjectOld.(*infrav1.AWSMachine)
+					newMachine := e.ObjectNew.(*infrav1.AWSMachine)
+
+					return !reflect.DeepEqual(oldMachine.Spec, newMachine.Spec)
```
fabriziopandini, line 227: Wondering if this can be simplified/generalized by looking at `medata.ObservedGeneration`

controllers/awsmachine_controller.go
```diff
@@ -384,28 +426,44 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
...
-	instance, err := r.getOrCreate(machineScope, ec2svc, secretSvc)
+	// Find existing instance
+	instance, err := r.findInstance(machineScope, ec2svc)
 	if err != nil {
+		conditions.MarkUnknown(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, err.Error())
 		return ctrl.Result{}, err
 	}
+	// Create new instance
+	if instance == nil {
+		conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceProvisionStartedReason, clusterv1.ConditionSeverityInfo, "")
```
fabriziopandini, line 450: If there are persistent problems in `createInstance` here we will have condition flickering at every reconciliation (due to the jump from `InstanceProvisionStartedReason` to `InstanceProvisionFailedReason`).
A possible approach to avoid this problem is to wrap this call into a 
```
if conditions.GetReason(machineScope.AWSMachine, infrav1.InstanceReadyCondition) != infrav1.InstanceProvisionFailedReason {
   conditions.MarkFalse(...)
}
```
And to add a comment explaining why we are adding this if

controllers/awsmachine_controller.go
```diff
@@ -510,17 +577,7 @@ func (r *AWSMachineReconciler) deleteEncryptedBootstrapDataSecret(machineScope *
 	return nil
 }
 
-func (r *AWSMachineReconciler) getOrCreate(scope *scope.MachineScope, ec2svc services.EC2MachineInterface, secretSvc services.SecretsManagerInterface) (*infrav1.Instance, error) {
-	instance, err := r.findInstance(scope, ec2svc)
```
fabriziopandini, line 514: Might be I'm missing something, but I don't see strong reason to move findInstance in reconcileNormal vs implementing here the conditions and the extra patching logic, but I leave the final call to @detiber on this

controllers/awsmachine_controller.go
detiber: > Where I should look to double-check this?

I'm not sure of an automated way to check it, but you could spin up the controller overriding the default sync period and using `-v 4` or greater and you could verify through the predicate logging

controllers/awsmachine_controller.go
detiber: I'm not overly opinionated on where we call `findInstance` as long as we are properly guarding against eventual consistency bugs, such as the potential one I described in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1740#discussion_r440444519

controllers/awsmachine_controller.go
gab-satchi: yep it can be. From what I see, CAPA doesn't have any of the types updated to include ObservedGeneration yet

fabriziopandini:
The PR LGTM to me, but if I'm not wrong there is https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1740#discussion_r440444519 still to be addressed

controllers/awsmachine_controller.go
randomvariable: Extracted this comment to https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1769

randomvariable:
Agree with Fabrizio here.
/lgtm

vincepri:
@gab-satchi Whenever you got some time, could you resolve some of the conversations?

controllers/awsmachine_controller.go
detiber: Unless we have a driving reason to add this predicate checking, I think we should likely avoid it for now. At least until we better understand potential implications of it.

controllers/awsmachine_controller.go
detiber: Unless we have a driving reason to add this predicate checking, I think we should likely avoid it for now. At least until we better understand potential implications of it.

gab-satchi: The reason we added it is to avoid a build up of events without any spec changes that would need to be processed by the reconciler. For AWSMachine, it's just one instance where we manually patch the object. For AWSCluster, there's a few more

detiber: > The reason we added it is to avoid a build up of events without any spec changes that would need to be processed by the reconciler. For AWSMachine, it's just one instance where we manually patch the object. For AWSCluster, there's a few more

I think my worry is that we are potentially blocking actions where we would want to reconcile.

For example, if the paused annotation is added and then removed from the AWSMachine. With this predicate we would block processing of the AWSMachine when the annotation is removed (the resource is unpaused).

We would also block reconciliation if other metadata changes on the resource (which may be okay).

I mainly want to make sure that we aren't introducing any other potential behavioral changes, since we rely on reconciliation not just for trying to create/update resources based on the spec, but also for exposing the state of the backing AWS resources.

gab-satchi: We could also check for changes on annotations similar to spec but I understand the broader point of this change potentially introducing unwanted behaviour changes. 

If the predicate function is removed altogether, it may exacerbate any reentrancy bugs we may have. AWSController changes can potentially patch the object 3 times in a single reconcile loop. I wonder if that has the chance to grow exponentially with each of the reconciliation loop patching the objects again.

pkg/cloud/scope/machine.go
```diff
@@ -260,7 +260,10 @@ func (m *MachineScope) GetRawBootstrapData() ([]byte, error) {
 
 // PatchObject persists the machine spec and status.
 func (m *MachineScope) PatchObject() error {
-	return m.patchHelper.Patch(context.TODO(), m.AWSMachine)
+	return m.patchHelper.Patch(
+		context.TODO(),
+		m.AWSMachine,
+		patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{infrav1.InstanceReadyCondition, infrav1.SecurityGroupsReadyCondition, infrav1.ELBAttachedCondition}})
```
fabriziopandini, line 266: nit (not blocking)
split this line into multiple lines for better visibility

controllers/awsmachine_controller.go
```diff
@@ -186,6 +215,20 @@ func (r *AWSMachineReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					if e.ObjectOld.GetObjectKind().GroupVersionKind().Kind != "AWSMachine" {
+						return true
+					}
+
+					oldMachine := e.ObjectOld.(*infrav1.AWSMachine)
+					newMachine := e.ObjectNew.(*infrav1.AWSMachine)
+
+					return !reflect.DeepEqual(oldMachine.Spec, newMachine.Spec)
+				},
```
detiber, line 229: I believe this will work for blocking reconciliation for incremental status updates during reconciliation without affecting any other mutations to the resource.
```suggestion
				// Avoid reconciling if the event triggering the reconciliation is related to incremental status updates
				// for AWSMachine resources only
				UpdateFunc: func(e event.UpdateEvent) bool {
					if e.ObjectOld.GetObjectKind().GroupVersionKind().Kind != "AWSMachine" {
						return true
					}

					oldMachine := e.ObjectOld.(*infrav1.AWSMachine).DeepCopy()
					newMachine := e.ObjectNew.(*infrav1.AWSMachine).DeepCopy()
					
					oldMachine.Status = infrav1.AWSMachineStatus{}
					newMachine.Status = infrav1.AWSMachineStatus{} 

					return !reflect.DeepEqual(oldMachine, newMachine)
				},
```

controllers/awsmachine_controller.go
detiber: Added a suggestion that should be safer while still avoiding triggering additional reconciliations for incremental status updates

gab-satchi: I like it 👍  Thanks!

pkg/cloud/scope/machine.go
gab-satchi: Thanks. Updated

PR #1743, opened 2020-06-08
[WIP] ✨ Adds Conditions to AWSCluster

### LINKED ISSUE #1742: Add conditions to AWSCluster
As an operator, I would like to see Conditions on AWSClusters summarizing the state of cluster infrastructure so that I could see at a glance the provisioning and operating state of an AWS cluster.

During the provisioning of an AWS cluster, the following is reconciled:
- Network infrastructure 
- Bastion host (optional)
- Load balancer

Proposing to create the following Conditions and Reasons:

**Network Infrastructure**
- VpcReadyCondition
  - VpcCreationStartedReason (Info) set when attempting to create a VPC. Won't be set for unmanaged. won't be reset if condition already exists from previous reconcile
  - VpcFailedReason (Error) when VPC reconciliation fails for any reason
- SubnetReadyCondition
  -  SubnetFailedReason (Error) when reconciling subnets (and tags) fails
- InternetGatewayReadyCondition (only for managed)
  - InternetGatewayFailedReason (Error) 
- NatGatewayReadyCondition (only for managed)
  - NatGatewayCreationStartedReason (Info) set before starting creation of 1/n nat gateways. won't be reset if condition already exists from previous reconcile
  - NatGatewayReconciliationFailedReason (Error) set when any of the NGWs fail
- RouteTablesReadyCondition (only for managed)
  - RouteTableFailedReason (Error) set for any failure while reconciling route tables
- SecurityGroupsReadyCondition
  - SecurityGroupsFailedReason (Error) set for any failures with SGs

**Bastion (optional depending on setup)**
- BastionReadyCondition
  - BastionCreationStartedReason (Info) set before creating a bastion instance. won't be reset if condition already exists from previous reconcile
  - BastionFailedReason (Error) set for any failures while reconciling bastion host

**Load Balancer**
- LoadBalancerReadyCondition
  - WaitingForDNSNameReason (Info) set while cluster waits for a DNS name to be populated for the LB. won't be reset if condition already exists from previous reconcile
  - WaitForDNSNameResolve (Info) set while the cluster waits for DNS name to resolve
  - LoadBalancerFailedReason (Error) set for any errors that occur while reconciling the LB
/kind feature

gab-satchi:
This PR adds the following Conditions to AWSCluster: 
- NetworkInfrastructureReadyCondition
- BastionHostReadyCondition (only if bastion host creation was attempted)
- LoadBalancerReadyCondition

Fixes #1742 

Needs #1740 merged first
Adding hold until we have a release of CAPI to use
/hold


controllers/awscluster_controller.go
```diff
@@ -98,6 +99,12 @@ func (r *AWSClusterReconciler) Reconcile(req ctrl.Request) (_ ctrl.Result, reter
 
 	// Always close the scope when exiting this function so we can persist any AWSCluster changes.
 	defer func() {
+		if c := conditions.Get(clusterScope.AWSCluster, infrav1.BastionHostReadyCondition); c != nil {
```
detiber, line 102: Would it be better to use the spec configuration related to whether a bastion instance should be configured for this rather than trying to determine from the conditions here?

controllers/awscluster_controller.go
```diff
@@ -155,16 +162,21 @@ func reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error)
 	elbService := elb.NewService(clusterScope)
 
 	if err := ec2Service.ReconcileNetwork(); err != nil {
+		conditions.MarkFalse(awsCluster, infrav1.NetworkInfrastructureReadyCondition, infrav1.NetworkInfrastructureFailedReason, clusterv1.ConditionSeverityError, err.Error())
```
detiber, line 165: I worry that this condition might be overly broad, I'm wondering if it would make sense to be more granular here and possibly aggregate those more granular conditions using [`conditions.SetAggregate()`](https://github.com/kubernetes-sigs/cluster-api/blob/b5962688a4b44c008cd04c138a14a76ae1f2d40c/util/conditions/setter.go#L128-L134) with the [`WithConditions()`](https://github.com/kubernetes-sigs/cluster-api/blob/b5962688a4b44c008cd04c138a14a76ae1f2d40c/util/conditions/merge_strategies.go#L39-L51) MergeOption

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,69 @@
...
+const (
+	WorkerConditionCount             = 2
+	ControlPlaneConditionCount       = 3
+	ClusterConditionCount            = 2
+	ClusterConditionWithBastionCount = 3
+)
+
+// AWSCluster Conditions and Reasons
+const (
+	NetworkInfrastructureReadyCondition = "NetworkInfrastructureReady"
```
jzhoucliqr, line 30:
```suggestion
	NetworkInfrastructureReadyCondition clusterv1.ConditionType = "NetworkInfrastructureReady"
```

controllers/awsmachine_controller.go
```diff
@@ -476,8 +492,10 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		// Ensure that the security groups are correct.
 		_, err = r.ensureSecurityGroups(ec2svc, machineScope, machineScope.AWSMachine.Spec.AdditionalSecurityGroups, existingSecurityGroups)
 		if err != nil {
+			conditions.MarkFalse(machineScope.AWSMachine, infrav1.SecurityGroupsReadyCondition, infrav1.SecurityGroupsFailedReason, clusterv1.ConditionSeverityError, "")
```
jzhoucliqr, line 495:
```suggestion
			conditions.MarkFalse(machineScope.AWSMachine, infrav1.SecurityGroupsReadyCondition, infrav1.SecurityGroupsFailedReason, clusterv1.ConditionSeverityError, err.Error())
```

controllers/awscluster_controller.go
gab-satchi: yep, agree with not checking a condition that the controller itself sets. I avoided it since there were multiple reasons for it to silently skip bastion creation:
- Unmanaged VPC (updating it to check for that)
- Bastion disabled in spec (updating it to check for that)
- Missing private subnet. (This one feels like it's expected to get resolved in future reconciliations so not including this in the check)

controllers/awscluster_controller.go
gab-satchi: The SetAggregate looks for `Ready` conditions in a set of objects and updates a condition based on their values so can't use that to aggregate. I do hear the argument for more granular reasons for the infrastructure condition, but given we also populate the message with the error on failure, what's the added value in having unique reasons?

gab-satchi:
closing this for now. Will update the issue with a new set of conditions that will show incremental progress for AWSCluster

PR #1747, opened 2020-06-11
✨ Allow user defined ingress rules for CNI

### LINKED ISSUE #1744: Ability to customize CNI ports
/kind feature

While #392 requests the ability to customize security group rules, the solution for that will require rethinking how we declaratively manage multiple aspects of networking (VPC, security groups, etc). We'd like to tackle that as part of a new v1alpha4 API version.

In the meantime, it would be nice to modify v1alpha3 to allow the user to customize which ports are open for a CNI to be functional. We currently hard-code rules for BGP and IP-in-IP (for Calico). Some of our users are interested in trying Antrea, which requires one UDP port for the overlay protocol (VXLAN or Geneve, for example), and another TCP port to enable Antrea agent-to-controller communications.

To add Antrea support, we could add these rules on top of the Calico ones, but that would open more ports than required. Rather than do this, we'd like to do the following:

1. Do not make any changes to the current set of defaults (i.e., continue to configure for Calico)
1. Do not make any changes that break current CAPA clusters
1. Adjust the AWSCluster spec to allow CNI port configuration
    1. Maybe something like spec.networkSpec.cni (optional / pointer), which looks like this (names TBD):

```go
type CNISpec struct {
  IngressRules infrav1.IngressRules
}
```

4. Add defaulting webhook logic for this cni field
    1. If nil, fill in the Calico rules that are currently hard-coded in securitygroups.go
5. Remove the Calico rules from securitygroups.go
6. Modify securitygroups.go to merge the current set of hard-coded rules (SSH, etcd, etc.) with the new CNI rules from the AWSCluster spec

cc @detiber @vincepri @randomvariable - did I capture this accurately?

gab-satchi:
- Adds ability for users to define ingress rules that would be applied to worker and control plane security groups. 
- Moves adding Calico default rules into a defaulting webhook

Fixes #1744 

cc: @jayunit100


api/v1alpha2/awscluster_conversion.go
```diff
@@ -187,3 +190,8 @@ func Convert_v1alpha3_AWSLoadBalancerSpec_To_v1alpha2_AWSLoadBalancerSpec(in *in
 func Convert_v1alpha3_ClassicELBAttributes_To_v1alpha2_ClassicELBAttributes(in *infrav1alpha3.ClassicELBAttributes, out *ClassicELBAttributes, s apiconversion.Scope) error { //nolint
 	return autoConvert_v1alpha3_ClassicELBAttributes_To_v1alpha2_ClassicELBAttributes(in, out, s)
 }
+
+//Convert_v1alpha3_NetworkSpec_To_v1alpha2_NetworkSpec
+func Convert_v1alpha3_NetworkSpec_To_v1alpha2_NetworkSpec(in *infrav1alpha3.NetworkSpec, out *NetworkSpec, s apiconversion.Scope) error {
```
ncdc, line 195: I don't think this is needed here, as it's part of the autogenerated bits?

api/v1alpha2/awscluster_conversion.go
```diff
@@ -59,6 +59,9 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 	if restored.Spec.ControlPlaneLoadBalancer != nil {
 		dst.Spec.ControlPlaneLoadBalancer = restored.Spec.ControlPlaneLoadBalancer
 	}
+	//Manually convert CNI
```
ncdc, line 62: Given we don't have this comment for the other fields, I would remove this.

api/v1alpha2/awscluster_conversion.go
```diff
@@ -59,6 +59,9 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
...
+	dst.Spec.NetworkSpec.CNI = restored.Spec.NetworkSpec.CNI
+
```
ncdc, line 64:
```suggestion
```

api/v1alpha3/types.go
```diff
@@ -179,6 +179,10 @@ type NetworkSpec struct {
 	// Subnets configuration.
 	// +optional
 	Subnets Subnets `json:"subnets,omitempty"`
+
+	// CNI configuration
+	// +optional
+	CNI CNISpec `json:"cni,omitempty"`
```
ncdc, line 185: This needs to be a pointer

api/v1alpha3/awscluster_webhook.go
```diff
@@ -80,3 +82,25 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 
 	return aggregateObjErrors(r.GroupVersionKind().GroupKind(), r.Name, allErrs)
 }
+
+func (r *AWSCluster) Default() {
+	// Default to Calico ingress rules if no rules have been set
+	if reflect.DeepEqual(r.Spec.NetworkSpec.CNI, CNISpec{}) {
```
ncdc, line 88: Once you make CNI a pointer, I think you should be able to check for nil.

I also think if we have a non-nil CNI field that has 0 IngressRules, we should **not** install the default Calico rules. WDYT @detiber @vincepri @randomvariable?

api/v1alpha3/types.go
```diff
@@ -294,6 +298,13 @@ func (s Subnets) FilterByZone(zone string) (res Subnets) {
 	return
 }
 
+// CNISpec defines configuration for CNI
+type CNISpec struct {
+	// IngressRules specify rules to apply to control plane and worker node security groups
+	// The source for the rule will be set to control plane and worker security group IDs
```
ncdc, line 304: Godoc/OpenAPI docs don't respect the whitespace. I recommend ending sentences with periods. 😄

api/v1alpha3/types.go
```diff
@@ -294,6 +298,13 @@ func (s Subnets) FilterByZone(zone string) (res Subnets) {
...
+	IngressRules IngressRules `json:"ingressRules,omitempty"`
```
ncdc, line 305: Do we want omitempty here?

api/v1alpha3/awscluster_webhook.go
```diff
@@ -39,8 +39,10 @@ func (r *AWSCluster) SetupWebhookWithManager(mgr ctrl.Manager) error {
 }
 
 // +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1alpha3-awscluster,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=awsclusters,versions=v1alpha3,name=validation.awscluster.infrastructure.cluster.x-k8s.io,sideEffects=None
+// +kubebuilder:webhook:verbs=create;update,path=/mutate-infrastructure-cluster-x-k8s-io-v1alpha3-awscluster,mutating=true,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=awsclusters,versions=v1alpha3,name=validation.awscluster.infrastructure.cluster.x-k8s.io,sideEffects=None
```
ncdc, line 42: Did you use `kubebuilder create webhook` for this? The name looks wrong (it still has `validation`). Also I think `sideEffects` shouldn't be `None`? cc @vincepri

api/v1alpha3/awscluster_webhook_test.go
```diff
@@ -107,3 +108,48 @@ func TestAWSCluster_ValidateUpdate(t *testing.T) {
...
+									Protocol:    SecurityGroupProtocolIPinIP,
+									FromPort:    -1,
+									ToPort:      65535,
+								},
+							},
+						},
+					},
+				},
+			},
+		},
```
ncdc, line 146: Please add a test to validate that we don't add Calico when we have other ingress rules defined.

api/v1alpha3/awscluster_webhook.go
randomvariable: Yes, that'd make sense to me.

vincepri: +1

api/v1alpha3/awscluster_webhook.go
vincepri: Side effect should be None, see https://github.com/kubernetes-sigs/cluster-api/issues/2851

api/v1alpha3/types.go
vincepri: Do we want to require at least 1 rule? To avoid things like `cni: {}`

api/v1alpha3/awscluster_webhook.go
gab-satchi: will update name. thanks

api/v1alpha3/types.go
vincepri: Also just to note, the `IngressRules` has other fields as well which we mentioned they wouldn't be exposed, like `sourceSecurityGroupIds`

ncdc: 0 vs 1+ rule - I'm 🤷‍♂️. Maybe someone would want to manage CNI rules with additionalSecurityGroups instead?

We should probably define our own scoped-down struct specifically for this (to exclude sourceSecurityGroupIds)?

vincepri: If they want to use a different mechanism, a user can set (or never set) `cni: nil`. If we allow for empty ingress rules, the spec would look something like `cni: { ingressRules: [] }` which seems odd

api/v1alpha3/awscluster_webhook.go
gab-satchi: IngressRules is currently the only field for the CNISpec. Would the existence of a non nil CNI spec just mean that the cluster is misconfigured?

api/v1alpha3/types.go
ncdc: Good point. Let's require at least 1

api/v1alpha3/awscluster_webhook.go
ncdc: Ignore my comment about non-nil with 0 rules (see other comment thread). If you have a non-nil CNI field, it means you are configuring at least 1 rule. So nil CNI = don't configure anything. non-nil CNI = configure at least 1 rule.

api/v1alpha2/awscluster_conversion.go
gab-satchi: The autogenerated code was [removing](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1747/files#diff-94d92e5ec55d2db907b66f93901cc2e2L212-L216) the Convert function so added it here since I saw some other fields here as well.

api/v1alpha3/types.go
gab-satchi: Updated

api/v1alpha3/types.go
randomvariable: Just checked the VPC CNI plugin. This would work through additional security groups and no rules in the CNI. I think we should therefore allow the zero length slice and respect it.

ncdc: Thanks for checking @randomvariable! How do additional security groups tie in here?

randomvariable: By default, if there's SGs attached to the primary interface, the VPC CNI will consume them. Though, as there's a chance that they'll start conflicting with CAPI in managing the rules, you would probably create new security groups, use `AdditionalSecurityGroups` in the AWSMachineTemplate, and use the VPC CNI plugin's custom ENIConfig CRD to force it to use that SG. The VPC CNI plugin manages the rules by itself, so no extra rules are needed.

ncdc: Got it, thanks!

@gab-satchi could you please modify your PR to allow a non-nil CNI field with 0 rules (and make sure it does not default Calico in this scenario)?

jayunit100:
im going to test this now will let you know if it works thanks gab

jayunit100:
it works !  FWIW , these are the ports that I added .  Am currently running conformance tests against this cluster, but so far it looks happy. 

```
 cniIngressRules:
 - description: antrea1
   fromPort: 10349
   protocol: tcp
   toPort: 10349
 - description: antrea2
   fromPort: 4789
   protocol: udp
   toPort: 4789
```

jayunit100:
can we rebase and merge this :)

PR #1749, opened 2020-06-17
Bump cluster creation timeout for conformance in multi-AZ

randomvariable:

Need to wait longer for cluster up time because of #1721 .
Witnessed an observed time of 547s, but also seen a failure on a 600s timeout, so setting at 700s here.

Fixes #


detiber:
/lgtm
/approve

I don't remember which region we use by default, but we could potentially also hit limits if running in us-east-1 and we are trying to create nat gateways for all available AZs by default

randomvariable:
Defaults to using the first 3 sorted AZs rather than all, so hoping that's fine.

PR #1750, opened 2020-06-17
✨ 1713 multitenancy (WIP)

andrewmyhre:
**What this PR does / why we need it**: Enables CAPA to use multiple AWS principals

Addresses #1713 

This is not complete - posting for visibility. We have tested using static principals and will be testing for IAM role assumption soon. Service account usage is not implemented yet.


andrewmyhre:
Closing in favor of #1753

Issue #1715, opened 2020-05-06
Allow Specifying Internal Subnets for Control Plane 


voor:
/kind feature

As a `controlPlaneLoadBalancer` with a `scheme` set to `internal` I would like to be able to specify which subnets that load balancer goes into, because there are security controls expecting it to go into certain subnets.

**Describe the solution you'd like**
Related to https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1714 it's currently just looping through the internal subnets, I'd like to be able to map it intentionally.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

bagnaram:
I believe in unmanaged mode you can specify the subnets intentionally in the `awscluster.spec.networkSpec.subnets`

voor:
@bagnaram That might fit the use case -- are you saying that if you specify the VPC ID you can specify a subset of subnets to use?  @scottslowe made it seem like if you specify a VPC ID it'll ignore the subnets and populate it automatically.

scottslowe:
@voor @bagnaram It is my understanding (based on conversations I had while creating the content for #1560) that a) the list of subnets isn't required when operating in unmanaged mode, *and* b) you can't "constrain" CAPA to use only a subset of the subnets by providing the list (it will use all discovered subnets regardless of what you provide).

voor:
Just want to confirm that specifying the VPC ID will overwrite and otherwise ignore whatever is in the subnet section.  If you try to set something as `isPublic` for example, it'll just overwrite it with the actual state of the world.

scottslowe:
@voor That is my understanding, yes, although I have not specifically tested the behavior.

randomvariable:
Fixed in #1931

PR #1756, opened 2020-06-19
✨ Bring your own control plane subnets

andrewmyhre:
**What this PR does / why we need it**: Presently it is not possible to specify which subnets the control plane should use. CAPA will discover and use all available subnets in the VPC. Some organizations have specific subnets designated for particular traffic e.g: service vs workloads need to be able to specify these in the cluster spec. 

Addresses #1715

Also related: #1726


api/v1alpha3/awscluster_types.go
```diff
@@ -107,6 +107,10 @@ type AWSLoadBalancerSpec struct {
 	// Defaults to false.
 	// +optional
 	CrossZoneLoadBalancing bool `json:"crossZoneLoadBalancing,omitempty"`
+
+	// Subnets specifies the subnets that should be used by the load balancer
+	// +optional
+	Subnets Subnets `json:"subnets,omitempty"`
```
randomvariable, line 113: We should scope this down to the bits that we're actually using, otherwise the API is going to be problematic, as it includes references to NAT gateways and public and/or private subnets. Copying the types to be more local to the task in hand is fine.

voor:
@andrewmyhre any idea how these changes are going?

api/v1alpha3/awscluster_types.go
voor: It might make more sense to just add the same `subnetIds` array that's in status up into the spec section.

gab-satchi: Since we're adding ability to specify subnets with filters with #1776, can make it a slice of AWSResourceReference

ncdc:
@andrewmyhre is this something you're still interested in working on? If not, would you like someone else from the community to take it on?

andrewmyhre:
@ncdc I have the bandwidth to pick this up again.

andrewmyhre:
Closing in favor of #1931 as this is very out of date

PR #1758, opened 2020-06-19
🐛 Check before (de)registering with ELB

### LINKED ISSUE #1751: Single cluster can hit rate limiting by continuously attempting EC2 instance registration against ELB
/kind bug
/help

Running e2e tests
See many `SuccessfulAttachControlPlaneELB` events, interspersed with `FailedAttachControlPlaneELB`
```
	message: Failed to register control plane instance "i-03a4545af5c5855e0" with load balancer: failed to describe classic load balancer "votta5hv10rk9t4lssguwjsa7o0r-k8s" attributes: Throttling: Rate exceeded
	status code: 400, request id: 714e1627-5fb8-4725-aed7-9a098d6a78a2
```

Completes successfully, with no rate limiting

`elbsvc.RegisterInstanceWithAPIServerELB(i)` does not check first if the instance is registered. Although this is not an issue operationally, it counts as a write request against the ELB. There should be a check that it is already registered prior to attempting again.

**Environment:**

- Cluster-api-provider-aws version: v0.5.4

ncdc:
Check to see if an instance is already registered with the apiserver ELB
before adding it.

Check to see if an instance is already deregistered from the apiserver
ELB before removing it.

Fixes #1751

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -575,6 +575,17 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 	// In order to prevent sending request to a "not-ready" control plane machines, it is required to remove the machine
 	// from the ELB as soon as the machine gets deleted or when the machine is in a not running state.
 	if !machineScope.AWSMachine.DeletionTimestamp.IsZero() || !machineScope.InstanceIsRunning() {
+		registered, err := elbsvc.InstanceIsRegisteredWithAPIServerELB(i)
+		if err != nil {
+			r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedDetachControlPlaneELB",
+				"Failed to deregister control plane instance %q from load balancer: failed to determine registration status: %v", i.ID, err)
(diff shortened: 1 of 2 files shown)


controllers/awsmachine_controller.go
```diff
@@ -575,6 +575,17 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 	// In order to prevent sending request to a "not-ready" control plane machines, it is required to remove the machine
 	// from the ELB as soon as the machine gets deleted or when the machine is in a not running state.
 	if !machineScope.AWSMachine.DeletionTimestamp.IsZero() || !machineScope.InstanceIsRunning() {
+		registered, err := elbsvc.InstanceIsRegisteredWithAPIServerELB(i)
```
ncdc, line 578: @vincepri I know we discussed moving this up and only calling it once, but I didn't (quickly) see an easy way to generate the right event based on the appropriate action (attach vs detach). Happy to move it around if you have suggestions.

vincepri: Makes sense, thanks for following up!

PR #1759, opened 2020-06-21
create a bastion host if it's enabled in the spec, regardless of managed or unmanaged VPC

### LINKED ISSUE #1748: Support bastion with unmanaged (bring your own) VPC
/kind feature

If you are using an unmanaged (bring your own) VPC, we do not create or reconcile a bastion host, but we do reconcile the security group for the bastion. We recently added the ability to enable/disable the bastion host. It would be nice if we could create a bastion host if it's enabled in the spec, regardless of managed or unmanaged VPC.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d4ade5922e04a2327f4043effd16252126dea223/pkg/cloud/services/ec2/bastion.go#L39-L42 is where we skip working on the bastion if the VPC is unmanaged.

**Environment:**

- Cluster-api-provider-aws version: v0.5.4

evalsocket:
Presently if you are using an unmanaged (bring your own) VPC, we do not create or reconcile a bastion host, but we do reconcile the security group for the bastion.Now  we could create a bastion host if it's enabled in the spec, regardless of managed or unmanaged VPC.

Fixes #1748

### DIFF
--- a/pkg/cloud/services/ec2/bastion.go
+++ b/pkg/cloud/services/ec2/bastion.go
@@ -36,12 +36,9 @@ const (
 
 // ReconcileBastion ensures a bastion is created for the cluster
 func (s *Service) ReconcileBastion() error {
-	if s.scope.VPC().IsUnmanaged(s.scope.Name()) {
-		s.scope.V(4).Info("Skipping bastion reconcile in unmanaged mode")
-		return nil
-	}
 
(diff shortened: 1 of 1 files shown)


ncdc:
/ok-to-test

Thanks @evalsocket!!!

PR #1761, opened 2020-06-23
✨ allow configuration of Bastion host's ingress rules with cidrBlocks field

### LINKED ISSUE #1473: Allow configuration of allowed CIDR in ingress rule
/kind feature

It seems to be hardcoded at the moment

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/75271f7b51f4a1e358792754a1c890b167d62460/pkg/cloud/services/ec2/securitygroups.go#L370

Would be nice to make it configurable.

None

**Environment:**

Not relevant

aaroniscode:
This adds a field `allowedCIDRBlocks` to the Bastion struct in `AWSCluster`. 

```
bastion:
  enabled: true
  allowedCIDRBlocks:
  - 192.168.0.10/32
```

I tested a number of configurations:
* Omiting `cidrBlocks` (nil) results in same behavior as today with `0.0.0.0/0` as allowed ingress.
* ~~Including an empty array `[]` results in no rules~~ (edit[ncdc]: we've added `disableIngressRules` for this)
* Including single or multiple CIDR's results in those rules added to the bastion security group.

~~I left out `omitEmpty` so anyone using the types as a library can distinguish between `nil` and `[]`.~~ (edit[ncdc]: we've added `disableIngressRules` for this)

I read through all the notes in the issue, but want to confirm I covered all use cases. Did I forget anything? Or does this implementation have any unintended consequences?  It was pretty straightforward.

Fixes #1473


api/v1alpha3/awscluster_webhook.go
```diff
@@ -84,6 +84,11 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 }
 
 func (r *AWSCluster) Default() {
+	// Default to allow open access to the bastion host if no CIDR Blocks have been set
+	if r.Spec.Bastion.CIDRBlocks == nil {
+		r.Spec.Bastion.CIDRBlocks = []string{"0.0.0.0/0"}
+	}
```
randomvariable, line 90: Might be worth adding validation of the CIDRs to the webhook as well.

aaroniscode: Sure!  I can add that!

aaroniscode:
So ... going through a bunch of test errors. Should have run `make test` BEFORE I opened the PR. 😃

The AWSCluster Reconciler test is failing when creating a k8s client. The error is:

```
Status: "Failure",
Message: "AWSCluster.infrastructure.cluster.x-k8s.io \"foo\" is invalid: spec.bastion.cidrBlocks: Invalid value: \"null\": spec.bastion.cidrBlocks in body must be of type array: \"null\"",
Reason: "Invalid",
Details: {
    Name: "foo",
    Group: "infrastructure.cluster.x-k8s.io",
    Kind: "AWSCluster",
    UID: "",
    Causes: [
        {
            Type: "FieldValueInvalid",
            Message: "Invalid value: \"null\": spec.bastion.cidrBlocks in body must be of type array: \"null\"",
            Field: "spec.bastion.cidrBlocks",
        },
    ],
    RetryAfterSeconds: 0,
}
```

Not sure how to resolve this. `nil` is the default value for the String Slice in Go. And it clearly works as I've tested. Any advice on this?

Looking into the conversion errors as well.

randomvariable:
I think you may also want to call the defaulter inside reconciliation to handle existing non-modified resources but with an upgraded CAPA controller.

ncdc:
I played around with this briefly, adding `omitempty` just to see what happen. It got past the create issue, and failed in the fuzzing tests. I ran out of time to dig into that failure, but you might want to work on that and we can continue to think about what to do with `omitempty`.

aaroniscode:
Thanks @ncdc !  I fixed 2 of the 3 failing tests late last night. Working on the last one. As soon as I get the last one fixed I'll force push an update.  The failing fuzzing tests I believe was due to a mistake on my part as I didn't understand how conversion worked. On the conversion from v1alpha2 to v1alpha3, I was able to take the CIDRBlocks in the restored data (I think this is unmarshalled from annotations) and had to manually add it back. I thought it was done automatically but it appears if you want the restored data you have to take the step yourself to apply it.

api/v1alpha3/awscluster_webhook.go
aaroniscode: added validation of the CIDR blocks. I used regex to validate. Let me know what you think

api/v1alpha2/awscluster_conversion.go
```diff
@@ -53,6 +53,14 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error { // nolint
 		dst.Spec.SSHKeyName = nil
 	}
 
+	// 	Preserve the pre-v1alpha3 behavior of allowing open access to the Bastion host if this is an upgrade
+	if restored.Spec.Bastion.CIDRBlocks == nil {
+		dst.Spec.Bastion.CIDRBlocks = []string{"0.0.0.0/0"}
```
ncdc, line 58: If the restored value is nil, it means the user's intent was no CIDRs. (Restored is the v1alpha3 version.)

aaroniscode:
> I think you may also want to call the defaulter inside reconciliation to handle existing non-modified resources but with an upgraded CAPA controller.

@randomvariable I did some thinking about this and whether it was the right way to handle upgrades and ended up with a different solution, but really need validation from the team if it's correct.

Here's my thinking. Right now in v1alpha3, we default Bastion rules to `0.0.0.0/0`. Say in v1alpha4 we decided to default to no rules `[]` because AWS really discourages against SSH open to the world. This would allow users to add rules for their individual IP's or their own ranges after creation.  If we change the default, we wouldn't want to call the default in the reconciliation loop because it would break the v1alpha2 behavior.

So I added `0.0.0.0/0` to the CIDR rules in the conversion from v1alpha2 to v1alpha3 if the annotation isn't present which I think should cover an upgrade scenario. And it will allow us to change the default in the future without breaking backward compatitibility.

api/v1alpha2/awscluster_conversion.go
aaroniscode: @ncdc My thought here is that if we upgrade the controller in a v1alpha2 cluster to v1alpha3, the nil value needs to be replaced because it will crash the API server. I don't believe the Default() function will be called. I could be wrong.

And I thought we wanted to use `[]` to mean no rules, and `nil` was for the default. Please correct me if I'm wrong.

aaroniscode: To add more detail ... and to correct my thought process if I don't understand the conversion process:

We need to support roundtrip between v1alpha3 and v1alpha2. So in this conversion to v1alpha3, IF the annotation is there, it's been roundtripped and we want to restore the CIDRBlocks. If the annotation is not there and the value is `nil` (not 100% sure that is what happens so please correct me if I'm wrong), we are moving from v1alpha2 to v1alpha3 in an upgrade and we want to preserve the v1alpha2 behavior.

That is my thought process above for the if/else.

ncdc: Sorry, multi-tasking bit me again 😄. You are right that `nil` means keep the current 0/0 default.

`restored` is the v1alpha3 copy of the resource stored as json in an annotation. Given that the defaulting logic you're adding in this PR sets it to 0/0 if you pass in `nil`, I don't see it being possible for `restored` to have a `nil` value here.

ncdc: > If the annotation is not there and the value is nil (not 100% sure that is what happens so please correct me if I'm wrong), we are moving from v1alpha2 to v1alpha3 in an upgrade and we want to preserve the v1alpha2 behavior

This is definitely a confusing area 😄. If the annotation is not there, then `ok` on line 47 will be false, and we'll return early on line 48.

aaroniscode: @ncdc Thanks for the Zoom and clearing up how the conversion works. So I've removed the if/else which is not correct and left the assignment of CIDRBlocks for when we roundtrip and the we want the restored data from the annotation.

api/v1alpha3/awscluster_types.go
```diff
@@ -89,6 +89,11 @@ type Bastion struct {
 	// with a public ip to access the VPC private network.
 	// +optional
 	Enabled bool `json:"enabled"`
+
+	// CIDRBlocks is a list of CIDR blocks allowed to access the bastion host.
+	// They are set as ingress rules for the Bastion host's Security Group (defaults to 0.0.0.0/0).
+	// +optional
+	CIDRBlocks []string `json:"cidrBlocks"`
```
ncdc, line 96: This will need to be `omitempty` and we will not be able to distinguish between nil and [] (SIG API Machinery advises against trying to distinguish). We'll likely want to use a separate bool field to disable ingress.

aaroniscode: ok, so if the value is empty we default to `0.0.0.0/0` and will use a boolean as an option to have to rules.

aaroniscode:
Well, I learned I really need to run `make lint` as well. Learned something new about misaligned structs.

```
api/v1alpha3/awscluster_types.go:87:14: struct of size 40 bytes could be of size 32 bytes (maligned)
```

So I moved the boolean `NoIngressRules` in front of `CIDRBlocks` to resolve the struct alignment issue.

api/v1alpha3/awscluster_types.go
aaroniscode: @ncdc Added a new boolean field called `NoIngressRules`.  I couldn't think of a better name. 😀

Let me know if you want something different. Also, I wasn't sure if we want `omitempty` for the boolean. So just want to confirm that.

api/v1alpha3/awscluster_types.go
```diff
@@ -89,6 +89,16 @@ type Bastion struct {
 	// with a public ip to access the VPC private network.
 	// +optional
 	Enabled bool `json:"enabled"`
+
+	// NoIngressRules will ensure no Ingress rules in the bastion host's security group.
+	// Requires CIDRBlocks to be empty.
+	// +optional
+	NoIngressRules bool `json:"noIngressRules,omitempty"`
```
ncdc, line 96: I'd probably prefer one of:
- DisableIngressRules
- DisallowIngress
- DisableIngress

@detiber I believe you previously mentioned that even if bastion.disabled is true, you thought it's still worthwhile to create the bastion security group, as users might want to BYO bastion VM, and the security group allows SSH access from the bastion's security group to the control plane & node security groups. If people are doing BYO bastion, do you think they'd be using CAPA's bastion security group for it, or something else (along with additionalSecurityGroups for the control plane and node machines)?

cc @randomvariable @rudoi @sethp-nr

randomvariable: Personally think they'll be bringing their own security groups, or the bastion is more likely not even in the same VPC and is in a management one of some sort.

ncdc: Right, that's mostly what I was thinking. I would love to be able to use `Enabled` to control both the creation of the bastion instance and the security group. But that is a change in behavior, and would require manual configuration + additionalSecurityGroups to replicate.

aaroniscode: @ncdc I updated the boolean name to `DisableIngressRules`

api/v1alpha3/awscluster_webhook.go
```diff
@@ -80,10 +85,38 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
...
+	var errs field.ErrorList
+
+	if r.Spec.Bastion.DisableIngressRules && len(r.Spec.Bastion.CIDRBlocks) > 0 {
+		errs = append(errs,
+			field.Forbidden(field.NewPath("spec", "Bastion", "CIDRBlocks"), "cannot be set if spec.Bastion.DisableIngressRules is true"),
+		)
+		return errs
+	}
+
+	re, _ := regexp.Compile(`^([0-9]{1,3}\.){3}[0-9]{1,3}(\/([0-9]|[1-2][0-9]|3[0-2]))?$`)
```
detiber, line 103: Rather than using a regex here it might be better to use [net.ParseCIDR](https://golang.org/pkg/net/#ParseCIDR) instead, which would also make it easier to support ipv6 in the future as well.

aaroniscode: thanks @detiber ! I wasn't aware of `net.ParseCIDR` and will update the validation to take advantage of it.

api/v1alpha3/awscluster_types.go
```diff
@@ -89,6 +89,16 @@ type Bastion struct {
...
+
+	// DisableIngressRules will ensure there are no Ingress rules in the bastion host's security group.
+	// Requires CIDRBlocks to be empty.
+	// +optional
+	DisableIngressRules bool `json:"disableIngressRules,omitempty"`
+
+	// CIDRBlocks is a list of CIDR blocks allowed to access the bastion host.
+	// They are set as ingress rules for the Bastion host's Security Group (defaults to 0.0.0.0/0).
+	// +optional
+	CIDRBlocks []string `json:"cidrBlocks,omitempty"`
```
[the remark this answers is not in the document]
aaroniscode: this sounds like a good idea. Happy to update.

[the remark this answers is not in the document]
ncdc: +1, thanks for the suggestion @detiber

api/v1alpha3/awscluster_webhook.go
```diff
@@ -80,10 +85,38 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
...
+
 	return aggregateObjErrors(r.GroupVersionKind().GroupKind(), r.Name, allErrs)
 }
 
+func (r *AWSCluster) validateCIDRBlocks() []*field.Error {
+	var errs field.ErrorList
+
+	if r.Spec.Bastion.DisableIngressRules && len(r.Spec.Bastion.CIDRBlocks) > 0 {
+		errs = append(errs,
+			field.Forbidden(field.NewPath("spec", "Bastion", "CIDRBlocks"), "cannot be set if spec.Bastion.DisableIngressRules is true"),
```
ncdc, line 98: Field path segments need to be lowercase

aaroniscode: thanks for catching @ncdc ... resolved.

api/v1alpha3/awscluster_webhook.go
aaroniscode: @detiber switched to `net.ParseCIDR`

api/v1alpha3/awscluster_types.go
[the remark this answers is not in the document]
aaroniscode: updated to `AllowedCIDRBlocks`

aaroniscode:
Looks like we have a timeout in the test as it passes locally for me.

api/v1alpha3/awscluster_webhook.go
```diff
@@ -80,10 +85,37 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
...
+	var errs field.ErrorList
+
+	if r.Spec.Bastion.DisableIngressRules && len(r.Spec.Bastion.AllowedCIDRBlocks) > 0 {
+		errs = append(errs,
+			field.Forbidden(field.NewPath("spec", "bastion", "allowedCIDRBlocks"), "cannot be set if spec.bastion.disableIngressRules is true"),
+		)
+		return errs
+	}
+
+	for _, cidr := range r.Spec.Bastion.AllowedCIDRBlocks {
```
ncdc, line 103:
```suggestion
	for i, cidr := range r.Spec.Bastion.AllowedCIDRBlocks {
```

api/v1alpha3/awscluster_webhook.go
```diff
@@ -80,10 +85,37 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
...
+		if _, _, err := net.ParseCIDR(cidr); err != nil {
+			errs = append(errs,
+				field.Invalid(field.NewPath("spec", "bastion", "allowedCIDRBlocks"), cidr, "must be a valid CIDR block"),
```
ncdc, line 106:
```suggestion
				field.Invalid(field.NewPath("spec", "bastion", fmt.Sprintf("allowedCIDRBlocks[%d]", i)), cidr, "must be a valid CIDR block"),
```

I think we're going to want the array index included in the field path

api/v1alpha3/awscluster_webhook.go
aaroniscode: done

api/v1alpha3/awscluster_webhook.go
aaroniscode: thanks @ncdc !  I didn't realize the Field Path could or should include the array index. Updated!

ncdc:
@aaroniscode can you add unit tests for the new validation logic?

aaroniscode:
> @aaroniscode can you add unit tests for the new validation logic?

@ncdc added unit tests for both validating the CIDRs and the new defaulting logic.

ncdc:
/lgtm
/approve

Thanks!

PR #1763, opened 2020-06-23
:bug: clusterawsadm: ensure role policies work

randomvariable:
Attempting to add extra statements for a real use case led to the CloudFormation template failing because it was missing the policy name. I had also accidentally appended these statements to the managed policy, which is not the documented intent.

This also adds an "AllowedEC2InstanceProfiles" field to be set for the controller to override which EC2 IAM roles can be used by Cluster API.

Fixes #

Advice on testing strategies. Should I create a test directory with "correct" CloudFormation YAML's to compare the generator to?


randomvariable:
tests added
/unhold

randomvariable:
urgh typos
/hold

cmd/clusterawsadm/cloudformation/bootstrap/template_test.go
```diff
@@ -0,0 +1,115 @@
...
+
+		tData, err := c.template().RenderCloudFormation().YAML()
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if string(tData) != string(data) {
+			dmp := diffmatchpatch.New()
+			diffs := dmp.DiffMain(string(tData), string(data), false)
+			// out := string(tData)
```
detiber, line 110: Looks like left over from earlier iteration
```suggestion
```

PR #1767, opened 2020-06-24
🐛 Preserve all VPCSpec fields

### LINKED ISSUE #1757: Constant attempts to patch AWSCluster to set AvailabilityZoneUsageLimit and AvailabilityZoneSelection to null
/kind bug

1. Create a normal cluster
1. Run `kubectl get awscluster -w`
1. You'll see that it is updated about once a second, but the only things that change are generation and resourceVersion

1. No constant updates

1. @vincepri and I enabled audit logging and discovered events like this:

```json
{
  "kind": "Event",
  "apiVersion": "audit.k8s.io/v1",
  "level": "Request",
  "auditID": "7cb7fb43-b02d-4495-86b6-5ea1a4e3a3f1",
  "stage": "ResponseComplete",
  "requestURI": "/apis/infrastructure.cluster.x-k8s.io/v1alpha3/namespaces/default/awsclusters/andy",
  "verb": "patch",
  "user": {
    "username": "system:serviceaccount:capa-system:default",
    "uid": "7f286c55-d41c-40da-8cfe-a4d48b31e2fb",
    "groups": [
      "system:serviceaccounts",
      "system:serviceaccounts:capa-system",
      "system:authenticated"
    ]
  },
  "sourceIPs": [
    "10.244.0.18"
  ],
  "userAgent": "manager/v0.0.0 (linux/amd64) kubernetes/$Format",
  "objectRef": {
    "resource": "awsclusters",
    "namespace": "default",
    "name": "andy",
    "apiGroup": "infrastructure.cluster.x-k8s.io",
    "apiVersion": "v1alpha3"
  },
  "responseStatus": {
    "metadata": {},
    "code": 200
  },
  "requestObject": {
    "spec": {
      "networkSpec": {
        "vpc": {
          "availabilityZoneSelection": null,
          "availabilityZoneUsageLimit": null
        }
      }
    }
  },
  "requestReceivedTimestamp": "2020-06-19T20:56:49.541849Z",
  "stageTimestamp": "2020-06-19T20:56:49.562848Z",
  "annotations": {
    "authorization.k8s.io/decision": "allow",
    "authorization.k8s.io/reason": "RBAC: allowed by ClusterRoleBinding \"capa-manager-rolebinding\" of ClusterRole \"capa-manager-role\" to ServiceAccount \"default/capa-system\"",
    "mutation.webhook.admission.k8s.io/round_0_index_0": "{\"configuration\":\"capa-mutating-webhook-configuration\",\"webhook\":\"default.awscluster.infrastructure.cluster.x-k8s.io\",\"mutated\":false}"
  }
}
```

We are constantly trying to patch availabilityZoneSelection and availabilityZoneUsageLimit to null.

This is because when we create and describe the VPC in AWS, we instantiate a new `VPCSpec`:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c718d8e8aaf2b5634cf7b233834818ff07d9d613/pkg/cloud/services/ec2/vpc.go#L195-L199

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c718d8e8aaf2b5634cf7b233834818ff07d9d613/pkg/cloud/services/ec2/vpc.go#L264-L268

And then we copy this newly-created `VPCSpec` into the AWSCluster, overwriting the existing values retrieved from the apiserver:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c718d8e8aaf2b5634cf7b233834818ff07d9d613/pkg/cloud/services/ec2/vpc.go#L86

Because of this, we are losing the values of availabilityZoneSelection and availabilityZoneUsageLimit, so the patch helper tries to clear them out during every reconcile. These 2 fields have OpenAPI defaulting, which prevents their values from being cleared out, but this flow results in an infinite loop of patching attempts.

In the long term, we probably should update our logic so instead of creating entirely new VPCSpecs and overwriting the existing field in the AWSCluster, we copy the described values into AWSCluster.Spec.NetworkSpec.VPC.

**Environment:**

- Cluster-api-provider-aws version: commit c718d8e8

ncdc:
When creating/describing the VPC from AWS, we were creating a new
VPCSpec and then using that for the actual VPCSpec in the AWSCluster,
unintentionally wiping out a couple of newly-added fields to VPCSpec
that are user-intent and not information that comes from AWS. This is a
temporary band-aid that ensures we do not lose these fields during
reconciliation. Without this change, the controller sends a patch to set
AvailabilityZoneSelection and AvailabilityZoneUsageLimit to null, but
because of defaulting, the null patches are essentially
rejected/overwritten by the mutating webhook. The result is that the
AWSCluster is reconciled continuously, over and over. This change fixes
that issue.

Fixes #1757

### DIFF
--- a/pkg/cloud/services/ec2/vpc.go
+++ b/pkg/cloud/services/ec2/vpc.go
@@ -53,6 +53,19 @@ func (s *Service) reconcileVPC() error {
 		return errors.Wrap(err, "failed to describe VPCs")
 	}
 
+	// This function creates a new infrav1.VPCSpec, populates it with data from AWS, and then deep copies into the
+	// AWSCluster's VPC spec (see the DeepCopyInto lines below). This is potentially problematic, as it completely
+	// overwrites the data for the VPC spec as retrieved from the apiserver. This is a temporary band-aid to restore
+	// recently-added fields that descripe user intent and do not come from AWS resource descriptions.
+	//
+	// FIXME(ncdc): rather than copying these values from the scope to vpc, find a better way to merge AWS information
+	// with data in the scope retrieved from the apiserver. Could use something like mergo.
+	//
+	// NOTE: it may look like we are losing InternetGatewayID because it's not populated by describeVPC/createVPC or
+	// restored here, but that's ok. It is restored by reconcileInternetGateways, which is invoked after this.
+	vpc.AvailabilityZoneSelection = s.scope.VPC().AvailabilityZoneSelection
+	vpc.AvailabilityZoneUsageLimit = s.scope.VPC().AvailabilityZoneUsageLimit
+
 	if vpc.IsUnmanaged(s.scope.Name()) {
 		vpc.DeepCopyInto(s.scope.VPC())
 		s.scope.V(2).Info("Working on unmanaged VPC", "vpc-id", vpc.ID)
--- a/pkg/cloud/services/ec2/vpc_test.go
+++ b/pkg/cloud/services/ec2/vpc_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/golang/mock/gomock"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/utils/diff"
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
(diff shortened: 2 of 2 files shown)


ncdc:
Hoping to find a way to add unit tests for this

ncdc:
I realized the unit test for this was not actually checking if there were errors - I have updated it to do so, and adjusted the tests to ensure we're not losing these 2 fields.

PR #1777, opened 2020-06-30
✨ Adds cluster conditions

### LINKED ISSUE #1742: Add conditions to AWSCluster

gab-satchi:
Adds conditions to clusters to show incremental progress along with feedback when reconciliation runs into errors. 

Fixes #1742 

/hold while we get #1740 merged.


api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,131 @@
...
+	// VpcCreationStartedReason used when attempting to create a VPC for a managed cluster.
+	// Will not be applied to unmanaged clusters.
+	VpcCreationStartedReason = "VpcCreationStarted"
+	// VpcReconciliationFailedReason used when errors occur during VPC reconciliation
+	VpcReconciliationFailedReason = "VpcReconciliationFailed"
+)
+
+const (
+	// SubnetReady condition reports on the successful reconciliation of subnets.
+	SubnetReadyCondition clusterv1.ConditionType = "SubnetReady"
```
detiber, line 33: Likely want to pluralize this one, since it would have to represent the aggregate of all subnets we are managing.
```suggestion
	SubnetsReadyCondition clusterv1.ConditionType = "SubnetsReady"
```

api/v1alpha3/conditions_consts.go
```diff
@@ -0,0 +1,131 @@
...
+)
+
+const (
+	// NatGatewayReady condition reports successful reconciliation of NAT gateways.
+	// Only applicable to managed clusters.
+	NatGatewayReadyCondition clusterv1.ConditionType = "NatGatewayReady"
+	// NatGatewayCreationStartedReason set once when creating new NAT gateways.
+	NatGatewayCreationStartedReason = "NatGatewayCreationStarted"
+	// NatGatewayReconciliationFailedReason used when any errors occur during reconciliation of NAT gateways.
+	NatGatewayReconciliationFailedReason = "NatGatewayReconciliationFailed"
```
detiber, line 53: Similar here, these should likely be pluralized to represent the aggregate of all NatGateways

controllers/awscluster_controller.go
```diff
@@ -204,6 +230,20 @@ func (r *AWSClusterReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+			predicate.Funcs{
+				UpdateFunc: func(e event.UpdateEvent) bool {
+					if e.ObjectOld.GetObjectKind().GroupVersionKind().Kind != "AWSCluster" {
+						return true
+					}
+
+					oldCluster := e.ObjectOld.(*infrav1.AWSCluster)
+					newCluster := e.ObjectNew.(*infrav1.AWSCluster)
+
+					return !reflect.DeepEqual(oldCluster.Spec, newCluster.Spec)
```
detiber, line 243: Similar concerns here as with AWSMachine around potentially blocking reconciliations that we were previously relying on.

pkg/cloud/services/ec2/network.go
```diff
@@ -26,33 +29,53 @@ func (s *Service) ReconcileNetwork() (err error) {
 
 	// VPC.
 	if err := s.reconcileVPC(); err != nil {
+		conditions.MarkFalse(s.scope.AWSCluster, infrav1.VpcReadyCondition, infrav1.VpcReconciliationFailedReason, clusterv1.ConditionSeverityError, err.Error())
 		return err
 	}
+	conditions.MarkTrue(s.scope.AWSCluster, infrav1.VpcReadyCondition)
```
fabriziopandini, line 35: Wondering if we can move this call inside the `reconcileVPC` func, just before the final `return nil` call
Similar consideration applies for the following `MarkTrue` calls

controllers/awscluster_controller.go
```diff
@@ -204,6 +230,20 @@ func (r *AWSClusterReconciler) SetupWithManager(mgr ctrl.Manager, options contro
...
+				},
+			},
+		).
```
detiber, line 246:
```suggestion
I believe this will work for blocking reconciliation for incremental status updates during reconciliation without affecting any other mutations to the resource.
```suggestion
				// Avoid reconciling if the event triggering the reconciliation is related to incremental status updates
				// for AWSCluster resources only
				UpdateFunc: func(e event.UpdateEvent) bool {
					if e.ObjectOld.GetObjectKind().GroupVersionKind().Kind != "AWSCluster" {
						return true
					}

					oldCluster := e.ObjectOld.(*infrav1.AWSCluster).DeepCopy()
					newCluster := e.ObjectNew.(*infrav1.AWSCluster).DeepCopy()
					
					oldCluster.Status = infrav1.AWSClusterStatus{}
					newCluster.Status = infrav1.AWSClusterStatus{} 

					return !reflect.DeepEqual(oldCluster, newCluster)
				},
```
```

File: api/v1alpha3/conditions_consts.go [hunk 6]
```diff
@@ -0,0 +1,131 @@
...
+import clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+
+const (
+	// VpcReady condition reports on the successful reconciliation of a VPC
+	VpcReadyCondition clusterv1.ConditionType = "VpcReady"
+	// VpcCreationStartedReason used when attempting to create a VPC for a managed cluster.
+	// Will not be applied to unmanaged clusters.
+	VpcCreationStartedReason = "VpcCreationStarted"
+	// VpcReconciliationFailedReason used when errors occur during VPC reconciliation
+	VpcReconciliationFailedReason = "VpcReconciliationFailed"
```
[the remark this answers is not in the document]
gab-satchi: The failure reasons get used for any failures during reconciliation and not just creation.

pkg/cloud/services/ec2/bastion.go
```diff
@@ -67,6 +69,12 @@ func (s *Service) ReconcileBastion() error {
 	// Describe bastion instance, if any.
 	instance, err := s.describeBastionInstance()
 	if awserrors.IsNotFound(err) {
+		if conditions.GetReason(s.scope.AWSCluster, infrav1.BastionHostReadyCondition) == "" {
```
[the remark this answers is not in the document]
gab-satchi: Just in case there's an issue with creating an instance on AWS to avoid the flickering between `BastionCreationStartedReason` and `BastionHostFailedReason`

controllers/awscluster_controller.go
gab-satchi: Thanks. Updated

pkg/cloud/services/ec2/network.go
gab-satchi: yes they can. They were left there because I was originally trying to contain condition updates to the controller but now they're all over the services package

pkg/cloud/services/ec2/natgateways.go
```diff
@@ -75,6 +77,14 @@ func (s *Service) reconcileNatGateways() error {
 			continue
 		}
 
+		// set NatGatewayCreationStarted if the condition has never been set before
+		if conditions.GetReason(s.scope.AWSCluster, infrav1.NatGatewaysReadyCondition) == "" {
```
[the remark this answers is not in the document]
gab-satchi: Thanks. Updated

pkg/cloud/services/ec2/gateways.go
```diff
@@ -47,7 +48,7 @@ func (s *Service) reconcileInternetGateways() error {
 
 		ig, err := s.createInternetGateway()
 		if err != nil {
-			return nil
```
gab-satchi, line 50: Not related to PR. I assumed we'd want to return the error

randomvariable:
Remove "k8s.io/utils/diff" import from vpc_test.go, and run `make lint` again.

pkg/cloud/services/ec2/natgateways.go
```diff
@@ -41,9 +43,21 @@ func (s *Service) reconcileNatGateways() error {
 
 	if len(s.scope.Subnets().FilterPrivate()) == 0 {
 		s.scope.V(2).Info("No private subnets available, skipping NAT gateways")
+		conditions.MarkFalse(
+			s.scope.AWSCluster,
+			infrav1.NatGatewaysReadyCondition,
+			infrav1.NatGatewaysReconciliationFailedReason,
+			clusterv1.ConditionSeverityWarning,
+			"No private subnets available, skipping NAT gateways")
```
detiber, line 51: I'm wondering if marking the condition false here will lead to us never being able to consider the AWSCluster as "Ready" by conditions. I'm also not quite sure if we support deploying with "only" public subnets, @randomvariable what are your thoughts?

randomvariable: Not sure we ever allow all-public subnets in managed mode, since we have this in subnets.go

```go
	// Check that we need at least 1 private and 1 public subnet after we have updated the metadata
	if len(subnets.FilterPrivate()) < 1 {
		record.Warnf(s.scope.AWSCluster, "FailedNoPrivateSubnet", "Expected at least 1 private subnet but got 0")
		return errors.New("expected at least 1 private subnet but got 0")
	}
	if len(subnets.FilterPublic()) < 1 {
		record.Warnf(s.scope.AWSCluster, "FailedNoPublicSubnet", "Expected at least 1 public subnet but got 0")
		return errors.New("expected at least 1 public subnet but got 0")
	}
```

detiber: Gotcha, sounds like this shouldn't be an issue (at least for now) then :)

PR #1760, opened 2020-06-23
:running: e2e: Rewrite kubernetes main branch conformance to use clusterctl

randomvariable:
Signed-off-by: Naadir Jeewa <jeewan@vmware.com>

**What this PR does / why we need it**:
Moves the conformance test to be part of the new clusterctl e2e test suite, which should make it easier to maintain in the long term.

test/e2e_new/conformance is designed to be copied to Cluster API core. Some concurrent testing has been done with capd.

Additional changes to e2e suite include:
- Cleanup clusters when hitting ctrl+c
- Continuously dump namespaces on the bootstrap cluster every 5s
- k/k is not imported so separate go mod files are not required
- SSM session manager plugin downloaded by make, and consumed to scrape commands from EC2 instances without an SSH key.
- Accepts kubetest viper configurations as input
- Conformance can run in parallel, bringing test time to sub 45 minutes

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1610 
Fixes #1587 
Fixes https://github.com/kubernetes/kubernetes/issues/91860

PR #1789, opened 2020-07-06
🐛 Restricts AZ limit to 1 in e2e tests

gab-satchi:
The e2e tests have been failing for a while: https://k8s-testgrid.appspot.com/sig-cluster-lifecycle-cluster-api-provider-aws#periodic-e2e-v1alpha3&width=20
They all seem to hit the 20 minute timeout while waiting for the cluster to reach `InfrastructureReady`. One of the recent changes that increased the time needed for an AWScluster to reach ready is the subnet defaulting logic #1721 

These changes restrict the number of AZs to provision infrastructure to 1. While testing locally, this made an improvement from 12 minutes down to 5 minutes. 
Note: We still have a test case that specifies subnets in multiple AZs and validates machines are created appropriately but there is now a gap in e2e testing the defaulting logic.

### DIFF
--- a/test/e2e/aws_test.go
+++ b/test/e2e/aws_test.go
@@ -58,6 +58,7 @@ import (
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/kubeconfig"
 	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+	"github.com/gophercloud/gophercloud"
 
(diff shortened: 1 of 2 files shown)


gab-satchi:
It's not ideal to have most test cases not exercising the default behaviour so if it's better to just increase the timeout, I could update it.

randomvariable:
I've been avoiding doing this in favour of fixing up the new e2e tests in #1760

randomvariable:
I'm going to 
/lgtm
this and open an issue to restore the defaults when we have time to sort it out.

PR #1790, opened 2020-07-07
:running: Makefile: Build clean binaries on release

randomvariable:
Small fix to the release process so we don't end up publishing binaries that appear to be built from a non-tagged git tree.


vincepri:
We can probably merge this in today, given that this PR is syncing our OWNERS file with what's in `k8s.io/org` already. @randomvariable already has permissions, I'm +1 moving this along.

PR #1798, opened 2020-07-07
🏃 Migrate rest of e2e tests to use clusterctl framework

gab-satchi:
- Migrates remaining tests to use the clusterctl testing framework. 
- Removes parallel option for Ginkgo tests. Tests were competing for resources and hitting the 5 EIP resource limit
- Consolidate some test cases so fewer clusters are created.

In an effort to speed up the total time, some of the tests are combined so a new cluster isn't required per test: 
- Cluster name validations and provisioning extra AWS resources creates one cluster with the following tests:
  - name with more than 22 characters
  - name with '.'
  - name that starts with 'sg-'
  - Provisioning LoadBalancer dynamically and deleting on cluster deletion
  - Provisioning volumes dynamically and retain on cluster deletion
- multiple workload clusters in different namespaces with machine failures also tests:
  - Delete infra node directly from infra provider
  - MachineDeployment will replace a deleted Machine

/hold for https://github.com/kubernetes/test-infra/pull/17590 to merge

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -135,7 +135,7 @@ test-e2e: $(GINKGO) $(KIND) ## Run e2e tests
 
 .PHONY: test-e2e-new ## Run new e2e tests using clusterctl
 test-e2e-new: $(GINKGO) $(KIND) $(SSM_PLUGIN) $(KUSTOMIZE) e2e-image ## Run e2e tests
-	time $(GINKGO) -trace -progress -nodes=$(GINKGO_NODES) -v -tags=e2e -focus=$(E2E_FOCUS) $(GINKGO_ARGS) ./test/e2e_new/... -- -config-path="$(E2E_CONF_PATH)" -artifacts-folder="$(ARTIFACTS)" $(E2E_ARGS)
+	time $(GINKGO) -trace -progress -v -tags=e2e -focus=$(E2E_FOCUS) $(GINKGO_ARGS) ./test/e2e_new/... -- -config-path="$(E2E_CONF_PATH)" -artifacts-folder="$(ARTIFACTS)" $(E2E_ARGS)
 
 .PHONY: e2e-image
 e2e-image:
--- a/test/e2e_new/data/infrastructure-aws/cluster-template-multi-az.yaml
+++ b/test/e2e_new/data/infrastructure-aws/cluster-template-multi-az.yaml
@@ -1,0 +1,125 @@
+---
+apiVersion: cluster.x-k8s.io/v1alpha3
+kind: Cluster
+metadata:
+  name: "${CLUSTER_NAME}"
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks: ["192.168.0.0/16"]
+  infrastructureRef:
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+    kind: AWSCluster
+    name: "${CLUSTER_NAME}"
+  controlPlaneRef:
+    kind: KubeadmControlPlane
+    apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
+    name: "${CLUSTER_NAME}-control-plane"
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AWSCluster
+metadata:
+  name: "${CLUSTER_NAME}"
+spec:
+  region: "${AWS_REGION}"
+  sshKeyName: "${AWS_SSH_KEY_NAME}"
+  networkSpec:
+    subnets:
+      - availabilityZone: "${AWS_AVAILABILITY_ZONE_1}"
+        cidrBlock: "10.0.0.0/24"
+      - availabilityZone: "${AWS_AVAILABILITY_ZONE_1}"
+        cidrBlock: "10.0.1.0/24"
+        isPublic: true
+      - availabilityZone: "${AWS_AVAILABILITY_ZONE_2}"
+        cidrBlock: "10.0.2.0/24"
+      - availabilityZone: "${AWS_AVAILABILITY_ZONE_2}"
+        cidrBlock: "10.0.3.0/24"
+        isPublic: true
+---
+kind: KubeadmControlPlane
+apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
+metadata:
+  name: "${CLUSTER_NAME}-control-plane"
+spec:
+  replicas: ${CONTROL_PLANE_MACHINE_COUNT}
+  infrastructureTemplate:
+    kind: AWSMachineTemplate
+    apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+    name: "${CLUSTER_NAME}-control-plane"
+  kubeadmConfigSpec:
+    initConfiguration:
+      nodeRegistration:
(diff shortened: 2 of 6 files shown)


test/e2e_new/data/infrastructure-aws/cluster-template-multi-az.yaml
```diff
@@ -0,0 +1,125 @@
...
+apiVersion: infrastructure.cluster.x-k8s.io/v1alpha3
+kind: AWSCluster
+metadata:
+  name: "${CLUSTER_NAME}"
+spec:
+  region: "${AWS_REGION}"
+  sshKeyName: "${AWS_SSH_KEY_NAME}"
+  networkSpec:
+    subnets:
+      - availabilityZone: "${AWS_AVAILABILITY_ZONE_1}"
```
randomvariable, line 28: What was the reason to do it this way instead of
```
  networkSpec:
    vpc:
      availabilityZoneUsageLimit: 3
```
?

Makefile
```diff
@@ -156,7 +156,7 @@ test-e2e: $(GINKGO) $(KIND) ## Run e2e tests
 
 .PHONY: test-e2e-new ## Run new e2e tests using clusterctl
 test-e2e-new: $(GINKGO) $(CONFORMANCE_CI_TEMPLATE) $(KIND) $(SSM_PLUGIN) e2e-image ## Run e2e tests
-	time $(GINKGO) -trace -progress -nodes=$(GINKGO_NODES) -v -tags=e2e -focus=$(E2E_FOCUS) $(GINKGO_ARGS) ./test/e2e_new/... -- -config-path="$(E2E_CONF_PATH)" -artifacts-folder="$(ARTIFACTS)" $(E2E_ARGS)
+	time $(GINKGO) -trace -progress -v -tags=e2e -focus=$(E2E_FOCUS) $(GINKGO_ARGS) ./test/e2e_new/... -- -config-path="$(E2E_CONF_PATH)" -artifacts-folder="$(ARTIFACTS)" $(E2E_ARGS)
```
detiber, line 159: One potential option that we may have rather than trying to force serialization for everything, we could follow the upstream convention of marking specs that need to be serialized with [Serial], then we could have two separate make targets, which set the focus appropriately to either target [Serial] specs or exclude [Serial] specs.

test/e2e_new/data/infrastructure-aws/cluster-template-multi-az.yaml
gab-satchi: The test that uses this template forces creating machine deployments in specific subnets. Without it, the instances would get created in the first available subnet and there's no validation on multi-az for worker nodes.

Makefile
gab-satchi: I originally went that route as well but then realized that none of the tests can be run in parallel since they all create at least 1 cluster and creating 2 clusters would surpass the resource limit.

gab-satchi:
It looks like the artifacts folder is already set as an environment variable: 
```
time hack/tools/bin/ginkgo -trace -progress -v -tags=e2e -focus="functional tests"  ./test/e2e_new/... -- -config-path="/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/data/e2e_conf.yaml" -artifacts-folder="/logs/artifacts"
```

but the `e2e_conf.yaml` specifically looks under `<repo_root>/_artifacts`

test/e2e_new/data/infrastructure-aws/cluster-template-multi-az.yaml
randomvariable: From other conversations, this is needed for worker node distribution.

Makefile
randomvariable: fwiw, ginkgo recommend separate packages for serial vs. parallel. Either that or an external lock.

randomvariable:
should be able to rebase now

randomvariable:
/lgtm
/hold 
pending e2e pass

randomvariable:
Maybe need a little longer here:
```
/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/aws_test.go:72
  Cluster name validations and provisioning extra AWS resources
  /home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/aws_test.go:85
    Should create a cluster, AWS load balancer, and volume [It]
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/aws_test.go:102
    Timed out after 600.000s.
      
    Expected
        <bool>: false
    to be true
    /home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/aws_test.go:889
    Full Stack Trace
    sigs.k8s.io/cluster-api-provider-aws/test/e2e_new.waitForStatefulSetRunning(0x27d0c1c, 0x11, 0x27bea92, 0x7, 0x2, 0xc0001ab890, 0x27cc1f0, 0xf, 0x27c8de2, 0xd, ...)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/aws_test.go:889 +0x205
    sigs.k8s.io/cluster-api-provider-aws/test/e2e_new.createStatefulSet(0x27d0c1c, 0x11, 0x27bea92, 0x7, 0x2, 0xc0001ab890, 0x27cc1f0, 0xf, 0x27c8de2, 0xd, ...)
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/aws_test.go:487 +0x379
    sigs.k8s.io/cluster-api-provider-aws/test/e2e_new.glob..func1.2.1()
    	/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e_new/aws_test.go:120 +0x406
    github.com/onsi/ginkgo/internal/leafnodes.(*runner).runSync(0xc000428f60, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 0x0, 
```

gab-satchi:
Ya I just saw the flake locally as well. It looks like it fails to schedule the pod as it waits for the worker node to become ready

gab-satchi:
There's also a point of failure with zone-affinity. If the control plane is created in one of the zones other than the first one, it tends to create the Volume in that zone as well. Adding a label to force storage classes to be created in the first AZ. This assumes AWS zones are returned consistently. If it's not reliable, I can move the volume test into one of the other tests that restricts the AZ used. 

/test pull-cluster-api-provider-aws-e2e-new

test/e2e_new/e2e_suite_test.go
```diff
@@ -203,7 +210,9 @@ var _ = SynchronizedBeforeSuite(func() []byte {
 	kubetestConfig = &conf.KubetestConfig
 	bootstrapClusterProxy = framework.NewClusterProxy("bootstrap", conf.KubeconfigPath, initScheme())
 	e2eConfig = &conf.E2EConfig
-
+	azs := getAvailabilityZones()
```
randomvariable, line 213: Assume every AWS API call is backed by DynamoDB on the backend, which does guarantee order of results.

randomvariable:
`DescribeAvailabilityZones` does appear to be deterministic, a surprise when it comes to the AWS API.

PR #1802, opened 2020-07-08
🐛  Fixes cluster creation with names starting with 'sg-'

### LINKED ISSUE #1725: Allow clusters to be created when named starting sg-
/kind bug

Create a cluster with the prefix `sg-`

The controller will fail with the following error:
```
1 controller.go:258] controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile network for AWSCluster default/sg-cluster: failed to create security group \"bastion\" in vpc \"vpc-0db5779f71b3ed663\": InvalidParameterValue: Value (sg-cluster-bastion) for parameter GroupName is invalid. Group names may not be in the format sg-.\n\tstatus code: 400, request id: 525f5d67-7926-42f0-931e-0def5cab93da"  "controller"="awscluster" "request"={"Namespace":"default","Name":"sg-cluster"}*
```

The cluster gets created.

The restriction is listed in API docs at https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateSecurityGroup.html

**Environment:**

- Cluster-api-provider-aws version: 0.5.3

gab-satchi:
- Fixes security group failures when attempting to create a cluster that has a name starting with 'sg-'

Fixes #1725

### DIFF
--- a/pkg/cloud/services/ec2/securitygroups.go
+++ b/pkg/cloud/services/ec2/securitygroups.go
@@ -18,6 +18,7 @@ package ec2
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/ec2/securitygroups.go
```diff
@@ -455,7 +456,7 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
 }
 
 func (s *Service) getSecurityGroupName(clusterName string, role infrav1.SecurityGroupRole) string {
-	return fmt.Sprintf("%s-%v", clusterName, role)
+	return fmt.Sprintf("%s-%v", strings.TrimPrefix(clusterName, "sg-"), role)
```
randomvariable, line 459: Is there a way to account for two clusters with names "test" and "sg-test"?

randomvariable, line 459:
```suggestion
    groupPrefix := clusterName
	if strings.HasPrefix(clusterName, "sg-") {
		groupPrefix = "@" + clusterName 
	}
	return fmt.Sprintf("%s-%v", groupPrefix, role)
```

Reason for this is that Kubernetes names are DNS RFC-1123, and AWS allows 
> Allowed characters are a-z, A-Z, 0-9, spaces, and ._-:/()#,@[]+=;{}!$*.
Prefixing with @ for the `sg-` special case won't conflict with the same name without the `sg-` prefix.

PR #1807, opened 2020-07-10
✨ controllers: Capture metrics for every AWS API request

### LINKED ISSUE #78: Metrics/Tracing

randomvariable:
Captures count and latency metrics for all AWS API requests, with per-region and per-controller labels.

Fixes #78

### DIFF
--- a/pkg/cloud/metrics/metrics.go
+++ b/pkg/cloud/metrics/metrics.go
@@ -1,0 +1,104 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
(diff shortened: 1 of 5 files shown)


pkg/cloud/metrics/metrics.go
```diff
@@ -0,0 +1,103 @@
...
+			statusCode = strconv.Itoa(r.HTTPResponse.StatusCode)
+		}
+		if r.Error != nil {
+			var ok bool
+			if errorCode, ok = awserrors.Code(r.Error); !ok {
+				errorCode = "internal"
+			}
+		}
+		awsRequestCount.WithLabelValues(controller, service, region, operation, statusCode, errorCode).Inc()
+		awsRequestDurationSeconds.WithLabelValues(controller, service, region, operation).Observe(duration.Seconds())
```
detiber, line 88: Should there be a call here for updating `awsCallRetries` to record the `RetryCount` from the `Request`?

randomvariable: Yes, thanks. Updated :|

PR #1823, opened 2020-07-15
e2e: Fixups for conformance template generation

randomvariable:
Resolves a CI difference where _artifacts is not the same as the ARTIFACTS directory, by moving the logic from GNU Make to go

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,5 @@ junit.*.xml
 
 # test results
 _artifacts
+awsiamconfiguration.yaml
+cloudformation.yaml
--- a/Makefile
+++ b/Makefile
@@ -39,15 +39,12 @@ BIN_DIR := bin
 REPO_ROOT := $(shell git rev-parse --show-toplevel)
 TEST_E2E_DIR := test/e2e
 TEST_E2E_NEW_DIR := test/e2e_new
-OVERLAY_DIR := $(ARTIFACTS)/overlay
-OVERLAY_SOURCE := $(TEST_E2E_NEW_DIR)/data/kubetest/kustomization
 
 # Files
 E2E_DATA_DIR ?= $(REPO_ROOT)/test/e2e_new/data
 E2E_CONF_PATH  ?= $(E2E_DATA_DIR)/e2e_conf.yaml
 KUBETEST_CONF_PATH ?= $(abspath $(E2E_DATA_DIR)/kubetest/conformance.yaml)
 KUBETEST_FAST_CONF_PATH ?= $(abspath $(REPO_ROOT)/test/e2e_new/data/kubetest/conformance-fast.yaml)
-CONFORMANCE_CI_TEMPLATE := $(ARTIFACTS)/templates/cluster-template-conformance-ci-artifacts.yaml
 
 # Binaries.
 CLUSTERCTL := $(BIN_DIR)/clusterctl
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -47,7 +47,8 @@ E2E_DATA_DIR ?= $(REPO_ROOT)/test/e2e_new/data
 E2E_CONF_PATH  ?= $(E2E_DATA_DIR)/e2e_conf.yaml
 KUBETEST_CONF_PATH ?= $(abspath $(E2E_DATA_DIR)/kubetest/conformance.yaml)
 KUBETEST_FAST_CONF_PATH ?= $(abspath $(REPO_ROOT)/test/e2e_new/data/kubetest/conformance-fast.yaml)
-CONFORMANCE_CI_TEMPLATE := $(ARTIFACTS)/templates/cluster-template-conformance-ci-artifacts.yaml
+CONFORMANCE_CI_TEMPLATE := _artifacts/templates/cluster-template-conformance-ci-artifacts.yaml
```
detiber, line 50: Is this related to something hardcoded in a kustomize file? I'm wondering if using envsubst or similar might be better than having a hardcoded path to _artifacts

detiber: Maybe something we could add to the e2e config parser?

randomvariable: Let me check

randomvariable: TBH, don't like this much logic in the Makefile as it prohibits `go test`

detiber: Yeah, that's one of the reasons I was thinking that maybe we need to push that support into the e2e config parser

randomvariable: modifying at run time in go, but github is not showing the changes...

randomvariable:
Pod never came up for test
/test pull-cluster-api-provider-aws-make-conformance

randomvariable:
hmm, looks like github is causing problems.

.gitignore
```diff
@@ -58,3 +58,5 @@ junit.*.xml
 
 # test results
 _artifacts
+awsiamconfiguration.yaml
```
gab-satchi, line 61: do we need these? thought they also get created under the artifacts folder

test/e2e_new/conformance_test.go
```diff
@@ -99,3 +105,56 @@ var _ = Describe("conformance tests", func() {
 	})
 
 })
+
+func prepConformanceConfig(artifactsDir string, e2eConfig *clusterctl.E2EConfig) error {
+	for i, p := range e2eConfig.Providers {
+		if p.Name != "aws" {
+			continue
+		}
+		e2eConfig.Providers[i].Files = append(p.Files, clusterctl.Files{
```
gab-satchi, line 114: we could also leave the copying and kustomize call in the Makefile and then append the file here after ensuring it exists.

.gitignore
randomvariable: Haven't got to the bottom as to why, but sometimes it gets created in the test directory. Will open a follow up issue to resolve than try to solve it now.

test/e2e_new/conformance_test.go
randomvariable: Idea is to start moving towards not needing the makefile, and allowing `go test` to work if you wanted to use it that way.

.gitignore
randomvariable: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1827

test/e2e_new/conformance_test.go
```diff
@@ -99,3 +105,56 @@ var _ = Describe("conformance tests", func() {
...
+			SourcePath: path.Join(artifactFolder, "/templates/cluster-template-conformance-ci-artifacts.yaml"),
+			TargetName: "cluster-template-conformance-ci-artifacts.yaml",
```
detiber, line 116: Is there no way currently to trigger a conformance test against a released version of Kubernetes? It looks like everything is hard coded towards running against ci artifacts.

randomvariable: We always generate the conformance template, but https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/test/e2e_new/conformance_test.go#L59 determines whether or not it actually gets used.

PR #1833, opened 2020-07-20
✨  Use filters to select subnet for machines

### LINKED ISSUE #1776: Implement subnet filter capability currently in the API for AWSMachineTemplate
/kind feature

I need to specify which subnets my [AWSMachineTemplate](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml#L200-L233) can land in.  Currently, that capability is restricted to just the [`.spec.template.spec.subnet.id`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml#L231-L233) field, I would like to expand this to include the already existing [`filters`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml#L208-L230) object that lives in the spec.  This would allow me to optionally tag, or look up subnets where I do not know the `id` ahead of time (either because it's dynamically created, or I am using the same templates for multiple deployments).

Reference: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinetemplates.yaml#L200-L233

Does not require changes to the CRDs, as the `filters` object is already there.

**Environment:**

gab-satchi:
Currently the only way to specify a subnet to be used by a machine is by using the subnetID. It's restrictive as the user may not always have the subnet provisioned beforehand or have access to the subnetID. These changes allow the user to specify AWS filters to be used to select a subnet for a given machine template. 

It looks for subnet to use in the following order: 
- subnetID specified in template
- subnet filters specified in template (will add availability-zone filter if failureDomain is also specified)
- filter available private subnets by failureDomain if specified
- select first available private subnet

**If multiple subnets are returned by the filter parameters, the first one is used.** 

Fixes #1776

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -159,48 +159,11 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
 		}
 	}
 
-	// Prefer AWSMachine.Spec.FailureDomain for now while migrating to the use of
-	// Machine.Spec.FailureDomain. The MachineController will handle migrating the value for us.
-	failureDomain := scope.AWSMachine.Spec.FailureDomain
-	if failureDomain == nil {
-		failureDomain = scope.Machine.Spec.FailureDomain
+	subnetID, err := s.findSubnet(scope)
+	if err != nil {
+		return nil, err
 	}
-
-	// Pick subnet from the machine configuration, or based on the availability zone specified,
-	// or default to the first private subnet available.
-	// TODO(vincepri): Move subnet picking logic to its own function/method.
-	switch {
-	case scope.AWSMachine.Spec.Subnet != nil && scope.AWSMachine.Spec.Subnet.ID != nil:
-		input.SubnetID = *scope.AWSMachine.Spec.Subnet.ID
-
-	case failureDomain != nil:
-		subnets := s.scope.Subnets().FilterPrivate().FilterByZone(*failureDomain)
-		if len(subnets) == 0 {
-			record.Warnf(scope.AWSMachine, "FailedCreate",
-				"Failed to create instance: no subnets available in availability zone %q", *failureDomain)
-
-			return nil, awserrors.NewFailedDependency(
-				errors.Errorf("failed to run machine %q, no subnets available in availability zone %q",
-					scope.Name(),
-					*failureDomain,
-				),
-			)
-		}
-		input.SubnetID = subnets[0].ID
-
-		// TODO(vincepri): Define a tag that would allow to pick a preferred subnet in an AZ when working
-		// with control plane machines.
-
-	case input.SubnetID == "":
-		sns := s.scope.Subnets().FilterPrivate()
-		if len(sns) == 0 {
-			record.Eventf(s.scope.InfraCluster(), "FailedCreateInstance", "Failed to run machine %q, no subnets available", scope.Name())
-			return nil, awserrors.NewFailedDependency(
-				errors.Errorf("failed to run machine %q, no subnets available", scope.Name()),
-			)
-		}
-		input.SubnetID = sns[0].ID
-	}
+	input.SubnetID = subnetID
 
 	if s.scope.Network().APIServerELB.DNSName == "" {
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/ec2/instances.go
```diff
@@ -259,6 +222,82 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
...
 }
 
+// findSubnet attempts to retrieve a subnet ID in the following order:
+// - subnetID specified in machine configuration,
+// - subnet based on filters in machine configuration
+// - subnet based on the availability zone specified,
+// - default to the first private subnet available.
+func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
+	// Prefer AWSMachine.Spec.FailureDomain for now while migrating to the use of
+	// Machine.Spec.FailureDomain. The MachineController will handle migrating the value for us.
```
detiber, line 232: This should likely be reversed, otherwise we could potentially ignore the failure domain spreading logic in CAPI.

gab-satchi: That bit was copied over from the existing code: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/ec2/instances.go#L162-L167

I'll do some digging to see if it needs updating

gab-satchi: I think this is working as expected. 
- In most cases, the failureDomain on the AWSMachine will be nil and the failure domain spreading logic will work as expected. 
- If a failureDomain is defined by the user in the AWSMachineTemplate then all machines would be created in that AZ regardless of the rotation that CAPI attempts.

detiber: > If a failureDomain is defined by the user in the AWSMachineTemplate then all machines would be created in that AZ regardless of the rotation that CAPI attempts.

This is the situation I'm most worried about, consider someone has a v1alpha2-based cluster and upgrades to v1alpha3, when they perform the steps for adopting their existing control plane components into a KubeadmControlPlane they accidentally set AWSMachineTemplate.Spec.Template.Spec.FailureDomain to a specific failure domain, then any additional actions taken on their control plane will now result in a control plane that is entirely stored in a single failure domain overriding the spreading logic in KubeadmControlPlane.

gab-satchi: Ah I didn't think about the upgrade flow and users accidentally restricting the failure domain. Flipping the logic would mean the user can't dictate the failure domain for control plane with that field. But they still have ways through the subnet field so that sounds reasonable. Will update.

pkg/cloud/services/ec2/instances.go
```diff
@@ -259,6 +222,81 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
...
+func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
+	// Check Machine.Spec.FailureDomain first as it's used by KubeadmControlPlane to spread machines across failure domains.
+	failureDomain := scope.Machine.Spec.FailureDomain
+	if failureDomain == nil {
+		failureDomain = scope.AWSMachine.Spec.FailureDomain
+	}
+
+	switch {
+	case scope.AWSMachine.Spec.Subnet != nil && scope.AWSMachine.Spec.Subnet.ID != nil:
+		return *scope.AWSMachine.Spec.Subnet.ID, nil
```
detiber, line 239: I'm wondering if it would make sense to add a validation step here that the specified Subnet.ID does not conflict with the failureDomain if set...  Not something that needs to be solved as part of this PR, though.

gab-satchi: Yeah it definitely makes sense to validate that. Created #1838

rudoi:
this looks good, thank you!

/approve

PR #1844, opened 2020-07-27
📖 Updates docs to upload existing keypair to AWS

srm09:
The `aws ec2 import-key-pair` requires the input for the `--public-key-material` flag to be base64 encoded.

n/a

### DIFF
--- a/docs/prerequisites.md
+++ b/docs/prerequisites.md
@@ -81,7 +81,7 @@ aws ssm put-parameter --name "/sigs.k8s.io/cluster-api-provider-aws/ssh-key" \
 # Replace with your own public key
 aws ec2 import-key-pair \
   --key-name default \
-  --public-key-material "$(cat ~/.ssh/id_rsa.pub)"
+  --public-key-material "$(base64 -i ~/.ssh/id_rsa.pub)"
 ```
 
 > **NB**: Only RSA keys are supported by AWS.


docs/prerequisites.md
```diff
@@ -81,7 +81,7 @@ aws ssm put-parameter --name "/sigs.k8s.io/cluster-api-provider-aws/ssh-key" \
 # Replace with your own public key
 aws ec2 import-key-pair \
   --key-name default \
-  --public-key-material "$(cat ~/.ssh/id_rsa.pub)"
+  --public-key-material "$(base64 -i ~/.ssh/id_rsa.pub)"
```
detiber, line 84: I went to look to see if this works for both OSX and Linux, and I believe it does, but only coincidentally since the `-i` flag is different between the two. https://www.unix.com/man-page/osx/1/base64/ vs https://linux.die.net/man/1/base64

I verified the expected behavior on Linux, but this probably should have a comment to avoid confusion later on.

ncdc: I don't think this change is needed. From the AWS CLI docs:

```
        --public-key-material (blob)
          The public key. For API calls, the text must be base64-encoded.  For
          command line tools, base64 encoding is performed for you.
```

srm09: I agree that this requirement is made clear in the CLI docs. But if an end user is following the CAPA docs, they might encounter an error, thus introducing an additional hoop in the installation process.

srm09:
Anyway this isn't super important, something I just happened to stumble across. I am happy to close it, wasn't any effort on my end.

docs/prerequisites.md
ncdc: What error?

srm09:
Not needed, misread the documentation

PR #1848, opened 2020-07-30
refactor: tags package to support other services

### LINKED ISSUE #1843: Refactor tags package (specifically Apply)
/kind feature

The tags package is used throughout the services to apply infrav1.Tags to resources created in AWS. The current implementation in the `tags` package is very EC2 specific:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/tags/tags.go#L36

So it would be good to refactor this so that other services can take advantage of the tags differencing (and apply) functionality.

This will help with #1787 in a tags only update scenario.

**Environment:**

richardcase:
The existing tags package applies tags to EC2 only. This is a small refactor to allow a 'apply' function to be passed in and this function contains resource/service specific functionality to applythe tags. The tags package still keeps the diff functionality. This will help with apply tags in the EKS control plane reconciler ( #1787 ).

Fixes #1843


pkg/cloud/tags/tags.go
```diff
@@ -79,10 +74,10 @@ func BuildParamsToTagSpecification(ec2ResourceType string, params infrav1.BuildP
 }
 
 // Ensure applies the tags if the current tags differ from the params.
-func Ensure(current infrav1.Tags, params *ApplyParams) error {
-	diff := computeDiff(current, params.BuildParams)
+func Ensure(current infrav1.Tags, params *infrav1.BuildParams, fn TagsApplyFunc) error {
```
vincepri, line 77: What if we make a builder struct instead of passing in an apply function? This way we could always extend the tags applier in the future as part of this package.

The reason I'm suggesting to go this direction is to make sure the `tags` package is still useful. In the current form, it seems very limited.

Consider something like:

```go
tags.WithEC2(..client..).Ensure/Apply()
```

richardcase: @vincepri - I like that idea, that would feel cleaner. I'll rework it now.

richardcase: @vincepri - i've made some changes to the package along the lines you mentioned.

rudoi:
/lgtm 

i'll leave final approve up to @vincepri? :D

richardcase:
> i'll leave final approve up to @vincepri? :D

@vincepri - does this look ok to you?

randomvariable:
@vincepri is off this week, I'll take a look.

PR #1853, opened 2020-07-31
🌱 refactor AWSMachine to use scope interfaces

rudoi:
**What this PR does / why we need it**: Refactors AWSMachine reconciliation to utilize the new scope interfaces. This will make onboarding the AWSManagedControlPlane stuff more simple.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1852

PR #1856, opened 2020-08-03
✨ refactor: split security groups from network service

### LINKED ISSUE #1857: Refactor network service to separate out security groups
/kind feature

For the EKS implementation we need to be able to reconcile the network without the security groups. The security groups will be handled differently with the managed side of the provider

Needed for #1787 

**Environment:**

richardcase:

The reconciliation of security groups has been split out from the network service. This is to enable the EKS implementation to reconcile the network without the security groups.

Fixes #1857

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -24,16 +24,14 @@ import (
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
-	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/ec2"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/elb"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/network"
+
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
 	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/conditions"
+
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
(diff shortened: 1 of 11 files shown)


rudoi:
@richardcase I think this generally looks good! Just to understand what happens next - the EKS controlplane code will use a different path to sort out security groups?

I think we'll still need the node and LB security groups "created" - just the controlplane SG can't be populated until after the EKS cluster itself is created.

richardcase:
> I think this generally looks good! Just to understand what happens next - the EKS controlplane code will use a different path to sort out security groups?

@rudoi  - yes thats correct. The EKS control plane implements its own logic with security groups. Specifically it looks up the automatically created security group and attaches the details of that to the `Status` on the `AWSManagedControlPlane`:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f01d62a133b9501e765ccbb983da2238adf68455/pkg/cloud/services/eks/securitygroup.go#L37

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f01d62a133b9501e765ccbb983da2238adf68455/exp/api/v1alpha3/awsmanagedcontrolplane_types.go#L134

rudoi:
@richardcase got it - ok so following the AWSMachine reconciliation path, I'm looking at this here:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d59f6777ce348d5e2fb790b1eebd371785a58416/pkg/cloud/services/ec2/instances.go#L302-L326

With the new interfaces, s.scope.SecurityGroups will be provided by the AWSManagedControlPlaneScope, which is fine, but this function here is going to expect a map containing "node" and "lb" keys. "node" sounds like it could be the one created by EKS, but do we need to add something additional to get the "lb" key in there?

richardcase:
> @richardcase got it - ok so following the AWSMachine reconciliation path, I'm looking at this here:
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d59f6777ce348d5e2fb790b1eebd371785a58416/pkg/cloud/services/ec2/instances.go#L302-L326
> 
> With the new interfaces, s.scope.SecurityGroups will be provided by the AWSManagedControlPlaneScope, which is fine, but this function here is going to expect a map containing "node" and "lb" keys. "node" sounds like it could be the one created by EKS, but do we need to add something additional to get the "lb" key in there?



> "node" sounds like it could be the one created by EKS, but do we need to add something additional to get the "lb" key in there?

A very good point @rudoi . I will investigate and get back to you. And thanks this thread has made me realize i forgot to update the managed scope around security groups.

rudoi:
/lgtm 

Made a note on #1853 as well

richardcase:
Thanks @randomvariable

PR #1858, opened 2020-08-04
🌱 Support using custom endpoints for AWS services

### LINKED ISSUE #1855: Custom Service Endpoint Configuration
/kind feature

Similar to a feature that Terraform has: https://registry.terraform.io/providers/hashicorp/aws/latest/docs/guides/custom-service-endpoints
Allow the cluster api provider, "to connect to non-default AWS service endpoints and AWS compatible solutions. This may be useful for environments with specific compliance requirements, such as using AWS FIPS 140-2 endpoints, connecting to AWS Snowball, SC2S, or C2S environments, or local testing."

Maybe this isn't an issue in a C2S or FIPS environment due to the routing model in the aws-sdk: https://github.com/aws/aws-sdk-go/blob/master/models/endpoints/endpoints.json

**Environment:**

- Cluster-api-provider-aws version: v0.5.5
- Kubernetes version: (use `kubectl version`): v1.18.6

Promaethius:
**What this PR does / why we need it**: Allows custom endpoints for aws services to be specified for the aws provider. Not only would this allow for local stack testing but would pave the road for Cluster API into AWS IC partitions which require FIPS named endpoints.

Fixes #1855

### DIFF
--- a/pkg/cloud/endpoints/endpoints.go
+++ b/pkg/cloud/endpoints/endpoints.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package endpoints
+
+import (
+	"errors"
+	"net/url"
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws/endpoints"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
+)
+
+var (
+	errServiceEndpointFormat             = errors.New("must be formatted as ${ServiceID}=${URL}")
+	errServiceEndpointSigningRegion      = errors.New("must be formatted as ${SigningRegion}:${ServiceID1}=${URL1},${ServiceID2}=${URL2...}")
+	errServiceEndpointURL                = errors.New("must use a valid URL as a service-endpoint")
+	errServiceEndpointServiceID          = errors.New("must use a valid serviceID from the AWS GO SDK")
+	errServiceEndpointDuplicateServiceID = errors.New("same serviceID defined twice for signing region")
+)
+
+func serviceEnum() []string {
+	var serviceIDs = []string{}
+	resolver := endpoints.DefaultResolver()
+	partitions := resolver.(endpoints.EnumPartitions).Partitions()
+	for _, p := range partitions {
+		for id := range p.Services() {
+			var add bool = true
+			for _, s := range serviceIDs {
+				if id == s {
+					add = false
+				}
+			}
+			if add {
+				serviceIDs = append(serviceIDs, id)
+			}
+		}
+	}
+
+	return serviceIDs
+}
+
+// ParseFlag parses the command line flag of service endponts in the format ${SigningRegion1}:${ServiceID1}=${URL1},${ServiceID2}=${URL2}...;${SigningRegion2}...
+// returning a set of ServiceEndpoints
+func ParseFlag(serviceEndpoints string) ([]scope.ServiceEndpoint, error) {
+	if serviceEndpoints == "" {
+
+		return nil, nil
+	}
+	serviceIDs := serviceEnum()
+	signingRegionConfigs := strings.Split(serviceEndpoints, ";")
+	endpoints := []scope.ServiceEndpoint{}
+	for _, regionConfig := range signingRegionConfigs {
+		components := strings.SplitN(regionConfig, ":", 2)
+		if len(components) != 2 {
+
+			return nil, errServiceEndpointSigningRegion
+		}
+		signingRegion := components[0]
+		servicePairs := strings.Split(components[1], ",")
+		seenServices := []string{}
+		for _, servicePair := range servicePairs {
+			kv := strings.Split(servicePair, "=")
+			if len(kv) != 2 {
+
+				return nil, errServiceEndpointFormat
+			}
+			var serviceID string = ""
+			for _, id := range serviceIDs {
+				if kv[0] == id {
+					serviceID = kv[0]
+
+					break
+				}
+			}
+			if serviceID == "" {
+				return nil, errServiceEndpointServiceID
+			}
+			if containsString(seenServices, serviceID) {
+				return nil, errServiceEndpointDuplicateServiceID
+			}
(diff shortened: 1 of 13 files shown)


api/v1alpha3/awscluster_types.go
```diff
@@ -82,6 +82,23 @@ type AWSClusterSpec struct {
 	// Bastion contains options to configure the bastion host.
 	// +optional
 	Bastion Bastion `json:"bastion"`
+
+	// Endpoints contains options to map custom AWS endpoints to services.
+	// +optional
+	Endpoints []Endpoint `json:"endpoints,omitempty"`
```
detiber, line 88: Would you consider endpoints to be distinct from credentials in that the same set of credentials would be used across multiple sets of endpoints? 

I'm wondering if it would make sense to tie the endpoints more closely with the credentials used (which we currently configured at the controller-manager level), but there is an outstanding issue to add multitenancy support to a single controller instance here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1552

@randomvariable thoughts?

Promaethius: That makes sense; I built the schema around what AWS presents in its SDK.

Promaethius:
I'm thinking that this might have to extend to the `cloud.conf` that's generated on controller nodes.
```
[ServiceOverride "IC"]
  Service=ec2
  Region=us-iso-east-1
  URL=https://ec2.us-iso-east-1.c2s.ic.gov
  SigningRegion=us-iso-east-1
```

randomvariable:
Think this one might be tricky if we need to modify the cloud controller manager on the created workload cluster. As @detiber notes, this might be more appropriate as part of the multi-tenancy work in progress. Will get back to you on how to deal with this.

Promaethius:
@randomvariable did you get direction? I got word back from aws-go-sdk that C2S endpoints should be specified through this custom parser. I'm worried about scope now cause that may mean if I define custom service endpoints here that they should become available to the Python CLI and the kubelet Cloud provider plugin.

randomvariable:
@Promaethius I didn't get round to it this week, but promise to answer on Monday.

That said, the custom service endpoints won't have downstream impacts on other processes. The main issue we have here is how to make sure the cloud provider also follows the endpoints. It may be the answer is that you use a kubeadmconfigtemplate and KCP with the appropriate settings.

The wider issue is that we're implementing a CRD that will hold different types of AWS credentials, and we might want to extend it to also include the service endpoints. This is being tracked for v0.6.1 in #1753

Promaethius:
@randomvariable what I'm thinking for configuring endpoints of the aws cli and kubelet cloud provider:
* Kubelet
  * cloud-init arbitrary file: /etc/kubernetes/cloud.conf
  * kubeadm.conf rendered with `--cloud-config /etc/kubernetes/cloud.conf` flag
* AWS CLI
  * Render cloud-init script with `--endpoint-url` in AWS CLI calls OR
  * Extend the CAPA image builder with an AWS CLI plugin which allows service endpoint configuration from a file

Promaethius:
@randomvariable any updates? I can make the time to join collaborative sessions if that'll push this through quicker

randomvariable:
We had a Zoom call to discuss this yesterday. Our conclusions were that in an environment where you would need to set this endpoints, you would never not want to use them, and to avoid users accidentally shooting themselves in the foot by selecting the wrong endpoint, this is going to be added as a controller start up flag, with the appropriate envsubst options in infrastructure-components to support this via clusterctl.

randomvariable:
Also, given we've bumped the SDK versions, can you take a look at https://github.com/aws/aws-sdk-go/blob/master/CHANGELOG.md#release-v1343-2020-08-12 as it looks like there's been some additions with relation to FIPS bits.

randomvariable (commented):
some minor comments and suggestions

randomvariable:
Number of things:

Slight misunderstanding of the configuration format. Basically wanted to dedupe the signing region, so you could do:

`us-iso:ec2=https://special-source.amazonaws.com,elb=https://foobar.com,sts=https://blahblah.com,gb-iso:ec2=https://ec2.very-normal-country.co.uk,elb=https://blahblah.com`

Additionally, the parsing should be moved from main.go into `pkg` somewhere, and then have unit tests.

I have an example of what this might look like in https://github.com/randomvariable/cluster-api-provider-aws/commit/543f916e5bbbeaf346c8533799b5d528afb00da5

Promaethius:
@randomvariable changes made before I rush off to the airport, additionally rebased #1924 on it. Thanks for the heavy lifting!

pkg/cloud/scope/session.go
```diff
@@ -20,18 +20,40 @@ import (
...
+		for _, s := range endpoint {
+			if service == s.ServiceID {
+				return endpoints.ResolvedEndpoint{
+					URL:           s.URL,
+					SigningRegion: s.SigningRegion,
+				}, nil
+			}
+		}
+		return endpoints.DefaultResolver().EndpointFor(service, region, optFns...)
+	}
```
randomvariable, line 52: Move this anonymous function out. Might be the former ruby dev in me, but easier to unit test this way.

pkg/cloud/endpoints/endpoints.go
```diff
@@ -0,0 +1,122 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
randomvariable, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors.
```

I know it's my copy-pasta, soz.

randomvariable (commented):
Thanks for this. Some minor changes.

Also, can we please wire this up in the EKS and bootstrap controllers, so we have consistency across them?

Promaethius:
I backtracked the session handler functionality through both EKS and
Machine so it should be working for both.

As for the resolver, this is pretty much lifted out from the aws resolver
example.

randomvariable:
thanks @Promaethius . could you also update the copyright header to the correct year?

randomvariable:
> I backtracked the session handler functionality through both EKS and Machine so it should be working for both.
Need to add the command line flag to the respective main.go .

Promaethius:
> > I backtracked the session handler functionality through both EKS and Machine so it should be working for both.
> 
> Need to add the command line flag to the respective main.gos in `bootstrap/eks/main.go` and `controlplane/main.go` .

@randomvariable Added it to `controlplane/main.go` but couldn't find a place for AWS Service Context in `bootstrap/eks/main.go` looks like it only operates against Kubernetes objects so service endpoints is useless there.

randomvariable:
Can you rebase the branch rather than merging master into it?

randomvariable:
`git rebase HEAD~2` should do it

randomvariable:
I think there's something wrong with the resultant bootstrap script. Please take a look, and verify with a local run against AWS.

Promaethius:
> I think there's something wrong with the resultant bootstrap script. Please take a look, and verify with a local run against AWS.

the bash if statement didn't evaluate correctly if there weren't spaces around the quotes :P

randomvariable:
Ah, the classic `[` is actually a program.

Promaethius:
@randomvariable e2e tests passed :D

PR #1864, opened 2020-08-08
✨ Custom Bastion AMI

### LINKED ISSUE #1441: Allow bastion host to use user specified image
/kind feature

Allow bastion host to use user specified image even in the managed mode. I can use my images for control plane and worker nodes today but not bastion host in managed mode. 

**Environment:**

- Cluster-api-provider-aws version: 0.4.6

Promaethius:
Allows us to specify a custom bastion AMI image. Useful for private partitions in AWS.

Fixes #1441

### DIFF
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -104,6 +104,11 @@ type Bastion struct {
 	// Cluster API Provider AWS will use t3.micro for all regions except us-east-1, where t2.micro
 	// will be the default.
 	InstanceType string `json:"instanceType,omitempty"`
+
+	// InstanceImage will use a specified AMI to boot the bastion. If not specified,
+	// the image will default to one picked out in public space.
+	// +optional
+	AMI string `json:"ami,omitempty"`
 }
 
 // AWSLoadBalancerSpec defines the desired state of an AWS load balancer
--- a/pkg/cloud/services/ec2/bastion.go
+++ b/pkg/cloud/services/ec2/bastion.go
@@ -76,7 +76,7 @@ func (s *Service) ReconcileBastion() error {
 				return errors.Wrap(err, "failed to patch conditions")
 			}
 		}
-		instance, err = s.runInstance("bastion", s.getDefaultBastion(s.scope.Bastion().InstanceType))
+		instance, err = s.runInstance("bastion", s.getDefaultBastion(s.scope.Bastion().InstanceType, s.scope.Bastion().AMI))
 		if err != nil {
 			record.Warnf(s.scope.InfraCluster(), "FailedCreateBastion", "Failed to create bastion instance: %v", err)
 			return err
@@ -151,7 +151,7 @@ func (s *Service) describeBastionInstance() (*infrav1.Instance, error) {
 	return nil, awserrors.NewNotFound("bastion host not found")
 }
 
-func (s *Service) getDefaultBastion(instanceType string) *infrav1.Instance {
+func (s *Service) getDefaultBastion(instanceType, ami string) *infrav1.Instance {
 	name := fmt.Sprintf("%s-bastion", s.scope.Name())
 	userData, _ := userdata.NewBastion(&userdata.BastionInput{})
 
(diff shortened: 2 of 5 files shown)


api/v1alpha3/awscluster_types.go
```diff
@@ -104,6 +104,11 @@ type Bastion struct {
 	// Cluster API Provider AWS will use t3.micro for all regions except us-east-1, where t2.micro
 	// will be the default.
 	InstanceType string `json:"instanceType,omitempty"`
+
+	// InstanceImage will use a specified AMI to boot the bastion. If not specified,
+	// the image will default to one picked out in public space.
+	// +optional
+	InstanceImage string `json:"instanceImage,omitempty"`
```
randomvariable, line 111:
```suggestion
	AMI string `json:"ami,omitempty"`
```

or should we actually have the same interface as awsmachine_types.go, i.e.

```suggestion
	AMI AWSResourceReference `json:"ami,omitempty"`
```

WDYT?

detiber: +1 to using AMI here for naming rather than InstanceImage

@randomvariable Unless the intent is to honor all the fields of AWSResourceReference, I don't think we should add additional uses of it.

randomvariable: Fair.

pkg/cloud/services/ec2/bastion.go
```diff
@@ -151,7 +151,7 @@ func (s *Service) describeBastionInstance() (*infrav1.Instance, error) {
 	return nil, awserrors.NewNotFound("bastion host not found")
 }
 
-func (s *Service) getDefaultBastion(instanceType string) *infrav1.Instance {
+func (s *Service) getDefaultBastion(instanceType string, instanceImage string) *infrav1.Instance {
```
ncdc, line 154:
```suggestion
func (s *Service) getDefaultBastion(instanceType, ami string) *infrav1.Instance {
```

pkg/cloud/services/ec2/bastion.go
```diff
@@ -171,10 +171,14 @@ func (s *Service) getDefaultBastion(instanceType string) *infrav1.Instance {
 		}
 	}
 
+	if instanceImage == "" {
```
ncdc, line 174:
```suggestion
	if ami == "" {
```

pkg/cloud/services/ec2/bastion.go
```diff
@@ -171,10 +171,14 @@ func (s *Service) getDefaultBastion(instanceType string) *infrav1.Instance {
...
+		instanceImage = s.defaultBastionAMILookup(s.scope.Region())
```
ncdc, line 175:
```suggestion
		ami = s.defaultBastionAMILookup(s.scope.Region())
```

pkg/cloud/services/ec2/bastion.go
```diff
@@ -171,10 +171,14 @@ func (s *Service) getDefaultBastion(instanceType string) *infrav1.Instance {
...
+	}
+
 	i := &infrav1.Instance{
 		Type:       instanceType,
 		SubnetID:   subnet.ID,
-		ImageID:    s.defaultBastionAMILookup(s.scope.Region()),
+		ImageID:    instanceImage,
```
ncdc, line 181:
```suggestion
		ImageID:    ami,
```

ncdc:
Do we want/need to allow the user to be able to specify custom user data for the bastion VM, and if so, do we want to do it as part of this PR or a follow-up?

randomvariable:
> Do we want/need to allow the user to be able to specify custom user data for the bastion VM, and if so, do we want to do it as part of this PR or a follow-up?

Never been convinced in the utility of that bastion user data script, almost would prefer to get rid of it.

ncdc:
@randomvariable I'm fine doing it separately and not in this PR.

@Promaethius do you think you'll be able to make the requested changes soon?

randomvariable:
Looks like the rebase has gone a bit awry and is pulling in lots of other PRs.

Promaethius:
Yeah I messed up somewhere. I'm going to rebuild the MR with patches today

randomvariable:
Because we're adding a new field to v1alpha3, we need an equivalent round-trip conversion in https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/api/v1alpha2/awscluster_conversion.go#L31:L31

Something along the lines of

`dst.Spec.Bastion.AMI = restored.Spec.Bastion.AMI`

should then succeed when you run `make test`

Promaethius:
That worked. Any ideas on why verify is failing?

randomvariable:
looks good. can you squash these commits, and then i'll lgtm it.

randomvariable:
if you're having trouble with `make generate`, then we do have a current issue with the controller gen tools.
You'll need to do the following:

Explicitly set `GOPATH` even though it's not technically needed any more.
Make sure the repo is checked out to ${GOPATH}/src/sigs.k8s.io/cluster-api-provider-aws

Promaethius:
@randomvariable done

PR #1866, opened 2020-08-10
✨ [cloud/ec2] Support AWSMachineSpec.PublicIP=true

dilyevsky:

Currently setting `.publicIP` is ignored when allocating ec2 instance
subnet. Support allocating public subnet when the above field is set (as
already specified in the API definition).

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -226,6 +226,7 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
 // - subnet based on filters in machine configuration
 // - subnet based on the availability zone specified,
 // - default to the first private subnet available.
+// Filters above subnets for public if .publicIP is specified in machine spec.
 func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
 	// Check Machine.Spec.FailureDomain first as it's used by KubeadmControlPlane to spread machines across failure domains.
 	failureDomain := scope.Machine.Spec.FailureDomain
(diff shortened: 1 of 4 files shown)


dilyevsky:
Actually reading the API definition more carefully (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/162b70cb36d0c9d35a230265f3ba6b626c0cbfd8/api/v1alpha3/awsmachine_types.go#L73) that required more aggressive subnet matching so updated the logic to reflect that. Also made the test a little more useful. Now ready for review.

dilyevsky:
@justinsb @detiber did you have a chance to take a look?

pkg/cloud/services/ec2/instances.go
```diff
@@ -226,30 +226,63 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
...
-		subnetInput := &ec2.DescribeSubnetsInput{
+		subnet := scope.AWSMachine.Spec.Subnet
+
+		// Describe subnet to check for public IP support.
+		out, err := s.EC2Client.DescribeSubnets(&ec2.DescribeSubnetsInput{
+			SubnetIds: []*string{subnet.ID},
 			Filters: []*ec2.Filter{
-				filter.EC2.SubnetStates(ec2.SubnetStatePending, ec2.SubnetStateAvailable),
-				filter.EC2.VPC(s.scope.VPC().ID),
+				filter.EC2.PublicSubnets(),
```
randomvariable, line 248: Does this not result in filtering for public subnets regardless of whether or not the user wanted or not?

I'm slightly cautious of adding extraneous EC2 calls, so can we do the following:
* Break this logic out to a separate function so it's easier to read
* Conditionally do the check based on whether or not the user specified wanting public IP support or not
* add test cases for the new function

ncdc: @dilyevsky have you had a chance to see @randomvariable's question above?

dilyevsky: > Does this not result in filtering for public subnets regardless of whether or not the user wanted or not?

@randomvariable good catch, thanks - I need extra condition for `len(out.Subnets) == 0` check.

> I'm slightly cautious of adding extraneous EC2 calls, so can we do the following:

I'll look into how I can possibly cache the result from previous calls...

Working on tests now

PR #1876, opened 2020-08-19
[WIP] Add ability to sync aws machine and node labels

alexander-demichev:
Add the ability to sync AWS machine and node labels.
I created a draft PR for so we can start the discussion about implementation details. The easiest way to sync node labels is probably to use `ClusterCacheTracker`, similar to what MHC does, get remote client, and sync all labels that start with "cluster.x-k8s.io", so users can understand where the label comes from.
I'll add unit tests after we agree on implementation.

### DIFF
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/base64"
 	"fmt"
+	"strings"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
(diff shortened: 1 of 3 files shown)


pkg/cloud/scope/machine.go
```diff
@@ -181,6 +182,42 @@ func (m *MachineScope) SetAnnotation(key, value string) {
 	m.AWSMachine.Annotations[key] = value
 }
 
+// SyncMachineNodeLabel sets a key value label on the AWSMachine node.
+func (m *MachineScope) SyncMachineNodeLabel(ctx context.Context, remoteClient client.Client) error {
```
randomvariable, line 186: Is `SyncMachineNodeLabel` really accurate as a function name?

Looks like we only ever copy labels if they have the correct prefix. There's no ability to remove these labels.

alexander-demichev:
Closing this PR in favor of https://github.com/kubernetes-sigs/cluster-api/pull/3668

PR #1878, opened 2020-08-19
 ✨Add support for multiple block device mappings

alexander-demichev:
In order to avoid having the main root partition to be fulled out by logs or empty dir pods data or docker images, user can prefer to have a dedicated partition for each. This PR adds support for creating multiple block device mappings. 
Happy to hear what others think about this change.


rudoi:
/approve

thanks for this!

PR #1879, opened 2020-08-19
✨ clusterawsadm: Add actions to IAM policy required for EKS

### LINKED ISSUE #1780: clusterawsadm needs updating for EKS support
/kind feature

The `controllers.cluster-api-provider-aws.sigs.k8s.io` will need updating with permissions for EKS.

**Environment:**

michaelbeaumont:
Add IAM for managed EKS control planes

Fixes #1780


cmd/clusterawsadm/api/bootstrap/v1alpha1/types.go
```diff
@@ -98,6 +98,8 @@ type ClusterAPIControllers struct {
 	// consumed by Cluster API when creating an ec2 instance. Defaults to
 	// *.<suffix>, where suffix is defaulted to .cluster-api-provider-aws.sigs.k8s.io
 	AllowedEC2InstanceProfiles []string `json:"allowedEC2InstanceProfiles,omitempty"`
+	// EnableEKS
+	EnableEKS bool `json:"enableEKS,omitempty"`
```
randomvariable, line 102: Might be better to convert this into a struct in case additional options need to be added at a later date, so something like

```suggestion
	EKS EKSControllerConfig `json:"enableEKS,omitempty"`
```

with a struct somewhere
``` go
type EKSControllerConfig struct {
    Enable bool `json:"enable,omitempty"`
```

randomvariable (commented):
One comment, might make it easier to make non API breaking changes in the future.

Additionally, do you expect people to want the policy to specifically exclude the EC2/VPC based controller permissions? We may want to consider a follow up issue to allow more fine grained policies.

randomvariable:
need to run `make generate`, and can you squash commits as well?

randomvariable:
So sorry, I just remembered, I would also add a fixture for the expected output and add a test case in https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/master/cmd/clusterawsadm/cloudformation/bootstrap/fixtures and add a test case to https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/cmd/clusterawsadm/cloudformation/bootstrap/template_test.go

After that, think this is good to go, but would like @rudoi to answer if this is sufficient as an end user.

michaelbeaumont:
I could certainly imagine users not needing some of those permissions, I think some of the ELB permissions might be unnecessary with EKS as well? I'll defer to @richardcase here.

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
```diff
@@ -51,105 +51,160 @@ func (t Template) controllersTrustPolicy() *iamv1.PolicyDocument {
...
-					"elasticloadbalancing:DescribeTags",
-					"elasticloadbalancing:ModifyLoadBalancerAttributes",
-					"elasticloadbalancing:RegisterInstancesWithLoadBalancer",
-					"elasticloadbalancing:DeregisterInstancesFromLoadBalancer",
-					"elasticloadbalancing:RemoveTags",
+					"iam:GetRole",
+					"iam:ListAttachedRolePolicies",
+					"iam:DetachRolePolicy",
+					"iam:DeleteRole",
+					"iam:CreateRole",
```
richardcase, line 157: I wonder if we should have an additional explicit option that the user must specify to say that are going to use the "automatic IAM role per cluster" functionality when provisioning an EKS cluster (the partner to the **EKSEnableIAM** feature flag on the controller) which would then include the actions that can modify IAM:
 - iam:DetachRolePolicy
- iam:DeleteRole
- iam:CreateRole
- iam:TagRole
- iam:AttachRolePolicy

And without that option the policy just gets:
- iam:GetRole
- iam:ListAttachedRolePolicies

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
```diff
@@ -51,105 +51,160 @@ func (t Template) controllersTrustPolicy() *iamv1.PolicyDocument {
...
+					"eks:TagResource",
+					"eks:UpdateClusterVersion",
+					"eks:DeleteCluster",
+					"eks:UpdateClusterConfig",
+					"eks:UntagResource",
 				},
-				Condition: iamv1.Conditions{
-					iamv1.StringLike: map[string]string{"iam:AWSServiceName": "elasticloadbalancing.amazonaws.com"},
+				Resource: iamv1.Resources{
+					"arn:aws:eks:*:*:cluster/*",
```
richardcase, line 185: Could we restrict this down to the specific account? Although i see elsewhere that * is used. Perhaps a discussion outside the scope of this?

randomvariable: IAM can't be scoped beyond an account anyway, so this is fine. Only the trust policies control which foreign IAM roles you can assume, which will need their own roles, etc...

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
michaelbeaumont: done!

richardcase:
Great @michaelbeaumont 👍

rudoi:
/approve

Looks great. Thanks so much!

PR #1884, opened 2020-08-25
✨ Unmanaged VPCs don't require public subnets

### LINKED ISSUE #1696: Accept provided VPC infrastructure with only a private subnet
/kind feature

As a user, I may want to provide CAPA with a VPC that has no associated public subnets or internet gateways. It is unclear whether CAPA would support this today, even with an `internal` load balancing scheme for the apiserver.

The current subnet code may issue an error when reconciling an unmanaged VPC with no public subnets: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/99349b8abeb34123b2e13d025946a2e66d6d6b40/pkg/cloud/services/ec2/subnets.go#L77-L87

I can't find anywhere else we mandate public subnets outside of points-of-use, and it looks like we may not even hit that check if the VPC has more than one subnet in it. More testing would be required to validate CAPA's behavior in this scenario.

andrewmyhre:
**What this PR does / why we need it**: As a user, I may want to provide CAPA with a VPC that has no associated public subnets or internet gateways. It is unclear whether CAPA would support this today, even with an internal load balancing scheme for the apiserver.

Fixes #1696

### DIFF
--- a/pkg/cloud/services/network/subnets.go
+++ b/pkg/cloud/services/network/subnets.go
@@ -104,14 +104,21 @@ func (s *Service) reconcileSubnets() error {
 		}
 	}
 
-	// Check that we need at least 1 private and 1 public subnet after we have updated the metadata
-	if len(subnets.FilterPrivate()) < 1 {
-		record.Warnf(s.scope.InfraCluster(), "FailedNoPrivateSubnet", "Expected at least 1 private subnet but got 0")
-		return errors.New("expected at least 1 private subnet but got 0")
-	}
-	if len(subnets.FilterPublic()) < 1 {
-		record.Warnf(s.scope.InfraCluster(), "FailedNoPublicSubnet", "Expected at least 1 public subnet but got 0")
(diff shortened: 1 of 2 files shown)


randomvariable:
@andrewmyhre can you add a test case to subnets_test?

andrewmyhre:
@randomvariable I only need to update a few tests. I'll describe the changes here in case there are implications I'm not aware of:

- A couple of tests expected reconciliation to fail when an unmanaged VPC had only private subnets - now expect to succeed
- A test expected reconciliation to fail when an unmanaged VPC had subnets specified in the Spec which differ from those in the VPC. In this case reconciliation should end up with the single subnet which exists in both the Spec and the VPC, and succeed. I've removed the test because I don't believe it's actually testing anything, but maybe there's a better test I could replace it with?

andrewmyhre:
Closing in favor of #1934 which is up to date with master

randomvariable:
looks good, can we squash the commits?

PR #1901, opened 2020-08-31
 🌱Fix for Bastion struct field comment

sedefsavas:
This PR fixes Bastion struct AMI field comment.


sedefsavas:
/hold 
Will also regenerate manifests.

PR #1915, opened 2020-09-02
✨ Persists changes early when using default subnets

### LINKED ISSUE #1906: Persist default subnets early
/kind feature

When there are no subnets specified and we use the default subnet logic [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/network/subnets.go#L71) we should persist the subnet specs as soon as possible, especially before creating them. This will be helpful if there is an transient error part way thorough creating the subnets.

We should think about exiting early from the reconciliation loop after we have determined the subnets for the VPC but before actually creating them and force a requeue (i.e. `reconcile.Result{Requeue: true}`). This will persist the definitions and on the next reconciliation loop the subnets will be created from the definition.

**Environment:**

- Cluster-api-provider-aws version: 0.5.5

gab-satchi:
To address a potential issue raised [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1721#discussion_r441821606)

Needed a way to short the rest of the reconciliation after default subnets is determined. I'm not super fond of this implementation where the `requeue` is bubbled up from the ec2 services so open to suggestions. Some other alternatives considered: 
- Patch the AWSCluster like how do for conditions. Calling patch on the spec outside of the controller felt wrong. Conditions was only modifying the status field. 
- Just return after determining the default subnets and continue. This may cause issues further down the line where subnets are expected to have been created. 

Fixes #1906

### DIFF
--- a/pkg/cloud/services/network/subnets.go
+++ b/pkg/cloud/services/network/subnets.go
@@ -73,6 +73,10 @@ func (s *Service) reconcileSubnets() error {
 			record.Warnf(s.scope.InfraCluster(), "FailedDefaultSubnets", "Failed getting default subnets: %v", err)
 			return errors.Wrap(err, "failed getting default subnets")
 		}
+		// Persist the new default subnets to AWSCluster
+		if err := s.scope.PatchObject(); err != nil {
+			return err
+		}
 	}
 
 	for _, sub := range subnets {


randomvariable:
I generally run e2e on behavioural changes, so

richardcase:
Great, thanks @gab-satchi. Its difficult to know which way is best. The way you have implemented it is how it was originally [discussed](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1721#discussion_r441821606). I agree that having `requeue` bubble up out of the network service doesn't sit well. The named return value could be changed to something else like `usingDefaults` (or something better) so thats its not saying `requeue`...it just so happens at the moment it does equal requeue to the controller.

The `PatchObject` alternative you suggest would be the easiest as this could be called on the scope straight after adding the subnet specs.  As you say the only time `PatchObject` is called currently in the services (and not in the controllers) is when the conditions or a status field have been set.  So if we call `PatchObject` in the network service when updating the `spec` with the subnets its going against the current convention.

The `PatchObject` alternative feels cleaner and keeps the `reconcileSubnets/ReconcileNetwork` functions in line with all the other reconcile functions.

It would be good to get @randomvariable or @rudoi view on it.

randomvariable:
I think it's reasonable to call `PatchObject` for when the spec is updated for the same reason we do so for updating the status. The history of all this is that some of the fields are stored in spec to allow `clusterctl move` as well as backup and restore tools like Velero to function so that state can be fully reconstructed from `spec` alone.

gab-satchi:
Thanks for the input. Significantly reduced the amount of changes. 

randomvariable:
Not sure why the test is failing. Mind taking a look?
/retest

gab-satchi:
Not sure either. will take a look if it fails again

gab-satchi:
There were some stricter gomock matcher added by [these changes](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1926/files#diff-e241beb9fc7a09f68e5212ac9172b644R1197-R1219). The tag specifications isn't deterministic so introduced a failure.

randomvariable:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1937 fixes it

PR #1924, opened 2020-09-04
✨ SSM (AWS Systems Manager Parameter Store) as a Secret Backend

### LINKED ISSUE #1900: Secret Backend Support
/kind feature

Support the storage of ClusterAPI secrets in AWS Systems Manager - Secure Parameter Storage. Why? AWS Secrets Manager is not available in DoD IL6+ environments but Secure Parameter Storage is available in all environments including the Secret Region. This issue starts the conversation around using Secure Parameter Storage for the chunked secrets used by the CAPA Image.

* Boolean switch between using SPS and Secrets Manager in the ClusterAPI objects.
* `pkg/cloud/services/sps` following a similar pattern to `pkg/cloud/services/secretsmanager`
* `pkg/cloud/services/sps/secret_fetch_script.go` can be normalized between the two methods since the CAPA contract provides the sps interface

https://aws.amazon.com/compliance/services-in-scope/

**Environment:**

Promaethius:
Adds an additional secret backend for environments which do not have access to AWS Secrets Manager. SSM is available as an EC2 feature in all possible scopes: https://aws.amazon.com/compliance/services-in-scope/

Additionally, the format of SSM API and CLI calls are similar to SecretsManager so this should be a relatively painless addition.

Fixes #1900


api/v1alpha3/awscluster_types.go
```diff
@@ -82,6 +82,11 @@ type AWSClusterSpec struct {
 	// Bastion contains options to configure the bastion host.
 	// +optional
 	Bastion Bastion `json:"bastion"`
+
+	// SecretBackend is an optional string allowing the user to specify either Secrets Manger or SSM Parameter Storage.
+	// Only possible input is 'ssm', any other entry will default to Secrets Manager.
+	// +optional
```
richardcase, line 88: It might be worth adding extra kubebuilder validation here. Perhaps something like
```go
// +kubebuilder:default=secret-manager
// +kubebuilder:validation:Enum=secret-manager;parameter-store
```

api/v1alpha3/awscluster_types.go
```diff
@@ -82,6 +82,11 @@ type AWSClusterSpec struct {
...
+	SecretBackend string `json:"secretBackend,omitempty`
```
richardcase, line 89: How about creating a type for the secret backend? And then having 2 instances of that type for the secrets manager and parameters store? Like we do for the [ELB scheme](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/api/v1alpha3/types.go#L82). The values for the 2 types would need to be the same as the enum values in the validation above.

controllers/awsmachine_controller.go
```diff
@@ -296,7 +305,11 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 	machineScope.Info("Handling deleted AWSMachine")
 
 	ec2Service := r.getEC2Service(ec2Scope)
-	secretSvc := r.getSecretsManagerService(clusterScope)
+	if clusterScope.SecretBackend() == "ssm" {
```
richardcase, line 308: Do you think it would be possible to have a single function call (i.e. something like `getSecretBackend`) that has the conditional logic to return the specific secrets backend?

pkg/cloud/scope/managedcontrolplane.go
```diff
@@ -267,3 +267,8 @@ func (s *ManagedControlPlaneScope) ImageLookupOrg() string {
 func (s *ManagedControlPlaneScope) ImageLookupBaseOS() string {
 	return s.ControlPlane.Spec.ImageLookupBaseOS
 }
+
+// SecretBackend returns the secret backend to use when looking up AMIs
```
randomvariable, line 271: > // SecretBackend returns the secret backend to use when looking up AMIs

That doesn't seem right to me as a description? The AMI lookup is either EC2 API or SSM?

controllers/awsmachine_controller.go
randomvariable: Would be in favour of this. Not least I suspect we'll need to add an S3 backend too for Ignition, which will solve things for @dongsupark

api/v1alpha3/awscluster_types.go
randomvariable: Nit: secrets-manager, not secret-manager
```
// +kubebuilder:default=secrets-manager
// +kubebuilder:validation:Enum=secrets-manager;parameter-store
```

pkg/cloud/services/ssm/helpers.go
```diff
@@ -0,0 +1,30 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package ssm
+
+import "bytes"
+
+func splitBytes(data []byte, maxSize int, iterFunc func([]byte)) {
```
randomvariable, line 21: Can we dedupe this from both the secrets manager and ssm backends into a single shared library, probably in `pkg/internal/splitbytes` ?

api/v1alpha3/awscluster_types.go
randomvariable: Additionally, I feel more confortable about moving this into the Cloudinit section of the AWSMachine spec. I understand that it's a bit more verbose, but it keeps a cleaner separation of things that relate to the overall cluster and those that relate to machines.

In addition, as we add support for Ignition, none of these methods will work, so we're strongly tying the AWSCluster resource to the Linux distribution here, which feels wrong.

randomvariable (changes requested):

api/v1alpha3/awscluster_types.go
Promaethius: @randomvariable that's a good place to put it. I see that there is an option `EnableSecureSecretsManager` already present. Should this be replaced by a variable type that can encompass S3, Secrets Manager, and SSM?

randomvariable: Ideally we would, but we can't replace it until the API is revved to v1alpha4, which will be 0.7.0.

Maybe change the description of `EnableSecureSecretsManager` to be to describe enabling the use of an external service in general, and then add something like `SecureSecretsBackend` to select on type.

Promaethius:
@randomvariable struggling with the `GenerateCloudInitMIMEDocument` functionality since I moved it from being a generic function to a service specific function. Not enough experience with go/mock. Had to generalize the `SecretsManagerInterface` and the `SSMInterface` into `SecretInterface` to allow either backend in function arguments or return values.

pkg/internal/splitbytes/helpers.go
```diff
@@ -14,11 +14,11 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package secretsmanager
+package splitbytes
```
randomvariable, line 17:
```suggestion
package bytes
```

I've probably misadvised you here, as i'm sure golangci_lint is going to complain about stuttering function names.

Should probably rename the package bytes, and the function split, so we have `bytes.Split(data []byte, maxSize int, iterFunc func([]byte))`

Promaethius: Done. I'm surprised it doesn't count `bytes` as a protected namespace or something similar.

randomvariable:
> @randomvariable struggling with the GenerateCloudInitMIMEDocument functionality since I moved it from being a generic function to a service specific function. Not enough experience with go/mock:

I'll take a look as to how best to unblock you after doing the v0.6.0 release. Also, apologies if I don't recognise you in Slack, but worth joining the channel in https://kubernetes.slack.com/archives/CD6U2V71N and we can pair, zoom etc...

Promaethius:
@randomvariable took some of your advice in half and broke the MIME generation out into an external package and unit test while leaving UserData generation Service specific.

api/v1alpha3/awsmachine_types.go
```diff
@@ -138,11 +149,19 @@ type AWSMachineSpec struct {
...
 	// By default, a cloud-init boothook shell script is prepended to download
 	// the userdata from Secrets Manager and additionally delete the secret.
 	InsecureSkipSecretsManager bool `json:"insecureSkipSecretsManager,omitempty"`
 
+	// SecureSecretsBackend, when set to parameter-storer will utilize the AWS Systems Manager
+	// Parameter Storage to distribute secrets. By default or with the value of secrets-manager,
+	// will use AWS Secrets Manager instead.
+	// +optional
+	// +kubebuilder:default=secrets-manager
+	// +kubebuilder:validation:Enum=secrets-manager;parameter-store
```
randomvariable, line 162:
```suggestion
	// +kubebuilder:validation:Enum=secrets-manager;ssm-parameter-store
```

randomvariable (changes requested):
some changes, but otherwise looking good I think

api/v1alpha3/awsmachine_types.go
```diff
@@ -28,6 +28,17 @@ const (
...
 
+// SecretBackend defines variants for backend secret storage.
+type SecretBackend string
+
+var (
+	// SSM defines AWS Systems Manager Parameter Store as the secret backend
+	SSM = SecretBackend("parameter-storer")
+
+	// SecretsManager defines AWS Secrets Manager as the secret backend
+	SecretsManager = SecretBackend("secrets-manager")
```
randomvariable, line 39:
```suggestion
	SecretBackendSecretsManager = SecretBackend("secrets-manager")
```
Makes it easier to see what the var stands for when importing the package

controllers/awsmachine_controller.go
```diff
@@ -68,14 +70,29 @@ func (r *AWSMachineReconciler) getEC2Service(scope scope.EC2Scope) services.EC2M
...
+
+	return ssm.NewService(scope)
+}
+
+func (r *AWSMachineReconciler) getSecretService(machineScope *scope.MachineScope, scope cloud.ClusterScoper) services.SecretInterface {
+	if machineScope.SecureSecretsBackend() == "parameter-store" {
+		return r.getSSMService(scope)
+	}
+	return r.getSecretsManagerService(scope)
+}
```
randomvariable, line 94:
```suggestion
func (r *AWSMachineReconciler) getSecretService(machineScope *scope.MachineScope, scope cloud.ClusterScoper) (services.SecretInterface, error) {
	switch machineScope.SecureSecretsBackend() {
		case infrav1.SecretBackendSSMParameterStore {
			return r.getSSMService(scope), nil
		}
		case infrav1.SecretBackendSecretsManager {
			return r.getSecretsManagerService(scope), nil
		}
	}
	return nil, errors.New("invalid secret backend")
}
```

controllers/awsmachine_controller.go
```diff
@@ -296,7 +313,7 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 	machineScope.Info("Handling deleted AWSMachine")
 
 	ec2Service := r.getEC2Service(ec2Scope)
-	secretSvc := r.getSecretsManagerService(clusterScope)
+	secretSvc := r.getSecretService(machineScope, clusterScope)
```
randomvariable, line 316:
```suggestion
	secretSvc, err := r.getSecretService(machineScope, clusterScope)
	if err != nil {
		return ctrl.Result{}, err
	}
```

Promaethius:
with #1858 in the same release, it's important to note that this PR would require it

Promaethius:
@randomvariable depends on and refactored for #1858

randomvariable:
this lgtm, but hold remains as it also includes #1858 at present.

randomvariable:
Need to
/hold
still as this will need to be rebased?

Promaethius:
@randomvariable rebased, I pre-emptively did the unhold as I worked on the rebase my bad
/unhold

Promaethius:
Cluster isn't coming up. I suspect the userdata doesn't come through correctly

randomvariable:
note to self, review the regex for excluding blocked prefixes for secret backends.

Promaethius:
> note to self, review the regex for excluding blocked prefixes for secret backends.

On this topic, I feel that validation for secret names has to be per secret backend due to how each handles it differently. Such that,
```go
prefixRe = regexp.MustCompile(`(?i)^[\/]?(aws|ssm)[.]?`)
// SSM Validation does not allow (/)aws|ssm in the beginning of the string
prefix = prefixRe.ReplaceAllString(prefix, "")
// Because the secret name has a slash in it, whoe name must validate as a whole path
if prefix[0] != byte('/') {
	prefix = "/" + prefix
}
```

api/v1alpha3/awsmachine_webhook.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2020 The Kubernetes Authors.
```
randomvariable, line 2: Don't need to update copyright year on existing files.

Promaethius:
@randomvariable changes made, tests passing. Am I missing anything?

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_node.go
```diff
@@ -17,20 +17,36 @@ limitations under the License.
...
-		Effect: iamv1.EffectAllow,
-		Resource: iamv1.Resources{
-			"arn:*:secretsmanager:*:*:secret:aws.cluster.x-k8s.io/*",
-		},
-		Action: iamv1.Actions{
-			"secretsmanager:DeleteSecret",
-			"secretsmanager:GetSecretValue",
-		},
+func (t Template) secretPolicy(secureSecretsBackend infrav1.SecretBackend) iamv1.StatementEntry {
+	switch secureSecretsBackend {
```
richardcase, line 25: This is very similar to  cluster_api_controller.go above. Do you think it would be possible to consolidate into 1 place and add the additional actions for the controller policy?

Not essential though.

richardcase:
This seems ok to me, small comment about the policies but generally

/lgtm

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_node.go
Promaethius: This kinda thing pops up a lot in the clusterctl code, I think it can be handled in a separate PR as a mini refactor?

randomvariable (approved):

randomvariable:
One last `git pull upstream --rebase` to account for the addition of the machinepool webhook

This should do it.

``` yaml
---
apiVersion: admissionregistration.k8s.io/v1beta1
kind: MutatingWebhookConfiguration
metadata:
  creationTimestamp: null
  name: mutating-webhook-configuration
webhooks:
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /mutate-infrastructure-cluster-x-k8s-io-v1alpha3-awscluster
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: default.awscluster.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsclusters
  sideEffects: None
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /mutate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine
  failurePolicy: Fail
  name: mutation.awsmachine.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsmachines
  sideEffects: None
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /mutate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachinepool
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: default.awsmachinepool.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsmachinepools
  sideEffects: None
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /mutate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmanagedmachinepool
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: default.awsmanagedmachinepool.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsmanagedmachinepools
  sideEffects: None

---
apiVersion: admissionregistration.k8s.io/v1beta1
kind: ValidatingWebhookConfiguration
metadata:
  creationTimestamp: null
  name: validating-webhook-configuration
webhooks:
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awscluster
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: validation.awscluster.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsclusters
  sideEffects: None
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachine
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: validation.awsmachine.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsmachines
  sideEffects: None
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachinetemplate
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: validation.awsmachinetemplate.infrastructure.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsmachinetemplates
  sideEffects: None
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmachinepool
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: validation.awsmachinepool.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsmachinepools
  sideEffects: None
- clientConfig:
    caBundle: Cg==
    service:
      name: webhook-service
      namespace: system
      path: /validate-infrastructure-cluster-x-k8s-io-v1alpha3-awsmanagedmachinepool
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: validation.awsmanagedmachinepool.infrastructure.cluster.x-k8s.io
  rules:
  - apiGroups:
    - infrastructure.cluster.x-k8s.io
    apiVersions:
    - v1alpha3
    operations:
    - CREATE
    - UPDATE
    resources:
    - awsmanagedmachinepools
  sideEffects: None
```

randomvariable:
@detiber asks:
> Not sure if this was discussed previously, but is there a specific use case in mind for having this as an API field rather than a configuration option on the controller manager?
> Is there a specific use case in mind where one would want to choose different backends for individual Clusters/Machines vs having it a global configuration for an instance of the infrastructure provider?

In retrospect, yes, maybe we could have done at the controller flag level. I think we should be working towards finding a better solution overall for v1alpha4, given the issues we've run into with cloud-init.

The one use case I did however have in mind for having it on the AWSMachineTemplate is for FlatCar Linux which doesn't support multi-part MIME docs, but where Ignition does natively support S3.

PR #1925, opened 2020-09-07
 ✨ feat: default iam role for eks

### LINKED ISSUE #1877: Create default EKS IAM Role using clusterawsadm
/kind feature

It would help users if a default IAM role was created for EKS. This could then be used in the EKS control plane if no role was specified and if the EKSEnableIAM flag is false (i.e. not specified). It feels like this could be done using clusterawsadm.

The default role should have a tag that denotes it as the default EKS role. This would then enable customers to create their own role with the tag and this be used instead.

We should also allow specifying addition polices when creating the role 

**Environment:**

### LINKED ISSUE #1908: Update EKS service to use default role for EKS control plane
/kind feature

When reconciling the **AWSManagedControlPlane** we should use the default IAM role for EKS if:

- No `roleName` is specified AND
- the `EKSEnableIAM` feature flag isn't enabled

A default role will be created by `clusterawsadm` (see issue #1877). When looking up the default role we need to use a label on a IAM role to allow users of CAPA to create their own default role if they want.

This could be done in parallel to #1877 as long as there is agreement on the name of the label.

**Environment:**

- Cluster-api-provider-aws version: > 0.6.0

richardcase:

A default role can be created by `clusterawsadm` for use by the eks control plane. The role creation needs to be enabled using a config file. The default role will be used using AWSManagedControlPlane reconciliation if the following are both true:

* Flag **EKSEnableIAM** is not enabled
* No role name is supplied in the AWSManagedControlPlane spec

This is so that dynamic role creation can be disabled in the controller and the IAM role required for EKS can be created ahead of time by `clusterawsadm` which enables users of the EKS functionality to run the controller with reduced privileges...especially IAM.

Fixes #1908 
Fixes #1877


docs/prerequisites.md
```diff
@@ -54,6 +55,30 @@ spec:
...
+#### With EKS Support
+
+If you want to use the the EKS support in the provider then you will need to enable these features via the configuration file. For example:
+
+```yaml
+apiVersion: bootstrap.aws.infrastructure.cluster.x-k8s.io/v1alpha1
+kind: AWSIAMConfiguration
+spec:
+  bootstrapUser:
+    enable: true
```
randomvariable, line 67:
```suggestion
```

Don't necessarily need the bastion for EKS do you?

richardcase: Good point, thats not required. I have removed it.

PR #1931, opened 2020-09-08
✨ Allow specifying subnet IDs for control plane load balancers

andrewmyhre:
**What this PR does / why we need it**: Presently it is not possible to specify which subnets the control plane should use. CAPA will discover and use all available subnets in the VPC. Some organizations have specific subnets designated for particular traffic e.g: service vs workloads need to be able to specify these in the cluster spec. 

Addresses #1715

Also related: #1726


randomvariable:
some of the files need go fmt'ing.

`make lint` should help here.

randomvariable:
looking good. could you squash these commits?

andrewmyhre:
> looking good. could you squash these commits?

@randomvariable Done!

PR #1860, opened 2020-08-05
✨ Implement AWSMachinePools

mytunguyen:
**What this PR does / why we need it**:

This PR implements machine pools for CAPA using AWS Auto Scaling Groups. In this PR, an /exp path is added similar to CAPI to be the home for the AWSMachinePool API and controller.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1784

PR #1936, opened 2020-09-09
:bug: Add capi exp schema and fix manager args and rbac

mytunguyen:
* eks-bootstrap doesn't know about the CRD `machinepools.exp.cluster.x-k8s.io` and `User "system:serviceaccount:capa-eks-bootstrap-system:default" cannot list resource "machinepools" in API group "exp.cluster.x-k8s.io"`
* I want to be able to use `"EXP_MACHINE_POOL": "true"` like in the capi-controller-manager and capa-controller-manager

This needs to be merged so that awsmachinepools can connect to an eks controlplane
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1860

### DIFF
--- a/bootstrap/eks/config/manager/kustomization.yaml
+++ b/bootstrap/eks/config/manager/kustomization.yaml
@@ -5,3 +5,4 @@ patchesStrategicMerge:
 - manager_image_patch.yaml
 - manager_pull_policy.yaml
 - manager_auth_proxy_patch.yaml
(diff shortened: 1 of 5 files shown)


randomvariable:
Think you'll need to add a patch for this otherwise controller-gen will keep fighting.

mytunguyen:
sweet! i think i figured it out

PR #1943, opened 2020-09-11
 ✨ chore: changed template name from managed to eks

### LINKED ISSUE #1942: Change name of the template used for EKS
/kind feature

The template for use with EKS is currently called `cluster-template-managed.yaml` and so the flavor is `managed`. When we introduce different flavors of the EKS template in the future the flavor name may stutter if we still use `managed` and introduce a template for EKS with managed node pools...we may get `cluster-template-managed-managedpool.yaml` or something similar. It doesn't feel right.

I think we should change the flavor from `managed` to `eks`. Then in the future we can have sub flavors of eks like `eks-machine-pool` or `eks-managed-machine-pool`.

**Environment:**

richardcase:
The template flavor used for EKS has been renamed from `managed` to`eks`. This is to enable different flavors of eks templates in the future whose names may stutter without this change, especially when we introduce managed machine pools. It also makes it more obvious what the flavor is for with `eks`.

The eks docs have been updated to use the new flavor and also the release doc.

Additionally a new target for release will copy the templates to the release directory.

Fixes #1942


randomvariable:
That makes more sense to me anyway
/lgtm
/approve

Issue #1896, opened 2020-08-28
Ensure clusterawsadm is documented


randomvariable:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]
clusterawsadm needs full documentation, including all configuration options for IAM.

Consider creating a site as is being done in for CAPZ:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/pull/858

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

PR #1947, opened 2020-09-14
:book: Add netlify configuration

randomvariable:

Part of #1896 

* Refactored the makefiles a fair amount so that the top-level one is cleaner, and there's more code-reuse.
* Added the mdbook bits
* Added the initial site structure


randomvariable:
/hold

have a slight error in the makefile

randomvariable:
/unhold

I think this is good to go, but I might need to chase up the TLS certificate for cluster-api-aws.sigs.k8s.io

randomvariable:
Modifed the makefile substantially, so just wanna check it all works.

randomvariable:
make targets do appear to work correctly.

rebased.

Makefile
```diff
@@ -12,29 +12,19 @@
...
-# Active module mode, as we use go modules to manage dependencies
-export GO111MODULE=on
+# If you update this file, please follow
+# https://suva.sh/posts/well-documented-makefiles
 
 # Directories.
 ARTIFACTS ?= $(REPO_ROOT)/_artifacts
 TOOLS_DIR := hack/tools
+TOOLS_DIR_DEPS := $(TOOLS_DIR)/go.sum $(TOOLS_DIR)/go.mod $(TOOLS_DIR)/Makefile
 TOOLS_BIN_DIR := $(TOOLS_DIR)/bin
```
ncdc, line 26: These are now in common.mk - do you need them here too?

docs/book/src/introduction.md
```diff
@@ -0,0 +1 @@
+{{#include ../../../README.md:2:}}
```
ncdc, line 1: What does the 2 mean?

ncdc (commented):
Just a couple of minor questions/comments but nothing you have to change before getting this merged if you don't want to. Thanks for doing this!!!

docs/book/src/introduction.md
randomvariable: Ignore the first two lines, so we don't get

```
# Cluster API

# Cluster API
```
on the front page.

PR #1950, opened 2020-09-16
:seedling: Update to controller runtime 0.5.11

randomvariable:

Fixes #


ncdc:
FWIW, the memory leak we fixed in controller-runtime only really occurs if you're dynamically adding a watch to a controller, and then you close/stop the cache (shared informers) associated with the that watch (e.g. when watching something from a workload cluster, **not** from the Manager's client/cache). It's fine to upgrade here, but it's not strictly required for CAPA from a memory leak perspective, I don't think.

vincepri:
+1 to what Andy said, although we can merge this now if we want to ship CAPA with the fix earlier than CAPI

/approve

CecileRobertMichon:
should we be doing this in CAPZ as well?

randomvariable:
> should we be doing this in CAPZ as well?

Won't hurt.

PR #1952, opened 2020-09-16
🐛 Retry with listing all ELBs when listing by tag fails

### LINKED ISSUE #1935: Remove criticality from code paths using APIs not available as VPC endpoints
/kind bug

If operating in an AWS account without internet connectivity, access to AWS APIs may only be available via VPC endpoints in some environments.

In this scenario, some operations fail, e.g. cluster deletes do not succeed because ELBs cannot be listed via the Resource Tagging API, which is not endpoint enabled.

Cluster creates, scales, and deletes work.

Review code for using VPC endpoint APIs and provide workaround where needed.

**Environment:**

- Cluster-api-provider-aws version: 0.5.5

/help

gab-satchi:
GetResources (with tags) aws call fails in an air gapped environment that relies on VPC endpoints. We were using that API for retrieving the two types of loadbalancers during cluster delete: 
- API server loadbalancer
- any loadbalancers created by the cloud provider for loadbalancer type services. 

With these changes we: 
- delete API server LB by name instead of searching by name
- fetch all LBs when the `GetResources` call fails and filter the ones created by the cloud provider using tags and delete those.

Fixes #1935


randomvariable:
Looks like there were connectivity issues between Prow and AWS

We should maybe as follow up add an issue to add a test for deleting service type load balancer.

gab-satchi:
> We should maybe as follow up add an issue to add an e2e test for deleting service type load balancer.

[This test](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/test/e2e/aws_test.go#L129-L143) should cover that. Unless you meant one that would go into the fallback codepath

randomvariable:
> This test should cover that. Unless you meant one that would go into the fallback codepath

Nope, that's fine, given the constraint of https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1935#issuecomment-693645683

PR #1953, opened 2020-09-17
🌱 Verify e2e code compiles as part of pull-cluster-api-provider-aws-build

### LINKED ISSUE #1429: CI presubmits should make sure that test/e2e/... compiles
#1426 passed all presubmits but broke compilation of test/e2e. We should verify that all the test code can build and compile in our presubmits (but leave running the tests to other jobs).

varunpurohit76:
added compile-e2e

Fixes #1429

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -179,6 +179,10 @@ manager-eks-bootstrap:
 clusterawsadm: ## Build clusterawsadm binary.
 	go build -ldflags "$(LDFLAGS)" -o $(BIN_DIR)/clusterawsadm ./cmd/clusterawsadm
 
+.PHONY: compile-e2e
+compile-e2e:
+	go test -c -o /dev/null -tags=e2e ./test/e2e
+
 ## --------------------------------------
 ## Tooling Binaries
 ## --------------------------------------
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -160,7 +160,7 @@ test-conformance-fast: ## Run clusterctl based conformance test on workload clus
 ## Binaries
 ## --------------------------------------
 .PHONY: binaries
-binaries: managers clusterawsadm ## Builds and installs all binaries
+binaries: managers clusterawsadm compile-e2e ## Builds and installs all binaries
```
ncdc, line 163: I think it might be better to remove this and instead update scripts/ci-build.sh to have it `make binaries compile-e2e`. WDYT?

varunpurohit76: could be done that way as well if we want to keep compilation of test separate.
Question is, should `.PHONY: binaries` make sure that all essential items compile or we want to keep that control outside?

ncdc: I would consider `binaries` to cover the executables that are user-facing and not include testing.

varunpurohit76: makes sense. will push the update.

ncdc:
LGTM. Could you please squash your commits?

varunpurohit76:
@ncdc all squashed into a single commit.

varunpurohit76:
@ncdc @randomvariable let me know if any input is required from my side..

PR #1957, opened 2020-09-19
🐛 fix: bastion reconcilation and connection error when using `eks` flavor

### LINKED ISSUE #1956: Creating a bastion with the `eks` flavor fails
/kind bug

Create a kind cluster, init the aws and eks-bootstrap providers. Render the `eks` flavour (i.e. cluster-template-eks.yaml) and then add fields to enable the bastion. For example the **AWSManagedControlPlane** definition should look like this:

```yaml
kind: AWSManagedControlPlane
metadata:
  name: "capi-managed-test-control-plane"
spec:
  region: "eu-west-2"
  sshKeyName: "capi-management-2"
  version: "v1.17.0"
  bastion:
    enabled: true
```

When this is applied reconciliation of the **AWSManagedControlPlane** fails with the following error:

```bash
[manager] E0919 04:35:36.278342       8 controller.go:248] controller-runtime/controller "msg"="Reconciler error" "error"="failed to reconcile bastion host for AWSManagedControlPlane default/capi-managed-test-control-plane: failed to run instance: MissingParameter: When specifying a security group you must specify a group id for each item\n\tstatus code: 400, request id: a22fa765-cdaf-4de6-9993-8ae99b0565d4" "controller"="awsmanagedcontrolplane" "name"="capi-managed-test-control-plane" "namespace"="default"
```

A would expect the reconciliation to be successful and the bastion instance to be created.

Found as part of investigating #1954 

**Environment:**

- Cluster-api-provider-aws version: 0.6.0
- Kubernetes version: (use `kubectl version`): 1.17

### LINKED ISSUE #1958: AWSCluster defaults/validation are not applied in the managed side
/kind bug

The defaults/validation that are done as part of the [AWSCluster webhook](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/api/v1alpha3/awscluster_webhook.go) aren't applied to the managed/eks side of the provider. For example if you enable a bastion in the **AWSManagedControlPlane** spec then the default for the AllowedCIDRBlocks isn't set and so you can't SSH to the bastion. 

Most of the same defaults and validation should apply to both umanaged/managed sides of the provider, especially for the network service.

Found as part of investigating #1954 

**Environment:**

- Cluster-api-provider-aws version: 0.6.0

richardcase:
The bastion was being reconciled before the security groups and this based the bastion reconciliation to fail. The reconciliation order has been updated.

Fixes #1956
Fixes #1958

### DIFF
--- a/api/v1alpha3/awscluster_webhook.go
+++ b/api/v1alpha3/awscluster_webhook.go
@@ -18,7 +18,6 @@ package v1alpha3
 
 import (
 	"fmt"
-	"net"
 	"reflect"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -50,7 +49,7 @@ var (
 func (r *AWSCluster) ValidateCreate() error {
 	var allErrs field.ErrorList
 
-	allErrs = append(allErrs, r.validateAllowedCIDRBlocks()...)
+	allErrs = append(allErrs, r.Spec.Bastion.Validate()...)
 
 	return aggregateObjErrors(r.GroupVersionKind().GroupKind(), r.Name, allErrs)
 }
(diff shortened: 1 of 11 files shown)


richardcase:
Changing back to WIP to include other bastion related fixes

randomvariable:
I assume this will need to be rebased on #1949

richardcase:
> I assume this will need to be rebased on #1949

Yes it will....thanks for approving the refactor.

PR #1962, opened 2020-09-23
🐛 fix: adding node to EKS cluster fails

### LINKED ISSUE #1870: AWSManagedControlPlane controller should set up aws-auth ConfigMap automatically
/kind feature

By default, the only IAM entity that has access to an EKS cluster is the entity that created it. Users must create an `aws-auth` ConfigMap specifying additional mappings. See: https://docs.aws.amazon.com/eks/latest/userguide/add-user-role.html

Additional mappings must also be added for instance roles on EC2 instances looking to join the cluster as Nodes. 

It would be great if:
1. this ConfigMap were created for me automatically
2. I could specify in the `spec` of AWSManagedControlPlane an IAM role to be added to the cluster as an admin
3. I could specify in the `spec` of AWSManagedControlPlane an InstanceRole to be added to the cluster as a node

cc @richardcase @mytunguyen

### LINKED ISSUE #1954: Error adding node with `eks` flavor
/kind bug

Created a kind cluster, did `tilt up` with a tilt config that enabled the aws and eks-bootstratp providers. Then used the [cluster-template-eks.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/templates/cluster-template-eks.yaml) to create a workload cluster.

The EKS cluster is created and other AWS resources like the EC2 instance for the worker node. The `AWSMachine` is in a `running` state. However, there is an error reported reported by the capi controller:

```bash
[manager] E0918 13:20:09.701862       8 machine_controller.go:267] controllers/Machine "msg"="Reconciliation for Machine asked to requeue" "error"="cannot assign NodeRef to Machine \"capi-managed-test-md-0-9f94564f4-64bv9\" in namespace \"default\", no matching Node: requeue in 20s" "cluster"="capi-managed-test" "machine"="capi-managed-test-md-0-9f94564f4-64bv9" "namespace"="default" 
```

And then if you use the `user` kubeconfig to query the EKS cluster there is no node `kubectl --kubeconfig=./managed-test.kubeconfig get nodes`

I expect there to be no error and the node to be listed when querying.

**Environment:**

richardcase:
When using the `eks` flavor to create a cluster the node isn't added to the cluster. This is due due to a mixture of missing IAM permissions on the nodes role and also not generating the `aws-auth` config map to authorize the node. This paves the way to fix #1870 in the future.

Fixes #1954


.golangci.yml
```diff
@@ -24,6 +24,7 @@ issues:
     - "G108: Profiling endpoint is automatically exposed on /debug/pprof"
 run:
   timeout: 10m
+  tests: false
```
randomvariable, line 27: Why did we do this?

richardcase: Mainly to stop linting failures in tests....shall i revert it?

richardcase: Its a bit lazy but it sops having to add `//nolint` in tests

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_node.go
```diff
@@ -48,6 +48,19 @@ func (t Template) sessionManagerPolicy() iamv1.StatementEntry {
...
 }
 
+func (t Template) nodeManagedPolicies() []string {
+	polices := t.Spec.Nodes.ExtraPolicyAttachments
+
+	if t.Spec.Nodes.EKS.Enable {
+		polices = append(polices,
+			"arn:aws:iam::aws:policy/AmazonEKSWorkerNodePolicy",
+			"arn:aws:iam::aws:policy/AmazonEKS_CNI_Policy",
+			"arn:aws:iam::aws:policy/AmazonEC2ContainerRegistryReadOnly")
```
randomvariable, line 58: `AmazonEC2ContainerRegistryReadOnly` would be a pretty useful toggle on the nodes level as it's not EKS specific.

If `AmazonEKS_CNI_Policy` is basically equivalent to that required by the VPC CNI, maybe also at the node level (and then use defaulter-gen to enable it when EKS is on for nodes).

randomvariable (commented):
Some comments and suggestions.

.golangci.yml
randomvariable: If we're doing it in lots of places, I guess that's fine.

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_node.go
richardcase: Good thinking, i'll make those changes.

randomvariable:
/approve

/assign @rudoi 
for lgtm

rudoi:
/lgtm

Tested via Tilt and it worked 👍🏻

PR #1992, opened 2020-09-30
🐛 Add externalManagedControlPlane Status to fix node drain under EKS

dthorsen:
**What this PR does / why we need it**: When using the AWSManagedControlPlane, if you delete a Machine resource it does not gracefully cordon and drain the associated Node. The fix for this in CAPI is to have the managed control plane provider set the `externalManagedControlPlane: true` value in its status to signal to cluster-api that the control plane Nodes are not visible in the cluster.

Related to: https://github.com/kubernetes-sigs/cluster-api/issues/3631


vincepri:
/ok-to-test
/approve
/assign @randomvariable 

/cc @alexeldeib @CecileRobertMichon 
(for CAPZ)

PR #1995, opened 2020-10-01
✨ feat: enable additional user & role mapping for aws-iam-authenticator

### LINKED ISSUE #1850: AWSManagedCluster should support granting additional IAM roles access to the EKS cluster
/kind feature

By default, only the IAM entity that has access to an EKS cluster is the entity that created it. Additional IAM roles must be added manually using an `aws-auth` ConfigMap in the `kube-system` namespace. 

It would be great if we could specify additional roles (maybe users, too?) in the spec and have that ConfigMap be set up automatically.

Find information about the format of the auth configuration here: https://github.com/kubernetes-sigs/aws-iam-authenticator#full-configuration-format

richardcase:
The EKS control plane has been extended so that you can specifyadditional IAM user and role mappings for use when generating the config for `aws-iam-authenticator`. This can be used to enabled additional users to have access to the cluster
without any additional work.

Fixes #1850


controlplane/eks/api/v1alpha3/types.go
```diff
@@ -83,3 +83,39 @@ var (
...
+// RoleMapping represents a mapping from a IAM role
+type RoleMapping struct {
+	// RoleARN is the AWS ARN for the role to map
+	// +kubebuilder:validation:MinLength:=31
+	RoleARN string `json:"rolearn"`
+	// KubernetesMapping holds the RBAC details for the mapping
+	KubernetesMapping `json:",inline"`
+}
+
+// UserMapping represents a mapping from a IAM user
```
randomvariable, line 114:
```suggestion
// UserMapping represents a mapping from an IAM user to Kubernetes users and groups
```

randomvariable:
two documentation updates

controlplane/eks/api/v1alpha3/types.go
```diff
@@ -83,3 +83,39 @@ var (
...
+
+// KubernetesMapping represents the kubernetes RBAC mapping
+type KubernetesMapping struct {
+	// UserName is a kubernetes RBAC user subject
+	UserName string `json:"username"`
+	// Groups is a list of kubernetes RBAC groups
+	Groups []string `json:"groups"`
+}
+
+// RoleMapping represents a mapping from a IAM role
```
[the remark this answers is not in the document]
richardcase: Changed

controlplane/eks/api/v1alpha3/types.go
richardcase: Changed

richardcase:
> two documentation updates

Thanks @randomvariable  - both changed.

PR #2002, opened 2020-10-02
.dockerignore: Add cmd package

randomvariable:
Fixes broken docker builds


ncdc:
/approve

I'm just curious which make target(s) are broken?

randomvariable:
the eks controlplane, because it's consuming the IAM types. See #2004

ncdc:
Ahhhhhhh 👍

PR #2005, opened 2020-10-02
🌱 Fix volume description to remove 'root' since it's also used for non-root volume

jzhoucliqr:

Now that we have support for non-root volume, we should remove the 'root' keyword in description for Volume.

Fixes #

### DIFF
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -660,18 +660,18 @@ type Instance struct {
 	Tenancy string `json:"tenancy,omitempty"`
 }
 
-// Volume encapsulates the configuration options for the root volume
+// Volume encapsulates the configuration options for the storage device
 type Volume struct {
 	// Device name
 	// +optional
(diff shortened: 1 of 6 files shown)


api/v1alpha3/types.go
```diff
@@ -660,18 +660,18 @@ type Instance struct {
...
+// Volume encapsulates the configuration options for the storage device
 type Volume struct {
 	// Device name
 	// +optional
 	DeviceName string `json:"deviceName,omitempty"`
 
-	// Size specifies size (in Gi) of the root storage device.
-	// Must be greater than the image root snapshot size or 8 (whichever is greater).
+	// Size specifies size (in Gi) of the storage device.
+	// Must be greater than the image snapshot size or 8 (whichever is greater).
```
jzhoucliqr, line 670: Question for this, Is this 8G minimum check still valid now that this is also used for non-root volume?

randomvariable: No. Only applies to the root volume.

PR #2006, opened 2020-10-03
 ✨ Add support for DefaultCooldown to AWSMachinePool

### LINKED ISSUE #1987: Add support for DefaultCooldown options to AWSMachinePool
/kind feature
/help

Broken out of #1968 

Add support for configuring cooldowns https://docs.aws.amazon.com/autoscaling/ec2/userguide/Cooldown.html to AWSMachinePool

**Environment:**

arghya88:

Add support for configuring cooldowns https://docs.aws.amazon.com/autoscaling/ec2/userguide/Cooldown.html to AWSMachinePool

Fixes #1987


exp/api/v1alpha3/awsmachinepool_types.go
```diff
@@ -67,6 +67,11 @@ type AWSMachinePoolSpec struct {
 	// This field must match the provider IDs as seen on the node objects corresponding to a machine pool's machine instances.
 	// +optional
 	ProviderIDList []string `json:"providerIDList,omitempty"`
+
+	// The amount of time, in seconds, after a scaling activity completes before another scaling activity can start.
+	// +kubebuilder:default=300
+	// +optional
```
randomvariable, line 73: Not sure having it optional but defaulting to 300 results in anything useful. Either remove the optional flag, or move defaulting logic to the controller.

randomvariable: Maybe also change the type to the more Kubernetes style of using time.ParseDuration, so users can specify "5m" etc... That would hint towards removing the optional flag, and setting the default.

arghya88: Thanks for the suggestion...you mean change the type to string and then in the controller somewhere use time.ParseDuration and use Seconds() to get the value in float64. Now I am having a bit of trouble converting that to *int64 because that's what the aws lib expects

exp/api/v1alpha3/awsmachinepool_types.go
```diff
@@ -67,6 +68,10 @@ type AWSMachinePoolSpec struct {
 	// This field must match the provider IDs as seen on the node objects corresponding to a machine pool's machine instances.
 	// +optional
 	ProviderIDList []string `json:"providerIDList,omitempty"`
+
+	// The amount of time, in seconds, after a scaling activity completes before another scaling activity can start.
+	// +kubebuilder:default=300
+	DefaultCoolDown time.ParseDuration `json:"defaultCoolDown,omitempty"`
```
randomvariable, line 74:
```suggestion
	DefaultCoolDown metav1.Duration `json:"defaultCoolDown,omitempty"`
```

exp/api/v1alpha3/awsmachinepool_types.go
randomvariable: converting to int64 is fine. You'll need to use metav1.Duration

exp/api/v1alpha3/awsmachinepool_types.go
arghya88: I have done some changes..PTAL

arghya88:
@randomvariable what do you think of this in it's current state? Let me know if any change is required like moving the defaulting to controller instead of relying on kubebuilder.

richardcase:
> Let me know if any change is required like moving the defaulting to controller instead of relying on kubebuilder.

Thanks @arghya88 . I think we need to add a validation check in the [webhook](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/exp/api/v1alpha3/awsmanagedmachinepool_webhook.go#L52) to make sure that nothing crazy like a negative duration is added.

arghya88:
@richardcase as I can see there is no webhook for `AWSMachinePool` but there is one for `AWSManagedMachinePool`. So we need to create a new webhook for `AWSMachinePool` because I added the `DefaultCoolDown` in `AWSMachinePoolSpec`  not in `AWSManagedMachinePoolSpec`?

richardcase:
Good point @arghya88....i'd missed that, sorry.

I think we should create the webhooks for AWSMachinePool and do this Duration check along with other validation checks for AWSMachinePool.

But perhaps we should create a separate issue for adding the webhooks for AWSMachinePool? What do you think @randomvariable ?

exp/api/v1alpha3/awsmachinepool_webhook.go
```diff
@@ -0,0 +1,91 @@
...
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+)
+
+var log = logf.Log.WithName("awsmachinepool-resource")
+
+// SetupWebhookWithManager will setup the webhooks for the AWSMachinePool
+func (r *AWSMachinePool) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
richardcase, line 18: This will need to be called from the main if the EKS flag is enabled like [we do for the AWSManagedMachinePool](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/main.go#L227)

arghya88: Thanks..added

arghya88:
@richardcase I am getting this error
```

Boilerplate header is wrong for: /home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/exp/api/v1alpha3/awsmachinepool_webhook.go
make: *** [Makefile:486: verify-boilerplate] Error 1
```

richardcase:
>  I am getting this error

You will need to add the license header to the new webhook file.

richardcase:
Any new files need to have the license comment in them.

arghya88:
> Any new files need to have the license comment in them.

Got it. Thanks for the guidance.

arghya88:
@randomvariable I have done below changes. PTAL

1. Made it optional
2. Setting default value in a webhook now
2. Validating negative value in the webhook

cc @richardcase

exp/api/v1alpha3/awsmachinepool_webhook.go
```diff
@@ -0,0 +1,107 @@
...
+}
+
+// ValidateUpdate will do any extra validation when updating a AWSMachinePool
+func (r *AWSMachinePool) ValidateUpdate(old runtime.Object) error {
+	var allErrs field.ErrorList
+	if errs := r.validateDefaultCoolDown(); errs != nil || len(errs) == 0 {
+		allErrs = append(allErrs, errs...)
+	}
+
+	if len(allErrs) == 0 {
```
richardcase, line 85: In the future should we should change this to use: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/28800c8339350a4d9036ebc9b71a35a6107261e3/api/v1alpha3/webhooks.go#L25

richardcase:
Seems ok to me

/lgtm

exp/api/v1alpha3/awsmachinepool_webhook.go
arghya88: At the moment this `webhook.go` file does not exist in `/exp/api/v1alpha3`. I will do a follow up PR to add it there and and use it for all the webhooks in `/exp/api/v1alpha3`

PR #2011, opened 2020-10-06
🌱 Validate subnet's az with failure domain when subnet id is passed

### LINKED ISSUE #1838: Specifying subnetID for an AWSMachine and a conflicting failureDomain will ignore the failureDomain
/kind bug

Summarizing from this comment: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1833#discussion_r459665277

- create an AWSMachine while specifying the subnetID it should attach to
- also specify a failureDomain that **isn't** the failureDomain of the subnet

Expect it to validate the subnet is in the correct failureDomain and run into errors trying to create the instance. 
Actual: machine is attached to subnet specified by subnetID

shysank:
Validates subnet's availability zone with failure domain when subnet id is passed

Fixes #1838

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -262,26 +262,50 @@ func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
 
 	switch {
 	case scope.AWSMachine.Spec.Subnet != nil && scope.AWSMachine.Spec.Subnet.ID != nil:
+		if failureDomain != nil {
+			subnet := s.scope.Subnets().FindByID(*scope.AWSMachine.Spec.Subnet.ID)
+			if subnet == nil {
+				record.Warnf(scope.AWSMachine, "FailedCreate",
+					"Failed to create instance: subnet with id %q not found", aws.StringValue(scope.AWSMachine.Spec.Subnet.ID))
+				return "", awserrors.NewFailedDependency(
+					fmt.Sprintf("failed to run machine %q, subnet with id %q not found",
+						scope.Name(),
+						aws.StringValue(scope.AWSMachine.Spec.Subnet.ID),
+					),
+				)
+			}
+
+			if subnet.AvailabilityZone != *failureDomain {
+				record.Warnf(scope.AWSMachine, "FailedCreate",
+					"Failed to create instance: subnet's availability zone %q does not match with the failure domain %q",
+					subnet.AvailabilityZone,
+					*failureDomain)
+				return "", awserrors.NewFailedDependency(
+					fmt.Sprintf("failed to run machine %q, subnet's availability zone %q does not match with the failure domain %q",
+						scope.Name(),
+						subnet.AvailabilityZone,
+						*failureDomain,
+					),
+				)
+			}
+		}
 		return *scope.AWSMachine.Spec.Subnet.ID, nil
 	case scope.AWSMachine.Spec.Subnet != nil && scope.AWSMachine.Spec.Subnet.Filters != nil:
-		subnetInput := &ec2.DescribeSubnetsInput{
-			Filters: []*ec2.Filter{
-				filter.EC2.SubnetStates(ec2.SubnetStatePending, ec2.SubnetStateAvailable),
-				filter.EC2.VPC(s.scope.VPC().ID),
-			},
+		criteria := []*ec2.Filter{
+			filter.EC2.SubnetStates(ec2.SubnetStatePending, ec2.SubnetStateAvailable),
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/ec2/instances.go
```diff
@@ -262,26 +262,54 @@ func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
 
 	switch {
 	case scope.AWSMachine.Spec.Subnet != nil && scope.AWSMachine.Spec.Subnet.ID != nil:
+		if failureDomain != nil {
+			subnets, err := s.getFilteredSubnets(filter.EC2.SubnetID(aws.StringValue(scope.AWSMachine.Spec.Subnet.ID)))
```
shysank, line 266: Should VPC ID also be passed here?

gab-satchi:
linter giving a suggestion for one of the for loops: 
```pkg/cloud/services/ec2/instances.go:355:2: S1011: should replace loop with `subnetInput.Filters = append(subnetInput.Filters, criteria...)` (gosimple)```

File: pkg/cloud/filter/ec2.go [hunk 2]
```diff
@@ -156,3 +157,10 @@ func (ec2Filters) IgnoreLocalZones() *ec2.Filter {
 		Values: aws.StringSlice([]string{"opt-in-not-required"}),
 	}
 }
+
```
richardcase, line 160: Small nit, a function comment would be a nice to have but not all the exported functions in this file have them

pkg/cloud/services/ec2/instances.go
```diff
@@ -262,26 +262,54 @@ func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
...
+			if err != nil {
+				return "", err
```
richardcase, line 268: Could potentially wrap the error, but a nice to have

richardcase (approved):
Thanks @shysank for this change. A couple of minor comments but these are only nice to haves, so:

/lgtm

pkg/cloud/filter/ec2.go
shysank: Fixed it.

pkg/cloud/services/ec2/instances.go
shysank: Fixed it, Thanks!

pkg/cloud/services/ec2/instances.go
```diff
@@ -262,26 +262,55 @@ func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
...
+				return "", errors.Wrapf(err, "failed to filter subnets for criteria %q", criteria)
+			}
+
+			if len(subnets) == 0 {
+				record.Warnf(scope.AWSMachine, "FailedCreate",
+					"Failed to create instance: subnet with id %q not found", aws.StringValue(scope.AWSMachine.Spec.Subnet.ID))
+				return "", awserrors.NewFailedDependency(
+					fmt.Sprintf("failed to run machine %q, subnet with id %q not found",
+						scope.Name(),
+						aws.StringValue(scope.AWSMachine.Spec.Subnet.ID),
```
randomvariable, line 278: Can we introspect the subnet information from the AWSCluster object instead? I'm wary of adding an additional EC2 API call if we can avoid it.

shysank: Do you mean something like this `s.scope.Subnets().FindByID("id")` ?

randomvariable: Yup. It might be we need to persist more information in AWSCluster status to enable this.

shysank: I see the subnets being populated [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a229052b19a7b4d976f312a6bf6a42a9ce09b5d8/pkg/cloud/services/network/subnets.go#L148), but I am not entirely sure, what extra data need to go in there. Is there some documentation that I can refer to?
Should we also modify `getFilteredSubnets` to look for subnets in status, and fallback to ec2 api if it's not there?

richardcase: I think there should be everything needed already persisted:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a229052b19a7b4d976f312a6bf6a42a9ce09b5d8/pkg/cloud/services/network/subnets.go#L264

Does that have everything needed @shysank ?

shysank: I've made the changes, PTAL.

richardcase:
This is looking good to me.

@shysank - would you be able to squash your commits?

Issue #1990, opened 2020-09-29
Default remote access security groups/bastion host for AWSManagedMachinePool


michaelbeaumont:
/kind feature
/kind cleanup
```
// SourceSecurityGroups specifies which security groups are allowed access
// An empty array opens port 22 to the public internet
```

This feels like a very dangerous default to me.

Appreciate thats the [default behavior of the SDK](https://docs.aws.amazon.com/sdk-for-go/api/service/eks/#RemoteAccessConfig) but feels like we could protect the user a bit better. 

Perhaps for a follow up issue?

_Originally posted by @richardcase in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1916#discussion_r496694858_

### Bastion
Ideas:

* If `remoteAccess` specified without `SSHKeyName`, take the SSH key from the control plane spec.
* Add the `SecurityGroupEKSNodeAdditional` to `sourceSecurityGroups`

randomvariable:
only comment is yowch

PR #2013, opened 2020-10-07
🌱 Improve AWSManagedMachinePool remote access support

michaelbeaumont:
Addresses part of #1990, making it less cumbersome to set up remote access for AWSManagedMachinePool

### DIFF
--- a/pkg/cloud/services/eks/nodegroup.go
+++ b/pkg/cloud/services/eks/nodegroup.go
@@ -108,6 +108,37 @@ func ngTags(key string, additionalTags infrav1.Tags) map[string]string {
 	return tags
 }
 
+func (s *NodegroupService) remoteAccess() (*eks.RemoteAccessConfig, error) {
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/eks/nodegroup.go
```diff
@@ -118,6 +149,12 @@ func (s *NodegroupService) createNodegroup() (*eks.Nodegroup, error) {
 	}
 	managedPool := s.scope.ManagedMachinePool.Spec
 	tags := ngTags(s.scope.Name(), additionalTags)
+
+	remoteAccess, err := s.remoteAccess()
+	if err != nil {
+		return nil, err
```
richardcase, line 155: Maybe wrap the error?

richardcase (approved):
One minor comment but:

/lgtm

pkg/cloud/services/eks/nodegroup.go
michaelbeaumont: changed

PR #2018, opened 2020-10-12
:seedling: docs: Provide AMI IDs for rebuilt Ubuntu amis and fix e2e conformance with CI Artifacts jobs

randomvariable:
Due to https://github.com/kubernetes-sigs/image-builder/pull/406,
have had to rebuild Ubuntu AMIs.

Am also bumping e2e versions and fixing the conformance with ci artifacts job so we have signal.

Fixes https://github.com/kubernetes/kubernetes/issues/95730

### DIFF
--- a/test/e2e/conformance_test.go
+++ b/test/e2e/conformance_test.go
@@ -57,6 +57,7 @@ var _ = Describe("conformance tests", func() {
 	Measure(specName, func(b Benchmarker) {
 
 		name := fmt.Sprintf("cluster-%s", util.RandomString(6))
+		setEnvVar("USE_CI_ARTIFACTS", "true", false)
 		kubernetesVersion := e2eConfig.GetVariable(KubernetesVersion)
 		flavor := clusterctl.DefaultFlavor
 		if useCIArtifacts {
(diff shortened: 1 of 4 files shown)


randomvariable:
/hold
for conformance test

/test ?

randomvariable:
/unhold

conformance tests are running successfully. The result of kubetest is independent of the cluster creation and deletion.

test/e2e/conformance_test.go
```diff
@@ -57,6 +57,7 @@ var _ = Describe("conformance tests", func() {
 	Measure(specName, func(b Benchmarker) {
 
 		name := fmt.Sprintf("cluster-%s", util.RandomString(6))
+		setEnvVar("USE_CI_ARTIFACTS", "true", false)
```
detiber, line 60: Why the override on the env variable, wouldn't we want to use the value set from Prow?

test/e2e/conformance_test.go
```diff
@@ -105,6 +106,7 @@ var _ = Describe("conformance tests", func() {
 	}, 1)
 
 	AfterEach(func() {
+		setEnvVar("USE_CI_ARTIFACTS", "false", false)
```
detiber, line 109: Similar here, what is the need for overriding the env variable here rather than using the value provided by Prow.

detiber (commented):
Question about hardcoding the values for USE_CI_ARTIFACTS, but otherwise lgtm

test/e2e/conformance_test.go
randomvariable: It's coming in as an explicit command line argument `--use-ci-artifacts`, and setting env var for the purpose of envsubst rendering from the CAPI package.

Maybe should have wired up differently

PR #2032, opened 2020-10-14
:seedling: Refactor error handling in webhooks to use API Machinery

### LINKED ISSUE #2020: Use `kerrors.NewAggregate` in all webhooks
/kind cleanup

As an example https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/exp/api/v1alpha3/awsmanagedmachinepool_webhook.go#L71

**Environment:**

arghya88:

Use `kerrors.NewAggregate` in all webhooks under `exp/api/v1alpha3`

Fixes #2020

### DIFF
--- a/api/v1alpha3/awscluster_webhook.go
+++ b/api/v1alpha3/awscluster_webhook.go
@@ -18,6 +18,7 @@ package v1alpha3
 
 import (
 	"fmt"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"reflect"
 
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -47,12 +48,11 @@ var (
 )
 
 func (r *AWSCluster) ValidateCreate() error {
-	var allErrs field.ErrorList
+	var allErrs []error
 
 	allErrs = append(allErrs, r.Spec.Bastion.Validate()...)
 	allErrs = append(allErrs, isValidSSHKey(r.Spec.SSHKeyName)...)
-
-	return aggregateObjErrors(r.GroupVersionKind().GroupKind(), r.Name, allErrs)
+	return kerrors.NewAggregate(allErrs)
 }
 
 func (r *AWSCluster) ValidateDelete() error {
@@ -60,7 +60,7 @@ func (r *AWSCluster) ValidateDelete() error {
 }
 
 func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
-	var allErrs field.ErrorList
+	var allErrs []error
 
 	oldC, ok := old.(*AWSCluster)
 	if !ok {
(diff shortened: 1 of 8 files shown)


exp/api/v1alpha3/webhooks.go
```diff
@@ -0,0 +1,35 @@
...
+
+package v1alpha3
+
+import (
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+)
+
+func aggregateObjErrors(gk schema.GroupKind, name string, allErrs field.ErrorList) error {
```
michaelbeaumont, line 25: Maybe cleaner to just accept an `"k8s.io/api/core/v1".ObjectReference`?

detiber: I'd suggest avoiding using ObjectReference wherever possible for new code, especially since it's on the roadmap for removal in v1alpha4

detiber, line 25: This seems like something that would be more generally useful outside of just exp/api/v1alpha3. No objection to adopting this as a short term solution, but I think it would make sense to consider adding this to core cluster-api for more general use in the long term.

arghya88: Fair enough..I will open an issue in cluster-api to get feedback..once that materializes..will refactor this to use the function from cluster-api in a follow up PR.

randomvariable, line 25: I think we should be also refactoring all of the webhooks to use kerrors.NewAggregate https://github.com/kubernetes/apimachinery/blob/master/pkg/util/errors/errors.go#L46. Can then remove the length check as that's already implemented in kerrors.

arghya88: You mean to get rid of this `aggregateObjErrors` and use `kerrors.NewAggregate` like it's being done here https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/0c45468b245feb28565bba57938511c3cf3f5f59/exp/api/v1alpha3/azuremachinepool_webhook.go#L92

randomvariable: Well, you don't even need to do
```
if len(errs) > 0 {
		return kerrors.NewAggregate(errs)
	}
```
as NewAggregate does that for you.

arghya88: Agree..made some code changes.PTAL

randomvariable:
Think this is good, but wouldn't mind a second opinion from @vincepri or @ncdc who first told me about this approach.

ncdc:
@randomvariable I don't think we want to do this? Don't we want to return `apierrors.NewInvalid(...)` from validation methods?

arghya88:
@randomvariable what do you suggest to do on this?

PR #2034, opened 2020-10-14
WIP: e2e: Add a scale test.

randomvariable:
Adds an e2e scenario and generators for cluster resources such
that we create one cluster, and then use the existing VPC method
to generate X number of new clusters and attempt to create them all at
once. This saves on having to have an account that will have 50,100,200
VPCs in them.

Did initially start adding different ramp up rates based on https://gatling.io/docs/current/general/simulation_structure/
but don't really need them.

Fixes #


randomvariable:
cc @sedefsavas

PR #2036, opened 2020-10-14
✨ SSM (AWS Systems Manager Parameter Store) as a Secret Backend

### LINKED ISSUE #1900: Secret Backend Support

randomvariable:
I've just rebased #1924 because I think @Promaethius is away, but do want to get into 0.6.1.

Adds an additional secret backend for environments which do not have access to AWS Secrets Manager. SSM is available as an EC2 feature in all possible scopes: https://aws.amazon.com/compliance/services-in-scope/

Additionally, the format of SSM API and CLI calls are similar to SecretsManager so this should be a relatively painless addition.

Fixes #1900


api/v1alpha3/awsmachine_types.go
```diff
@@ -28,6 +28,17 @@ const (
...
+// SecretBackend defines variants for backend secret storage.
+type SecretBackend string
+
+var (
+	// SecretBackendSSMParameterStore defines AWS Systems Manager Parameter Store as the secret backend
+	SecretBackendSSMParameterStore = SecretBackend("ssm-parameter-store")
+
+	// SecretBackendSecretsManager defines AWS Secrets Manager as the secret backend
+	SecretBackendSecretsManager = SecretBackend("secrets-manager")
+)
```
detiber, line 40: Not sure if this was discussed previously, but is there a specific use case in mind for having this as an API field rather than a configuration option on the controller manager?

Is there a specific use case in mind where one would want to choose different backends for individual Clusters/Machines vs having it a global configuration for an instance of the infrastructure provider?

randomvariable: Oops, closed it because @Promaethius is around, I'll reply in the other PR,

PR #2037, opened 2020-10-14
✨ Watches instance state changes and reconcile AWSMachine if necessary

### LINKED ISSUE #1871: Consume EventBridge for EC2 instance state notifications
/kind feature

When there's EC2 state changes, CAPA won't reconcile this until the next run which could be up to 10 mins.

I think it's also of benefit if we recorded the EC2 status somewhere, which means you don't have to continuosly compare AWSMachine with the EC2 console / favourite AWS EC2 access mechanism.

We should find a mechanism to consume these events, probably via some combination of EventBridge, SNS & SQS.
CAPA should consume all instance notifications and then compare against the AWSMachines on the API Server. We may need to add some indexes to be able to do this efficiently without retrieving all AWSMachines.

**Environment:**

gab-satchi:
- Reconciles an SQS queue per cluster (with queue name modified to omit '.' similar to ELBs)
- Reconciles an EventBridge rule per cluster on the default event bus (Custom event buses can't accept aws service events from the same account)
- Adds queue as a target to the cluster's rule
- Creates a policy for rule on the queue so rule can publish to the SQS queue
- Adds instances to rule as instances are created
- Adds instance state controller to track queue URLs of all available clusters
- global scope that isn't tied to a cluster
- triggers reconcile on AWSMachine when consuming messages

Fixes #1871


gab-satchi:
test /pull-cluster-api-provider-aws-e2e

gab-satchi:
The e2e test I modified passed but a different one timed out. Running it locally, I also saw it flake while waiting for the ELB name to register. Slightly increased the timeout for waiting on cluster to reach provisioned state. 

gab-satchi:
seems to be failing due to the test that creates a cluster with name: `sg-test.` the `.` is not allowed for queue name.

gab-satchi:
/retitle ✨ Watches instance state changes and reconcile AWSMachine if necessary

pkg/cloud/services/instancestate/rule.go
```diff
@@ -0,0 +1,214 @@
...
+	})
+
+	if err != nil && resourceNotFoundError(err) {
+		return nil
+	}
+
+	return errors.Wrapf(err, "unable to delete rule %s", s.getEC2RuleName())
+}
+
+func (s Service) AddInstanceToEventPattern(instanceID string) error {
```
sedefsavas, line 160: I see deleted instances still exist on the rule, we may need to have a cleanup for this during reconcileDelete.

sedefsavas (commented):
Thanks for this feature @gab-satchi 
I tested it with multiple clusters, it works well.
Have a few comments, other than that LGTM.

sedefsavas:
Another nit. 
In cluster-api, we decided to add a new package for each feature going forward so that cleaning it up when promoted is easier. We can do the same here, instead of adding all features under exp.

exp/controllers/awsinstancestate_controller.go
```diff
@@ -0,0 +1,238 @@
...
+				}
+				for _, msg := range resp.Messages {
+					m := message{}
+					err := json.Unmarshal([]byte(*msg.Body), &m)
+
+					if err != nil {
+						r.Log.Error(err, "unable to marshall")
+						return
+					}
+					r.processMessage(ctx, m)
```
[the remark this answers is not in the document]
sedefsavas: Shall we add a TODO to explain this behavior?

exp/controllers/awsinstancestate_controller.go
```diff
@@ -0,0 +1,237 @@
...
+	ctx := context.TODO()
+	awsClusterList := &infrav1.AWSClusterList{}
+	if err := r.Client.List(ctx, awsClusterList); err == nil {
+		for i, cluster := range awsClusterList.Items {
+			if URL, err := r.getQueueURL(&awsClusterList.Items[i]); err == nil {
+				r.queueURLs.Store(cluster.Name, queueParams{region: cluster.Spec.Region, URL: URL})
+			}
+		}
+	}
+	for range time.Tick(1 * time.Second) {
```
sedefsavas, line 127: We may hit the api call limits if it is too frequent (assuming it is using the api request tokens too?). Maybe make this a parameter or increase the interval?

sedefsavas:
Forgot to ask: what was the pros and cons of using a single queue for all clusters?

pkg/cloud/services/instancestate/rule.go
```diff
@@ -0,0 +1,214 @@
...
+
+	return nil
+}
+
+func (s Service) createRule() error {
+	eventPattern := eventPattern{
+		Source:     []string{"aws.ec2"},
+		DetailType: []string{Ec2StateChangeNotification},
+		EventDetail: &eventDetail{
+			States: []infrav1.InstanceState{infrav1.InstanceStateShuttingDown, infrav1.InstanceStateTerminated},
```
sedefsavas, line 126: Since we get all messages at once at each interval (tick), there is no api limit issue if we watch all events, right? It'd be good to see the exact state of ec2 instances. This can be tracked in a follow up issue.

gab-satchi: yeah I don't think there's rate limiting for the sqs api. I know ec2 does rate limiting. With SQS there's a different non-free tier after a certain number of requests. Seeing the ec2 state would definitely be useful.

gab-satchi:
> Forgot to ask: what was the pros and cons of using a single queue for all clusters?

that was something I thought through initially. Main issue was handling the lifecycle of that one queue? Creating is easy but when does it get deleted?

api/v1alpha3/awsmachine_webhook.go
```diff
@@ -83,6 +83,10 @@ func (r *AWSMachine) ValidateUpdate(old runtime.Object) error {
 	delete(oldAWSMachineSpec, "providerID")
 	delete(newAWSMachineSpec, "providerID")
 
+	// allow changes to instanceID
```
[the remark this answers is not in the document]
gab-satchi: It should stay the same but we don't have instanceID on creation of the AWSMachine. When we do have access to it when we create the ec2 instance, we'd need to update it in the spec.

pkg/cloud/services/instancestate/rule.go
gab-satchi: I had this on my initial plan but seemed to have skipped it as failed/terminated instances won't be emitting any new events. Will update to clean up the rule before an intentional deletion of the machine.

sedefsavas:
/lgtm
Should be squashed before approval.

sedefsavas:
I think this PR is ready to go after rebasing.

gab-satchi:
During the rebase had to do the following as well: 
- Add `ServiceLimiter` function to GlobalScope
- Regenerate mock eventbridge interface

sedefsavas:
I think we can merge this after rebase.
Let's prioritize to merge this as this has been rebased couple of times.
cc @richardcase @randomvariable

richardcase:
Oh no, it needs another rebase 😞

richardcase:
@gab-satchi - does this use short polling on the SQS queue? It looks like it from the code but wanted to check as I may have missed the wait time setting

richardcase:
It does use long polling (i missed the attribute setting on first pass), so this also:

/lgtm

richardcase:
e2e test are passing so:

/approve

PR #2048, opened 2020-10-19
🐛: handle all subnets properties for AWSMachinePool

### LINKED ISSUE #2047: AWSMachinePool - Subnets aren't fetched using filter
/kind bug

When using `AWSMachinePool` you're required to specify the `Subnets` property in the spec otherwise the ASG fails to create. This value is typed as `infrav1.AWSResourceReference` which provides the ability to specify using IDs, ARNs or Filters but the `AWSMachinePool` only uses ID for subnets.

The `Filters` (or `ARN`) are also valid options to provide and cluster-api will look up the appropriate subnets.

_Alternatively..._ it would be good is `Subnets` was optional and if not provided it defaults to the private subnets associated with the control plane.

Relevant code block: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f6a97de9ba98338e58bf1afafb71ece60db5b394/pkg/cloud/services/autoscaling/autoscalinggroup.go#L143-L146

**Environment:**

- Kubernetes version: (use `kubectl version`): 1.18
- OS (e.g. from `/etc/os-release`): N/a

MarcusNoble:
Ensures all properties provided for `Subnets` on the `AWSMachinePool` resource are handled. If none are provided it falls back to the control-plane subnets.

Fixes #2047

### DIFF
--- a/pkg/cloud/services/autoscaling/autoscalinggroup.go
+++ b/pkg/cloud/services/autoscaling/autoscalinggroup.go
@@ -21,6 +21,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/autoscaling"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
 	expinfrav1 "sigs.k8s.io/cluster-api-provider-aws/exp/api/v1alpha3"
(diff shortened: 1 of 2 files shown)


richardcase (commented):
Thanks @MarcusNoble . We probably shouldn't be doing another AWS call for this as we already have the subnet information on the InfraCluster (i.e. on the AWSManagedControlPlane or AWSCluster). 

Do you think we need the filter functionality on the subnets defined in the AWSMachinePool? Or could that just be a list of subnets ids? If  a filter is needed perhaps we can filter on the subnets already  on the InfraCluster?

MarcusNoble:
I think you might be right, we wouldn't likely want to use a subnet that isn't already known to InfraCluster so maybe using `AWSResourceReference` for the `Subnets` field isn't the right choice. A change to an (optional) list of IDs would cause a breaking change though. I can get another PR put up to cover that use case and we can compare.

MarcusNoble:
See: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2050

richardcase:
@MarcusNoble - i've been thinking about the filtering some more. Implementing the filtering against the InfraCluster subnets could be very useful. For instance, it may be helpful to supply a filter that essentially said "use subnets that IsPublic == false" instead of actually having to supply the subnet ids. 

What do you think?

MarcusNoble:
I can see that being useful but not sure how to cleanly expose that as a property while still supporting a list of IDs 🤔

MarcusNoble:
Thinking more about this, is there actually many (any?) instances where you'd want to run the worker nodes in the public subnets? If those are edge cases we could default to filtering to the private subnets from InfraCluster and if public are needed then the IDs can be provided which will take precedence.

randomvariable:
Can we open up an equivalent issue for filtering for AWSMachine then please, if this is the consensus?
We need consistent APIs.

MarcusNoble:
Closing until decision is made on https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2087 as to how to proceed.

PR #2050, opened 2020-10-20
⚠️: change AWSMachinePool subnets to an optional list of IDs

### LINKED ISSUE #2047: AWSMachinePool - Subnets aren't fetched using filter

MarcusNoble:
Change the `Subnets` field on `AWSMachinePool` to be an optional string slice containing subnet IDs rather than `AWSResourceReference` as only the IDs are currently being used. If none are provided it falls back to the subnets associated with `InfraCluster`

See also: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2048 

Fixes #2047

### DIFF
--- a/exp/api/v1alpha3/awsmachinepool_types.go
+++ b/exp/api/v1alpha3/awsmachinepool_types.go
@@ -48,8 +48,9 @@ type AWSMachinePoolSpec struct {
 	// AvailabilityZones is an array of availability zones instances can run in
 	AvailabilityZones []string `json:"availabilityZones,omitempty"`
 
-	// Subnets is an array of subnet configurations
-	Subnets []infrav1.AWSResourceReference `json:"subnets,omitempty"`
+	// Subnets is an array of subnet IDs
+	// +optional
+	Subnets []string `json:"subnets,omitempty"`
 
 	// AdditionalTags is an optional set of tags to add to an instance, in addition to the ones added by default by the
 	// AWS provider.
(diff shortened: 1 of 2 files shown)


arghya88:
I think we have to wait till `v1alpha4` to introduce breaking change.

MarcusNoble:
@arghya88 I also opened #2048 as an alternative (non-breaking) approach but I have also though that we could introduce the fallback logic without making changes to the current schema which would also solve the issue I'm seeing. Would this be preferable?

MarcusNoble:
e.g. https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2051

richardcase:
This functionality is experimental so i don't think  there isn't such strict requirements around changing the contract. The fallback to using the InfraCluster subnets is good.

randomvariable:
@mytu and @dthorsen are potential end users and may get hit by the API change, so would like their opinion

mytu:
> @mytu and @dthorsen are potential end users and may get hit by the API change, so would like their opinion

I think you go the wrong person as I have nothing to do with your project. Good luck!

dthorsen:
@randomvariable @MarcusNoble Thanks for the heads up. This looks good to me. I think the list of subnet ID strings makes more sense as well.

randomvariable:
Oops, I meant @mytunguyen 

@mytunguyen, @dthorsen We also have an alternative presented in #2048 that adds support for filtering but keeps the existing API. I would like that ported to AWSMachine then.

randomvariable:
After discussion in Slack, we've decided to not break the API in v1alpha3.

/close

PR #2051, opened 2020-10-20
🐛 Add fallback if subnets not provided on AWSMachinePool

MarcusNoble:
Fallback to subnets associated with cluster if none are provided.

Related #2047

See also #2048 and #2050 as alternatives


arghya88 (commented):
/lgtm

richardcase:
This change does look good, but we should probably try to keep discussions and the change to the 1 PR.

@MarcusNoble @arghya88 - i've added a comment to #2048 about filtering.

PR #2068, opened 2020-10-23
🌱 Updated Bastion node's AMI. Changed the OS version to Ubuntu 20.04 from Ubuntu 16.04

LochanRn:

Updated the OS version of the bastion node. The current AMI's are Ubuntu 16.04 which is two years old and it has security vulnerabilities.

Fixes #


LochanRn:
I signed it

PR #2071, opened 2020-10-26
🐛Make the regex less restrictive to allow underscore etc. for ssh key file name

### LINKED ISSUE #2060: ssh key validation regex is too restrictive
/kind bug

created an AWSMachine with an ssh key with an underscore in the name, got denied by webhook.

it should work, as I believe all ascii characters should work for ssh key name:

https://docs.aws.amazon.com/cli/latest/reference/ec2/create-key-pair.html

**Environment:**

arghya88:
Make the regex less restrictive to allows underscore etc. for ssh key file name
Fixes #2060

### DIFF
--- a/api/v1alpha3/webhooks.go
+++ b/api/v1alpha3/webhooks.go
@@ -23,6 +23,10 @@ import (
 	"regexp"
 )
 
+var (
+	sshKeyValidNameRegex = regexp.MustCompile(`^[[:graph:]]+([[:print:]]*[[:graph:]]+)*$`)
+)
+
 func aggregateObjErrors(gk schema.GroupKind, name string, allErrs field.ErrorList) error {
 	if len(allErrs) == 0 {
 		return nil
@@ -38,15 +42,9 @@ func aggregateObjErrors(gk schema.GroupKind, name string, allErrs field.ErrorLis
 func isValidSSHKey(sshKey *string) field.ErrorList {
 	var allErrs field.ErrorList
 	if sshKey != nil {
-		reg, err := regexp.Compile("[^-A-Za-z0-9-]+")
-		if err != nil {
-			return append(allErrs, field.Invalid(field.NewPath("sshKey"), sshKey, "SSHKey contains invalid character"))
+		if sshKey != nil && !sshKeyValidNameRegex.Match([]byte(*sshKey)) {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("sshKey"), sshKey, "Name is invalid. Must be specified in ASCII and must not start or end in whitespace"))
 		}
-		processedString := reg.ReplaceAllString(*sshKey, "")
-		if *sshKey == processedString {
-			return nil
-		}
-		allErrs = append(allErrs, field.Invalid(field.NewPath("sshKey"), sshKey, "SSHKey contains invalid character"))
 	}
 
 	return allErrs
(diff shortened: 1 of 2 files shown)


arghya88:
cc @randomvariable

api/v1alpha3/webhooks.go
```diff
@@ -38,7 +38,7 @@ func aggregateObjErrors(gk schema.GroupKind, name string, allErrs field.ErrorLis
 func isValidSSHKey(sshKey *string) field.ErrorList {
 	var allErrs field.ErrorList
 	if sshKey != nil {
-		reg, err := regexp.Compile("[^-A-Za-z0-9-]+")
+		reg, err := regexp.Compile("[^-A-Za-z0-9_-]+")
```
randomvariable, line 41:
```suggestion
		reg, err := regexp.Compile("[\x00-\x7F]+")
```

Also, can we move this into variable using `regexp.MustCompile`, so we're not reallocating memory on every call to the webhook?

arghya88: If I do `reg := regexp.MustCompile(`[\x00-\x7F]+`)` the tests with ssh key name `test-key`, `test_key`, `test` fails
`

arghya88: I have done some changes. with this change `test\t` is a valid name of ssh keyfile.Hence had to make some changes in the test as well. PTAL

api/v1alpha3/webhooks.go
```diff
@@ -38,10 +38,7 @@ func aggregateObjErrors(gk schema.GroupKind, name string, allErrs field.ErrorLis
 func isValidSSHKey(sshKey *string) field.ErrorList {
 	var allErrs field.ErrorList
 	if sshKey != nil {
-		reg, err := regexp.Compile("[^-A-Za-z0-9-]+")
-		if err != nil {
-			return append(allErrs, field.Invalid(field.NewPath("sshKey"), sshKey, "SSHKey contains invalid character"))
-		}
+		reg := regexp.MustCompile(`^\x00-\x7F]+`)
```
randomvariable, line 41: Please move this to the package variable as mentioned previously.

randomvariable, line 41: Let's make sure we get this right:

Valid SSH key names are the intersection of ISO-8859, ASCII and valid XML entity names.

That gives us a regex of 

```go
var (
  sshKeyValidNameRegex regexp.MustCompile(`^[[:graph:]]+([[:print:]]*[[:graph:]]+)*$`)
)`
```

This should go in a package variable, such that the controller panics as close as possible to startup if the regex is invalid, and that we minimise memory reallocations.

then we can do something like the following (assuming we can change isValidSSHKey's function return  signature from `field.ErrorList` to `error`.

```suggestion
		if sshKey != nil && !sshKeyValidNameRegex.Match([]byte(*sshKey))
		{
		  return field.Invalid(field.NewPath("sshKey"), sshKey, "Name is invalid. Must be specified in ASCII and must not start or end in whitespace")
		}
		return nil
```

arghya88: Make sense..I think the change for error handling needs to be done in a separate PR

randomvariable: Yup, that's fine.

randomvariable (approved):

PR #2082, opened 2020-10-29
🐛  Updates AWSMachine CloudInit defaulting logic

### LINKED ISSUE #2064: AWSMachineTemplate defaulting and validation for `spec.cloudInit.insecureSkipSecretsManager` are in conflict
/kind bug

The AWSMachineTemplate defaulting webhook [sets `spec.cloudInit.secureSecretsBackend` to `secrets-manager`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/16515f764e272018be1cf0a56c2ea7c64468a508/api/v1alpha3/awsmachine_webhook.go#L179) if it is not specified, even when  `spec.cloudInit.insecureSkipSecretsManager` is set to `true`.

This seems to conflict with the validating webhook, which [forbids this combination](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/16515f764e272018be1cf0a56c2ea7c64468a508/api/v1alpha3/awsmachine_webhook.go#L124).

FWIW, the AWSMachineController _requires_ that [`spec.cloudInit.secureSecretsBackend` be set either to `secrets-manager`, or `ssm-parameter-store`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/16515f764e272018be1cf0a56c2ea7c64468a508/controllers/awsmachine_controller.go#L101). If some value is required, then I my guess is that the validating rule linked to above should be removed.

I discovered this by creating this AWSMachineTemplate:
```
---
kind: AWSMachineTemplate
metadata:
  name: dlipovetsky-test
  namespace: default
spec:
  template:
    spec:
      instanceType: t3.large
      sshKeyName: dlipovetsky
      cloudInit:
        insecureSkipSecretsManager: true
```

And finding that the defaulting webhook had modified it to:
```
---
kind: AWSMachineTemplate
metadata:
  name: dlipovetsky-test
  namespace: default
spec:
  template:
    spec:
      instanceType: t3.large
      sshKeyName: dlipovetsky
      cloudInit:
        insecureSkipSecretsManager: true
        secureSecretsBackend: secrets-manager
```

**Environment:**

CAPA v0.6.1

srm09:
This patch checks for the presence of the `spec.cloudInit.insecureSkipSecretsManager` before setting the default value for the `spec.cloudInit.secureSecretsBackend`.

Fixes #2064

### DIFF
--- a/api/v1alpha3/awsmachine_webhook_test.go
+++ b/api/v1alpha3/awsmachine_webhook_test.go
@@ -20,6 +20,7 @@ import (
 	"github.com/aws/aws-sdk-go/aws"
 	"testing"
 
+	. "github.com/onsi/gomega"
 	"k8s.io/utils/pointer"
 )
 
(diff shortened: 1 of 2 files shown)


api/v1alpha3/awsmachine_webhook_test.go
```diff
@@ -217,3 +218,49 @@ func TestAWSMachine_ValidateUpdate(t *testing.T) {
...
+			ProviderID:               nil,
+			AdditionalTags:           nil,
+			AdditionalSecurityGroups: nil,
+		},
+	}
+
+	tests := []struct{
+		name string
+		cloudInit CloudInit
+		assertFunc func(*AWSMachine, *WithT)
```
ncdc, line 234: This works, but I find it preferable in simple comparisons like this to have something like `expected string` and have a `g.Expect` comparison directly in the subtest, instead of using a function pointer.

PR #2085, opened 2020-11-02
🐛 AWSManagedMachinePool: Fix cloud provider key usage and nodegroup IAM role name

michaelbeaumont:
Fixes incorrectly using the nodegroup name for tagging resources instead of cluster name.
Fixes potential nodegroup IAM role name collisions.


richardcase (approved):
/lgtm

PR #2086, opened 2020-11-02
✨ Secondary cidr support for EKS workload clusters

### LINKED ISSUE #2065: [EKS] Support secondary CIDR
/kind feature

The ability to create workload clusters on EKS that make use of secondary CIDR ranges (`AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG` in [amazon-vpc-cni-k8s](https://github.com/aws/amazon-vpc-cni-k8s/)) to avoid the need of large IP allocation while still having worker nodes that a routeable over DirectConnect.

Ideally this should also be supported when using existing AWS infrastructure where the control plane and worker nodes make use of the provided VPC and subnets and cluster-api create the needed additional subnets and route tables for the secondary CIDR.

For this to work fully without intervention cluster-api would be required to reconfigure the aws-node daemonset that comes with new EKS clusters as well as adding some `ENIConfig` custom resources. This needs to be done prior to the worker nodes joining the cluster for the config to take effect.

https://www.eksworkshop.com/beginner/160_advanced-networking/secondary_cidr/
https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-eks-now-supports-additional-vpc-cidr-blocks/

MarcusNoble:
Add support for creating subnets from a secondary CIDR range to be used with aws-node CNI for pod IPs.

Fixes #2065


pkg/cloud/services/awsnode/cni.go
```diff
@@ -0,0 +1,134 @@
...
+					SecurityGroups: sgs,
+				},
+			}
+
+			if err := s.client.Create(context.Background(), &eniConfig, &client.CreateOptions{}); err != nil {
+				return err
+			}
+		}
+
+		s.scope.Info("Updating ENIConfig", "cluster-name", s.scope.Name(), "cluster-namespace", s.scope.Namespace(), "subnet", subnet.ID, "availability-zone", subnet.AvailabilityZone)
```
michaelbeaumont, line 78: If we're supporting updates, do the existing but no longer needed `ENIConfig`s need to be deleted?

MarcusNoble: That's a good point. It's also made me think we should be labeling these as being managed by cluster-api so we don't make changes to any created by the user outside of CAPA.

MarcusNoble: Updated 🙂

pkg/cloud/services/awsnode/cni.go
```diff
@@ -51,6 +52,11 @@ func (s *Service) ReconcileCNI() error {
 		return err
 	}
 
+	metaLabels := map[string]string{
```
michaelbeaumont, line 55: Is there a CAPI way to mark workload cluster resources as owned by some management cluster?

MarcusNoble: Nothing that I've been able to find.

randomvariable:
I guess given that e2e tests are done, this is unblocked for potential merging?

pkg/cloud/services/awsnode/cni.go
MarcusNoble: @randomvariable do you have any input on this at all? It's the only outstanding thing.

richardcase:
With the fix from #2199, the eks e2e should now run after a rebase.

richardcase:
Hopefully after a rebase the e2e tests will run.

richardcase:
e2e tests [passing](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2086/pull-cluster-api-provider-aws-e2e-eks/1350065907971395584)

/lgtm

PR #2097, opened 2020-11-06
Fix rolebinding for new serviceaccounts

MarcusNoble:
Fix the role bindings that were missed when moving to specific service accounts for the managers


randomvariable:
/close

superseded by #2099

PR #2100, opened 2020-11-09
✨ Add support for AWS_CONTROLLER_IAM_ROLE env var

### LINKED ISSUE #2053: Support using IAM for Service Accounts on the management cluster components
/kind feature

Add the ability to make use of [IAM for Service Accounts](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html) as an alternative to providing credentials (Access key and secret from the `AWS_B64ENCODED_CREDENTIALS` env var) for the management components to make use of.

N/A

**Environment:**

MarcusNoble:
Second attempt. See previous #2052 

Provide the ability to set the AWS_CONTROLLER_IAM_ROLE environment var which will then be used to add an EKS annotation to the service account the management components run as. This will allow those components to run as that IAM role when using IAM for Service Accounts.

- Includes Kiam support taken from: #1975
- Fixes issues with Kustomize name conflicts
- Adds service account to auth proxy role binding
- Includes docs: https://deploy-preview-2052--kubernetes-sigs-cluster-api-provider-aws.netlify.app/topics/specify-management-iam-role.html

Fixes #2053

/assign @randomvariable


randomvariable:
This isn't working because the service account is also being applied to the webhook server (which isn't needed), and the service account is only being created in capa-system.

bootstrap/eks/config/rbac/serviceaccount.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: controller-manager
+  namespace: system
+  labels:
+    control-plane: controller-manager
+  annotations:
+    ${AWS_CONTROLLER_IAM_ROLE/#arn/eks.amazonaws.com/role-arn: arn}
```
randomvariable, line 9: Can we add a comment to this one, as this is a very special hack that is specific to https://github.com/drone/envsubst, and doesn't work with gettext?

MarcusNoble: Added. 

Is there any scenario where this would be used with envsubst from gettext? Or are you just thinking about potential issues in the future?

randomvariable: potential issues in the future. when we started using envsubst in clusterctl and  that was within a v1alpha3 release, so in case something happened again it'd be a little easier to diagnose.

MarcusNoble: Good to know! 👍

PR #2103, opened 2020-11-10
:book: Update consuming-existing-aws-infrastructure.md to provide context to where networkSpec belongs

cdenneen:
Provide context as to where the `networkSpec` is to be updated inside the `kind: AWSManagedControlPlane`

### DIFF
--- a/docs/book/src/topics/consuming-existing-aws-infrastructure.md
+++ b/docs/book/src/topics/consuming-existing-aws-infrastructure.md
@@ -34,6 +34,10 @@ Finally, if the controller manager isn't started with the `--configure-cloud-rou
 Specifying existing infrastructure for Cluster API to use takes place in the specification for the AWSCluster object. Specifically, you will need to add an entry with the VPC ID and the IDs of all applicable subnets into the `networkSpec` field. Here is an example:
 
 ```yaml
+apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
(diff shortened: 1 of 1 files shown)


docs/book/src/topics/consuming-existing-aws-infrastructure.md
```diff
@@ -34,6 +34,10 @@ Finally, if the controller manager isn't started with the `--configure-cloud-rou
 Specifying existing infrastructure for Cluster API to use takes place in the specification for the AWSCluster object. Specifically, you will need to add an entry with the VPC ID and the IDs of all applicable subnets into the `networkSpec` field. Here is an example:
 
 ```yaml
+apiVersion: controlplane.cluster.x-k8s.io/v1alpha3
```
richardcase, line 37: For EKS based clusters it lives on **AWSManagedControlPlane** but for EC2 based clusters it lives on **AWSCluster**. For each of these kinds it is in the spec.

So i think we should we should either:
1. Change this part `Specifying existing infrastructure for Cluster API to use takes place in the specification for the AWSCluster object`
2. Have a separate example or section for each cluster type

richardcase (changes requested):

cdenneen:
I signed it

cdenneen:
I signed it

richardcase (approved):
/lgtm

PR #2120, opened 2020-11-22
🌱 add Interruptible field to AWSMachine status

### LINKED ISSUE #1899: Spot Instance interruption notice support
/kind feature

Spot instance support has come in with #1868 however when a spot instance terminates, it drops the workload without notice.
Augment the spot instance support with interruption notice polling. When an instance receives the normal 2 minute notice, attempt to drain it using the same lifecycle process that the provider uses when scaling down a pool.

https://docs.aws.amazon.com/AmazonCloudWatch/latest/events/EventTypes.html#spot-instance-event-types

**Environment:**

akash-gautam:
set Interruptible as true when SpotMarketOptions for AWSMachine is non-nil
fixes issue 1899

**Which issue(s) this PR fixes** *
Fixes #1899

### DIFF
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -337,3 +337,10 @@ func (m *MachineScope) AWSMachineIsDeleted() bool {
 func (m *MachineScope) IsEKSManaged() bool {
 	return m.InfraCluster.InfraCluster().GetObjectKind().GroupVersionKind().Kind == "AWSManagedControlPlane"
 }
+
+// SetInterruptible sets the AWSMachine status Interruptible
+func (m *MachineScope) SetInterruptible() {
+	if m.AWSMachine.Spec.SpotMarketOptions != nil {
+		m.AWSMachine.Status.Interruptible = true
+	}
+}
--- a/api/v1alpha2/awsmachine_conversion.go
+++ b/api/v1alpha2/awsmachine_conversion.go
@@ -40,6 +40,7 @@ func (src *AWSMachine) ConvertTo(dstRaw conversion.Hub) error {
 		return err
 	}
 	restoreAWSMachineSpec(&restored.Spec, &dst.Spec, &src.Spec)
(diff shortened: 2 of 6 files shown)


akash-gautam:
verify CLA again.

pkg/cloud/scope/machine.go
```diff
@@ -337,3 +337,10 @@ func (m *MachineScope) AWSMachineIsDeleted() bool {
 func (m *MachineScope) IsEKSManaged() bool {
 	return m.InfraCluster.InfraCluster().GetObjectKind().GroupVersionKind().Kind == "AWSManagedControlPlane"
 }
+
+// SetInterruptible sets the AWSMachine status Interruptible
+func (m *MachineScope) SetInterruptible() {
+	if m.AWSMachine.Spec.SpotMarketOptions != (&infrav1.SpotMarketOptions{}) {
```
richardcase, line 343: If `m.AWSMachine.Spec.SpotMarketOptions` is nil then interuptible will still be set which based on the comments in the issue would be wrong.

richardcase (changes requested):

richardcase:
@akash-gautam - thanks for the PR 👍  As you have made some changes to the CRDs you will need to run `make generate`. Also its looks like there may be a linting failure, you can run the linter locally using `make lint`

akash-gautam:
@richardcase Thanks a lot for reviewing the PR and providing feedback. I worked on your suggestion but still, two checks are failing for me.

richardcase:
@akash-gautam  - i'll have a quick look

richardcase:
The "fuzzy" test are failing on the new field in the status. The fuzzy tests convert v1alpha2<->v1alpha3 and with the new field in the machine status you'll have "restore" the field value when it round trips and converts from v1alpha2->v1alpha3. If you are on the Kubernetes slack we can chat further there as it may be easier.

richardcase:
Thanks @akash-gautam for this change.

PR #2124, opened 2020-11-26
Add externalInfraCluster

enxebre:
This decouples the machine controller from the Infrastructure resource.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2125

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -18,6 +18,7 @@ package controllers
 
 import (
 	"context"
+	"fmt"
 	"reflect"
 
 	"github.com/go-logr/logr"
@@ -26,7 +27,18 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/utils/pointer"
+	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha3"
+	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/controlplane/eks/api/v1alpha3"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/ec2"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/elb"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/secretsmanager"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/ssm"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/userdata"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
+	"sigs.k8s.io/cluster-api/controllers/external"
 	"sigs.k8s.io/cluster-api/controllers/noderefutil"
 	capierrors "sigs.k8s.io/cluster-api/errors"
 	"sigs.k8s.io/cluster-api/util"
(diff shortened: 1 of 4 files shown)


controllers/awsmachine_controller.go
```diff
@@ -553,16 +557,18 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 	if machineScope.InstanceIsOperational() {
 		machineScope.SetAddresses(instance.Addresses)
 
-		existingSecurityGroups, err := ec2svc.GetInstanceSecurityGroups(*machineScope.GetInstanceID())
-		if err != nil {
-			return ctrl.Result{}, err
-		}
+		if !machineScope.IsExternalInfraCluster() {
```
richardcase, line 560: Do you think we will need to do something similar for machine pools? If so, then we probably want to move the [getInfraCluster](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f02e98508e6b09b0c5aaebcf631469fa603c9695/controllers/awsmachine_controller.go#L770) to somewhere where it can be shared.

richardcase (commented):
A very interesting change and generally looks good. A couple of comments and the checks appear to be failing due to linting issues.

Makefile
```diff
@@ -146,7 +146,7 @@ managers:
 
 .PHONY: manager-aws-infrastructure
 manager-aws-infrastructure: ## Build manager binary.
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
+	CGO_ENABLED=0 go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
```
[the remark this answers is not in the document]
enxebre: nope, that can use a different pr, I dropped it.

randomvariable:
You're hitting this dreaded thing:

```
missing call(s) to *mock_services.MockEC2MachineInterface.GetInstanceSecurityGroups(is anything) /home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_controller_test.go
```

Otherwise agree with rich here.

controllers/awsmachine_controller.go
```diff
@@ -796,6 +801,29 @@ func (r *AWSMachineReconciler) getInfraCluster(ctx context.Context, log logr.Log
 		return managedControlPlaneScope, nil
 	}
 
+	if cluster.Spec.InfrastructureRef != nil &&
+		cluster.Spec.InfrastructureRef.Kind != AWSClusterInfrastructureRefKind {
+		externalInfraCluster, err := external.Get(ctx, r.Client, cluster.Spec.InfrastructureRef, cluster.Spec.InfrastructureRef.Namespace)
+		if err != nil {
+			return nil, fmt.Errorf("error fetching externalInfraCluster: %w", err)
```
randomvariable, line 808:
```suggestion
			log.Error(err, "error fetching externalInfraCluster")
			return nil, err
```
Doing this means we get the source code line rather than it pointing in controller runtime. Will make it easier to debug if there's a problem.

controllers/awsmachine_controller.go
```diff
@@ -796,6 +801,29 @@ func (r *AWSMachineReconciler) getInfraCluster(ctx context.Context, log logr.Log
...
+		// Create the cluster scope
+		extenalInfraClusterScope, err := scope.NewExternalInfraClusterScope(scope.ExternalInfraClusterScopeParams{
+			Client:               r.Client,
+			Logger:               log,
+			Cluster:              cluster,
+			ExternalInfraCluster: externalInfraCluster,
+			ControllerName:       "externalInfraCluster",
+		})
+		if err != nil {
+			return nil, err
```
randomvariable, line 820:
```suggestion
			log.Error(err, "unable to create external cluster scope")
			return nil, err
```

same as above

randomvariable (changes requested):
Some small logging suggestions to make it easier to debug because of controller runtime, otherwise lgtm.

randomvariable (approved):

PR #2135, opened 2020-12-02
:bug: Fix OpenAPI defaulting for secrets backend and add OpenAPI testing

### LINKED ISSUE #1999: Figure out how to validate kubebuilder validations (regex particularly)
/kind feature

Regexes are painful, but they're part of the kubebuilder validations.
We don't yet have tests for them, but we should.

Probably involves setting up envtest with the webhooks, creating resources and testing the result.

**Environment:**

### LINKED ISSUE #2064: AWSMachineTemplate defaulting and validation for `spec.cloudInit.insecureSkipSecretsManager` are in conflict

randomvariable:
Also includes fake CRDs for Cluster API clusters and machines

Creates a multi-package envtest helper, and converts controllers and webhook tests over from fake client to envtest*

Mirrors that of Cluster API where multiple controllers are in the same repo, so webhooks may need to modified so that they can run together.

`* The one exception is awsmachine_controller_test.go, which needs a major refactor to be able to use envtest, as it relies on sequential ordering of test statements. (https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2142)

Fixes #1999 
Fixes #2064

### DIFF
--- a/api/v1alpha3/awscluster_webhook_test.go
+++ b/api/v1alpha3/awscluster_webhook_test.go
@@ -17,10 +17,13 @@ limitations under the License.
 package v1alpha3
 
 import (
-	"github.com/aws/aws-sdk-go/aws"
+	"context"
 	"testing"
 
+	"github.com/aws/aws-sdk-go/aws"
 	. "github.com/onsi/gomega"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1alpha3"
 )
 
@@ -51,7 +54,13 @@ func TestAWSCluster_ValidateCreate(t *testing.T) {
 	}
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			if err := tt.cluster.ValidateCreate(); (err != nil) != tt.wantErr {
+			cluster := tt.cluster.DeepCopy()
+			cluster.ObjectMeta = metav1.ObjectMeta{
+				GenerateName: "cluster-",
+				Namespace:    "default",
+			}
+			ctx := context.TODO()
+			if err := testEnv.Create(ctx, cluster); (err != nil) != tt.wantErr {
 				t.Errorf("ValidateCreate() error = %v, wantErr %v", err, tt.wantErr)
 			}
 		})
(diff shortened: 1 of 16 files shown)


randomvariable:
@richardcase can you check everything is as expected in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2135/files#diff-494951c8cd9bd229b1d727afe059c557be518ac1a69e910243f6e3f895edfd07 . Had to modify some of the tests to pass API server validation.

/assign @sedefsavas 

ps: @ncdc, the copy-pasta from the CAPI repo is in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2135/files#diff-6ff1d74b482f5f7e1c2727b17204096fe4bf243e8b7bbb0099b5701b7d3f77b1 , which does three things:
- Mutates webhook configurations coming from different directories so that they co-exist
- Adds the waiter for the webhook port to stop test flakes

randomvariable:
will do a follow up issue about awsmachine_controller_test.go

api/v1alpha3/awscluster_webhook_test.go
```diff
@@ -389,8 +431,14 @@ func TestAWSCluster_DefaultAllowedCIDRBlocks(t *testing.T) {
...
 		t.Run(tt.name, func(t *testing.T) {
-			tt.beforeCluster.Default()
-			g.Expect(tt.beforeCluster.Spec.Bastion).To(Equal(tt.afterCluster.Spec.Bastion))
+			ctx := context.TODO()
+			cluster := tt.beforeCluster.DeepCopy()
+			cluster.ObjectMeta = metav1.ObjectMeta{
+				GenerateName: "cluster-",
+				Namespace:    "default",
+			}
+			g.Expect(testEnv.Create(ctx, cluster)).To(Succeed())
```
sedefsavas, line 440: Does this call Default()?

randomvariable: envtest's API Server is calling the running webhook

sedefsavas: right, and Defaulting is called during Create right? Forgot if it is only being called during get.

randomvariable: It's called on Create or Update depending on the kubebuilder verbs. Only conversion webhooks are called on get.

sedefsavas:
Eventually do we want to do this change in pkg/ tests too?

richardcase:
Seems good to me, also:

/lgtm

randomvariable:
> Eventually do we want to do this change in pkg/ tests too?

@sedefsavas I did consider it, and I thought maybe the cost-reward ratio is a little too much if we're thinking about unit tests for individual cloud service functions. I compared against CAPI and fake client is still being used in a number of places. But I don't feel strongly about it either.

richardcase:
And a:

/lgtm

PR #2151, opened 2020-12-04
📖 add config example to docs specifying IAM role

MarcusNoble:
Includes an example of how to make use of the `controllers.cluster-api-provider-aws.sigs.k8s.io` role when specifying `AWS_CONTROLLER_IAM_ROLE`.


richardcase:
Seems ok to me but will let someone else to the approve

/lgtm

PR #2184, opened 2021-01-04
🐛 AWSMachinePool: Remove unused ID field on launch template spec

michaelbeaumont:

This field isn't actually respected if set on an `AWSMachinePool`, and in particular a user cannot reference an existing launch template in an `AWSMachinePool`, as far as I can tell. The `ID` field and the `AWSLaunchTemplate` struct are sort of abused internally to hold the results of fetching a launch template based on `status.launchTemplateID` but we don't need to hold the ID.
Ideally this use case would work, but until it does, I think it best to remove this field.

### DIFF
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinepools.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinepools.yaml
@@ -130,9 +130,6 @@ spec:
                   iamInstanceProfile:
                     description: The name or the Amazon Resource Name (ARN) of the instance profile associated with the IAM role for the instance. The instance profile contains the IAM role.
                     type: string
-                  id:
-                    description: The ID of the launch template.
-                    type: string
                   imageLookupBaseOS:
                     description: ImageLookupBaseOS is the name of the base operating system to use for image lookup the AMI is not set.
                     type: string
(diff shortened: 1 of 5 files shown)


randomvariable:
Just to check, this isn't a breaking change is it?

michaelbeaumont:
I suppose technically it is since it changes the CRD. It's an unfortunate accident that it's there at all, I think. As far as I can tell, it doesn't end up persisted unless set by a user. If it is set by a user, it's ignored.
What's the policy for breaking changes in `/exp`?

richardcase:
> What's the policy for breaking changes in `/exp`?

Interesting question.  Does the fact that its contained in`/exp` mean its exempt from the recommendations on [API versions](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#alpha-beta-and-stable-versions)? 

If we follow those recommendations for **Alpha** versions then any incompatible changes (which i take as the same as a breaking change) would require us to increment the API version:

> developers should still increment the API version when object schema or semantics change in an incompatible way

Which would mean that it would have to go into **v1alpha4**. But I could be misinterpreting the recommendations and also the fact that stuff under `/exp` is different.

I think @randomvariable may have a better and more correct answer.

michaelbeaumont:
Interesting, if we can't make changes without incrementing the version then resources need to be carefully introduced into `/exp`. If launch templates could be referenced by ID I don't think this would be the ideal schema.

randomvariable:
We chatted on Monday. I forgot this was in exp, I meant it more in the sense is this ok for any of your users?

Given that meeting, I think we're ok with it.

/lgtm

I'll leave to @richardcase  for the approve

Issue #2192, opened 2021-01-06
AWSMachinePool doesn't update ASG template upon MachinePool version change


sedefsavas:
/kind bug

**What steps did you take and what happened:**
MachinePool K8s version change does not trigger an update on ASG machines.
This can be replicated with CAPI e2e MachinePool test.

**What did you expect to happen:**

**Anything else you would like to add:**
Here the new ImageId based on the version in MachinePool is discovered.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f6a97de9ba98338e58bf1afafb71ece60db5b394/exp/controllers/awsmachinepool_controller.go#L391

But when deciding if ASG template needs to be updated, we do not consider that new imageID:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f6a97de9ba98338e58bf1afafb71ece60db5b394/exp/controllers/awsmachinepool_controller.go#L418
**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sedefsavas:
Update: after fixing above mentioned issue, a new version of LaunchTemplate is created but the instances are not updated. So probably we also need to set a update strategy to trigger update when anything changes on the LaunchTemplate.

PR #2193, opened 2021-01-07
🐛Trigger rolling replacement upon launch template change

sedefsavas:
This PR refreshes machine pools instances when there are changes in the launch template (a.k.a, a new template version is created).

For instance, when `version` field is changed in MachinePool, a new AMI ID is calculated, and a new version of the initial template is created. But this does not trigger the rollout of the existing MachinePool machines. Need to explicitly refresh instances.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2192


exp/controllers/awsmachinepool_controller.go
```diff
@@ -427,6 +427,13 @@ func (r *AWSMachinePoolReconciler) reconcileLaunchTemplate(machinePoolScope *sco
 		if err := ec2svc.CreateLaunchTemplateVersion(machinePoolScope, imageID, userData); err != nil {
 			return err
 		}
+		machinePoolScope.Info("start instance refresh", "number of instances", machinePoolScope.MachinePool.Spec.Replicas)
+		asgSvc := r.getASGService(ec2Scope)
+		// After creating a new version of launch template, instance refresh is required
+		// to trigger a rolling replacement of all previously launched instances.
+		if err := asgSvc.StartASGInstanceRefresh(machinePoolScope); err != nil {
+			return err
```
richardcase, line 435: Do we need to check for `InstanceRefreshInProgress` or check the status up front for the scenario here we are going around the reconciliation loop again and the refresh is still happening?

pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -303,6 +304,19 @@ func (s *Service) UpdateASG(scope *scope.MachinePoolScope) error {
 	return nil
 }
 
+func (s *Service) StartASGInstanceRefresh(scope *scope.MachinePoolScope) error {
+	input := &autoscaling.StartInstanceRefreshInput{
```
richardcase, line 308: Would it be worth exposing the preferences (i.e. instance warmup and min health percentage)? Might be useful to users that want more control. WDYT?

pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -303,6 +304,19 @@ func (s *Service) UpdateASG(scope *scope.MachinePoolScope) error {
...
+		AutoScalingGroupName: aws.String(scope.Name()), //TODO: define dynamically - borrow logic from ec2
+		Strategy:             pointer.StringPtr(autoscaling.RefreshStrategyRolling),
+	}
+
+	if _, err := s.ASGClient.StartInstanceRefresh(input); err != nil {
```
richardcase, line 313: Should we set a condition when refreshing?

pkg/cloud/services/autoscaling/autoscalinggroup.go
randomvariable: If we can do it now, that'd be preferable.

sedefsavas: Do you mean exposing them in CRD?

exp/controllers/awsmachinepool_controller.go
sedefsavas: Good point, if there is an InstanceRefreshInProgress we should block updating the template, because until an existing refresh request finalizes, cannot do new refresh request. 
Doing this change now.

pkg/cloud/services/autoscaling/autoscalinggroup.go
richardcase: Sorry, yes in the CRD and then using them here.

sedefsavas:
Thanks for the comments. PTAL.

dthorsen:
@sedefsavas I have been away from the code for a while but did someone add a lifecycle hook for draining nodes? I'll try to make some time next week to catch up on the status of this stuff.

sedefsavas:
@dthorsen no, that'd be useful.
We have a PR https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2037 open for creating Eventbridges, I think this might helps with lifecycle hooks notifications.

randomvariable:
i think the eventbridge stuff might be separate to this. it's a similar issue to what cecile was asking about in the CAPI meeting the other week about wanting to represent the instances inside the machinepool.

this could be done with the AWS written controller maybe though?

/lgtm
/approve

PR #2202, opened 2021-01-13
✨ feat:  add EKS addon support

### LINKED ISSUE #2139: Support EKS Addons
/kind feature
/area provider/eks

Add support for EKS addons in CAPA.

See the [API docs](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateAddon.html)

**Environment:**

***Tasks***

- [ ] Validating webhook changes to ensure minimum k8s version
- [x] Control Plane type and controller changes - create
- [x] Control Plane type and controller changes - update
- [x] Control Plane type and controller changes - delete
- [x] clusterawesadm command to list available addons/version
- [ ] docs update
- [ ] e2e test added

richardcase:
This change adds support for installing/managing EKS addons as part of the control plane reconciliation.

It includes:
* EKS Addon reconciliation
* New clusterawsadm commands to list installed/available addons
* Additional permissions to the controller policy
* Webhook check to make sure we have a supported k8s version
* Updated e2e tests

Additionally a number of minor bugs have been fixed with the e2e tests/

Fixes #2139


cmd/clusterawsadm/cmd/eks/addons/list-available.go
```diff
@@ -0,0 +1,140 @@
...
+		Short: "List available EKS addons",
+		Long:  "Lists the addons that are available for use with an EKS cluster",
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return listAvailableAddons(&region, &clusterName, &outputPrinter)
+		},
+	}
+
+	newCmd.Flags().StringVarP(&region, "region", "r", "", "The AWS region containing the EKS cluster")
+	newCmd.Flags().StringVarP(&clusterName, "cluster-name", "n", "", "The name of the cluster to get the list of available addons for")
+	newCmd.Flags().StringVarP(&outputPrinter, "output", "o", "table", "The output format of the results. Possible values: table,json.yaml")
```
MarcusNoble, line 48:
```suggestion
	newCmd.Flags().StringVarP(&outputPrinter, "output", "o", "table", "The output format of the results. Possible values: table,json,yaml")
```

cmd/clusterawsadm/cmd/eks/addons/list-available.go
```diff
@@ -0,0 +1,140 @@
...
+			AddonName: addon,
+		}
+		describeOutput, err := eksClient.DescribeAddonVersions(describeInput)
+		if err != nil {
+			return fmt.Errorf("describing addon versions %s: %w", *addon, err)
+		}
+
+		for _, info := range describeOutput.Addons {
+			for _, version := range info.AddonVersions {
+				availableAddon := availableAddon{
```
MarcusNoble, line 99: Can we rename this variable to avoid the same name as the type?

cmd/clusterawsadm/cmd/eks/addons/list-installed.go
```diff
@@ -0,0 +1,142 @@
...
+
+func listInstalledCmd() *cobra.Command {
+	clusterName := ""
+	region := ""
+	outputPrinter := ""
+
+	newCmd := &cobra.Command{
+		Use:   "list-installed",
+		Short: "List installed EKS addons",
+		Long:  "Lists the addons that are installed for a EKS cluster",
```
MarcusNoble, line 40:
```suggestion
		Long:  "Lists the addons that are installed for an EKS cluster",
```

cmd/clusterawsadm/cmd/eks/addons/list-installed.go
```diff
@@ -0,0 +1,142 @@
...
+		RunE: func(cmd *cobra.Command, args []string) error {
+			return listInstalledAddons(&region, &clusterName, &outputPrinter)
+		},
+	}
+
+	newCmd.Flags().StringVarP(&region, "region", "r", "", "The AWS region containing the EKS cluster")
+	newCmd.Flags().StringVarP(&clusterName, "cluster-name", "n", "", "The name of the cluster to get the list of installed addons for")
+	newCmd.Flags().StringVarP(&outputPrinter, "output", "o", "table", "The output format of the results. Possible values: table,json.yaml")
```
MarcusNoble, line 48:
```suggestion
	newCmd.Flags().StringVarP(&outputPrinter, "output", "o", "table", "The output format of the results. Possible values: table,json,yaml")
```

docs/book/src/topics/eks/addons.md
```diff
@@ -0,0 +1,59 @@
+# EKS Addons
+
+[EKS Addons](https://aws.amazon.com/blogs/containers/introducing-amazon-eks-add-ons/) can be used with EKS clusters created using Cluster API Provider AWS.
+
+Addons are supported in EKS clusters using Kubernetes v1.18 or greater. 
+
+## Installing addons
+
+To install an addon you need to declare them by specifying the name, version and optionally how conflicts should be resolved in the `AWSManagedControlPlane`. For example:
```
MarcusNoble, line 9: How would conflicts arrise? Conflicts with what exactly?

pkg/cloud/services/eks/addons.go
```diff
@@ -0,0 +1,188 @@
...
+	for _, procedure := range procedures {
+		s.scope.V(2).Info("Executing addon procedure", "name", procedure.Name())
+		if err := procedure.Do(ctx); err != nil {
+			s.scope.Error(err, "failed executing addon procedure", "name", procedure.Name())
+			return fmt.Errorf("%s: %w", procedure.Name(), err)
+		}
+	}
+
+	// 5. Update status with addons installed details
+	// NOTE: we should be able to construct this froom the plan/procedures????
```
MarcusNoble, line 77: Could we add a `getAllDesired()` to `addonsPlan` so we don't need to make the extra API calls here?

MarcusNoble (changes requested):

randomvariable:
Would you want to control addons declaratively in the future? If so, might want to move the commands to underneath alpha if it's not going to be longterm.

pkg/cloud/services/eks/addons.go
richardcase: I did wonder whether we should construct the Status from the operations. I was undecided and also I wanted to leave `Plan\Procedure` generic. But I guess I can return `AddonsPlan` from the `New` instead of `Plan` and then have it on the `AddonsPlan`  instead. I'll make that change....thanks for the feedback on this

richardcase:
> Would you want to control addons declaratively in the future? If so, might want to move the commands to underneath alpha if it's not going to be longterm.

I think we'll want a utility for users to help them see what addons are available (and installed) for the longterm.  This would be purely read and any operations have to be via the `AWSManagedControlPlane`

docs/book/src/topics/eks/addons.md
richardcase: This is on the AWS API for addons and so I exposed it. Theres only a  single addon available at present so this may make more sense in the future.

cmd/clusterawsadm/cmd/eks/addons/list-available.go
richardcase: We sure can

docs/book/src/topics/eks/addons.md
MarcusNoble: Yeah, just took a look at the AWS SDK
```
    // How to resolve parameter value conflicts when applying the new version of
    // the add-on to the cluster.
    ResolveConflicts *string `locationName:"resolveConflicts" type:"string" enum:"ResolveConflicts"`
```
Still not clear exactly how a conflict may arise but I'm not too worried about it at this stage.

pkg/cloud/services/eks/addons.go
MarcusNoble: Just looked into it a bit more and it doesn't look like we're capturing the ARN during the plan/procedure bit.

richardcase: In the `CreateAddonProcedure` i can get that from the output and save it.

randomvariable:
Need #2216 to let the e2e-image build correctly

docs/book/src/topics/eks/addons.md
michaelbeaumont: The AWS API uses server side apply in the implementation. This means for the CNI, for example, (the only addon atm) enabling the addon can fail if say the `image` field is managed by some other `manager` (the addons API manages as `eks`) but will ignore this if you set `overwrite`.

controlplane/eks/api/v1alpha3/types.go
```diff
@@ -119,3 +119,46 @@ type UserMapping struct {
...
+	RoleArn *string `json:"roleARN,omitempty"`
+}
+
+// AddonResolution defines the method for resolving parameter conflicts
+type AddonResolution string
+
+var (
+	// AddonResolutionOverwite indicates that if there are parameter conflicts then
+	// resolution will be accomplished via overwriting
+	AddonResolutionOverwite = AddonResolution("overwrite")
```
michaelbeaumont, line 147:
```suggestion
	AddonResolutionOverwrite = AddonResolution("overwrite")
```

pkg/cloud/services/eks/addons.go
```diff
@@ -0,0 +1,188 @@
...
+	if len(procedures) == 0 {
+		s.scope.Info("addon plan has no actions required, no action needed")
+		return nil
+	}
+
+	// 4. Perform required operations
+	for _, procedure := range procedures {
+		s.scope.V(2).Info("Executing addon procedure", "name", procedure.Name())
+		if err := procedure.Do(ctx); err != nil {
+			s.scope.Error(err, "failed executing addon procedure", "name", procedure.Name())
```
michaelbeaumont, line 71: could the addon status be set to failed in this case?

pkg/cloud/services/eks/addons.go
```diff
@@ -0,0 +1,188 @@
...
+	// 5. Update status with addons installed details
+	// NOTE: we should be able to construct this froom the plan/procedures????
+	s.scope.V(2).Info("getting installed eks addons for status", "cluster", eksClusterName)
+	installed, err = s.getClusterAddons()
+	if err != nil {
+		return fmt.Errorf("getting installed eks addons: %w", err)
+	}
+	addonStatus := []ekscontrolplanev1.AddonStatus{}
+	for _, installedAddon := range installed {
+		addonStatus = append(addonStatus, ekscontrolplanev1.AddonStatus{
```
michaelbeaumont, line 85: may have missed it but can we show `Status` and if appropriate `Health` on the resource?

pkg/eks/addons/procedures.go
```diff
@@ -0,0 +1,160 @@
...
+	requestToken := string(uuid.NewUUID())
+	input := &eks.CreateAddonInput{
+		AddonName:          &desired.Name,
+		AddonVersion:       &desired.Version,
+		ClusterName:        &p.plan.clusterName,
+		ResolveConflicts:   desired.ResolveConflict,
+		Tags:               convertTags(desired.Tags),
+		ClientRequestToken: &requestToken,
+	}
+	if desired.ServiceAccountRoleARN != "" {
```
michaelbeaumont, line 141: It might be worth creating an issue to reexamine after some time the fact that it's relatively cumbersome to define these roles externally, especially because securing the role means setting a trust relationship that depends on the name/namespace of the service account, which is well known but not provided by the API.
At the moment I have to create a role beforehand, with knowledge of the needed policies and eventual metadata of the service account, before creating the addon.

cmd/clusterawsadm/cmd/eks/addons/list-available.go
richardcase: Thanks, i have changed this.

cmd/clusterawsadm/cmd/eks/addons/list-available.go
richardcase: I have renamed this.

cmd/clusterawsadm/cmd/eks/addons/list-installed.go
richardcase: Changed.

cmd/clusterawsadm/cmd/eks/addons/list-installed.go
richardcase: Thanks, i have changed this.

pkg/cloud/services/eks/addons.go
richardcase: I decided to go with querying the addons again for the scenario where there are no actions to take with the plan but where there might be an update to the addon...for example the status moving from UPDATING to ACTIVE.

controlplane/eks/api/v1alpha3/types.go
richardcase: Thanks, changed.

pkg/cloud/services/eks/addons.go
richardcase: I've added the status to the state of the addon and this will be set via the state returned by the API. So I don't want to manually set a status.  

In the scenario where there are multiple addons and one fails do you think we should carry on with the others or stop processing? At the moment we stop processing but I'm thinking we should continue on after reporting the error.

pkg/cloud/services/eks/addons.go
richardcase: I have added status and health/issues

pkg/eks/addons/procedures.go
richardcase: Yeah good thinking. Created #2219

richardcase:
e2e test are [passing](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2202/pull-cluster-api-provider-aws-e2e-eks/1351640791038038016).

cmd/clusterawsadm/cmd/eks/addons/list-available.go
```diff
@@ -0,0 +1,140 @@
+/*
```
MarcusNoble, line 1: Can we rename this file (and the other similar) to replace the `-` to either `_` or using camal case to match the rest of the files in the repo?

pkg/cloud/converters/eks.go
```diff
@@ -0,0 +1,50 @@
...
+import (
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/eks"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/controlplane/eks/api/v1alpha3"
+)
+
+func AddonSDKToAddonState(eksAddon *eks.Addon) *ekscontrolplanev1.AddonState {
```
MarcusNoble, line 28: Can we get a doc comment added to this function please?

MarcusNoble (changes requested):

cmd/clusterawsadm/cmd/eks/addons/list-available.go
richardcase: I've renamed to use `_`.

pkg/cloud/converters/eks.go
richardcase: Good spot, thanks. I've added a doc comment to this,

PR #2228, opened 2021-01-26
controllers: Add conditionals around fetching secret backend service

randomvariable:

Refactors some of the controller code around secrets backend handling. Early returns when we're not using it allowing insecureSkipSecrets to function in its entirety.

Fixes #


rsmitty:
Tested this with Talos and our bootstrap/controlplane providers and it worked out well. Thanks for the quick fix.

PR #2243, opened 2021-02-07
🌱 AWSManagedMachinePool: Make public access to SSH explicit

### LINKED ISSUE #1990: Default remote access security groups/bastion host for AWSManagedMachinePool

michaelbeaumont:
Restricts access to SSH for `AWSManagedMachinePool`s to specific security groups by default. Until now the default matches the AWS API defaults where if SSH access is enabled with some SSH key then access is allowed from any IP. Now the user must set `public: true`.

Fixes #1990 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
action required: New `AWSManagedMachinePool` resources with non-empty `remoteAccess` now require `remoteAccess.public: true` in order to allow public access to SSH on port 22
```

### DIFF
--- a/exp/api/v1alpha3/awsmanagedmachinepool_types.go
+++ b/exp/api/v1alpha3/awsmanagedmachinepool_types.go
@@ -129,8 +129,10 @@ type ManagedRemoteAccess struct {
 	SSHKeyName *string `json:"sshKeyName,omitempty"`
 
 	// SourceSecurityGroups specifies which security groups are allowed access
-	// An empty array opens port 22 to the public internet
 	SourceSecurityGroups []string `json:"sourceSecurityGroups,omitempty"`
+
+	// Public specifies whether to open port 22 to the public internet
+	Public bool `json:"public,omitempty"`
 }
 
 // AWSManagedMachinePoolStatus defines the observed state of AWSManagedMachinePool
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmanagedmachinepools.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmanagedmachinepools.yaml
@@ -83,8 +83,11 @@ spec:
               remoteAccess:
                 description: RemoteAccess specifies how machines can be accessed remotely
                 properties:
+                  public:
+                    description: Public specifies whether to open port 22 to the public internet
+                    type: boolean
                   sourceSecurityGroups:
-                    description: SourceSecurityGroups specifies which security groups are allowed access An empty array opens port 22 to the public internet
+                    description: SourceSecurityGroups specifies which security groups are allowed access
                     items:
                       type: string
                     type: array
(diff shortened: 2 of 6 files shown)


richardcase:
@michaelbeaumont - i think we a release note here as we are changing behavior from where it would be public if there was an SSH key to even if there is an SSH key they need to set `publicAccess: true`

exp/api/v1alpha3/awsmanagedmachinepool_types.go
```diff
@@ -129,8 +129,10 @@ type ManagedRemoteAccess struct {
 	SSHKeyName *string `json:"sshKeyName,omitempty"`
 
 	// SourceSecurityGroups specifies which security groups are allowed access
-	// An empty array opens port 22 to the public internet
 	SourceSecurityGroups []string `json:"sourceSecurityGroups,omitempty"`
+
+	// Public specifies whether to open port 22 to the public internet
+	Public bool `json:"public,omitempty"`
```
richardcase, line 135: Small nit, in the PR description its `publicAccess` but here its `public`. Guessing the preferred naming is public?

michaelbeaumont: Yes, I switched to `public` to get rid of the `remoteAccess.publicAccess` duplication.

michaelbeaumont:
> @michaelbeaumont - i think we a release note here as we are changing behavior from where it would be public if there was an SSH key to even if there is an SSH key they need to set `publicAccess: true`

True, I've now added an `action required` release note although I'm not 100% sure that fits since it only affects _new_ resources?

richardcase:
As soon as the e2e tests pass I will approve

richardcase:
Ah, ran the wrong tests:

richardcase:
the eks e2e tests [passed](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2243/pull-cluster-api-provider-aws-e2e-eks/1358753105021243392), so:

/approve

richardcase:
the eks e2e tests [passed](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2243/pull-cluster-api-provider-aws-e2e-eks/1358753105021243392), so:

/lgtm

Issue #1786, opened 2020-07-06
EKS Fargate profiles support


richardcase:
/kind feature

**Describe the solution you'd like**
When creating a EKS cluster i want the option to be able to use fargate for the workloads. Adding support for [fargate profiles](https://docs.aws.amazon.com/eks/latest/userguide/fargate.html) would be needed.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

michaelbeaumont:
Mapping this to `MachinePool` feels like a little bit of a mismatch given that we don't have `version`, `replicas`, `bootstrap`, etc. We have `Node`s but we don't really have `spec.replicas`. Do we think this is important?

richardcase:
> Mapping this to `MachinePool` feels like a little bit of a mismatch given that we don't have `version`, `replicas`, `bootstrap`, etc. We have `Node`s but we don't really have `spec.replicas`. Do we think this is important?

As this doesn't really match well to upstream CAPI kinds I wonder if we just create an `AWSFargateProfile` kind with associated controller? And maybe it just cares about its matching `AWSManagedControlPlane`. WDYT?

richardcase:
@michaelbeaumont @randomvariable  - i think we should probably capture the decision on how we implement this in an ADR

PR #2250, opened 2021-02-15
📖 Add AWSFargatePool ADR

michaelbeaumont:
Documents a major decision around EKS Fargate support, whether they map to a `MachinePool`.

Part of #1786 (https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1786#issuecomment-777599215)

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
New ADR to document the decision of how Fargate Profiles will be represented.
```

### DIFF
--- a/docs/adr/0004-fargate-pools.md
+++ b/docs/adr/0004-fargate-pools.md
@@ -1,0 +1,45 @@
+# 4. EKS AWSFargateProfiles
+
+- Status: accepted
+- Date: 2021-02-15
+- Authors: @michaelbeaumont
+- Deciders: @michaelbeaumont, @richardcase
+
+## Context
+
+At the time of writing CAPA has the AWS-specific infrastructure resources `AWSMachinePool`s
+and `AWSManagedMachinePool`s.
+These are referred to in `MachinePool`s as `infrastructureRef`s. The
+underlying AWS resources fit the `MachinePool` model very closely. Data about
+the pools, such as `replicas` is reflected back to the `MachinePool` and
+the responsible controllers reconcile the AWS resources using data specified on the
+`MachinePool`, like `.spec.template.spec.version` and `.spec.template.spec.bootstrap`.
+
+Issue #1786 is concerned with supporting EKS Fargate profiles in CAPA. Fargate
+profiles allow users to run pods on an on-demand, serverless compute platform
+called Fargate.
+Nodes are spun up dynamically for pods in specified namespaces or with specific
+labels. These selectors are packaged into a "profile" that also specifies which
+subnets are available for pods to run in.
+
+With CAPA, users would like to create a resource representing such a profile.
+
+This ADR concerns itself with how these resources are specified. The major
+decision is whether a Fargate profile can be represented in the `MachinePool` abstraction.
+
+Some of the fields on `MachinePool` don't really make sense in the Fargate
+context, like the entire `MachineSpec` under `spec.template.spec` (`version`,
+`bootstrap`), `spec.strategy` and the idea of desired replicas.
+
+## Decision
+
+The `AWSFargateProfile` resource is implemented independently from any existing CAPI resource
+and users must only create and manipulate `AWSFargateProfile`s to create and manage
+Fargate profiles. `AWSFargateProfile`s are nevertheless associated with an
+`AWSManagedControlPlane`/`AWSManagedCluster`/`Cluster`.
+
+## Consequences
+
+Users don't have to create a sort of "empty" `MachinePool` resource that points
+to an `AWSFargateProfile` and no ambiguity or confusion arises as to the semantics of the `MachinePool`
+spec in a Fargate context.


docs/adr/0004-fargate-pools.md
```diff
@@ -0,0 +1,44 @@
...
+This ADR concerns itself with how these resources are specified. The major
+decision is whether a Fargate profile can be represented in the `MachinePool` abstraction.
+
+Some of the fields on `MachinePool` don't really make sense in the Fargate
+context, like the entire `MachineSpec` under `spec.template.spec` (`version`,
+`bootstrap`), `spec.strategy` and the idea of desired replicas.
+
+## Decision
+
+The `FargatePool` resource is implemented independently from any existing CAPI resource
```
richardcase, line 36: I think we should probably go with `AWSFargatePool` for the kind

docs/adr/0004-fargate-pools.md
```diff
@@ -0,0 +1,44 @@
...
+Some of the fields on `MachinePool` don't really make sense in the Fargate
+context, like the entire `MachineSpec` under `spec.template.spec` (`version`,
+`bootstrap`), `spec.strategy` and the idea of desired replicas.
+
+## Decision
+
+The `FargatePool` resource is implemented independently from any existing CAPI resource 
+and users must only create and manipulate `FargatePool`s to create and manage
+Fargate profiles.
+
```
richardcase, line 39: Is it worth mentioning that even though we aren't using the MachinePool CAPI kind that the AWSFaregatePool will still be associated with a Cluster/AWSManagedControlPlane?

richardcase (commented):
Looks good @michaelbeaumont . A few minor comments...

docs/adr/0004-fargate-pools.md
richardcase: Thanks for updating

docs/adr/0004-fargate-pools.md
richardcase: Thanks

richardcase:
This seems good to me. Lets get a few others to review and get some consensus...until then:

/hold

docs/adr/0004-fargate-pools.md
```diff
@@ -0,0 +1,45 @@
+# 4. EKS AWSFargatePools
+
+- Status: proposed
+- Date: 2021-02-15
+- Authors: @michaelbeaumont
+- Deciders:
+
+## Context
+
+At the time of writing CAPA has the AWS-specific infrastructure resources `AWSMachinePool`s
```
richardcase, line 10: If we are not mapping to a MachinePool then shall we remove 'Pool' from the kind name and go with something like `AWSFargateProfile`?

randomvariable:
I wonder if @CecileRobertMichon and @devigned have any ideas on related concepts in CAPZ and proposed changes for MachinePools overall.

docs/adr/0004-fargate-pools.md
michaelbeaumont: We could. I like the word "pool" better in general but it does have the `MachinePool` connotation. Profile is kind of meaningless IMO, but since that's what the AWS object is called, it wouldn't be wrong.

richardcase:
> I wonder if @CecileRobertMichon and @devigned have any ideas on related concepts in CAPZ and proposed changes for MachinePools overall.

I wonder if using Azure Container Instances with AKS is similar? ....don't know enough about ACI to answer that.

docs/adr/0004-fargate-pools.md
richardcase: Yeah i see your point....'profile' is a bit meaningless. It is a 'pool' of compute resource to use but it sort of implies MachinePool to me. But if machinepools are changing in v1alpha4 then perhaps it would be better to keep it as is?

devigned:
Wow, this is timely. @CecileRobertMichon and I have been prototyping a similar idea on Azure Container Instances (https://github.com/kubernetes-sigs/cluster-api-provider-azure/compare/master...devigned:aci?expand=1). There is no proposal yet, but you can see the direction via the diff above.

The way we are currently approaching the problems is by running virtual kubelet as a container on ACI to then act as a virtual node in the cluster with infinite resources. We have a AzureContainerInstanceMachine which acts like any normal workload machine, bootstrap + infrastructure reference. We are mounting the kubeadm bootstrap data into the container, running cloud-init in the container :vomiting_face:, and joining the virtual kubelet to the cluster. This is not totally working yet, but super close.

I had thought about using the machine pool abstraction, but the ideas didn't align in my mind. The abstraction for machine pool is more focused on running machines, VMs, not containers. If we could start a node image image on ACI or Fargate, like Kind does, then run docker in docker, then I think this would be the right abstraction. Unfortunately, for ACI, I don't know about Fargate, we are unable to run privileged workloads. That pretty much kills the docker in docker scenario, afaik.

We should totally discuss serverless CAPI :rocket:

michaelbeaumont:
It sounds like Fargate operates at a higher abstraction level in EKS than ACI in AKS. Fargate profiles are basically just lists of pod label and namespace selectors which match pods with a certain annotation and then run these pods in their own environment. In fact, every pod running in a Fargate profile appears as a Kubernetes node.
There's no clear link between nodes and their profile though, but a replica count wouldn't be especially meaningful anyway.

richardcase:
@michaelbeaumont - i would agree that it operates on a higher level of abstraction. But the commonality here i think is that ACI and Fargate don't match the machine pool. Based on this i think what is proposed in the ADR is good.

CecileRobertMichon:
It seems like the goal here is slightly different than what @devigned and I are prototyping. This seems to be about enabling Fargate-backed virtual nodes on EKS (the equivalent of the AKS virtual node addon). We're looking at implementing an ACI infra provider for machines (similar to CAPD machines, but in ACI).

docs/adr/0004-fargate-pools.md
CecileRobertMichon: I think keeping the "Pool" name might get confusing, especially if there is no mapping with `MachinePools`. MachinePools might change but they will still have different objectives and there is no plan to change the name. See https://github.com/kubernetes-sigs/cluster-api/issues/4063 for details.

richardcase: @michaelbeaumont - what do you think could be a good alternative to using Pool or Profile?  Just some ideas: `AWSFargateRule`, `AWSFargatePodExecution`, `AWSFargateExecutionConfig`

michaelbeaumont: Let's go with the original suggestion, `AWSFargateProfile`

richardcase:
This has been decided now, so:

/unhold
/lgtm
/approve

PR #2285, opened 2021-03-05
Set the UserAgent to "capa-controller"

sedefsavas:
/kind cleanup

Doing the same change with CAPI: https://github.com/kubernetes-sigs/cluster-api/pull/4257
> Set the UserAgent to "capa-controller".
> This string is used to identify actors within a Kubernetes system; for instance it appears in managedFields.

-->
```release-note
NONE
```

### DIFF
--- a/bootstrap/eks/main.go
+++ b/bootstrap/eks/main.go
@@ -124,7 +124,9 @@ func main() {
 		}()
 	}
 
-	mgr, err := ctrl.NewManager(ctrl.GetConfigOrDie(), ctrl.Options{
+	restConfig := ctrl.GetConfigOrDie()
+	restConfig.UserAgent = "cluster-api-provider-aws-controller"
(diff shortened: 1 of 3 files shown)


randomvariable:
We have version strings in CAPA already, so I think we should feed these through, see

```
// DefaultKubernetesUserAgent returns a User-Agent string built from static global vars.
func DefaultKubernetesUserAgent() string {
	return buildUserAgent(
		adjustCommand(os.Args[0]),
		adjustVersion(version.Get().GitVersion),
		gruntime.GOOS,
		gruntime.GOARCH,
		adjustCommit(version.Get().GitCommit))
```

for example.

randomvariable:
Had a look at other things, and we don't need the version string.

Can we change the string to "cluster-api-provider-aws-controller" ?

PR #2292, opened 2021-03-10
feat: ability to delete amazon vpc cni

### LINKED ISSUE #2287: Declarative way to delete AWS VPC CNI
/kind feature
/area provider/eks
/priority important-soon
/help

When provisioning a EKS cluster for Kubernetes v1.17 or less using CAPA where and i want to use a CNI that is different from the default Amazon VPC CNI (i.e something like Calico) i would like an easy way to ensure that the AWS VPC CNI is deleted.  For an unmanaged cluster (i.e. EC2/Kubeadm) this can be done using the **postKubeadmCommands** but this isn't available on the managed side. 

One potential way to do this is via a job that could be applied via a **ClusterResourceSet** or something like Flux2. And a longer-term solution may come in the way of the [machineadm bootstrapper poposal](https://github.com/kubernetes-sigs/cluster-api/pull/4221).

However, short term it would be good if we had a `disableVPVCNI` (or a name similar) that caused the AWS VPC CNI to be removed as part of the `AWSManagedControlPlane` reconciliation.

This means deleting the `aws-node` daemonset from the `kube-system` namespace.

We would need to add some checks to the validating webhook that would error if the new flag was set AND the vpc cni addon was defined.

**Environment:**

- Cluster-api-provider-aws version: 0.6.4

**Tasks**

- [x] API Changes
- [x] Controller reconciliation changes
- [x] Record events for the deletion
- [x] Validation webhook check for VPC CNI and error
- [x] Validation webhook check for secondary CIDR and warn
- [x] Update docs

richardcase:
**What type of PR is this?**

/kind feature

A new property called **disableVPCCNI** has been introduced that can be used to ensure that the Amazon VPC CNI is deleted as part of the **AWSManageControlPlane** reconciliation. This is useful in scenarios where a different CNI is used.

Fixes #2287 

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Ability to declaratively remove the Amazon VPC CNI when using an alternate CNI
```

### DIFF
--- a/pkg/cloud/services/awsnode/cni.go
+++ b/pkg/cloud/services/awsnode/cni.go
@@ -21,8 +21,6 @@ import (
 	"fmt"
 
 	amazoncni "github.com/aws/amazon-vpc-cni-k8s/pkg/apis/crd/v1alpha1"
-	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/awserrors"
-	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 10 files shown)


pkg/cloud/services/awsnode/cni.go
```diff
@@ -166,3 +180,28 @@ func (s *Service) filterEnv(env []corev1.EnvVar) []corev1.EnvVar {
...
+
+	s.scope.V(2).Info("The aws-node DaemonSet found, deleting")
+	if err := s.client.Delete(ctx, ds, &client.DeleteOptions{}); err != nil {
+		if errors.IsNotFound(err) {
+			s.scope.V(2).Info("The aws-node DaemonSet is not found, not deleted")
+			return nil
+		}
+		return fmt.Errorf("deleting aws-node DaemonSet: %w", err)
+	}
+	record.Eventf(s.scope.InfraCluster(), "DeletedVPCCNI", "The AWS CPC CNI has been removed from the cluster. Ensure you enable a CNI via another mechanism")
```
sedefsavas, line 204:
```suggestion
	record.Eventf(s.scope.InfraCluster(), "DeletedVPCCNI", "The AWS VPC CNI has been removed from the cluster. Ensure you enable a CNI via another mechanism")
```

richardcase: Good spot, thanks @sedefsavas - changed

PR #2302, opened 2021-03-15
fix: machine pool failure domains honoured

### LINKED ISSUE #2145: AWSMachinePool do not check use availabilityZones field
/kind bug

I don't see where we use `availabilityZones` field in the AWSMachinePool code.
For unmanaged VPCs, AZs are selected based on `AZSelectionScheme`, so the `availabilityZones` defined for AWSMachinePool cannot always be satisfied if those zones are not selected as part of Cluster VPC.

Expect AWSMachinePool to satisfy `availabilityZones` provided.

**Environment:**

### LINKED ISSUE #2191: Set ASG subnet to only private subnets
In AWSMachinePool, if subnets are not specified, we add all available subnets to the AutoScalingGroup.
Instead, the default should be only private subnets IMO. 

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/da8c0c4539a2aaaf64f570d95da7596d2bc683d0/pkg/cloud/services/autoscaling/autoscalinggroup.go#L148-L152
What do you all think?

### LINKED ISSUE #2277: `MachinePool` FailureDomains ignored for managed machine pool
/kind bug
/area provider/eks
/priority important-soon
/help

When creating an EKS nodegroup via `AWSManagedMachinePool` and no subnets are set in the spec then we should make sure that we select subnets that honor the [FailureDomains](https://github.com/kubernetes-sigs/cluster-api/blob/master/exp/api/v1alpha3/machinepool_types.go#L64) on the parent `MachinePool`. Currently this is [ignored](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/eks/nodegroup.go#L88:L93)

There is a scenario where users want to run a nodegroup per AZ

**Environment:**

- Cluster-api-provider-aws version: 0.6.4

**Tasks**

- [ ] Modify NodeGroupService.subnets to use failure domain from MachinePool if no subnets on managed machine 
- [ ] Validate failuredomains from MachinePool are in the list on the ControlPlane
- [ ] Create e2e test

### LINKED ISSUE #2301: `MachinePool` FailureDomains ignored for machine pool
/kind bug
/priority important-soon
/milestone v0.6.5

When creating an EKS cluster and using AWSMachinePool and no subnets are set in the spec then we should make sure that we select subnets that honour the FailureDomains on the parent MachinePool. Currently this is [ignored](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/fd4cc9d3f8db29b226b3e1d24dd8e14602beb2e3/pkg/cloud/services/autoscaling/autoscalinggroup.go#L153)

There is a scenario where users want to run a nodegroup per AZ.

Similar to #2277 

**Environment:**

- Cluster-api-provider-aws version: 0.6.4

richardcase:
**What type of PR is this?**
/kind bug

When using machine pools (managed and unmanaged) this is the precedence of choosing which subnets the machine pool instances will sit it:

1. Subnets defined explicitly in the `spec` of `AWSMachinePool/AWSManagedMachinePool`
2. If `AvailabilityZones` is specified on `AWSMachinePool/AWSManagedMachinePool` then use the subnets associated with those AZs
3. If `failureDomains` are specified in the `MachinePool` then subnets that are in the az (a.k.a failureDomain) that matche this are used
4. All the private subnets from the control plane are used 

Fixes #2277 
Fixes #2301
Fixes #2145
Fixes #2191  

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
action required 
The behaviour when selecting which subnets to use with machine pools (AWSMachinePool & AWSManagedMachinePool) when subnets ids haven't been included has changed. There is now a defined order of precedenace that will determine which subnets to use:

1. Subnets defined explicitly in the `spec` of `AWSMachinePool/AWSManagedMachinePool`
2. If `AvailabilityZones` is specified on `AWSMachinePool/AWSManagedMachinePool` then the subnets associated with those AZs will be used
3. If `failureDomains` are specified in the `MachinePool` then subnets that are in the az (a.k.a failureDomain) that match this are used
4. All the private subnets from the control plane are used 

```


richardcase:
Whilst the e2e tests are run:

/hold

richardcase:
The [e2e tests passed](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2302/pull-cluster-api-provider-aws-e2e-eks/1371478427349028864)

sedefsavas:
This PR also fixes: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2145

pkg/cloud/scope/shared.go
```diff
@@ -0,0 +1,117 @@
...
+// defaultSubnetPlacementStrategy is the default strategy for subnet placement
+type defaultSubnetPlacementStrategy struct {
+	logger logr.Logger
+}
+
+// Place works out the subnet placement based on the following precedence:
+// 1. Explicit definition of subnet IDs in the spec
+// 2. If the spec has Availability Zones then get the subnets for these AZs
+// 3. If the parent resource has Availability Zones then get the subnets for these AZs
+// 4. All the subnets from the control plane are used
```
sedefsavas, line 68: Should this be all the subnets or only private ones?
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2191

pkg/cloud/scope/shared.go
```diff
@@ -0,0 +1,117 @@
...
+	if len(input.SpecSubnetIDs) > 0 {
+		p.logger.V(2).Info("using subnets from the spec")
+		return input.SpecSubnetIDs, nil
+	}
+
+	if len(input.SpecAvailabilityZones) > 0 {
+		p.logger.V(2).Info("determining subnets to use from the spec availability zones")
+		subnetIDs, err := p.getSubnetsForAZs(input.SpecAvailabilityZones, input.ControlplaneSubnets)
+		if err != nil {
+			return nil, fmt.Errorf("getting subnets for spec azs: %w", err)
```
sedefsavas, line 80: Do we want to return an error if there are no subnets for the specified availability zones, or use the next option? Because this is a breaking change if users had wrong Availability zones as we did not have this check before.

pkg/cloud/scope/shared.go
richardcase: I kept the 'use all control plane subnets' existing behaviour. However, happy to change it to just the private subnets. Is that what we want the default behaviour to be?

pkg/cloud/scope/shared.go
richardcase: How about we log the error and then like you say continue to the next rule?

sedefsavas: Sounds good to me.

pkg/cloud/scope/shared.go
sedefsavas: We default to private subnet in Machines: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/48f64dc6d364d7448bb43c5c7a80bfbc979024a2/pkg/cloud/services/ec2/instances.go#L253

But this may be considered as breaking change.

richardcase: I have made this change to use private subnets. In addition, I have created an `action required` release note so that we can inform users of the behaviour change as part of the release notes.

pkg/cloud/scope/shared.go
richardcase: I was thinking about this, and I've left the error as is. The logic being if you have explicitly said you want the machine pool in a specific AZ/failure domain and there are no subnets in that az/failure domain then I would see that as an error. What do you think @sedefsavas @randomvariable ?

(like the review point above I have changed the release note to `action required` so that when we do a release users are aware of the change in behaviour)

PR #2324, opened 2021-03-23
Make conversion-gen output location explicit

### LINKED ISSUE #2323: Generated code may go to `$GOPATH/src` instead of in repo
/kind bug

Found in https://github.com/kubernetes-sigs/cluster-api/issues/4185

We [conversion-gen](https://pkg.go.dev/k8s.io/code-generator/cmd/conversion-gen) to generate some of the code. This command has an optional argument of `--output-base` that controls where the generated files are placed.

The default for this arg can vary. If `GOPATH` is set, then it will place files under `$GOPATH/src`. But if not set, it defaults to `./`, which is what we want to have the generated files within the repo.

The dependency on an environment variable is non-obvious and can cause stale files to be used if the user doesn't know about it. There also appears to be a difference whether running with the repo cloned within `$GOPATH/src` or elsehwere.

Files generated/over-written within the repo. In environment with GOPATH set, files were generated under `$GOPATH/src/api`.

stmcginnis:
**What type of PR is this?**

/kind bug

The conversion-gen command has an --output-base argument to control
where generated files are placed. The default value for this argument
can vary depending on whether or not $GOPATH is set or not. This results
in our generated files being created in the wrong place for some people
in a subtle and non-obvious way.

For those running `make generate` with GOPATH set, the files are created
in `$GOPATH/src/[file path]`. For those without GOPATH set, files are
created where we expect them to be within the repo at `./[file path]`.

To make sure files are always generated to the location where we expect
them to be, this updates our calls to conversion-gen to always be
explicit when not within GOPATH by setting  `--output-base=[repo_path]`.

Fixes #2323

Should be able to reproduce locally to validate by cloning repo outside of GOPATH. Check and remove any remnants found under `ls "$(go env GOPATH)/src"`, then run `make generate-go-core`. Prior to applying this change, you will see `awsclient` and others show up under `$GOPATH/src`, after you will see updates within the repo.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,6 +61,11 @@ PATH := $(abspath $(TOOLS_BIN_DIR)):$(PATH)
 DOCKER_CLI_EXPERIMENTAL=enabled
 DOCKER_BUILDKIT=1
 
+# Set --output-base for conversion-gen if we are not within GOPATH
+ifneq ($(abspath $(REPO_ROOT)),$(shell go env GOPATH)/src/sigs.k8s.io/cluster-api-provider-aws)
+	GEN_OUTPUT_BASE := --output-base=$(REPO_ROOT)
+endif
+
 # Release variables
 
 STAGING_REGISTRY ?= gcr.io/k8s-staging-cluster-api-aws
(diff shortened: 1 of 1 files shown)


sedefsavas:
Thanks for porting this change @stmcginnis

Issue #2227, opened 2021-01-25
Can not delete cluster when VPC creation failed


jzhoucliqr:
/kind bug

**What steps did you take and what happened:**
A cluster creation failed because of VPC quota limit, when try to delete the cluster, it stuck at delete security groups because there is no VPC-ID. 

```
[capa-controller-manager-ccd64cb9-bzs9v] 1 controller.go:257] controller-runtime/controller "msg"="Reconciler error" "error"="error deleting security groups for AWSCluster cluster-6005cc2b93ed23f55890cde8/aws-test-centos-os-patch-01: failed to describe cluster-owned security groups in vpc \"\": InvalidParameterValue: vpc-id\n\tstatus code: 400, request id: dfe92f29-2b56-4aea-9115-c355e38a8d22" "controller"="awscluster" "name"="aws-test-centos-os-patch-01" "namespace"="cluster-6005cc2b93ed23f55890cde8"

```

We are using capa 0.6.3, but by looking at the code, 0.6.4 should have the same issue. 

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version:  0.6.3
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

PR #2327, opened 2021-03-24
🐛fix for vpc deletion failure

LochanRn:
**What type of PR is this?**
/kind bug

Fixes cluster deletion failure.
If a cluster is created with VPC quota limit being exhausted vpc id is not provided, when we try to delete the cluster, it gets stuck at delete security groups because there is no VPC-ID.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2227

**Checklist**:
- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/services/securitygroup/securitygroups.go
+++ b/pkg/cloud/services/securitygroup/securitygroups.go
@@ -194,6 +194,11 @@ func (s *Service) DeleteSecurityGroups() error {
 		}
 	}
 
+	if s.scope.VPC().ID == "" {
+		s.scope.V(0).Info("skipping cluster owned security group delete, vpc id nil", "vpc-id", s.scope.VPC().ID)
+		return nil
+	}
+
 	clusterGroups, err := s.describeClusterOwnedSecurityGroups()
 	if err != nil {
 		return err
--- a/pkg/cloud/services/securitygroup/securitygroups.go
+++ b/pkg/cloud/services/securitygroup/securitygroups.go
@@ -162,6 +162,12 @@ func (s *Service) ReconcileSecurityGroups() error {
 
 func (s *Service) DeleteSecurityGroups() error {
 	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
+
+	if s.scope.VPC().ID == "" {
+		conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
+		return nil
+	}
+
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -162,6 +162,12 @@ func (s *Service) ReconcileSecurityGroups() error {
 
 func (s *Service) DeleteSecurityGroups() error {
 	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
+
+	if s.scope.VPC().ID == "" {
+		conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
```
sedefsavas, line 167: Cluster is not yet being deleted here:
```suggestion
		conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
```

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -162,6 +162,12 @@ func (s *Service) ReconcileSecurityGroups() error {
...
+		return nil
+	}
+
 	if err := s.scope.PatchObject(); err != nil {
```
sedefsavas, line 171: Just noticed that we are not patching the conditions on early return. We can move this to a defer func above.

pkg/cloud/services/securitygroup/securitygroups.go
LochanRn: Actually we are not setting cluster deletion here. We are just setting securityGroup condition to be deleted as the security group does not exist.

sedefsavas: You are right.

pkg/cloud/services/securitygroup/securitygroups.go
sedefsavas: This is not true as well. It is being patched on close.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -162,10 +162,17 @@ func (s *Service) ReconcileSecurityGroups() error {
 
 func (s *Service) DeleteSecurityGroups() error {
 	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
+
```
enxebre, line 165: remove extra space?

PR #2447, opened 2021-06-01
fix: added missing kms permissions required for EKS encryption

### LINKED ISSUE #2432: Permission error when enable EKS encryption
/kind bug
/area provider/eks
/priority important-soon
/milestone v0.7.0

Create an EK cluster and enable encruption using a KMS key. For example:
```yaml
 encryptionConfig:
    provider: "arn:aws:kms:us-east-1:12345678:alias/aws/eks"
    resources:
    - "secrets"
```

You then receive the following error:
```
 Warning  FailedCreateEKSControlPlane                   2s    aws-controller  Failed to initiate creation of a new EKS control plane: InvalidRequestException: User not authori │
│ zed to perform kms:CreateGrant operatio
```

I expect the cluster to provision with an error or warning and the EKS encryption settings to be set. 

**Environment:**

- Cluster-api-provider-aws version: 0.6.6

richardcase:
Signed-off-by: Richard Case <richard@weave.works>

**What type of PR is this?**
/kind bug

If you are using encryption with EKS there are missing permission on the controllers policy to allow using the specified KMS key.

Added missing IAM permission required to use EKS encryption. The permissions are scoped to KMS keys that have an alias name with a prefix of cluster-api-provider-aws-*.

The prefix is configurable via clusterawsadm config file.

Fixes #2432 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
action required
Controllers policy updated with missing KMS permissions required to use EKS encryption, if you are planning to use EKS encryption then you will need to update your controllers policy by running `clusterawsadm bootstrap iam create-cloudformation-stack` again.  And then when you create a cluster with encryption enabled you will need to use a KMS key that has an alias name starting with `cluster-api-provider-aws-`. For further information see docs. 

```

### DIFF
--- a/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
+++ b/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
@@ -379,6 +379,21 @@ func (t Template) ControllersPolicy() *infrav1.PolicyDocument {
 				},
 				Effect: infrav1.EffectAllow,
 			},
+			{
+				Action: infrav1.Actions{
+					"kms:CreateGrant",
+					"kms:DescribeKey",
+				},
+				Resource: infrav1.Resources{
+					"*",
+				},
+				Effect: infrav1.EffectAllow,
+				Condition: infrav1.Conditions{
+					"ForAnyValue:StringLike": map[string]string{
+						"kms:ResourceAliases": fmt.Sprintf("alias/%s", t.Spec.EKS.KMSAliasPrefix),
+					},
+				},
+			},
 		}...)
 	}
 
--- a/cmd/clusterawsadm/api/bootstrap/v1alpha1/defaults.go
+++ b/cmd/clusterawsadm/api/bootstrap/v1alpha1/defaults.go
@@ -30,6 +30,8 @@ const (
 	DefaultStackName = "cluster-api-provider-aws-sigs-k8s-io"
 	// DefaultPartitionName is the default security partition for AWS ARNs.
 	DefaultPartitionName = "aws"
+	// DefaultKMSAliasPattern is the default KMS alias.
+	DefaultKMSAliasPattern = "cluster-api-provider-aws-*"
 )
 
 func addDefaultingFuncs(scheme *runtime.Scheme) error {
@@ -84,6 +86,9 @@ func SetDefaults_AWSIAMConfigurationSpec(obj *AWSIAMConfigurationSpec) { //nolin
 		obj.SecureSecretsBackends = []infrav1.SecretBackend{
 			infrav1.SecretBackendSecretsManager,
 		}
+	}
+	if len(obj.EKS.KMSAliasPrefix) == 0 {
+		obj.EKS.KMSAliasPrefix = DefaultKMSAliasPattern
 	}
 }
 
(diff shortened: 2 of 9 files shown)


cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
```diff
@@ -360,6 +360,16 @@ func (t Template) ControllersPolicy() *infrav1.PolicyDocument {
 				},
 				Effect: infrav1.EffectAllow,
 			},
+			{
+				Action: infrav1.Actions{
+					"kms:CreateGrant",
+					"kms:DescribeKey",
+				},
+				Resource: infrav1.Resources{
+					"arn:aws:kms:*:*:key/*",
```
randomvariable, line 369: `"arn:aws:kms:*:*:key/*",`

allows the creation of grants across all KMS keys which would defeat the usage of customer managed keys. I think this will need to be configurable. Is there some default key that this can be restricted to?

richardcase: I don't think there is a default. (@martin-ducar-gd do you know?)

We could do this instead perhaps:
- restrict to a default key name / prefix and put in the docs saying the key must have an alias of `XYZ` 
- have a feature flag or override where you explicitly change this key name/pattern

What do you think?

richardcase: @martin-ducar-gd - i'd be keen to get your thoughts on this.

richardcase:
For the time being:

/hold

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
martin-ducar-gd: @richardcase @randomvariable there is default alias from AWS if you do ```sh
aws kms list-aliases
```
there is this for example:

Comment (martin-ducar-gd) [reply cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go:369]: aliases are visible from API/CLI only unless you use them only then they are in Console UI

Comment (martin-ducar-gd) [reply cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go:369]: the final key alias points to still looks like wildcard you specified.
```
arn:aws:kms:<region>:<accountid>:key/<uuid>
```
no special key ids for aws managed keys

martin-ducar-gd: In your IAM code not sure if you specify alias "arn:aws:kms:*:*:alias/capa-alias" as an example if it will work as really its just an alias real kms is nothing that can be abstracted, but I would say we need to test

richardcase: Interesting, i don't have that alias if i run that command.

richardcase: I will make some changes based on the above.

richardcase:
@randomvariable @sedefsavas  - currently the KMS permissions are only added when EKS is enabled. It feels like this should always be added. WDYT?

randomvariable:
> It feels like this should always be added. WDYT?

I don't think so, since the more common use case is to use default KMS encryption key for EBS volume provisioning. In terms of KMS encryption provider for EC2 CAPI, i'm not convinced of the benefit over root volume encryption. SIG Security also does not want to make a strong recommendation to use it because of concerns over performance.

richardcase:
> I don't think so, since the more common use case is to use default KMS encryption key for EBS volume provisioning

Ok, understood. In that case i will move the configuration of the custom prefix under `EKS` in the config file. Changing back to WIP

PR #2463, opened 2021-06-02
Add root storage device tags through additionalTags in ec2 instance

### LINKED ISSUE #1392: Forward port of 1394 - EC2 root volume should inherit additionalTags from instance
/kind feature

Additional tags should pass through to EC2 instance root volumes. This makes analyzing resources in a shared account much simpler. 

I already have this working in a branch, but wanted to allow for any discussion.

Ankitasw:
**What type of PR is this?**

/kind bug

Adds fields related to root storage device in EC2 instance

Fixes #1392

**Checklist**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Add root storage device tags through additionalTags in ec2 instance
```


yuzhang17:
wip

controllers/awsmachine_controller.go
```diff
@@ -911,3 +918,32 @@ func (r *AWSMachineReconciler) indexAWSMachineByInstanceID(o client.Object) []st
...
+	i := &infrav1.Volume{
+		DeviceName:    aws.StringValue(v.VolumeId),
+		Size:          aws.Int64Value(v.Size),
+		Type:          aws.StringValue(v.VolumeType),
+		IOPS:          aws.Int64Value(v.Iops),
+		Encrypted:     aws.BoolValue(v.Encrypted),
+		EncryptionKey: aws.StringValue(v.KmsKeyId),
+	}
+	return i
+}
```
randomvariable, line 949: Unnecessary conversion. Only need a volume ID to make changes

controllers/awsmachine_controller.go
```diff
@@ -911,3 +918,32 @@ func (r *AWSMachineReconciler) indexAWSMachineByInstanceID(o client.Object) []st
...
+
+func (r *AWSMachineReconciler) ensureStorageTags(ec2svc services.EC2MachineInterface, instanceID *string, machine *infrav1.AWSMachine) error {
+	volumes, err := ec2svc.GetEC2Volumes(instanceID)
+	if err != nil {
+		return err
+	}
+
+	for _, ebsVolume := range volumes {
+		volume := sdkToVolume(ebsVolume)
+		_, err = r.ensureVolumeTags(ec2svc, machine, volume, instanceID, machine.Spec.Tags)
```
randomvariable, line 930: We only need to pass in a volume ID. Don't need a full `infrav1.Volume` in the signature. 
We only want to apply the existing AdditionalTags in the API to the volume, not apply a different set of tags.

api/v1alpha4/awsmachine_types.go
```diff
@@ -147,6 +147,9 @@ type AWSMachineSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum:=default;dedicated;host
 	Tenancy string `json:"tenancy,omitempty"`
+
+	// Tags for tracking the additionalTags related to volume
+	Tags map[string]string `json:"tags,omitempty"`
```
randomvariable, line 152: Don't think we want an additional field here. We have an existing AdditionalTags API that does what we want.

controllers/awsmachine_tags.go
```diff
@@ -62,6 +70,35 @@ func (r *AWSMachineReconciler) ensureTags(svc service.EC2MachineInterface, machi
...
+	annotations := make(map[string]interface{})
+	annotations["Volumes"] = volume
+
+	// Check if the volume tags were changed. If they were, update them.
+	// It would be possible here to only send new/updated tags, but for the
+	// moment we send everything, even if only a single tag was created or
+	// updated.
+	changed, created, deleted, newAnnotation := r.tagsChanged(annotations, additionalTags)
+	if changed {
+		err := svc.UpdateResourceTags(instanceID, created, deleted)
```
randomvariable, line 87: We want to update the tags on the volume, not the EC2 instance, which is handled in the `ensureTags` function. The input to `UpdateResourceTags` should start with the volume ID

controllers/awsmachine_tags.go
```diff
@@ -62,6 +70,35 @@ func (r *AWSMachineReconciler) ensureTags(svc service.EC2MachineInterface, machi
 	return changed, nil
 }
 
+// Ensure that the tags of the volumes in the machine are correct
+// Returns bool, error
+// Bool indicates if changes were made or not, allowing the caller to decide
+// if the machine should be updated.
+func (r *AWSMachineReconciler) ensureVolumeTags(svc service.EC2MachineInterface, machine *infrav1.AWSMachine, volume *infrav1.Volume, instanceID *string, additionalTags map[string]string) (bool, error) {
+	annotations := make(map[string]interface{})
```
randomvariable, line 78: Updating the annotation in this function means the annotation will get overwritten with every EC2 volume.
Either, move the annotation up a level into the calling function, as we end up with
`map[string]map[string]interface{}` or, instead store a separate annotation per volume.

randomvariable (commented):
Thanks for this. Left some comments.

controllers/awsmachine_tags.go
Ankitasw: does the new changes look ok? I have moved annotation outside to the calling func now

controlplane/eks/api/v1alpha3/conversion.go
```diff
@@ -89,3 +89,13 @@ func Convert_v1alpha3_NetworkSpec_To_v1alpha4_NetworkSpec(in *infrav1alpha3.Netw
...
+
+// Convert_v1alpha3_Instance_To_v1alpha4_Instance is an autogenerated conversion function.
+func Convert_v1alpha3_Instance_To_v1alpha4_Instance(in *infrav1alpha3.Instance, out *infrav1alpha4.Instance, s apiconversion.Scope) error {
+	return infrav1alpha3.Convert_v1alpha3_Instance_To_v1alpha4_Instance(in, out, s)
+}
+
+// Convert_v1alpha4_Instance_To_v1alpha3_Instance is an autogenerated conversion function.
+func Convert_v1alpha4_Instance_To_v1alpha3_Instance(in *infrav1alpha4.Instance, out *infrav1alpha3.Instance, s apiconversion.Scope) error {
+	return infrav1alpha3.Convert_v1alpha4_Instance_To_v1alpha3_Instance(in, out, s)
+}
```
randomvariable, line 101: Remove these as they're no longer needed. Generated files should revert to their original.

api/v1alpha3/conversion.go
```diff
@@ -195,3 +195,8 @@ func Convert_v1alpha3_AWSClusterStaticIdentitySpec_To_v1alpha4_AWSClusterStaticI
 func Convert_v1alpha4_AWSClusterStaticIdentitySpec_To_v1alpha3_AWSClusterStaticIdentitySpec(in *v1alpha4.AWSClusterStaticIdentitySpec, out *AWSClusterStaticIdentitySpec, s apiconversion.Scope) error {
 	return autoConvert_v1alpha4_AWSClusterStaticIdentitySpec_To_v1alpha3_AWSClusterStaticIdentitySpec(in, out, s)
 }
+
+// Convert_v1alpha4_AWSMachineSpec_To_v1alpha3_AWSMachineSpec is an autogenerated conversion function.
+func Convert_v1alpha4_AWSMachineSpec_To_v1alpha3_AWSMachineSpec(in *v1alpha4.AWSMachineSpec, out *AWSMachineSpec, s apiconversion.Scope) error {
+	return autoConvert_v1alpha4_AWSMachineSpec_To_v1alpha3_AWSMachineSpec(in, out, s)
+}
```
randomvariable, line 202: Remove as there's no API changes to handle.

controllers/awsmachine_controller.go
```diff
@@ -911,3 +913,19 @@ func (r *AWSMachineReconciler) indexAWSMachineByInstanceID(o client.Object) []st
 
 	return nil
 }
+
+func (r *AWSMachineReconciler) ensureStorageTags(ec2svc services.EC2MachineInterface, instanceID *string, machine *infrav1.AWSMachine) {
+	volumes, err := ec2svc.GetEC2Volumes(instanceID)
```
sedefsavas, line 918: I think this is too much overhead in terms of api calls as we get the volumes first and then check if there is a change in the tags. So GetEC2Volumes is called at each reconcile.

Let's think about a way to avoid it.
We get the volumeId info from `v.BlockDeviceMappings[0].Ebs.VolumeId` in this method:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c53c107cbddb9e32e79455b1b7f96d46a7aafb08/pkg/cloud/services/ec2/instances.go#L761

We have annotations on the AWSMachine, we have additionaltags. If there is no change, we don't call GetEC2Volumes(). If there is change, we still don't call GetEC2Volumes(), just update the tags with UpdateResourceTags() using the volumeId and then update the annotation.

Does this sound reasonable?

Ankitasw: sounds fair, but again to get this v *ec2.instance, we need to call describeInstances which is again an API call, is my understanding correct?

randomvariable: We will have already have called describeInstances by the time we have got here, so the suggestion is to change the `ensureStorageTags` signature to pass in a infrav1.Instance .

@sedefsavas  is suggesting that if we know from the annotation that there is no change in tags to be done, we do not need to call GetEC2Volumes, which will minimise calls to the EC2 API.

Ankitasw: As per the comment --> If there is change, we still don't call GetEC2Volumes(), just update the tags with UpdateResourceTags() using the volumeId and then update the annotation ---> i understood from this statement that we shouldn't use getEC2Volumes at all, and make use of volume id from` v.BlockDeviceMappings[0].Ebs.VolumeId`

pkg/cloud/services/ec2/instances.go
```diff
@@ -794,6 +794,10 @@ func (s *Service) SDKToInstance(v *ec2.Instance) (*infrav1.Instance, error) {
 
 	i.AvailabilityZone = aws.StringValue(v.Placement.AvailabilityZone)
 
+	if len(v.BlockDeviceMappings) > 0 {
+		i.VolumeID = v.BlockDeviceMappings[0].Ebs.VolumeId
```
sedefsavas, line 798: We should do this for all the volumes as before and keep it in a slice.

sedefsavas (commented):
Changes look good, ,minor comment. Otherwise LGTM.

api/v1alpha4/types.go
```diff
@@ -672,6 +672,10 @@ type Instance struct {
 	// Tenancy indicates if instance should run on shared or single-tenant hardware.
 	// +optional
 	Tenancy string `json:"tenancy,omitempty"`
+
+	// ID of the instance's volume
+	// +optional
+	VolumeIDs []*string `json:"volumeIDs,omitempty"`
```
sedefsavas, line 678:
```suggestion
	VolumeIDs []string `json:"volumeIDs,omitempty"`
```
We are moving away from using pointer slices due to a conversion issue. 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2495

controllers/awsmachine_controller.go
```diff
@@ -912,3 +918,20 @@ func (r *AWSMachineReconciler) indexAWSMachineByInstanceID(o client.Object) []st
 
 	return nil
 }
+
+func (r *AWSMachineReconciler) ensureStorageTags(ec2svc services.EC2MachineInterface, instance *infrav1.Instance, machine *infrav1.AWSMachine) {
+	annotations := make(map[string]interface{})
+	for _, volumeID := range instance.VolumeIDs {
+		subAnnotations, err := r.ensureVolumeTags(ec2svc, aws.String(volumeID), annotations, machine.Spec.AdditionalTags)
```
sedefsavas, line 925: Sending empty annotations here?

controllers/awsmachine_controller.go
```diff
@@ -912,3 +918,36 @@ func (r *AWSMachineReconciler) indexAWSMachineByInstanceID(o client.Object) []st
 
 	return nil
 }
+
+func (r *AWSMachineReconciler) ensureStorageTags(ec2svc services.EC2MachineInterface, instance *infrav1.Instance, machine *infrav1.AWSMachine) {
+	annotations := make(map[string]interface{})
+	for _, volumeID := range instance.VolumeIDs {
+		if _, ok := annotations[volumeID]; ok {
```
sedefsavas, line 925: `annotations` is an empty map, this will always be false.

`machineAnnotation` and `additionalTags` needs to be compared here. See `ensureTags()` as an example.

Ankitasw: If i do that check outside if condition I am getting below error 
```
[manager] panic: interface conversion: interface {} is map[string]interface {}, not string
[manager] 
[manager] goroutine 673 [running]:
[manager] sigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).tagsChanged(0xc000200230, 0xc000b3f7d0, 0xc000cb3830, 0x40e278, 0x10, 0x212cd40, 0x1)
[manager] 	/Users/swamyan/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_tags.go:111 +0x5c8
[manager] sigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).ensureVolumeTags(0xc000200230, 0x2aa0490, 0xc000992630, 0xc0006291e0, 0xc000b3f7d0, 0xc000cb3830, 0x0, 0x1, 0x100000000000002)
[manager] 	/Users/swamyan/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_tags.go:76 +0x45
[manager] sigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).ensureStorageTags(0xc000200230, 0x2aa0490, 0xc000992630, 0xc0003ce3c0, 0xc00082e840)
[manager] 	/Users/swamyan/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_controller.go:932 +0x225
[manager] sigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).reconcileNormal(0xc000200230, 0x2a7e838, 0xc000cb37a0, 0xc0000bbf40, 0x2aa33c8, 0xc00068cd20, 0x2aa7278, 0xc00068cd20, 0x2aa4610, 0xc00068cd20, ...)
[manager] 	/Users/swamyan/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_controller.go:589 +0xe25
[manager] sigs.k8s.io/cluster-api-provider-aws/controllers.(*AWSMachineReconciler).Reconcile(0xc000200230, 0x2a7e838, 0xc000cb37a0, 0xc000b996a9, 0x7, 0xc0004f3ec0, 0x19, 0xc000cb3700, 0x0, 0x0, ...)
[manager] 	/Users/swamyan/go/src/sigs.k8s.io/cluster-api-provider-aws/controllers/awsmachine_controller.go:201 +0x8b4
[manager] sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).reconcileHandler(0xc0007be280, 0x2a7e790, 0xc000b311c0, 0x239b1c0, 0xc0000e0e00)
[manager] 	/Users/swamyan/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0/pkg/internal/controller/controller.go:298 +0x30d
[manager] sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).processNextWorkItem(0xc0007be280, 0x2a7e790, 0xc000b311c0, 0x5d4a00)
[manager] 	/Users/swamyan/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0/pkg/internal/controller/controller.go:253 +0x205
[manager] sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2.2(0xc000690070, 0xc0007be280, 0x2a7e790, 0xc000b311c0)
[manager] 	/Users/swamyan/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0/pkg/internal/controller/controller.go:214 +0x6b
[manager] created by sigs.k8s.io/controller-runtime/pkg/internal/controller.(*Controller).Start.func2
[manager] 	/Users/swamyan/go/pkg/mod/sigs.k8s.io/controller-runtime@v0.9.0/pkg/internal/controller/controller.go:210 +0x425
[manager] Exiting with code 2
```

Ankitasw: jsonAnnotation in `machineAnnotationJSON` func is giving the value as null in map , "{\"vol-0e5e5b65521f6e259\":{\"foo\":\"bar\",\"rootDeviceID\":\"volume-id\",\"rootDeviceSize\":\"40\"}}"=null

randomvariable:
Looks in a good state to me.

/lgtm

/assign @sedefsavas  for approval;

sedefsavas:
Thanks for the changes @Ankitasw 
/approve

PR #2505, opened 2021-06-14
Update EKS clusters to enable encryption

### LINKED ISSUE #2452: Update EKS clusters to enable encryption
/kind feature
/area provider/eks
/priority important-soon
/help
/milestone v0.7.0

You can enable EKS encryption at cluster creation time by using `encryptionConfig`. However, if you initially create a cluster without encryption and then later want to enable encryption this isn't currently supported by CAPA. 

We need to detect the scenario where `AWSManagedControlPlane` has been updated an `encryptionConfig` has gone from nil ro being specified. And then call [AssociateEncryptionConfig](https://docs.aws.amazon.com/sdk-for-go/api/service/eks/#EKS.AssociateEncryptionConfig)

Also, once enabled you are not allowed to change your key. We also need to update the validating webhook to stop changing the key.

**Environment:**

- Cluster-api-provider-aws version: 0.6.6

Ankitasw:
**What type of PR is this?**
/kind feature

Enables to add encryption config in update path if it was nil at the time of cluster creation. It also disallows any modifications in further update if the config was already defined at the time of creation

Fixes #2452

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
* Add encryption-config in update if missed in cluster creation
* Add webhook validation checks for encryption-config
Action required
Controllers policy updated with missing IAM permissions required to enable EKS encryption, if you are planning to enable EKS encryption then you will need to update your controllers policy by running `clusterawsadm bootstrap iam create-cloudformation-stack` again.
```

### DIFF
--- a/controlplane/eks/api/v1alpha4/awsmanagedcontrolplane_webhook.go
+++ b/controlplane/eks/api/v1alpha4/awsmanagedcontrolplane_webhook.go
@@ -131,6 +131,20 @@ func (r *AWSManagedControlPlane) ValidateUpdate(old runtime.Object) error {
 		)
 	}
 
+	// If encryptionConfig is already set, do not allow removal of it.
+	if oldAWSManagedControlplane.Spec.EncryptionConfig != nil && r.Spec.EncryptionConfig == nil {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "encryptionConfig"), r.Spec.EncryptionConfig, "disabling EKS encryption is not allowed after it has been enabled"),
+		)
+	}
+
+	// If encryptionConfig is already set, do not allow change in provider
+	if r.Spec.EncryptionConfig != nil && oldAWSManagedControlplane.Spec.EncryptionConfig != nil && *r.Spec.EncryptionConfig.Provider != *oldAWSManagedControlplane.Spec.EncryptionConfig.Provider {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "encryptionConfig", "provider"), r.Spec.EncryptionConfig.Provider, "changing EKS encryption is not allowed after it has been enabled"),
+		)
+	}
+
 	// If a identityRef is already set, do not allow removal of it.
 	if oldAWSManagedControlplane.Spec.IdentityRef != nil && r.Spec.IdentityRef == nil {
 		allErrs = append(allErrs,
--- a/pkg/cloud/services/eks/cluster.go
+++ b/pkg/cloud/services/eks/cluster.go
@@ -20,8 +20,9 @@ import (
 	"context"
 	"fmt"
 	"net"
-	"reflect"
 	"time"
+
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/internal/cmp"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/awserr"
@@ -219,15 +220,25 @@ func (s *Service) deleteClusterAndWait(cluster *eks.Cluster) error {
 }
 
 func makeEksEncryptionConfigs(encryptionConfig *controlplanev1.EncryptionConfig) []*eks.EncryptionConfig {
+	cfg := []*eks.EncryptionConfig{}
+
 	if encryptionConfig == nil {
-		return []*eks.EncryptionConfig{}
+		return cfg
 	}
-	return []*eks.EncryptionConfig{{
+	//TODO: change EncryptionConfig so that provider and resources are required  if encruptionConfig is specified
+	if encryptionConfig.Provider == nil || len(*encryptionConfig.Provider) == 0 {
+		return cfg
+	}
+	if len(encryptionConfig.Resources) == 0 {
+		return cfg
+	}
+
+	return append(cfg, &eks.EncryptionConfig{
 		Provider: &eks.Provider{
 			KeyArn: encryptionConfig.Provider,
 		},
 		Resources: encryptionConfig.Resources,
-	}}
+	})
 }
 
 func makeVpcConfig(subnets infrav1.Subnets, endpointAccess controlplanev1.EndpointAccess, securityGroups map[infrav1.SecurityGroupRole]infrav1.SecurityGroup) (*eks.VpcConfigRequest, error) {
@@ -481,16 +492,23 @@ func (s *Service) reconcileVpcConfig(vpcConfig *eks.VpcConfigResponse) (*eks.Vpc
 	return nil, nil
 }
 
-func (s *Service) reconcileEKSEncryptionConfig(encryptionConfig []*eks.EncryptionConfig) error {
+func (s *Service) reconcileEKSEncryptionConfig(currentClusterConfig []*eks.EncryptionConfig) error {
+	s.Info("reconciling encryption configuration")
+	if currentClusterConfig == nil {
+		currentClusterConfig = []*eks.EncryptionConfig{}
+	}
+
 	encryptionConfigs := s.scope.ControlPlane.Spec.EncryptionConfig
 	updatedEncryptionConfigs := makeEksEncryptionConfigs(encryptionConfigs)
 
 	switch {
-	case reflect.DeepEqual(encryptionConfig, updatedEncryptionConfigs):
+	case compareEncryptionConfig(currentClusterConfig, updatedEncryptionConfigs):
+		s.V(2).Info("encryption configuration unchanged, no action")
 		return nil
-	case encryptionConfig == nil || len(encryptionConfig) == 0 && updatedEncryptionConfigs != nil:
+	case len(currentClusterConfig) == 0 && len(updatedEncryptionConfigs) > 0:
+		s.V(2).Info("enabling encryption for eks cluster", "cluster", s.scope.KubernetesClusterName())
 		if err := s.updateEncryptionConfig(updatedEncryptionConfigs); err != nil {
-			record.Warnf(s.scope.ControlPlane, "FailedUpdateEKSControlPlane", "failed to update the EKS control plane: %v", err)
+			record.Warnf(s.scope.ControlPlane, "FailedUpdateEKSControlPlane", "failed to update the EKS control plane encryption configuration: %v", err)
 			return errors.Wrapf(err, "failed to update EKS cluster")
 		}
 	default:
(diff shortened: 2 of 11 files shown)


controlplane/eks/api/v1alpha4/awsmanagedcontrolplane_webhook.go
```diff
@@ -131,6 +131,13 @@ func (r *AWSManagedControlPlane) ValidateUpdate(old runtime.Object) error {
 		)
 	}
 
+	// If encryptionConfig is already set, do not allow removal of it.
+	if oldAWSManagedControlplane.Spec.EncryptionConfig != nil && r.Spec.EncryptionConfig == nil {
+		allErrs = append(allErrs,
+			field.Invalid(field.NewPath("spec", "encryptionConfig"), r.Spec.EncryptionConfig, "field cannot be set to nil"),
```
richardcase, line 137: It may be good to give a more user friendly message. Something like "disabling EKS encryption is not allowed after it has been enabled".

pkg/cloud/services/eks/cluster.go
```diff
@@ -114,6 +114,10 @@ func (s *Service) reconcileCluster(ctx context.Context) error {
 		return errors.Wrap(err, "failed reconciling cluster config")
 	}
 
+	if err := s.reconcileEKSEncryptionConfig(cluster.EncryptionConfig); err != nil {
```
richardcase, line 117: We will need some more logic around this call.  For example, if i am creating a new cluster then the encryption will have been enabled in `createCluster` and then it will update the encryption config again in `reconcileEKSEncryptionConfig`. 

What do you think to this:
1. We have the existing configuration  from AWS [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/dad36a88f506a7dd05ba19a5e5ff25bb69bdf9a0/pkg/cloud/services/eks/cluster.go#L51)
2.  The existing encryption config is passed into `reconcileEKSEncryptionConfig` (as you are doing now)
3. The desired encryption configuration from the AWSManagedControlPlane is coming from [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/dad36a88f506a7dd05ba19a5e5ff25bb69bdf9a0/pkg/cloud/services/eks/cluster.go#L483)
4. So we have the actual and desired config at this point
5.  We add some logic to `reconcileEKSEncryptionConfig` based on actual/desired that determines if there is a difference and what action needs to be taken (i.e. no action, AssociateEncryptionConfig or raise error)

What do you think?

Ankitasw: I think it sounds fair, i will take into consideration about updating the encryption config again (first in createCluster, then in reconcileEKSEncryptionConfig)

Ankitasw: in what scenario, we can have error in `reconcileEKSEncryptionConfig`

richardcase: An error shouldn't happen with the validating webhook. However, the Webhooks could be disabled and it's probably best to check for the incorrect update scenario to be safe.

pkg/cloud/services/eks/cluster.go
```diff
@@ -550,6 +573,29 @@ func (s *Service) describeEKSCluster(eksClusterName string) (*eks.Cluster, error
 	return out.Cluster, nil
 }
 
+func (s *Service) updateEncryptionConfig(updatedEncryptionConfigs []*eks.EncryptionConfig) error {
+	input := &eks.AssociateEncryptionConfigInput{
+		ClusterName:      &s.scope.Cluster.ClusterName,
```
richardcase, line 578: You will need to use `s.scope.KubernetesClusterName()` instead here just in case someone has given an explicit EKS cluster name in the AWSManagedControlPlane.

richardcase (changes requested):
This is looking good @Ankitasw :+1:  

A few minor comments.

pkg/cloud/services/eks/cluster.go
```diff
@@ -475,6 +480,24 @@ func (s *Service) reconcileVpcConfig(vpcConfig *eks.VpcConfigResponse) (*eks.Vpc
...
+
+	if reflect.DeepEqual(encryptionConfig, updatedEncryptionConfigs) {
+		return nil
+	} else if encryptionConfig == nil || len(encryptionConfig) == 0 && updatedEncryptionConfigs != nil {
+		if err := s.updateEncryptionConfig(updatedEncryptionConfigs); err != nil {
+			record.Warnf(s.scope.ControlPlane, "FailedUpdateEKSControlPlane", "failed to update the EKS control plane: %v", err)
+			return errors.Wrapf(err, "failed to update EKS cluster")
+		}
+	} else {
+		record.Warnf(s.scope.ControlPlane, "FailedUpdateEKSControlPlane", "failed to update the EKS control plane: disabling EKS encryption is not allowed after it has been enabled")
```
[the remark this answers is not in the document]
Ankitasw: yes that's correct

Ankitasw:
@richardcase : Gentle reminder to review the change.

pkg/cloud/services/eks/cluster.go
```diff
@@ -550,6 +573,29 @@ func (s *Service) describeEKSCluster(eksClusterName string) (*eks.Cluster, error
...
+	}
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+		if _, err := s.EKSClient.AssociateEncryptionConfig(input); err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				return false, aerr
+			}
+			return false, err
+		}
+
+		conditions.MarkTrue(s.scope.ControlPlane, eks.UpdateTypeAssociateEncryptionConfig)
```
richardcase, line 589: I like including a condition. But this will require some additional code to ensure we mark false as well. When we call `AssociateEncryptionConfig` what status will the EKS cluster change to?

I'm guessing it migt change to: `ClusterStatusUpdating = "UPDATING"`

As an example of how we use existing conditions [EKSControlPlaneUpdatingCondition](https://github.com/kubernetes-sigs/cluster-api-provider-aws/search?q=EKSControlPlaneUpdatingCondition) is good to look at.

Ankitasw: I am not sure what sttaus will EKSCluster change to, in what case we should mark false?

Ankitasw: do we want to add something like [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/e55d3ed297b9608aa229cf96caa658e64ef6373d/pkg/cloud/services/eks/cluster.go)

richardcase: I'm thinking that we may need to change the condition to `EKSControlPlaneUpdatingCondition` as well to indicate that an update is in progress.....assuming that the `AssociateEncryptionConfig` call results in the status of the cluster updating.

richardcase: I'll have a quick look at the status now....

richardcase: The status of the cluster does change to to "UPDATING":
image

Ankitasw:
@richardcase : not sure why the tests are failing, by looking at the report, I couldn't understand the root cause, could you help me understand the issue

richardcase:
> @richardcase : not sure why the tests are failing, by looking at the report, I couldn't understand the root cause, could you help me understand the issue

I'll take a look at the logs.

richardcase:
There are a lot of errors in the logs which are unrelated to this change (i think). I need to investigate

richardcase:
I just ran the EKS e2e test on #2530 and they passed. I will try again it case its a flake:

richardcase:
Investigating the e2e failures

richardcase:
@Ankitasw - i found a couple of issues and made the following changes:

1. Added missing IAM permission for the AssociateEncryptionConfig call to the controllers policy. This means that people who want to use this feature will need to update their bootstrap stack using clusterawsadm. I would add this in the release notes section in the PR description ^^^^^^
2. Changed the logic for the enabling to check the length of the config in the spec is greater than zero instead of checking not null
3. Added some additional logging.

Feel free to ping me on slack if you want to discuss.

richardcase:
There are numerous errors with the e2e tests: 

https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2505/pull-cluster-api-provider-aws-e2e-eks/1408071070308634624/artifacts/clusters/bootstrap/controllers/capa-eks-control-plane-controller-manager/capa-eks-control-plane-controller-manager-79dfc4848d-f7lj6/manager.log

pkg/internal/cmp/slice.go
```diff
@@ -0,0 +1,32 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package cmp
+
+import "k8s.io/utils/pointer"
+
+func Equals(slice1, slice2 []*string) bool {
```
richardcase, line 21: Would we consider 2 string slices to be equal if they had the same elements but in a different order?

For example:
- "test1","test2"
- "test2", "test1"

Ankitasw: I was hoping to add the sort func for both the slices, but since the slices are []*string we dont have inbuilt sort func, the only way I have in mind is to convert []*string to []string and then sort and then do the above logic, do you have any other alternative in mind?

Ankitasw: added customized sorting

richardcase:
Tests are passing (including e2e), so:

/lgtm

@randomvariable @sedefsavas - would you be able to take a look and do the final approval ?

Ankitasw:
> New changes are detected. LGTM label has been removed.

Rebased the branch

richardcase:
@Ankitasw - thanks for rebasing. After the e2e passes again i will take al ook

randomvariable:
/approve

/assign @richardcase  
for lgtm

richardcase:
The e2e tests have [passed](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2505/pull-cluster-api-provider-aws-e2e-eks/1411975745751748608) and the change is looking good, so:

/lgtm

PR #2525, opened 2021-06-19
:bug: AWSMachinePool: Prune old Launch Template versions

### LINKED ISSUE #2368: AWSMachinePool controller should clean up unused Launch Template Versions
/kind bug

The AWSMachinePool controller creates new Launch Template Versions when the configuration changes, or the MachinePool is scaled. (See #2354). AWS imposes a limit on the number of Launch Template Versions:

> You are limited to creating 5,000 launch templates per Region and 10,000 versions per launch template.
> -- https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-launch-templates.html#launch-template-restrictions

When this limit is reached, further updates are not possible, and therefore further configuration changes or scale operations would be blocked.

dlipovetsky:
**What type of PR is this?**

/kind bug

Remove one old version before creating a new version.

Fixes #2368

I'm not sure what sort of unit test we want to see.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AWSMachinePool controller removes one old LaunchTemplate version before creating a new version, preventing the number of versions from growing without bound, and reaching the maximum limit. 
```

### DIFF
--- a/exp/controllers/awsmachinepool_controller.go
+++ b/exp/controllers/awsmachinepool_controller.go
@@ -456,7 +456,7 @@ func (r *AWSMachinePoolReconciler) reconcileLaunchTemplate(machinePoolScope *sco
 	if needsUpdate || tagsChanged || *imageID != *launchTemplate.AMI.ID || launchTemplateUserDataHash != bootstrapDataHash {
 		machinePoolScope.Info("creating new version for launch template", "existing", launchTemplate, "incoming", machinePoolScope.AWSMachinePool.Spec.AWSLaunchTemplate)
 		// There is a limit to the number of Launch Template Versions.
-		// We ensure that the number of versions does not grow without bound by following a simple rule: For each version we create, we delete one old version.
+		// We ensure that the number of versions does not grow without bound by following a simple rule: Before we create a new version, we delete one old version, if there is at least one old version that is not in use.
 		if err := ec2svc.PruneLaunchTemplateVersions(machinePoolScope.AWSMachinePool.Status.LaunchTemplateID); err != nil {
 			return err
 		}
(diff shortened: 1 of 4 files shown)


dlipovetsky:
Once this is approved, I'd like to backport to `release-0.6`

randomvariable:
> Once this is approved, I'd like to backport to release-0.6

sgtm

exp/controllers/awsmachinepool_controller.go
```diff
@@ -455,6 +455,11 @@ func (r *AWSMachinePoolReconciler) reconcileLaunchTemplate(machinePoolScope *sco
 	// userdata, OR we've discovered a new AMI ID.
 	if needsUpdate || tagsChanged || *imageID != *launchTemplate.AMI.ID || launchTemplateUserDataHash != bootstrapDataHash {
 		machinePoolScope.Info("creating new version for launch template", "existing", launchTemplate, "incoming", machinePoolScope.AWSMachinePool.Spec.AWSLaunchTemplate)
+		// There is a limit to the number of Launch Template Versions.
+		// We ensure that the number of versions does not grow without bound by following a simple rule: For each version we create, we delete one old version.
+		if err := ec2svc.PruneLaunchTemplateVersions(machinePoolScope.AWSMachinePool.Status.LaunchTemplateID); err != nil {
```
sedefsavas, line 460: Is it better to create one before pruning? In case creation fails we don't want to delete the previous one.
We create a new one, it is tagged as latest, so the previous can be deleted.

`CreateLaunchTemplateVersion` returns the version created, how about directly trying to delete the previous version? Assuming the numbering is strictly increasing.

dlipovetsky: > Is it better to create one before pruning?

If we create one before removing one, then we may end up never removing any. Whether or not we return an error when we fail to remove one, we'll create more than we'll remove, and therefore eventually the limit.

> In case creation fails we don't want to delete the previous one.
> We create a new one, it is tagged as latest, so the previous can be deleted.

I took care to never remove the version that is currently "in use." Please see the `PruneLaunchTemplateVersions` godoc, and let me know if anything is unclear. Thanks!

dlipovetsky: I've revised the comment at this line. I hope it's more clear now.

dlipovetsky:
Thanks for reviewing, @sedefsavas!

Issue #2546, opened 2021-06-25
[e2e] Route Table error when running EKS tests


richardcase:
/kind bug
/kind failing-test
/area testing
/help
/priority important-soon
/milestone v0.7.x

**What steps did you take and what happened:**
When running the EKS e2e tests, i see the following error message in the control plane manager logs:

```
E0625 07:42:28.399270       1 controller.go:302] controller-runtime/manager/controller/awsmanagedcontrolplane "msg"="Reconciler error" "error"="failed to reconcile network for AWSManagedControlPlane eks-nodes-6l160t/cluster-6vem15-control-plane: failed to associate route table \"rtb-0f2c0010004a7eac4\" to subnet \"\": MissingParameter: The request must contain the parameter SubnetId or GatewayId\n\tstatus code: 400, request id: 931867e5-04a0-4708-9ccd-dadc3b0239d5" "name"="cluster-6vem15-control-plane" "namespace"="eks-nodes-6l160t" "reconciler group"="controlplane.cluster.x-k8s.io" "reconciler kind"="AWSManagedControlPlane" 
```

And then when I look in the AWS console, i see an additional route table created for the ^^^^ above ID that isn't associated with any subnets:

![image](https://user-images.githubusercontent.com/198425/123391455-f73fe280-d593-11eb-83c8-6e08ddddebdc.png)

In this example, we end up with a duplicate route table for eu-north-1a

**What did you expect to happen:**
I expect there to be no error and for there to be no unused/un-needed route tables

**Environment:**

- Cluster-api-provider-aws version: 0.6.6

PR #2549, opened 2021-06-26
Enabled GPU optimised AMIs for EKS

### LINKED ISSUE #2275: Enable usage of GPU optimized AMIs for EKS
/kind feature
/area provider/eks
/help

When creating an EKS cluster i would like to indicate that the GPU optimized AMI be used. Currently we use `/aws/service/eks/optimized-ami/%s/amazon-linux-2/recommended/image_id` to lookup the [AMI in parameter store](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/master/pkg/cloud/services/ec2/ami.go#L232).

To lookup the GPU accelerated AMIs we need to use `/aws/service/eks/optimized-ami/%s/amazon-linux-2-gpu/recommended/image_id`

It would be good if there was an option to say u want this machinepool to use the GPU accelerated AMI and then look this up from SSM.

> Is the accelerated AMI even needed? And what about https://github.com/NVIDIA/gpu-operator

It is possible with the `AWSManagedMachinePool` using `amiType: "AL2_x86_64_GPU"` but not with AWSMachinePool or AWSMachine

**Environment:**

shivi28:
**What type of PR is this?**
/kind feature

Enable usage of GPU optimized AMIs for EKS.
Removed unused fields like ARN and filters from AMIReference.
Fixes #2275

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Enable usage of GPU optimized AMIs for EKS
Removed unused fields like ARN and Filters from AMIReference
```


api/v1alpha4/types.go
```diff
@@ -724,3 +724,13 @@ type SpotMarketOptions struct {
 	// +kubebuilder:validation:pattern="^[0-9]+(\.[0-9]+)?$"
 	MaxPrice *string `json:"maxPrice,omitempty"`
 }
+
+// MachineAMIType specifies which AWS AMI to use for a AWSMachine and AWSMachinePool.
+type MachineAMIType string
+
+const (
+	// Al2x86_64 is the default AMI type.
+	Al2x86_64 MachineAMIType = "AL2_x86_64"
```
richardcase, line 733: I wonder if we should have names that are a bit more user-friendly. Perhaps we could have `AmazonLinux` and `AmazonLinuxGPU` or something similar.  It would be better UX, and also we have an open ticket to support Bottlerocket as well.

exp/api/v1alpha4/types.go
```diff
@@ -112,6 +112,12 @@ type AWSLaunchTemplate struct {
 	// at the cluster level or in the actuator.
 	// +optional
 	AdditionalSecurityGroups []infrav1.AWSResourceReference `json:"additionalSecurityGroups,omitempty"`
+
+	// AMIType defines the AMI type
+	// +kubebuilder:validation:Enum:=AL2_x86_64;AL2_x86_64_GPU
+	// +kubebuilder:default:=AL2_x86_64
+	// +optional
+	AMIType *infrav1.MachineAMIType `json:"amiType,omitempty"`
```
richardcase, line 120: Probably worth adding a note this is only applicable to EKS

api/v1alpha4/awsmachine_types.go
```diff
@@ -147,6 +147,12 @@ type AWSMachineSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum:=default;dedicated;host
 	Tenancy string `json:"tenancy,omitempty"`
+
+	// AMIType defines the AMI type
+	// +kubebuilder:validation:Enum:=AL2_x86_64;AL2_x86_64_GPU
+	// +kubebuilder:default:=AL2_x86_64
+	// +optional
+	AMIType *MachineAMIType `json:"amiType,omitempty"`
```
richardcase, line 155: Worth adding a comment that this i sonly applicable for EKS

exp/api/v1alpha4/types.go
shivi28: added

api/v1alpha4/types.go
shivi28: done

api/v1alpha4/awsmachine_types.go
```diff
@@ -147,6 +147,12 @@ type AWSMachineSpec struct {
 	// +optional
 	// +kubebuilder:validation:Enum:=default;dedicated;host
 	Tenancy string `json:"tenancy,omitempty"`
+
+	// AMIType defines the AMI type. AmazonLinuxGPU type is only available for EKS
+	// +kubebuilder:validation:Enum:=AmazonLinux;AmazonLinuxGPU
+	// +kubebuilder:default:=AmazonLinux
+	// +optional
+	AMIType *MachineAMIType `json:"amiType,omitempty"`
```
randomvariable, line 155: Not sure about adding `AMIType` when we already have an `AMI` field. It makes for a confusing UX.

I would rather change `AMI` from `AWSResourceReference` to a new `AMIReference` and add the EKS option there. Note that since we support Ubuntu, Amazon Linux, CentOS, and we will support Flatcar soon, having a field that only allows either AmazonLinux or AmazonLinuxGPU is also confusing.

If we have the customise AMIReference type, we get something like:

```
// AMIReference is a reference to a specific AWS resource by ID, ARN, or filters.
// Only one of ID, ARN or Filters may be specified. Specifying more than one will result in
// a validation error.
type AMIReference struct {
	// ID of resource
	// +optional
	ID *string `json:"id,omitempty"`

	// ARN of resource
	// +optional
	ARN *string `json:"arn,omitempty"`

	// Filters is a set of key/value pairs used to identify a resource
	// They are applied according to the rules defined by the AWS API:
	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Filtering.html
	// +optional
	Filters []Filter `json:"filters,omitempty"`

        // EKSOptimizedLookupType If specified, will look up an EKS Optimized image in SSM Parameter store
        // +kubebuilder:validation:Enum:=AmazonLinux;AmazonLinuxGPU
        EKSOptimizedLookupType *EKSAMILookupType `json:"eksLookupType,omitempty"`
}
```

Note, there should be no default here, as we default to not using the optimised image type via nil.

randomvariable (changes requested):
Please review the API here as it causes confusion for non-optimised use case

api/v1alpha4/awsmachine_types.go
richardcase: I like this idea of `AMIReference` as its more explicit and easier to understand. So a 

+1

shivi28: `EKSOptimizedLookupType` field should be optional, right? If user is not providing any `EKSOptimizedLookupType` then it should use AmazonLinux optimised Non GPU AMIs from path `/aws/service/eks/optimized-ami/%s/amazon-linux-2/recommended/image_id` exactly like the current behaviour

richardcase: For EKS, yes it should.

randomvariable: Good catch, thanks!

randomvariable:
Happy with this from an API perspective.
/assign @richardcase

richardcase:
@shivi28 - i think the eks e2e test failures are being caused by #2546 and not anything to do with your code. I am going to look at 2546 today. (FYI @randomvariable )

randomvariable:
@shivi28 having a look at #2580, it probably makes sense to delete the unused fields in AMIResourceReference.

shivi28:
Sure @randomvariable I will remove it

randomvariable:
Can you rebase this on top of main and rerun tests. Should be sorted now.

richardcase:
The EKS e2e have [passed](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2549/pull-cluster-api-provider-aws-e2e-eks/1417048617188855808), so:

/lgtm

PR #2556, opened 2021-06-30
Changed Volume encypted to pointer for v1alpha4

### LINKED ISSUE #2233: For v1alpha4, encrypted value of Volume struct should be pointer
/kind feature

If you're using an encrypted EBS snapshot for the root volume AMI, you explicitly need to set

``` yaml
rootVolume:
  encrypted: true
```

You shouldn't need to do this.

**Environment:**

shivi28:
**What type of PR is this?**
/kind feature

Change v1alpha4 `Volume. Encrypted` field to `*bool` from `bool`

Fixes #2233 

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
No longer mandatory to set encryption value to True for root volumes when using encrypted AMIs
```


sedefsavas (commented):
One thing to clarify from the issue:

> If you're using an encrypted EBS snapshot for the root volume AMI, you shouldn't need to set encrypted to true.

We will still need KMS key for encrypted AMIs. 
@randomvariable Can you clarify how we won't need the encrypted field is set?

api/v1alpha4/types.go
```diff
@@ -706,7 +706,8 @@ type Volume struct {
 
 	// Encrypted is whether the volume should be encrypted or not.
 	// +optional
-	Encrypted bool `json:"encrypted,omitempty"`
+	// +kubebuilder:default=true
```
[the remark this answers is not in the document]
randomvariable: Yes, that's correct. We don't want this on by default.

shivi28:
@randomvariable ,  if **encrypted = nil** in v1alpha4 then we are converting it to **encrypted = false** in v1alpha3.
But my concern is:
1. If **AMI is encrypted** in v1alpha4, then it will work. But in v1alpha3 it might not work because if `encrypted=nil`  and `AMI is encrypted` in v1alpha4, then it will get converted to `encrypted = false` in `v1alpha3` which might fail on the AWS side (since AMI is encrypted and value is false). Is this behavior acceptable?
2. Even if we convert the encrypted nil value in v1alpha4  to true in v1alpha3, then the same problem might occur for non-encrypted AMIs.
Please correct if understanding is wrong

api/v1alpha4/awsmachine_webhook.go
```diff
@@ -145,6 +146,11 @@ func (r *AWSMachine) validateRootVolume() field.ErrorList {
 		return allErrs
 	}
 
+	// be default rootVolume encrypted should be true
+	if r.Spec.RootVolume != nil && r.Spec.RootVolume.Encrypted == nil {
+		r.Spec.RootVolume.Encrypted = aws.Bool(true)
```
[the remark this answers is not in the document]
shivi28: sure, default is removed

api/v1alpha4/types.go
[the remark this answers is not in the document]
shivi28: agree

sedefsavas:
@shivi28 

>  if encrypted = nil in v1alpha4 then we are converting it to encrypted = false in v1alpha3.
But my concern is: 
If AMI is encrypted in v1alpha4, then it will work. 

Above sentence should not be accurate: When encrypted is nil, we don't default it to true, hence it becomes false in v1alpha4 too.


Currently, if the AMI is encrypted, user still needs to set the encrypted value to true. For not requiring encrypted field to be true when using an encrypted snapshot:
We can try to find out if the AMI the instance is created from encrypted or not. `Volume` struct has that field. If the instance is encrypted, then continue as if the encrypted field is set to true.

This is how an encrypted AMI looks from the GUI
`Block Devices
/dev/sda1=snap-03aaaa7cd1e181cf:8:true:gp2:encrypted`

api/v1alpha4/types.go
randomvariable, line 709:
```suggestion
	// +kubebuilder:default=true
```

We still don't want any defaulting here. Remove this.

api/v1alpha3/conversion_test.go
```diff
@@ -42,6 +43,78 @@ func AWSClusterStaticIdentityFuzzer(obj *AWSClusterStaticIdentity, c fuzz.Contin
...
+	}
+}
+
+func AWSClusterFuzzer(obj *v1alpha4.AWSCluster, c fuzz.Continue) {
+	c.FuzzNoCustom(obj)
+
+	if obj.Status.Bastion == nil {
+		obj.Status.Bastion = &v1alpha4.Instance{}
+	}
+	obj.Status.Bastion.RootVolume = &v1alpha4.Volume{Encrypted: aws.Bool(false)}
```
randomvariable, line 58:
```suggestion
	obj.Status.Bastion.RootVolume = &v1alpha4.Volume{Encrypted: pointer.BoolPtr(false)}
```

Only use aws.Bool when working with AWS services.

api/v1alpha3/conversion_test.go
```diff
@@ -42,6 +43,78 @@ func AWSClusterStaticIdentityFuzzer(obj *AWSClusterStaticIdentity, c fuzz.Contin
...
+
+	if obj.Status.Bastion.NonRootVolumes == nil || len(obj.Status.Bastion.NonRootVolumes) == 0{
+		obj.Status.Bastion.NonRootVolumes = []v1alpha4.Volume{{Encrypted: aws.Bool(false)}}
```
randomvariable, line 61:
```suggestion
		obj.Status.Bastion.NonRootVolumes = []v1alpha4.Volume{{Encrypted: pointer.BoolPtr(false)}}
```

api/v1alpha3/conversion_test.go
```diff
@@ -42,6 +43,78 @@ func AWSClusterStaticIdentityFuzzer(obj *AWSClusterStaticIdentity, c fuzz.Contin
...
+func AWSMachineFuzzer(obj *v1alpha4.AWSMachine, c fuzz.Continue) {
+	c.FuzzNoCustom(obj)
+
+	if obj.Spec.RootVolume == nil {
+		obj.Spec.RootVolume = &v1alpha4.Volume{}
+	}
+	obj.Spec.RootVolume.Encrypted = aws.Bool(false)
+
+	if obj.Spec.NonRootVolumes == nil || len(obj.Spec.NonRootVolumes) == 0 {
+		obj.Spec.NonRootVolumes = []v1alpha4.Volume{{Encrypted: aws.Bool(false)}}
```
randomvariable, line 85:
```suggestion
		obj.Spec.NonRootVolumes = []v1alpha4.Volume{{Encrypted: pointer.BoolPtr(false)}}
```

api/v1alpha3/conversion_test.go
```diff
@@ -42,6 +43,78 @@ func AWSClusterStaticIdentityFuzzer(obj *AWSClusterStaticIdentity, c fuzz.Contin
...
+	}
+
+	for i, nonRootVolume := range obj.Spec.NonRootVolumes {
+		nonRootVolume.Encrypted = aws.Bool(false)
```
randomvariable, line 89:
```suggestion
		nonRootVolume.Encrypted = pointer.BoolPtr(false)
```

api/v1alpha3/conversion_test.go
```diff
@@ -42,6 +43,78 @@ func AWSClusterStaticIdentityFuzzer(obj *AWSClusterStaticIdentity, c fuzz.Contin
...
+	}
+}
+
+func AWSMachineTemplateFuzzer(obj *v1alpha4.AWSMachineTemplate, c fuzz.Continue) {
+	c.FuzzNoCustom(obj)
+
+	if obj.Spec.Template.Spec.RootVolume == nil {
+		obj.Spec.Template.Spec.RootVolume = &v1alpha4.Volume{}
+	}
+	obj.Spec.Template.Spec.RootVolume.Encrypted = aws.Bool(false)
```
randomvariable, line 106:
```suggestion
	obj.Spec.Template.Spec.RootVolume.Encrypted = pointer.BoolPtr(false)
```

api/v1alpha3/conversion_test.go
```diff
@@ -42,6 +43,78 @@ func AWSClusterStaticIdentityFuzzer(obj *AWSClusterStaticIdentity, c fuzz.Contin
...
+
+	if obj.Spec.Template.Spec.NonRootVolumes == nil || len(obj.Spec.Template.Spec.NonRootVolumes) == 0 {
+		obj.Spec.Template.Spec.NonRootVolumes = []v1alpha4.Volume{{Encrypted: aws.Bool(false)}}
```
randomvariable, line 109:
```suggestion
		obj.Spec.Template.Spec.NonRootVolumes = []v1alpha4.Volume{{Encrypted: pointer.BoolPtr(false)}}
```

api/v1alpha3/conversion_test.go
```diff
@@ -42,6 +43,78 @@ func AWSClusterStaticIdentityFuzzer(obj *AWSClusterStaticIdentity, c fuzz.Contin
...
+	}
+
+	for i, nonRootVolume := range obj.Spec.Template.Spec.NonRootVolumes {
+		nonRootVolume.Encrypted = aws.Bool(false)
```
randomvariable, line 113:
```suggestion
		nonRootVolume.Encrypted = pointer.BoolPtr(false)
```

config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
```diff
@@ -1144,6 +1144,7 @@ spec:
                           description: Device name
                           type: string
                         encrypted:
+                          default: true
```
randomvariable, line 1147: All this defaulting needs to go by removing the kubebuilder default and running `make generate`

api/v1alpha3/conversion_test.go
```diff
@@ -22,6 +22,7 @@ import (
 	. "github.com/onsi/gomega"
 
 	fuzz "github.com/google/gofuzz"
+	"github.com/aws/aws-sdk-go/aws"
```
randomvariable, line 25:
```suggestion
	""k8s.io/utils/pointer"
```

Only use aws.Bool when working with AWS services.

api/v1alpha3/conversion.go
```diff
@@ -76,14 +88,39 @@ func (r *AWSClusterList) ConvertFrom(srcRaw conversion.Hub) error {
...
+	if err := Convert_v1alpha3_AWSMachine_To_v1alpha4_AWSMachine(r, dst, nil); err != nil {
+		return err
+	}
+
+	restored := &v1alpha4.AWSMachine{}
+	if ok, err := utilconversion.UnmarshalData(r, restored); err != nil || !ok {
+		return err
+	}
+
+	if restored.Spec.RootVolume != nil && restored.Spec.RootVolume.Encrypted == nil {
```
randomvariable, line 100:
```suggestion
	if dst.Spec.RootVolume != nil && dst.Spec.RootVolume.Encrypted == false && restored.Spec.RootVolume != nil && restored.Spec.RootVolume.Encrypted == nil {
```

Although this shouldn't happen, we shouldn't change true back to false.

api/v1alpha3/conversion.go
```diff
@@ -76,14 +88,39 @@ func (r *AWSClusterList) ConvertFrom(srcRaw conversion.Hub) error {
...
+		dst.Spec.RootVolume.Encrypted = nil
+	}
+
+	for i := range restored.Spec.NonRootVolumes {
```
randomvariable, line 104:
```suggestion
	for i := range dst.Spec.NonRootVolumes {
```

api/v1alpha3/conversion.go
```diff
@@ -76,14 +88,39 @@ func (r *AWSClusterList) ConvertFrom(srcRaw conversion.Hub) error {
...
+		if restored.Spec.NonRootVolumes[i].Encrypted == nil {
```
randomvariable, line 105:
```suggestion
		if len(restored.Spec.NonRootVolumes) > i && restored.Spec.NonRootVolumes[i].Encrypted == nil {
```

defend against mismatched slice lengths.

api/v1alpha3/conversion.go
```diff
@@ -104,14 +141,35 @@ func (r *AWSMachineList) ConvertFrom(srcRaw conversion.Hub) error {
...
+
+	if restored.Spec.Template.Spec.RootVolume != nil && restored.Spec.Template.Spec.RootVolume.Encrypted == nil {
+		dst.Spec.Template.Spec.RootVolume.Encrypted = nil
+	}
+
+	for i := range restored.Spec.Template.Spec.NonRootVolumes {
+		if restored.Spec.Template.Spec.NonRootVolumes[i].Encrypted == nil {
+			dst.Spec.Template.Spec.NonRootVolumes[i].Encrypted = nil
+		}
+	}
```
randomvariable, line 161: Instead of copy pasting the same code as for AWSMachine, can create a 

`Convert_v1alpha3_AWSMachineSpec_To_v1alpha4_AWSMachineSpec` and reuse the same code.

randomvariable:
> Currently, if the AMI is encrypted, user still needs to set the encrypted value to true. For not requiring encrypted field to be true when using an encrypted snapshot:
> We can try to find out if the AMI the instance is created from encrypted or not. Volume struct has that field. If the instance is encrypted, then continue as if the encrypted field is set to true.

This is overcomplicating things a bit. In the controller, for v1alpha3, we were doing `aws.Bool(blah)`, by converting to pointer, we send in the raw value. If we send in `nil`, then AWS uses the encryption value of the AMI automatically - it doesn't require us to to know whether the AMI is encrypted or not. The EC2 API is like this because for block volumes, we have two choices - create a volume from a snapshot (as in an AMI), or create a raw volume. For volumes created from snapshots, it's fine to use `nil`, and AWS will just copy the value from the AMI/snapshot. For blank raw volumes, nil is the same as false.

randomvariable:
> If AMI is encrypted in v1alpha4, then it will work. But in v1alpha3 it might not work because if encrypted=nil and AMI is encrypted in v1alpha4, then it will get converted to encrypted = false in v1alpha3 which might fail on the AWS side (since AMI is encrypted and value is false). Is this behavior acceptable?
> Even if we convert the encrypted nil value in v1alpha4 to true in v1alpha3, then the same problem might occur for non-encrypted AMIs.
Please correct if understanding is wrong

I think the question here is what happens if the controller gets downgraded back from v1alpha4 to v1alpha3. At that point there's not much we can do, data is going to be lost. We only care about roundtrip conversion for v1alpha3 clients, the v1alpha4 controller will always be the reconciler. At most we'll document the breaking change.

api/v1alpha3/conversion.go
```diff
@@ -258,3 +298,36 @@ func restoreInstance(restored, dst *v1alpha4.Instance) {
...
+			if len(dstVolumes) <= i {
+				dstVolumes = append(dstVolumes, restoredVolumes[i])
+			} else {
+				dstVolumes[i].Encrypted = nil
+			}
+		}
+	}
+}
+
+func RestoreRootVolume(restored, dst *v1alpha4.Volume) {
```
randomvariable, line 320: Can we add some commentary in this as to why we have this logic so it's clear to other people looking at the code as to why we're doing this?

randomvariable (changes requested):
Let's unexport these functions and add some commentary as to what's happening.

api/v1alpha3/conversion.go
```diff
@@ -258,3 +298,36 @@ func restoreInstance(restored, dst *v1alpha4.Instance) {
 	}
 	dst.VolumeIDs = restored.VolumeIDs
 }
+
```
randomvariable, line 301: Add comment for exported function.

api/v1alpha3/conversion.go
```diff
@@ -258,3 +298,36 @@ func restoreInstance(restored, dst *v1alpha4.Instance) {
...
+		if restoredVolumes[i].Encrypted == nil {
+			if len(dstVolumes) <= i {
+				dstVolumes = append(dstVolumes, restoredVolumes[i])
+			} else {
+				dstVolumes[i].Encrypted = nil
+			}
+		}
+	}
+}
+
```
randomvariable, line 319: Add comment for exported function

randomvariable:
ignore the pull-cluster-api-provider-aws-verify errors.

/override pull-cluster-api-provider-aws-verify

randomvariable (approved):

Issue #2563, opened 2021-07-02
CAPA created 2 VPCs and attached the ELB to the wrong VPC


sedefsavas:
/kind bug

**What steps did you take and what happened:**

In a very old cluster, there seems to be 2 different VPCs that are created for the same cluster. Controller might be down after the first VPC reconcile fails and when it comes back up since we don't persist VPC ID early in the reconciliation, the controller created a new one and attached to the loadbalancer the wrong VPC.

Instead we need to persist VPC ID very early in the reconcile loop like we do in subnets:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1915

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sedefsavas:
/assign

@praveenghuge assigning to myself as there is a bigger change required for this problem. Sorry for not assigning in the beginning.

PR #2566, opened 2021-07-05
Cleanup ReconcileVPC and set id early in reconciliation

sedefsavas:
/kind bug
1) Remove wait logic that wait until vpc is created. Instead returns early if `vpc.id` is set but AWS resource could not be retrieved.
2) Don't allow `spec.vpc.id` modifications in the webhook. This may cause problems such as if an unmanaged VPC is used and the VPC ID is nulled, then we create a new managed VPC but do not handle the ELB logic afterwards.
3) Create VPC and patch the ID immediately.
4) Also, instead of allowing not providing VPC ids in DescribeVPCs logic, modified it to describe only the VPC when the id is set in the spec.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2563

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Patch VPC ID immediately after VPC creation.
```

### DIFF
--- a/api/v1alpha4/awscluster_webhook.go
+++ b/api/v1alpha4/awscluster_webhook.go
@@ -109,6 +109,19 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 		)
 	}
 
+	// Modifying VPC id is not allowed because it will cause a new VPC creation if set to nil.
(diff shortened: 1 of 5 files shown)


randomvariable:
looks good, but want to go over the logic one more time before merging.

randomvariable:
it won't let me push a rebase, so i'll re-open this PR

PR #2571, opened 2021-07-08
Renaming spec.networkSpec to spec.network

### LINKED ISSUE #2119: Consider renaming spec.networkSpec to spec.network
/kind feature

The field called `networkSpec` seems redundant given that's already under `spec`, we can simplify by calling it just `network`.

Ankitasw:
**What type of PR is this?**
/kind feature

This PR renames the field `networkSpec` which seems redundant given that's already under spec, so we simplify it by changing it to `network`

Fixes #2119

**Checklist**:
- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Rename field `networkSpec` as `network`
```


api/v1alpha3/conversion.go
```diff
@@ -251,6 +252,24 @@ func Convert_v1alpha4_Instance_To_v1alpha3_Instance(in *v1alpha4.Instance, out *
...
+
+// Convert_v1alpha4_NetworkStatus_To_v1alpha3_Network is an autogenerated conversion function.
+func Convert_v1alpha4_NetworkStatus_To_v1alpha3_Network(in *v1alpha4.NetworkStatus, out *Network, s apiconversion.Scope) error {
+	out.SecurityGroups = *(*map[SecurityGroupRole]SecurityGroup)(unsafe.Pointer(&in.SecurityGroups))
+	if err := Convert_v1alpha4_ClassicELB_To_v1alpha3_ClassicELB(&in.APIServerELB, &out.APIServerELB, s); err != nil {
+		return err
+	}
+	return nil
+}
+
```
randomvariable, line 272: Why was this copied from the autogenerated conversions?

Ankitasw: shall we directly set it as out.SecurityGroups = in.SecurityGroups ?

randomvariable: Let's catch up tomorrow morning. Noticed a few other errors in the conversions that aren't related to this PR that need fixing first.

randomvariable: Had a check of this. Can confirm there isn't another way to do this. Can we update the comment about it being an "autogenerated conversion function".

for the specific comment, i would put something like
`Convert_v1alpha4_NetworkStatus_To_v1alpha3_Network is based on the autogenerated function and handles the renaming of the Network struct to NetworkStatus`

If you can, can you also get rid of the "is an autogenerated conversion function" text from the other comments in the conversion.go file, as it's not accurate.

Ankitasw: sure I will update the comments

randomvariable:
Can you rebase on top of main and then rerun the test, it should work then.

Ankitasw:
> /lgtm
> /approve

rebased the branch now

PR #2580, opened 2021-07-12
Add webhook to ensure that only ID references AMI

alexander-demichev:
**What type of PR is this?**
/kind feature

Currently, only ID is used to reference an AMI https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/ec2/instances.go#L132 however we expose two other API fields - ARN and Filters. This PR adds a webhook to prevent users from using unsupported values.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add webhook to ensure that only ID references AMI 
```

### DIFF
--- a/api/v1alpha4/awsmachinetemplate_webhook.go
+++ b/api/v1alpha4/awsmachinetemplate_webhook.go
@@ -55,6 +55,14 @@ func (r *AWSMachineTemplate) ValidateCreate() error {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "template", "spec", "providerID"), "cannot be set in templates"))
 	}
 
+	if spec.AMI.ARN != nil {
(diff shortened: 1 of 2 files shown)


randomvariable:
As we haven't implemented this, I think the better option is to delete these fields from AMIResourceReference in #2549

randomvariable:
Left a comment on the other PR

shivi28:
Removed unused fields like ARN and filters from AMIReference in PR #2549

randomvariable:
#2549 is queued for merge, so

/close

@alexander-demichev  might be worth pinging the channel in Slack before picking one of the other cleanups so we're not stepping on each other's toes. Sorry about that.

PR #2587, opened 2021-07-16
Cleanup ReconcileVPC() and set id early in reconciliation.

randomvariable:
/kind bug
1) Remove wait logic that wait until vpc is created. Instead returns early if `vpc.id` is set but AWS resource could not be retrieved.
2) Don't allow `spec.vpc.id` modifications in the webhook. This may cause problems such as if an unmanaged VPC is used and the VPC ID is nulled, then we create a new managed VPC but do not handle the ELB logic afterwards.
3) Create VPC and patch the ID immediately.
4) Also, instead of allowing not providing VPC ids in DescribeVPCs logic, modified it to describe only the VPC when the id is set in the spec.

Reopening @sedefsavas's PR rebased on top of main whilst she's on :palm_tree: 

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2563

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Patch VPC ID immediately after VPC creation, to deal with edge case where multiple VPCs may get created with the same tags.
```

### DIFF
--- a/api/v1alpha4/awscluster_webhook.go
+++ b/api/v1alpha4/awscluster_webhook.go
@@ -109,6 +109,19 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 		)
 	}
 
+	// Modifying VPC id is not allowed because it will cause a new VPC creation if set to nil.
+	if !reflect.DeepEqual(oldC.Spec.NetworkSpec, NetworkSpec{}) &&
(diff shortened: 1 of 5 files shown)


randomvariable:
I'm going to override verify's for now because of Go 1.17 rc changes.

/override pull-cluster-api-provider-aws-verify

PR #2601, opened 2021-07-26
Finalizers added after first read operation on AWS in AWSMachines

### LINKED ISSUE #2270: AWSMachine cannot be removed if controller is installed with invalid credentials
/kind bug

1. Install AWS provide with invalid credentials (malformed, revoked etc.)
2. Create a cluster
3. Delete a cluster

Controller gets stuck by calling `ec2svc.GetRunningInstanceByTags`, which fails because of wrong credentials. This make removal of `AWSMachine` object stuck.

Actually, the same happens for `AWSCluster`.

Controller should perhaps handle this case somehow.

**Environment:**

- Cluster-api-provider-aws version: fbf880963b07b2fc5f869faab0e0ab07aa06a729

shivi28:
**What type of PR is this?**
/kind bug

Update finalizer addition logic for AWSMachine. If we apply AWS workload cluster yaml with invalid credentials, then AWSMachine object is created and stuck because of invalid credentials. Once we try to delete that yaml, AWSMachine object stuck. So adding finalizer in AWSMachine only after first successful AWS read operation.

Fixes #2270 

With this change AWSMachine object is deleted in case of invalid credentials, same needs to be done for AWSCluster. WIP because AWSCluster finalizer updation requires a bit of refactoring in reconciliation logic

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
AWSMachine objects successfully deleted in case of invalid credentials
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -470,14 +470,6 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		return ctrl.Result{}, nil
 	}
 
-	// If the AWSMachine doesn't have our finalizer, add it.
-	controllerutil.AddFinalizer(machineScope.AWSMachine, infrav1.MachineFinalizer)
-	// Register the finalizer immediately to avoid orphaning AWS resources on delete
-	if err := machineScope.PatchObject(); err != nil {
-		machineScope.Error(err, "unable to patch object")
-		return ctrl.Result{}, err
-	}
-
 	if !machineScope.Cluster.Status.InfrastructureReady {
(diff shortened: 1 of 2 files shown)


randomvariable:
Need to update the tests appropriately, as the logic has changed.

sedefsavas:
One problem might be: this does not solve the case where the credentials were valid initially (the first findInstance() call succeeded), then for example it expired.

I don't think there is an easy solution for this.

randomvariable:
No probably not. I don't think there's much more we can do here. It at least unblocks the most trivial use case.

PR #2555, opened 2021-06-30
Do not delete overriden security groups

sedefsavas:
/kind bug

**What this PR does / why we need it**:
We were deleting SGs looking at `Status`, which also includes overridden SGs.
Instead, during SG deletion, this PR only deletes the ones that have `sigs.k8s.io/cluster-api-provider-aws/cluster/name: owned` tag.

Also, allows overriding SGs partially. Before all default SGs must have been overridden or seeing this error:
`"msg"="failed to reconcile security groups" "error"="security group overrides have been provided for some but not all roles - missing security group for role bastion" "cluster"="tagleme" "name"="tagleme" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AWSCluster" `

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2548

**Special notes for your reviewer**:

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:
```release-note
Do not delete security groups when provided as overrides
```

PR #2620, opened 2021-07-29
🐛 Managed SecurityGroups should be filtered correctly

sedefsavas:
/kind bug

There was an old bug in `describeClusterOwnedSecurityGroups` which filters SGs by provider-owned filter instead of cluster-owned. This bug is surfaced and fixed during the changes: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2555. But during backport, I forgot to fix that which causes e2e's fail: SGs are not being deleted and hence VPC could not be deleted.
-aws/pull/2560/

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Managed SecurityGroups should be filtered correctly.
```


randomvariable:
oh wow, that was a nice find.

/lgtm
/approve

Issue #2609, opened 2021-07-28
awsmachinepool do not respect subnets filters


calvix:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
`AWSMachinePool.Spec.Subnets` is type `[]AWSResourceReference` which allow AWS resource specification via:
* ID
* ARN
* Filters

But the implementation only considers IDs and ignores the rest.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/scope/machinepool.go#L224-L227

**What did you expect to happen:**
defined subnets filters are respected as it is implemented for MachineDeployment:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/ec2/instances.go#L293-L319

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 0.4.0
- Kubernetes version: (use `kubectl version`): any
- OS (e.g. from `/etc/os-release`): any

randomvariable:
I'm already working on this one in fact. I'll be removing the unused fields until such a time they actually get implemented.

/lifecycle active
/assign

Part of #2087

randomvariable:
basically, Subnets  --> []SubnetReference
[]SecurityGroups --> []SecurityGroupReference

and so forth.

Deleting AWSResourceReference from v1alpha4 so we don't end up doing this again :(

calvix:
> 

> I'm already working on this one in fact. I'll be removing the unused fields until such a time they actually get implemented.
> 
> /lifecycle active
> /assign
> 
> Part of #2087

I actually have a prepared code to implement the lookup by filters, I would like to have the option there

randomvariable:
@calvix sure, i can rebase over the top.

If you can, but not 100% necessary, can you convert AWSResourceReference into SubnetReference?

Also, this problem isn't just in AWSMachinePool but also AWSCluster so might be worth seeing if you can put it in both places.

calvix:
jumping back on this, @randomvariable  where is type SubnetReference located? in which package?

sedefsavas:
@calvix I `SubnetReference` would be a new struct which holds only the fields that is used (for instance if ARN is not used, it should not be there) but we can no longer do any changes in the api (AWSResourceRef to SubnetRef) until v1beta2 release, so that change has to wait for a bit.
[...]

Issue #2087, opened 2020-11-02
AWSResourceReference unimplemented logic


MarcusNoble:
Collating all instances of `AWSResourceReference` being used in the codebase without implementing all lookup logic (e.g. only using the `ID` property)

| Type | Property | References | ID | ARN | Filters |
| --- | --- | --- | --- | --- | --- |
| `AWSMachineSpec` | `AMI` | https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2e24bd127d875995ed13699966e60b9e12060cb2/pkg/cloud/services/ec2/instances.go#L132-L134 | ✅ | ❌ | ❌ |
| `AWSMachineSpec` | `AdditionalSecurityGroups` | https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2e24bd127d875995ed13699966e60b9e12060cb2/controllers/awsmachine_security_groups.go#L62-L64 | ✅ | ❌ | ❌ |
| `AWSMachineSpec` | `Subnet`| https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2e24bd127d875995ed13699966e60b9e12060cb2/pkg/cloud/services/ec2/instances.go#L261-L343 | ✅ | ❌ | ✅ |
| `AWSMachinePoolSpec` | `Subnets` | https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2e24bd127d875995ed13699966e60b9e12060cb2/pkg/cloud/services/autoscaling/autoscalinggroup.go#L144-L146 | ✅ | ❌ | ❌ |

MarcusNoble:
It looks as though there aren't as many instances as first thought.

If ignoring test files and generated files `AWSResourceReference` is only made use of in 4 places.

All 4 make use of the `ID` property, one makes use of the `Filters` property and there are no instances where `ARN` is used.

sedefsavas:
lifecycle/stale

Ankitasw:
@sedefsavas @richardcase Just to clarify, here we need to implement lookup logic for fields which are of type  `AWSResourceReference` and using any of the properties(ID, Filters) from `AWSResourceReference` in the respective codes.

shivi28:
Pasting slack discussion here for others reference:
Scope of this issue is 
```
if subnet id is there, do a lookup using that
if filter is there, do a lookup using that
If both are given then throw a validation error 
```
But currently in our code we are only doing lookup by IDs.

[...]

PR #2634, opened 2021-08-03
fix: awsmachinepool use subnet filters

calvix:
**What type of PR is this?**

/kind bug

implements  was subnet lookup specified by filters in `AWSMachinePool.Spec.Subnets`

The previous implementation only worked with IDs, but now users can specify filters to select specific subnets that will be used for the machinepool.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2609

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Add implementation of subnets filters for AWSMachinePool CR
```


randomvariable:
I'd prefer for this to be implemented everywhere there is a Subnet currently using a AWSResourceReference, rather than having this very ad hoc implementation in this one specific CRD. It doesn't really fix the API issue we have.

I'll try and include this in the API cleanup.

calvix:
> I'd prefer for this to be implemented everywhere there is a Subnet currently using a AWSResourceReference, rather than having this very ad hoc implementation in this one specific CRD. It doesn't really fix the API issue we have.
> 
> I'll try and include this in the API cleanup.

As far as I know, the only other resource that has reference to the subnet is AWSMachine and the machine controller already has the implementation for the filters (I used it as a template for this) 

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/327e9b8a0a81d05ede3a9f61c30fef2947fa05ef/pkg/cloud/services/ec2/instances.go#L293-L320

randomvariable:
In which case, can we dedupe the code and have it in a single plac?

sedefsavas:
@calvix Can we try to implement a common method for all other subnets to use as @randomvariable pointed in the comment above?
We cannot have an API change but I think if we at least implement a common logic to handle filter field, it will be straightforward to fix the api in the v1beta2 release.

AverageMarcus:
@sedefsavas @randomvariable Do you have any suggestions for how best to refactor this in a non-API-changing way? 

The inconsistencies with `AWSResourceReference` have been [long-standing](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2087) and not yet addressed in any of the API changes so would hate to have this blocked on that.

`AWAMachinePool` and `AWSMachine` are (as far as I can see) the only two types that use `AWSResourceReference` for defining subnets. The former as a slice, the later as a single pointer, so the logic between the two isn't exactly the same so I'm not sure a single function even makes sense for this.

calvix:
@sedefsavas @randomvariable  I also think the logic between the two isn't exactly the same so I'm not sure a single function even makes sense for this. Unless you have some good suggestions on how to do that.

shivi28:
Thank you so much @calvix for raising this PR and doing initial research/discussion to resolve #2609. 🎉 

And due to priority reasons, I also have raised a [PR](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3255) to implement the remaining logic and feeling happy to add you as a co-author :)

Ankitasw:
/close
in favor of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3255

calvix:
thanks for sorting it out :)

PR #2271, opened 2021-02-25
✨ Add support for Ignition-based bootstrap data and Flatcar Container Linux

invidian:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

This PR adds support for handling bootstrap data in Ignition format, which will hopefully be produced by CABPK v1alpha4 implemented by PR https://github.com/kubernetes-sigs/cluster-api/pull/4172 (not merged yet).

As Ignition does not plan to support multi part mime for user-data (https://github.com/coreos/ignition/issues/1072), this PR implements support for putting bootstrap data in Ignition format in S3 bucket, to workaround the 64kb limitation on user-data for EC2 instances.

Similarly to SSM support, bootstrap data is removed after node has successfully bootstrapped.

Right now single S3 bucket is used for each cluster, as we need a single place to control bucket policies, which are used to restrict access for control-plane and worker nodes to only be able to access their own bootstrap data.

Finally, new template is added tailored for Flatcar Container Linux, which contains references to right now unofficial AMIs, though Flatcar support is fully added in `image-builder` already.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1979
Refs #1875 

**Special notes for your reviewer**:

This PR can't probably be merged at this point yet. We probably need to wait until CAPI PR is merged and released, but as this PR goes in pair with CAPI one, it would be awesome to be able to get some early feedback on the approach taken.

@dongsupark is also working on adding e2e tests for it.

**Checklist**:

- [x] includes documentation
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Added support for bootstrap data in Ignition format and for Flatcar Container Linux
```

PR #2662, opened 2021-08-11
Separate EKS-related IAM resources

### LINKED ISSUE #2642: Work around AWS managed policy size limit
AWS has a [character limit](https://docs.aws.amazon.com/IAM/latest/UserGuide/reference_iam-quotas.html) of **6144** for managed policies.

As of e69740275ca5a14b51eeca3ed74cff2d88dd8cdc (current tip of `main`) we are nearly at this limit for the `AWSIAMManagedPolicyControllers` policy:

```console
cd /tmp
git clone git@github.com:kubernetes-sigs/cluster-api-provider-aws.git
cd cluster-api-provider-aws
git checkout e69740275ca5a14b51eeca3ed74cff2d88dd8cdc
make clusterawsadm
```

Generate the config file (this specific file is used in the e2e tests):

```console
cat <<EOF >/tmp/config.yaml
---
apiVersion: bootstrap.aws.infrastructure.cluster.x-k8s.io/v1alpha1
kind: AWSIAMConfiguration
spec:
  nameSuffix: ".cluster-api-provider-aws.sigs.k8s.io"
  controlPlane:
    disable: false
    disableCloudProviderPolicy: false
    enableCSIPolicy: false
  clusterAPIControllers:
    disable: false
    allowedEC2InstanceProfiles:
    - "*.cluster-api-provider-aws.sigs.k8s.io"
  nodes:
    disable: false
    disableCloudProviderPolicy: false
    ec2ContainerRegistryReadOnly: false
  bootstrapUser:
    enable: true
    userName: bootstrapper.cluster-api-provider-aws.sigs.k8s.io
    extraStatements:
    - Effect: Allow
      Action:
      - sts:AssumeRole
      Resource:
      - eyAiRm46OkdldEF0dCIgOiBbICJDQVBBTXVsdGlUZW5hbmN5U2ltcGxlIiwgIkFybiIgXSB9
      - eyAiRm46OkdldEF0dCIgOiBbICJDQVBBTXVsdGlUZW5hbmN5SnVtcCIsICJBcm4iIF0gfQ==
    - Effect: Allow
      Action:
      - servicequotas:GetServiceQuota
      - servicequotas:RequestServiceQuotaIncrease
      - servicequotas:ListRequestedServiceQuotaChangeHistory
      - elasticloadbalancing:DescribeAccountLimits
      - ec2:DescribeAccountLimits
      - cloudtrail:LookupEvents
      - ssm:StartSession
      - ssm:DescribeSessions
      - ssm:GetConnectionStatus
      - ssm:DescribeInstanceProperties
      - ssm:GetDocument
      - ssm:TerminateSession
      - ssm:ResumeSession
      Resource:
      - "*"
    - Effect: Allow
      Action:
      - iam:CreateServiceLinkedRole
      Resource:
      - arn:*:iam::*:role/aws-service-role/autoscaling.amazonaws.com/AWSServiceRoleForAutoScaling
      - arn:*:iam::*:role/aws-service-role/servicequotas.amazonaws.com/AWSServiceRoleForServiceQuotas
      - arn:*:iam::*:role/aws-service-role/support.amazonaws.com/AWSServiceRoleForSupport
      - arn:*:iam::*:role/aws-service-role/trustedadvisor.amazonaws.com/AWSServiceRoleForTrustedAdvisor
  stackName: cluster-api-provider-aws-sigs-k8s-io
  region: eu-central-1
  eks:
    enable: true
    defaultControlPlaneRole:
      disable: false
    managedMachinePool:
      disable: false
    fargate:
      disable: true
    kmsAliasPrefix: cluster-api-provider-aws-*
  eventBridge:
    enable: true
  partition: aws
  secureSecretBackends:
  - secrets-manager
  - ssm-parameter-store
EOF
```

Render the IAM policy we use for controllers, strip the whitespace and count the characters:

```console
./bin/clusterawsadm bootstrap iam print-policy \
    --document AWSIAMManagedPolicyControllers \
    --config /tmp/config.yaml | tr -d '[:space:]' | wc -c
# 6012
```

Adding just a few lines to this policy breaks the e2e tests:

```console
cat <<EOF | git apply --ignore-whitespace -
diff --git a/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go b/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
index a36a6755..258fd2b4 100644
--- a/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
+++ b/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
@@ -206,6 +206,19 @@ func (t Template) ControllersPolicy() *infrav1.PolicyDocument {
                                "iam:PassRole",
                        },
                },
+               {
+                       Effect: infrav1.EffectAllow,
+                       Resource: infrav1.Resources{
+                               "arn:*:s3:::*",
+                       },
+                       Action: infrav1.Actions{
+                               "s3:CreateBucket",
+                               "s3:DeleteBucket",
+                               "s3:PutObject",
+                               "s3:DeleteObject",
+                               "s3:PutBucketPolicy",
+                       },
+               },
        }
        for _, secureSecretBackend := range t.Spec.SecureSecretsBackends {
                switch secureSecretBackend {
EOF
```

Rebuild and check the template size:

```console
make clusterawsadm
./bin/clusterawsadm bootstrap iam print-policy \
    --document AWSIAMManagedPolicyControllers \
    --config /tmp/config.yaml | tr -d '[:space:]' | wc -c
# 6160!
```

Try to run the e2e tests:

```console
AWS_REGION=eu-central-1 make test-e2e GINKGO_ARGS=-stream E2E_ARGS=-skip-cloudformation-deletion
```

The above fails with the following error in CloudFormation:

```
Cannot exceed quota for PolicySize: 6144 (Service: AmazonIdentityManagement; Status Code: 409; Error Code: LimitExceeded; Request ID: 6b6a1bc0-2c6c-440e-80a8-7e976304e69e; Proxy: null)
```

I've encountered this issue while working on https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2271.

We should figure out a solution which allows evolving the various IAM policies over time in a scalable manner. Some options:

- Break down long policies into smaller policies.
- Simplify the policies by grouping permissions with identical resource + effect into a single statement.
- Use wildcards such as `ec2:Describe*`. **Warning:** Wildcards could have the side effect of adding unwanted permissions as AWS adds new IAM actions.

johananl:
**What type of PR is this?**
/kind bug

This PR works around the AWS character limit for managed policies. In the e2e tests the `AWSIAMManagedPolicyControllers` policy is currently very close to the AWS limit, which means that the next time a feature adds a few lines to this policy, the CloudFormation template would error out (I've already hit this limit while working on #2271, hence this issue + PR). More info and repro instructions in #2642.

Fixes #2642

**Checklist**:

- [x] squashed commits
- [x] adds unit tests

**Release note**:

```release-note
Move EKS-related resources from `AWSIAMManagedPolicyControllers` to a separate policy to work around AWS size limit for managed policies.
```

### DIFF
--- a/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
+++ b/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
@@ -235,169 +235,6 @@ func (t Template) ControllersPolicy() *infrav1.PolicyDocument {
 			})
 		}
 	}
-	if !t.Spec.EKS.Disable {
-		allowedIAMActions := infrav1.Actions{
-			"iam:GetRole",
-			"iam:ListAttachedRolePolicies",
-		}
-		statement = append(statement, infrav1.StatementEntry{
-			Effect: infrav1.EffectAllow,
-			Resource: infrav1.Resources{
-				"arn:*:ssm:*:*:parameter/aws/service/eks/optimized-ami/*",
-			},
-			Action: infrav1.Actions{
-				"ssm:GetParameter",
-			},
-		})
-
-		statement = append(statement, infrav1.StatementEntry{
-			Effect: infrav1.EffectAllow,
-			Action: infrav1.Actions{
-				"iam:CreateServiceLinkedRole",
-			},
-			Resource: infrav1.Resources{
-				"arn:*:iam::*:role/aws-service-role/eks.amazonaws.com/AWSServiceRoleForAmazonEKS",
-			},
-			Condition: infrav1.Conditions{
-				infrav1.StringLike: map[string]string{"iam:AWSServiceName": "eks.amazonaws.com"},
-			},
-		})
-
-		statement = append(statement, infrav1.StatementEntry{
-			Effect: infrav1.EffectAllow,
-			Action: infrav1.Actions{
-				"iam:CreateServiceLinkedRole",
-			},
-			Resource: infrav1.Resources{
-				"arn:*:iam::*:role/aws-service-role/eks-nodegroup.amazonaws.com/AWSServiceRoleForAmazonEKSNodegroup",
-			},
-			Condition: infrav1.Conditions{
-				infrav1.StringLike: map[string]string{"iam:AWSServiceName": "eks-nodegroup.amazonaws.com"},
-			},
-		})
-
-		statement = append(statement, infrav1.StatementEntry{
-			Effect: infrav1.EffectAllow,
-			Action: infrav1.Actions{
-				"iam:CreateServiceLinkedRole",
-			},
-			Resource: infrav1.Resources{
-				"arn:aws:iam::*:role/aws-service-role/eks-fargate-pods.amazonaws.com/AWSServiceRoleForAmazonEKSForFargate",
-			},
-			Condition: infrav1.Conditions{
-				infrav1.StringLike: map[string]string{"iam:AWSServiceName": "eks-fargate.amazonaws.com"},
-			},
-		})
-
-		if t.Spec.EKS.AllowIAMRoleCreation {
-			allowedIAMActions = append(allowedIAMActions, infrav1.Actions{
-				"iam:DetachRolePolicy",
-				"iam:DeleteRole",
-				"iam:CreateRole",
-				"iam:TagRole",
-				"iam:AttachRolePolicy",
-			}...)
-
-			statement = append(statement, infrav1.StatementEntry{
-				Action: infrav1.Actions{
-					"iam:ListOpenIDConnectProviders",
-					"iam:CreateOpenIDConnectProvider",
-					"iam:AddClientIDToOpenIDConnectProvider",
-					"iam:UpdateOpenIDConnectProviderThumbprint",
-					"iam:DeleteOpenIDConnectProvider",
-				},
-				Resource: infrav1.Resources{
-					"*",
-				},
-				Effect: infrav1.EffectAllow,
-			})
-		}
-		statement = append(statement, []infrav1.StatementEntry{
-			{
-				Action: allowedIAMActions,
-				Resource: infrav1.Resources{
-					"arn:*:iam::*:role/*",
-				},
-				Effect: infrav1.EffectAllow,
-			}, {
-				Action: infrav1.Actions{
-					"iam:GetPolicy",
-				},
-				Resource: infrav1.Resources{
-					t.generateAWSManagedPolicyARN(eksClusterPolicyName),
-				},
-				Effect: infrav1.EffectAllow,
-			}, {
-				Action: infrav1.Actions{
-					"eks:DescribeCluster",
-					"eks:ListClusters",
-					"eks:CreateCluster",
-					"eks:TagResource",
-					"eks:UpdateClusterVersion",
-					"eks:DeleteCluster",
-					"eks:UpdateClusterConfig",
-					"eks:UntagResource",
-					"eks:UpdateNodegroupVersion",
-					"eks:DescribeNodegroup",
-					"eks:DeleteNodegroup",
-					"eks:UpdateNodegroupConfig",
-					"eks:CreateNodegroup",
-					"eks:AssociateEncryptionConfig",
-				},
-				Resource: infrav1.Resources{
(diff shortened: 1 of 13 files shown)


cmd/clusterawsadm/cloudformation/bootstrap/template.go
```diff
@@ -96,6 +97,16 @@ func (t Template) RenderCloudFormation() *cloudformation.Template {
 		Roles:             t.controllersPolicyRoleAttachments(),
 	}
 
+	if t.Spec.EKS.Enable {
```
sedefsavas, line 100: This field is going to be changed with https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2648

Let's wait for it to be merged first and then rebase.
/hold

johananl:
>Let's wait for it to be merged first and then rebase.

@sedefsavas done.

randomvariable:
Given the impact of this is creating new policies, I think we may need to bump the minor version to v0.8.0

johananl:
> Given the impact of this is creating new policies, I think we may need to bump the minor version to v0.8.0

OK. Is there anything I should do to make this happen @randomvariable?

Also, if we're incrementing the version for this, should we proactively move any other IAM statements to new policies while we're at it? Right now the EKS-related statements seem to make the biggest difference in terms of policy size.

randomvariable:
> Also, if we're incrementing the version for this, should we proactively move any other IAM statements to new policies while we're at it?

I think so. Can you create an issue documenting any suggested changes as we need to think carefully about upgrades here.

I'm placing the hold back on because we haven't decided on a branch split yet - will raise at next office hours.

/hold

randomvariable:
This is big enough impact in terms of upgrades that changes should be recorded in an ADR too, see https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/docs/adr

johananl:
>Can you create an issue documenting any suggested changes as we need to think carefully about upgrades here.

@randomvariable I took a closer look at the various IAM policies. AFAICT the only policy which has a risk of hitting the 6144-char size limit is `AWSIAMManagedPolicyControllers`. Within that policy, EKS-related resources amount to ~50% of the policy size, which makes them good candidates for moving into a new policy, especially given the fact that these statements already live neatly under a conditional based on the `t.Spec.EKS.Disable` knob.

Given the above and given that we have to balance between the **per-policy size limit** and the limit on the **number of policies** which can be attached to an IAM entity (default: **10**), I don't think we need to do anything other than break out the EKS-related resources from `AWSIAMManagedPolicyControllers` at the moment. We *could*, of course, go more fine-grained than that, but in terms of policy size nothing else seems to justify creating a new policy.

As far as upgrades and ADRs go, I'm happy to follow any necessary process, however do you feel this is indeed such a big change? We're talking about moving IAM statements which are already applied to a new policy that's attached **to the same entities** (roles/users/groups). Assuming we don't make any policy logic mistakes and the "allow/deny" end result is identical, AFAICT the users don't need to do anything special when upgrading and I can't find any reason for downtime or other problems. Following the same logic, this change also doesn't feel like an architectural change to me.

What do you think?

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
```diff
@@ -235,190 +235,198 @@ func (t Template) ControllersPolicy() *infrav1.PolicyDocument {
...
 		})
+	}
+
+	return &infrav1.PolicyDocument{
+		Version:   infrav1.CurrentVersion,
+		Statement: statement,
+	}
+}
+
+// ControllersPolicyEKS creates a policy from a template for AWS Controllers.
```
invidian, line 266: @johananl I think it would be nice to document the 10 policies limit somewhere in here, to indicate why other resources has not been put in separate policies. We could as well do this in #2271 though.

randomvariable: I think it's ok for now. The 10 policies limit should be relatively well understood by folk in the IAM space.

PR #2673, opened 2021-08-16
Add log-level customization for artifact logs in E2E runs

Ankitasw:
**What type of PR is this?**
/kind feature

This PR adds log-level for artifacts logs of the E2E test runs which would be useful for development/debugging.

**Checklist**:
- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Controller log level can be set using the CAPA_LOGLEVEL variable. Defaults to 0.
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -21,6 +21,7 @@ spec:
         - "--metrics-bind-addr=127.0.0.1:8080"
         - "--leader-elect"
         - "--feature-gates=EKS=${CAPA_EKS:=true},EKSEnableIAM=${CAPA_EKS_IAM:=false},EKSAllowAddRoles=${CAPA_EKS_ADD_ROLES:=false},EKSFargate=${EXP_EKS_FARGATE:=false},MachinePool=${EXP_MACHINE_POOL:=false},EventBridgeInstanceState=${EVENT_BRIDGE_INSTANCE_STATE:=false},AutoControllerIdentityCreator=${AUTO_CONTROLLER_IDENTITY_CREATOR:=true}"
+        - "--v=${CAPA_LOGLEVEL:=0}"
         image: controller:latest
         imagePullPolicy: Always
         name: manager


config/manager/manager.yaml
```diff
@@ -21,6 +21,7 @@ spec:
         - "--metrics-bind-addr=127.0.0.1:8080"
         - "--leader-elect"
         - "--feature-gates=EKS=${CAPA_EKS:=true},EKSEnableIAM=${CAPA_EKS_IAM:=false},EKSAllowAddRoles=${CAPA_EKS_ADD_ROLES:=false},EKSFargate=${EXP_EKS_FARGATE:=false},MachinePool=${EXP_MACHINE_POOL:=false},EventBridgeInstanceState=${EVENT_BRIDGE_INSTANCE_STATE:=false},AutoControllerIdentityCreator=${AUTO_CONTROLLER_IDENTITY_CREATOR:=true}"
+        - "--v=${EKS_CP_LOGLEVEL:=4}"
```
richardcase, line 24: Now we have merged the controllers together, it's not specific to EKS. 

perhaps we should change its name to `CAPA_LOGLEVEL`, wdyt?

richardcase (changes requested):

config/manager/manager.yaml
Ankitasw: sounds fine, will make the change

config/manager/manager.yaml
```diff
@@ -21,6 +21,7 @@ spec:
         - "--metrics-bind-addr=127.0.0.1:8080"
         - "--leader-elect"
         - "--feature-gates=EKS=${CAPA_EKS:=true},EKSEnableIAM=${CAPA_EKS_IAM:=false},EKSAllowAddRoles=${CAPA_EKS_ADD_ROLES:=false},EKSFargate=${EXP_EKS_FARGATE:=false},MachinePool=${EXP_MACHINE_POOL:=false},EventBridgeInstanceState=${EVENT_BRIDGE_INSTANCE_STATE:=false},AutoControllerIdentityCreator=${AUTO_CONTROLLER_IDENTITY_CREATOR:=true}"
+        - "--v=${CAPA_LOGLEVEL:=4}"
```
randomvariable, line 24:
```suggestion
        - "--v=${CAPA_LOGLEVEL:=0}"
```

Let's leave the default as 0, and devs can bump it to 4 when they want to.

PR #2740, opened 2021-09-07
Validation added for empty instanceType in AWSMachine and AWSMachineTemplate

### LINKED ISSUE #2686: CAPA lets you define an AWSMachine and AWSMachineTemplate without specifying an instance type and then it fails.
/kind bug

Specify an AWSMachineTemplate without setting instanceType
CAPA will try to create a resultant AWSMachine and then error out with 

```
lastTransitionTime: "2021-08-19T18:02:16Z"
    message: "failed to create AWSMachine instance: failed to run instance: InvalidParameterValue:
```

Have the template and AWSMachine rejected by validating webhook

**Environment:**

- Cluster-api-provider-aws version: v0.6.6

shivi28:
**What type of PR is this?**
/kind bug

Validation added for empty InstanceType in AWSMachine and AWSMachineTemplate.

Fixes #2686

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Validation added for empty InstanceType in AWSMachine and AWSMachineTemplate.
```

### DIFF
--- a/api/v1alpha4/awsmachine_types.go
+++ b/api/v1alpha4/awsmachine_types.go
@@ -72,7 +72,9 @@ type AWSMachineSpec struct {
 	ImageLookupBaseOS string `json:"imageLookupBaseOS,omitempty"`
 
 	// InstanceType is the type of instance to create. Example: m4.xlarge
-	InstanceType string `json:"instanceType,omitempty"`
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength:=2
+	InstanceType string `json:"instanceType"`
 
 	// AdditionalTags is an optional set of tags to add to an instance, in addition to the ones added by default by the
 	// AWS provider. If both the AWSCluster and the AWSMachine specify the same tag name with different values, the
--- a/api/v1alpha4/awsmachine_webhook_test.go
+++ b/api/v1alpha4/awsmachine_webhook_test.go
@@ -49,6 +49,7 @@ func TestAWSMachine_Create(t *testing.T) {
 					RootVolume: &Volume{
 						Type: "io1",
 					},
+					InstanceType: "test",
 				},
 			},
 			wantErr: true,
@@ -60,6 +61,7 @@ func TestAWSMachine_Create(t *testing.T) {
 					RootVolume: &Volume{
 						Type: "io2",
 					},
+					InstanceType: "test",
 				},
 			},
 			wantErr: true,
@@ -71,6 +73,7 @@ func TestAWSMachine_Create(t *testing.T) {
 					RootVolume: &Volume{
 						Throughput: aws.Int64(-125),
 					},
+					InstanceType: "test",
 				},
 			},
 			wantErr: true,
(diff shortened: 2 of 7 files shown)


api/v1alpha4/awsmachine_webhook.go
```diff
@@ -49,6 +49,10 @@ var (
 func (r *AWSMachine) ValidateCreate() error {
 	var allErrs field.ErrorList
 
+	if r.Spec.InstanceType == "" {
+		allErrs = append(allErrs, field.Required(field.NewPath("spec.instancetype"), "instancetype is required"))
```
pydctw, line 53: `InstanceType` is an optional field in AWSMachine with json tag of `omitempty` so I don't think we can error out when instancetype is empty.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1alpha4/awsmachine_types.go#L75

If we want to make it a required field, we need to change the API. @randomvariable, thoughts?

randomvariable: This being optional and omitempty was a mistake I think, since the reconciler just ended up passing "" to ec2.RunInstance and failing, so I think it's reasonable to fix.

Ideally, we should mark the package as validation required and  then use the optional tags properly. I think in this case, add `// +kubebuilder:validation:Required` to the field and use OpenAPI validation instead of doing it in the webhook.

pydctw: Agreed that it's good to fix. Just wanted to point out that changing a field from optional to required is a breaking API change and we need to inform the users. 
Using kubebuilder validation marker sounds like a good idea.

api/v1alpha4/awsmachine_types.go
```diff
@@ -72,7 +72,8 @@ type AWSMachineSpec struct {
 	ImageLookupBaseOS string `json:"imageLookupBaseOS,omitempty"`
 
 	// InstanceType is the type of instance to create. Example: m4.xlarge
-	InstanceType string `json:"instanceType,omitempty"`
+	// +kubebuilder:validation:Required
```
richardcase, line 75: I'd also be tempted to add a min length to stop people just using 1 character.

```go
```

dlipovetsky:
I recommend adding this to the release note:

```
An AWSMachine or AWSMachineTemplate must define the `instanceType` field. Previously, these resources did not have to define this field.  An undefined field would not have been used in practice, because the controller would fail to create an EC2 instance. 
```

dlipovetsky:
lgtm in general, but I'm not sure whether it's ok to update v1alpha4, or whether we need to start v1alpha5.

sedefsavas:
I think since this is broken as is, this change won't break a working Cluster. I opt for adding it into v1alpha4 release.

Since we are targeting for a beta release soon, another option would be to add this in that release.
/lgtm

dlipovetsky:
I was also concerned that users may be leaving the instanceType field empty, and deferring its value to a mutating webhook, and this change could break this workflow.

However, the change cannot break this workflow: This is a validating webhook, so it sees [the final state of the requested object](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#guaranteeing-the-final-state-of-the-object-is-seen), _after_ all mutating webhooks have run.

Therefore, I believe we can make this change without breaking any users.

/approve

PR #2751, opened 2021-09-09
EKS: support role additional policies for AWSManagedMachinePool

### LINKED ISSUE #2750: Support RoleAdditionalPolicies for ManagedMachinePool
/kind feature

Currently, `AWSManagedControlPlane` supports both,`roleName` and `roleAdditionalPolicies`.
`AWSManagedMachinePool` supports only `roleName`.

Just like in MCP, we need to allow `roleAdditionalPolicies` in `AWSManagedMachinePool`, to allow attaching additional policies applicable to the node group role.

abhinavnagaraj:
**What type of PR is this?**
/kind feature

The node group role that is created per node group contains a list of default/pre-defined policies. This PR adds a property in AWSManagedMachinePool to allow attaching additional policies to this node group role.

Fixes #2750 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
EKS: Support role additional policies for AWSManagedMachinePool
```

### DIFF
--- a/pkg/cloud/services/eks/roles.go
+++ b/pkg/cloud/services/eks/roles.go
@@ -204,6 +204,14 @@ func (s *NodegroupService) reconcileNodegroupIAMRole() error {
 	}
 
 	policies := NodegroupRolePolicies()
+	if len(s.scope.ManagedMachinePool.Spec.RoleAdditionalPolicies) > 0 {
+		if !s.scope.AllowAdditionalRoles() {
+			return ErrCannotUseAdditionalRoles
+		}
+
+		policies = append(policies, s.scope.ManagedMachinePool.Spec.RoleAdditionalPolicies...)
+	}
+
 	_, err = s.EnsurePoliciesAttached(role, aws.StringSlice(policies))
 	if err != nil {
 		return errors.Wrapf(err, "error ensuring policies are attached: %v", policies)
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmanagedmachinepools.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmanagedmachinepools.yaml
@@ -351,6 +351,13 @@ spec:
                       plane is used.
                     type: string
                 type: object
+              roleAdditionalPolicies:
+                description: RoleAdditionalPolicies allows you to attach additional
+                  polices to the node group role. You must enable the EKSAllowAddRoles
+                  feature flag to incorporate these into the created role.
+                items:
+                  type: string
+                type: array
               roleName:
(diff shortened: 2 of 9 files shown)


richardcase:
@abhinavnagaraj - i see the issue and its with the fuzzy tests because you've added a new field that doesn't exist on the AWSManagedControlPlane in v1alpha3. Pinged you on slack.

pkg/cloud/services/eks/roles.go
```diff
@@ -204,6 +204,14 @@ func (s *NodegroupService) reconcileNodegroupIAMRole() error {
 	}
 
 	policies := NodegroupRolePolicies()
+	if s.scope.ManagedMachinePool.Spec.RoleAdditionalPolicies != nil {
```
richardcase, line 207: Could we do a **len** check here instead of a nil check?

abhinavnagaraj: ack

pkg/cloud/services/eks/roles.go
```diff
@@ -204,6 +204,14 @@ func (s *NodegroupService) reconcileNodegroupIAMRole() error {
 	}
 
 	policies := NodegroupRolePolicies()
+	if len(s.scope.ManagedMachinePool.Spec.RoleAdditionalPolicies) > 0 {
```
shivi28, line 207: Can we have a `nil` exception check at `s.scope.ManagedMachinePool`

richardcase: This shouldn't be nil but looking at the creation of the scope we don't explicitly check for nil on the managed machine pool. I'd favor the nil check at the time the scope is created instead
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/scope/managednodegroup.go#L62

wdyt @shivi28 @abhinavnagaraj  ?

abhinavnagaraj: Nil check is probably not required because we return if managedMachinePool is missing -- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3949f5376410f694defdf0ffb257a27835d7f9e5/exp/controllers/awsmanagedmachinepool_controller.go#L93
Let me add it during scope creation.

richardcase: Great, thanks.

richardcase:
e2e tests [passed](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2751/pull-cluster-api-provider-aws-e2e-eks/1443511254709702656) so:

/lgtm
/approve

PR #2768, opened 2021-09-14
:bug: Rename the ELB scheme from Internet-facing to internet-facing

### LINKED ISSUE #2534: AWSCluster `controlPlaneEndpoint.scheme` defaults - validation issue blocks editing neighbouring mutable fields
/kind bug
Hey guys, not sure if this has been reported before or if it's expected behaviour, but I've found that the default value for 

`awscluster.spec.controlPlaneEndpoint.scheme` Is validated as `Internet-facing` but actual value set by applying an AWSCluster without `spec.controlPlaneEndpoint` is `internet-facing`. When you then set other `controlPlaneEndpoint` values, such as additional SG rules for the control plane load balancer, the object fails validation with the below error:

`The AWSCluster "capi-dev" is invalid: spec.controlPlaneLoadBalancer.scheme: Invalid value: "Internet-facing": field is immutable`

For the value to be set as the proper default `Internet-facing` when applying an AWSCluster without a controlPlaneLoadBalancer struct, and to be able to edit other keys under controlPlaneLoadBalancer that aren't immutable.

**Environment:**

- Cluster-api-provider-aws version: 9b294114b
- Kubernetes version: (use `kubectl version`): 1.21.0
- OS (e.g. from `/etc/os-release`): KIND - ubuntu 18.04

`awscluster.spec.controlPlaneEndpoint.scheme `

is set as:

```
awscluster_types.go:147:   // Scheme sets the scheme of the load balancer (defaults to Internet-facing)
awscluster_types.go:148:   // +kubebuilder:default=Internet-facing
awscluster_types.go:149:   // +kubebuilder:validation:Enum=Internet-facing;internal
```

Note the `Internet-facing` with capital `I`

But, when you create an AWSCluster without providing `controlPlaneLoadBalancer` - await creation, then:

```
k get awscluster -oyaml | grep scheme
scheme: internet-facing
```

Now, this value is immutable (Why?) - but if you then go and attempt to add any other options under controlPlaneEndpoint, such as an additional SG for the control plane lb, you are faced with the validation error mentioned above.

Additionally, creating an AWSCluster with the following:

```
  controlPlaneLoadBalancer:
    scheme: Internet-facing
```

Will begin to fail if CAPA is restarted with the following error: 

```
ELB names must be unique within a region: \"capi-dev-apiserver\" ELB already exists in this region with a different scheme \"internet-facing\"
```

dlipovetsky:
**What type of PR is this?**

/kind bug

According to AWS documentation, the scheme is lowercase. If the field is
explicitly set to Internet-facing, CAPA fails to recognize the ELB it
has created, because the ELB's scheme is lowercase, i.e.,
internet-facing.

Fixes #2534

Ordinarily, changing the accepted value is a breaking API change. However, cluster create _always fails_ with the current value ( `Internet-facing`), therefore we can assume that no one is using the current value.

I added this PR to the agenda for our next meeting.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Changes the accepted ELB scheme from "Internet-facing" to "internet-facing." Although this is a breaking API change, it should not impact any user, because cluster create always fails with the previous value.
```

### DIFF
--- a/api/v1alpha3/awscluster_types.go
+++ b/api/v1alpha3/awscluster_types.go
@@ -146,9 +146,9 @@ type Bastion struct {
 
 // AWSLoadBalancerSpec defines the desired state of an AWS load balancer.
 type AWSLoadBalancerSpec struct {
-	// Scheme sets the scheme of the load balancer (defaults to Internet-facing)
-	// +kubebuilder:default=Internet-facing
-	// +kubebuilder:validation:Enum=Internet-facing;internal
+	// Scheme sets the scheme of the load balancer (defaults to internet-facing)
+	// +kubebuilder:default=internet-facing
+	// +kubebuilder:validation:Enum=internet-facing;internal
 	// +optional
 	Scheme *ClassicELBScheme `json:"scheme,omitempty"`
 
--- a/api/v1alpha3/types.go
+++ b/api/v1alpha3/types.go
@@ -89,7 +89,7 @@ type ClassicELBScheme string
 var (
 	// ClassicELBSchemeInternetFacing defines an internet-facing, publicly
 	// accessible AWS Classic ELB scheme.
-	ClassicELBSchemeInternetFacing = ClassicELBScheme("Internet-facing")
+	ClassicELBSchemeInternetFacing = ClassicELBScheme("internet-facing")
 
 	// ClassicELBSchemeInternal defines an internal-only facing
 	// load balancer internal to an ELB.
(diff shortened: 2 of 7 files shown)


sedefsavas:
Following up on the discussion during the office hours:

```
const (
	// LoadBalancerSchemeEnumInternetFacing is a LoadBalancerSchemeEnum enum value
	LoadBalancerSchemeEnumInternetFacing = "internet-facing"

	// LoadBalancerSchemeEnumInternal is a LoadBalancerSchemeEnum enum value
	LoadBalancerSchemeEnumInternal = "internal"
)
```

@dlipovetsky what you are saying is correct.
Since it was broken before, although this is an API change, we should fix it and make a release soon.

sedefsavas:
/approve
cc @richardcase @randomvariable

Issue #2801, opened 2021-09-24
Add defaulter & conversion for Internet-facing to internet-facing ELB scheme change


sedefsavas:
/kind bug

**What steps did you take and what happened:**
We fixed an issue about ELB scheme enum values https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2768, but forgot to add the conversion for the clusters.

We should fix this before doing a release with #2768 fix in.
Also, we need to backport it to v1alpha3 too.

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sedefsavas:
cc @dilyevsky

randomvariable:
I think we also need this in the defaulter so that it applies to existing v1alpha3 or existing v1alpha4 clusters.

randomvariable:
Discussing with @sedefsavas , I think what we want is:

v1alpha3 --> v1alpha3 & v1alpha4 --> v1alpha4, default on Update to corrected value
v1alpha3 --> v1alpha4 , v1alpha3 --> v1beta1, v1alpha4 --> v1beta1, correct in conversion

We may also want to call this particular defaulter inside the controller, such that it allows the ELB to complete reconciliation. Calling patch will then update with the corrected value:

sedefsavas:
Changes for the v1alpha4 release is described in the open PR: [see](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2807)


TODO:

**Changes to be added to the next v1alpha3 release:**
 No change in the API or webhooks, just use "Internet-facing" scheme and translate it to "internet-facing" internally in the controllers. This is enough to unblock existing users without any change to their objects.

**Changes to be added to the next v1beta1 release:**
`Internet-facing` to `internet-facing` modification will only happen in the conversion webhook and remove the logic that accepts `Internet-facing` scheme from the controllers. No need for defaulting.

dlipovetsky:
Thanks for catching this @sedefsavas!

(I missed this issue because of a tag typo above :grimacing: )

PR #2807, opened 2021-09-28
Fix for internet-facing ELB scheme

sedefsavas:
/kind bug
This PR
1. Adds `Internet-facing` to `internet-facing` modification to v1alpha3 to 4 conversion to support v1alpha3 clusters created with scheme set to Internet-facing.

2. Adds `Internet-facing` to `internet-facing` modification in the defaulting webhook for v1alpha4 clusters that was created with scheme set to `Internet-facing` (using v0.7.0 release). Adding this modification to defaulting webhook means that, we will continue to allow creating `AWSClusters` with `Internet-facing` scheme. But this has to be added to the defaulting to support updates on the existing v1alpha4 clusters: since the API has changed, during update saving the `Internet-facing` scheme is not possible.

3. Adds the logic to consider `Internet-facing` scheme as `internet-facing` in the controller for `AWSClusters` created by v0.7.0 and have not been modified. (Since they won't hit defaulting webhook, they will keep having `Internet-facing` as scheme, even though the API no longer supports it)

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2801

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
For v1alpha4 and upgraded v1alpha3 clusters, having "Internet-facing" scheme in the load balancer spec will work exactly as "internet-facing" scheme. In v1beta1, only "internet-facing" scheme will be used. 
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -7,7 +7,7 @@ replace sigs.k8s.io/cluster-api => sigs.k8s.io/cluster-api v0.4.1
 require (
 	github.com/apparentlymart/go-cidr v1.1.0
 	github.com/aws/amazon-vpc-cni-k8s v1.9.1
-	github.com/aws/aws-lambda-go v1.26.0
+	github.com/aws/aws-lambda-go v1.27.0
 	github.com/aws/aws-sdk-go v1.40.33
 	github.com/awslabs/goformation/v4 v4.19.5
 	github.com/blang/semver v3.5.1+incompatible
--- a/go.sum
+++ b/go.sum
@@ -114,8 +114,8 @@ github.com/asaskevich/govalidator v0.0.0-20180720115003-f9ffefc3facf/go.mod h1:l
 github.com/asaskevich/govalidator v0.0.0-20190424111038-f61b66f89f4a/go.mod h1:lB+ZfQJz7igIIfQNfa7Ml4HSf2uFQQRzpGGRXenZAgY=
 github.com/aws/amazon-vpc-cni-k8s v1.9.1 h1:nUvjYFCQCoNu/qP0equnyrtdb34mIQ8KPJm2jk9XY9s=
 github.com/aws/amazon-vpc-cni-k8s v1.9.1/go.mod h1:OLzS+K66W2hyykBZcHyHEodeK0HaBlSbULgICBJ2tZY=
-github.com/aws/aws-lambda-go v1.26.0 h1:6ujqBpYF7tdZcBvPIccs98SpeGfrt/UOVEiexfNIdHA=
-github.com/aws/aws-lambda-go v1.26.0/go.mod h1:jJmlefzPfGnckuHdXX7/80O3BvUUi12XOkbv4w9SGLU=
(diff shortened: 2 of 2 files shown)


richardcase:
Apart from the conflict this:

/lgtm

api/v1alpha3/awscluster_types.go
```diff
@@ -146,9 +146,9 @@ type Bastion struct {
 
 // AWSLoadBalancerSpec defines the desired state of an AWS load balancer.
 type AWSLoadBalancerSpec struct {
-	// Scheme sets the scheme of the load balancer (defaults to internet-facing)
-	// +kubebuilder:default=internet-facing
-	// +kubebuilder:validation:Enum=internet-facing;internal
+	// Scheme sets the scheme of the load balancer (defaults to Internet-facing)
+	// +kubebuilder:default=Internet-facing
+	// +kubebuilder:validation:Enum=Internet-facing;internal
```
randomvariable, line 151:
```suggestion
	// +kubebuilder:default=internet-facing
	// +kubebuilder:validation:Enum=internet-facing;Internet-facing;internal
```

Or have I misunderstood the problem entirely?

api/v1alpha4/types.go
```diff
@@ -105,6 +105,9 @@ var (
 	// accessible AWS Classic ELB scheme.
 	ClassicELBSchemeInternetFacing = ClassicELBScheme("internet-facing")
 
+	// ClassicELBSchemeIncorrectInternetFacing was inaccurately used to define an internet-facing LB in v0.6 releases > v0.6.6 and v0.7.0 release
```
randomvariable, line 108:
```suggestion
	// [Deprecated] ClassicELBSchemeIncorrectInternetFacing was inaccurately used to define an internet-facing LB in v0.6 releases > v0.6.6 and v0.7.0 release
```

Might as well do an immediate deprecation.

randomvariable:
/priority critical-urgent
/area api

/lgtm 

too, pending rebase.

api/v1alpha3/awscluster_types.go
randomvariable, line 151:
```suggestion
	// Scheme sets the scheme of the load balancer (defaults to Internet-facing)
	// +kubebuilder:default=internet-facing
	// +kubebuilder:validation:Enum=internet-facing;Internet-facing;internal
```

Also thinking about the backport here, and I think this is what we'll be accepting?

randomvariable: We also want to fix the v1alpha3 branch though, which means allowing lower case `internet-facing`. Let's discuss this when you're around.

randomvariable:
/hold

After chatting to @voor  about desired behaviour, there's some changes to make. We want to persist the correct values back so will handle that here as well. Will follow up as Sedef is on PTO.

sedefsavas:
Pushed an empty commit accidentally. I don't know why it shows merged because of this.

Issue #2795, opened 2021-09-24
If AWSCluster managed security group ingress rule is renamed, this causes AWSCluster to fail


randomvariable:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

If someone tries to manually replace an ingress rule with the same port but different description, AWSCluster will fail to reconcile the security group:

![image](https://user-images.githubusercontent.com/1441070/134682835-6ddb2200-cb27-4d82-8cb8-7860aaf563bc.png) 

with the following error:

```
ingress rules: [protocol=tcp/range=[20449-20449]/description= protocol=tcp/range=[22-22]/description=SSH protocol=tcp/range=[20049-20049]/description= protocol=tcp/range=[2049-2049]/description=]: InvalidPermission.NotFound: The specified rule does not exist in this security group.\n\tstatus code: 400, request id: 85392b86-a8fc-493e-b678-5d5594340200" "controller"="awscluster" "name"="test" "namespace"="default"
```

This then triggers #2794 

**What did you expect to happen:**

Rules revert back to their original state correctly.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**
Cluster API Provider AWS: v0.6.4

Issue #2798, opened 2021-09-24
Allow customisation of node rules


randomvariable:
/kind feature

**Describe the solution you'd like**

One of the most common requests we have is to allow customisation of the node security group. In particular, users want to often change the SSH ingress rule when bastion is enabled or not, and either disable or change the node port ruleset.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: v0.7.0
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`): 

/area api
/area networking
/priority important-soon
/milestone v0.8.0

sedefsavas:
Closing in favor of https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/392

PR #2813, opened 2021-09-30
Fix reconciliation of AWS ingress rules of same port

pydctw:
**What type of PR is this?**
/kind bug

Fix `ingressRulesFromSDKType` function.
Changed the logic to generate separate Ingress rules for userIdGroupPairs and IPRanges from EC2 IpPermission API Response.
An IngressRule cannot have SourceSecurityGroupIDs and CidrBlocks at the same time and combining them into a single IngressRule was the cause of the reported bug.

Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2795

There are three commits in this PR, which I plan to keep separate.
* The first commit refactors duplicate codes.
* The second commit back-fills unit tests for Difference function.
* The third commit is an actual fix for the issue.

**Checklist**:
- [ ] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Fix reconciliation of AWS ingress rules of same port
```

### DIFF
--- a/api/v1alpha4/types_test.go
+++ b/api/v1alpha4/types_test.go
@@ -1,0 +1,107 @@
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha4
+
+import (
+	"testing"
+
+	. "github.com/onsi/gomega"
+)
+
+func TestSG_Difference(t *testing.T) {
+	tests := []struct {
+		name     string
+		self     IngressRules
+		input    IngressRules
+		expected IngressRules
+	}{
+		{
+			name:     "self and input are nil",
+			self:     nil,
+			input:    nil,
+			expected: nil,
+		},
+		{
+			name: "input is nil",
+			self: IngressRules{
+				{
+					Description:            "SSH",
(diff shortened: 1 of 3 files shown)


pydctw:
Tested the changes in actual AWS cluster deployment.

#### Before the change (the bug)
The rule to revoke has cidrBlocks and sourceSecurityGroupIds combined, which results in `The specified rule does not exist in this security group` error response from AWS API. Please see the issue for the example.
```
{"description":"SSH","protocol":"tcp","fromPort":22,"toPort":22,"cidrBlocks":["0.0.0.0/0"],"sourceSecurityGroupIds":["sg-02a13358455ea8b96"]}
```

#### After the change
The rule to revoke only has a user-created portion of SSH rule.
```
[{"description":"MY-SSH","protocol":"tcp","fromPort":22,"toPort":22,"cidrBlocks":["0.0.0.0/0"]}]
```

Observed that this rule is successfully revoked.

randomvariable:
Looks fine, should we add some commentary for future archaeologists as to the absolutely maddening AWS API?

randomvariable:
This logic works, but wouldn't on the next reconciliation these then revert to what's specified. I think it'll be better for AWSCluster to be authoritative for rules, and that's always been our support statement, and instead allow #2798 for node port rule customisation.

Users have also been instructed to add their own additionalSecurityGroups with their own rules. Supporting customisation via the EC2 console maybe introducing more uncertainty.

pydctw:
> This logic works, but wouldn't on the next reconciliation these then revert to what's specified in AWSCluster? I think it'll be better for AWSCluster to be authoritative for rules, and that's always been our support statement, and instead allow #2798 for node port rule customization.

You are correct that the rules will revert to what's specified in AWSCluster. I thought that's the behavior we want for this issue. 

I think #2798, allowing node rule customization, can be addressed separately but let's sync up next week.

pydctw:
@randomvariable, I updated https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2813#issuecomment-931643192.

randomvariable:
I'll forward port this to v1beta1.

randomvariable:
Fixed #2795

PR #2866, opened 2021-10-21
[WIP]Add AWSCluster as ownerRef and finalizer to identityRef to block deletion of identityRef resources

### LINKED ISSUE #2371: AWSClusterRoleIdentity can be deleted when its still referenced
/kind bug

When deleting a cluste, which is using the new AWS multitenancy (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2253) CR `AWSClusterRoleIdentity` referenced in `AWSCuster.spec.identityRef`, you can accidentally delete the Role resource and the whole deletion process will fail, because the controller doesn't know how to access the AWS API.

`AWSClusterRoleIdentity` would have `finalizer` set on to prevent deletion ( or `ownerReferences:` to block deletion), until there is no identity reference on it (no AWSCluster is pointing to it)

I have not tested this with `AWSClusterStaticIdentity` and `AWSClusterControllerIdentity` but there could be similar behavior

**Environment:**

- Cluster-api-provider-aws version: 0.6.5
- Kubernetes version: (use `kubectl version`): 1.19.8

Ankitasw:
**What type of PR is this?**
/kind bug

This PR adds the ownerReferences and finalizers to cluster IdentityRefs `AWSClusterStaticIdentity`, `AWSClusterControllerIdentity` and `AWSClusterRoleIdentity` to block accidental deletion of identityRef resources before cluster deletion. This PR also adds controllers for `AWSClusterStaticIdentity` and `AWSClusterRoleIdentity` to accommodate `updateIdentity in AWSCluster` use-case.

Fixes #2371 

**Checklist**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
- Add ownerReferences and finalizer to cluster IdentityRefs to block deletion of identityRef resources before cluster deletion
- Add controllers for `AWSClusterStaticIdentity` and `AWSClusterRoleIdentity`
```


sedefsavas:
I might be missing something, but we need to add ownerReferences to identity objects, not to the secrets it is referred to.
1- AWSClusters that use an identity will add themselves to identities as owner reference. 
2- During deletion of an AWSCluster, AWSCluster will move itself from the identity's owner ref list.
3- Identity will add a finalizer and block deletion if there is any AWSCluster as owner reference.
4- Also, we need to add tests for these.

controllers/awscluster_controller.go
```diff
@@ -413,3 +428,123 @@ func (r *AWSClusterReconciler) requeueAWSClusterForUnpausedCluster(ctx context.C
 		}
 	}
 }
+
+func (r *AWSClusterReconciler) deleteAWSClusterFromIdentityOwnerRef(ctx context.Context, clusterScope *scope.ClusterScope) error {
+	identityObjectKey := client.ObjectKey{Name: clusterScope.IdentityRef().Name}
```
sedefsavas, line 433: IdentityRef is an optional field, nil check is needed here.

controllers/awscluster_controller.go
```diff
@@ -413,3 +428,123 @@ func (r *AWSClusterReconciler) requeueAWSClusterForUnpausedCluster(ctx context.C
...
+	hasOwnerRef := false
+	switch clusterScope.IdentityRef().Kind {
```
sedefsavas, line 435: We don't need to know the kind of the identity here, just need to reach to ownerReferences. Unstructured struct could be used here.

sedefsavas:
Is this PR ready for another review? I see that a new version is pushed.

Ankitasw:
> Is this PR ready for another review? I see that a new version is pushed.

@sedefsavas the review comments have been taken care of, only unit tests for this scenario is pending, that why it's still WIP.

controllers/awscluster_controller.go
sedefsavas: @Ankitasw I think you missed this comment. Is this doable?

Ankitasw: @sedefsavas totally forgot about this, will do it, thanks

Ankitasw: @sedefsavas it's ready now(still need to add unit tests which might take time)

Ankitasw: @sedefsavas I have added a scenario in unit test to check if finalizer and owner ref is updated properly. Do you think we need to cover any other scenario, not sure how to test in case of reconcileDelete() as after deleting cluster we can't check the identity ref?

Ankitasw:
E2E tests are passing https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2866/pull-cluster-api-provider-aws-e2e/1460505083463602176

sedefsavas:
@Ankitasw Wondering if you tested this scenario manually:

- Creating multiple clusters that use the same identity
- Deleting the identity 
- Confirming that it is there
- Deleting one of the clusters and confirming that the identity isn't deleted
- Deleting last cluster and confirming the identity is gone.

Ankitasw:
> @Ankitasw Wondering if you tested this scenario manually:
> 
> * Creating multiple clusters that use the same identity
> * Deleting the identity and confirming that finalizer is there and deletion timestamp is set
> * Deleting one of the clusters and confirming that the identity isn't deleted
> * Deleting last cluster and confirming the identity is gone.

@sedefsavas I didn't test these scenarios manually, will do it

controllers/awscluster_controller.go
```diff
@@ -194,26 +232,37 @@ func reconcileDelete(clusterScope *scope.ClusterScope) (reconcile.Result, error)
 	// Cluster is deleted so remove the finalizer.
 	controllerutil.RemoveFinalizer(clusterScope.AWSCluster, infrav1.ClusterFinalizer)
```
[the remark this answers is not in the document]
Ankitasw: How can we find out these unrecoverable errors(like if the ownerRef is deleted manually)?

Ankitasw:
@sedefsavas I tested the scenarios listed by you. Findings:

controllers/awscluster_controller.go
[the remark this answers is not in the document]
Ankitasw: I think it's difficult to track manual intervention, anybody can do kubectl edit and delete the finalizers and ownerRef without getting tracked, we can probably provide a warning somewhere to avoid such interventions.

randomvariable:
> Deleting all the clusters deletes the identityRef (as expected)

We don't want this behaviour.

Ankitasw:
> We don't want this behaviour. Deletion of an identity resource should be explicit.

@randomvariable thanks for pointing this out.
Correction to what i stated earlier:
Deleting all the clusters deletes the ownerRef and finalizer from identityRef, while the identityRef remains in the system (as expected)

sedefsavas:
> > Deleting all the clusters deletes the identityRef (as expected)
> 
> 
> 
> We don't want this behaviour. Deletion of an identity resource should be explicit.

The test I described above is to check when the identitiy is deleted by the user but has ownerrefs, does the deletion get completed once the owner clusters are gone? So identity should be gone at the end of the manual test. So @Ankitasw if the identity is still there as you mentioned that is a problem.

Ankitasw:
@sedefsavas sorry I misunderstood it, if at first we delete the identityRef when ownerRef is there, deletion timestamp gets attached and after cluster gets deleted, the identityRef deletion completes successfully.
What I described above is the scenario where we delete all the clusters, then identityRef remains as is, and all the ownerRefs and finalizer is deleted from identityRef.
(got a bit confused at first :) )

Ankitasw:
@randomvariable regarding what we discussed about [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2866#issuecomment-948880693) comment, still having doubts while I started implementing: 

1. suppose we have a cluster `cluster-1` and we have attached it to `identityRef-1`
2. now in update case, we update the `cluster-1` and attach it to `identityRef-2`
3. now we don't have reference to `identityRef-1`, and since we are updating the ownerRef in `identityRef-1` in 1st step, we dont have access to identityRef-1 after the update, so how do we delete ownerRef from `identityRef-1`?

Ankitasw:
@randomvariable regarding the [comment](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2866#issuecomment-948880693) which we discussed, I have some open points while implementing it:

1. Suppose we have created `cluster-1` with `identityRef-1`
2. We update the cluster by updating identityRef to `identityRef-2 `
3. Since we add `cluster-1` as ownerRef of `identityRef-1` in 1st step, after updating the cluster, we dont have access to `identityRef-1` , so how do we keep track of the old `identityRef-1` to remove `cluster-1` from its ownerRef

sedefsavas:
> I would also add, that we implicitly support the change of the identityRef as well as per the original proposal, so to support that we need to track that change and update ownerRefs as well.

This is a critical comment. One thing we can do is to create identity controllers and do a sanity check there. 
 -> Go over all ownerRefs and check if the identity is still referred by the cluster. If not, delete the owner ref.

controllers/awscluster_controller.go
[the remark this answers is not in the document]
sedefsavas: We should proceed on notFound error.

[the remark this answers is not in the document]
Ankitasw: I have pushed a new change to ignore deletion of finalizer and ownerRef from identity to delete ClusterFinalizer, please have a look.

Ankitasw:
@sedefsavas reminder for review

controllers/awsclusterstatic_identity_controller.go
```diff
@@ -0,0 +1,96 @@
...
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			return reconcile.Result{}, err
+		}
+		return reconcile.Result{}, nil
+	}
+
+	// Fetch all static identities
+	staticIdentities := &infrav1.AWSClusterStaticIdentityList{}
+	err = r.List(ctx, staticIdentities)
```
sedefsavas, line 61: After any changes (and periodically), `AWSClusterStaticIdentity` object will start reconciling. We don't need to list all identities here.
Similar to other controllers, get the object from the request and only do the checks on that particular resource.

controllers/awscluster_controller.go
```diff
@@ -66,9 +77,41 @@ var (
...
+	ec2ServiceFactory     func(scope.EC2Scope) services.EC2MachineInterface
+	networkServiceFactory func(scope.ClusterScope) services.NetworkInterface
+	elbServiceFactory     func(scope.ClusterScope) services.ELBClusterInterface
+	securityGroupFactory  func(scope.ClusterScope) services.SecurityGroupInterface
+	Endpoints             []scope.ServiceEndpoint
+	WatchFilterValue      string
+}
+
+func (r *AWSClusterReconciler) getEC2Service(scope scope.EC2Scope) services.EC2MachineInterface {
+	if r.ec2ServiceFactory != nil {
```
sedefsavas, line 90: It'd be good to have comments for all these factories to make it clear that these are added for enabling testing.

controllers/awscluster_controller.go
```diff
@@ -190,28 +234,39 @@ func (r *AWSClusterReconciler) reconcileDelete(clusterScope *scope.ClusterScope)
...
+		return reconcile.Result{}, err
+	}
 	// Cluster is deleted so remove the finalizer.
 	controllerutil.RemoveFinalizer(clusterScope.AWSCluster, infrav1.ClusterFinalizer)
 
 	return reconcile.Result{}, nil
 }
 
-func (r *AWSClusterReconciler) reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error) {
+// TODO(ncdc): should this be a function on ClusterScope?
```
sedefsavas, line 247:
```suggestion
```

controllers/awscluster_controller.go
```diff
@@ -411,3 +466,101 @@ func (r *AWSClusterReconciler) requeueAWSClusterForUnpausedCluster(ctx context.C
...
+		if err != nil {
+			return err
+		}
+		ownerReferences := unstructuredIdentityObject.GetOwnerReferences()
+		ownerReferences = util.EnsureOwnerRef(ownerReferences, metav1.OwnerReference{
+			APIVersion:         infrav1.GroupVersion.String(),
+			Kind:               awsClusterKind,
+			Name:               clusterScope.InfraClusterName(),
+			UID:                clusterScope.InfraCluster().GetUID(),
+			BlockOwnerDeletion: aws.Bool(true),
```
sedefsavas, line 514: We should not block the deletion of AWSCluster when the identity is not deleted.
The definition of BlockOwnerDeletion is:
// If true, AND if the owner has the "foregroundDeletion" finalizer, then
	// the owner cannot be deleted from the key-value store until this
	// reference is removed.

controllers/awsclusterstatic_identity_controller.go
Ankitasw: The problem with that approach is we will be able to get the identity object, but we wont be able to get the cluster object from request, as identities are cluster scoped, there wont be any namespaces

sedefsavas (commented):
Also, is there a reason why we don't need AWSClusterControllerIdentity controller?

api/v1beta1/awsclusterstaticidentity_webhook.go
```diff
@@ -65,16 +65,11 @@ func (r *AWSClusterStaticIdentity) ValidateDelete() error {
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
 func (r *AWSClusterStaticIdentity) ValidateUpdate(old runtime.Object) error {
-	oldP, ok := old.(*AWSClusterStaticIdentity)
+	_, ok := old.(*AWSClusterStaticIdentity)
 	if !ok {
 		return apierrors.NewBadRequest(fmt.Sprintf("expected an AWSClusterStaticIdentity but got a %T", old))
 	}
 
-	if oldP.Spec.SecretRef != r.Spec.SecretRef {
```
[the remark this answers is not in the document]
Ankitasw: in update case, if we change the identity, this check will fail

controllers/awscluster_controller.go
Ankitasw: this is the add ownerRef case, so when we add the ownerRef shall we not set `BlockOwnerDeletion` to true

Ankitasw:
> Also, is there a reason why we don't need AWSClusterControllerIdentity controller?

there already is a controller for [`AWSClusterControllerIdentity`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1b304535450392b7762558768efbb8e7672f6713/exp/controlleridentitycreator/awscontrolleridentity_controller.go), just wanted to check if we should modify that controller only for adding ownerRef and finalizer?

controllers/awscluster_controller.go
Ankitasw: done

controllers/awscluster_controller.go
Ankitasw: done

controllers/awscluster_controller.go
Ankitasw: so in what case we should set this? or is it not required to set it in our case?

controllers/awsclusterstatic_identity_controller.go
sedefsavas: Right, I remember now. Cluster-scoped objects should not have namespace scoped owner references normally. 

I wonder if this becomes a problem in terms of Garbage collection.
Since cluster-scoped identity will have owner refs that is not cluster-scoped, garbage collector will not be able to find the owners and tries to delete the object? Not sure, but since you did tests I am guessing something is blocking this to happen, probably finalizer. 
WDYT is happening?

controllers/awscluster_controller.go
sedefsavas: For example MachineDeployment is owner of MachineSet and should not be deleted before the MachineSet.

For our case, deletion of the Cluster should not be affected by the identities. Actually, there is no owner relation between identities and clusters so thinking of a better way to do this.

sedefsavas:
> > Also, is there a reason why we don't need AWSClusterControllerIdentity controller?
> 
> there already is a controller for [`AWSClusterControllerIdentity`](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1b304535450392b7762558768efbb8e7672f6713/exp/controlleridentitycreator/awscontrolleridentity_controller.go), just wanted to check if we should modify that controller only for adding ownerRef and finalizer?

That is an experimental feature, we cannot rely on that.

api/v1beta1/awsclusterstaticidentity_webhook.go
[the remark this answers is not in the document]
sedefsavas: I might be missing something but this checks if there is any change on the static identity itself. How does this block changing identity in AWSCluster?

controllers/awsclusterstatic_identity_controller.go
sedefsavas: This is a relatively new change, which makes the identities with absent owners not to be garbage collected:

```
Cluster-scoped dependents can only specify cluster-scoped owners. In v1.20+, if a cluster-scoped dependent specifies a namespaced kind as an owner, it is treated as having an unresolvable owner reference, and is not able to be garbage collected.
```

Ankitasw: After brainstorming a bit, I found that we can use only annotations or addition of new field to the identity object (like list of owners) to make sure deletion goes through properly.
In any case, we can't use ownerReferences if the owner and dependent are not in same scope.

api/v1beta1/awsclusterstaticidentity_webhook.go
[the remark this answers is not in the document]
Ankitasw: oldP.Spec.SecretRef will point to the old identity secretRef, but after update case, r.Spec.SecretRef will point to the new identity, hence these two wont match and it will fail this check.

richardcase:
@Ankitasw - apart from the rebase is there much more to complete on this?

Ankitasw:
Yes @richardcase only rebase is remaining, will do it once i am back from PTO.

sedefsavas:
> Yes @richardcase only rebase is remaining, will do it once i am back from PTO.

I remember there was a problem of assigning a namespaced resource as an owner ref of cluster scoped resource in this PR. And this is not supposed to be the case:

`Cluster-scoped dependents can only specify cluster-scoped owners.`
https://kubernetes.io/docs/concepts/overview/working-with-objects/owners-dependents/

If this is still the case, let's close this PR and reopen once if/when the issue is solved.

Ankitasw:
Yes @sedefsavas i misread this as another resource ref PR, this had the problem of cross namespace reference, we can close this as of now.
/close

PR #2892, opened 2021-10-28
build(deps): Remove indirect dependencies in go.mod

PushkarJ:
**What type of PR is this?**

/kind cleanup

Indirect deps were added in the past, to ensure that go build picked up latest versions for transitive dependencies that had fixes for CVEs

Since then the indirect dependencies are upgraded in the direct dependencies go.mod such that no vulnerabilities should show up for build time dependencies

This PR removes these indirect dependencies to make the go.mod simpler

- Slack discussion: https://kubernetes.slack.com/archives/C022K4F2W4W/p1635287763001800?thread_ts=1634678681.026600&cid=C022K4F2W4W
- Not sure if release notes are needed here, but happy to add it in any format that is preferred

**Checklist**:

- [X] squashed commits

**Release note**:

```release-note
NONE
```


shivi28:
Thanks @PushkarJ for the PR. I think we can do similar thing in CAPI and other providers as well. Right ?

PushkarJ:
> Thanks @PushkarJ for the PR. I think we can do similar thing in CAPI and other providers as well. Right ?

@shivi28: Tried this for `cluster-api` repo and they get added right back after I do `go mod tidy`. The best explanation (may not be the right one) I found for this is here: https://golangbyexample.com/direct-indirect-dependency-module-go/

PR #2898, opened 2021-11-01
Always add controllers role to IAMAuthenticator config

codablock:
**What type of PR is this?**
/kind bug

This ensures that access to clusters is never lost by CAPA controllers,
which might have happened in different situations, e.g. when creating
a management cluster from a temporary bootstrap cluster and then moving
the management cluster to an EKS managed cluster (probably to itself).

In that case, the initial bootstrapping of the EKS cluster used the users
IAM identity to create the cluster, which gives it admin access to the
whole cluster. When the cluster is then moved to itself, and at the same
time the creator forgot to add the controllers role to the clusters
"iamAuthenticatorConfig", access to the cluster is lost by the controller.

From that point on, it can not generate valid admin kubeconfigs anymore
and also fails to update resources on the cluster (e.g. via CRS) from that
point on. Also, IAM authenticator reconcilation itself will fail as this
requires a valid kubeconfig.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Always add controllers role to IAMAuthenticator config
```

### DIFF
--- a/pkg/cloud/services/iamauth/reconcile.go
+++ b/pkg/cloud/services/iamauth/reconcile.go
@@ -48,9 +48,9 @@ func (s *Service) ReconcileIAMAuthenticator(ctx context.Context) error {
 		return fmt.Errorf("getting aws-iam-authenticator backend: %w", err)
 	}
 
-	roleARN := fmt.Sprintf("arn:aws:iam::%s:role/nodes%s", accountID, iamv1.DefaultNameSuffix)
+	nodesRoleARN := fmt.Sprintf("arn:aws:iam::%s:role/nodes%s", accountID, iamv1.DefaultNameSuffix)
 	nodesRoleMapping := ekscontrolplanev1.RoleMapping{
-		RoleARN: roleARN,
+		RoleARN: nodesRoleARN,
 		KubernetesMapping: ekscontrolplanev1.KubernetesMapping{
 			UserName: EC2NodeUserName,
 			Groups:   NodeGroups,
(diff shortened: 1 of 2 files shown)


randomvariable:
seems about right to me.

pkg/cloud/services/iamauth/reconcile.go
```diff
@@ -61,6 +61,19 @@ func (s *Service) ReconcileIAMAuthenticator(ctx context.Context) error {
 		return fmt.Errorf("mapping iam node role: %w", err)
 	}
 
+	controllersRoleARN := fmt.Sprintf("arn:aws:iam::%s:role/controllers%s", accountID, iamv1.DefaultNameSuffix)
```
richardcase, line 64: Looking at this has made me wonder how handle the option of bootstrapping with a [prefix](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d41f8cc91c992585fa4e1e0d693d408c89dbfb1d/cmd/clusterawsadm/cloudformation/bootstrap/template.go#L71) on the role names. But thats out of scope of this change.

richardcase:
This seems good to me. 

Prior to this change we relied on the behaviour of EKS in granting admin access to the creator of the cluster which requiring explicit modification of aws-auth...the creator being the encoded credentials via the encoded credentials /  identity types. But we didn't consider IRSA/kiam/kube2iam in the original implementation (i don't think IRSA was even a thing then).

However, i'm undecided if it constitutes a change in behaviour (i don't think it does) or if it will impact any assumptions' people currently have when creating EKS clusters with CAPA. For example, do users assume that just the nodes mapping is there and have built tooling/processes around this and will adding a new entry automatically break something? Potentially.

If we think it will impact users assumptions maybe we introduce this via an option you can set to add the mapping and then in a later version make it default. 

Perhaps i'm other thinking.

What are you thoughts @codablock @randomvariable @sedefsavas?

Temporarily to allow discussion:

/hold

richardcase:
@sedefsavas - it would be good to chat about this one at the office hours.

PR #2917, opened 2021-11-05
Test flavors small improvements

invidian:
**What type of PR is this?**

/kind cleanup

Just some small stuff found along the way.

**Release note**:
```release-note
NONE
```


randomvariable:
Sure it's fine, but has this been tested on both MacOS and Linux?

invidian:
>Sure it's fine, but has this been tested on both MacOS and Linux? (I only have Linux to test with)

Tested on MacOS 12.0.1 with `grep -v -E '(foo|bar)'` and it works fine if that's what you mean.

sedefsavas:
Works for MacOs.
/lgtm
/approve

Issue #2936, opened 2021-11-10
Convert identity management into a library


sedefsavas:
We reconcile CAPA resources by creating sessions based on the identity reference provided in their corresponding `AWSCluster`.

For creating those sessions, we use some helper functions for creating an AWS SDK identity provider based on the CAPA identity resource. These helper functions would be useful for external providers that reconcile CAPA resources and want to support multi-tenancy.

This PR is to track exposing that logic as a library.

PR #2937, opened 2021-11-10
Refactor identity management

sedefsavas:
/kind cleanup

This PR does the initial refactoring for https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2936

Before, `buildProvidersForRef()` was returning an array of providers but this never is the case as for each identity there must be exactly 1 provider. This is not a bug because in the logic, we were never ending up returning multiple providers.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/api/v1beta1/awsidentity_types.go
+++ b/api/v1beta1/awsidentity_types.go
@@ -148,7 +148,7 @@ type AWSClusterRoleIdentitySpec struct {
 
 	// SourceIdentityRef is a reference to another identity which will be chained to do
 	// role assumption. All identity types are accepted.
-	SourceIdentityRef *AWSIdentityReference `json:"sourceIdentityRef,omitempty"`
+	SourceIdentityRef *AWSIdentityReference `json:"sourceIdentityRef"`
 }
 
 // +kubebuilder:object:root=true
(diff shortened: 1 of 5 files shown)


sedefsavas:
Since this is just refactoring, let's merge this after we make a release this week (~Nov. 18).
/hold

api/v1beta1/awsidentity_types.go
```diff
@@ -148,7 +148,7 @@ type AWSClusterRoleIdentitySpec struct {
 
 	// SourceIdentityRef is a reference to another identity which will be chained to do
 	// role assumption. All identity types are accepted.
-	SourceIdentityRef *AWSIdentityReference `json:"sourceIdentityRef,omitempty"`
+	SourceIdentityRef *AWSIdentityReference `json:"sourceIdentityRef"`
```
richardcase, line 151: Is this still an optional field?

richardcase: (i'm guessing it is so might be worth adding `+optional`)

Issue #2915, opened 2021-11-05
Creating a new ec2 instance for the same AWSMachine


sedefsavas:
/kind bug

**What steps did you take and what happened:**

In a fully functional v1alpha3 cluster, while a healthy ec2 instance exists, a second instance is created by the AWSMachine controller for the same `AWSMachine` due to unknown reasons. 

During AWSMachine `reconcileNormal()`, we check if the instance exists; if not, a new instance is created: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1a61dfc39649749b8b8dfad8334909a388d675a7/controllers/awsmachine_controller.go#L489
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1a61dfc39649749b8b8dfad8334909a388d675a7/controllers/awsmachine_controller.go#L504-L505

In this particular case, providerID on the AWSMachine was set, hence findInstance() called InstanceIfExists() method:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1a61dfc39649749b8b8dfad8334909a388d675a7/pkg/cloud/services/ec2/instances.go#L79

The only case where a new instance is created is when `InstanceIfExists()` returns nil instance and nil error.
InstanceIfExists() calls `DescribeInstances` and two things might have happened for returning nil instance and nil error.:
1) DescribeInstances() returned a 404 error for an existing ec2 instance
2) DescribeInstances() returned 0 instances for an existing ec2 instance

I could not pinpoint what caused this but we should not be creating a new ec2 instance when ProviderID is set in `AWSMachine`, as it does not propagate to `Machine` object. Remediation is either handled in KCP or with MHC.

dlipovetsky:
Thanks for filing this! I have seen at least two internal reports of what appears to be this issue. The reports are for CAPA v0.7.0. If we can create a failing test for this issue, let's check if it also fails on the release-0.7 branch.

dlipovetsky:
The EC2 API is eventually consistent: One server might fulfill the RunInstances API call. Later another server, without the latest information on the set of created instances, might fulfill the DescribeInstances call, and return no matching instances.

dlipovetsky:
To address the issue of eventual consistency, we should make our RunInstances API calls idempotent.

We'll need to create a client token (64 character string), store it with the AWSMachine, and use that token every time we call RunInstances for that AWSMachine.

[...]

PR #2957, opened 2021-11-16
Return error if providerID is set and instance cannot be found

sedefsavas:
/kind bug

If providerID is set for an `AWSMachine`, we should never create a new instance for it again. Because the new instance id does not propagate to `Machine.status.nodeRef`.

This PR has some reorg for better readability. The main change is to return error (instead of nil instance and nil error), when the instance could not be found in `InstanceIfExists()`.

Partially fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2915, more investigation needed.

```release-note
Do not create a new instance if AWSMachine already has ProviderID field set.
```

### DIFF
--- a/pkg/cloud/services/ec2/instances_test.go
+++ b/pkg/cloud/services/ec2/instances_test.go
@@ -61,8 +61,8 @@ func TestInstanceIfExists(t *testing.T) {
 					Return(nil, awserrors.NewNotFound("not found"))
 			},
 			check: func(instance *infrav1.Instance, err error) {
-				if err != nil {
-					t.Fatalf("did not expect error: %v", err)
+				if err == nil {
+					t.Fatalf("expects error when instance could not be found: %v", err)
 				}
 
 				if instance != nil {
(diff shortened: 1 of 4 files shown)


pkg/cloud/services/ec2/instances_test.go
```diff
@@ -61,8 +61,8 @@ func TestInstanceIfExists(t *testing.T) {
 					Return(nil, awserrors.NewNotFound("not found"))
 			},
 			check: func(instance *infrav1.Instance, err error) {
-				if err != nil {
-					t.Fatalf("did not expect error: %v", err)
+				if err == nil {
```
sedefsavas, line 64: If instance could not be found when the provider id is set, returns `ErrInstanceNotFoundByID` error. So that during reconcileNormal(), we don't create a new instance. In `reconcileDelete()`, when ErrInstanceNotFoundByID is seen, deletion continues to clean up even if the instance is gone (may be manually deleted).

sedefsavas:
@randomvariable  @dlipovetsky 
Is there any scenario I am missing where we need to create a new instance even if the provider id is already set?

richardcase:
This seems good to me. As discussed on slack the only comment is that we should consider a follow-up PR to add an e2e test that covers the scenario where we delete an EC2 instance to make sure it isn't created. But like you said this will result in a long running test as the instance stay in a terminated state for 1 hour. So:

/lgtm

sedefsavas:
@dlipovetsky Do you have any comments regarding this PR? I think this is a good start blocking the controller to create a second instance. To merge it today before the release, any feedback is appreciated.

randomvariable:
/approve
/hold 
in case @dlipovetsky has any comments.

PR #2914, opened 2021-11-04
fix: multiple addons not being reconciled

richardcase:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

When the EKS addons functionality was added to CAPA there was only 1
addon available (vpc) and so it was never tested with multiple addons.
Now there are multiple addons there are issues.

The unit and e2e tests only tested 1 addons. On investigation the issue
was caused by implicit memory aliasing in a loop. This change fixes the
loop and also expands the unit and e2e tests to include multiple eks
addons

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2911 

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Fixes an issue with using multiple EKS addons.
```

Issue #2911, opened 2021-11-04
AWSManagedControlPlane can only install one EKS Addon at a time


Jacobious52:
/kind bug

It looks like the Add-on manager has a bug where it will only install the last add-on you specific in the config spec of a AWSManagedControlPlane. This means you cannot use more than one EKS Addon at a time currently.

**What steps did you take and what happened:**

- Create an AWSManagedControlPlane plan with all 3 EKS addons
```yaml
  addons:
  - name: "kube-proxy"
    version: "v1.20.4-eksbuild.2"
    conflictResolution: "overwrite"
  - name: "vpc-cni"
    version: "v1.8.0-eksbuild.1"
    conflictResolution: "overwrite"
  - name: "coredns"
    version: "v1.8.3-eksbuild.1"
    conflictResolution: "overwrite"
```
- Wait for reconcile. Check EKS console for add ons
<img width="1322" alt="image" src="https://user-images.githubusercontent.com/12645162/140268717-fb239563-bb23-4d5e-aad7-cd24b7e6c4b4.png">
- Note how only one is installed.

- Reorder the same add ons such that a different one is at the bottom. In this case `vpc-cni`

```yaml
  - name: "kube-proxy"
    version: "v1.20.4-eksbuild.2"
    conflictResolution: "overwrite"
  - name: "coredns"
    version: "v1.8.3-eksbuild.1"
    conflictResolution: "overwrite"
  - name: "vpc-cni"
    version: "v1.8.0-eksbuild.1"
    conflictResolution: "overwrite"
```

- Check the console again to find that `coredns` has been uninstalled, and `vpc-cni` has replaced it

<img width="1336" alt="image" src="https://user-images.githubusercontent.com/12645162/140269006-acd36b1b-b115-4efa-bb17-6f499072498f.png">

- You can repeat this in any configuration and get the same result for all add ons.

If you check the spec in cluster, all are in the spec, only the last one is in the status as active, and the EKSAddonsConfigured condition is complete.

```
  Addons:
    Conflict Resolution:    overwrite
    Name:                   kube-proxy
    Version:                v1.20.4-eksbuild.2
    Conflict Resolution:    overwrite
    Name:                   coredns
    Version:                v1.8.3-eksbuild.1
    Conflict Resolution:    overwrite
    Name:                   vpc-cni
    Version:                v1.8.0-eksbuild.1
``` 

```
Status:
  Addons:
[...]

PR #2963, opened 2021-11-16
[backport 0.6] fix: support more than 1 eks addon

richardcase:
**What type of PR is this?**

/kind bug

When the EKS addons functionality was added to CAPA there was only 1
addon available (vpc) and so it was never tested with multiple addons.
Now there are multiple addons there are issues.

The unit and e2e tests only tested 1 addons. On investigation the issue
was caused by implicit memory aliasing in a loop. This change fixes the
loop and also expands the unit and e2e tests to include multiple eks
addons

Backport of #2914

Relates #2911 

**Checklist**:

- [x] squashed commits
- [x] adds unit tests

**Release note**:

```release-note
Fixes an issue with using multiple EKS addons.
```

### DIFF
--- a/pkg/eks/addons/procedures.go
+++ b/pkg/eks/addons/procedures.go
@@ -23,6 +23,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/eks"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/wait"
 )
 
 var (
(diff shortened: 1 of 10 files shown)


pkg/eks/addons/procedures.go
```diff
@@ -166,10 +167,12 @@ func (p *CreateAddonProcedure) Name() string {
 }
 
 // WaitAddonActiveProcedure is a procedure that will wait for an EKS addon
-// to be active in a cluster
+// to be active in a cluster. Abd optionally include the degraded state.
```
shivi28, line 170: Minor typo change. `And optionally ...`

richardcase:
I will look into these failures. Although, do we really need to backport this to 0.6?

sedefsavas:
> I will look into these failures. Although, do we really need to backport this to 0.6?


@richardcase 
Not really if we consider supporting more than 1 addon as a feature.

sedefsavas:
Shall we close this?

richardcase:
@sedefsavas - tempted to close it but I can't help wondering if we need to do another 0.6 release, in which case this should go into it

richardcase:
@sedefsavas - as 0.6 will be deprecated soon and the fact that this is fixed in 0.7/1.X i will close this.

PR #2976, opened 2021-11-19
Fix the AWS.SimpleQueueService.NonExistentQueue errors in CAPA logs

### LINKED ISSUE #2912: Fix the `AWS.SimpleQueueService.NonExistentQueue` errors
/kind bug
/help

When looking at the logs for e2e test runs, there are a lot of errors like this:

```bash
E1103 18:11:17.046414       1 controller.go:317] controller/awscluster "msg"="Reconciler error" "error"="AWS.SimpleQueueService.NonExistentQueue: The specified queue does not exist for this wsdl version.\n\tstatus code: 400, request id: 346776dc-e171-5851-b3a6-2f2032741436" "name"="cluster-mbp3le" "namespace"="functional-gpu-cluster-dock0l" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AWSCluster" 
```

[Source log file](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2871/pull-cluster-api-provider-aws-e2e/1455958494833283072/artifacts/clusters/bootstrap/controllers/capa-controller-manager/capa-controller-manager-7bd56cd58-w68gp/manager.log)

This makes it harder to look for the real errors of failures with these log messages polluting the log.

I expect these errors not to be logged.....unless its  actually the issue.

**Environment:**

Ankitasw:
**What type of PR is this?**
/kind bug

This PR fixes the SQS queue non existing errors from CAPA logs.

Fixes #2912 

**Checklist**:
- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Fix recurring AWS.SimpleQueueService.NonExistentQueue error messages from CAPA logs
```

### DIFF
--- a/exp/instancestate/awsinstancestate_controller.go
+++ b/exp/instancestate/awsinstancestate_controller.go
@@ -23,6 +23,7 @@ import (
 	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
 	"github.com/aws/aws-sdk-go/service/sqs"
 	"github.com/aws/aws-sdk-go/service/sqs/sqsiface"
 	"github.com/go-logr/logr"
@@ -95,6 +96,9 @@ func (r *AwsInstanceStateReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	if _, ok := r.queueURLs.Load(awsCluster.Name); !ok {
 		URL, err := r.getQueueURL(awsCluster)
 		if err != nil {
+			if queueNotFoundError(err) {
+				return reconcile.Result{}, nil
+			}
 			return reconcile.Result{}, err
 		}
 		r.queueURLs.Store(awsCluster.Name, queueParams{region: awsCluster.Spec.Region, URL: URL})
@@ -219,6 +223,15 @@ func (r *AwsInstanceStateReconciler) getQueueURL(cluster *infrav1.AWSCluster) (s
 	}
 
 	return *resp.QueueUrl, nil
+}
+
+func queueNotFoundError(err error) bool {
+	if aerr, ok := err.(awserr.Error); ok {
+		if aerr.Code() == sqs.ErrCodeQueueDoesNotExist {
+			return true
+		}
+	}
+	return false
 }
 
 type queueParams struct {


exp/instancestate/awsinstancestate_controller.go
```diff
@@ -94,7 +95,7 @@ func (r *AwsInstanceStateReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	// retrieve queue URL if it isn't already tracked
 	if _, ok := r.queueURLs.Load(awsCluster.Name); !ok {
 		URL, err := r.getQueueURL(awsCluster)
-		if err != nil {
+		if err != nil && !queueNotFoundError(err) {
```
sedefsavas, line 98: I think we show queueNotFoundError, only when feature is enabled like this:
`	if feature.Gates.Enabled(feature.EventBridgeInstanceState) {}
`
So that users that enable it, sees this error and understand what is wrong.

Ankitasw: we are setting this feature gate as [true](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/data/e2e_conf.yaml#L212) for E2E tests, so this log flooding will keep on existing in E2E artifacts logs

richardcase: Do we actually test the event bridge functionality in the e2e tests? Also, what creates the SQS queue in the first place (i.e is it auto-created by AWS or is it something that we need to create manually).

I agree with @sedefsavas i think we need to display the error here if the queue doesn't exist but we need to work out why the queue doesn't exist and fix that as the root cause.

Ankitasw: I see it's enabled [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/shared/template.go#L96)
Not sure where it is used.

Ankitasw: Is it ok if we disable this feature flag?

richardcase: We could disable the event bridge feature if we aren't testing it in the e2e and that would stop the messages. However, i'm not sure that truly gets to the bottom of what is causing the error.

I'm curious to know why we are getting this error if the event bridge feature is enabled and CAPA is responsible for creating the queues? This implies there is a bug somewhere in either the e2e tests or the reconciliation loop.

This is a weird part of the error as well: `The specified queue does not exist for this wsdl version.`

pydctw: Based on my brief investigation during e2e runs, it seems that `AWS.SimpleQueueService.NonExistentQueue` error occurs because the awsinstancestate_controller tries to retrieve a queue url before a queue is created and after a queue is deleted. 

I think there is a bug in the reconciliation loop that needs to be fixed.

sedefsavas: @Ankitasw let's see if this is actually a bug and if so, we can do the fix in this PR. 
Also, we are missing documentation for this feature https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2978, in case you are interested.

Ankitasw: @sedefsavas @richardcase I agree to what @pydctw is telling, I checked the logs, that are populating even before queue is created. I think we should check for NonExistentQueue error [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/instancestate/awsinstancestate_controller.go#L95-L101) for reconcile loop.

sedefsavas: One issue with not logging error in case queue is not found is if AWSCluster controller fails to create queue, user won't know why this feature is not working.

We can add a condition with information severity here:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3e61428d885f2eaeab1a7dffc4d17270a20c6675/controllers/awscluster_controller.go#L238-L245

Ankitasw: I think if queue is not created successfully, it would be captured [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/instancestate/queue.go#L36-L48)

sedefsavas: 👍 let us know if the pr is ready.

Ankitasw: it's ready now.

exp/instancestate/awsinstancestate_controller.go
```diff
@@ -135,7 +136,7 @@ func (r *AwsInstanceStateReconciler) watchQueuesForInstanceEvents() {
 					return
 				}
 				resp, err := sqsSvs.ReceiveMessage(&sqs.ReceiveMessageInput{QueueUrl: aws.String(qp.URL)})
-				if err != nil {
+				if err != nil && !queueNotFoundError(err) {
```
Ankitasw, line 139: This will execute every second so it will flood the logs again if the queue is not created and we are trying to receive msg on that queue.

pkg/cloud/services/instancestate/rule.go
```diff
@@ -65,7 +65,7 @@ func (s Service) reconcileRules() error {
 		QueueName: aws.String(GenerateQueueName(s.scope.Name())),
 	})
 
-	if err != nil {
+	if err != nil && !queueNotFoundError(err) {
```
Ankitasw, line 68: This handling is done wherever `GetQueueURL` API is called.

Ankitasw:
Logs looks better now https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2976/pull-cluster-api-provider-aws-e2e/1467534737634496512/artifacts/clusters/bootstrap/controllers/capa-controller-manager/capa-controller-manager-7bd56cd58-ph4c4/manager.log
The GPU test failure is because of resource limit exceeded, which is a separate issue.

exp/instancestate/awsinstancestate_controller.go
sedefsavas: Does `ReceiveMessage()` return `queueNotFoundError`?

Also, it iterates over `r.queueURLs.Range`, so this point is reached only when the queue is created.

Ankitasw: Yes ReceiveMessage() is also flooding the logs with this error:
controllers/AWSInstanceStateController "msg"="failed to receive messages" "error"="AWS.SimpleQueueService.NonExistentQueue: The specified queue does not exist for this wsdl version.\n\tstatus code: 400, request id: 6b0eae44-fc62-5fb1-ab3f-f46dc2151960"

Ankitasw: ^^this one seems to be a bug, coz even after the queue is created, it's throwing this error

Ankitasw: @sedefsavas in current runs also I can see these [logs](https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-provider-aws-e2e/1468777921022791680/artifacts/clusters/bootstrap/controllers/capa-controller-manager/capa-controller-manager-7bd56cd58-g4ql6/manager.log)

Ankitasw: @sedefsavas I have fixed the issue in latest commit

Ankitasw: Logs are comparatively reduced now but still seeing this error:
> controllers/AWSInstanceStateController "msg"="failed to receive messages" "error"="AWS.SimpleQueueService.NonExistentQueue: The specified queue does not exist for this wsdl version.\n\tstatus code: 400, request id: f4974e87-8498-56fc-968c-14ac7bef669c" 

The problem is when the queue gets deleted, we remove `queueURL` value from the map [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/instancestate/awsinstancestate_controller.go#L91) , but the map will still contain the `key` as URL, so [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/instancestate/awsinstancestate_controller.go#L129) loop gets executed even if `val` is empty

Ankitasw: One solution is to execute [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/instancestate/awsinstancestate_controller.go#L130) go routine only when val is not empty

PR #3009, opened 2021-12-08
Allow defining Cloudformation Stack tags

### LINKED ISSUE #3015: Allow to define tags in the bootstrap cloudformation stack
/kind feature

When creating a bootstrap IAM Cloudformation stack via `clusterawsadm bootstrap iam create-cloudformation-stack` we could have a `--tags` parameter, so that we don't need to run a new command to modify the stack adding tags.

None.

**Environment:**

clusterawsadm version: `&version.Info{Major:"1", Minor:"1", GitVersion:"v1.1.0", GitCommit:"0b2e34680d117b1d8146965f3123c04709d37951", GitTreeState:"clean", BuildDate:"2021-11-17T22:30:32Z", GoVersion:"go1.17.3", AwsSdkVersion:"v1.40.56", Compiler:"gc", Platform:"linux/amd64"}`
- OS (e.g. from `/etc/os-release`):  N/A

jonathanbeber:
**What type of PR is this?**

/kind feature

It's not possible to add tags to the IAM bootstrap Cloudformation Stack
without exporting it as a template and manually applying or editing the
created stack.

This commit adds the `--tags` flag to the command
`bootstrap iam create-cloudformation-stack`.

Fixes #3015 and [this slack thread](https://kubernetes.slack.com/archives/C0LRMHZ1T/p1638810698316500).

none

**Checklist**:

- [x] squashed commits

**Release note**:

```release-note
Add the --tags to the bootstrap iam [create|update]-cloudformation-stack command.
```


sedefsavas:
Instead of a new flag, adding `StackTags` field in `AWSIAMConfiguration` and read if from config file during the stack creation would make defining multiple tags easier. WDYT?

jonathanbeber:
@sedefsavas thanks for taking a look. 

I agree it would make it easier to add multiple flags. I considered this option and in the end opted for a flag because it seemed to me more like tags work in CF. i.e. I can't define a tag for a stack inside its template, I have to add it in the command creating the stack. This seemed to me more like that, more the AWS way. 

But your point is valid, it would be way easier to define multiple tags and could be versioned, template shared, etc. I will change the implementation to this form. I'm also adding a new E2E test for that.

jonathanbeber:
had to rerun `make generate`, sorry.

jonathanbeber:
it seems ready from my side @sedefsavas

sedefsavas:
We need to have conversion for the new tags field like this (First marshall it during convertFrom, then unmarshall convertto):

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0672892f699ca235fbff8f78448ae27ffedf60c1/bootstrap/eks/api/v1alpha4/conversion.go#L34-L39

sedefsavas:
Auto-generated conversion code handles existing fields. But for a scenario where 

1. Your API version is v1beta1
2. Create a v1beta1 config resource with tags field.
3. Someone else reads it with v1alpha1 client
4. Tags field is not saved anywhere, when v1alpha1 client writes back, it will hit the conversion webhook to be saved as v1beta1. At this point, we could not recover the initial tags field.

Instead in step 3, we should have marshalled the data, and in step 4 during unmarshall it and set it in the v1beta1 object during conversion.

sedefsavas:
Will run the blocking e2e test to make sure e2e is not failing after verify is fixed.

PR #3027, opened 2021-12-13
Add new condition (EKSNodegroupUpdateSucceededCondition) to the AWSManagedMachinePool resource

### LINKED ISSUE #2966: Add Creating/Updating conditions to EKS node group
/kind feature

We have [Ready/Creating/Updating conditions](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4274a5ab5c9dc391045840d6e45e17fe0cbab3dc/controlplane/eks/api/v1beta1/conditions_consts.go#L23-L29) for EKS control plan, however there is only [Ready condition](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4274a5ab5c9dc391045840d6e45e17fe0cbab3dc/exp/api/v1beta1/conditions_consts.go#L49) for EKS node group. It would be great if we can also add Creating/Updating conditions to EKS node group as well.

**Environment:**

jon-fearer:
**What type of PR is this?**
/kind feature

This PR adds a new condition (`EKSNodegroupUpdateSucceededCondition`) to the `AWSManagedMachinePool` resource, as well as related reasons (`Creating`, `Updating`, `FailedToCreate`, `FailedToUpdate`). See issue link below as well as the discussion in this PR.

fixes #2966 

 I haven't found any places in the documentation that would need manual updating as a result of this change, but please let me know if I missed something. Thanks in advance for the review.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add new condition (EKSNodegroupUpdateSucceededCondition) to the AWSManagedMachinePool resource.
```

### DIFF
--- a/pkg/cloud/services/eks/nodegroup.go
+++ b/pkg/cloud/services/eks/nodegroup.go
@@ -36,6 +36,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	"sigs.k8s.io/cluster-api/controllers/noderefutil"
+	"sigs.k8s.io/cluster-api/util/conditions"
 )
 
 func (s *NodegroupService) describeNodegroup() (*eks.Nodegroup, error) {
@@ -337,6 +338,13 @@ func (s *NodegroupService) reconcileNodegroupVersion(ng *eks.Nodegroup) error {
 			updateMsg = fmt.Sprintf("to AMI version %s", *input.ReleaseVersion)
 		}
 
+		conditions.MarkFalse(
+			s.scope.ManagedMachinePool,
+			expinfrav1.EKSNodegroupUpdateSucceededCondition,
+			expinfrav1.EKSNodegroupUpdatingReason,
+			clusterv1.ConditionSeverityInfo,
+			"",
+		)
 		if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
 			if _, err := s.EKSClient.UpdateNodegroupVersion(input); err != nil {
 				if aerr, ok := err.(awserr.Error); ok {
@@ -344,10 +352,21 @@ func (s *NodegroupService) reconcileNodegroupVersion(ng *eks.Nodegroup) error {
 				}
 				return false, err
 			}
-			record.Eventf(s.scope.ManagedMachinePool, "SuccessfulUpdateEKSNodegroup", "Updated EKS nodegroup %s %s", eksClusterName, updateMsg)
+			record.Eventf(s.scope.ManagedMachinePool, "UpdatingEKSNodegroup", "Updating EKS nodegroup %s %s", eksClusterName, updateMsg)
 			return true, nil
 		}); err != nil {
-			record.Warnf(s.scope.ManagedMachinePool, "FailedUpdateEKSNodegroup", "failed to update the EKS nodegroup %s %s: %v", eksClusterName, updateMsg, err)
+			errorMessage := "failed to update the EKS nodegroup %s %s: %v"
+			record.Warnf(s.scope.ManagedMachinePool, "FailedUpdateEKSNodegroup", errorMessage, eksClusterName, updateMsg, err)
+			conditions.MarkFalse(
+				s.scope.ManagedMachinePool,
+				expinfrav1.EKSNodegroupUpdateSucceededCondition,
+				expinfrav1.EKSNodegroupFailedToUpdateReason,
+				clusterv1.ConditionSeverityError,
+				errorMessage,
+				eksClusterName,
+				updateMsg,
+				err,
+			)
 			return errors.Wrapf(err, "failed to update EKS nodegroup")
 		}
 	}
@@ -460,8 +479,23 @@ func (s *NodegroupService) reconcileNodegroupConfig(ng *eks.Nodegroup) error {
 		return errors.Wrap(err, "created invalid UpdateNodegroupConfigInput")
 	}
 
+	conditions.MarkFalse(
+		s.scope.ManagedMachinePool,
+		expinfrav1.EKSNodegroupUpdateSucceededCondition,
+		expinfrav1.EKSNodegroupUpdatingReason,
+		clusterv1.ConditionSeverityInfo,
+		"",
+	)
 	_, err = s.EKSClient.UpdateNodegroupConfig(input)
 	if err != nil {
+		conditions.MarkFalse(
+			s.scope.ManagedMachinePool,
+			expinfrav1.EKSNodegroupUpdateSucceededCondition,
+			expinfrav1.EKSNodegroupFailedToUpdateReason,
+			clusterv1.ConditionSeverityError,
+			"failed to update nodegroup config: %v",
+			err,
+		)
 		return errors.Wrap(err, "failed to update nodegroup config")
 	}
 
@@ -475,8 +509,23 @@ func (s *NodegroupService) reconcileNodegroup() error {
 	}
 
 	if eksClusterName, eksNodegroupName := s.scope.KubernetesClusterName(), s.scope.NodegroupName(); ng == nil {
+		conditions.MarkFalse(
+			s.scope.ManagedMachinePool,
+			expinfrav1.EKSNodegroupReadyCondition,
+			expinfrav1.EKSNodegroupCreatingReason,
+			clusterv1.ConditionSeverityInfo,
+			"",
+		)
 		ng, err = s.createNodegroup()
 		if err != nil {
+			conditions.MarkFalse(
+				s.scope.ManagedMachinePool,
+				expinfrav1.EKSNodegroupReadyCondition,
+				expinfrav1.EKSNodegroupFailedToCreateReason,
+				clusterv1.ConditionSeverityError,
+				"failed to create nodegroup: %v",
+				err,
+			)
 			return errors.Wrap(err, "failed to create nodegroup")
 		}
 		s.scope.Info("Created EKS nodegroup in AWS", "cluster-name", eksClusterName, "nodegroup-name", eksNodegroupName)
@@ -496,12 +545,12 @@ func (s *NodegroupService) reconcileNodegroup() error {
 	switch *ng.Status {
 	case eks.NodegroupStatusCreating, eks.NodegroupStatusUpdating:
 		ng, err = s.waitForNodegroupActive()
+		if err != nil {
+			return errors.Wrap(err, "failed to wait for nodegroup to be active")
+		}
+		s.markReadyAndUpdateConditionsTrue()
 	default:
 		break
-	}
-
-	if err != nil {
-		return errors.Wrap(err, "failed to wait for nodegroup to be active")
 	}
 
 	if err := s.reconcileNodegroupVersion(ng); err != nil {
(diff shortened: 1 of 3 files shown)


sedefsavas:
Creating and updating sounds more like reasons to explain operational status.
There is `EKSNodegroupReadyCondition`, `Creating` can be a reason to why NodeGroup is not ready yet.

And for update, a new condition like `EKSNodegroupUpdateSucceededCondition` with failure reasons `updating`, `failedToUpdate` make more sense to me.

I will defer to @richardcase.

jon-fearer:
`EKSNodegroupReadyCondition` = False with reason of `Creating` makes sense to me.

Would setting `EKSNodegroupUpdateSucceededCondition` to False (with reason `Updating`) during the update be misleading at all? It technically hasn't failed or succeeded at that point

sedefsavas:
> Would setting `EKSNodegroupUpdateSucceededCondition` to False (with reason `Updating`) during the update be misleading at all? It technically hasn't failed or succeeded at that point

`Succeeded = false` does not mean it failed, and explaining with a `updating` reason means that update is not succeeded due to a continuing update. 
 
Examle from cluster api:

```
// DrainingSucceededCondition provide evidence of the status of the node drain operation which happens during the machine
	// deletion process.
	DrainingSucceededCondition ConditionType = "DrainingSucceeded"

	// DrainingReason (Severity=Info) documents a machine node being drained.
	DrainingReason = "Draining"

	// DrainingFailedReason (Severity=Warning) documents a machine node drain operation failed.
	DrainingFailedReason = "DrainingFailed"
```

sedefsavas:
Is this PR ready for another review?

jon-fearer:
No, should I move forward with the requested changes? On a related note, would it make sense to do something similar for the EKS control plane conditions/reasons?

sedefsavas:
A refactoring is needed there as well.
For example `EKSControlPlaneUpdatingCondition` becomes false when updating is completed, no way to understand what is wrong when the update is failed by looking at this. Instead, we could have a `EKSControlPlaneUpdateSucceededCondition` which will only show up if an update is started so that user will now an update is in progress and if the update is failed, the reason will also show up.

For these changes, we can define new conditions (like EKSControlPlaneUpdateSucceededCondition) and add deprecate the redundant ones (like EKSControlPlaneUpdatingCondition) and eventually remove deprecated ones in 2.x release.
But this needs a separate issue/PR.

richardcase:
> A refactoring is needed there as well. For example `EKSControlPlaneUpdatingCondition` becomes false when updating is completed, no way to understand what is wrong when the update is failed by looking at this. Instead, we could have a `EKSControlPlaneUpdateSucceededCondition` which will only show up if an update is started so that user will now an update is in progress and if the update is failed, the reason will also show up.

I'm not sure about this to be honest. If i start an update to the control i think it makes sense to see `EKSControlPlaneUpdatingCondition` as true. If it the update fails then i see no problem with seeing `EKSControlPlaneUpdatingCondition` as false with a `failed` reason and message that contains the failure message.

jon-fearer:
@sedefsavas This is ready for another look. Let me know your thoughts

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -337,17 +338,38 @@ func (s *NodegroupService) reconcileNodegroupVersion(ng *eks.Nodegroup) error {
...
+		conditions.MarkFalse(
+			s.scope.ManagedMachinePool,
+			expinfrav1.EKSNodegroupUpdateSucceededCondition,
+			expinfrav1.EKSNodegroupUpdatingReason,
+			clusterv1.ConditionSeverityInfo,
+			"",
+		)
 		if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
 			if _, err := s.EKSClient.UpdateNodegroupVersion(input); err != nil {
+				conditions.MarkFalse(
```
richardcase, line 350: We probably don't need to set the condition here as it is covered by the error handling block 364-384

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -337,17 +338,38 @@ func (s *NodegroupService) reconcileNodegroupVersion(ng *eks.Nodegroup) error {
...
+			record.Eventf(s.scope.ManagedMachinePool, "UpdatingEKSNodegroup", "Updating EKS nodegroup %s %s", eksClusterName, updateMsg)
 			return true, nil
 		}); err != nil {
 			record.Warnf(s.scope.ManagedMachinePool, "FailedUpdateEKSNodegroup", "failed to update the EKS nodegroup %s %s: %v", eksClusterName, updateMsg, err)
+			conditions.MarkFalse(
+				s.scope.ManagedMachinePool,
+				expinfrav1.EKSNodegroupUpdateSucceededCondition,
+				expinfrav1.EKSNodegroupFailedToUpdateReason,
+				clusterv1.ConditionSeverityError,
+				"",
```
richardcase, line 371: We should probably set the error message here so that we can see this when we do a describe.

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -460,8 +482,22 @@ func (s *NodegroupService) reconcileNodegroupConfig(ng *eks.Nodegroup) error {
...
+		"",
+	)
 	_, err = s.EKSClient.UpdateNodegroupConfig(input)
 	if err != nil {
+		conditions.MarkFalse(
+			s.scope.ManagedMachinePool,
+			expinfrav1.EKSNodegroupUpdateSucceededCondition,
+			expinfrav1.EKSNodegroupFailedToUpdateReason,
+			clusterv1.ConditionSeverityInfo,
+			"",
```
richardcase, line 499: Probably best to set the error message here

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -475,8 +511,22 @@ func (s *NodegroupService) reconcileNodegroup() error {
...
+		)
 		ng, err = s.createNodegroup()
 		if err != nil {
+			conditions.MarkFalse(
+				s.scope.ManagedMachinePool,
+				expinfrav1.EKSNodegroupReadyCondition,
+				expinfrav1.EKSNodegroupFailedToCreateReason,
+				clusterv1.ConditionSeverityError,
+				"",
```
richardcase, line 528: Also, here it would be good to set the error message.

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -496,14 +546,18 @@ func (s *NodegroupService) reconcileNodegroup() error {
 	switch *ng.Status {
 	case eks.NodegroupStatusCreating, eks.NodegroupStatusUpdating:
 		ng, err = s.waitForNodegroupActive()
+		if err != nil {
+			return errors.Wrap(err, "failed to wait for nodegroup to be active")
+		}
+		conditions.MarkTrue(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupReadyCondition)
+		if conditions.IsFalse(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition) &&
```
richardcase, line 553: Maybe we could just set the `EKSNodegroupUpdateSucceededCondition` to true if we get here and if `*ng.Status==eks.NodegroupStatusUpdating`? wdyt?

jon-fearer: I think the status will have changed by the time we get to this line, because we are calling `ng, err := s.waitForNodegroupActive()` which looks like it will update `ng.Status` to something else: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f4c22203805d14b5bc57ca7605a9ee611d175e76/pkg/cloud/services/eks/nodegroup.go#L640-L646

pkg/cloud/services/eks/nodegroup.go
jon-fearer: Updated

pkg/cloud/services/eks/nodegroup.go
jon-fearer: Updated

pkg/cloud/services/eks/nodegroup.go
jon-fearer: Updated

pkg/cloud/services/eks/nodegroup.go
jon-fearer: Updated

pkg/cloud/services/eks/nodegroup.go
richardcase: Thats a good point @jon-fearer, i missed that...thanks :)

The intent of the condition check logic on [552-555 ](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/cec813cf16c54cd1e3570926cc50342f44f9d7a0/pkg/cloud/services/eks/nodegroup.go#L552:L555) isn't immediately obvious to me on reading the code. 

I'm wonder if one of the following would make it more obvious:
*  Extract the `conditions.IsFalse(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition) && conditions.GetReason(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition) == expinfrav1.EKSNodegroupUpdatingReason` into a function with a name that indicates its intent
* Split the ` eks.NodegroupStatusCreating, eks.NodegroupStatusUpdating` into 2 separate cases in the switch statement, with the `eks.NodegroupStatusUpdating` case doing the additional `conditions.MarkTrue(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition)`. We'd probably want to extra the common code into a function that is called in each case.

wdyt?

jon-fearer: I'm wondering if those if statements are even necessary. I think the main reason for putting them in there was wanting to avoid setting the `EKSNodegroupUpdateSucceededCondition` to true when the node group is created for the first time (since it is a "create" and not an "update"). But if that's ok, then always marking them both true when the node group is active should be fine, so I just moved both `markTrue(...)` statements into a separate function. Does that behavior make sense to you (always setting both conditions to true when cluster becomes active)?

jon-fearer: If we only want to set the`EKSNodegroupUpdateSucceededCondition` to true on updates (and not creates), then I could do something like this instead:
```
// markReadyAndUpdateConditionsTrue sets the EKSNodegroupReadyCondition and EKSNodegroupUpdateSucceededCondition
// to true for the ManagedMachinePool.
func (s *NodegroupService) markReadyAndUpdateConditionsTrue() {
	conditions.MarkTrue(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupReadyCondition)
	if (s.isNodegroupConditionUpdating()) {
		conditions.MarkTrue(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition)
	}
}

func (s *NodegroupService) isNodegroupConditionUpdating() bool {
	return conditions.IsFalse(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition) &&
		conditions.GetReason(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition) == expinfrav1.EKSNodegroupUpdatingReason
}
```

richardcase: > then always marking them both true when the node group is active should be fine, so I just moved both `markTrue(...)` statements into a separate function. Does that behavior make sense to you (always setting both conditions to true when cluster becomes active)

Thats a good question :+1:  I don't think we should set `expinfrav1.EKSNodegroupUpdateSucceededCondition` to true on
create as this gives the impression that an update has been done when it hasn't.

jon-fearer: Ok sounds good. Here are those [changes](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3027/files#diff-ed6ead8e5e6ab86b4693e189f4568484aeda9b9c660e21f70f08fab9f5da13e3R652-R666). Let me know your thoughts

Ankitasw:
@sedefsavas @richardcase I think this PR is ready for another review. Since I was not involved from the start, I dont have much context so just wanted to highlight this PR for review.

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -599,3 +648,19 @@ func (s *NodegroupService) waitForNodegroupActive() (*eks.Nodegroup, error) {
...
+	conditions.MarkTrue(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupReadyCondition)
+	if s.isNodegroupConditionUpdating() {
+		conditions.MarkTrue(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition)
+	}
+}
+
+// isNodegroupConditionUpdating checks if the node group is currently in an updating condition (i.e. an update
+// was previously initiated).
+func (s *NodegroupService) isNodegroupConditionUpdating() bool {
+	return conditions.IsFalse(s.scope.ManagedMachinePool, expinfrav1.EKSNodegroupUpdateSucceededCondition) &&
```
richardcase, line 664: I'm no in favour of using conditions for logic control. If a not group is updating we should be able to get this from the status of the nodegroup from AWS.

richardcase: @jon-fearer - just checking in on this to see what you though?

PR #3063, opened 2022-01-06
[WIP]Fetch ELB instance port from KCP's kubeconfig of cluster

### LINKED ISSUE #2316: ELB Instance port for API should not be hardcoded
/kind feature

The ELB instance port for API is hardcoded to `6443` here https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/5d3c76211239b3700d1f69a36c7819f94c5b2609/pkg/cloud/services/elb/loadbalancer.go#L323

Even though the port can be configured in `KubeadmControlPlane` CR in `InitConfiguration` part:

https://pkg.go.dev/k8s.io/kubernetes/cmd/kubeadm/app/apis/kubeadm/v1beta1#APIEndpoint
`KubeadmControlPlane.Spec.KubeadmConfigSpec.InitConfiguration.LocalAPIEndpoint.BindPort` will set the port on API manifest to the desired value

If someone sets that value to something else than 6443 then it will create a broken CAPA cluster.

I would expect the controller to read this information and adjust the ELB config accordingly.

**Environment:**

- Cluster-api-provider-aws version: v1alapha3

Ankitasw:
**What type of PR is this?**
/kind bug

This PR removes the hardcoded ELB instance port value, and reads it's value from the KubeadmControlPlane kubeconfig's InitConfig BindPort, so as to keep the port information consistent throughout the cluster.

Fixes #2316 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Fetch ELB instance port from KCP kubeconfig's APIEndpoint
```

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -296,3 +296,11 @@ rules:
   - get
   - patch
   - update
+- apiGroups:
+  - controlplane.cluster.x-k8s.io
+  resources:
+  - kubeadmcontrolplanes
+  verbs:
+  - get
+  - list
+
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -75,6 +75,7 @@ type AWSClusterReconciler struct {
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsclusters,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=kubeadmcontrolplanes,verbs=get;list
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsclusterroleidentities;awsclusterstaticidentities,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsclustercontrolleridentities,verbs=get;list;watch;create;
 
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -23,15 +23,20 @@ import (
 	awsclient "github.com/aws/aws-sdk-go/aws/client"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/klog/v2/klogr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/throttle"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
+
+const defaultELBPort int32 = 6443
 
 // ClusterScopeParams defines the input parameters used to create a new Scope.
 type ClusterScopeParams struct {
@@ -96,6 +101,28 @@ type ClusterScope struct {
 	session         awsclient.ConfigProvider
 	serviceLimiters throttle.ServiceLimiters
 	controllerName  string
+}
+
+// KubeadmControlPlaneBindPort returns the bind port used in KCP kubeconfig.
+func (s *ClusterScope) KubeadmControlPlaneBindPort() (int32, error) {
+	if s.Cluster.Spec.ControlPlaneRef != nil {
+		cpObjectKey := client.ObjectKey{
+			Namespace: s.Cluster.Namespace,
+			Name:      s.Cluster.Spec.ControlPlaneRef.Name,
+		}
+		var kubeadmCP controlplanev1.KubeadmControlPlane
+		err := s.client.Get(context.TODO(), cpObjectKey, &kubeadmCP)
+
+		if err != nil && !apierrors.IsNotFound(err) {
+			return 0, err
+		}
+
+		var apiEndpoint bootstrapv1.APIEndpoint
+		if kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration != nil && kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration.LocalAPIEndpoint != apiEndpoint {
+			return kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration.LocalAPIEndpoint.BindPort, nil
+		}
+	}
+	return defaultELBPort, nil
 }
 
 // Network returns the cluster network object.
--- a/pkg/cloud/services/elb/loadbalancer_test.go
+++ b/pkg/cloud/services/elb/loadbalancer_test.go
@@ -29,6 +29,8 @@ import (
 	rgapi "github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
 	"github.com/golang/mock/gomock"
 	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/utils/pointer"
@@ -38,6 +40,8 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/elb/mock_elbiface"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/elb/mock_resourcegroupstaggingapiiface"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	cabpkv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
+	controlplanev1 "sigs.k8s.io/cluster-api/controlplane/kubeadm/api/v1beta1"
 	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 )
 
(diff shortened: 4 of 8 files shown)


pkg/cloud/scope/cluster.go
```diff
@@ -98,6 +100,23 @@ type ClusterScope struct {
...
+	if s.Cluster.Spec.ControlPlaneRef != nil {
+		cpObjectKey := client.ObjectKey{
+			Namespace: s.Cluster.Namespace,
+			Name:      s.Cluster.Spec.ControlPlaneRef.Name,
+		}
+		var kubeadmCP controlplanev1.KubeadmControlPlane
+		_ = s.client.Get(context.TODO(), cpObjectKey, &kubeadmCP)
+
+		var apiEndpoint bootstrapv1.APIEndpoint
+		if kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration != nil && kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration.LocalAPIEndpoint != apiEndpoint {
```
sedefsavas, line 113:
```suggestion
		if kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration != nil && kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration.LocalAPIEndpoint != bootstrapv1.APIEndpoint{} {
```

pkg/cloud/scope/cluster.go
```diff
@@ -98,6 +100,23 @@ type ClusterScope struct {
 	controllerName  string
 }
 
+func (s *ClusterScope) KubeadmControlPlaneBindPort() int32 {
```
sedefsavas, line 103: Missing description of what this method does.

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -1000,6 +1003,109 @@ func TestChunkELBs(t *testing.T) {
...
+								BindPort: 443,
+							},
+						},
+					},
+				},
+			},
+			mocks: func(m *mock_ec2iface.MockEC2APIMockRecorder) {},
+			expect: func(t *testing.T, res *infrav1.ClassicELB) {
+				if !res.Attributes.CrossZoneLoadBalancing {
+					t.Error("Expected load balancer to have cross-zone load balancing enabled")
```
sedefsavas, line 1037: Do we need `CrossZoneLoadBalancing` check here?

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -1000,6 +1003,109 @@ func TestChunkELBs(t *testing.T) {
...
+			}
+
+			tc.mocks(ec2Mock.EXPECT())
+
+			s := &Service{
+				scope:     clusterScope,
+				EC2Client: ec2Mock,
+			}
+
+			spec, err := s.getAPIServerClassicELBSpec(clusterScope.Name())
```
sedefsavas, line 1099: Where do you check if the correct ELB internal port is used?

pkg/cloud/services/elb/loadbalancer_test.go
Ankitasw: No, we don't, I will remove it.

pkg/cloud/scope/cluster.go
Ankitasw: Adding this gives the compile time error, Type 'bootstrapv1.APIEndpoint' is not an expression. That's why added the variable and then compared.

pkg/cloud/scope/cluster.go
```diff
@@ -98,6 +100,23 @@ type ClusterScope struct {
...
+	if s.Cluster.Spec.ControlPlaneRef != nil {
+		cpObjectKey := client.ObjectKey{
+			Namespace: s.Cluster.Namespace,
+			Name:      s.Cluster.Spec.ControlPlaneRef.Name,
+		}
+		var kubeadmCP controlplanev1.KubeadmControlPlane
+		_ = s.client.Get(context.TODO(), cpObjectKey, &kubeadmCP)
```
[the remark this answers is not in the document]
Ankitasw: Done

pkg/cloud/scope/cluster.go
Ankitasw: is there any better way to do this?

pkg/cloud/scope/cluster.go
Ankitasw: Done

pkg/cloud/services/elb/loadbalancer_test.go
Ankitasw: Added the check now

Ankitasw:
Looks like e2e is failing with following error 
failed to list *v1beta1.KubeadmControlPlane: kubeadmcontrolplanes.controlplane.cluster.x-k8s.io is forbidden: User "system:serviceaccount:capa-system:capa-controller-manager" cannot list resource "kubeadmcontrolplanes" in API group "controlplane.cluster.x-k8s.io" at the cluster scope
Do I have to update the role.yaml?

config/rbac/role.yaml
```diff
@@ -295,4 +296,4 @@ rules:
   verbs:
   - get
   - patch
-  - update
+  - update
```
shivi28, line 299: nit, add new line

controllers/awscluster_controller.go
```diff
@@ -75,6 +75,7 @@ type AWSClusterReconciler struct {
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsclusters,verbs=get;list;watch;create;update;patch;delete
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsclusters/status,verbs=get;update;patch
 // +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=clusters;clusters/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=kubeadmcontrolplanes,verbs=get;list
```
shivi28, line 78: Do we need watch permission as well?

pkg/cloud/scope/cluster.go
```diff
@@ -98,6 +101,28 @@ type ClusterScope struct {
...
 
+// KubeadmControlPlaneBindPort returns the bind port used in KCP kubeconfig.
+func (s *ClusterScope) KubeadmControlPlaneBindPort() (int32, error) {
+	if s.Cluster.Spec.ControlPlaneRef != nil {
+		cpObjectKey := client.ObjectKey{
+			Namespace: s.Cluster.Namespace,
+			Name:      s.Cluster.Spec.ControlPlaneRef.Name,
+		}
+		var kubeadmCP controlplanev1.KubeadmControlPlane
+		err := s.client.Get(context.TODO(), cpObjectKey, &kubeadmCP)
```
shivi28, line 112: Should we use `context.Background()` instead of `context.TODO()`. wdyt?

pkg/cloud/scope/cluster.go
```diff
@@ -98,6 +101,28 @@ type ClusterScope struct {
...
+
+		if err != nil && !apierrors.IsNotFound(err) {
```
shivi28, line 114:
```
if err := s.client.Get(context.TODO(), cpObjectKey, &kubeadmCP); err !=nil && !apierrors.IsNotFound(err){
return 0, err
}
```

pkg/cloud/scope/cluster.go
shivi28: Can use `reflect.DeepEqual(a,b)`

pkg/cloud/scope/cluster.go
```diff
@@ -98,6 +101,28 @@ type ClusterScope struct {
...
+			return 0, err
+		}
+
+		var apiEndpoint bootstrapv1.APIEndpoint
+		if kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration != nil && kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration.LocalAPIEndpoint != apiEndpoint {
+			return kubeadmCP.Spec.KubeadmConfigSpec.InitConfiguration.LocalAPIEndpoint.BindPort, nil
+		}
+	}
+	return int32(6443), nil
```
shivi28, line 123: Can we make a constant for this? something with name  `DefaultELBPort`

pkg/cloud/scope/cluster.go
Ankitasw: IMO context.TODO() should be used, as we mostly use context.Background() from where the call starts, since this will come somewhere during reconciliation, we might already have the context in place, so we should use context.TODO()

pkg/cloud/scope/cluster.go
Ankitasw: IMO the current one is more readable as otherwise it will become a longer statement.

Ankitasw: > Can use reflect.DeepEqual(a,b)

Since this is just a nil check and not any complex structure where we have to match each and every property of the struct with each other, i think it should be fine, wdyt?

controllers/awscluster_controller.go
Ankitasw: We only need get and list permission, removed it now.

pkg/cloud/scope/cluster.go
Ankitasw: Since this is just a nil check and not any complex structure where we have to compare each and every attribute of the struct with each other, i think it should be fine, wdyt?

pkg/cloud/scope/cluster.go
```diff
@@ -23,16 +23,21 @@ import (
 	awsclient "github.com/aws/aws-sdk-go/aws/client"
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/klog/v2/klogr"
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/throttle"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	bootstrapv1 "sigs.k8s.io/cluster-api/bootstrap/kubeadm/api/v1beta1"
```
CecileRobertMichon, line 32: It seems that this PR is causing the AWS infrastructure provider to take a dependency on both the kubeadm control plane and bootstrap providers. We should be very careful with making assumptions that Clusters use either of those as they are **providers** (as opposed to core CAPI) and by design should be modular. Doing this would essentially be saying "AWS is now only compatible with the kubeadm control plane and bootstrap providers" which would cause users to be forced to use those implementations. 

It's easy to forget since the kubeadm providers are part of the CAPI repo and releases, and there hasn't been much happening in terms of other (public) control-plane/bootstrap providers so far but if we want to keep the core design principle that CAPI is modular in its bootstrap and control plane implementations we should be adding tight coupling between infra providers and other providers.

cc @vincepri @fabriziopandini @enxebre

pkg/cloud/scope/cluster.go
shivi28: Yes your point is right as its just a nil check so we can ignore it but in case if you want to avoid creating new variable then `reflect.DeepEqual(a,b)` works.

pkg/cloud/scope/cluster.go
sedefsavas: Good point @CecileRobertMichon, this is not the way to go about this.

I don't know if there were discussions around adding common fields to the bootstrap providers as contract. The port  that api server will run on might be a common field.

I don't think of a way to extract this info from somewhere else without some API changes.

We can continue discussing this in the relevant CAPI issue: https://github.com/kubernetes-sigs/cluster-api/issues/5517

enxebre: I agree, this default infra implementation shouldn't preclude a consumer from plugging in a different Control Plane provider implementation, that modularity is critical for adoption and enables consumers to experiment and to develop support for new scenarios. I'm curious doesn't this even break your `ekscontrolplanev1.AWSManagedControlPlane` workflow? At the very minimum this should check the Control Plane ref API group and kind for this assumption to be valid. Though for a more sustainable approach that applies generically we could discuss about having the API server port as part of the contract ([1] and [2]) and let it be input for the control plane CR spec or something in that line. I'm also curious on how does this relate to the LB proposal.

[1] https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/architecture/controllers/cluster.md#contracts

[2] https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/architecture/controllers/control-plane.md#contracts

sedefsavas:
/hold 
until discussions resolved.

pkg/cloud/scope/cluster.go
Ankitasw: @enxebre The proposal for adding contract looks good to me for a sustainable approach but I am not sure of it's relation to LB proposal. @randomvariable can you comment here?

yastij: I think we need to hold this and take this to upstream, we can't make the current apiserver port field as part of the contract, given where it is located in the `kubeadmConfig` spec. Let's take the discussion to https://github.com/kubernetes-sigs/cluster-api/issues/5517

Ankitasw:
This has not been fixed in CAPI yet.

PR #3071, opened 2022-01-11
Include test package to linter checks

sedefsavas:
/kind cleanup

Test package was not being checked by linter before due to its tags. Added it with this PR.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add test package linter checks.
```

### DIFF
--- a/test/e2e/shared/aws.go
+++ b/test/e2e/shared/aws.go
@@ -1,3 +1,4 @@
+//go:build e2e
 // +build e2e
 
 /*
@@ -101,7 +102,7 @@ func NewAWSSessionWithKey(accessKey *iam.AccessKey) client.ConfigProvider {
 	return sess
 }
 
-// createCloudFormationStack ensures the cloudformation stack is up to date
+// createCloudFormationStack ensures the cloudformation stack is up to date.
 func createCloudFormationStack(prov client.ConfigProvider, t *cfn_bootstrap.Template, tags map[string]string) error {
 	Byf("Creating AWS CloudFormation stack for AWS IAM resources: stack-name=%s", t.Spec.StackName)
 	CFN := cfn.New(prov)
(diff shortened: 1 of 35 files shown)


test/e2e/shared/aws.go
```diff
@@ -141,11 +142,12 @@ func deleteResourcesInCloudFormation(prov client.ConfigProvider, t *cfn_bootstra
...
 		if tayp == configservice.ResourceTypeAwsIamRole {
 			role := val.(*cfn_iam.Role)
-			iamSvc.DeleteRole(&iam.DeleteRoleInput{RoleName: aws.String(role.RoleName)})
+			_, err := iamSvc.DeleteRole(&iam.DeleteRoleInput{RoleName: aws.String(role.RoleName)})
+			Expect(err).NotTo(HaveOccurred())
 		}
 		if val.AWSCloudFormationType() == "AWS::IAM::InstanceProfile" {
 			profile := val.(*cfn_iam.InstanceProfile)
-			iamSvc.DeleteInstanceProfile(&iam.DeleteInstanceProfileInput{InstanceProfileName: aws.String(profile.InstanceProfileName)})
+			_, _ = iamSvc.DeleteInstanceProfile(&iam.DeleteInstanceProfileInput{InstanceProfileName: aws.String(profile.InstanceProfileName)})
```
alexander-demichev, line 150: Just out of curiosity, why returned error here is silenced?

sedefsavas: Failing on error here does not give any benefit, it will panic and not continue cleaning up the remaining resources down the function.
Instead of failing, we could add retry mechanism for all but best way is to have Boskos do the remaining clean up for us, which we have: https://github.com/kubernetes-sigs/boskos/blob/master/aws-janitor/resources/iam_instance_profiles.go

alexander-demichev: thanks for explanation

PR #3073, opened 2022-01-12
Increase unit test coverage for AWSCluster controller

### LINKED ISSUE #2910: Increase unit test coverage
This issue is to track improving our current unit-test coverage which is as follows:

Screen Shot 2021-11-03 at 1 08 31 PM

[Prow link](https://storage.googleapis.com/kubernetes-jenkins/logs/periodic-cluster-api-provider-aws-coverage/1455632468185976832/artifacts/coverage.html#sigs.k8s.io/cluster-api-provider-aws/) for the interactive UI above.

Artifacts of the test-coverage prow job are [here](https://gcsweb.k8s.io/gcs/kubernetes-jenkins/logs/periodic-cluster-api-provider-aws-coverage/1455632468185976832/artifacts/) for more details.

Ankitasw:
**What type of PR is this?**
/area testing

This PR increases the unit tests coverage for AWSCluster controller. All the scenarios covered are listed [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2910#issuecomment-989621532). 
Coverage of `controllers` pkg increased from 42.5% to 60%.

**Which issue(s) this PR fixes (optional, in fixes #(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged)**:
Part of #2910

Plan to submit a separate PR per each controller.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Add unit tests for AWSCluster controller
```


Ankitasw:
Coverage has now increased from 5% to 62% for awscluster_controller.

Madhur97:
Hey @Ankitasw, thanks for the PR. 
As per offline conversation, adding comments here: 
1. Use testenv instead of fakeclient for integration tests as much as possible. 
2. Let's not mock any reconcile call on service objects during integration test and only mock AWS/3rd party calls.
3. We will add missing scenario's (apart from critical ones) in the original issue to be tracked later so that we can unblock this PR( and other testing related PRs as well) after these changes.
 
cc: @sedefsavas @shivi28

Ankitasw:
@Madhur97 feel free to list the scenarios asap, as I will try to cover those in this PR only, don't want things to pile up

Madhur97:
@Ankitasw Have added some more scenarios [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2910#issuecomment-1022844927). Will review again once integration tests are updated with testenv.

controllers/awscluster_controller_test.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
```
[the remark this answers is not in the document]
Ankitasw: Since this whole file has been changed, I have added that, and yes for all these file it was added manually

controllers/awscluster_controller.go
```diff
@@ -159,10 +196,10 @@ func (r *AWSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
-	elbsvc := elb.NewService(clusterScope)
-	networkSvc := network.NewService(clusterScope)
-	sgService := securitygroup.NewService(clusterScope, awsSecurityGroupRoles)
+	ec2svc := r.getEC2Service(clusterScope)
+	elbsvc := r.getELBService(*clusterScope)
+	networkSvc := r.getNetworkService(*clusterScope)
+	sgService := r.getSecurityGroupService(*clusterScope)
 
 	if feature.Gates.Enabled(feature.EventBridgeInstanceState) {
 		instancestateSvc := instancestate.NewService(clusterScope)
```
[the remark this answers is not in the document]
Ankitasw: I have intentionally not created it here, as EventBridgeInstanceState is tested in E2E we don't want to test it explicitly here

controllers/awscluster_controller.go
```diff
@@ -68,9 +69,45 @@ var (
 // AWSClusterReconciler reconciles a AwsCluster object.
 type AWSClusterReconciler struct {
 	client.Client
-	Recorder         record.EventRecorder
-	Endpoints        []scope.ServiceEndpoint
-	WatchFilterValue string
+	Recorder              record.EventRecorder
+	ec2ServiceFactory     func(scope.EC2Scope) services.EC2MachineInterface
+	networkServiceFactory func(scope.ClusterScope) services.NetworkInterface
+	elbServiceFactory     func(scope.ClusterScope) services.ELBClusterInterface
```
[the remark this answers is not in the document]
Ankitasw: I will rename it

controllers/awscluster_controller_test.go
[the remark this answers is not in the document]
shivi28: Makes sense, but as it's an existing file so not sure if we should change it

controllers/awscluster_controller.go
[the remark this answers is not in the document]
shivi28: Yes we can test this with E2E/Integration test but if you think we cab add this change in this PR only as you have done for other services.
But it's up to you, you can leave if you want.

controllers/awscluster_controller.go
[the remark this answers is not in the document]
Ankitasw: I will rename it

controllers/awscluster_controller.go
```diff
@@ -159,10 +196,10 @@ func (r *AWSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 func (r *AWSClusterReconciler) reconcileDelete(clusterScope *scope.ClusterScope) (reconcile.Result, error) {
```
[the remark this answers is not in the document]
Ankitasw: IMO the above checks should be done inside during `DeleteBastion()` & `DeleteSecurityGroups()` only, as we are not doing any such sanity checks in controllers.
You might want to add these comments for the pkg `securitygroups` and `bastion` in original issue so that it can be taken care of.

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,637 @@
...
+		name         string
+		awsCluster   *infrav1.AWSCluster
+		ownerCluster *clusterv1.Cluster
+		expectError  bool
+		requeue      bool
+	}{
+		{
+			name:        "Should Reconcile AWSCluster with requeue",
+			awsCluster:  &infrav1.AWSCluster{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("aws-test-%s", util.RandomString(5)), Namespace: "default"}},
+			requeue:     true,
```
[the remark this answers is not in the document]
Ankitasw: I think the meaning of that variable is not interpreted by you properly here, i will change the variable name, it should be shouldRequeue, and if it is false then [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3073/files#diff-c35a4662a40320fd2d431281da35fb99765cfb82809236943751c323004986e7R120) check should be done

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,637 @@
...
+	}
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			g := NewWithT(t)
+			ctx := context.Background()
+			log := ctrl.LoggerFrom(ctx)
+			reconciler := &AWSClusterReconciler{
+				Client: testEnv.Client,
+			}
+			if tc.awsCluster != nil {
```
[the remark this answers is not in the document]
Ankitasw: actually above lines contain defer statement which should be executed after the func execution completes.
If i extract it to other func, it will change the defer behaviour. That's why not extracted to other func

controllers/awscluster_controller_unit_test.go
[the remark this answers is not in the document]
Ankitasw: Done

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,637 @@
...
+func TestAWSClusterReconciler_Reconcile(t *testing.T) {
+	testCases := []struct {
+		name         string
+		awsCluster   *infrav1.AWSCluster
+		ownerCluster *clusterv1.Cluster
+		expectError  bool
+		requeue      bool
+	}{
+		{
+			name:        "Should Reconcile AWSCluster with requeue",
```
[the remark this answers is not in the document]
Ankitasw: Done

controllers/awscluster_controller.go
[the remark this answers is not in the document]
Ankitasw: Done

controllers/awscluster_controller.go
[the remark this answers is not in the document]
Ankitasw: I meant it is already tested in E2E so no need to add more tests explicitly, we are not testing this in integration/unit tests, we need this change only when we use it in integration/unit test

controllers/awscluster_controller.go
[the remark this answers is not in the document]
Ankitasw: Done

controllers/awscluster_controller_test.go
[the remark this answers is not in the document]
Ankitasw: I will revert this change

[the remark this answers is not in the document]
Ankitasw: Done

Ankitasw:
Integration tests added till now:

1.  Reconcile AWSCluster, Should not create VPC if using unmanaged VPC
2.  Reconcile delete AWSCluster, Should delete route tables while deleting VPC
3.  Reconcile AWSCluster, Should fail VPC creation if VPC limit exceeded

sedefsavas:
> 3. Reconcile AWSCluster, Should fail VPC creation if VPC limit exceeded

We can test this in the services package: reconcileVPC()

Ankitasw:
> We can test this in the services package: reconcileVPC()
thanks @sedefsavas I have added it now here, but from next time we will make sure that specific service are tested in the service pkg itself

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,632 @@
...
+		}
+		csClient := fake.NewClientBuilder().WithObjects(awsCluster, secret).Build()
+
+		mockCtrl = gomock.NewController(t)
+		ec2Svc = mock_services.NewMockEC2MachineInterface(mockCtrl)
+		elbSvc = mock_services.NewMockELBInterface(mockCtrl)
+		networkSvc = mock_services.NewMockNetworkInterface(mockCtrl)
+		sgSvc = mock_services.NewMockSecurityGroupInterface(mockCtrl)
+
+		recorder = record.NewFakeRecorder(2)
```
shivi28, line 163: use `32` instead of `2` in NewfakeRecorder

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,632 @@
...
+		reconciler AWSClusterReconciler
+		mockCtrl   *gomock.Controller
+		ec2Svc     *mock_services.MockEC2MachineInterface
+		elbSvc     *mock_services.MockELBInterface
+		networkSvc *mock_services.MockNetworkInterface
+		sgSvc      *mock_services.MockSecurityGroupInterface
+		recorder   *record.FakeRecorder
+	)
+
+	setup := func(t *testing.T, g *WithT, awsCluster *infrav1.AWSCluster) client.WithWatch {
```
shivi28, line 142: No need to pass `g *WithT` in arguments as you are not using it anywhere

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,632 @@
...
+			},
+			securityGroupFactory: func(clusterScope scope.ClusterScope) services.SecurityGroupInterface {
+				return sgSvc
+			},
+			Recorder: recorder,
+		}
+		return csClient
+	}
+
+	teardown := func(t *testing.T, g *WithT) {
```
shivi28, line 184: No need to pass `g *WithT` in arguments as you are not using it anywhere

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,632 @@
...
+		t.Helper()
+		mockCtrl.Finish()
+	}
+
+	t.Run("Reconciling an AWSCluster", func(t *testing.T) {
+		t.Run("reconcile success", func(t *testing.T) {
+			t.Run("should successfully create AWSCluster with Cluster Finalizer and LoadBalancerReady status true on AWSCluster", func(t *testing.T) {
+				g := NewWithT(t)
+				runningCluster := func(t *testing.T, g *WithT) {
```
shivi28, line 193: No need to pass `g *WithT` in arguments as you are not using it anywhere. Consider it at other places also if any

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,632 @@
...
+				g.Expect(err).To(BeNil())
+				expectAWSClusterConditions(g, cs.AWSCluster, []conditionAssertion{{infrav1.LoadBalancerReadyCondition, corev1.ConditionTrue, "", ""}})
+				g.Expect(awsCluster.GetFinalizers()).To(ContainElement(infrav1.ClusterFinalizer))
+			})
+		})
+		t.Run("reconcile failure", func(t *testing.T) {
+			expectedErr := errors.New("failed to get resource")
+			t.Run("should fail AWSCluster create with reconcile network failure", func(t *testing.T) {
+				g := NewWithT(t)
+				awsCluster := getAWSCluster("test", "test")
```
shivi28, line 252: You are creating objects in same ns for multiple test cases, try to avoid it and create objects in separated ns

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,632 @@
...
+				})
+				_, err = reconciler.reconcileNormal(cs)
+				g.Expect(err).To(BeNil())
+				expectAWSClusterConditions(g, cs.AWSCluster, []conditionAssertion{{infrav1.LoadBalancerReadyCondition, corev1.ConditionTrue, "", ""}})
+				g.Expect(awsCluster.GetFinalizers()).To(ContainElement(infrav1.ClusterFinalizer))
+			})
+		})
+		t.Run("reconcile failure", func(t *testing.T) {
+			expectedErr := errors.New("failed to get resource")
+			t.Run("should fail AWSCluster create with reconcile network failure", func(t *testing.T) {
```
shivi28, line 250: use capital letter to start your test case name

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1154 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+
+		capaNS := &corev1.Namespace{
+			ObjectMeta: metav1.ObjectMeta{
+				Name: "capa-system",
+			},
+		}
+		g.Expect(testEnv.Create(ctx, capaNS)).To(Succeed())
+
+		controllerIdentity := createControllerIdentity(g)
+		ns, err := testEnv.CreateNamespace(ctx, fmt.Sprintf("integ-test-%s", util.RandomString(5)))
```
Madhur97, line 99: No need to create a random name ourself as it is already handled by GenerateName. 
Ref: https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#idempotency 
Suggestion: Can we consider using this for naming other resources like AWSCluster during our integration tests. This will safeguard us against name collision.

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1154 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
-func TestAWSClusterReconciler(t *testing.T) {
-	g := NewWithT(t)
-	ctx := context.Background()
+func TestAWSClusterReconciler_IntegrationTests(t *testing.T) {
+	var (
+		reconciler AWSClusterReconciler
+		mockCtrl   *gomock.Controller
+		recorder   *record.FakeRecorder
+	)
+	var secret = &corev1.Secret{
```
Madhur97, line 51: What is the use of this secret in test? 
We are using cluster controller identity, does it need any secret?

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1154 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		ns, err := testEnv.CreateNamespace(ctx, fmt.Sprintf("integ-test-%s", util.RandomString(5)))
+		g.Expect(err).To(BeNil())
+		setup(t, g)
+		awsCluster := getAWSCluster("test", ns.Name)
+		g.Expect(reconciler.Client.Create(ctx, secret)).To(Succeed())
+
+		g.Expect(testEnv.Create(ctx, &awsCluster)).To(Succeed())
+		defer teardown(t, g)
+		defer t.Cleanup(func() {
+			g.Expect(testEnv.Cleanup(ctx, controllerIdentity)).To(Succeed())
```
Madhur97, line 108: testEnv.Cleanup() - This method accepts multiple arguments. Can we check if we can merge two CleanUp() calls into one?

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1154 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		controllerIdentity := createControllerIdentity(g)
+		ns, err := testEnv.CreateNamespace(ctx, fmt.Sprintf("integ-test-%s", util.RandomString(5)))
+		g.Expect(err).To(BeNil())
+		setup(t, g)
+		awsCluster := getAWSCluster("test", ns.Name)
+		g.Expect(reconciler.Client.Create(ctx, secret)).To(Succeed())
+
+		g.Expect(testEnv.Create(ctx, &awsCluster)).To(Succeed())
+		defer teardown(t, g)
+		defer t.Cleanup(func() {
```
Madhur97, line 107: Should we clean up other resources as well like secret, namespace, etc which were created as part of this test?

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1154 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+			g.Expect(testEnv.Cleanup(ctx, controllerIdentity)).To(Succeed())
+			g.Expect(testEnv.Cleanup(ctx, &awsCluster)).To(Succeed())
+		})
+
+		cs, err := scope.NewClusterScope(
```
Madhur97, line 112: We can refactor NewClusterScope into a method as this piece is repeated in all tests.

controllers/awscluster_controller_unit_test.go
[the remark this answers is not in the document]
shivi28: What if  we do something similar to this

```
func cleanupCluster(g *WithT, awsCluster *infrav1.AWSCluster) {
	if awsCluster !=nil{
		g.Expect(testEnv.Cleanup(ctx, awsCluster)).To(Succeed())
	}
}
func setupCluster(g *WithT, awsCluster *infrav1.AWSCluster) {
	if awsCluster != nil {
		awsCluster.Default()
		g.Expect(testEnv.Create(ctx, awsCluster)).To(Succeed())
	}
}
```

And in your test cases just call 
```
setupCluster(g, tc.awsCluster)
defer cleanupCluster(g, tc.awsCluster)
```

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1154 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		reconciler.networkServiceFactory = func(clusterScope scope.ClusterScope) services.NetworkInterface {
+			return networkSvc
+		}
+
+		ec2Svc := ec2Service.NewService(cs)
+		ec2Svc.EC2Client = ec2Mock
+		reconciler.ec2ServiceFactory = func(scope scope.EC2Scope) services.EC2MachineInterface {
+			return ec2Svc
+		}
+		_, err = ec2Svc.EC2Client.CreateSubnet(&ec2.CreateSubnetInput{
```
Madhur97, line 135: What is the purpose of this CreateSubnet() call here?

Madhur97 (commented):
Have given some initial comments. Still reviewing.

controllers/awscluster_controller_test.go
Ankitasw: here we need random naming otherwise the cleanup sometimes does not happen properly.
I can add random name for AWSCluster in integ tests

controllers/awscluster_controller_test.go
Ankitasw: Yes i will combine to single cleanup command

controllers/awscluster_controller_test.go
Ankitasw: I am creating namespace and secret only once so that it will be used across all tests

controllers/awscluster_controller_test.go
Ankitasw: In case of unmanaged VPC, we should have subnet pre created, see the doc [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/book/src/topics/consuming-existing-aws-infrastructure.md#consuming-existing-aws-infrastructure)

controllers/awscluster_controller.go
```diff
@@ -68,9 +69,45 @@ var (
...
+	ec2ServiceFactory     func(scope.EC2Scope) services.EC2MachineInterface
+	networkServiceFactory func(scope.ClusterScope) services.NetworkInterface
+	elbServiceFactory     func(scope.ELBScope) services.ELBInterface
+	securityGroupFactory  func(scope.ClusterScope) services.SecurityGroupInterface
+	Endpoints             []scope.ServiceEndpoint
+	WatchFilterValue      string
+}
+
+// getEC2Service factory func is added for testing purpose so that we can inject mocked EC2Service to the AWSClusterReconciler.
+func (r *AWSClusterReconciler) getEC2Service(scope scope.EC2Scope) services.EC2MachineInterface {
```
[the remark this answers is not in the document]
Ankitasw: I have specified it for testing purpose as the same is followed in AWSMachine controller, and we really did that refactoring for testing purpose.

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1154 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+			},
+		}
+		g.Expect(testEnv.Create(ctx, capaNS)).To(Succeed())
+
+		controllerIdentity := createControllerIdentity(g)
+		ns, err := testEnv.CreateNamespace(ctx, fmt.Sprintf("integ-test-%s", util.RandomString(5)))
+		g.Expect(err).To(BeNil())
+		setup(t, g)
+		awsCluster := getAWSCluster("test", ns.Name)
+		g.Expect(reconciler.Client.Create(ctx, secret)).To(Succeed())
```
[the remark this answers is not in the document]
Ankitasw: it is used in all tests, so created only once. No need to create everytime.

controllers/awscluster_controller_unit_test.go
```diff
@@ -0,0 +1,632 @@
...
+func TestAWSClusterReconciler_Reconcile(t *testing.T) {
+	testCases := []struct {
+		name         string
+		awsCluster   *infrav1.AWSCluster
+		ownerCluster *clusterv1.Cluster
+		expectError  bool
+	}{
+		{
+			name:        "Should not reconcile if owner reference is not set",
+			awsCluster:  &infrav1.AWSCluster{ObjectMeta: metav1.ObjectMeta{Name: fmt.Sprintf("aws-test-%s", util.RandomString(5)), Namespace: "default"}},
```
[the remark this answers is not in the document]
Ankitasw: I think we follow that in E2E tests not for integration tests, but if you have something to support this case, happy to change the namespace for each test.

controllers/awscluster_controller_unit_test.go
Ankitasw: any specific reason for this?

controllers/awscluster_controller_unit_test.go
Ankitasw: answered that above

controllers/awscluster_controller_unit_test.go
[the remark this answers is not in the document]
Ankitasw: I can do that 👍

controllers/awscluster_controller.go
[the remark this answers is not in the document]
shivi28: Not sure if we really need to mention it, we can change at AWSMachine controller file if required

controllers/awscluster_controller_test.go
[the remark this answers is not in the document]
Ankitasw: yes you are right, I am saying it's used across all test cases that's why I created it only in 1st test case, but to avoid your confusion, I will move it to 1st test only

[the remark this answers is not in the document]
Ankitasw: just observed, if I fulfill this [comment](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3073/files#r795777924), then I have to keep the change as is.

controllers/awscluster_controller_test.go
Ankitasw: good point, I think it's used in static identity, I misread it to controller identity, I will remove it.

controllers/awscluster_controller.go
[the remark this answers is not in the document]
Ankitasw: I will let @sedefsavas take a call here

pydctw:
Catching up on the PR, I have some high-level questions.

1) I see some tests use fakeClient while other tests use envTest in awscluster_controller_unit_test.go.
* fakeClient - `TestAWSClusterReconcileOperations`
* envTest - `TestAWSClusterReconciler_Reconcile` and `TestAWSClusterReconciler_RequeueAWSClusterForUnpausedCluster`

Having two different test set up in a single file makes understanding the code difficult. Any reason for this?

2) What differentiates unit vs. integration tests for controllers? 

I see there are three use cases tested in awscluster_controller_test.go (integration tests)
* Reconcile AWSCluster, Should not create VPC if using unmanaged VPC
* Reconcile delete AWSCluster, Should delete route tables, subnets, NAT gateways, EIPs while deleting VPC
* Reconcile AWSCluster, Should fail VPC creation if VPC limit exceeded

With current tests where every single call to AWS API is mocked, 70% of the file is just for mock setups. I feel these can be tested better in each subfunction like vpc_test.go. Any thoughts?

Ankitasw:
Hey @pydctw 

> I see some tests use fakeClient while other tests use envTest in awscluster_controller_unit_test.go.
fakeClient - TestAWSClusterReconcileOperations
envTest - TestAWSClusterReconciler_Reconcile and TestAWSClusterReconciler_RequeueAWSClusterForUnpausedCluster

The reason I have used fakeClient is to reduce the number of mocking calls as you have seen in case of integration tests, and as we have discussed [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3073#issuecomment-1020840327), wherever we can use testenv we are using it.

> What differentiates unit vs. integration tests for controllers?
> I see there are three use cases tested in awscluster_controller_test.go (integration tests)
> 
> Reconcile AWSCluster, Should not create VPC if using unmanaged VPC
> Reconcile delete AWSCluster, Should delete route tables, subnets, NAT gateways, EIPs while deleting VPC
> Reconcile AWSCluster, Should fail VPC creation if VPC limit exceeded

The integration tests actually testing the end to end usecase scenarios of the controllers, such scenarios are not checked during the E2E tests. In unit tests we are only checking functions of the awscluster controller.

> With current tests where every single call to AWS API is mocked, 70% of the file is just for mock setups. I feel these can be tested better in each subfunction like vpc_test.go. Any thoughts?

Since it is recommended to right the integration tests, what I have tried to cover is basic scenario of reconcile case, delete reconcile case, and one is related to VPC limit exceeded, which can be moved to vpc_test as suggested by @sedefsavas [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3073#issuecomment-1023955297), but since I have already written it, and that doesn't contain a lot of mocking, I have kept it here as is.

controllers/awscluster_controller_test.go
Madhur97: Cleanup should not be an issue as API server returns the object with the 'Name' field set. We can use this name during cleanup. This will also solve your above concern for using GenerateName with test objects like AWSCluster.   GenerateName helps to solve the problem of maintaining uniqueness of object names without using any custom naming technique in our tests.

controllers/awscluster_controller_test.go
Madhur97: It is required to precreate these resources when calling actual AWS apis like in case of e2e tests, but here in integration tests we are mocking the AWS calls so we don’t need to add these calls to satisfy AWS related prerequisites. This statement is not doing anything during this test so we can remove it.
Similar concern is there in other integration test - ‘Should fail VPC creation if VPC limit exceeded’ 
`_, err = s.EC2Client.CreateVpc(&ec2.CreateVpcInput{})`
This statement is also not required, we can return error in the mocked CreateVpc() call without making these dummy calls in test.

Ankitasw: If I do not pre-create the subnet here, it gives error and gets stuck at this [point](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/323f0a8506ca9483006e381cc9ac26fb8743dcbe/pkg/cloud/services/ec2/bastion.go#L62-L68). i have added it only to overcome this issue.

controllers/awscluster_controller_test.go
Ankitasw: @Madhur97 I didn't get you point here

Ankitasw: what i meant with cleanup is, t.Cleanup sometimes doesn't clear the objects, that's  why we used random namespaces, and we cannot use random name for AWSCluster as you asked, since we face problem during mocking as mentioned here:

> We can't add random name for AWSCluster in integ tests, as some of the mocking call tags contain AWSCluster name so we won't know what to return as the AWSCluster name is random, here is an example. That's why we are using random namespace.

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1122 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+				ID:               "subnet-1",
+				AvailabilityZone: "us-east-1a",
+				CidrBlock:        "10.0.10.0/24",
+				IsPublic:         false,
+			},
+		})
+		_, err = reconciler.reconcileNormal(cs)
+		g.Expect(err).To(BeNil())
+		g.Expect(cs.VPC().ID).To(Equal("vpc-exists"))
+		expectAWSClusterConditions(g, cs.AWSCluster, []conditionAssertion{
```
Madhur97, line 181: Can we reload the AWSCluster object from API Server and then check the conditions? This will help in ensuring that the object was patched successfully to etcd.

Ankitasw: cs.AWSCluster is shows the exact conditions as it fetches the AWSCluster object from scope. ho do you suggest to reload the object?

controllers/awscluster_controller_unit_test.go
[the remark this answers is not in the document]
shivi28: since we are using testenv in this test, so we should create objects for different test in different namespace, so that they are isolated from each other.

controllers/awscluster_controller_unit_test.go
shivi28: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3073#discussion_r796543557

controllers/awscluster_controller_unit_test.go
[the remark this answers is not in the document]
Ankitasw: To simplify, I will try to create in different namespace.

controllers/awscluster_controller_test.go
Madhur97: I am suggesting to remove custom code for random name generation for namespaces and rely on 'GenerateName' to do that for you. 
Creating random names for other objects like AWSCluster was just a suggestion, but I guess we do not need to that if namespaces are already different.

controllers/awscluster_controller_test.go
Madhur97: > If I do not pre-create the subnet here, it gives error and gets stuck at this point.

Tried it on my system, it is working fine.

controllers/awscluster_controller_test.go
Madhur97:
```
awsCluster := &infrav1.AWSCluster{}
err := r.Get(ctx, req.NamespacedName, awsCluster)
```
We can then use the assertions against this 'awsCluster' object.

Ankitasw: so this is the object we create, but the object gets updated in scope, I tried using this object, it contains no conditions, that's why using scope.AWSCluster

controllers/awscluster_controller_test.go
Ankitasw: I will give it a try.

Ankitasw: seems right, thanks :) maybe with some mocking that changed that I didn't notice. I will remove it.

controllers/awscluster_controller_unit_test.go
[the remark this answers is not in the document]
Ankitasw: Done

controllers/awscluster_controller.go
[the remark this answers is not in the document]
Ankitasw: since this is nit, we can ignore for now

pydctw:
> The reason I have used fakeClient is to reduce the number of mocking calls as you have seen in case of integration tests, and as we have discussed here, wherever we can use testenv we are using it. Since we want to test only a function and not subfunctions, we are using fakeClient in some places so that we don't end up adding expected mock calls for each AWS API.

I am confused by this statement. I've used both fakeClient and envTest before and know that there are pros and cons of using one vs. another. See https://cluster-api.sigs.k8s.io/developer/testing.html?highlight=envtest#envtest.

However, I don't think the choice affects the number of mocking calls. Could you help me understand what mocking calls were avoided by using a fakeClient in TestAWSClusterReconcileOperations?

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1065 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
 
-	reconciler := &AWSClusterReconciler{
-		Client: testEnv.Client,
+	teardown := func() {
+		mockCtrl.Finish()
 	}
 
-	instance := &infrav1.AWSCluster{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"}}
-	instance.Default()
+	t.Run("Should not create VPC in case of unmanaged VPC", func(t *testing.T) {
```
pydctw, line 66: This test case is covered in [vpc_test.go](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/vpc_test.go#L82). As suggested by @sedefsavas previously, vpc specific behaviors should be tested in vpc_test.go.

controllers/awscluster_controller_test.go
```diff
@@ -13,42 +13,1065 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
...
+		g.Expect(err).To(BeNil())
+		g.Expect(cs.VPC().ID).To(Equal("vpc-exists"))
+		expectAWSClusterConditions(g, cs.AWSCluster, []conditionAssertion{
+			{conditionType: infrav1.ClusterSecurityGroupsReadyCondition, status: corev1.ConditionTrue, severity: "", reason: ""},
+			{conditionType: infrav1.BastionHostReadyCondition, status: corev1.ConditionTrue, severity: "", reason: ""},
+			{conditionType: infrav1.VpcReadyCondition, status: corev1.ConditionTrue, severity: "", reason: ""},
+			{conditionType: infrav1.SubnetsReadyCondition, status: corev1.ConditionTrue, severity: "", reason: ""},
+		})
+	})
+	t.Run("Should fail VPC creation if VPC limit exceeded", func(t *testing.T) {
```
pydctw, line 155: Can we also move it to vpc_test.go?

pydctw (changes requested):

sedefsavas:
For integration tests, we should define couple of realistic scenarios. And checking conditions to see if they are as expected is a good signal.

The scenarios I am thinking are:
**For happy path:**
 1- Testing successful AWSCluster creation
2- Testing successful AWSCluster deletion

**For unhappy path:**
 1- Testing failure during AWSCluster creation
2- Testing failure during  AWSCluster deletion

@pydctw please add further suggestion on the unhappy path testing.

Ankitasw:
@sedefsavas For happy path, we already have delete scenario in PR, and the unmanaged VPC test case covers happy path just that it wont reconcile VPC, rest all it flows it goes through, do you suggest to add separate case to PR to test happy path fully?

Ankitasw:
> However, I don't think the choice affects the number of mocking calls. Could you help me understand what mocking calls were avoided by using a fakeClient in TestAWSClusterReconcileOperations?

If you check mocking in `TestAWSClusterReconcileOperations` tests, you would see I have mocked whole subfunctions like ReconcileBastion, ReconcileLoadbalancers. Wherever we want the flow to go through, we wont mock that service, and mock other services for which the flow has not to be tested, this saves lots of AWS API mocking calls as we have seen in integ tests.

controllers/awscluster_controller_test.go
Madhur97: > This will help in ensuring that the object was patched successfully to etcd
As mentioned earlier, it is advisable to run our checks on actual object present in etcd as it confirms that object is patched successfully and it is closer to the real world usecase. You can find similar approach being followed in cluster-api as well.

Ankitasw: not sure if the object gets persisted to etcd in these flows, as some of my tests are failing to get the conditions correctly which were fetched properly from scope

Ankitasw: I think it's better we leave it as is, but if you feel like we should do like this, then we have to find root cause on why no conditions are getting updated in AWSCluster object fetched from etcd

Ankitasw: `[this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/vpc.go#L80)` is one such example where we see that conditions are patched only on scope objects. Likewise, all the other services patch the conditions to scope object in same way. That's why when we try to get AWSCluster object from `testenv.Get` it doesn't contain required condition.

controllers/awscluster_controller_unit_test.go
shivi28: I have seen in cluster API `32` is mainly used as a buffer size for NewFakeRecorder(). If you are sure `2` will work then you can leave it

Ankitasw: atleast in CAPA I see it mostly as 2, so let's leave it for now, we can always increase it later on need basis.

pydctw:
@Ankitasw, first of all, thanks for all the hard work and great discussion.

I had an impression that envTest is a preferred way and CAPI is moving away from fakeClient but after more discussion with some folks (@fabriziopandini, @yastij, @sedefsavas), the recommendation is to pick a right approach depending on test cases. You probably know but to summarize the pros and cons of two approaches here - 

fakeClient 
- Fast and easier to write
- Cannot test resourceVersion, webhooks and etc

envTest
- Closer to the real env with a local kube-apiserver and etcd
- Webhooks and controllers can be registered, hence more realistic uses cases can be tested.
- Cache sync delays need to be accounted for, meaning created/deleted resources can take time to be updated in cache. We've observed tests being flaky in the past and it is why many tests based on envTest uses `g.Eventually` for assertion and help functions like [CreateAndWait](https://github.com/kubernetes-sigs/cluster-api/blob/c014ed1eef86053742e05d42c20d96f0b76e4f7e/internal/test/envtest/environment.go#L363) or [CleanupAndWait](https://github.com/kubernetes-sigs/cluster-api/blob/c014ed1eef86053742e05d42c20d96f0b76e4f7e/internal/test/envtest/environment.go#L329)
- Slower

What I was trying to ask [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3073#issuecomment-1027594239) was the whys of choosing fakeClient vs. envTest for different unit tests (not across unit vs. integration tests).

As long as we understand tradeoffs of using fakeClient vs. envTest and picks an appropriate one, I am ok with the choices. And if we change our mind, we can always iterate and refactor later 😄

pydctw:
As discussed, let's split the integration test out of this PR so that we can merge unit tests first.

Ankitasw:
> What I was trying to ask https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3073#issuecomment-1027418716 was the whys of choosing fakeClient vs. envTest for different unit tests (not across unit vs. integration tests).

@pydctw I hope I answered your question correctly here :) 

> As long as we understand tradeoffs of using fakeClient vs. envTest and picks an appropriate one, I am ok with the choices. And if we change our mind, we can always iterate and refactor later 😄

Thanks @pydctw I hope the current tests are ok for now and I agree we can always refactor it later based on what is good for our tests and how good is the maintainability of mocks

> As discussed, let's split the integration test out of this PR so that we can merge unit tests first.

I will do that

Ankitasw:
The scope of the PR is now changed, added integration tests in separate PR after this PR merges.

controllers/awscluster_controller_test.go
Ankitasw: moved to other PR

controllers/awscluster_controller_test.go
Ankitasw: moved to other PR

controllers/awscluster_controller.go
richardcase, line 82: I'm wondering if we could change the pattern we are using here slightly so that we don't have these extra functions and instead we handle via arguments to `New` . So, something like this (I haven't tested this and will need changing/investigating):

```go
type NewClusterReconcilerInput struct {
	Manager ctrl.Manager
	WatchFilterValue string
	Endpoints []scope.ServiceEndpoint
}

type NewClusterReconcilerServices struct {
	EC2ServiceFactory     func(scope.EC2Scope) services.EC2MachineInterface
	NetworkServiceFactory func(scope.ClusterScope) services.NetworkInterface
	ElbServiceFactory     func(scope.ELBScope) services.ELBInterface
	SecurityGroupFactory  func(scope.ClusterScope) services.SecurityGroupInterface
}

func NewClusterReconciler(input NewClusterReconcilerInput) reconcile.Reconciler {
	return &awsClusterReconciler{
		Client:           input.Manager.GetClient(),
		Recorder:         input.Manager.GetEventRecorderFor("awscluster-controller"),
		Endpoints:        input.Endpoints,
		WatchFilterValue: input.WatchFilterValue,

		ec2ServiceFactory: ec2.NewService,
		elbServiceFactory: elb.NewService,
		networkServiceFactory: network.NewService,
		securityGroupFactory: securitygroup.NewService,
	}
}

func NewClusterReconcilerWithServices(input NewClusterReconcilerInput, svcs NewClusterReconcilerServices ) reconcile.Reconciler {
	return &awsClusterReconciler{
		Client:           input.Manager.GetClient(),
		Recorder:         input.Manager.GetEventRecorderFor("awscluster-controller"),
		Endpoints:        input.Endpoints,
		WatchFilterValue: input.WatchFilterValue,

		ec2ServiceFactory: svcs.EC2ServiceFactory,
		elbServiceFactory: svcs.ElbServiceFactory,
		networkServiceFactory: svcs.NetworkServiceFactory,
		securityGroupFactory: svcs.SecurityGroupFactory,
	}
}

// awsClusterReconciler reconciles a AwsCluster object.
type awsClusterReconciler struct {
	client.Client
	Recorder         record.EventRecorder
	Endpoints        []scope.ServiceEndpoint
	WatchFilterValue string

	ec2ServiceFactory     func(scope.EC2Scope) services.EC2MachineInterface
	networkServiceFactory func(scope.ClusterScope) services.NetworkInterface
	elbServiceFactory     func(scope.ELBScope) services.ELBInterface
	securityGroupFactory  func(scope.ClusterScope) services.SecurityGroupInterface
}
```

And maybe adopt functional options, which could be something along these lines (again not tested and would need to be changed/thought about):

```go
type ReconcilerOption func(*AWSReconciler)

type EC2ServiceFactory func(scope.EC2Scope) services.EC2MachineInterface

func WithEC2ServiceFactory(fn EC2ServiceFactory) ReconcilerOption {
	return func(r *AWSReconciler) {
		r.EC2ServiceFactory  = fn
	}
}


func NewClusterReconciler(input NewClusterReconcilerInput, opt ...ReconcilerOption) reconcile.Reconciler {
	clusterReconciler := &awsClusterReconciler{
		Client:           input.Manager.GetClient(),
		Recorder:         input.Manager.GetEventRecorderFor("awscluster-controller"),
		Endpoints:        input.Endpoints,
		WatchFilterValue: input.WatchFilterValue,

		EC2ServiceFactory: ec2.NewService,
		ELBServiceFactory: elb.NewService,
		NetworkServiceFactory: network.NewService,
		SecurityGroupFactory: securitygroup.NewService,
	}

	for _, opt := range opt {
		opt(clusterReconciler)
	}

	return clusterReconciler
}
```

sedefsavas: @richardcase 
A question to understand better:
Reconcilers are created in main and we do not have the session created there yet. How do we create `ec2.NewService` in this case when we call NewClusterReconciler()?

AWSMachineReconciler have serviceFactories in its struct, which is better in the sense that we don't have to call `ec2.NewService()` multiple times for reconcileNormal() and reconcileDelete().

sedefsavas: Do you want to track this in a new issue? We will probably need to do changes everywhere if we want to proceed. So out of scope of this PR.

richardcase: > Reconcilers are created in main and we do not have the session created there yet. How do we create ec2.NewService in this case when we call NewClusterReconciler()?

We don't create an instance of the services in or when calling `NewClusterReconciler` but instead we pass in functions. So in this example, we are passing in the function `ec2.NewService` and not calling it `ec2.NewService(scope)`. The function we pass in should adhere to the same factory function definition. We can then call the function later when we have the scope.  But it also allows us to override the factory function when we create an new instance of the reconciler in the tests. And if we don't pass in an override it can default to what we want outside of testing, `ec2.NewService` in this example.

Instead of having logic like this in the controllers:

```go
func (r *AWSClusterReconciler) getEC2Service(scope scope.EC2Scope) services.EC2MachineInterface {
	if r.ec2ServiceFactory != nil {
		return r.ec2ServiceFactory(scope)
	}
	return ec2.NewService(scope)
}
```

it feels to me that we should handle this during creation of the controller (i.e. via a concreational pattern) and so anywhere we need an ec2 service in the reconciler we just go `r.ec2ServiceFactory(scope)`.

richardcase: > Do you want to track this in a new issue? We will probably need to do changes everywhere if we want to proceed with the change. So out of scope of this PR.

These changes are being made to aid testability so i would argue its in scope but i should've picked up on this earlier and as you point out it requires changes across a number of controllers. 

So, i will create a new issue and follow up after these testing PRs merge :+1:

Ankitasw: @richardcase @sedefsavas so shall we move this discussion to a new issue and go ahead with the PR changes?

richardcase: Yes i will create the new issue with the above :)

Ankitasw: cc @pydctw

PR #3099, opened 2022-01-23
Added unit test cases for launchtemplate.go

### LINKED ISSUE #2910: Increase unit test coverage

shivi28:
**What type of PR is this?**
/area testing

This PR increases the unit tests coverage for launchtemplate.go file.

Part of #2910 

Kindly **add any missing scenario in the original issue #2910**. Decided to track all missing scenarios in future PR

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Added unit test cases for launchtemplate.go and increased coverage to 82%
```

### DIFF
--- a/pkg/cloud/services/ec2/ami_test.go
+++ b/pkg/cloud/services/ec2/ami_test.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/ssm"
 	"github.com/golang/mock/gomock"
 	. "github.com/onsi/gomega"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/awserrors"
@@ -184,10 +185,16 @@ func TestAMIs(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewWithT(t)
 
+			scheme, err := setupScheme()
+			g.Expect(err).NotTo(HaveOccurred())
+			client := fake.NewClientBuilder().WithScheme(scheme).Build()
+
 			ec2Mock := mock_ec2iface.NewMockEC2API(mockCtrl)
 			tc.expect(ec2Mock.EXPECT())
 
-			clusterScope := setupClusterScope(g)
+			clusterScope, err := setupClusterScope(client)
+			g.Expect(err).NotTo(HaveOccurred())
+
 			s := NewService(clusterScope)
 			s.EC2Client = ec2Mock
 
@@ -443,12 +450,18 @@ func TestEKSAMILookUp(t *testing.T) {
 		t.Run(tt.name, func(t *testing.T) {
 			g := NewWithT(t)
 
+			scheme, err := setupScheme()
+			g.Expect(err).NotTo(HaveOccurred())
+			client := fake.NewClientBuilder().WithScheme(scheme).Build()
+
 			ssmMock := mock_ssmiface.NewMockSSMAPI(mockCtrl)
 			if tt.expect != nil {
 				tt.expect(ssmMock.EXPECT())
 			}
 
-			clusterScope := setupClusterScope(g)
+			clusterScope, err := setupClusterScope(client)
+			g.Expect(err).NotTo(HaveOccurred())
+
 			s := NewService(clusterScope)
 			s.SSMClient = ssmMock
 
--- a/pkg/cloud/services/ec2/helper_test.go
+++ b/pkg/cloud/services/ec2/helper_test.go
@@ -17,33 +17,97 @@ limitations under the License.
 package ec2
 
 import (
-	. "github.com/onsi/gomega"
+	"sort"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
+	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/controlplane/eks/api/v1beta1"
 	expinfrav1 "sigs.k8s.io/cluster-api-provider-aws/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/exp/api/v1beta1"
 )
 
-func setupScheme() (*runtime.Scheme, error) {
-	scheme := runtime.NewScheme()
-	if err := clusterv1.AddToScheme(scheme); err != nil {
-		return nil, err
+func setupClusterScope(cl client.Client) (*scope.ClusterScope, error) {
+	return scope.NewClusterScope(scope.ClusterScopeParams{
+		Client:     cl,
+		Cluster:    newCluster(),
+		AWSCluster: newAWSCluster(),
+	})
+}
+
+func setupNewManagedControlPlaneScope(cl client.Client) (*scope.ManagedControlPlaneScope, error) {
+	return scope.NewManagedControlPlaneScope(scope.ManagedControlPlaneScopeParams{
+		Client:       cl,
+		Cluster:      newCluster(),
+		ControlPlane: newAWSManagedControlPlane(),
+	})
+}
+
+func setupMachinePoolScope(cl client.Client, ec2Scope scope.EC2Scope) (*scope.MachinePoolScope, error) {
+	return scope.NewMachinePoolScope(scope.MachinePoolScopeParams{
+		Client:         cl,
+		InfraCluster:   ec2Scope,
+		Cluster:        newCluster(),
+		MachinePool:    newMachinePool(),
+		AWSMachinePool: newAWSMachinePool(),
+	})
+}
+
+func defaultEC2Tags(name, clusterName string) []*ec2.Tag {
+	return []*ec2.Tag{
+		{
+			Key:   aws.String("Name"),
+			Value: aws.String(name),
+		},
+		{
+			Key:   aws.String(infrav1.ClusterAWSCloudProviderTagKey(clusterName)),
+			Value: aws.String("owned"),
+		},
+		{
+			Key:   aws.String(infrav1.ClusterTagKey(clusterName)),
+			Value: aws.String("owned"),
+		},
+		{
+			Key:   aws.String(infrav1.NameAWSClusterAPIRole),
+			Value: aws.String("node"),
+		},
 	}
-	if err := corev1.AddToScheme(scheme); err != nil {
-		return nil, err
+}
+
+func newAWSMachinePool() *expinfrav1.AWSMachinePool {
+	return &expinfrav1.AWSMachinePool{
+		TypeMeta: metav1.TypeMeta{
+			Kind:       "AWSMachinePool",
+			APIVersion: "v1",
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      "aws-mp-name",
+			Namespace: "aws-mp-ns",
+		},
+		Spec: expinfrav1.AWSMachinePoolSpec{
+			AvailabilityZones: []string{"us-east-1"},
+			AdditionalTags:    infrav1.Tags{},
+			AWSLaunchTemplate: expinfrav1.AWSLaunchTemplate{
+				Name:                     "aws-launch-template",
+				IamInstanceProfile:       "instance-profile",
+				AMI:                      infrav1.AMIReference{},
+				InstanceType:             "t3.large",
+				SSHKeyName:               aws.String("default"),
+				AdditionalSecurityGroups: []infrav1.AWSResourceReference{{ID: aws.String("1")}},
+			},
+		},
+		Status: expinfrav1.AWSMachinePoolStatus{
+			LaunchTemplateID: "launch-template-id",
+		},
 	}
-	if err := infrav1.AddToScheme(scheme); err != nil {
-		return nil, err
-	}
-	if err := expinfrav1.AddToScheme(scheme); err != nil {
-		return nil, err
-	}
-	return scheme, nil
 }
 
 func newCluster() *clusterv1.Cluster {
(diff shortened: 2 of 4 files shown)


pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -100,45 +121,151 @@ func TestGetLaunchTemplate(t *testing.T) {
...
+				if userDataHash != "" {
+					t.Fatalf("Did not expect a userdata hash, but got something: %s", userDataHash)
+				}
+				if launchTemplate != nil {
+					t.Fatalf("Did not expect anything but got something: %+v", launchTemplate)
+				}
+			},
+		},
+		{
+			name:               "successfully fetched launch template versions",
```
Ankitasw, line 185: In success case, is it possible to add the following scenario:
Should add additionalSecurityGroups in launch template if `DescribeLaunchTemplateVersions` returns output with additionalSecurityGroup.
No need for separate test case, you can add this in existing test `successfully fetched launch template versions`. Wdyt?

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -100,45 +121,151 @@ func TestGetLaunchTemplate(t *testing.T) {
...
+					t.Fatalf("launchTemplate mismatch: got %v, want %v", launchTemplate, wantLT)
+				}
+				if !reflect.DeepEqual(userDataHash, testUserDataHash) {
+					t.Fatalf("userDataHash mismatch: got %v, want %v", userDataHash, testUserDataHash)
+				}
+				if err != nil {
+					t.Fatalf("did not expect error: %v", err)
 				}
 			},
 		},
```
Ankitasw, line 248: Can we also add a scenario:
Should return computed userdata if userdata is nil in launchtemplate

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -309,3 +436,850 @@ func TestService_LaunchTemplateNeedsUpdate(t *testing.T) {
 		})
 	}
 }
+
```
Ankitasw, line 439: In TestService_LaunchTemplateNeedsUpdate func, is it possible to add following scenarios:
- Should return true if incoming IamInstanceProfile is not same as existing IamInstanceProfile
- Should return true if incoming InstanceType is not same as existing InstanceType
- Should return false if failed to get core node security groups

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -309,3 +436,850 @@ func TestService_LaunchTemplateNeedsUpdate(t *testing.T) {
...
+		}
+	}
+
+	var userData = []byte{1, 0, 0}
+	testCases := []struct {
+		name   string
+		expect func(scope *mocks.MockEC2ScopeMockRecorder, m *mock_ec2iface.MockEC2APIMockRecorder)
+		check  func(s string, e error)
+	}{
+		{
```
Ankitasw, line 684: Is it possible to add following scenario here:
Should fail to form launch template data

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -83,13 +87,30 @@ func TestGetLaunchTemplate(t *testing.T) {
...
 		name               string
 		launchTemplateName string
-		expect             func(m *mock_ec2iface.MockEC2APIMockRecorder)
-		check              func(launchtemplate *expinfrav1.AWSLaunchTemplate, userdatahash string, err error)
+		expect             func(scope *mocks.MockEC2ScopeMockRecorder, m *mock_ec2iface.MockEC2APIMockRecorder)
+		check              func(launchTemplate *expinfrav1.AWSLaunchTemplate, userDataHash string, err error)
 	}{
 		{
-			name:               "does not exist",
+			name:   "empty launch template name passed",
```
Ankitasw, line 94: nit: unit tests naming conventions cam be taken care in this file. You can use any way from the [reference](https://medium.com/@stefanovskyi/unit-test-naming-conventions-dd9208eadbea).
I generally follow `Should_ExpectedBehavior_If_StateUnderTest`

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -309,3 +436,850 @@ func TestService_LaunchTemplateNeedsUpdate(t *testing.T) {
...
+	}
+
+	var userData = []byte{1, 0, 0}
+	testCases := []struct {
+		name    string
+		imageID *string
+		expect  func(scope *mocks.MockEC2ScopeMockRecorder, m *mock_ec2iface.MockEC2APIMockRecorder)
+		check   func(e error)
+	}{
+		{
```
Ankitasw, line 857: Is it possible to add following scenario here:
Should fail to form launch template data

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -309,3 +436,850 @@ func TestService_LaunchTemplateNeedsUpdate(t *testing.T) {
...
+	defer mockCtrl.Finish()
+
+	testCases := []struct {
+		name              string
+		awsLaunchTemplate expinfrav1.AWSLaunchTemplate
+		machineTemplate   clusterv1.MachineTemplateSpec
+		expect            func(m *mock_ec2iface.MockEC2APIMockRecorder)
+		check             func(*string, error)
+	}{
+		{
```
Ankitasw, line 1074: Is it possible to dd following scenarios here:
- Should fetch imageLookupFormat if not present in launchtemplate
- Should fetch ImageLookupOrg if not present in launchtemplate
- Should fetch ImageLookupBaseOS if not present in launchtemplate
- Should do eks EMI lookup in case of EKSManaged cluster(you can think of covering eksAMILookup() func here with different combinations to cover all edge cases)

Ankitasw (changes requested):

Ankitasw:
@shivi28 You can also check past issues occurred due to issues in launchtemplate and cover those as well in these tests(like integration tests)

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -309,3 +436,850 @@ func TestService_LaunchTemplateNeedsUpdate(t *testing.T) {
...
+			scheme := runtime.NewScheme()
+			_ = infrav1.AddToScheme(scheme)
+			client := fake.NewClientBuilder().WithScheme(scheme).Build()
+
+			scope, err := scope.NewClusterScope(scope.ClusterScopeParams{
+				Client:     client,
+				Cluster:    &clusterv1.Cluster{},
+				AWSCluster: &infrav1.AWSCluster{},
+			})
+			if err != nil {
```
alexander-demichev, line 1273: what do you think about using `gomega` matchers? like here https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/50af6ec5ce7a183b67d23a187e687c42fef24176/pkg/cloud/services/iamauth/configmap_test.go#L150

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -309,3 +436,850 @@ func TestService_LaunchTemplateNeedsUpdate(t *testing.T) {
...
+				m.DeleteLaunchTemplateVersions(gomock.Eq(
+					&ec2.DeleteLaunchTemplateVersionsInput{
+						LaunchTemplateId: aws.String("id"),
+						Versions:         aws.StringSlice([]string{"12"}),
+					},
+				)).Return(nil, nil)
+			},
+			check: func(err error) {
+				if err != nil {
+					t.Fatalf("did not expect an erro, but got %+v", err)
```
alexander-demichev, line 1254:
```suggestion
					t.Fatalf("did not expect an error, but got %+v", err)
```

Ankitasw:
@shivi28 You can change it Part of #2910, otherwise the parent issue will get closed with this PR merge.

> Which issue(s) this PR fixes (optional, in fixes #(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
> Part of Fixes #2910

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: done

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: done

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: done

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: done

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: Covered 1 and 2

> Should return false if failed to get core node security groups
This piece of code is covered in other test case

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: Added a test case `Test_LaunchTemplateDataCreation `

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: done

pkg/cloud/services/ec2/ami_test.go
```diff
@@ -184,10 +185,16 @@ func TestAMIs(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			g := NewWithT(t)
 
+			scheme, err := setupScheme()
+			g.Expect(err).NotTo(HaveOccurred())
+			client := fake.NewClientBuilder().WithScheme(scheme).Build()
```
Ankitasw, line 190: Can we extract a func out of these lines as same is used in line 453-456

pkg/cloud/services/ec2/instances.go
```diff
@@ -420,13 +420,16 @@ func (s *Service) GetCoreNodeSecurityGroups(scope *scope.MachinePoolScope) ([]st
 	}
 
 	ids := make([]string, 0, len(sgRoles))
-	for _, sg := range sgRoles {
-		if _, ok := s.scope.SecurityGroups()[sg]; !ok {
-			return nil, awserrors.NewFailedDependency(
-				fmt.Sprintf("%s security group not available", sg),
-			)
+	existingSGs := s.scope.SecurityGroups()
+	if len(existingSGs) != 0 {
```
Ankitasw, line 424: Generally we follow convention len(<variable>) > 0

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -82,10 +87,18 @@ func TestGetLaunchTemplate(t *testing.T) {
...
-		check              func(launchtemplate *expinfrav1.AWSLaunchTemplate, userdatahash string, err error)
+		check              func(g *WithT, launchTemplate *expinfrav1.AWSLaunchTemplate, userDataHash string, err error)
 	}{
 		{
-			name:               "does not exist",
+			name: "Should not return launch template if empty launch template name passed",
+			check: func(g *WithT, launchTemplate *expinfrav1.AWSLaunchTemplate, userDataHash string, err error) {
+				g.Expect(err).NotTo(HaveOccurred())
+				g.Expect(userDataHash).Should(BeEmpty())
+				g.Expect(launchTemplate).Should(BeNil())
```
Ankitasw, line 97: Can we extract this also to a function as it is used in multiple places?

pkg/cloud/services/ec2/ami_test.go
shivi28: Actually I have already refactored it using `setupScheme()` method and I have checked other test files also and kept it like this to maintain the consistency.

pkg/cloud/services/ec2/launchtemplate_test.go
shivi28: Yes I could do that if we are doing same check in all the test cases but in last 2 test case I am doing different check. So that's why I thought to put it in a separate `check field`.
```
g.Expect(err).NotTo(HaveOccurred())
g.Expect(userDataHash).Should(Equal(testUserDataHash))
g.Expect(launchTemplate).Should(Equal(wantLT))
```

shivi28: Yes I could do that if we are doing same check in all the test cases but in last 2 test case I am doing different check. So that's why I thought to put it in a separate check field.
```
g.Expect(err).NotTo(HaveOccurred())
g.Expect(userDataHash).Should(Equal(testUserDataHash))
g.Expect(launchTemplate).Should(Equal(wantLT))
```

pkg/cloud/services/ec2/instances.go
Ankitasw: is this addressed?

shivi28: done

pkg/cloud/scope/mocks/ec2_mock.go
```diff
@@ -0,0 +1,525 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Code generated by MockGen. DO NOT EDIT.
+// Source: ../ec2.go
```
sedefsavas, line 18: Where is this new mock being used?

shivi28: Initially thought to create all mock files in scope folder, but later found it should be within each service package.
So I forgot to cleanup this, but now updated the PR. Thanks

pkg/cloud/services/ec2/helper_test.go
```diff
@@ -80,18 +147,59 @@ func newAWSCluster() *infrav1.AWSCluster {
 	}
 }
 
-func setupClusterScope(g *WithT) *scope.ClusterScope {
-	scheme, err := setupScheme()
-	g.Expect(err).NotTo(HaveOccurred())
+func newAWSClusterForEKS() *infrav1.AWSCluster {
```
richardcase, line 150: This is the wrong type, it needs to be `AWSManagedControlPlane`

shivi28: Thanks @richardcase , it makes sense. I have updated the PR

pkg/cloud/services/ec2/instances.go
```diff
@@ -420,13 +420,16 @@ func (s *Service) GetCoreNodeSecurityGroups(scope *scope.MachinePoolScope) ([]st
 	}
 
 	ids := make([]string, 0, len(sgRoles))
-	for _, sg := range sgRoles {
-		if _, ok := s.scope.SecurityGroups()[sg]; !ok {
-			return nil, awserrors.NewFailedDependency(
-				fmt.Sprintf("%s security group not available", sg),
-			)
+	existingSGs := s.scope.SecurityGroups()
```
richardcase, line 423: I'm curious as to why this part has been changed as part of this?

shivi28: There are 2 reasons for this
1. In old code we are accessing `s.scope.SecurityGroups()[sg]` where `s.scope.SecurityGroups()` is a slice. And as in former condition we are checking length of `s.scope.SecurityGroups()` so it might given index out of range exception if `len(s.scope.SecurityGroups()) == 0`
2. And as we are calling `s.scope.SecurityGroups()` multiple times so thought to store it in a variable `exisitingSGs`

Ankitasw: > In old code we are iterating over sgRoles slice and then check if that role exisit in existingSGs or not. But in case of len(existingSGs)==0 then there is no need to range over sgRoles slice.

If the `sgRoles` length is 0, then automatically it wont range over `sgroles` and skip this block, so I think we dont need to do explicit check for length.

shivi28: Thanks everyone for suggestion. Now it makes sense to me as `s.scope.SecurityGroups()` is only accessing infra cluster spec, so no issues in calling it multiple times and also no use of checking length for existingSGs as when there is no element in the map then accessing its element return ok=false.

sedefsavas:
/approve

/hold

until tests pass.

PR #3103, opened 2022-01-24
Replace predicate functions with CAPI util functions

pydctw:
**What type of PR is this?**
/kind cleanup

While reading the code, noticed that awsmachine and awscluster controllers use its own predicate functions for watching a Cluster that could be replaced with CAPI util functions. 

Note that we are already using them in [awsmanagedcontrolplane_controller.go](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/50af6ec5ce7a183b67d23a187e687c42fef24176/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go#L94)

Continuously refactoring the code to make it cleaner and easy to read.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
NONE
```

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -324,45 +324,7 @@ func (r *AWSClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 	return controller.Watch(
 		&source.Kind{Type: &clusterv1.Cluster{}},
 		handler.EnqueueRequestsFromMapFunc(r.requeueAWSClusterForUnpausedCluster(ctx, log)),
-		predicate.Funcs{
-			UpdateFunc: func(e event.UpdateEvent) bool {
-				oldCluster := e.ObjectOld.(*clusterv1.Cluster)
-				newCluster := e.ObjectNew.(*clusterv1.Cluster)
-				log := log.WithValues("predicate", "updateEvent", "namespace", newCluster.Namespace, "cluster", newCluster.Name)
-				switch {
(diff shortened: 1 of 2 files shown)


controllers/awscluster_controller.go
```diff
@@ -324,45 +324,7 @@ func (r *AWSClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
...
-				log.V(4).Info("Cluster did not match expected conditions, will not attempt to map associated AWSCluster.")
-				return false
-			},
-			GenericFunc: func(e event.GenericEvent) bool {
-				log := log.WithValues("predicate", "genericEvent", "namespace", e.Object.GetNamespace(), "cluster", e.Object.GetName())
-				log.V(4).Info("Cluster did not match expected conditions, will not attempt to map associated AWSCluster.")
-				return false
-			},
-		},
+		predicates.ClusterUnpaused(log),
```
pydctw, line 327: @sedefsavas, following up our discussion, 

For AWSCluster, `ClusterUnpaused` function should be good enough as [CAPI gets InfrastructureReady information from AWSCluster](https://github.com/kubernetes-sigs/cluster-api/blob/main/internal/controllers/cluster/cluster_controller_phases.go#L166) and AWSCluster controller is already watching itself, meaning this is already known state.

So we don't need to use `ClusterUnpausedAndInfrastructureReady` function in this case.

sedefsavas: Thanks for confirming.

Issue #3104, opened 2022-01-26
unknown field "emeritus_maintainers" in the OWNERS file


neolit123:
there is a [new tool](https://github.com/dims/maintainers) for parsing the kubernetes/community/sigs.yaml file and it reports an error in the OWNERS file in this repository:
> ERROR: unable to parse owners file at https://raw.githubusercontent.com/kubernetes-sigs/cluster-api-provider-aws/master/OWNERS url - error unmarshaling JSON: while decoding JSON: json: unknown field "emeritus_maintainers"

**solution**

1) move the entries under `emeritus_maintainers` https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/OWNERS#L19-L24 to `emeritus_approvers`.

2) remove the `emeritus_maintainers` key.

neolit123:
/kind bug
/good-first-issue
/help
(see instructions in the OP)

Skarlso:
@neolit123 Hi. :)

How likely is this change going to break existing tooling which isn't updated yet to use this new tooling?

I wonder if there is a way of rather ignoring certain fields or having both in the file until such a time comes when only the new tooling is used?

Skarlso:
Would have been helpful to include that it's the `export` command that fails... :)

Skarlso:
I feel like this tool is too new to be taken seriously. It has zero tests, no validations, no guidelines, PR checks, releases or relevant information. How come this was picked up or why is it not under the kube-sigs project structure.

Skarlso:
Sorry, I didn't mean to sound so pendantic. Just please help me understand the adoption process of a new tool in this ecosystem. 😊

richardcase:
@Skarlso - when you make this change could you move the existing `emeritus_approvers` into the `emeritus_reviewers` list? And then change `emeritus_maintainers` to `emeritus_approvers`?

Skarlso:
Yes, and thank you for the background and clarifications! :)

neolit123:
It can take some time before a tool like that becomes stable. I think what
matters in this case is that there is no such field (_maintainers) defined
in the YAML spec of the bot that reads the owners files.

Skarlso:
`emeritus_reviewers` -> is also an unknown field

neolit123:
Related issue https://github.com/kubernetes/community/issues/6334

Skarlso:
[...]

PR #3116, opened 2022-01-28
Fix markdown mentions to proper links

Skarlso:
Follow up to https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3104 which fixes the mentions, which is not a supported thing in Markdown, to be proper links. :)

```release-note
Fix the list of contributors in the README to be links instead of @ mentions.
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,20 +185,20 @@ and/or other countries."
 Thank you to all contributors and a special thanks to:
 
 emeritus reviewers:
-- @ashish-amarnath
-- @davidewatson
-- @enxebre
-- @ingvagabund
-- @michaelbeaumont
-- @sethp-nr
+- [@ashish-amarnath](https://github.com/ashish-amarnath)
+- [@davidewatson](https://github.com/davidewatson)
+- [@enxebre](https://github.com/enxebre)
+- [@ingvagabund](https://github.com/ingvagabund)
+- [@michaelbeaumont](https://github.com/michaelbeaumont)
+- [@sethp-nr](https://github.com/sethp-nr)
 
 emeritus approvers/maintainers:
-- @chuckha
-- @detiber
-- @ncdc
-- @randomvariable
-- @rudoi
-- @vincepri
+- [@chuckha](https://github.com/chuckha)
+- [@detiber](https://github.com/detiber)
+- [@ncdc](https://github.com/ncdc)
+- [@randomvariable](https://github.com/randomvariable)
+- [@rudoi](https://github.com/rudoi)
+- [@vincepri](https://github.com/vincepri)
 
 <!-- References -->
 [slack]: https://kubernetes.slack.com/messages/CD6U2V71N


README.md
```diff
@@ -185,20 +185,20 @@ and/or other countries."
 Thank you to all contributors and a special thanks to:
 
 emeritus reviewers:
```
pydctw, line 187: I thought emeritus_reviewers doesn't exist in OWNERs file.

README.md
```diff
@@ -185,20 +185,20 @@ and/or other countries."
...
-- @michaelbeaumont
-- @sethp-nr
+- [@ashish-amarnath](https://github.com/ashish-amarnath)
+- [@davidewatson](https://github.com/davidewatson)
+- [@enxebre](https://github.com/enxebre)
+- [@ingvagabund](https://github.com/ingvagabund)
+- [@michaelbeaumont](https://github.com/michaelbeaumont)
+- [@sethp-nr](https://github.com/sethp-nr)
 
 emeritus approvers/maintainers:
```
pydctw, line 195: Do we need to keep emeritus approvers section on README? I would rather give special thanks to current maintainers, @sedefsavas and @richardcase 😀

If we decide to keep them, let's change `emeritus approvers/maintainers` to `emeritus approvers` as that was what's requested in #3104.

README.md
Skarlso: Correct. The new tool does not parse those fields, but Richard thought that we still should thank these people, so we added them to README instead. :)

README.md
Skarlso: Technically, they are maintainers, but also people who where approvers. OWNERS has now the correct format, but missing the right section for these people we invented these new sections so they don't get lost and are still thanked. :)

README.md
pydctw: Thanks for the context. I didn't know about that discussion - I am fine with the decision.

Skarlso: Yeah, sorry about that, should have put more context into the ticket. It was all in my head only. :D

sedefsavas: This list is not aligning with the OWNER file's content. It'd be good to make them consistent (and then delete those from the OWNERs list in a follow-up PR).

Skarlso: It does not, because they are different people. We had to rename the reviewers to approvers and remove the approvers and the maintainers. :D And put them into the Readme. Because the new tool was unable to parse either. :/

sedefsavas: There is `emeritus_reviewers:` section in the OWNERS file.

Skarlso: Are you sure you are looking at the right revision? We removed that section in a previous PR because of the breaking owners tool.

sedefsavas: Yes looking at the PR https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3115/files merged.
My comment holds, we can use the list that is deleted in the PR merged.

Skarlso: Ah you mean there was a reviewers section. :D Okay, so, we merged that with, errr... the maintainers and pull them together that's the discrepancy.  But honestly, I'm a bit confused now here. :D I'm hoping @richardcase can shed a bit of better light on this. :)

sedefsavas: I just saw the comments from the other issue, ok this is the merged list. LGTM. Thanks for the PR.

PR #3117, opened 2022-01-28
Add unit tests for package pkg/cloud/tags

Ankitasw:
**What type of PR is this?**
/area testing

This PR increases the unit tests coverage for package pkg/cloud/tags. The coverage of package has increased from 3.9% to 98%

Part of #2910 

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Add unit tests for package pkg/cloud/tags
```

### DIFF
--- a/pkg/cloud/tags/tags.go
+++ b/pkg/cloud/tags/tags.go
@@ -78,6 +78,9 @@ func (b *Builder) Apply() error {
 
 // Ensure applies the tags if the current tags differ from the params.
 func (b *Builder) Ensure(current infrav1.Tags) error {
+	if b.params == nil {
+		return ErrBuildParamsRequired
+	}
 	if diff := computeDiff(current, *b.params); len(diff) > 0 {
 		return b.Apply()
 	}
--- a/pkg/cloud/tags/tags_test.go
+++ b/pkg/cloud/tags/tags_test.go
@@ -20,7 +20,44 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/eks"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
+
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/ec2/mock_ec2iface"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/eks/mock_eksiface"
+)
+
+var (
+	bp = infrav1.BuildParams{
+		Lifecycle:   infrav1.ResourceLifecycleOwned,
+		ClusterName: "testcluster",
+		Name:        aws.String("test"),
+		Role:        aws.String("testrole"),
+		Additional:  map[string]string{"k1": "v1"},
+	}
+	tags = []*ec2.Tag{
+		{
+			Key:   aws.String("Name"),
+			Value: aws.String("test"),
+		},
+		{
+			Key:   aws.String("k1"),
+			Value: aws.String("v1"),
+		},
+		{
+			Key:   aws.String("sigs.k8s.io/cluster-api-provider-aws/cluster/testcluster"),
+			Value: aws.String("owned"),
+		},
+		{
+			Key:   aws.String("sigs.k8s.io/cluster-api-provider-aws/role"),
+			Value: aws.String("testrole"),
+		},
+	}
 )
 
 func TestTags_ComputeDiff(t *testing.T) {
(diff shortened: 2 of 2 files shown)


pkg/cloud/tags/tags.go
```diff
@@ -78,6 +78,9 @@ func (b *Builder) Apply() error {
 
 // Ensure applies the tags if the current tags differ from the params.
 func (b *Builder) Ensure(current infrav1.Tags) error {
+	if b.params == nil {
```
sedefsavas, line 81: This check is already in `Apply()`.
`Ensure()` does not need to have this check as it checks for diff. Having a nil params will cause diff to be empty and return with success (this is not valid use case, we always have tags but still, I don't think this check is the responsibility of this method)

pkg/cloud/tags/tags_test.go
```diff
@@ -94,3 +129,73 @@ func TestTags_ComputeDiff(t *testing.T) {
 		})
 	}
 }
+
+func TestTags_Ensure(t *testing.T) {
+	tests := []struct {
+		name          string
+		builder       Builder
+		builderOption BuilderOption
```
sedefsavas, line 137: This is not used in the tests.

pkg/cloud/tags/tags_test.go
```diff
@@ -94,3 +129,73 @@ func TestTags_ComputeDiff(t *testing.T) {
...
+	}
+
+	mockCtrl := gomock.NewController(t)
+	ec2Mock := mock_ec2iface.NewMockEC2API(mockCtrl)
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			var builder *Builder
+			if tc.expect != nil {
+				tc.expect(ec2Mock.EXPECT())
+				builder = New(tc.builder.params, WithEC2(ec2Mock))
```
sedefsavas, line 180: Would be good to add WithEKS() test as it is in the same file.

pkg/cloud/tags/tags_test.go
```diff
@@ -20,7 +20,42 @@ import (
 	"reflect"
 	"testing"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/golang/mock/gomock"
+	"github.com/pkg/errors"
+	"github.com/stretchr/testify/assert"
```
sedefsavas, line 27: To be consistent with the rest of the tests, cluster api ecosystem is using gomega assertions.
cc @shivi28 @Madhur97

pkg/cloud/tags/tags.go
Ankitasw: The reason this check has been added here, if you see this line, it uses `*b.params` as input to `computeDiff()`, and if we pass `b.params` as `nil`, we will get panic here, hence to avoid this, `b.params` should be check first, before computing the diff

pkg/cloud/tags/tags_test.go
Ankitasw: Done

pkg/cloud/tags/tags_test.go
Ankitasw: Added the tests

pkg/cloud/tags/tags_test.go
Ankitasw: Removed

pkg/cloud/tags/tags.go
sedefsavas: Makes sense, did not see that we were dereferencing it.

Ankitasw: Done

pkg/cloud/tags/tags_test.go
```diff
@@ -94,3 +131,121 @@ func TestTags_ComputeDiff(t *testing.T) {
...
+}
+
+func TestTags_EnsureWithEKS(t *testing.T) {
+	tests := []struct {
+		name    string
+		builder Builder
+		expect  func(m *mock_eksiface.MockEKSAPIMockRecorder)
+	}{
+		{
+			name:    "Should return error when applyFunc params fails",
```
sedefsavas, line 200: What does this mean?
Maybe "Should return error when tags apply function fails"?

Using `applyFunc` in test names is not descriptive enough.

Ankitasw: will add appropriate names

Ankitasw: Done

pkg/cloud/tags/tags.go
```diff
@@ -63,9 +63,6 @@ func New(params *infrav1.BuildParams, opts ...BuilderOption) *Builder {
 
 // Apply tags a resource with tags including the cluster tag.
 func (b *Builder) Apply() error {
-	if b.params == nil {
```
sedefsavas, line 66: Let's keep this check here too, `applyFunc()` that this method calls dereference params (the check should actually be there but that method does not return error).
>  tags := infrav1.Build(*params)

Does not matter if the current caller of this method (Ensure()) has this check. If it is called somewhere else with nil params, will panic.

Issue #3085, opened 2022-01-20
Conversion is not completely implemented for all API types


sbueringer:
/kind bug

Hey folks,
in core CAPI we have a tool called conversion-verifier which loops over all types and ensures the controller-runtime conversion interfaces are implemented correctly, i.e. it checks the types have the Hub / ConvertTo / ConvertFrom functions.

The results for CAPA are:
* type "infrastructure.cluster.x-k8s.io/AWSClusterTemplateList" in package "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1" marked as storage version but it's not convertible, missing Hub() method
* type "infrastructure.cluster.x-k8s.io/AWSClusterTemplateList" in package "sigs.k8s.io/cluster-api-provider-aws/api/v1alpha4" it's not convertible, missing ConvertFrom() and ConvertTo() methods

I think it would be great to:
* implement the missing methods
* integrate the conversion-verifier in CI

The conversion-verifier can be found here:
* Code: https://github.com/kubernetes-sigs/cluster-api/tree/main/hack/tools/conversion-verifier
* Usage: https://github.com/kubernetes-sigs/cluster-api/blob/main/Makefile#L339-L341 (core CAPI runs it as part of the verify job)

I can provide additional details if needed / if you want to implement this issue.

**Environment:**

- Cluster-api-provider-aws version: latest
- Kubernetes version: (use `kubectl version`): -
- OS (e.g. from `/etc/os-release`): -

sedefsavas:
Thanks for the detailed explanation @sbueringer 

This is not critical as ClusterClass support is in v1beta1, but we need to fix it anyway.

/triage accepted
/priority important-soon
/milestone v1.3.0

richardcase:
As discussed in the office hours i think you need to do this as well for AWSClusterTemplateList conversion:

* Create a new file `api/v1alpha4/awsclustertemplate_conversion.go` that contains something like this:

```go
func (r *AWSClusterTemplateList) ConvertFrom(srcRaw conversion.Hub) error {
	src := srcRaw.(*infrav1.AWSClusterTemplateList)

	return autoConvert_v1beta1_AWSClusterTemplateList_To_v1alpha4_AWSClusterTemplateList(src, r, nil)

	return nil
}

func (r *AWSClusterTemplateList) ConvertTo(dstRaw conversion.Hub) error {
	dst := dstRaw.(*infrav1.AWSClusterTemplateList)

[...]

Issue #3003, opened 2021-12-06
V1.1.0 is not compatible with v1.0.0 release due to setting up an optional field in the controller


sedefsavas:
In our community meetings, we have decided to only do minor releases starting from v1 and do a major release for breaking changes. So, all v1.x releases need to be compatible with each other.

In v1.1.0, we introduced a bug by adding an optional field

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3e61428d885f2eaeab1a7dffc4d17270a20c6675/api/v1beta1/awscluster_types.go#L156

and then setting it a value in the controller:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3e61428d885f2eaeab1a7dffc4d17270a20c6675/pkg/cloud/services/elb/loadbalancer.go#L72

Now, external controllers that use CAPA API v1.0.0, is failing to update AWSCluster because admission webhook denies the request with `invalid: spec.controlPlaneLoadBalancer.name: Invalid value: "null": field is immutable` error

/kind bug
/priority critical-urgent
/milestone v1.2.0
/triage accepted

sedefsavas:
@dilyevsky What would not be working if we remove the LB name assignment in BYO LB?

dlipovetsky:
The external controller doesn't know about the Name field, and the defaulting webhook is setting it to nil, but the Name field has already been set by the v1.1.0 CAPA controller, so that's causing the validating webhook to reject the update as a change? 

@sedefsavas and I talked about possible workarounds.

The first option I'll try is to generate the default ELB name in the code, and to _not_ persist it in the AWSCluster Spec. The reasoning is: Any client that is not aware of the Name field would get the default value, and if the default value is not persisted, then that client can update the AWSCluster without hitting the validation error. Any client that _is_ aware of the Name field can use it. And although the default value won't be persisted to the Spec, it _will_ still be in the Status, and so it will be discoverable.

PR #3124, opened 2022-02-01
Add support for custom protocol for ELB health checks

### LINKED ISSUE #1657: Use naive TCP instead of SSL for ELB health checks

Ankitasw:
**What type of PR is this?**
/kind bug

This PR adds support for providing the custom protocol for Classic ELB health checks so as to support different ciphers.

Fixes #1657 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
Add support for custom protocol for ELB health checks
```

### DIFF
--- a/api/v1alpha4/awscluster_conversion.go
+++ b/api/v1alpha4/awscluster_conversion.go
@@ -52,6 +52,7 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 // Assumes restored and dst are non-nil.
 func restoreControlPlaneLoadBalancer(restored, dst *infrav1.AWSLoadBalancerSpec) {
 	dst.Name = restored.Name
+	dst.HealthCheckProtocol = restored.HealthCheckProtocol
 }
 
 // ConvertFrom converts the v1beta1 AWSCluster receiver to a v1alpha4 AWSCluster.
--- a/api/v1beta1/awscluster_types.go
+++ b/api/v1beta1/awscluster_types.go
@@ -177,6 +177,11 @@ type AWSLoadBalancerSpec struct {
 	// +optional
 	Subnets []string `json:"subnets,omitempty"`
 
+	// HealthCheckProtocol sets the protocol type for classic ELB health check target
+	// default value is ClassicELBProtocolSSL
+	// +optional
+	HealthCheckProtocol *ClassicELBProtocol `json:"healthCheckProtocol,omitempty"`
+
 	// AdditionalSecurityGroups sets the security groups used by the load balancer. Expected to be security group IDs
 	// This is optional - if not provided new security groups will be created for the load balancer
 	// +optional
--- a/api/v1beta1/awscluster_webhook.go
+++ b/api/v1beta1/awscluster_webhook.go
@@ -115,6 +115,16 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 					r.Spec.ControlPlaneLoadBalancer.Name, "field is immutable"),
 			)
 		}
+
+		// Block the update for HealthCheckProtocol :
+		// - if it was not set in old spec but added in new spec
+		// - if it was set in old spec but changed in new spec
+		if !reflect.DeepEqual(newLoadBalancer.HealthCheckProtocol, existingLoadBalancer.HealthCheckProtocol) {
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "healthCheckProtocol"),
+					newLoadBalancer.HealthCheckProtocol, "field is immutable once set"),
+			)
+		}
 	}
 
 	if !reflect.DeepEqual(oldC.Spec.ControlPlaneEndpoint, clusterv1.APIEndpoint{}) &&
--- a/api/v1beta1/awscluster_webhook_test.go
+++ b/api/v1beta1/awscluster_webhook_test.go
@@ -46,7 +46,7 @@ func TestAWSCluster_ValidateCreate(t *testing.T) {
 		name    string
 		cluster *AWSCluster
 		wantErr bool
-		expect  func(t *testing.T, res *AWSLoadBalancerSpec)
+		expect  func(g *WithT, res *AWSLoadBalancerSpec)
 	}{
 		// The SSHKeyName tests were moved to sshkeyname_test.go
 		{
@@ -54,12 +54,8 @@ func TestAWSCluster_ValidateCreate(t *testing.T) {
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{},
 			},
-			expect: func(t *testing.T, res *AWSLoadBalancerSpec) {
-				t.Helper()
-
-				if res.Scheme.String() != ClassicELBSchemeInternetFacing.String() {
-					t.Error("Expected internet-facing defaulting for nil loadbalancer schemes")
-				}
+			expect: func(g *WithT, res *AWSLoadBalancerSpec) {
+				g.Expect(res.Scheme.String(), ClassicELBSchemeInternetFacing.String())
 			},
 			wantErr: false,
 		},
@@ -70,12 +66,8 @@ func TestAWSCluster_ValidateCreate(t *testing.T) {
 					ControlPlaneLoadBalancer: &AWSLoadBalancerSpec{Scheme: &ClassicELBSchemeIncorrectInternetFacing},
 				},
 			},
-			expect: func(t *testing.T, res *AWSLoadBalancerSpec) {
-				t.Helper()
-
-				if res.Scheme.String() != ClassicELBSchemeInternetFacing.String() {
-					t.Error("Expected internet-facing defaulting for supported incorrect scheme: Internet-facing")
-				}
+			expect: func(g *WithT, res *AWSLoadBalancerSpec) {
+				g.Expect(res.Scheme.String(), ClassicELBSchemeInternetFacing.String())
 			},
 			wantErr: false,
 		},
@@ -131,7 +123,7 @@ func TestAWSCluster_ValidateCreate(t *testing.T) {
 				return err == nil
 			}, 10*time.Second).Should(Equal(true))
 
-			tt.expect(t, c.Spec.ControlPlaneLoadBalancer)
+			tt.expect(g, c.Spec.ControlPlaneLoadBalancer)
 		})
 	}
 }
@@ -363,6 +355,56 @@ func TestAWSCluster_ValidateUpdate(t *testing.T) {
 						"":                         "value-2",
 						strings.Repeat("CAPI", 33): "value-3",
 						"key-4":                    strings.Repeat("CAPI", 65),
+					},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "Should fail if controlPlaneLoadBalancer healthcheckprotocol is updated",
+			oldCluster: &AWSCluster{
+				Spec: AWSClusterSpec{
+					ControlPlaneLoadBalancer: &AWSLoadBalancerSpec{
+						HealthCheckProtocol: &ClassicELBProtocolTCP,
+					},
+				},
+			},
+			newCluster: &AWSCluster{
+				Spec: AWSClusterSpec{
+					ControlPlaneLoadBalancer: &AWSLoadBalancerSpec{
+						HealthCheckProtocol: &ClassicELBProtocolSSL,
+					},
+				},
+			},
+			wantErr: true,
+		},
+		{
+			name: "Should pass if controlPlaneLoadBalancer healthcheckprotocol is same after update",
+			oldCluster: &AWSCluster{
+				Spec: AWSClusterSpec{
+					ControlPlaneLoadBalancer: &AWSLoadBalancerSpec{
+						HealthCheckProtocol: &ClassicELBProtocolTCP,
+					},
+				},
+			},
+			newCluster: &AWSCluster{
+				Spec: AWSClusterSpec{
+					ControlPlaneLoadBalancer: &AWSLoadBalancerSpec{
+						HealthCheckProtocol: &ClassicELBProtocolTCP,
+					},
+				},
+			},
+			wantErr: false,
+		},
+		{
+			name: "Should fail if controlPlaneLoadBalancer healthcheckprotocol is changed to non-default if it was not set before update",
+			oldCluster: &AWSCluster{
+				Spec: AWSClusterSpec{},
+			},
+			newCluster: &AWSCluster{
+				Spec: AWSClusterSpec{
+					ControlPlaneLoadBalancer: &AWSLoadBalancerSpec{
+						HealthCheckProtocol: &ClassicELBProtocolTCP,
 					},
 				},
 			},
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -431,7 +431,7 @@ func (s *Service) getAPIServerClassicELBSpec(elbName string) (*infrav1.ClassicEL
 			},
 		},
 		HealthCheck: &infrav1.ClassicELBHealthCheck{
-			Target:             fmt.Sprintf("%v:%d", infrav1.ClassicELBProtocolSSL, 6443),
+			Target:             fmt.Sprintf("%v:%d", s.getHealthCheckELBProtocol(), 6443),
 			Interval:           10 * time.Second,
 			Timeout:            5 * time.Second,
 			HealthyThreshold:   5,
@@ -779,6 +779,14 @@ func (s *Service) reconcileELBTags(lb *infrav1.ClassicELB, desiredTags map[strin
 	}
 
 	return nil
+}
+
+func (s *Service) getHealthCheckELBProtocol() *infrav1.ClassicELBProtocol {
+	controlPlaneELB := s.scope.ControlPlaneLoadBalancer()
+	if controlPlaneELB != nil && controlPlaneELB.HealthCheckProtocol != nil {
+		return controlPlaneELB.HealthCheckProtocol
+	}
+	return &infrav1.ClassicELBProtocolSSL
 }
 
 func fromSDKTypeToClassicELB(v *elb.LoadBalancerDescription, attrs *elb.LoadBalancerAttributes, tags []*elb.Tag) *infrav1.ClassicELB {
--- a/pkg/cloud/services/elb/loadbalancer_test.go
+++ b/pkg/cloud/services/elb/loadbalancer_test.go
@@ -28,6 +28,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/elb"
 	rgapi "github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
 	"github.com/golang/mock/gomock"
+	. "github.com/onsi/gomega"
 	"github.com/pkg/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -154,13 +155,13 @@ func TestGetAPIServerClassicELBSpec_ControlPlaneLoadBalancer(t *testing.T) {
 		name   string
 		lb     *infrav1.AWSLoadBalancerSpec
 		mocks  func(m *mock_ec2iface.MockEC2APIMockRecorder)
-		expect func(t *testing.T, res *infrav1.ClassicELB)
+		expect func(t *testing.T, g *WithT, res *infrav1.ClassicELB)
 	}{
 		{
 			name:  "nil load balancer config",
 			lb:    nil,
 			mocks: func(m *mock_ec2iface.MockEC2APIMockRecorder) {},
-			expect: func(t *testing.T, res *infrav1.ClassicELB) {
+			expect: func(t *testing.T, g *WithT, res *infrav1.ClassicELB) {
 				t.Helper()
 				if res.Attributes.CrossZoneLoadBalancing {
 					t.Error("Expected load balancer not to have cross-zone load balancing enabled")
@@ -173,7 +174,7 @@ func TestGetAPIServerClassicELBSpec_ControlPlaneLoadBalancer(t *testing.T) {
 				CrossZoneLoadBalancing: true,
 			},
 			mocks: func(m *mock_ec2iface.MockEC2APIMockRecorder) {},
-			expect: func(t *testing.T, res *infrav1.ClassicELB) {
+			expect: func(t *testing.T, g *WithT, res *infrav1.ClassicELB) {
 				t.Helper()
 				if !res.Attributes.CrossZoneLoadBalancing {
 					t.Error("Expected load balancer to have cross-zone load balancing enabled")
@@ -205,7 +206,7 @@ func TestGetAPIServerClassicELBSpec_ControlPlaneLoadBalancer(t *testing.T) {
 						},
 					}, nil)
 			},
-			expect: func(t *testing.T, res *infrav1.ClassicELB) {
+			expect: func(t *testing.T, g *WithT, res *infrav1.ClassicELB) {
 				t.Helper()
 				if len(res.SubnetIDs) != 2 {
 					t.Errorf("Expected load balancer to be configured for 2 subnets, got %v", len(res.SubnetIDs))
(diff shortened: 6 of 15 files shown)


sedefsavas:
I think we should not change the existing default behaviour, could let users to specify what protocol should be used during health checks.

I would check:

`	HealthCheck *ClassicELBHealthCheck
` field in ELB struct.

It has 	`Target             string ` field. We might use that field.

Ankitasw:
> HealthCheck *ClassicELBHealthCheck field in ELB struct.

@sedefsavas If we do this check, then shall we add new field in networkspec as well, coz `ClassicELB` is present in `NetworkStatus`, so user can't modify it.

sedefsavas:
In that case, we can just add a field to controlPlaneLoadbalancer in spec for selecting the protocol to be used in health checks and default to SSL if not provided.

api/v1beta1/awscluster_types.go
```diff
@@ -177,6 +177,11 @@ type AWSLoadBalancerSpec struct {
 	// +optional
 	Subnets []string `json:"subnets,omitempty"`
 
+	// ClassicELBProtocol sets the protocol type for classic ELB health check target
+	// +optional
+	// +nullable
+	ClassicELBProtocol *ClassicELBProtocol `json:"classicELBProtocol"`
```
sedefsavas, line 183:
```suggestion
	HealthCheckProtocol *ClassicELBProtocol `json:"healthCheckProtocol"`
```

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -780,6 +782,15 @@ func (s *Service) reconcileELBTags(lb *infrav1.ClassicELB, desiredTags map[strin
 	return nil
 }
 
+func (s *Service) getClassicELBProtocol() *infrav1.ClassicELBProtocol {
```
sedefsavas, line 785: This protocol will only be used for health checks so should be named accordingly.

api/v1beta1/awscluster_types.go
```diff
@@ -177,6 +177,11 @@ type AWSLoadBalancerSpec struct {
 	// +optional
 	Subnets []string `json:"subnets,omitempty"`
 
+	// ClassicELBProtocol sets the protocol type for classic ELB health check target
```
sedefsavas, line 180: Since it is defaulted to `SSL` when left empty, should add a comment here.

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -780,6 +782,15 @@ func (s *Service) reconcileELBTags(lb *infrav1.ClassicELB, desiredTags map[strin
...
+	elbProtocol := &infrav1.ClassicELBProtocolSSL
```
sedefsavas, line 786: We can default it to `SSL` in defaulting webhooks when it is not provided.

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -228,6 +228,20 @@ func TestGetAPIServerClassicELBSpec_ControlPlaneLoadBalancer(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "load balancer config with elb protocol specified",
```
sedefsavas, line 232:
```suggestion
			name: "load balancer config with elb health check protocol specified",
```

sedefsavas: Also needs a test to confirm it is set to SSL by default, but this test will be in the webhooks after doing the changes above.

api/v1alpha4/awscluster_conversion.go
```diff
@@ -52,6 +52,7 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 // Assumes restored and dst are non-nil.
 func restoreControlPlaneLoadBalancer(restored, dst *infrav1.AWSLoadBalancerSpec) {
 	dst.Name = restored.Name
+	dst.ClassicELBProtocol = restored.ClassicELBProtocol
```
sedefsavas, line 55: Just confirming: not adding this failed our fuzzy conversion tests, right?
There was an issue about them not catching errors.

sedefsavas (commented):
Also, adding TCP check to one of the e2e tests would be good.

api/v1alpha4/awscluster_conversion.go
Ankitasw: Yes it's creating problem in fuzzy conversion tests that's why added this is v1alpha3 and v1alpha4 conversions.

Ankitasw: is it related to this [issue](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3085)

sedefsavas: No, it was about EKS conversion tests. Just wanted to confirm for myself, not related to this issue.

pkg/cloud/services/elb/loadbalancer_test.go
Ankitasw: Done

pkg/cloud/services/elb/loadbalancer.go
Ankitasw: Done

api/v1beta1/awscluster_types.go
Ankitasw: Done

pkg/cloud/services/elb/loadbalancer.go
Ankitasw: Done

api/v1beta1/awscluster_types.go
Ankitasw: Done

Ankitasw:
@sedefsavas Manual testing is still pending, I will get that checked first, and then we can get this merged

Ankitasw:
> @sedefsavas Manual testing is still pending, I will get that checked first, and then we can get this merged

Added healthcheckprotocol as TCP, the cluster including  control plane node has come up successfully.

sedefsavas:
Forgot to mention, we should add update validations to block changing the protocol once set.

api/v1beta1/awscluster_webhook_test.go
```diff
@@ -87,6 +87,20 @@ func TestAWSCluster_ValidateCreate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "HealthCheckProtocol in ControlPlaneLoadBalancer should default to ClassicELBProtocolSSL",
```
shivi28, line 91: As we have agreed on [this](https://medium.com/@stefanovskyi/unit-test-naming-conventions-dd9208eadbea) convention, so can we use it here ?

api/v1beta1/awscluster_webhook_test.go
```diff
@@ -87,6 +87,20 @@ func TestAWSCluster_ValidateCreate(t *testing.T) {
...
+			cluster: &AWSCluster{
+				Spec: AWSClusterSpec{},
+			},
+			expect: func(t *testing.T, res *AWSLoadBalancerSpec) {
+				t.Helper()
+
```
shivi28, line 97: Can we use gomega `g := NewWithT(t)`
and also use same for comparing result with expected output

api/v1beta1/awscluster_types.go
```diff
@@ -177,6 +177,12 @@ type AWSLoadBalancerSpec struct {
 	// +optional
 	Subnets []string `json:"subnets,omitempty"`
 
+	// HealthCheckProtocol sets the protocol type for classic ELB health check target
+	// default value is ClassicELBProtocolSSL
+	// +optional
+	// +nullable
+	HealthCheckProtocol *ClassicELBProtocol `json:"healthCheckProtocol"`
```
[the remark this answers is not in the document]
sedefsavas: Also, `nullable` tag is unnecessary here.

api/v1beta1/awscluster_webhook.go
```diff
@@ -114,6 +114,12 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 					r.Spec.ControlPlaneLoadBalancer.Name, "field is immutable"),
 			)
 		}
+		if oldC.Spec.ControlPlaneLoadBalancer.HealthCheckProtocol.String() != ClassicELBProtocolSSL.String() && newLoadBalancer.HealthCheckProtocol != nil {
```
sedefsavas, line 117: If both `oldC.Spec.ControlPlaneLoadBalancer.HealthCheckProtocol` and`newLoadBalancer.HealthCheckProtocol` is TCP, then update will fail? It shouldn't.

Should add this test.

sedefsavas:
> Also, adding TCP check to one of the e2e tests would be good.

Especially, adding this to upgrade tests would show that we are not causing regression.

Ankitasw:
> Especially, adding this to upgrade tests would show that we are not causing regression.

I think with current changes upgrade tests are failing, we might want to wait for the merge then, for this release.

api/v1beta1/awscluster_webhook.go
shivi28, line 117: instead of `oldC.Spec.ControlPlaneLoadBalancer` use `existingLoadBalancer` variable. In [Line#97 ](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta1/awscluster_webhook.go#L97)we are creating deepcopy for `oldC.Spec.ControlPlaneLoadBalancer`

api/v1beta1/awscluster_webhook.go
```diff
@@ -114,6 +114,12 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
...
+			allErrs = append(allErrs,
+				field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "healthCheckProtocol"),
+					r.Spec.ControlPlaneLoadBalancer.HealthCheckProtocol, "field is immutable once set"),
```
shivi28, line 120: Use `newLoadBalancer` instead of `r.Spec.ControlPlaneLoadBalancer` --> same logic as above

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -780,6 +782,15 @@ func (s *Service) reconcileELBTags(lb *infrav1.ClassicELB, desiredTags map[strin
 	return nil
 }
 
+func (s *Service) getHealthCheckELBProtocol() *infrav1.ClassicELBProtocol {
```
shivi28, line 785:
```
func (s *Service) getHealthCheckELBProtocol() *infrav1.ClassicELBProtocol {
	controlPlaneELB := s.scope.ControlPlaneLoadBalancer()
	if controlPlaneELB != nil && controlPlaneELB.HealthCheckProtocol != nil {
		return controlPlaneELB.HealthCheckProtocol
	}
	return *infrav1.ClassicELBProtocolSSL
}
```
In this way we can reduce one new variable named elbProtocol and handle nil case if HealthCheckProtocol is nil in LBSpec. I understand we are handing nil case in defaults webhooks, but IMO it's better to not rely only on webhooks.

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -228,6 +228,20 @@ func TestGetAPIServerClassicELBSpec_ControlPlaneLoadBalancer(t *testing.T) {
...
+			mocks: func(m *mock_ec2iface.MockEC2APIMockRecorder) {},
+			expect: func(t *testing.T, res *infrav1.ClassicELB) {
+				t.Helper()
+				expectedTarget := fmt.Sprintf("%v:%d", infrav1.ClassicELBProtocolTCP, 6443)
+				if res.HealthCheck.Target != expectedTarget {
+					t.Errorf("Expected health check target to be %v, got %v", expectedTarget, res.HealthCheck.Target)
+				}
+			},
+		},
 	}
```
shivi28, line 245: Can you write a use case where
```
lb: &infrav1.AWSLoadBalancerSpec{}
```
In this case you need to call webhooks default function

pkg/cloud/services/elb/loadbalancer.go
Ankitasw: I added this before, but as per [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3124#discussion_r797313475) comment, I removed it.

pkg/cloud/services/elb/loadbalancer_test.go
Ankitasw: Can I add this check in one of the existing test case? to reduce duplication?

api/v1beta1/awscluster_types.go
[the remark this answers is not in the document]
Ankitasw: Done

api/v1beta1/awscluster_types.go
```diff
@@ -177,6 +177,12 @@ type AWSLoadBalancerSpec struct {
 	// +optional
 	Subnets []string `json:"subnets,omitempty"`
 
+	// HealthCheckProtocol sets the protocol type for classic ELB health check target
+	// default value is ClassicELBProtocolSSL
```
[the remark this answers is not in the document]
Ankitasw: Done

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -432,7 +434,7 @@ func (s *Service) getAPIServerClassicELBSpec(elbName string) (*infrav1.ClassicEL
 			},
 		},
 		HealthCheck: &infrav1.ClassicELBHealthCheck{
-			Target:             fmt.Sprintf("%v:%d", infrav1.ClassicELBProtocolSSL, 6443),
+			Target:             fmt.Sprintf("%v:%d", protocol, 6443),
```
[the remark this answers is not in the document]
Ankitasw: Done

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -228,6 +228,20 @@ func TestGetAPIServerClassicELBSpec_ControlPlaneLoadBalancer(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "load balancer config with elb health check protocol specified",
```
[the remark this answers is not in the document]
Ankitasw: Done

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -228,6 +228,20 @@ func TestGetAPIServerClassicELBSpec_ControlPlaneLoadBalancer(t *testing.T) {
...
+			lb: &infrav1.AWSLoadBalancerSpec{
+				HealthCheckProtocol: &infrav1.ClassicELBProtocolTCP,
+			},
+			mocks: func(m *mock_ec2iface.MockEC2APIMockRecorder) {},
+			expect: func(t *testing.T, res *infrav1.ClassicELB) {
+				t.Helper()
+				expectedTarget := fmt.Sprintf("%v:%d", infrav1.ClassicELBProtocolTCP, 6443)
+				if res.HealthCheck.Target != expectedTarget {
+					t.Errorf("Expected health check target to be %v, got %v", expectedTarget, res.HealthCheck.Target)
```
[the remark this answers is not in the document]
Ankitasw: Done

api/v1beta1/awscluster_webhook_test.go
Ankitasw: Done

api/v1beta1/awscluster_webhook_test.go
Ankitasw: Done

api/v1beta1/awscluster_webhook.go
Ankitasw: Done

Ankitasw: Done

pkg/cloud/services/elb/loadbalancer_test.go
Ankitasw: Done

api/v1beta1/awscluster_webhook.go
Ankitasw: Done

Ankitasw:
> Especially, adding this to upgrade tests would show that we are not causing regression.

@sedefsavas You mean to add custom protocol in ELB config for upgrade tests?

pkg/cloud/services/elb/loadbalancer.go
shivi28: Then you can write it as 
```
func (s *Service) getHealthCheckELBProtocol() *infrav1.ClassicELBProtocol {
	controlPlaneELB := s.scope.ControlPlaneLoadBalancer()
	if controlPlaneELB != nil && controlPlaneELB.HealthCheckProtocol != nil {
		return controlPlaneELB.HealthCheckProtocol
	}
	return nil
}
```
to reduce one new variable named `elbProtocol`

Ankitasw: Done

Ankitasw: Done

sedefsavas:
Defaulting optional fields might be breaking older v1beta1 clients. So, I think we should do this without defaulting to a value.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3003

sedefsavas:
> Defaulting optional fields might be breaking older v1beta1 clients. So, I think we should do this without defaulting to a value.

Note: Defaulting not just in the webhooks, also defaulting in the code and storing the value back in the API server causes this issue.

Ankitasw:
> Defaulting optional fields might be breaking older v1beta1 clients. So, I think we should do this without defaulting to a value.

The same thing with this issue might happen:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3003

@sedefsavas Then I shall revert the changes back to original without setting default value using kubebuilder. cc @shivi28

api/v1beta1/awscluster_types.go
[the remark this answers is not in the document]
Ankitasw: reverting this as per ([comment](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3124#issuecomment-1031101329)) by @sedefsavas

sedefsavas:
> @sedefsavas Then I shall revert the changes back to original without setting default value using kubebuilder. cc @shivi28

@Ankitasw We still need defaulting as now we have 2 options SSL/TCP, but should not persist the defaulting.

Ankitasw:
> @Ankitasw We still need defaulting as now we have 2 options SSL/TCP, but should not persist the defaulting.

Do we need defaulting in webhooks. or can we handle it in code itself, as we are doing [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/commit/3056568bad0a66ed73150f350db95dc8987624b6#diff-ee38bda5efe2606e7e7746a556c49a2d53d14eca60c029b8efebea657b8718f1R784)

api/v1beta1/defaults.go
```diff
@@ -67,6 +67,11 @@ func SetDefaults_AWSClusterSpec(s *AWSClusterSpec) { //nolint:golint,stylecheck
 	} else if s.ControlPlaneLoadBalancer.Scheme != nil && s.ControlPlaneLoadBalancer.Scheme.String() == ClassicELBSchemeIncorrectInternetFacing.String() {
 		s.ControlPlaneLoadBalancer.Scheme = &ClassicELBSchemeInternetFacing
 	}
+
+	// Set Default value of HealthCheckProtocol if not set
+	if s.ControlPlaneLoadBalancer != nil && s.ControlPlaneLoadBalancer.HealthCheckProtocol == nil {
```
sedefsavas, line 72: This is defaulting webhook, so stores the change back to api serves, which as per our discussions makes this a breaking change.

Ankitasw: so shall we add a check for immutability in cluster webhook?

Ankitasw: so shall we remove this check from here and keep it in loadbalancer code only?

Ankitasw: How can we do [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3124#issuecomment-1031140120) here

Ankitasw: Would defaulting [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/638278a552ff92c992e57a926f4598bc6bcccaec/pkg/cloud/services/elb/loadbalancer.go#L789) not suffice?

test/e2e/data/infrastructure-aws/kustomize_sources/limit-az/kustomization.yaml
```diff
@@ -2,3 +2,4 @@ resources:
   - ../default
 patchesStrategicMerge:
   - patches/limit-az.yaml
+  - patches/elb-health-check-protocol.yaml
```
Ankitasw, line 5: @sedefsavas Added a patch for custom protocol in existing test as we cannot reproduce it in upgrade test in remote mgmt cluster flavor.

sedefsavas:
Just to confirm that manual upgrade tests below are done:

1. from v1beta1 without this change to v1beta1 with this change (This is important as there won't be conversion webhooks for this scenario)
2. from v1alpha3 to v1beta1 with this change

Ankitasw:
> from v1beta1 without this change to v1beta1 with this change (This is important as there won't be conversion webhooks for this scenario)
from v1alpha3 to v1beta1 with this change

1st test is done, I will check the 2nd test manually tomorrow

sedefsavas:
> > from v1beta1 without this change to v1beta1 with this change (This is important as there won't be conversion webhooks for this scenario)
> > from v1alpha3 to v1beta1 with this change
> 
> 1st test is done, I will check the 2nd test manually tomorrow

Rethinking, regular upgrade tests cover this one. So, we are good.

api/v1beta1/awscluster_webhook.go
```diff
@@ -115,6 +115,12 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 					r.Spec.ControlPlaneLoadBalancer.Name, "field is immutable"),
 			)
 		}
+		if existingLoadBalancer.HealthCheckProtocol != nil && existingLoadBalancer.HealthCheckProtocol.String() != newLoadBalancer.HealthCheckProtocol.String() {
```
sedefsavas, line 118: `existingLoadBalancer = nil` means SSL is being used. Should block setting it to TCP.

For nil case, the only acceptable value is SSL.

sedefsavas: Unit tests should cover these scenarios.

Ankitasw: Done

Ankitasw:
Triggering final e2e test with all changes

api/v1beta1/awscluster_webhook.go
```diff
@@ -115,6 +115,17 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 					r.Spec.ControlPlaneLoadBalancer.Name, "field is immutable"),
 			)
 		}
+
+		// Block the update for HealthCheckProtocol :
+		// - if it was not set in old spec but added in new spec
+		// - if it was set in old spec but changed in new spec
+		if newLoadBalancer.HealthCheckProtocol != nil && existingLoadBalancer.HealthCheckProtocol == nil && newLoadBalancer.HealthCheckProtocol.String() != ClassicELBProtocolSSL.String() ||
```
shivi28, line 122: Thinking: could not we check this using `reflect.DeepEqual(newLoadBalancer.HealthCheckProtocol , existingLoadBalancer.HealthCheckProtocol)`

api/v1beta1/awscluster_webhook_test.go
```diff
@@ -368,6 +360,56 @@ func TestAWSCluster_ValidateUpdate(t *testing.T) {
 			},
 			wantErr: true,
 		},
+		{
+			name: "Should fail if controlPlaneLoadBalancer healthcheckprotocol is updated",
```
shivi28, line 364: What happens if during update we don't pass anything in `HealthCheckProtocol ` ? Might add a case mentioning this scenario is not possible

test/e2e/data/infrastructure-aws/kustomize_sources/limit-az/patches/elb-health-check-protocol.yaml
```diff
@@ -0,0 +1,8 @@
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
```
shivi28, line 2: Is there any reason to put it inside `limit-az` pkg?

api/v1beta1/awscluster_webhook_test.go
Ankitasw: Since we are not using defaukting webhooks and handling the default value in loadbalancer code itslef, have added the test for default case there itself.

test/e2e/data/infrastructure-aws/kustomize_sources/limit-az/patches/elb-health-check-protocol.yaml
Ankitasw: no reason, it was suggested by @sedefsavas that we take any of the existing test case to check for upgrade case for this optional field.

api/v1beta1/awscluster_webhook.go
Ankitasw: I think it makes sense to use `reflect.DeepEqual()`, thanks @shivi28.

sedefsavas:
I just saw a force push after the last e2e test so triggering test one last time.

Feel free to unhold when this tests pass.

PR #3157, opened 2022-02-05
Fix change condition conflict in reconcileDelete

### LINKED ISSUE #2180: Conditions error on deleting AWSManagedControlPlane
/kind bug
/area provider/eks
/help

When an EKS cluster (AWSManagedControlPlane) with managed node group (AWSManagedMachinePool) is deleted in the e2e tests the following error is seen:

```bash
E1229 22:54:05.485310       1 controller.go:257] controller-runtime/controller "msg"="Reconciler error" "error"="error deleting network for AWSManagedControlPlane eks-nodes-hxozyw/cluster-aoqabq-control-plane: error patching conditions: The condition \"ClusterSecurityGroupsReady\" was modified by a different process and this caused a merge/ChangeCondition conflict:   \u0026v1alpha3.Condition{\n  \tType:               \"ClusterSecurityGroupsReady\",\n  \tStatus:             \"False\",\n  \tSeverity:           \"Info\",\n- \tLastTransitionTime: v1.Time{Time: s\"2020-12-29 22:54:03 +0000 UTC\"},\n+ \tLastTransitionTime: v1.Time{Time: s\"2020-12-29 22:54:05 +0000 UTC\"},\n- \tReason:             \"Deleting\",\n+ \tReason:             \"Deleted\",\n  \tMessage:            \"\",\n  }\n" "controller"="awsmanagedcontrolplane" "name"="cluster-aoqabq-control-plane" "namespace"="eks-nodes-hxozyw"
```

This implies that 2 different manager processes are modifying the `ClusterSecurityGroupsReady` condition on the `AWSManagedControlPlane` instance.

I expect that this error not occur when deleting. If we are changing the `ClusterSecurityGroupsReady` condition on the `AWSManagedControlPlane` in 2 different managers then this needs to be changed. The only managed allowed to modify the `AWSManagedControlPlane` is the EKS control plane manager.

Logs from a e2e run with the error are [here](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/2175/pull-cluster-api-provider-aws-e2e-eks/1344046080525864960/artifacts/clusters/bootstrap/controllers/capa-eks-control-plane-controller-manager/capa-eks-control-plane-controller-manager-6c9995688-ntpw8/manager.log)

**Environment:**

- Cluster-api-provider-aws version: 0.6.3

richardchen331:
**What type of PR is this?**
/kind bug

https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2180 is occurring consistently when I looked at controller logs in https://prow.k8s.io/job-history/gs/kubernetes-jenkins/pr-logs/directory/pull-cluster-api-provider-aws-e2e-eks

I investigated into this a bit, and I think the root cause is that:
- In AWSMCP's `reconcileDelete`, it calls `DeleteSecurityGroups` and then `DeleteNetwork`.
- In `DeleteSecurityGroups`, it first marks `ClusterSecurityGroupsReadyCondition` to `Deleting`, patch it, and at the end, marks `ClusterSecurityGroupsReadyCondition` to `Deleted`.
- In `DeleteNetwork`, it patches again, which propagate the change to `ClusterSecurityGroupsReadyCondition` (changing it to `Deleted`) to the management cluster.
- If at the start of `reconcileDelete`, `ClusterSecurityGroupsReadyCondition` is already `Deleted` (this could happen, for example, if the controller tried `reconcileDelete` earlier, finished deleting security groups but didn't complete all the steps afterwards), then we are essentially patching `ClusterSecurityGroupsReadyCondition` from `Deleted` to `Deleting` then to `Deleted` again in one `reconcileDelete` loop. This triggers an error here (https://github.com/kubernetes-sigs/cluster-api/blob/v1.0.0/util/conditions/patch.go#L167) when the controller attempts to patch `ClusterSecurityGroupsReadyCondition` from `Deleting` to `Deleted` because `latestCondition` is `Deleting`, however `conditionPatch.Before` is `Deleted` (value retrieved from management cluster at the beginning of `reconcileDelete`) and `conditionPatch.After` is `Deleted` (the value the controller tries to patch).

This PR fixes the issue by checking if `ClusterSecurityGroupsReadyCondition` is already `Deleted`, if yes, then skip `DeleteSecurityGroups`.

The same issue happens to a number of other conditions (e.g. `RouteTablesReady`, `NatGatewaysReady`, `InternetGatewayReady`). The reason that only `ClusterSecurityGroupsReadyCondition` is observed is that the error is triggered in the first `PatchObject` call in `DeleteNetwork` (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/network.go#L103), and the controller doesn't have the change to set other conditions from `Deleted` to `Deleting` then back to `Deleted`.

If the fix looks good, I can apply the same fix to other conditions.

Fixes #2180

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix change condition conflict in reconcileDelete
```

### DIFF
--- a/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
+++ b/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
@@ -285,10 +285,8 @@ func (r *AWSManagedControlPlaneReconciler) reconcileDelete(ctx context.Context, 
 
 	if err := sgService.DeleteSecurityGroups(); err != nil {
 		log.Error(err, "error deleting general security groups for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
-		conditions.MarkFalse(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletionFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
 		return reconcile.Result{}, err
 	}
-	conditions.MarkFalse(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
 
 	if err := networkSvc.DeleteNetwork(); err != nil {
 		log.Error(err, "error deleting network for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
--- a/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
+++ b/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
@@ -283,9 +283,11 @@ func (r *AWSManagedControlPlaneReconciler) reconcileDelete(ctx context.Context, 
 		return reconcile.Result{}, err
 	}
 
-	if err := sgService.DeleteSecurityGroups(); err != nil {
-		log.Error(err, "error deleting general security groups for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
-		return reconcile.Result{}, err
+	if conditions.GetReason(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition) != clusterv1.DeletedReason {
+		if err := sgService.DeleteSecurityGroups(); err != nil {
+			log.Error(err, "error deleting general security groups for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
+			return reconcile.Result{}, err
+		}
 	}
 
 	if err := networkSvc.DeleteNetwork(); err != nil {
--- a/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
+++ b/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
@@ -283,12 +283,12 @@ func (r *AWSManagedControlPlaneReconciler) reconcileDelete(ctx context.Context, 
 		return reconcile.Result{}, err
 	}
 
-	if conditions.GetReason(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition) != clusterv1.DeletedReason {
-		if err := sgService.DeleteSecurityGroups(); err != nil {
-			log.Error(err, "error deleting general security groups for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
-			return reconcile.Result{}, err
-		}
+	if err := sgService.DeleteSecurityGroups(); err != nil {
+		log.Error(err, "error deleting general security groups for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
+		conditions.MarkFalse(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletionFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
+		return reconcile.Result{}, err
 	}
+	conditions.MarkFalse(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
 
 	if err := networkSvc.DeleteNetwork(); err != nil {
 		log.Error(err, "error deleting network for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
--- a/pkg/cloud/services/securitygroup/securitygroups.go
+++ b/pkg/cloud/services/securitygroup/securitygroups.go
@@ -267,7 +267,6 @@ func (s *Service) DeleteSecurityGroups() error {
 	if len(clusterGroups) == 0 {
 		return nil
 	}
-	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
 
 	for i := range clusterGroups {
 		sg := clusterGroups[i]
@@ -286,7 +285,7 @@ func (s *Service) DeleteSecurityGroups() error {
 	if err != nil {
 		return err
 	}
-	return err
+	return nil
 }
 
 func (s *Service) deleteSecurityGroup(sg *infrav1.SecurityGroup, typ string) error {
--- a/pkg/cloud/services/securitygroup/securitygroups.go
+++ b/pkg/cloud/services/securitygroup/securitygroups.go
@@ -268,10 +268,16 @@ func (s *Service) DeleteSecurityGroups() error {
 		return nil
 	}
 
+	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
+	if err := s.scope.PatchObject(); err != nil {
+		return err
+	}
+
 	for i := range clusterGroups {
 		sg := clusterGroups[i]
 		current := sg.IngressRules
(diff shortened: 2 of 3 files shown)


sedefsavas:
Adding logic that relies on the condition that is set by the very same controller is not preferred. 

Does it make sense to set this condition only during `DeleteSecurityGroups()` call? Because DeleteNetworks() does that based on the response coming from it.

controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
```diff
@@ -283,9 +283,11 @@ func (r *AWSManagedControlPlaneReconciler) reconcileDelete(ctx context.Context,
 		return reconcile.Result{}, err
 	}
 
-	if err := sgService.DeleteSecurityGroups(); err != nil {
-		log.Error(err, "error deleting general security groups for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
-		return reconcile.Result{}, err
+	if conditions.GetReason(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition) != clusterv1.DeletedReason {
```
richardcase, line 286: I don't think using the condition as part of reconciliation logic is the right way to go....when in the same reconciliation process.

Would something like this work:
* change `DeleteSecurityGroup` :
  * If the length of `clustergroups` is 0 then exit early 
  * Then if length of `clustergroups` is greater than set `clusterv1.DeletingReason` 
  * In any error handling blocks just return the error
* change AWSMCP reconcileDelete:
  *  In the error handler block for `sgService.DeleteSecurityGroups()` set the `DeletingFailed` reason
  * if successful set the `DeletedReason`

richardchen331: Thanks for the feedback! That sounds good. Updated logic according to your suggestion.

pydctw:
I've observed this patch condition conflict and am wondering why this is observed only in EKS. DeleteSecurityGroups is used both for regular and EKS clusters but I haven't seen it with a regular cluster. Any thoughts, @richardcase?

If we decide to go with the solution, awscluster_controller.go also needs a change as DeleteSecurityGroups is a common function.

pydctw:
@richardchen331, FYI, I opened a PR to [Add ClusterSecurityGroupsReadyCondition to managedcontrolplane's patchObject](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3234)

I still think checking the length of clusterGroups before setting the condition to DeletingReason as you implemented in the PR is needed.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -252,27 +252,27 @@ func (s *Service) ec2SecurityGroupToSecurityGroup(ec2SecurityGroup *ec2.Security
...
 	clusterGroups, err := s.describeClusterOwnedSecurityGroups()
 	if err != nil {
 		return err
 	}
 
+	// Security groups already deleted, exit early
+	if len(clusterGroups) == 0 {
+		return nil
+	}
+	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
```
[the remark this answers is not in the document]
richardchen331: You're right. Let me remove it.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -284,12 +284,9 @@ func (s *Service) DeleteSecurityGroups() error {
 	}
 
 	if err != nil {
-		conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, "DeletingFailed", clusterv1.ConditionSeverityWarning, err.Error())
 		return err
 	}
-	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletedReason, clusterv1.ConditionSeverityInfo, "")
-
-	return nil
+	return err
```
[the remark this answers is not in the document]
richardchen331: Makes sense. Let me change it to nil.

richardchen331:
Hi @pydctw , I updated the PR according to your comments, and also updated awscluster_controller.go as well. Could you take a look? Thanks!

controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
```diff
@@ -285,8 +285,10 @@ func (r *AWSManagedControlPlaneReconciler) reconcileDelete(ctx context.Context,
 
 	if err := sgService.DeleteSecurityGroups(); err != nil {
 		log.Error(err, "error deleting general security groups for AWSManagedControlPlane", "namespace", controlPlane.Namespace, "name", controlPlane.Name)
+		conditions.MarkFalse(managedScope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletionFailedReason, clusterv1.ConditionSeverityWarning, err.Error())
```
sedefsavas, line 288: Deletion is not failed in all scenarios where DeleteSecurityGroups() returns error. 
Having the condition set as deleting and erroring is more accurate.
For example, `describeClusterOwnedSecurityGroups()` could fail and this doesn't mean deletion process failed.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -252,27 +252,26 @@ func (s *Service) ec2SecurityGroupToSecurityGroup(ec2SecurityGroup *ec2.Security
 
 // DeleteSecurityGroups will delete a service's security groups.
 func (s *Service) DeleteSecurityGroups() error {
-	conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
```
sedefsavas, line 255: I'd rather keep this deleting state, if something fails between this and actual deletion, we will know exactly which section is problematic.

controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
richardchen331: I agree with you that failed call to `describeClusterOwnedSecurityGroups` and failed call to `revokeAllSecurityGroupIngressRules` are different. 

However if we mark the condition as Deleting first, we could run into the same issue that this PR is trying to address (patching the same condition multiple times in one reconciliation loop). 

One option is to mark conditions explicitly inside `DeleteSecurityGroups` (e.g. set the reason to something like DescribeFailed if `describeClusterOwnedSecurityGroups` failed and set the reason to DeleteFailed if `revokeAllSecurityGroupIngressRules` failed) and does not mark the condition in `awsmanagedcontrolplane_controller.go`. WDYT?

richardchen331:
> > Hi @pydctw , I updated the PR according to your comments, and also updated awscluster_controller.go as well. Could you take a look? Thanks!
> 
> Hi @richardchen331, I am not sure if you had a chance to look at the PR I linked, #3234, but the `error patching conditions` seems to be gone with it.
> 
> I think the scope of this PR can change to solve the problem of `ClusterSecurityGroupsReadyCondition` circling between Deleted <-> Deleting, which I have observed many times.
> 
> We can rearrange the code to set Deleting condition only when len(clusterGroups) > 0 and patch it right away.
> 
> ```
> if len(clusterGroups) == 0 {
> 	return nil
> }
> 
> conditions.MarkFalse(s.scope.InfraCluster(), infrav1.ClusterSecurityGroupsReadyCondition, clusterv1.DeletingReason, clusterv1.ConditionSeverityInfo, "")
> if err := s.scope.PatchObject(); err != nil {
> 		return err
> }
> ```
> 
> Thoughts?

For some reason i missed your PR. After looking at it I think your suggestion makes perfect sense :) Updated the PR to incorporate your feedback. Could you help take another look?

sedefsavas:
The oscillation between deleting and deleted reason is not unusual and could happen for any condition we set in `DeleteNetwork()` as we are setting `deleting` right before delete method is called.

Issue #3102, opened 2022-01-24
AWSCluster additional tags accepts empty key tags


jonathanbeber:
/kind bug

**What steps did you take and what happened:**
Create an `AWSCluster` resource with an additional tag where the key is empty (don't ask me why I did this):

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1alpha4
kind: AWSCluster
metadata:
  name: emptytag
  namespace: default
spec:
  additionalTags:
    owner: jbeber
    expiration: 1d
    "": ""
...
```

This leaves the capa controller failing to apply resources:

```log
E0124 1 awscluster_controller.go:209] controller/awscluster "msg"="failed to reconcile network" "error"="failed to create new vpc: failed to create vpc: InvalidParameter: Tag key must have a value\n\tstatus code: 400, request id: X" "cluster"="emptytag" "name"="emptytag" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AWSCluster" 
```

**What did you expect to happen:**

[AWS docs](https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html) says that:

> The tag key must be a minimum of 1 and a maximum of 128 Unicode characters in UTF-8.

So I'd expect the CRD to not even support this case. i.e.: `// +kubebuilder:validation:MinLength:=1` and `// +kubebuilder:validation:MaxLength:=128` in  https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/e07d4fdd5dd88319c370da692719d2c21960ec27/api/v1alpha4/tags.go#L29

**Anything else you would like to add:**
The cluster is also stuck to be deleted, since its API is never added.

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

jonathanbeber:
I'd like to work on that if possible!

jonathanbeber:
From what I see there's no easy way to validate a key in a `map[]` with kubebuilder, because it just becomes an object to OpenAPI and is never validated. I tried something like:

```go
// Tags defines a map of tags.
type Tags map[TagKey]TagValue

// TagKey is the key of a tag for AWS resources.
// +kubebuilder:validation:MinLength:=1
// +kubebuilder:validation:MaxLength:=128
type TagKey string

// TagValue is the value of a tag for AWS resources.
[...]

PR #3177, opened 2022-02-08
Validate additionalTags

jonathanbeber:
**What type of PR is this?**

/kind bug

As described in https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3102, tags keys currently can be empty or do not
respect other AWS rules for tagging, resulting in errors during
resources provisioning.

This commit adds validation to the tags in the webhook, for all the
resource where `additionalTags` is present, during creation and update.

Validation is done in the webhook instead of OpenAPI schema due to
limitations in the current `additionalTags` field type, as better
described in #3102.

Part of #3102

**Checklist**:

- [x] squashed commits
- [x] adds unit tests

**Release note**:

```release-note
additionalTags are now validated to comply with AWS API requirements
```

### DIFF
--- a/api/v1beta1/awscluster_webhook.go
+++ b/api/v1beta1/awscluster_webhook.go
@@ -54,6 +54,7 @@ func (r *AWSCluster) ValidateCreate() error {
 
 	allErrs = append(allErrs, r.Spec.Bastion.Validate()...)
 	allErrs = append(allErrs, r.validateSSHKeyName()...)
+	allErrs = append(allErrs, r.Spec.AdditionalTags.Validate()...)
 
 	return aggregateObjErrors(r.GroupVersionKind().GroupKind(), r.Name, allErrs)
 }
@@ -152,6 +153,7 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 	}
 
 	allErrs = append(allErrs, r.Spec.Bastion.Validate()...)
+	allErrs = append(allErrs, r.Spec.AdditionalTags.Validate()...)
 
 	return aggregateObjErrors(r.GroupVersionKind().GroupKind(), r.Name, allErrs)
 }
--- a/api/v1beta1/awscluster_webhook_test.go
+++ b/api/v1beta1/awscluster_webhook_test.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"context"
+	"strings"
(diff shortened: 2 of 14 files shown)


sedefsavas:
LGTM.

One comment: is it enough to add just a single test as tags_test.go, given that same tag validation is used by all webhooks?

richardcase:
Thanks @jonathanbeber 

We also use tags in a number of other resource kinds:

- [AWSManagedControlPlane](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go#L93)
- [AWSFargateProfile](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/api/v1beta1/awsfargateprofile_types.go#L59)
- [AWSMachinePool](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/api/v1beta1/awsmachinepool_types.go#L63)
- [AWSManagedMachinePool](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/api/v1beta1/awsmanagedmachinepool_types.go#L84)

Do you think it would be possible to add similar validation checks to the webhooks for these? This would be great as  it would give us consistent tag validation across all our resource kinds :tada:

jonathanbeber:
Thank you very much for your quick review and feedback.

> LGTM.
> 
> One comment: is it enough to add just a single test as tags_test.go, given that same tag validation is used by all webhooks?

Maybe I didn't follow completely this one, @sedefsavas. You mean like testing in just one of the webhooks?  That would leave us in a situation where a line like [this one](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/62e5730e6d32d51d4c106534de40308b431141ce/api/v1beta1/awsmachine_webhook.go#L57) is removed and the tests wouldn't break.

---

@richardcase sure thing, I'm not sure how I missed those :man_facepalming: but for sure I will work in adding them too.

jonathanbeber:
@richardcase I added the validation for other resources. I realised I ignored them because they were part of other packages, thanks for pointing it out. 

For some of the resources I had to add new test cases as the validation was not being tested.

richardcase:
Thanks for making those changes @jonathanbeber .

/approve

PR #3186, opened 2022-02-10
Use Wrapf for errors only when err is not nil

### LINKED ISSUE #3179: Errors missed out due to wrong usage of error.Wrapf()
/kind bug

While writing the tests for `pkg/cloud/services/natgateway` package, I encountered [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/natgateways.go#L290) line of code. Here the expectation was to return an error if `len(out.NatGateways)` is zero but actually it will return error as nil. This is because `error.Wrapf(err, args…)` returns nil if err is nil. 
Due to this bug, the method `WaitForWithRetryable()` will keep retrying and then `deleteNatGateway()` will timeout. 

On further investigation, I found similar instances in other places too.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/eks/fargate.go#L107
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/eks/nodegroup.go#L487

We should use wrap errors with error.Wrapf() only if error is non-nil.

DiptoChakrabarty:
**What type of PR is this?**

/kind bug
This PR checks if a error is present already and then only makes use of the errors.Wrapf functionality else it returns a custom error. This should prevent unnecessary timeouts during implementation

Fixes #3179 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/pkg/cloud/services/eks/fargate.go
+++ b/pkg/cloud/services/eks/fargate.go
@@ -104,7 +104,7 @@ func (s *FargateService) reconcileFargateProfile() (requeue bool, err error) {
 		tagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.ClusterName())
 		ownedTag := profile.Tags[tagKey]
 		if ownedTag == nil {
-			return false, errors.Wrapf(err, "owned tag not found for this cluster")
+			return false, errors.New("owned tag not found for this cluster")
 		}
 		s.scope.V(2).Info("Found owned EKS fargate profile", "cluster-name", eksClusterName, "profile-name", profileName)
 	}
--- a/pkg/cloud/services/eks/nodegroup.go
+++ b/pkg/cloud/services/eks/nodegroup.go
@@ -484,7 +484,7 @@ func (s *NodegroupService) reconcileNodegroup() error {
 		tagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.ClusterName())
 		ownedTag := ng.Tags[tagKey]
 		if ownedTag == nil {
-			return errors.Wrapf(err, "owner of %s mismatch: %s", eksNodegroupName, s.scope.ClusterName())
+			return errors.Errorf("owner of %s mismatch: %s", eksNodegroupName, s.scope.ClusterName())
 		}
 		s.scope.V(2).Info("Found owned EKS nodegroup in AWS", "cluster-name", eksClusterName, "nodegroup-name", eksNodegroupName)
 	}
--- a/pkg/cloud/services/network/natgateways.go
+++ b/pkg/cloud/services/network/natgateways.go
@@ -287,7 +287,7 @@ func (s *Service) deleteNatGateway(id string) error {
 		}
 
 		if len(out.NatGateways) == 0 {
-			return false, errors.Wrapf(err, "no NAT gateway returned for id %q", id)
+			return false, errors.New(fmt.Sprintf("no NAT gateway returned for id %q", id))
 		}
 
 		ng := out.NatGateways[0]


shivi28:
Thanks @DiptoChakrabarty for the PR 🎉 . Have added some suggestions, otherwise it's LGTM

pkg/cloud/services/network/natgateways.go
```diff
@@ -287,7 +287,11 @@ func (s *Service) deleteNatGateway(id string) error {
 		}
 
 		if len(out.NatGateways) == 0 {
-			return false, errors.Wrapf(err, "no NAT gateway returned for id %q", id)
+			if err != nil {
+				return false, errors.Wrapf(err, "no NAT gateway returned for id %q", id)
+			}
+			errmsg := fmt.Sprintf("no NAT gateway returned for id %q", id)
+			return false, errors.New(errmsg)
```
Ankitasw, line 294: nit: We can directly use `fmt.Sprintf("no NAT gateway returned for id %q", id)` as an argument to `errors.New()` without introducing new variable, wdyt?

pkg/cloud/tags/tags.go
```diff
@@ -115,7 +115,11 @@ func WithEC2(ec2client ec2iface.EC2API) BuilderOption {
 			}
 
 			_, err := ec2client.CreateTags(createTagsInput)
-			return errors.Wrapf(err, "failed to tag resource %q in cluster %q", params.ResourceID, params.ClusterName)
+			if err != nil {
+				return errors.Wrapf(err, "failed to tag resource %q in cluster %q", params.ResourceID, params.ClusterName)
+			}
+			errmsg := fmt.Sprintf("failed to tag resource %q in cluster %q", params.ResourceID, params.ClusterName)
+			return errors.Errorf(errmsg)
```
Ankitasw, line 122: nit: Same comment as above

Ankitasw:
Hi @DiptoChakrabarty , thanks for the PR.
I request you to run `make test` locally so that if your changes affect any existing test cases, you can modify them, or take a look if something went wrong with the implementation.

pkg/cloud/tags/tags.go
pydctw, line 122: We should return nil when err == nil, meaning CreateTags are successful. This is why multiple tests are failing.

pkg/cloud/services/eks/fargate.go
```diff
@@ -104,7 +104,10 @@ func (s *FargateService) reconcileFargateProfile() (requeue bool, err error) {
 		tagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.ClusterName())
 		ownedTag := profile.Tags[tagKey]
 		if ownedTag == nil {
-			return false, errors.Wrapf(err, "owned tag not found for this cluster")
+			if err != nil {
```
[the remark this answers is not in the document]
DiptoChakrabarty: yes done

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -484,7 +484,11 @@ func (s *NodegroupService) reconcileNodegroup() error {
 		tagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.ClusterName())
 		ownedTag := ng.Tags[tagKey]
 		if ownedTag == nil {
-			return errors.Wrapf(err, "owner of %s mismatch: %s", eksNodegroupName, s.scope.ClusterName())
+			if err != nil {
+				return errors.Wrapf(err, "owner of %s mismatch: %s", eksNodegroupName, s.scope.ClusterName())
+			}
+			errmsg := fmt.Sprintf("owner of %s mismatch: %s", eksNodegroupName, s.scope.ClusterName())
```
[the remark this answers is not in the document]
DiptoChakrabarty: modified

pkg/cloud/services/network/natgateways.go
DiptoChakrabarty: yes done them that way

pkg/cloud/tags/tags.go
DiptoChakrabarty: I reverted this back to the way it is in upstream since no error will return nil by errors.Wrapf and it was causing tests to fail

DiptoChakrabarty: Thanks @pydctw I reverted it back to the original way(since wrapf will return nil if err is nil) , the failing tests failed once but passed after squash

pkg/cloud/services/network/natgateways.go
```diff
@@ -287,7 +287,10 @@ func (s *Service) deleteNatGateway(id string) error {
 		}
 
 		if len(out.NatGateways) == 0 {
-			return false, errors.Wrapf(err, "no NAT gateway returned for id %q", id)
+			if err != nil {
```
shivi28, line 290: nit: remove this err check, as we are already checking it at [Line285](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3186/files#diff-23c065949155c8a08cd81009d31c08a2d86909abe59c908a374d11c9a641281eL285)
```
if err != nil {
   return false, errors.Wrapf(err, "no NAT gateway returned for id %q", id)
}
```
/lgtm

PR #3189, opened 2022-02-10
Test cases added for secondary cidr file

shivi28:
**What type of PR is this?**
/area testing

Added test case for secondarycidr.go file

Part of #2910 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Test coverage for secondarycidr.go updated to 96%
```

### DIFF
--- a/pkg/cloud/services/network/secondarycidr.go
+++ b/pkg/cloud/services/network/secondarycidr.go
@@ -17,11 +17,17 @@ limitations under the License.
 package network
 
 import (
+	"reflect"
+
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
 )
+
+func isVPCPresent(vpcs *ec2.DescribeVpcsOutput) bool {
+	return vpcs != nil && len(vpcs.Vpcs) > 0
+}
 
 func (s *Service) associateSecondaryCidr() error {
 	if s.scope.SecondaryCidrBlock() == nil {
@@ -35,8 +41,8 @@ func (s *Service) associateSecondaryCidr() error {
 		return err
 	}
 
-	if len(vpcs.Vpcs) != 1 {
-		return errors.Errorf("VPC not found")
+	if !isVPCPresent(vpcs) {
+		return errors.Errorf("failed to associateSecondaryCidr as there are no VPCs present")
 	}
 
 	existingAssociations := vpcs.Vpcs[0].CidrBlockAssociationSet
@@ -55,6 +61,7 @@ func (s *Service) associateSecondaryCidr() error {
 		return err
 	}
 
+	// once IPv6 is supported, we need to modify out.CidrBlockAssociation.AssociationId to out.Ipv6CidrBlockAssociation.AssociationId
 	record.Eventf(s.scope.InfraCluster(), "SuccessfulAssociateSecondaryCidr", "Associated secondary CIDR with VPC %q", *out.CidrBlockAssociation.AssociationId)
 
 	return nil
@@ -72,17 +79,16 @@ func (s *Service) disassociateSecondaryCidr() error {
 		return err
 	}
 
-	if len(vpcs.Vpcs) != 1 {
-		return errors.Errorf("VPC not found")
+	if !isVPCPresent(vpcs) {
+		return errors.Errorf("failed to associateSecondaryCidr as there are no VPCs present")
 	}
 
 	existingAssociations := vpcs.Vpcs[0].CidrBlockAssociationSet
 	for _, existing := range existingAssociations {
-		if existing.CidrBlock == s.scope.SecondaryCidrBlock() {
-			_, err := s.EC2Client.DisassociateVpcCidrBlock(&ec2.DisassociateVpcCidrBlockInput{
+		if reflect.DeepEqual(existing.CidrBlock, s.scope.SecondaryCidrBlock()) {
+			if _, err := s.EC2Client.DisassociateVpcCidrBlock(&ec2.DisassociateVpcCidrBlockInput{
 				AssociationId: existing.AssociationId,
-			})
-			if err != nil {
+			}); err != nil {
 				record.Warnf(s.scope.InfraCluster(), "FailedDisassociateSecondaryCidr", "Failed disassociating secondary CIDR with VPC %v", err)
 				return err
 			}
--- a/pkg/cloud/services/network/secondarycidr_test.go
+++ b/pkg/cloud/services/network/secondarycidr_test.go
@@ -1,0 +1,249 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package network
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/utils/pointer"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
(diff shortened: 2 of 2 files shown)


pkg/cloud/services/network/secondarycidr.go
```diff
@@ -35,7 +37,7 @@ func (s *Service) associateSecondaryCidr() error {
 		return err
 	}
 
-	if len(vpcs.Vpcs) != 1 {
+	if vpcs == nil || len(vpcs.Vpcs) != 1 {
 		return errors.Errorf("VPC not found")
 	}
```
Ankitasw, line 42: Since this is used in multiple places, can we extract a func out of it?

pkg/cloud/services/network/secondarycidr_test.go
```diff
@@ -0,0 +1,264 @@
...
+			wantErr: true,
+		},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			g := NewWithT(t)
+
+			scheme, err := setupScheme()
+			g.Expect(err).NotTo(HaveOccurred())
+			cl := fake.NewClientBuilder().WithScheme(scheme).Build()
```
Ankitasw, line 238: Since this client is used only in `setupNewManagedControlPlaneScope` func, can we declare it inside the func?

pkg/cloud/services/network/secondarycidr_test.go
```diff
@@ -0,0 +1,264 @@
...
```
Ankitasw, line 143: Since this client is used only in setupNewManagedControlPlaneScope func, can we declare it inside the func?

pkg/cloud/services/network/secondarycidr_test.go
```diff
@@ -0,0 +1,264 @@
...
+
+			ec2Mock := mock_ec2iface.NewMockEC2API(mockCtrl)
+
+			mcpScope, err := setupNewManagedControlPlaneScope(cl)
+			g.Expect(err).NotTo(HaveOccurred())
```
Ankitasw, line 148: Can this whole block be refactored to a separate func as used in both the tests?

pkg/cloud/services/network/secondarycidr.go
shivi28: I feel refactoring one or two lines into a separate function makes it less readable. Since, this is being used at one more place apart from this, can we keep it as it is?

pkg/cloud/services/network/secondarycidr_test.go
shivi28: This will couple the two test cases by tying their setup in a function. We have done similar thing across other test cases in CAPA as this helps each test case to evolve separately based on requirement. Wdyt?

pkg/cloud/services/network/secondarycidr_test.go
shivi28: We could do this, but the responsibility of `setupNewManagedControlPlaneScope` is to just return the managedControlPlaneScope object. And I have kept the schemeSetup and client creation out of this scope.
And from my previous experience I have observed that in test cases we might need to create some other scopes also like machineScope etc. So in those scenarios we can reuse the same client. That's why not merged it in the same func.

pkg/cloud/services/network/secondarycidr_test.go
shivi28: I have added my response in other same comment

pkg/cloud/services/network/secondarycidr.go
Ankitasw: I thought refactoring it makes more sense in readability here as we can rename the func as `isVPCPresent` which makes it clear what we are trying to do with these checks, wdyt?

pkg/cloud/services/network/secondarycidr_test.go
Ankitasw: IMO if any variable is used only for a particular func w should keep it in the function scope, so if at all it's required in future that other scopes may come into picture, we can always refactor it back.

shivi28: I got your point regarding variable scope, but responsibility of setupNewManagedControlPlaneScope() is to create managedControlPlaneScope object. Adding the creation of client inside this function seems against single responsibility principle to me. Do you still think client should be created inside the method?

pkg/cloud/services/network/secondarycidr.go
```diff
@@ -55,7 +64,9 @@ func (s *Service) associateSecondaryCidr() error {
 		return err
 	}
 
-	record.Eventf(s.scope.InfraCluster(), "SuccessfulAssociateSecondaryCidr", "Associated secondary CIDR with VPC %q", *out.CidrBlockAssociation.AssociationId)
+	if out != nil && out.CidrBlockAssociation != nil {
```
sedefsavas, line 67: Do we need this check here?

When we support IPv6, associated CIDR will show up in another field: `Ipv6CidrBlockAssociation`

I'd suggest either add a note saying this check should be modified when IPv6 is supported or check if there is any scenario where SDK returns nil output with nil error. Because we are only making the AssociateCIDR call if secondary CIDR is provided.

shivi28: Actually `out` is a pointer variable returned by SDK and to avoid `nil pointer exception` I have added it. Sure I will add a note `we should modify this check once IPv6 is supported `

sedefsavas: What I mean is if the error is nil, output probably won't return as nil ever. You can check the documentation if they mention anything. Most of the time, I cannot find this info but you can try.

And only leaving the `out`'s nil check should be enough, even if there is a scenario where error is nil and output is nil.
If `CidrBlockAssociation ` is nil, iteration on it won't happen anyways.

sedefsavas: My concern is not the nil check on out, it is CidrBlockAssociation check.

shivi28: Yes Sedef, I checked also I am not getting any scenario where out is nil in case of err ==nil . So it makes sense to remove both the checks `out != nil && out.CidrBlockAssociation != nil` . I will add a note to update the check when   IPv6 is supported

pkg/cloud/services/network/secondarycidr.go
```diff
@@ -17,12 +17,21 @@ limitations under the License.
...
 import (
+	"reflect"
+
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/record"
 )
 
+func isVPCPresent(vpcs *ec2.DescribeVpcsOutput) error {
```
richardcase, line 28: A function prefixed with `is` implies to be that it tests something and returns a boolean. 

So could it be something like this:

```go
func isVPCPresent(vpcs *ec2.DescribeVpcsOutput) bool {
  return vpcs != nil && len(vpcs.Vpcs) > 0
}
```

pkg/cloud/services/network/secondarycidr.go
```diff
@@ -72,17 +82,16 @@ func (s *Service) disassociateSecondaryCidr() error {
...
-		return errors.Errorf("VPC not found")
+	if err = isVPCPresent(vpcs); err != nil {
+		return err
 	}
 
 	existingAssociations := vpcs.Vpcs[0].CidrBlockAssociationSet
 	for _, existing := range existingAssociations {
-		if existing.CidrBlock == s.scope.SecondaryCidrBlock() {
-			_, err := s.EC2Client.DisassociateVpcCidrBlock(&ec2.DisassociateVpcCidrBlockInput{
+		if reflect.DeepEqual(existing.CidrBlock, s.scope.SecondaryCidrBlock()) {
```
richardcase, line 91: I'm wonder if we use [go-cmp](https://github.com/google/go-cmp) instead

Ankitasw: Actually we have many use cases where we are using `reflect.DeepEqual()`, shall we consider this as a separate issue, to see how feasible is go-cmp vs reflect?

richardcase: Yes, lets consider this a separate issue. There are some nice benefits to using go-cmp instead. Shall we discuss at next weeks office hours?

Ankitasw: Yeah sure @richardcase, meanwhile I will raise an issue for the same so that we can discuss  in office hours

sedefsavas: I liked the idea of providing custom equality functions.

richardcase: > I liked the idea of providing custom equality functions.

Same here @sedefsavas , thats why i like go-cmp and it using an `Equals` function if one is defined on the type.

pkg/cloud/services/network/secondarycidr.go
richardcase: @shivi28 - what do you tthink to this ^^^

PR #3202, opened 2022-02-11
Test cases added for eips.go

shivi28:
**What type of PR is this?**
/area testing

Added test cases for eips.go

Part of #2910 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Added test cases for eips.go and increase coverage to 78%
```

### DIFF
--- a/pkg/cloud/services/network/eips_test.go
+++ b/pkg/cloud/services/network/eips_test.go
@@ -1,0 +1,174 @@
+/*
+Copyright 2022 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package network
+
+import (
+	"testing"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/golang/mock/gomock"
+	. "github.com/onsi/gomega"
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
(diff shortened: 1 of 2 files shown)


Ankitasw:
The current failure in this PR is not because of the flake that we are facing, as this test ran for 3hr

pkg/cloud/services/network/eips_test.go
```diff
@@ -0,0 +1,174 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/awserrors"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope"
+	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/ec2/mock_ec2iface"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+func TestService_releaseAddresses(t *testing.T) {
```
Ankitasw, line 38: Are we not covering any case for `allocateAddresses` ?

shivi28: `allocateAddresses` code is already covered by `reconcileNatGateways` unit test, that's why not added here to avoid duplication. Added for `releaseAddresses` because this code is not tested by any code. If you feel any case is missing then kindly add it in the original issue.

pkg/cloud/services/network/eips_test.go
```diff
@@ -0,0 +1,174 @@
...
+						{
+							AssociationId: aws.String("association-id-1"),
+							PublicIp:      aws.String("public-ip"),
+							AllocationId:  aws.String("allocation-id"),
+						},
+					},
+				}, nil)
+				m.DisassociateAddress(gomock.AssignableToTypeOf(&ec2.DisassociateAddressInput{})).Return(nil, nil).Times(2)
+				m.ReleaseAddress(gomock.AssignableToTypeOf(&ec2.ReleaseAddressInput{})).Return(nil, awserr.New(awserrors.AuthFailure, awserrors.AuthFailure, errors.Errorf(awserrors.AuthFailure)))
+				m.ReleaseAddress(gomock.AssignableToTypeOf(&ec2.ReleaseAddressInput{})).Return(nil, nil)
```
Ankitasw, line 106: Why we have two different expect calls for `ReleaseAddress` here, since we have `WaitForWithRetryable` used in code and If we want to throw AuthError, wouldn't one expect call suffice?
P.S applicable to other failure cases as well.

shivi28: Actually in this test case I want to check retry happens when we receive `AuthFailure` error. So for first time I am mocking `ReleaseAddress` with AuthFailure to trigger retry. And after 1st retry I exited from retry function by returning without error from `ReleaseAddress`

PR #3212, opened 2022-02-14
Added unit test for pkg/cloud/services/secretsmanager

Madhur97:
**What type of PR is this?**
/area testing

This PR increases the unit tests coverage for pkg/cloud/services/secretmanager package.

Part of https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2910

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Add unit tests for `pkg/cloud/services/secretmanager` package
```


pkg/cloud/services/secretsmanager/secret_test.go
```diff
@@ -0,0 +1,291 @@
...
+	}
+}
+
+func setupClient() client.Client {
+	scheme := runtime.NewScheme()
+	_ = infrav1.AddToScheme(scheme)
+	return fake.NewClientBuilder().WithScheme(scheme).Build()
+}
+
+func getClusterScope(client client.Client) (*scope.ClusterScope, error) {
```
sedefsavas, line 266: nit. Duplicated code with other services.

pkg/cloud/services/secretsmanager/secret_test.go
```diff
@@ -0,0 +1,291 @@
...
+		},
+	}
+	return scope.NewClusterScope(scope.ClusterScopeParams{
+		Client:     client,
+		Cluster:    cluster,
+		AWSCluster: &infrav1.AWSCluster{},
+	})
+}
+
+func getMachineScope(client client.Client, clusterScope *scope.ClusterScope) (*scope.MachineScope, error) {
```
sedefsavas, line 279: nit. Duplicated code with other services.

pkg/cloud/services/secretsmanager/secret_test.go
```diff
@@ -0,0 +1,291 @@
...
+	tests := []struct {
+		name           string
+		bytesCount     int64
+		secretPrefix   string
+		expectedPrefix string
+		wantErr        bool
+		expect         func(g *WithT, m *mock_secretsmanageriface.MockSecretsManagerAPIMockRecorder)
+	}{
+		{
+			name:           "Should not store data in secret manager if data is having zero bytes",
```
sedefsavas, line 94: How is this scenario checked here by just checking nil error?

sedefsavas (commented):
Regarding the duplication comments: we can do a deduplication PR after all tests are merged.

pkg/cloud/services/secretsmanager/secret_test.go
Madhur97: The expect method of this test is nil, which means this test will succeed if there is no call to CreateSecret() method. If there is any call to CreateSecret(), the test will fail.
For other tests where we are expecting to store data in secret manager, we have set expected calls to CreateSecret() method.

I will add an expect method having `m.CreateSecret(gomock.Any()).Times(0)` to this test to make this more explicit.

pkg/cloud/services/secretsmanager/secret_test.go
Madhur97: As suggested in other comment, we can do this after other related PRs are also merged.

pkg/cloud/services/secretsmanager/secret_test.go
sedefsavas: No, this is enough. Thanks.

PR #3240, opened 2022-02-21
Block ELB re-creation after the cluster has been provisioned

### LINKED ISSUE #2608: Don't attempt to recreate an ELB after the cluster has been provisioned
/kind feature

If a cluster was created at time X, and then the user deletes the ELB , CAPA will try to recreate the ELB, which will have a new DNS endpoint which will not work. There's not much point doing this, and CAPA also has issues updating the resource in Kubernetes because its own webhook rejects the update as immutable.

In this scenario, we should mark AWSCluster as failed, and inform the user the cluster is unrecoverable.

Long term, we should switch to using NLBs and a recoverable endpoint.

**Environment:**

Ankitasw:
**What type of PR is this?**
/kind feature

Since ELB is created only when it's not found, this PR blocks the ELB creation in the scenario where if user deletes ELB post the cluster has been provisioned. The reason to do this is if ELB is created again after the cluster has been provisioned, then CAPA will create new ELB with new DNS endpoint which won't work.
As `AWSClusterSpec.ControlPlaneEndpoint` becomes immutable after cluster is provisioned, we are relying on the validity of this endpoint to know the status of provisioning of the cluster to check whether ELB has to be created or not. If this endpoint is valid and ELB is not found, then we inform the user that AWSCluster is in failed state and unrecoverable.

Fixes #2608 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
Block ELB re-creation after cluster is provisioned.
```

### DIFF
--- a/pkg/cloud/scope/cluster.go
+++ b/pkg/cloud/scope/cluster.go
@@ -193,6 +193,10 @@ func (s *ClusterScope) ControlPlaneLoadBalancerName() *string {
 	return nil
 }
 
+func (s *ClusterScope) ControlPlaneEndpoint() clusterv1.APIEndpoint {
+	return s.AWSCluster.Spec.ControlPlaneEndpoint
+}
+
 // ControlPlaneConfigMapName returns the name of the ConfigMap used to
 // coordinate the bootstrapping of control plane nodes.
 func (s *ClusterScope) ControlPlaneConfigMapName() string {
--- a/pkg/cloud/scope/elb.go
+++ b/pkg/cloud/scope/elb.go
@@ -19,6 +19,7 @@ package scope
 import (
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud"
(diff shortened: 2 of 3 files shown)


pkg/cloud/scope/cluster.go
```diff
@@ -193,6 +193,10 @@ func (s *ClusterScope) ControlPlaneLoadBalancerName() *string {
 	return nil
 }
 
+func (s *ClusterScope) ClusterControlPlaneEndpoint() clusterv1.APIEndpoint {
+	return s.Cluster.Spec.ControlPlaneEndpoint
```
sedefsavas, line 197: Is there a specific reason of using Cluster's ControlPlaneEndpoint here instead of AWSCluster's ControlPlaneEndpoint?

Cluster's field gets populated at a later time by Cluster controller using AWSCluster field.

Ankitasw: I thought we can add the Cluster's field as it will be the last operation to happen before setting cluster to provisioned state, but as discussed offline, will keep it as AWSCluster's endpoint as problems may arise before setting this endpoint in owner cluster and we may end up recreating ELB again.

PR #3255, opened 2022-02-25
Added subnet IDs lookup by AWSMachinePool Spec Subnets.Filters

shivi28:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Currently we are doing lookup by subnet IDs only. But with this PR we enable
```
if subnet id is there, do a lookup using that
if filter is there, do a lookup using that
If both are given then it will throw a validation error as per comment on AWSResourceReference
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3254 
Fixes #2609 
Part of #2087 

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Added subnet IDs lookup by AWSMachinePool Spec Subnets.Filters in case IDs not provided
```

PR #3257, opened 2022-02-25
Added lookup for fields of type AWSResourcesReference

### LINKED ISSUE #2087: AWSResourceReference unimplemented logic

Ankitasw:
**What type of PR is this?**
/kind bug

This PR does following:

- Remove unused conversion functions in APIs.
- Deprecate `ARN` field from `AWSResourceReference`.
- Add lookup for `AdditionalSecurityGroups` with filters.
- Refactored code for `GetAdditionalSecurityGroupIDs` so as to use common function in `AWSMachine` controller and `launchtemplate.go`.

Fixes #2087 

Subset [PR](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3255/) that handles the same for Subnet fields in AWSMachinePool

**Checklist**:
- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
- Remove unused conversion functions in APIs.
- Deprecate `ARN` field from `AWSResourceReference`.
- Add lookup for `AdditionalSecurityGroups` with filters.
- Code refactor for  `GetAdditionalSecurityGroupIDs` to use common function in `AWSMachine` controller and `launchtemplate.go`.
```

### DIFF
--- a/api/v1alpha4/conversion.go
+++ b/api/v1alpha4/conversion.go
@@ -17,22 +17,9 @@ limitations under the License.
 package v1alpha4
 
 import (
-	apiconversion "k8s.io/apimachinery/pkg/conversion"
-	conversion "k8s.io/apimachinery/pkg/conversion"
-	v1beta1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
-	clusterv1alpha4 "sigs.k8s.io/cluster-api/api/v1alpha4"
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"k8s.io/apimachinery/pkg/conversion"
+	"sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
 )
-
-// Convert_v1alpha4_APIEndpoint_To_v1beta1_APIEndpoint .
-func Convert_v1alpha4_ObjectMeta_To_v1beta1_ObjectMeta(in *clusterv1alpha4.ObjectMeta, out *clusterv1.ObjectMeta, s apiconversion.Scope) error {
-	return clusterv1alpha4.Convert_v1alpha4_ObjectMeta_To_v1beta1_ObjectMeta(in, out, s)
-}
-
-// Convert_v1beta1_APIEndpoint_To_v1alpha4_APIEndpoint .
-func Convert_v1beta1_ObjectMeta_To_v1alpha4_ObjectMeta(in *clusterv1.ObjectMeta, out *clusterv1alpha4.ObjectMeta, s apiconversion.Scope) error {
-	return clusterv1alpha4.Convert_v1beta1_ObjectMeta_To_v1alpha4_ObjectMeta(in, out, s)
-}
 
 func Convert_v1beta1_AWSClusterSpec_To_v1alpha4_AWSClusterSpec(in *v1beta1.AWSClusterSpec, out *AWSClusterSpec, s conversion.Scope) error {
 	return autoConvert_v1beta1_AWSClusterSpec_To_v1alpha4_AWSClusterSpec(in, out, s)
--- a/exp/api/v1alpha4/conversion_test.go
+++ b/exp/api/v1alpha4/conversion_test.go
@@ -20,8 +20,8 @@ import (
 	"testing"
 
 	. "github.com/onsi/gomega"
+	"k8s.io/apimachinery/pkg/runtime"
 
-	runtime "k8s.io/apimachinery/pkg/runtime"
 	"sigs.k8s.io/cluster-api-provider-aws/exp/api/v1beta1"
 	utilconversion "sigs.k8s.io/cluster-api/util/conversion"
 )
--- a/exp/api/v1beta1/awsmachinepool_webhook.go
+++ b/exp/api/v1beta1/awsmachinepool_webhook.go
@@ -89,12 +89,28 @@ func (r *AWSMachinePool) validateSubnets() field.ErrorList {
 	}
 
 	for _, subnet := range r.Spec.Subnets {
+		if subnet.ARN != nil {
+			log.Info("ARN field is deprecated and is no operation function.")
+		}
 		if subnet.ID != nil && subnet.Filters != nil {
 			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.subnets.filters"), "providing either subnet ID or filter is supported, should not provide both"))
 			break
 		}
 	}
 
+	return allErrs
+}
+
+func (r *AWSMachinePool) validateAdditionalSecurityGroups() field.ErrorList {
+	var allErrs field.ErrorList
+	for _, sg := range r.Spec.AWSLaunchTemplate.AdditionalSecurityGroups {
+		if sg.ID != nil && sg.Filters != nil {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.awsLaunchTemplate.AdditionalSecurityGroups"), "either ID or filters should be used"))
+		}
+		if sg.ARN != nil {
+			log.Info("ARN field is deprecated and is no operation function.")
+		}
+	}
 	return allErrs
 }
 
@@ -108,6 +124,7 @@ func (r *AWSMachinePool) ValidateCreate() error {
 	allErrs = append(allErrs, r.validateRootVolume()...)
 	allErrs = append(allErrs, r.Spec.AdditionalTags.Validate()...)
 	allErrs = append(allErrs, r.validateSubnets()...)
+	allErrs = append(allErrs, r.validateAdditionalSecurityGroups()...)
 
 	if len(allErrs) == 0 {
 		return nil
@@ -127,6 +144,7 @@ func (r *AWSMachinePool) ValidateUpdate(old runtime.Object) error {
 	allErrs = append(allErrs, r.validateDefaultCoolDown()...)
 	allErrs = append(allErrs, r.Spec.AdditionalTags.Validate()...)
 	allErrs = append(allErrs, r.validateSubnets()...)
+	allErrs = append(allErrs, r.validateAdditionalSecurityGroups()...)
 
 	if len(allErrs) == 0 {
 		return nil
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -990,26 +990,3 @@ func getInstanceMarketOptionsRequest(spotMarketOptions *infrav1.SpotMarketOption
 
 	return instanceMarketOptionsRequest
 }
-
-// GetFilteredSecurityGroupID get security group ID using filters.
-func (s *Service) GetFilteredSecurityGroupID(securityGroup infrav1.AWSResourceReference) (string, error) {
-	if securityGroup.Filters == nil {
-		return "", nil
-	}
-
-	filters := []*ec2.Filter{}
-	for _, f := range securityGroup.Filters {
-		filters = append(filters, &ec2.Filter{Name: aws.String(f.Name), Values: aws.StringSlice(f.Values)})
-	}
-
-	sgs, err := s.EC2Client.DescribeSecurityGroups(&ec2.DescribeSecurityGroupsInput{Filters: filters})
-	if err != nil {
-		return "", err
-	}
-
-	if len(sgs.SecurityGroups) == 0 {
-		return "", fmt.Errorf("failed to find security group matching filters: %q, reason: %w", filters, err)
-	}
-
-	return *sgs.SecurityGroups[0].GroupId, nil
-}
--- a/pkg/cloud/services/ec2/launchtemplate.go
+++ b/pkg/cloud/services/ec2/launchtemplate.go
@@ -18,6 +18,7 @@ package ec2
 
 import (
 	"encoding/base64"
+	"fmt"
 	"sort"
 	"strconv"
 	"strings"
@@ -185,9 +186,11 @@ func (s *Service) createLaunchTemplateData(scope *scope.MachinePoolScope, imageI
 	}
 
 	// add additional security groups as well
-	for _, additionalGroup := range scope.AWSMachinePool.Spec.AWSLaunchTemplate.AdditionalSecurityGroups {
-		data.SecurityGroupIds = append(data.SecurityGroupIds, additionalGroup.ID)
+	securityGroupIDs, err := s.GetAdditionalSecurityGroupsIDs(scope.AWSMachinePool.Spec.AWSLaunchTemplate.AdditionalSecurityGroups)
+	if err != nil {
+		return nil, err
 	}
+	data.SecurityGroupIds = append(data.SecurityGroupIds, aws.StringSlice(securityGroupIDs)...)
 
 	// set the AMI ID
 	data.ImageId = imageID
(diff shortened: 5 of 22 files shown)


sedefsavas:
This PR has contradicting changes with https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3255
We need to consolidate two and use a common approach as both touching the same methods.

For example, filters are used in SDK calls here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3255/files

exp/controllers/awsmachinepool_controller.go
```diff
@@ -356,20 +367,37 @@ func (r *AWSMachinePoolReconciler) updatePool(machinePoolScope *scope.MachinePoo
...
-	asg, err := asgsvc.CreateASG(machinePoolScope)
+	asg, err := asgsvc.CreateASG(machinePoolScope, subnetIDs)
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to create AWSMachinePool")
 	}
 
 	return asg, nil
 }
 
+func findSubnet(machinePoolScope *scope.MachinePoolScope, ec2svc services.EC2MachineInterface) ([]string, error) {
```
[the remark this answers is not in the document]
sedefsavas: The scope only includes the second part:

> Also, is the scope of issue also include to use filters, ARN fields wherever not used?

We don't need to add new calls to confirm subnet IDs exist in AWS.

pkg/cloud/services/ec2/instances.go
```diff
@@ -133,9 +133,27 @@ func (s *Service) CreateInstance(scope *scope.MachineScope, userData []byte) (*i
...
+		imageLookupBaseOS = scope.InfraCluster.ImageLookupBaseOS()
+	}
+
 	// Pick image from the machine configuration, or use a default one.
 	if scope.AWSMachine.Spec.AMI.ID != nil { // nolint:nestif
-		input.ImageID = *scope.AWSMachine.Spec.AMI.ID
+		input.ImageID, err = s.defaultAMIIDLookup(imageLookupFormat, imageLookupOrg, imageLookupBaseOS, *scope.Machine.Spec.Version)
+		if err != nil {
+			return nil, err
+		}
```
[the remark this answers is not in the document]
sedefsavas: If the AMI.ID is there, we should not do a lookup, just use the provided ID, right?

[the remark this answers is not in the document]
Ankitasw: yeah as explained below, misunderstood the scope. Will remove this change.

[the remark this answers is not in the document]
Ankitasw: Done

exp/controllers/awsmachinepool_controller.go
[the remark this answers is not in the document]
Ankitasw: Done

sedefsavas:
@Ankitasw 
One question: 
Have you confirmed the change is not breaking the existing v1beta1 clusters locally, as we cannot test that in e2e?

sedefsavas:
Rethinking about this, although ARN field was no-op, we still do not want to break clusters with that field specified.

**Test scenarios:**
v1alpha4 to v1beta1 upgrade test where ARN was set in a Cluster.
v1beta1 to v1beta1 upgrade test where ARN was set in a Cluster.
v1beta1 to v1beta1 upgrade test where ARN is not set in a Cluster.


@richardcase TL;DR deleting ARN field from AWSResourceReference as it is not used anywhere in the code. Since this is a  field deletion, although not used, do we still consider it as breaking change? Could deprecate then delete in the next API version change if you think that is better.

Ankitasw:
@sedefsavas I have not confirmed yet if upgrade works fine locally, I can do that and let you know.
Meanwhile @richardcase can give his inputs.

Ankitasw:
> Test scenarios:
v1alpha4 to v1beta1 upgrade test where ARN was set in a Cluster.
v1beta1 to v1beta1 upgrade test where ARN was set in a Cluster.
v1beta1 to v1beta1 upgrade test where ARN is not set in a Cluster.

Tested above scenarios successfully by setting/unsetting subnet field ARN in AWSMachine and found no issue with the removal of ARN field.

richardcase:
> @richardcase TL;DR deleting ARN field from AWSResourceReference as it is not used anywhere in the code. Since this is a field deletion, although not used, do we still consider it as breaking change? Could deprecate then delete in the next API version change if you think that is better.

Thats an interesting question. I would say we need to follow the k8s API deprecation policy because a user could be specifying the field (even if it has no effect) and then this would break for them. So i would be in favour of marking the field as deprecated and remove in the next api version. wdyt?

Ankitasw:
> Thats an interesting question. I would say we need to follow the k8s API deprecation policy because a user could be specifying the field (even if it has no effect) and then this would break for them. So i would be in favour of marking the field as deprecated and remove in the next api version. wdyt?

@richardcase If it's a no-op field, would it really break in any scenario, coz I checked the above scenarios and the cluster works all fine.

Ankitasw:
@sedefsavas @richardcase shall we move ahead with deprecating the field and nor removing? As a follow-up, I will open an issue and take care of it when we change the API.

pydctw:
+1  to follow the k8s API deprecation policy, deprecate the field first and remove in next API revision

richardcase:
> If it's a no-op field, would it really break in any scenario, coz I checked the above scenarios and the cluster works all fine(just thinking out loud). I am ok to wait for next API version as well.

@Ankitasw  - i know its a weird one. Even if it's a no-op field in reconciliation, it's a field we expose on our public API and as such someone may supply a value to it.  If they have manifests that uses it and we remove it without deprecation then `kubectl apply` will fail, and we have potentially then caused an issue somewhere with a user. So personally i think we need to follow the official k8s deprecation for this.

Ankitasw:
Thanks @richardcase, I received same feedback from community as well. Added only depreciation of field now, removal can be taken care in API version change :)

Ankitasw:
This PR is ready for another review.

exp/api/v1beta1/awsmachinepool_webhook.go
```diff
@@ -98,6 +98,16 @@ func (r *AWSMachinePool) validateSubnets() field.ErrorList {
 	return allErrs
 }
 
+func (r *AWSMachinePool) validateAdditionalSecurityGroups() field.ErrorList {
+	var allErrs field.ErrorList
+	for _, sg := range r.Spec.AWSLaunchTemplate.AdditionalSecurityGroups {
+		if sg.ID != nil && sg.Filters != nil {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.awsLaunchTemplate.AdditionalSecurityGroups"), "either ID or filters should be used"))
```
richardcase, line 105: Should we also log a warning if ARN is specified to say its deprecated?

exp/api/v1alpha4/conversion_test.go
```diff
@@ -18,14 +18,42 @@ package v1alpha4
...
+func fuzzFuncs(_ runtimeserializer.CodecFactory) []interface{} {
+	return []interface{}{
+		AWSMachinePoolFuzzer,
+	}
+}
+
+func AWSMachinePoolFuzzer(obj *AWSMachinePool, c fuzz.Continue) {
+	c.FuzzNoCustom(obj)
+	
+	// AWSMachinePool.Spec.AWSLaunchTemplate.AWSLaunchTemplate.AdditionalSecurityGroups.ARN and AWSMachinePool.Spec.Subnets.ARN have been removed in v1beta1, so setting it to nil in order to avoid v1alpha3 --> v1beta1 --> v1alpha3 round trip errors.
```
richardcase, line 42:
```suggestion
	// AWSMachinePool.Spec.AWSLaunchTemplate.AWSLaunchTemplate.AdditionalSecurityGroups.ARN and AWSMachinePool.Spec.Subnets.ARN has been deprecated in v1beta1, so setting it to nil in order to avoid v1alpha3 --> v1beta1 --> v1alpha3 round trip errors.
```

richardcase (commented):
A couple of minor nits but this looks ok to me

exp/api/v1alpha4/conversion_test.go
Ankitasw: oh i think this change has to be removed, we can add this change when we actually remove the field.

Ankitasw: There are some other conversions files as well, which will require the change only when we remove ARN field, I will revert those as well.

exp/api/v1beta1/awsmachinepool_webhook.go
Ankitasw: Done

exp/api/v1alpha4/conversion_test.go
Ankitasw: Done

Ankitasw:
@richardcase took care of comments, PTAL.

Ankitasw:
This PR is ready for review.

richardcase:
@Ankitasw - is there a way that we could add the `filter` functionality to an e2e test?

Ankitasw:
> Yes, I think that should be possible to add it in an existing test case. I will check and add it.

@richardcase Looks like there is no existing test case which is using additionalSecurityGroups, and this case should go into BYO SG test case. In addition to it, we don't have any test case using filters in templates in E2E. So I rethinking if we want to add it to E2E, or just keep it to unit tests only?

richardcase:
I think it would be good to add e2e tests around this but perhaps we create another issue to follow-up. wdyt?

Ankitasw:
> I think it would be good to add e2e tests around this but perhaps we create another issue to follow-up. wdyt?

Sounds good to me. Do you want to open new issue or shall I go ahead and do it?

Ankitasw:
@richardcase since I created a follow-up issue, is this PR good to merge?

Ankitasw:
@richardcase Also a follow-up question, do we require a new issue to keep track of removal of ARN field during API version change?
cc @sedefsavas

richardcase:
> @richardcase Also a follow-up question, do we require a new issue to keep track of removal of ARN field during API version change? cc @sedefsavas

@Ankitasw - yes thats a great idea as there are a few  fields that need removing. Shall we create a general v1beta2 API change issue?

Ankitasw:
> @Ankitasw - yes thats a great idea as there are a few fields that need removing. Shall we create a general v1beta2 API change issue?

yes that makes sense.

Ankitasw:
@richardcase is this PR good to merge?

Ankitasw:
This PR is ready to merge.

Ankitasw:
cc @sedefsavas @richardcase for approval

api/v1alpha4/conversion.go
```diff
@@ -15,19 +15,3 @@ limitations under the License.
 */
 
 package v1alpha4
-
-import (
```
sedefsavas, line 19: There is nothing left in this file, could be removed in a follow up PR.

pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -370,7 +378,15 @@ func (s *Service) LaunchTemplateNeedsUpdate(scope *scope.MachinePoolScope, incom
...
+		if ref.ID != nil {
+			incomingIDs[i] = aws.StringValue(ref.ID)
+		} else if ref.Filters != nil {
+			group, err := s.GetFilteredSecurityGroupID(ref)
+			if err != nil {
+				return false, err
+			}
+			incomingIDs[i] = group
+		}
 	}
```
sedefsavas, line 390: This logic is copy/pasted 3 times in this file, also I see the exact same logic in `awsmachine_security_groups.go`.

Is there a way to make a method out of this?

Ankitasw: I will refactor this code.

api/v1alpha4/awsmachine_conversion.go
```diff
@@ -32,7 +32,7 @@ func (src *AWSMachine) ConvertTo(dstRaw conversion.Hub) error {
 // ConvertFrom converts the v1beta1 AWSMachine to a v1alpha4 AWSMachine.
 func (dst *AWSMachine) ConvertFrom(srcRaw conversion.Hub) error {
 	src := srcRaw.(*infrav1.AWSMachine)
-
+
```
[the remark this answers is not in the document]
Ankitasw: This change is already reverted in newest commit.

sedefsavas:
Also, I remember we added filter searches to subnets before but can you confirm that we are good on that side?
All lookup should use filters for subnets as well. AWSMachinePool PR covered part of this I think:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3255/files

We should also confirm if this logic is there for AWSMachine also.

Ankitasw:
> Also, I remember we added filter searches to subnets before but can you confirm that we are good on that side?
> All lookup should use filters for subnets as well. AWSMachinePool PR covered part of this I think:
> https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3255/files
> 
> We should also confirm if this logic is there for AWSMachine.

The `AWSMachinePool` PR looked good to me, and I cross checked any other usage of `AWSResourceReference` where filter is not used, but couldn't find one.
Also, I have checked it, we are already covering subnets lookup by filters [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/ec2/instances.go#L273).

pkg/cloud/services/ec2/launchtemplate.go
Ankitasw: Done

api/v1alpha4/conversion.go
Ankitasw: Done

Ankitasw:
@sedefsavas Addressed the comments, PTAL.

Ankitasw:
E2E tests are passing now.

Ankitasw:
@richardcase rebased the PR.

Ankitasw:
Fixed the tests failing due to rebase.

Ankitasw:
@richardcase i think eks tests are failing continously for some reason(which I didn't check) https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-aws#periodic-eks-e2e-main

Ankitasw:
@sedefsavas now that we have a CAPA release, can we take this PR ahead for closure?

Ankitasw:
Reminder for PR review.

richardcase:
Seems good to me:

/lgtm

Ankitasw:
cc @sedefsavas for approval.

Ankitasw:
@sedefsavas I think this PR is ready to merge but feel free to review again.

PR #3264, opened 2022-03-01
fix: asgNeedsUpdates invalid condition (pointers)

mweibel:
**What type of PR is this?**
/kind bug

`machinePoolScope.MachinePool.Spec.Replicas` and `existingASG.DesiredCapacity` are both pointers.
The previous condition compared memory addresses instead of integer values.

I'm not sure if the check for `existingASG.DesiredCapacity == nil` is needed (i.e. can it happen at all?) but I played a bit on the safe side here. The check could still fail overall if `machinePoolScope.MachinePool.Spec.Replicas` is nil, but I guess that should never happen. Or does it?

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
fixes logic for determining if an AutoScalingGroup needs updates
```

### DIFF
--- a/exp/controllers/awsmachinepool_controller.go
+++ b/exp/controllers/awsmachinepool_controller.go
@@ -510,7 +510,11 @@ func (r *AWSMachinePoolReconciler) reconcileTags(machinePoolScope *scope.Machine
 
 // asgNeedsUpdates compares incoming AWSMachinePool and compares against existing ASG.
 func asgNeedsUpdates(machinePoolScope *scope.MachinePoolScope, existingASG *expinfrav1.AutoScalingGroup) bool {
-	if machinePoolScope.MachinePool.Spec.Replicas != nil && machinePoolScope.MachinePool.Spec.Replicas != existingASG.DesiredCapacity {
+	if machinePoolScope.MachinePool.Spec.Replicas != nil {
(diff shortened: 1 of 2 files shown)


pydctw:
> I'm not sure if the check for existingASG.DesiredCapacity == nil is needed (i.e. can it happen at all?) but I played a bit on the safe side here. The check could still fail overall if machinePoolScope.MachinePool.Spec.Replicas is nil, but I guess that should never happen. Or does it?

As existingASG.DesiredCapacity is an optional field, it can be nil. Same for MachinePool replicas spec. When a field is defined as a pointer, it is usually to distinguish between explicit zero and not specified.

Whether that is a real use case is a different story but we will need to check those conditions.

pydctw:
FYI, MachinePool spec replicas is defaulted to 1 if nil.
https://github.com/kubernetes-sigs/cluster-api/blob/main/exp/api/v1beta1/machinepool_webhook.go#L51-L53

mweibel:
restarting the timed out test (not sure if I can do that)

/test pull-cluster-api-provider-aws-verify

pydctw:
Thanks for adding missing tests in addition to what you are fixing in the PR.

/lgtm

Ankitasw:
/unassign @enxebre 
/assign @richardcase 
for approval

mweibel:
tests ran through :+1: thanks for retesting.

richardcase:
Both sets of e2e are passing so:

/approve

PR #3269, opened 2022-03-02
docs: add an initial page about e2e testing

mweibel:
**What type of PR is this?**
/kind documentation

Adds some initial docs about E2E testing which could be helpful for newcomers. 

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:
-->
```release-note
NONE
```

### DIFF
--- a/docs/book/src/crd/index.md
+++ b/docs/book/src/crd/index.md
@@ -1145,6 +1145,18 @@ string
 </tr>
 <tr>
 <td>
+<code>stackTags</code><br/>
+<em>
+map[string]string
+</em>
+</td>
+<td>
+<em>(Optional)</em>
+<p>StackTags defines the tags of the AWS CloudFormation stack.</p>
+</td>
+</tr>
+<tr>
+<td>
 <code>region</code><br/>
 <em>
 string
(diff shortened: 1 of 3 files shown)


docs/book/src/crd/index.md
```diff
@@ -1145,6 +1145,18 @@ string
 </tr>
 <tr>
 <td>
+<code>stackTags</code><br/>
```
sedefsavas, line 1148: Wondering why there are changes in the index.md.
Not seeing in my local when I build the book.

sedefsavas:
Thanks for adding this documentation @mweibel!

docs/book/src/crd/index.md
mweibel: I'm not 100% sure but it could be an issue with how Makefile targets work. Because the CRD files didn't change the index.md doesn't need updating. 
To double check I removed `docs/book/src/crd/index.md` and ran `make -C docs/book build` again. This produced the same output. 
If you try to do this, do you still have no changes or do you get the same output?

mweibel: If you prefer, I can also remove this change from this PR to ensure no unrelated change is added with this PR. LMK if I should do this.

PR #3285, opened 2022-03-07
Adds an E2E test for using peered VPCs and an internal ELB

### LINKED ISSUE #3249: Create E2E tests for managed infrastructure
/kind feature

Create additional e2e tests to validate the following use cases for the AWSCluster code base.
- Provision a Cluster with externally managed Security Groups
- Provision a Cluster with internal ELB
- Provision Management and Workload Clusters in peered VPCs

Some of the functions to support this were added in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3237. I'm finishing up adding security group functions now and have built most of the scaffolding for the test themselves.

**Environment:**

josh-ferrell:
**What type of PR is this?**
/kind feature

Creates an E2E test for creating clusters in peered VPCs and creating a workload cluster using an internal ELB.

Fixes #3249

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Adds an E2E test that validates clusters in peered VPCs and cluster creation using an internal ELB.
```

### DIFF
--- a/test/e2e/data/e2e_conf.yaml
+++ b/test/e2e/data/e2e_conf.yaml
@@ -226,6 +226,8 @@ providers:
           - sourcePath: "./infrastructure-aws/generated/cluster-template-upgrades.yaml"
           - sourcePath: "./infrastructure-aws/generated/cluster-template-external-infrastructure.yaml"
           - sourcePath: "./infrastructure-aws/generated/cluster-template-external-securitygroups.yaml"
+          - sourcePath: "./infrastructure-aws/generated/cluster-template-peered-remote.yaml"
+          - sourcePath: "./infrastructure-aws/generated/cluster-template-internal-elb.yaml"
           - sourcePath: "./infrastructure-aws/kustomize_sources/topology/clusterclass-quick-start.yaml"
           - sourcePath: "./shared/v1beta1_provider/metadata.yaml"
         replacements:
--- a/test/e2e/data/infrastructure-aws/kustomize_sources/internal-elb/kustomization.yaml
+++ b/test/e2e/data/infrastructure-aws/kustomize_sources/internal-elb/kustomization.yaml
@@ -1,0 +1,5 @@
+resources:
+  - ../remote-management-cluster
+patchesStrategicMerge:
+  - patches/internal-elb.yaml
+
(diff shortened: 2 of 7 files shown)


Ankitasw:
@josh-ferrell thank you for the PR.
Personally, I have never worked with peering VPC, so I would let @sedefsavas give feedback here.

josh-ferrell:
@sedefsavas when you get a free moment will you review this PR?

sedefsavas:
This is a great way to do both internal ELB and peering validation at the same time!

One question, instead of creating management cluster's infra, could we have used the ones auto-generated? In that case, we need to do some SDK calls to get the IDs and then do the peering with the workload cluster. Not asking to change the way it is, just curious.

Also, have you confirmed in your local that `DeleteInfrastructure()` is not leaving any resource behind? Not sure if we have the clean up logic in boskos for everything created here.

sedefsavas:
It would be great to have a documentation for deployment in internet-restricted environments, possibly with a diagram of the scenario tested here. Do you want to open an issue for this?

josh-ferrell:
> It would be great to have a documentation for deployment in internet-restricted environments, possibly with a diagram of the scenario tested here. Do you want to open an issue for this?

Technically the workload isn't Internet restricted egress wise, the API server is just on the internal ELB. I've been thinking we need to architect and document a validated airgapped deployment if we don't have one already. I can open an issue to do so.

josh-ferrell:
> This is a great way to do both internal ELB and peering validation at the same time!
> 
> One question, instead of creating management cluster's infra, could we have used the ones auto-generated? In that case, we need to do some SDK calls to get the IDs and then do the peering with the workload cluster. Not asking to change the way it is, just curious.
> 
> Also, have you confirmed in your local that `DeleteInfrastructure()` is not leaving any resource behind? Not sure if we have the clean up logic in boskos for everything created here.

I hadn't thought about using the CAPA created infrastructure to peer but I'm also still learning the SDK so it's now on my list to look into. 

DeleteInfrastructure took some time to get consistent but I think we're well covered. I could add some additional validation and probably will when refactoring the security groups test to use the same setup.

PR #3291, opened 2022-03-08
Change assertion for DetachInternetGateway to Eventually

### LINKED ISSUE #3283: Change Gomega assertion for DetachInternetGateway function in managed sec groups/VPC e2e test
/kind bug

Change the assertion for the DetachInternetGateway function to be Eventually.

Provide a timeout for achieving the IGW detachment so that the test does not fail despite validating use of managed security groups, VPCs and subnets. 

N/A

**Environment:**

josh-ferrell:
**What type of PR is this?**
/kind bug

Changes assertion for DetachInternetGateway function in the managed security group E2E test to Eventually. Provides timeout to avoid test failure due to longer running operations in AWS.

Fixes #3283 

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Provides timeout for DetachInternetGateway function in managed security group E2E test.
```

### DIFF
--- a/test/e2e/suites/unmanaged/unmanaged_functional_test.go
+++ b/test/e2e/suites/unmanaged/unmanaged_functional_test.go
@@ -647,7 +647,7 @@ var _ = ginkgo.Context("[unmanaged] [functional]", func() {
 			Expect(shared.ReleaseAddress(e2eCtx, allocationID)).To(BeTrue())
 
 			ginkgo.By("Detaching Internet gateway")
-			Expect(shared.DetachInternetGateway(e2eCtx, igwID, vpcID)).To(BeTrue())
+			Eventually(shared.DetachInternetGateway(e2eCtx, igwID, vpcID), 30*time.Second).Should(BeTrue())
 
 			ginkgo.By("Deleting Internet gateway")
 			Expect(shared.DeleteInternetGateway(e2eCtx, igwID)).To(BeTrue())


test/e2e/suites/unmanaged/unmanaged_functional_test.go
```diff
@@ -647,7 +647,7 @@ var _ = ginkgo.Context("[unmanaged] [functional]", func() {
 			Expect(shared.ReleaseAddress(e2eCtx, allocationID)).To(BeTrue())
 
 			ginkgo.By("Detaching Internet gateway")
-			Expect(shared.DetachInternetGateway(e2eCtx, igwID, vpcID)).To(BeTrue())
+			Eventually(shared.DetachInternetGateway(e2eCtx, igwID, vpcID), 60*time.Second).Should(BeTrue())
```
Ankitasw, line 650: Do we need 60s wait time? Does the detaching takes that long to finish? Or can this be reduces to 20-30s maybe. Wdyt?

josh-ferrell: @Ankitasw I honestly don't know. I've only seen it fail once and I don't have the time required on that test. I can change it to 30 seconds. My understanding was that Gomega would pass if the condition was meet before the timeout.

Ankitasw: yes, but it wont pass if it doesn't meet till 60s, so in worst case this test case will add additional 60s delay, so I think it would be better to keep it within 30s.

PR #3298, opened 2022-03-10
Removed hardcoding for AMIs in bastion host and added latest AMI lookup

### LINKED ISSUE #3108: Use AMI lookup for bastion hosts instead of using hardcoded values
We are using Ubuntu images published on AWS community AMIs for bastion host if an image id is not provided.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/50af6ec5ce7a183b67d23a187e687c42fef24176/pkg/cloud/services/ec2/ami.go#L191-L195

Currently, default AMIs used for bastion are hardcoded. To use the latest published images to avoid any security issues, we can do a lookup and automatically use the latest images published by Ubuntu.

Ankitasw:
**What type of PR is this?**
/kind feature

Currently, default AMIs used for bastion are hardcoded. This PR enables to lookup the latest images published by Ubuntu.

Fixes #3108 

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
- Added latest AMI lookup for bastion host and removed hardcoded AMIs.
- Added unit tests for bastion.go
```

### DIFF
--- a/docs/book/src/topics/accessing-ec2-instances.md
+++ b/docs/book/src/topics/accessing-ec2-instances.md
@@ -31,7 +31,7 @@ spec:
   bastion:
     enabled: true
 ```
-If this field is set, then by default the latest AMI(Ubuntu 20.04 LTS OS) is looked up from [Ubuntu cloud images](https://ubuntu.com/server/docs/cloud-images/amazon-ec2) by CAPA controller and used in bastion host creation.
+If this field is set and a specific AMI ID is not provided for the bastion (by setting spec.bastion.ami) then by default the latest AMI(Ubuntu 20.04 LTS OS) is looked up from [Ubuntu cloud images](https://ubuntu.com/server/docs/cloud-images/amazon-ec2) by CAPA controller and used in bastion host creation.
 
 #### Obtain public IP address of the bastion node
 
--- a/docs/book/src/topics/accessing-ec2-instances.md
+++ b/docs/book/src/topics/accessing-ec2-instances.md
@@ -31,6 +31,7 @@ spec:
   bastion:
     enabled: true
 ```
+If this field is set, then by default the latest AMI(Ubuntu 20.04 LTS OS) is looked up from [Ubuntu cloud images](https://ubuntu.com/server/docs/cloud-images/amazon-ec2) by CAPA controller and used in bastion host creation.
 
 #### Obtain public IP address of the bastion node
 
--- a/pkg/cloud/services/ec2/ami.go
+++ b/pkg/cloud/services/ec2/ami.go
@@ -40,6 +40,13 @@ const (
 	// https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/487
 	DefaultMachineAMIOwnerID = "258751437250"
 
+	// ubuntuOwnerID is Ubuntu owned account. Please see:
+	// https://ubuntu.com/server/docs/cloud-images/amazon-ec2
+	ubuntuOwnerID = "099720109477"
+
+	// Description regex for fetching Ubuntu AMIs for bastion host.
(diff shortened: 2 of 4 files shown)


Ankitasw:
I will add the documentation.
I have not tested it locally but i checked e2e tests looks fine, would that not suffice? If not, i will check it locally.

sedefsavas:
> I will add the documentation. I have not tested it locally but i checked e2e tests looks fine, would that not suffice? If not, i will check it locally.

If we have a bastion host enabled template probably enough, but probably need to look at the controller logs if creating bastion instance is successful or not because we don't have any checks in the tests. Would be good to make sure locally as well.

Ankitasw:
> If we have a bastion host enabled template probably enough, but probably need to look at the controller logs if creating bastion instance is successful or not because we don't have any checks in the tests. Would be good to make sure locally as well.

I checked that in any of E2E templates we are not setting bastion as enabled, shall we add it to one of the existing tests so that this is tested?
Also, I have tested locally, and checked the logs, it's creating bastion host successfully.

> I0311 08:37:10.465774       8 bastion.go:91] controller/awscluster "msg"="Created new bastion host" "cluster"="capi-bastion-test" "name"="capi-bastion-test" "namespace"="default" "reconciler group"="infrastructure.cluster.x-k8s.io" "reconciler kind"="AWSCluster" "id"="i-02b3c36d4135c037d"

Ankitasw:
> Would be good to have a section that explains which images we use by default for bastions under accessing ec2 instances section in the book.

@sedefsavas The default latest AMIs that are being used as of now shall be listed there? If yes, we will need to update it everytime the latest AMIs are published right?

docs/book/src/topics/accessing-ec2-instances.md
```diff
@@ -31,6 +31,7 @@ spec:
   bastion:
     enabled: true
 ```
+If this field is set, then by default the latest AMI is looked up from [Ubuntu cloud images](https://ubuntu.com/server/docs/cloud-images/amazon-ec2) by CAPA controller and used in bastion host creation.
```
Ankitasw, line 34: @sedefsavas Let me know if this statement would suffice?

docs/book/src/topics/accessing-ec2-instances.md
```diff
@@ -31,6 +31,7 @@ spec:
   bastion:
     enabled: true
 ```
+If this field is set, then by default the latest AMI(Ubuntu 20.04 LTS OS) is looked up from [Ubuntu cloud images](https://ubuntu.com/server/docs/cloud-images/amazon-ec2) by CAPA controller and used in bastion host creation.
```
Ankitasw, line 34: @sedefsavas Let me know if this statement would suffice?

sedefsavas:
No need to publish those images since not providing useful info, users can go to that link and see for themselves, and if users want more control, they can specify the AMI ID for the bastion host.

Also, enabling bastion in e2e might be difficult to test, what will be the success condition there? Merging this PR now as this looks complete as is, if you want to do add an e2e test, feel free to do a follow up PR.

docs/book/src/topics/accessing-ec2-instances.md
sedefsavas: Just added `this is true in case AMI ID is not specified`.

sedefsavas:
/lgtm
/approve

Can check `BastionHostReadyCondition` in the e2e test.
Also, might add a new reason for this condition: BastionAMILookupFailedReason.

Ankitasw:
@sedefsavas since we have integration tests in controllers niw, what would make more sense in this case, E2E or integration test? Wdyt?

PR #3303, opened 2022-03-11
Clean up readiness condition on legit deletion

enxebre:
**What type of PR is this?**
/kind bug
/kind api-change

During a successful deletion operation conditions that summarise readiness must be cleaned up before removing the finalizer.
Otherwise, scalable resources which aggregates this resource will see and permanently store ready=false for this resource right before it goes away.
This would mistakenly impact in the readiness of the owner scalable resource which should disregard the particular ready condition of a resource legitimately deleted.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -382,6 +382,15 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeNormal, "SuccessfulTerminate", "Terminated instance %q", instance.ID)
 	}
 
+	// During a successful deletion operation conditions that summarise readiness must be cleaned up before removing the finalizer.
+	// Otherwise, scalable resources which aggregates this resource will see and permanently store ready=false for this resource right before it goes away.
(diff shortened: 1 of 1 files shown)


controllers/awsmachine_controller.go
```diff
@@ -382,6 +382,15 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeNormal, "SuccessfulTerminate", "Terminated instance %q", instance.ID)
 	}
 
+	// During a successful deletion operation conditions that summarise readiness must be cleaned up before removing the finalizer.
+	// Otherwise, scalable resources which aggregates this resource will see and permanently store ready=false for this resource right before it goes away.
+	// This would mistakenly impact in the readiness of the owner scalable resource which should disregard the particular ready condition of a resource legitimately deleted.
+	if conditions.Has(machineScope.AWSMachine, infrav1.SecurityGroupsReadyCondition) || conditions.Has(machineScope.AWSMachine, infrav1.InstanceReadyCondition) {
```
richardcase, line 388: Could we remove the the `if` and just do the the `Delete`  & `Patch`

richardcase: @enxebre - checking in on this to see what you think?

PR #3308, opened 2022-03-15
Add sanity check for VPC ID in VPC deletion reconcile loop

### LINKED ISSUE #2758: Cannot delete AWSManagedControlPlane when using existing VPC
/kind bug

Maybe interesting. I played  with machinepools. I'm in an environment you need to use an existing vpc. I forgot to set the vpc in machinepool. So the AWSManagedControlPlane could not be deployed. FunFact I was not able to delete the resource either:

`failed to delete vpc "": UnauthorizedOperation: You are not authorized to perform this operation.`

So deleting AWSManagedControlPlane failed because it was not able (missing rights anyway) to delete a non-existing vpc (as it was never be created (missing rights)). So it hangs.
I'm not sure if that;s AWS-specific or something that can be handled by the provider. 

Simplified I created a Cluster without having the rights to create an VPC.

As it was my mistake. I would have expected to delete the resource AWSManagedControlPlane.

**Environment:**

- Cluster-api-provider-aws version:  0.7.0

- Kubernetes version: (use `kubectl version`): Client Version: version.Info{Major:"1", Minor:"21", GitVersion:"v1.21.0", GitCommit:"cb303e613a121a29364f75cc67d3d580833a7479", GitTreeState:"clean", BuildDate:"2021-04-08T21:15:16Z", GoVersion:"go1.16.3", Compiler:"gc", Platform:"darwin/amd64"}
- 
Laptop or Host?

Ankitasw:
**What type of PR is this?**
/kind bug

In case of unmanaged VPC, if user forgets to set the VPC ID in machinepool, AWSManagedControlPlane wont be deployed. If user wants to delete the resources to recreate the cluster, VPC reconciliation fails as it doesn't find any VPC ID to delete. This PR adds a nil check to validate if we really have VPC ID in place to proceed with the deletion.

Fixes #2758 

**Checklist**:
- [x] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:
```release-note
Add sanity check for VPC ID in VPC deletion.
```

### DIFF
--- a/pkg/cloud/services/network/vpc.go
+++ b/pkg/cloud/services/network/vpc.go
@@ -202,6 +202,12 @@ func (s *Service) createVPC() (*infrav1.VPCSpec, error) {
 func (s *Service) deleteVPC() error {
 	vpc := s.scope.VPC()
 
+	// Skip deletion if .spec.vpc.id is not set.
+	if vpc.ID == "" {
+		s.scope.V(4).Info("Skipping VPC deletion as VPC ID is not set")
+		return nil
+	}
+
 	if vpc.IsUnmanaged(s.scope.Name()) {
 		s.scope.V(4).Info("Skipping VPC deletion in unmanaged mode")
 		return nil
--- a/pkg/cloud/services/network/vpc_test.go
+++ b/pkg/cloud/services/network/vpc_test.go
@@ -402,6 +402,13 @@ func Test_DeleteVPC(t *testing.T) {
 				})).Return(nil, awserr.New("InvalidVpcID.NotFound", "not found", nil))
 			},
 		},
+		{
+			name:    "Should skip reconciliation if VPC ID is not mentioned",
+			input:   &infrav1.VPCSpec{ID: ""},
+			wantErr: false,
+			expect: func(m *mock_ec2iface.MockEC2APIMockRecorder) {
+			},
+		},
 	}
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {


pkg/cloud/services/network/vpc.go
```diff
@@ -202,6 +202,12 @@ func (s *Service) createVPC() (*infrav1.VPCSpec, error) {
 func (s *Service) deleteVPC() error {
 	vpc := s.scope.VPC()
 
+	// Skip deletion if VPC is deleted already or not mentioned in .spec.vpc.id.
```
sedefsavas, line 205: When VPC is deleted, this field is not being set to nil.

Ankitasw: Done

Ankitasw:
This PR is ready for review.

pkg/cloud/services/network/vpc.go
```diff
@@ -202,6 +202,12 @@ func (s *Service) createVPC() (*infrav1.VPCSpec, error) {
 func (s *Service) deleteVPC() error {
 	vpc := s.scope.VPC()
 
+	// Skip deletion if .spec.vpc.id is not set.
+	if vpc.ID == "" {
```
sedefsavas, line 206: Below we return if VPC does not exist:

> 		if code, ok := awserrors.Code(err); ok && code == awserrors.VPCNotFound {

When the ID is nil, don't we get VPCNotFound error?

Ankitasw: Since we are getting below error in this scenario:

> failed to delete vpc "": UnauthorizedOperation: You are not authorized to perform this operation.

which is arising from [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/scope/clients.go#L75), it doesn't throw `VPCNotFound` error.

Madhur97:
@Ankitasw Were you able to reproduce this issue?
There is already a check present for [vpc ID being missing](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/network.go#L78) before calling [deleteVPC()](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/vpc.go#L202) method.

pkg/cloud/services/network/vpc.go
sedefsavas: In this issue, it is stated that this error is seen due to auth issues.

Are you able to replicate this locally and saw the same error?

Ankitasw: I have not replicated the issue locally yet, will do and provide the findings.

Ankitasw: makes sense to me, I will close the PR and issue for now and maybe we can reopen if we face this issue again.

PR #3322, opened 2022-03-16
Remove .metadata.clusterName usage

### LINKED ISSUE #3321: CAPA should not use ObjectMeta.ClusterName
There is an upstream discussion about dropping ObjectMeta.ClusterName: https://github.com/kubernetes/kubernetes/pull/108717. Related CAPI issue: https://github.com/kubernetes-sigs/cluster-api/issues/6305

This field is not populated by Kubernetes and we should not use it as an CAPA Cluster Name.

This issue is about doing some research where we use it in CAPA and fix it if it is used.

/kind bug

pydctw:
**What type of PR is this?**
/kind bug

k8s upstream plans to drop ObjectMeta.ClusterName field as it has always been force cleared by the API server.
See kubernetes/kubernetes#108717.

Searched CAPA codebase and there is only one occasion where `ObjectMeta.ClusterName` is used.

Fixes #3321 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Remove .metadata.clusterName usage
```

### DIFF
--- a/pkg/cloud/services/eks/addons.go
+++ b/pkg/cloud/services/eks/addons.go
@@ -196,7 +196,7 @@ func (s *Service) translateAPIToAddon(addons []ekscontrolplanev1.Addon) []*eksad
 		convertedAddon := &eksaddons.EKSAddon{
 			Name:                  &addon.Name,
 			Version:               &addon.Version,
-			Tags:                  ngTags(s.scope.Cluster.ClusterName, s.scope.AdditionalTags()),
+			Tags:                  ngTags(s.scope.Cluster.Name, s.scope.AdditionalTags()),
 			ResolveConflict:       convertConflictResolution(*addon.ConflictResolution),
 			ServiceAccountRoleARN: addon.ServiceAccountRoleArn,
 		}


pkg/cloud/services/eks/addons.go
```diff
@@ -196,7 +196,7 @@ func (s *Service) translateAPIToAddon(addons []ekscontrolplanev1.Addon) []*eksad
 		convertedAddon := &eksaddons.EKSAddon{
 			Name:                  &addon.Name,
 			Version:               &addon.Version,
-			Tags:                  ngTags(s.scope.Cluster.ClusterName, s.scope.AdditionalTags()),
+			Tags:                  ngTags(s.scope.Cluster.Name, s.scope.AdditionalTags()),
```
sedefsavas, line 199: I feel like this name should be the Kubernetes cluster name instead of CAPI cluster name, since they are different in EKS.
Not sure though.

pydctw: You mean EKS cluster name? I don't know why the original code is using CAPI cluster name but that is a great point. 
Let's ask @richardcase.

pydctw:
Bad CI day
/retest

pkg/cloud/services/eks/addons.go
richardcase: I'd be tempted to use `s.scope.Cluster.Name` and not the eks cluster name. An addon is always a child of a EKS cluster (and so you have the name).

Issue #3329, opened 2022-03-20
Cannot create Service type LoadBalancer in EKS clusters


dkoshkin:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]
1. Create a cluster with https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/download/v1.2.0/cluster-template-eks.yaml with name `my-cluster`
2. Create a `Deployment` and a `Service` with type `LoadBalancer`
```
apiVersion: v1
kind: Namespace
metadata:
  name: e2e-loadbalancer-created
---
apiVersion: apps/v1
kind: Deployment
metadata:
  name: example-deployment
  namespace: e2e-loadbalancer-created
  labels:
    app: nginx
spec:
  replicas: 3
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: docker.io/library/nginx:stable
        ports:
        - containerPort: 80
---
apiVersion: v1
kind: Service
metadata:
  name: example-service
  namespace: e2e-loadbalancer-created
spec:
  selector:
    app: nginx
  ports:
    - port: 8765
      targetPort: 80
  type: LoadBalancer
```
---

Saw an error in the `Service`
```
Error syncing load balancer: failed to ensure load balancer: could not find any suitable subnets for creating the ELB
```

And when enabled controller-manager logs:
```
No tagged subnets found; will fall-back to the current subnet only.
```

I saw these tags on the Subnets which at first looked correct as it had both `kubernetes.io/role/elb` and `kubernetes.io/cluster/<clusterName>`

```
kubernetes.io/role/elb	1
kubernetes.io/cluster/my-cluster shared
sigs.k8s.io/cluster-api-provider-aws/role	public
sigs.k8s.io/cluster-api-provider-aws/cluster/my-cluster	owned
Name my-cluster-subnet-public-us-west-2b
```

**What did you expect to happen:**
An ELB for the `Service` should be created and ` curl <ELB>:8765` should return an nginx  page.

---

The `No tagged subnets found` was interesting so I looked more into the Subnet tags.
Turns out the tag `kubernetes.io/cluster/my-cluster` is not correct.
The EKS clusters generated by CAPA are named like `<namespace>_<cluster_name>-control-plane` (in AWS console)

[...]

PR #3343, opened 2022-03-23
:bug: fixes services for EKS clusters 

faiq:
**What type of PR is this?**
/kind bug

This PR creates  a failing test for #3329  as well as extending the existing test for non managed clusters to ensure the LB for service is actually backed by instances.

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
changes AWSCloudProvider tag for ec2 instances to be KubernetesClusterName() instead of ClusterName() to support load balancer services for EKS clusters 
```

### DIFF
--- a/test/e2e/shared/services.go
+++ b/test/e2e/shared/services.go
@@ -1,0 +1,195 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2021 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package shared
+
+import (
+	"context"
+	"strings"
+	"time"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/awserr"
+	"github.com/aws/aws-sdk-go/service/elb"
+	. "github.com/onsi/gomega"
+	v1 "k8s.io/api/apps/v1"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	apimachinerytypes "k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+func CreateLBService(e2eCtx *E2EContext, svcNamespace string, svcName string, k8sclient crclient.Client) string {
+	Byf("Creating service of type Load Balancer with name: %s under namespace: %s", svcName, svcNamespace)
+	svcSpec := corev1.ServiceSpec{
+		Type: corev1.ServiceTypeLoadBalancer,
+		Ports: []corev1.ServicePort{
+			{
+				Port:     80,
+				Protocol: corev1.ProtocolTCP,
+			},
+		},
+		Selector: map[string]string{
+			"app": "nginx",
+		},
+	}
+	CreateService(svcName, svcNamespace, nil, svcSpec, k8sclient)
+	elbName := ""
+	Eventually(func() bool {
+		svcCreated := &corev1.Service{}
+		err := k8sclient.Get(context.TODO(), apimachinerytypes.NamespacedName{Namespace: svcNamespace, Name: svcName}, svcCreated)
+		Expect(err).NotTo(HaveOccurred())
+		if lbs := len(svcCreated.Status.LoadBalancer.Ingress); lbs > 0 {
+			ingressHostname := svcCreated.Status.LoadBalancer.Ingress[0].Hostname
+			elbName = strings.Split(ingressHostname, "-")[0]
+			return true
+		}
+		return false
+	}, e2eCtx.E2EConfig.GetIntervals("", "wait-create-service")...).Should(BeTrue())
+	Byf("Created Load Balancer service and ELB name is: %s", elbName)
+	return elbName
+}
+
+func DeleteLBService(svcNamespace string, svcName string, k8sclient crclient.Client) {
+	svcSpec := corev1.ServiceSpec{
+		Type: corev1.ServiceTypeLoadBalancer,
+		Ports: []corev1.ServicePort{
+			{
+				Port:     80,
+				Protocol: corev1.ProtocolTCP,
+			},
+		},
+		Selector: map[string]string{
+			"app": "nginx",
+		},
+	}
+	deleteService(svcName, svcNamespace, nil, svcSpec, k8sclient)
+}
+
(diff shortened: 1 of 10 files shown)


dlipovetsky:
(Setting same priority as #3329)
 
/priority important-soon

pydctw:
For EKS lb creation and verification

pydctw:
pull-cluster-api-provider-aws-e2e test failure seems to be a flake as the PR didn't touch unmanaged cluster upgrade path.
```
capa-e2e: [unmanaged] [Cluster API Framework] Clusterctl Upgrade Spec [from latest v1beta1 release to main] Should create a management cluster and then upgrade all the providers expand_less | 21m4s
-- | --
{ Failure /home/prow/go/pkg/mod/sigs.k8s.io/cluster-api/test@v1.1.2/e2e/clusterctl_upgrade.go:147 Timed out after 300.003s.
```

test/e2e/suites/managed/eks_test.go
```diff
@@ -136,7 +137,13 @@ var _ = ginkgo.Describe("[managed] [general] EKS cluster tests", func() {
 			Name:      clusterName,
 		})
 		Expect(cluster).NotTo(BeNil(), "couldn't find CAPI cluster")
-
+		clusterClient := e2eCtx.Environment.BootstrapClusterProxy.GetWorkloadCluster(ctx, namespace.Name, clusterName).GetClient()
+		ginkgo.By("Creating the LB service")
```
pydctw, line 141: It may not be necessary for the test but it just feels strange to me to create a service without the target deployment or a pod when the service is targeting pods with `"app": "nginx"` selector.

pydctw:
@faiq, thanks for the PR. Overall, looks good with a few comments.

I am not sure why eks e2e test is not triggering but to verify the PR works, eks e2e test needs to pass. 

The command is `/test pull-cluster-api-provider-aws-e2e-eks`

test/e2e/suites/managed/eks_test.go
faiq: Let me see if the `unamanaged functional` tests are doing this as well. If not I'll add them in both places.

faiq:
Hi @pydctw I believe I addressed the changes you were looking for: 
- Eventually instead of Sleep 
- Backing the service with an actual deployment

In good news, the test failed as expected for EKS proving the issue reported in #3329 as seen [in this run](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/3343/pull-cluster-api-provider-aws-e2e-eks/1507050709399375872) -- note that this was before the changes you requested. 

Thanks for the review. If this all looks good, I'll start working on the fix next.

pydctw:
The change looks great. Thanks for adding nginx deployment, the flow makes more sense now. 

I was happy to see the failing test 😄 that proved instance is currently not attached to the LB. Please go ahead with the fix.

pydctw:
Just to add - having worked in a TDD shop for a few years, I really liked seeing a failing test before fixing it.

test/e2e/shared/services.go
```diff
@@ -0,0 +1,183 @@
...
+		svcCreated := &corev1.Service{}
+		err := k8sclient.Get(context.TODO(), apimachinerytypes.NamespacedName{Namespace: svcNamespace, Name: svcName}, svcCreated)
+		Expect(err).NotTo(HaveOccurred())
+		if lbs := len(svcCreated.Status.LoadBalancer.Ingress); lbs > 0 {
+			ingressHostname := svcCreated.Status.LoadBalancer.Ingress[0].Hostname
+			elbName = strings.Split(ingressHostname, "-")[0]
+			return true
+		}
+		return false
+	}, e2eCtx.E2EConfig.GetIntervals("", "wait-create-service")...).Should(BeTrue())
```
pydctw, line 64: eks e2e test failed because `wait-create-service` is not defined. I think unmanaged e2e test will also fail.

There are two options, whichever you prefer as I don't have a strong opinion about it.
* Use a specific interval like 60*time.Second
* Define the variable in e2e conf files. As the function is used for both e2e tests, it needs to be defined in [e2e_eks_conf.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/data/e2e_eks_conf.yaml#L224-L238) and [e2e_eks_conf.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/data/e2e_conf.yaml#L286-L302)

faiq: Ah perfect... I was wondering where those intervals were coming from. Thanks so much for solving that mystery for me. 

I'll go with the specific duration option. Just a second.

faiq:
@pydctw can i backport this to release 1.3 branch now that i also included a fix?

pydctw:
> @pydctw can i backport this to release 1.3 branch now that i also included a fix?

I will ask the maintainer, cc @richardcase @sedefsavas

dlipovetsky:
Since this a bug fix, +1 to backport onto `release-1.3`

sedefsavas:
Since a shared method (VerifyElbExists) is changed:

sedefsavas:
> Since this a bug fix, +1 to backport onto release-1.3

We will make v1.4 release soon, as its been 2 months since the last release and we have some feat work in.
@dlipovetsky is there any urgency for this bug to make a patch release to v1.2?

richardcase:
This looks good to me.

@faiq - could you squash your commits?

sedefsavas:
The additional deployment added here does not give any signal as is.

`Expect(len(elbsOutput.LoadBalancerDescriptions[0].Instances)).Should(BeNumerically(">=", 1))`

This is true for even if we don't have a deployment that matches the Service. The instances attached to the LB are just Out of Service if there are no pods in the instance.

You can either add a check that shows the instances with the pods are in service in the LB,
 or can remove the deployment all together.

faiq:
> The additional deployment added here does not give any signal as is.

> Expect(len(elbsOutput.LoadBalancerDescriptions[0].Instances)).Should(BeNumerically(">=", 1))

> This is true for even if we don't have a deployment that matches the Service. The instances attached to the LB are just Out of Service if there are no pods in the instance.
> You can either add a check that shows the instances with the pods are in service in the LB,
or can remove the deployment all together. 

I added the change because I was asked to here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3343#discussion_r834551091 

It makes the test code flow a little bit better, but I agree we don't _need_ it to test the functionality.

If you're still ok with having it removed, just let me know and I'll do it

sedefsavas:
> I added the change because I was asked to here [#3343 (comment)](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3343#discussion_r834551091)

Don't have a strong preference, we just don't need the code if we are not using it.
@pydctw WDYT?

sedefsavas:
Also there is a way to check if the instances are in service in the ELBs with (c *ELB) DescribeInstanceHealth(input *DescribeInstanceHealthInput).

If this check is added to verifyELB, then adding a deployment will make sense IMO.

faiq:
> Also there is a way to check if the instances are in service in the ELBs with (c *ELB) DescribeInstanceHealth(input *DescribeInstanceHealthInput).
> If this check is added to verifyELB, then adding a deployment will make sense IMO.

I'm a little confused on your ask here. To me it sounds like you want me to:

1. get the node the nginx deployment is running on 
2. check to see if that DescribeInstanceHealth to see if the node is in the list 


i would prefer to simply remove the deployment

Issue #3244, opened 2022-02-22
Ability to add custom environment variables to the `aws-node` DaemonSet


charlie-haley:
/kind feature

**Describe the solution you'd like**
Ability to add custom environment variables to the `aws-node` DaemonSet, ideally through an existing CRD such as `AWSManagedControlPlane`

**Anything else you would like to add:**
I've recently needed to bump up the pod limits on our EKS clusters and as a result, needed to set `ENABLE_PREFIX_DELEGATION=true` on the `aws-node` DaemonSet - having the ability to modify the env vars on the DS would be very useful! 

Happy to raise a PR if this solution is accepted.

**Environment:**

- Cluster-api-provider-aws version: v1.3.0 
- Kubernetes version: (use `kubectl version`): v1.21.5
- OS (e.g. from `/etc/os-release`): Amazon Linux 2

richardcase:
Thanks @charlie-haley. I think this is a reasonable request and if you'd like to create a PR that would be fantastic....feel free to reach out here or in slack if you need any help.

/triage accepted
/priority important-longterm
/help

richardcase:
I meant to say, feel free to assign the issue to yourself.

charlie-haley:
Awesome, thanks @richardcase!

/assign @charlie-haley

richardcase:
@charlie-haley - we have some existing code around aws-node...it may be worth looking at [this package](https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/24e3ea2e372e5c6deff3c32bb6e72e4207f0f24a/pkg/cloud/services/awsnode).

PR #3368, opened 2022-03-30
Apidiff test runs only if there are changes in api/ or exp/api/

meghanajangi:
Signed-off-by: Meghana Jangi <mjangi@vmware.com>

**What type of PR is this?**
/kind failing-test

**What this PR does / why we need it**:
This PR will limit the apidiff test to run only if there are changes made in the directory api/ and exp/api/

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3282 

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #3374, opened 2022-04-01
feat: add custom environment variables to the 'aws-node' DaemonSet

charlie-haley:
**What type of PR is this?**
/kind feature

Ability to add custom environment variables to the aws-node DaemonSet. This is mostly so you can increase pod limits on nodes through the `ENABLE_PREFIX_DELEGATION` environment variable, but this PR supports all environment variables.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3244

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests


richardcase:
@charlie-haley - For the verify issue i think you need to run `go mod tidy` and test failure looks to me a lint problem....worth just checking with `make lint`. Lets chat Monday morning on slack.

richardcase:
And thanks for working on this :+1:

pydctw:
> @charlie-haley - For the verify issue i think you need to run `go mod tidy`. The linter error is a weird one. Lets chat Monday morning on slack.

The linter error message is cryptic but basically the import order needs to be changed.

pkg/cloud/services/awsnode/service.go
```diff
@@ -19,6 +19,7 @@ package awsnode
 import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	corev1 "k8s.io/api/core/v1"
```
pydctw, line 22:
```suggestion
	corev1 "k8s.io/api/core/v1"
	"sigs.k8s.io/controller-runtime/pkg/client"
```

docs/book/src/topics/eks/pod-networking.md
```diff
@@ -7,6 +7,27 @@ When creating a EKS cluster the Amazon VPC CNI will be used by default for Pod N
...
+```yaml
+kind: AWSManagedControlPlane
+apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+metadata:
+  name: "capi-managed-test-control-plane"
+spec:
+  awsNodeEnvVars:
+  - name: ENABLE_PREFIX_DELEGATION
+    value: "true"
+  addons:
```
pydctw, line 23: Is there a reason to specify vpc-cni in addons when`disableVPCCNI: false` is already set? Is it to pick a specific version?

charlie-haley: No particular reason, I just wanted to be as explicit as possible with the example configuration! Happy to change it if you think it's redundant?

pydctw: Make sense. In that case, can we change the version to sth like REPLACE_WITH_VPC_CNI_VERSION instead of a specific version? As this is a doc, I worry that it will get outdated and we will need to update it regularly.

pydctw:
pull-cluster-api-provider-aws-verify - looks like generated conversion files are out of date. Could you run `make generate`? 
```
generated files are out of date, run make generate
make: *** [Makefile:319: verify-gen] Error 1
```

pkg/cloud/services/awsnode/service.go
```diff
@@ -17,10 +17,10 @@ limitations under the License.
 package awsnode
 
 import (
-	"sigs.k8s.io/controller-runtime/pkg/client"
-
+	corev1 "k8s.io/api/core/v1"
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
pydctw, line 23: So for this linter error, make lint-fix was supposed to fix the import order automatically but looks like gci's latest upgrade broke it - https://github.com/golangci/golangci-lint/issues/2604

As defined https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/.golangci.yml, we group imports in three parts. 

```suggestion
         corev1 "k8s.io/api/core/v1"
	"sigs.k8s.io/controller-runtime/pkg/client"
	
	infrav1 "sigs.k8s.io/cluster-api-provider-aws/api/v1beta1"
	"sigs.k8s.io/cluster-api-provider-aws/pkg/cloud"
```

charlie-haley:
> pull-cluster-api-provider-aws-verify - looks like generated conversion files are out of date. Could you run `make generate`?
> 
> ```
> generated files are out of date, run make generate
> make: *** [Makefile:319: verify-gen] Error 1
> ```

So, this is also confusing me, I've tried running `make generate` multiple times and I don't get anything new in the diff. The makefile target doesn't return any errors either

pkg/cloud/services/awsnode/service.go
charlie-haley: Looks like that worked 🎉, thanks @pydctw

pydctw:
e2e test result looks good to me

```
I0406 16:46:04.492514       1 cni.go:44]  "msg"="Reconciling aws-node DaemonSet in cluster"  "cluster-name"="cluster-yknmkm" "cluster-namespace"="eks-nodes-lrf3qi"
I0406 16:46:04.966908       1 cni.go:68]  "msg"="updating aws-node daemonset environment variables"  "cluster-name"="cluster-yknmkm" "cluster-namespace"="eks-nodes-lrf3qi"
```

[AWSManagedControlPlane](https://storage.googleapis.com/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/3374/pull-cluster-api-provider-aws-e2e-eks/1511730882350682112/artifacts/clusters/bootstrap/resources/eks-nodes-lrf3qi/AWSManagedControlPlane/cluster-yknmkm-control-plane.yaml)
```
spec:
  addons:
  - conflictResolution: overwrite
    name: vpc-cni
    version: v1.8.0-eksbuild.1
  - conflictResolution: overwrite
    name: coredns
    version: v1.8.3-eksbuild.1
  associateOIDCProvider: false
  awsNodeEnvVars:
  - name: FOO
    value: BAR
  - name: ENABLE_PREFIX_DELEGATION
    value: "true"
```

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
```diff
@@ -166,6 +167,10 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
 	// Amazon VPC CNI addon or if you have specified a secondary CIDR block.
 	// +kubebuilder:default=false
 	DisableVPCCNI bool `json:"disableVPCCNI,omitempty"`
+
+	// AWSNodeEnvVars defines a list of environment variables to apply to the `aws-node` DaemonSet
+	// +optional
+	AWSNodeEnvVars []corev1.EnvVar `json:"awsNodeEnvVars,omitempty"`
```
richardcase, line 173: I'm wondering if we create a new struct called `VPCCNI` or `AWSNode` that has a field called `EnvVars` instead? 

We could then do a follow-up change  to add a `Disable` field and start deprecation of  `DisableVPCCNI` so that anything related to aws node/the vpc cni is on its own struct.

wdyt?

pkg/cloud/services/awsnode/cni.go
```diff
@@ -67,6 +63,41 @@ func (s *Service) ReconcileCNI(ctx context.Context) error {
...
+				}
+				container.Env = append(existingVars, envVars...)
+				break
+			}
+		}
+		if s.scope.SecondaryCidrBlock() == nil {
+			err = remoteClient.Update(ctx, &ds, &client.UpdateOptions{})
+			if err != nil {
+				return err
+			}
```
richardcase, line 93: If the secondary cidr block is nil we could return here instead of doing the check again outside this block.

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
richardcase: Yes something like :+1:  We can then follow up later with:

```yaml
spec:
  VPCCNI:
    disable: true
```

charlie-haley:
Not sure if this will work, but I'll trigger the e2e to run again to ensure there hasn't been any regression.

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
```diff
@@ -166,6 +167,17 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
...
+	// VPCCNI defines a list of environment variables to apply to the `aws-node` DaemonSet
+	// +optional
+	VPCCNI VPCCNI `json:"VPCCNI,omitempty"`
+}
+
+// VPCCNI specifies configuration related to the VPC CNI.
+type VPCCNI struct {
+	// Environment defines a list of environment variables to apply to the `aws-node` DaemonSet
+	// +optional
+	Environment []corev1.EnvVar `json:"environment,omitempty"`
```
pydctw, line 180: I would prefer `Env` (similar to how env vars are defined in a [container](https://github.com/kubernetes/api/blob/7a897301bd567648ff3b967902baafca9f589d8f/core/v1/types.go#L2354))  or `EnvVars` rather than `Environment` as a field name but up to you.

pydctw:
apidiff failure is fine. It is just informing that there are API changes.
```bash
sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/services/awsnode
  Incompatible changes:
  - Scope.VPCCNI: added
sigs.k8s.io/cluster-api-provider-aws/pkg/cloud/scope
  Compatible changes:
  - (*ManagedControlPlaneScope).VPCCNI: added
sigs.k8s.io/cluster-api-provider-aws/controlplane/eks/api/v1beta1
  Compatible changes:
  - AWSManagedControlPlaneSpec.VPCCNI: added
  - VPCCNI: added
make: *** [Makefile:329: apidiff] Error 1
```

@meghanajangi, I thought you modified apidiff to run only on actual API changes. Is this test result expected?
* https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3368

pydctw:
Ah, needed rebase to get the latest apidiff. @meghanajangi, please ignore my previous question.
Thanks @charlie-haley for all your hard work.

/lgtm

mloiseleur:
Hello,

Any news on this, @richardcase ?

Skarlso:
Hello.

Really need this for IPv6 because the was cni needs extra environment properties.

Reviewing this PR.... 

@richardcase Ping for visibility. :)

controlplane/eks/api/v1alpha3/zz_generated.conversion.go
```diff
@@ -365,6 +365,7 @@ func autoConvert_v1beta1_AWSManagedControlPlaneSpec_To_v1alpha3_AWSManagedContro
 	out.Addons = (*[]Addon)(unsafe.Pointer(in.Addons))
 	// WARNING: in.OIDCIdentityProviderConfig requires manual conversion: does not exist in peer-type
 	out.DisableVPCCNI = in.DisableVPCCNI
+	// WARNING: in.VPCCNI requires manual conversion: does not exist in peer-type
```
Skarlso, line 368: @richardcase There needs to be something happening for this, right? :D

controlplane/eks/api/v1alpha4/conversion.go
```diff
@@ -23,21 +23,41 @@ import (
...
+	utilconversion "sigs.k8s.io/cluster-api/util/conversion"
 	"sigs.k8s.io/controller-runtime/pkg/conversion"
 )
 
 // ConvertTo converts the v1alpha4 AWSManagedControlPlane receiver to a v1beta1 AWSManagedControlPlane.
 func (r *AWSManagedControlPlane) ConvertTo(dstRaw conversion.Hub) error {
 	dst := dstRaw.(*v1beta1.AWSManagedControlPlane)
 
-	return Convert_v1alpha4_AWSManagedControlPlane_To_v1beta1_AWSManagedControlPlane(r, dst, nil)
+	if err := Convert_v1alpha4_AWSManagedControlPlane_To_v1beta1_AWSManagedControlPlane(r, dst, nil); err != nil {
```
Skarlso, line 34: @richardcase This is what's needed for the newly added fields for ipv6 too, right?

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
```diff
@@ -166,6 +167,17 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
 	// Amazon VPC CNI addon or if you have specified a secondary CIDR block.
 	// +kubebuilder:default=false
 	DisableVPCCNI bool `json:"disableVPCCNI,omitempty"`
+
+	// VPCCNI defines a list of environment variables to apply to the `aws-node` DaemonSet
+	// +optional
+	VPCCNI VPCCNI `json:"VPCCNI,omitempty"`
```
Skarlso, line 173: Hmmm, this shouldn't be fully uppercase, right?

pkg/cloud/services/awsnode/cni.go
```diff
@@ -67,6 +63,41 @@ func (s *Service) ReconcileCNI(ctx context.Context) error {
...
+		}
+	}
+
+	if s.scope.SecondaryCidrBlock() == nil {
+		if len(envVars) > 0 {
+			err = remoteClient.Update(ctx, &ds, &client.UpdateOptions{})
+			if err != nil {
+				return err
+			}
```
Skarlso, line 96:
```suggestion
			if err = remoteClient.Update(ctx, &ds, &client.UpdateOptions{}); err != nil {
				return err
			}
```

Skarlso:
This looks okay to me as well. Would be nice to see some manual test run logs? :)

Skarlso:
I'm wondering what this would do if, for example, there would be no CNI installed, but we change the environment properties.

Previously if you didn't have a secondary CIDR block defined it would just be nil.

But now it would start returning 
```go
ErrCNIMissing
```
I wonder if this would break any kind of existing flows. I guess, I'm just thinking about what that change would do to existing environments or changes to the config files. Should it create it, for example?

For example, I want to have a vpc-cni and set some environment properties. I guess that could be a different PR. It's surely needed for ipv6. I could do it in that PR.

controlplane/eks/api/v1alpha4/conversion.go
richardcase: Yes exactly :+1: 

Sorry i didn't get around to sending you the example

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
richardcase: @Skarlso - do you think it should be `VpcCni` instead?

richardcase:
> I wonder if this would break any kind of existing flows. I guess, I'm just thinking about what that change would do to existing environments or changes to the config files. Should it create it, for example?

@Skarlso - thats a good spot. Yes this would break the reconciliation of `AWSManagedControlPlane` in the scenario where we delete the vpc cni.

I would say that if the ds is not found then we continue on but this also means changing some of the later logic to be optional if we are deleting.

richardcase:
@charlie-haley - i'm keen to get this into the next release. Any chance you'd be able to rebase?

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
charlie-haley: Just to weigh in on this, I think `VPCCNI` makes more sense due to it being an acronym, it also matches the previous stylisation of `DisableVPCCNI`.

Happy to change it however if we think something like `VpcCni` or `VpcCNI` makes more sense

richardcase: I think all options don't sit well....2 acronyms together :vomiting_face: My thoughts on the options:

1. `VPCCNI` - i tend to agree with @charlie-haley, as they are acronyms capitals sort of makes sense and it does fit with the 
`DisableVPCCNI` field. But the `DisableVPCCNI` will be removed in **v1beta2**.
2. `VpcCNI` - i don't like this as it's mixing capitalization for acronyms
3.  `VpcCni`- if there is a strong option against fully caps then this would be my choice.

I don't have a super strong opinion apart from not choosing option 2. @Skarlso @sedefsavas  (or anyone else) any other views?

charlie-haley: I don't have strong feelings about this so I took the liberty of updating the PR to use `VpcCni`

charlie-haley:
Not sure what's happening here, I don't get any changes when running `make generate` locally, I also rebased onto master (which fixed the apidiff issue) any ideas @richardcase @pydctw?

richardcase:
> Not sure what's happening here, I don't get any changes when running `make generate` locally, I also rebased onto master (which fixed the apidiff issue) any ideas @richardcase @pydctw?

Interesting, i've seen this on another PR. Could you do the following to see if you get a diferent result:

```bash
make clean
make generate
```

richardcase:
@charlie-haley - there was an issue with code generation. If you rebase your PR on main this issue with the verify job should go away.

charlie-haley:
> @charlie-haley - there was an issue with code generation. If you rebase your PR on main this issue with the verify job should go away.

That did the trick, thanks @richardcase

mloiseleur:
@richardcase @pydctw @dlipovetsky Ping for visibility. This PR is ready to be merged.

richardcase:
The e2e tests are passing.

Thanks for your work on this @charlie-haley :bowing_man:  From my side:

/approve

@dlipovetsky @Ankitasw - any chance you could take a look and when you are happy lgtm?

Ankitasw:
@richardcase I will take a look.

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
```diff
@@ -167,6 +168,10 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
 	// +kubebuilder:default=false
 	DisableVPCCNI bool `json:"disableVPCCNI,omitempty"`
 
+	// VpcCni defines a list of environment variables to apply to the `aws-node` DaemonSet
```
Ankitasw, line 171: This is added right now for use case of aws-node DaemonSet but can be used for other env configuration in VPC CNI, so the comment should be generic

Ankitasw (commented):
nit, other LGTM

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
richardcase: Good point @Ankitasw :+1: 

Perhaps something like `// VpcCni is used to set configuration options for the VPC CNI (i.e. aws-node daemonset) such as environment variables.`

@charlie-haley - any chance you could change it to something a bit more generic and generate the crd definition again? After that, i think we are good to go.

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
```diff
@@ -167,6 +168,10 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
 	// +kubebuilder:default=false
 	DisableVPCCNI bool `json:"disableVPCCNI,omitempty"`
 
+	// VpcCni is used to set configuration options for the VPC CNI plugin
+	// +optional
+	VpcCni VpcCni `json:"VpcCni,omitempty"`
```
Skarlso, line 173: @richardcase Ah I haven't noticed that but this should be lowercase.

I'll fix it in my PR. :)

PR #3388, opened 2022-04-05
Fix for release log creation script

sedefsavas:
/kind bug
Removed getopts logic from the release notes script. The variables set in the Makefile that calls this script are passed to the script. getops was not working properly, values were being overriden by Makefile because same variables names were being used in the script.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -99,6 +99,7 @@ ALL_ARCH ?= amd64 arm arm64 ppc64le s390x
 # main controller
 CORE_IMAGE_NAME ?= cluster-api-aws-controller
 CORE_CONTROLLER_IMG ?= $(REGISTRY)/$(CORE_IMAGE_NAME)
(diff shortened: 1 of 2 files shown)


richardcase:
This is an interesting one as i asked @meghanajangi to add the flags so that it would make it easier for people to run this interactively from the command line. I'm no overly keen on scripts that are dependent on the right variables being passed in from the makefile...i prefer to make it more explicit. In hindsight, we should've assigned the flag values to local variable in the shell script instead.

sedefsavas:
I agree, I did not think it through, this is a workaround I used.
I can close this if we want to create an issue.

richardcase:
> I agree, I did not think it through, this is a workaround I used. I can close this if we want to create an issue.

We can create a follow-up issue. The change you made works and gets the release process working again so:

/lgtm
/approve

Issue #1746, opened 2020-06-11
Validate AWS tags


richardcase:
/kind feature

**Describe the solution you'd like**
When we create resources in AWS we add a number of Tags to the resources. There are certain restrictions on the tags, such as a maximum of 256 characters for a tag value.

There are variouus places in the code where we create these tags, for example:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d4ade5922e04a2327f4043effd16252126dea223/pkg/cloud/services/ec2/subnets.go#L323
and we should probably validate the tags we want to create.

The details around the limits and requirements can be found in the AWS documentation:
https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

rudoi:
/assign 

Will add context here such that this could be good-first-issue.

evalsocket:
@detiber @rudoi Can this assign to me ?

detiber:
/assign @evalsocket 

@evalsocket I believe you should be able to self-assign issues as long as you've commented on the issue (a comment with just `/assign` on it's own line should be sufficient to meet the requirement and for the bot to do the assignment)

evalsocket:
@richardcase what should we do if validation failed. Should we skip the tag or through the error?

richardcase:
I would say throw an error. The reason being that a lot of people rely in tags for assigning costs to different teams and without the lables it will cause headaches.

VibhorChinda:
Hii @Ankitasw @richardcase I was looking into the description of this issue and understood the following :
1. Tags are created for the resources which we create in AWS [System tags (AWS: prefix), user tags]
2. We need to validate the tags which user creates for the resources. Following the rules mentioned here under "Tag naming limits and requirements" [https://docs.aws.amazon.com/general/latest/gr/aws_tagging.html#tag-best-practices](url)

So the possible solution can be creating a function which treats a tag as a simple string and check all the rules on it.

Please guide, if I am thinking in right direction or not. 
And if yes how can I achieve this :))

richardcase:
[...]

PR #3398, opened 2022-04-08
Validate Tags created for the resources

VibhorChinda:
**What type of PR is this?**
/kind feature

This PR validates the tags created for resources
1. Added code for checking max number of user tags which can be created for a resource [50 user tags can be created]
2. Added code for checking characters which can be used in user tags created for a resource [letters, numbers, spaces and the following characters: _ . : / = + - @]

Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1746

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/api/v1beta1/tags.go
+++ b/api/v1beta1/tags.go
@@ -18,6 +18,7 @@ package v1beta1
 
 import (
 	"fmt"
+	"regexp"
 
 	"github.com/google/go-cmp/cmp"
 	"k8s.io/apimachinery/pkg/types"
@@ -74,10 +75,18 @@ func (t Tags) Merge(other Tags) {
 	}
 }
 
-// Validate checks if tags are valid for the AWS API. Keys must have at
-// least 1 character and max 128. Values must be max 256 characters long.
+// Validate checks if tags are valid for the AWS API/Resources.
+// Keys must have at least 1 and max 128 characters.
+// Values must be max 256 characters long.
+// Keys and Values can only have alphabets, numbers, spaces and _ . : / = + - @ as characters.
+// Tag's key cannot have prefix "aws:".
+// Max count of User tags for a specific resource can be 50.
 func (t Tags) Validate() []*field.Error {
+	// Defines the maximum number of user tags which can be created for a specific resource
+	const maxUserTagsAllowed = 50
 	var errs field.ErrorList
+	var userTagCount = len(t)
+	re := regexp.MustCompile(`^[a-zA-Z0-9\\s\_\.\:\=\+\-\@\/]*$`)
 
 	for k, v := range t {
 		if len(k) < 1 {
@@ -95,9 +104,37 @@ func (t Tags) Validate() []*field.Error {
 				field.Invalid(field.NewPath("spec", "additionalTags"), v, "value cannot be longer than 256 characters"),
 			)
 		}
+		if wrongUserTagNomenclature(k) {
+			errs = append(errs,
+				field.Invalid(field.NewPath("spec", "additionalTags"), k, "user created tag's key cannot have prefix aws:"),
+			)
+		}
+		val := re.MatchString(k)
+		if !val {
+			errs = append(errs,
+				field.Invalid(field.NewPath("spec", "additionalTags"), k, "key cannot have characters other than alphabets, numbers, spaces and _ . : / = + - @ ."),
+			)
+		}
+		val = re.MatchString(v)
+		if !val {
+			errs = append(errs,
+				field.Invalid(field.NewPath("spec", "additionalTags"), v, "value cannot have characters other than alphabets, numbers, spaces and _ . : / = + - @ ."),
+			)
+		}
+	}
+
+	if userTagCount > maxUserTagsAllowed {
+		errs = append(errs,
+			field.Invalid(field.NewPath("spec", "additionalTags"), t, "user created tags cannot be more than 50"),
+		)
 	}
 
 	return errs
+}
+
+// Checks whether the tag created is user tag or not.
+func wrongUserTagNomenclature(k string) bool {
+	return len(k) > 3 && k[0:4] == "aws:"
 }
 
 // ResourceLifecycle configures the lifecycle of a resource.
--- a/api/v1beta1/tags_test.go
+++ b/api/v1beta1/tags_test.go
@@ -271,6 +271,68 @@ func TestTags_Validate(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "key has aws: prefix",
+			self: Tags{
+				"aws:key": "validValue",
+			},
+			expected: []*field.Error{
+				{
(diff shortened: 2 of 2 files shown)


VibhorChinda:
@richardcase please have a look :))

VibhorChinda:
All tests passed @Ankitasw :)))))

api/v1beta1/tags.go
```diff
@@ -95,6 +98,27 @@ func (t Tags) Validate() []*field.Error {
...
+		}
+		val = re.MatchString(v)
+		if !val {
+			errs = append(errs,
+				field.Invalid(field.NewPath("spec", "additionalTags"), v, "value cannot have characters other than alphabets, numbers, spaces and _ . : / = + - @ ."),
+			)
+		}
+	}
+
+	if userTagCount > 50 {
```
Ankitasw, line 118: Can we define a constant for 50, with a comment indicating what this number represents?

api/v1beta1/tags.go
```diff
@@ -95,6 +98,27 @@ func (t Tags) Validate() []*field.Error {
 				field.Invalid(field.NewPath("spec", "additionalTags"), v, "value cannot be longer than 256 characters"),
 			)
 		}
+		if len(k) > 3 && k[0:4] != "aws:" {
```
Ankitasw, line 101: Can we add unit tests for the new validations, so as to check that the logic works fine?

VibhorChinda:
All tests passing @richardcase @Ankitasw @sedefsavas 👀

To the merge 🚀🚀 :))

Ankitasw (commented):
As discussed in office hours, it would be good to add a small documentation about the tagging best practices, as we cannot handle that in our code.

VibhorChinda:
Makes sense @Ankitasw will do. 

Could you please point to the place where I could add documentation :))

api/v1beta1/tags.go
Ankitasw: Can we refactor this code to a func, something like:
```go
func isUserOwnedTag() bool {
   return len(k) > 3 && k[0:4] != "aws:" 
}
```
so as to improve code readability?

Ankitasw:
> Could you please point to the place where I could add documentation :))

We have a doc where we list the conventions to be followed in code, I think we can add these changes [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/book/src/development/conventions.md).

api/v1beta1/tags.go
richardcase: Yes this would be good

VibhorChinda:
> > Could you please point to the place where I could add documentation :))
> 
> We have a doc where we list the conventions to be followed in code, I think we can add these changes [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/book/src/development/conventions.md).

@Ankitasw  @richardcase I was wondering not to add any documentation/check for this. Because

If we say users should create tags with consistent naming convention and give examples like "costcenter" "CostCenter" etc

How would user be able to add characters in the naming thing ?? [As characters like "."  "+"  "-"  "/"  "@" etc are allowed and users can use them while naming tags.

What do you think ??

api/v1beta1/tags.go
```diff
@@ -95,11 +99,37 @@ func (t Tags) Validate() []*field.Error {
...
+			field.Invalid(field.NewPath("spec", "additionalTags"), t, "user created tags cannot be more than 50"),
+		)
 	}
 
 	return errs
 }
 
+// Checks whether the tag created is user tag or not.
+func isUserOwnedTag(k string) bool {
+	return len(k) > 3 && k[0:4] != "aws:"
```
sedefsavas, line 130: Users cannot create tags start with `aws:`, this check is not needed. But we can add a check to disallow adding tags  with `aws:` prefix.

api/v1beta1/tags.go
```diff
@@ -95,11 +99,37 @@ func (t Tags) Validate() []*field.Error {
 				field.Invalid(field.NewPath("spec", "additionalTags"), v, "value cannot be longer than 256 characters"),
 			)
 		}
+		if isUserOwnedTag(k) {
```
sedefsavas, line 102: Before here, we should return error if tag starts with `aws:` and here just count all the tags.

api/v1beta1/tags.go
```diff
@@ -74,10 +75,13 @@ func (t Tags) Merge(other Tags) {
 	}
 }
 
-// Validate checks if tags are valid for the AWS API. Keys must have at
-// least 1 character and max 128. Values must be max 256 characters long.
+// Validate checks if tags are valid for the AWS API/Resources.
```
sedefsavas, line 78: Keeping validations we have in this method as comment is useful as we only have checks for a subset of them.

api/v1beta1/tags.go
VibhorChinda: I thought the list of tags we get to validate contained both user and system tags.

I tried to calculate the number of user tags we create because the upper limit is 50 for that. Thus having a check to know whether the tag is system created or not for counting purposes was important.

This is what I thought :))

what does everyone suggest ??

api/v1beta1/tags.go
VibhorChinda: okay will add comments describing the checks

richardcase:
> @Ankitasw @richardcase I was wondering not to add any documentation/check for this. Because
> 
> If we say users should create tags with consistent naming convention and give examples like "costcenter" "CostCenter" etc
> 
> How would user be able to add characters in the naming thing ?? [As characters like "." "+" "-" "/" "@" etc are allowed and users can use them while naming tags.
> 
> What do you think ??

@VibhorChinda i would agree that we can stay away from the "best practice" part and leave that up to the user. So lets not add documentation around the `costcentre` vs `CostCentre` advice. I see that @sedefsavas has also left a comment with the same thoughts.

api/v1beta1/tags.go
VibhorChinda: Can I get some clearance here please @richardcase @sedefsavas 
Kinda confused here :))

sedefsavas: additionalTags that we use this method for do not include system tags.
We can add a check to disallow tags with system prefix in this validation method.
I tried manually, AWS already errors when user tag name starts with `aws:`

Since additional tags are not supposed to have system tags, we don't need to check if a tag is a system tag here for the limit calculation.

VibhorChinda: cool thanks understood :))
Will soon commit changes.

VibhorChinda:
@richardcase @sedefsavas have a look please :))

VibhorChinda:
Yayyyyyyy :)) 
Onto the next 🚀🚀

PR #3460, opened 2022-05-05
bugfix: Fix log level for a couple of message of log messages

### LINKED ISSUE #3454: Some log messages are misleading and probably need to be moved to a lower level
/kind bug

`awsmachine_controller.go:763]  "msg"="Deleting unneeded entry from AWS S3"  "secretPrefix"=""`
and
`awsmachine_controller.go:432]  "msg"="Reconciling AWSMachine"`

occur with great frequency in the logs, and neither provides much information or insight into what's going on.  Additionally, CAPA is not actually deleting anything from AWS S3, and so might lead to a misleading action that doesn't actually occur.

I believe we would be best served by moving these out of `INFO` and into debug.

**Environment:**

- Cluster-api-provider-aws version: `v1.4.1`
- Kubernetes version: (use `kubectl version`): `v1.23.4`

tasdikrahman:
**What type of PR is this?**

/kind cleanup

Fixes #3454

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -429,7 +429,7 @@ func (r *AWSMachineReconciler) findInstance(scope *scope.MachineScope, ec2svc se
 }
 
 func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *scope.MachineScope, clusterScope cloud.ClusterScoper, ec2Scope scope.EC2Scope, elbScope scope.ELBScope, objectStoreScope scope.S3Scope) (ctrl.Result, error) {
-	machineScope.Info("Reconciling AWSMachine")
+	machineScope.V(4).Info("Reconciling AWSMachine")
 
 	// If the AWSMachine is in an error state, return early.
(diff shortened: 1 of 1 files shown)


shivi28:
In the issue #3454 we were discussing on handling
```
We should not log this unless ignition is used.
awsmachine_controller.go:763] "msg"="Deleting unneeded entry from AWS S3" "secretPrefix"=""
```
Are we not handling this scenario in this PR only? May be add [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmachine_controller.go#L763) line after [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmachine_controller.go#L774).

Otherwise changes lgtm.

tasdikrahman:
Hello @shivi28 , I tried incorporating your changes, please have a look

sedefsavas:
/approve

cc @voor for review.

PR #3512, opened 2022-06-08
Removed additionalTag application on unmamaged subnets

### LINKED ISSUE #3494: Additional Tags should not apply to unmanaged subnets
/kind bug

When using existing subnets and `additionalTags` the tags are getting applied to the unmanaged subnets and overwriting any tags that exist on the subnets. 

existing resources(unmanaged subnets) should not have tags applied to them since capi does not own those resources

**Environment:**

- Cluster-api-provider-aws version: v1.2.0
- Kubernetes version: (use `kubectl version`):  1.22.x
- OS (e.g. from `/etc/os-release`): aws-linux

Ankitasw:
**What type of PR is this?**
/kind bug

This PR removes the adding additional tags from subnets in case of using BYO VPC. This makes the tagging behavior consistent across all the network components(like routetables, etc.).

Fixes #3494 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


pkg/cloud/services/network/subnets.go
```diff
@@ -451,7 +451,6 @@ func (s *Service) getSubnetTagParams(unmanagedVPC bool, id string, public bool,
 	} else {
 		return infrav1.BuildParams{
 			ResourceID: id,
-			Additional: additionalTags,
```
sedefsavas, line 454: crit. There is a logic above that adds tags for CCM ,we still want those tags to be added to the unmanaged subnets so only additionalTags added by user should be removed from unmanaged subnets.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7f094b6e315e54b89e2aa3385d7838244c3e2c87/pkg/cloud/services/network/subnets.go#L422

pkg/cloud/services/network/subnets_test.go
```diff
@@ -224,137 +193,9 @@ func TestReconcileSubnets(t *testing.T) {
...
-
-				m.CreateTags(gomock.Eq(&ec2.CreateTagsInput{
-					Resources: aws.StringSlice([]string{"subnet-2"}),
-					Tags: []*ec2.Tag{
-						{
-							Key:   aws.String("kubernetes.io/cluster/test-cluster"),
-							Value: aws.String("shared"),
-						},
-						{
-							Key:   aws.String("kubernetes.io/role/internal-elb"),
```
sedefsavas, line 251: We want these tags to be tested for unmanaged subnets too after fixing the above comment.

sedefsavas:
Also, adding a note to external infrastructure doc for this behaviour will be useful. I think there is already a section stating that we add CCM tags for unmanaged resources, but need to clarify additionalTags will not get propagated.

pkg/cloud/services/network/subnets.go
Ankitasw: Done

pkg/cloud/services/network/subnets_test.go
Ankitasw: Done

Ankitasw:
This PR is ready for review.

pkg/cloud/services/network/subnets.go
```diff
@@ -428,11 +428,11 @@ func (s *Service) getSubnetTagParams(unmanagedVPC bool, id string, public bool,
 	// Add tag needed for Service type=LoadBalancer
 	additionalTags[infrav1.NameKubernetesAWSCloudProviderPrefix+s.scope.KubernetesClusterName()] = string(infrav1.ResourceLifecycleShared)
 
-	for k, v := range manualTags {
```
sedefsavas, line 431: `additionalTags := s.scope.AdditionalTags()` in line 418 already adds the additional tags from `s.AWSCluster.Spec.AdditionalTags`. 

That is what we are trying to avoid.

I don't know if manual tags are also relevant here, guessing not.

Ankitasw: Done

Ankitasw: For safer side, I have removed manual tags as well from unmanaged subnets.

sedefsavas: And what was in the manual tags?

Ankitasw: Manual tags are existing tags on that subnet which is fetched from scope, I think we need to retain it here, wdyt @sedefsavas ?

Ankitasw: Any views here @sedefsavas ?

richardcase: Our networking code really needs a complete re-write!

Ankitasw: I agree 😞

richardcase: I can't see this change causing issues.

Ankitasw:
cc @richardcase @sedefsavas for review

Issue #2420, opened 2021-05-20
Add IPv6 support


sedefsavas:
Add support to create IPv6 clusters and add e2e test for it.

/kind feature
/milestone next

sedefsavas:
Reminder:
Create issue for dual-stack
and replicate this for EKS

richardcase:
This gets messy with EKS.

randomvariable:
Support should be fixed in Cluster API, not by shoving lots of non-standard behaviour into CAPA - consistency across providers is paramount. We need to break this down. Please start a doc, we have a team at VMware looking into dual-stack in general who can assist.

richardcase:
@randomvariable  - started a [doc](https://docs.google.com/document/d/1-J3whrf2ewOJ22G3ETZYOAh6zysTf37ecCtttmggzlM/edit?usp=sharing)

randomvariable:
/assign

for initial scoping

yandooo:
Are there any updates here?

yandooo:
@richardcase would it be possible to share relevant documentation for the topic if any? As per [comment](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2420#issuecomment-959022716)  there seems like some progress is made towards standardizing CAPI to support dual-stack. Thanks

richardcase:
There isn't a huge amount of documentation. The doc mentioned in the thread was the start of a doc to capture notes but we didn't much further. Do you have access to that doc?

yandooo:
@richardcase don't have it. I requested access linking your comment in the request message. Appreciated it if it can be shared. Thanks

richardcase:
> @richardcase don't have it. I requested access linking your comment in the request message. Appreciated it if it can be shared. Thanks

Just sent you an email @yandooo :smile:

yandooo:
I tend to agree on getting IPv6 out first and look at dual stack cohesively to unlock AWS EKS ipv6 feature @sedefsavas.

richardcase:
I also think that adding IPv6 first is a good way forward.

richardcase:
@Skarlso - this was the issue i was thinking og.

Skarlso:
Hi! :) 

Let's tackle this! :) We have some experience creating the necessary resources in eksctl, so hopefully, I can be of assistance here.

nikimanoledaki:
[...]

PR #3513, opened 2022-06-08
Add ipv6 support for EKS

Skarlso:
**What type of PR is this?**

/kind feature

Part 1 for https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2420.

We are aiming to do as less as possible so the PR doesn't get too much bloated.

This will be Part 1 for IPv6 support. What does Part 1 contain you might ask?

- create an ipv6 cluster
- refactor the code to offer both options without needing to use a lot of `if`s
- validations for things that ipv6 doesn't support and capa wouldn't support in the beginning

E2E test run for EKS cluster:

```
Ran 1 of 3 Specs in 1488.316 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 2 Skipped
```

**Checklist**:

- [ ] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release Notes**:

To enable permissions to use ipv6 networking, `clusterawsadm` needs to be executed with additional prerequisites.
Please refer to the prerequisites documentation for details on what to set in the `clusterawsadm` config file.


api/v1alpha4/zz_generated.conversion.go
```diff
@@ -2170,18 +2187,16 @@ func Convert_v1alpha4_VPCSpec_To_v1beta1_VPCSpec(in *VPCSpec, out *v1beta1.VPCSp
 func autoConvert_v1beta1_VPCSpec_To_v1alpha4_VPCSpec(in *v1beta1.VPCSpec, out *VPCSpec, s conversion.Scope) error {
 	out.ID = in.ID
 	out.CidrBlock = in.CidrBlock
+	// WARNING: in.EnableIPv6 requires manual conversion: does not exist in peer-type
```
Skarlso, line 2190: @richardcase Halp. :D How do I make this work? :D I'm probably missing something from somewhere? :)

api/v1beta1/network_types.go
```diff
@@ -174,6 +174,20 @@ type VPCSpec struct {
...
 
+	// EnableIPv6 requests an Amazon-provided IPv6 CIDR block with a /56 prefix length for
+	// the VPC. You cannot specify the range of IP addresses, or the size of the
+	// CIDR block.
+	// +optional
+	EnableIPv6 bool `json:"enableIPv6"`
+
+	// IPv6CidrBlock is the CIDR block provided by Amazon when VPC has enabled IPv6.
+	// +optional
+	IPv6CidrBlock string `json:"ipv6CidrBlock,omitempty"`
```
Skarlso, line 185: I was on the fence if this should be a pointer or not. The other one isn't. 🤔 Technically, if we introduce this one, the other one for a fully ipv6 cluster, could also be empty. So if we would make this a pointer we would have to make the other one a pointer as well. So I decided to have this as a concrete value.

api/v1beta1/network_types.go
```diff
@@ -174,6 +174,20 @@ type VPCSpec struct {
...
+
+	// IPv6Pool is the IP pool which must be defined in case of BYO IP is defined.
+	// +optional
+	IPv6Pool string `json:"ipv6Pool,omitempty"`
```
Skarlso, line 189: I'm not using this one yet, but will once BYO is enabled... It is required for bring your own cidr for IPv6.

api/v1alpha4/zz_generated.conversion.go
richardcase: Ah yes, conversions. Lets have a call tomorrow and i can take you through it.

Skarlso:
As a note it's probably not worth wasting run time on this yet as I'm sure I messed up a bunch of tests 😅

pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -175,6 +175,17 @@ func (s *Service) createLaunchTemplateData(scope *scope.MachinePoolScope, imageI
 		KeyName:  sshKeyNamePtr,
 		UserData: pointer.StringPtr(base64.StdEncoding.EncodeToString(userData)),
 	}
+	if s.scope.VPC().EnableIPv6 {
```
Skarlso, line 178: Skarlso: Remove as it might be set automatically.

Skarlso:
Working subnet creation

Screenshot 2022-06-17 at 8 44 08

Skarlso:
Successfully created Ipv6 cluster:

```
I0617 07:01:35.730194     606 eks.go:34]  "msg"="Reconciling EKS control plane"  "cluster-name"="ipv6-cluster" "cluster-namespace"="default"
I0617 07:01:35.730243     606 roles.go:56]  "msg"="Reconciling EKS Control Plane IAM Role"  
I0617 07:01:35.730254     606 roles.go:60]  "msg"="no eks control plane role specified, using default eks control plane role"  
I0617 07:01:35.730262     606 roles.go:67]  "msg"="using eks control plane role"  "role-name"="eks-controlplane.cluster-api-provider-aws.sigs.k8s.io"
I0617 07:01:36.587591     606 roles.go:90]  "msg"="Skipping, EKS control plane role policy assignment as role is unamanged"  
I0617 07:01:36.587657     606 cluster.go:47]  "msg"="Reconciling EKS cluster"  
I0617 07:01:39.233322     606 cluster.go:413]  "msg"="Created EKS cluster in AWS"  "cluster-name"="default_ipv6-cluster-control-plane"
```

Now, testing pod access, machines, communication, etc... :D

Skarlso:
Screenshot 2022-06-17 at 9 16 55

Skarlso:
Let's see if we didn't break anything so far.

/retest

Skarlso:
Screenshot 2022-06-23 at 17 40 15


Hmm... The fuzzing passes locally.

Skarlso:
This fuzzy test thing is very :suspect: 
```
--- FAIL: TestFuzzyConversion (37.63s)
    --- FAIL: TestFuzzyConversion/for_AWSManagedControlPlane (37.63s)
        --- PASS: TestFuzzyConversion/for_AWSManagedControlPlane/spoke-hub-spoke (37.62s)
        --- FAIL: TestFuzzyConversion/for_AWSManagedControlPlane/hub-spoke-hub (0.01s)
FAIL
```
Passes locally.

Skarlso:
Nice!! Now that we know nothing existing is breaking ( as far as tests go :D ) the real work can begin of extending unit tests, adding e2e tests and adding documentation and validation. :)

Skarlso:
Things are still working after refactorings and small fixes:

```
clusterctl describe cluster ipv6-cluster
NAME                                                                READY  SEVERITY  REASON  SINCE  MESSAGE
Cluster/ipv6-cluster                                                True                     10m
├─ControlPlane - AWSManagedControlPlane/ipv6-cluster-control-plane  True                     10m
└─Workers
  └─MachineDeployment/ipv6-cluster-md-0                             True                     79s
    └─Machine/ipv6-cluster-md-0-64d8d795ff-58sm8                    True                     4m31s
```

Skarlso:
**NOTE**: I will be on vacation for two weeks so expect some downtime on this. :) I have not abandoned it, I'm just taking a bit of time off.

Things that will be still added:

- Documentation
- E2E test
- Validations
- Lots of manual testing for security groups and communications

Not much left now. :)

Skarlso:
Nice, the new code skips updating ipv6 if it's defined.
```
I0625 16:22:27.906893       9 cni.go:182]  "msg"="ipv6 environment properties already set, skipping update of vpc-cni addon" 
```

Skarlso:
Niceeee

```
bash-4.2# ping6 2600:1f13:859:ec02:545c::1
PING 2600:1f13:859:ec02:545c::1(2600:1f13:859:ec02:545c::1) 56 data bytes
64 bytes from 2600:1f13:859:ec02:545c::1: icmp_seq=1 ttl=64 time=0.047 ms
64 bytes from 2600:1f13:859:ec02:545c::1: icmp_seq=2 ttl=64 time=0.052 ms
64 bytes from 2600:1f13:859:ec02:545c::1: icmp_seq=3 ttl=64 time=0.049 ms
64 bytes from 2600:1f13:859:ec02:545c::1: icmp_seq=4 ttl=64 time=0.042 ms
```

Skarlso:
Let's create a couple pods and see IP assignment with `inflate`.

Skarlso:
Nice, looking good:

```
➜  ipv6 k get pods -A --kubeconfig test.kubeconfig -o wide
NAMESPACE     NAME                       READY   STATUS    RESTARTS      AGE   IP                                      NODE                                       NOMINATED NODE   READINESS GATES
default       inflate-866ccdf4c8-7lm65   1/1     Running   0             22s   2600:1f13:859:ec02:545c::8              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
default       inflate-866ccdf4c8-ghsmq   1/1     Running   0             22s   2600:1f13:859:ec02:545c::4              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
default       inflate-866ccdf4c8-j7x2s   1/1     Running   0             22s   2600:1f13:859:ec02:545c::6              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
default       inflate-866ccdf4c8-jsngz   0/1     Pending   0             22s   <none>                                  <none>                                     <none>           <none>
default       inflate-866ccdf4c8-jvnhn   0/1     Pending   0             22s   <none>                                  <none>                                     <none>           <none>
default       inflate-866ccdf4c8-m4xnv   1/1     Running   0             22s   2600:1f13:859:ec02:545c::7              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
default       inflate-866ccdf4c8-mlllj   1/1     Running   0             93s   2600:1f13:859:ec02:545c::2              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
default       inflate-866ccdf4c8-nw6dq   1/1     Running   0             22s   2600:1f13:859:ec02:545c::3              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
default       inflate-866ccdf4c8-ptp6s   0/1     Pending   0             22s   <none>                                  <none>                                     <none>           <none>
default       inflate-866ccdf4c8-wxk5n   1/1     Running   0             22s   2600:1f13:859:ec02:545c::5              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
kube-system   aws-node-4jcdw             1/1     Running   1 (10m ago)   12m   2600:1f13:859:ec02:7fb:b410:fe63:20e2   ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
kube-system   coredns-657694c6f4-hdz92   1/1     Running   0             23m   2600:1f13:859:ec02:545c::1              ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
kube-system   coredns-657694c6f4-l9btx   1/1     Running   0             23m   2600:1f13:859:ec02:545c::               ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
kube-system   kube-proxy-42kht           1/1     Running   0             12m   2600:1f13:859:ec02:7fb:b410:fe63:20e2   ip-10-0-93-63.us-west-2.compute.internal   <none>           <none>
```

Skarlso:
Node to node communication is working. This is all private and internal. Now, let's try creating a service and putting nodes into a public subnet.

Skarlso:
I just realised that EgressOnlyInternetGateway creation needs to be part of the Network creation conditions when IPv6 is enabled.

Skarlso:
# Manual testing of creation and deletion of resources with Managed Egress Only Internet Gateway

Creating the cluster...

All good, the egress only internet gateway is there:

Screenshot 2022-06-25 at 22 08 10

This is good, because previously it was just created during routing, but that actually opened up the thing to a bug. During deletion reconcile, CAPA couldn't get rid of the VPC because there was a remaining attachment. The EIWG.

Now, deleting the VPC... 

The VPC: 
Screenshot 2022-06-25 at 22 09 21

The running cluster:

```
k get pods -A --kubeconfig test.kubeconfig
NAMESPACE     NAME                       READY   STATUS    RESTARTS       AGE
kube-system   aws-node-j9z8n             1/1     Running   1 (4m6s ago)   5m54s
kube-system   coredns-657694c6f4-l8c8k   1/1     Running   0              15m
kube-system   coredns-657694c6f4-msfts   1/1     Running   0              15m
kube-system   kube-proxy-m7vqz           1/1     Running   0              5m54s
```

Deleting the resources:

```
k delete -f cluster-template.yaml
cluster.cluster.x-k8s.io "ipv6-cluster" deleted
awsmanagedcontrolplane.controlplane.cluster.x-k8s.io "ipv6-cluster-control-plane" deleted
awsmachinetemplate.infrastructure.cluster.x-k8s.io "ipv6-cluster-md-0" deleted
eksconfigtemplate.bootstrap.cluster.x-k8s.io "ipv6-cluster-md-0" deleted
Error from server (NotFound): error when deleting "cluster-template.yaml": machinedeployments.cluster.x-k8s.io "ipv6-cluster-md-0" not found
➜  ipv6
```

The last error is a bit suspicious still... @richardcase Is that normal?

Now waiting for the VPC to be completely gone...

Screenshot 2022-06-25 at 22 18 58

Yaay, worked. EIGWS:
Screenshot 2022-06-25 at 22 19 29

Also gone. Nice. :)

Logs:

```
I0625 20:18:20.302781      64 vpc.go:280]  "msg"="Deleted VPC" "vpc-id"="vpc-0a4f3ad1bcc21e91b"
I0625 20:18:20.302909      64 network.go:184]  "msg"="Delete network completed successfully" 
I0625 20:18:20.303353      64 recorder.go:103] events "msg"="Normal" "message"="Deleted managed VPC \"vpc-0a4f3ad1bcc21e91b\"" "object"={"kind":"AWSManagedControlPlane","namespace":"default","name":"ipv6-cluster-control-plane","uid":"c921c966-7f2f-4909-9bfc-f88c9cbcd1fd","apiVersion":"controlplane.cluster.x-k8s.io/v1beta1","resourceVersion":"14876"} "reason"="SuccessfulDeleteVPC"
I0625 20:18:20.303378      64 recorder.go:103] events "msg"="Normal" "message"="Deleted managed VPC \"vpc-0a4f3ad1bcc21e91b\"" "object"={"kind":"AWSManagedControlPlane","namespace":"default","name":"ipv6-cluster-control-plane","uid":"c921c966-7f2f-4909-9bfc-f88c9cbcd1fd","apiVersion":"controlplane.cluster.x-k8s.io/v1beta1","resourceVersion":"14876"} "reason"="SuccessfulDeleteVPC"
```

All good. :)

More logs on Egress Only Internet Gateway:

```
I0625 20:18:16.219663      64 egress_only_gateways.go:103]  "msg"="Deleted Egress Only Internet gateway in VPC" "egress-only-internet-gateway-id"="eigw-026ff0f2da2779e85" "vpc-id"="vpc-0a4f3ad1bcc21e91b"
I0625 20:18:16.220200      64 recorder.go:103] events "msg"="Normal" "message"="Deleted Egress Only Internet Gateway \"eigw-026ff0f2da2779e85\" previously attached to VPC \"vpc-0a4f3ad1bcc21e91b\"" "object"={"kind":"AWSManagedControlPlane","namespace":"default","name":"ipv6-cluster-control-plane","uid":"c921c966-7f2f-4909-9bfc-f88c9cbcd1fd","apiVersion":"controlplane.cluster.x-k8s.io/v1beta1","resourceVersion":"14876"} "reason"="SuccessfulDeleteEgressOnlyInternetGateway"
I0625 20:18:16.221264      64 recorder.go:103] events "msg"="Normal" "message"="Deleted Egress Only Internet Gateway \"eigw-026ff0f2da2779e85\" previously attached to VPC \"vpc-0a4f3ad1bcc21e91b\"" "object"={"kind":"AWSManagedControlPlane","namespace":"default","name":"ipv6-cluster-control-plane","uid":"c921c966-7f2f-4909-9bfc-f88c9cbcd1fd","apiVersion":"controlplane.cluster.x-k8s.io/v1beta1","resourceVersion":"14876"} "reason"="SuccessfulDeleteEgressOnlyInternetGateway"
```

Dooonee.

Skarlso:
@richardcase  and @sedefsavas So, I think this is ready to be reviewed. Two things are missing yet:

- Validation
  - [ ] Can't update an existing cluster to IPv6
  - [ ] ipv6Pool needs to be provided if ipv6CidrBlock is provided for a VPC
  - [ ] Addons need to be defined if IPv6 is enabled
  - [x] We could possibly check if the machine is nitro enabled hypervisor
  - [ ] Cluster version must be 1.21 or higher
  - [ ] Addon version of CNI must be 1.10 or higher in case of IPv6
  - [ ] SecondaryCIDRBlocks is not supported / needed with IPv6 cluster
- e2e testing
  - At least one test which will cover creating an IPv6 based cluster

But considering I'll be out for a couple of days, now is a good time to have a look over the changes so I can deal with them if there are any, as I'm sure there will be. :D

test/e2e/data/e2e_eks_conf.yaml
```diff
@@ -216,6 +216,8 @@ variables:
   EXP_EKS_ADD_ROLES: "false"
   VPC_ADDON_VERSION: "v1.11.0-eksbuild.1"
   COREDNS_ADDON_VERSION: "v1.8.7-eksbuild.1"
+  KUBE_PROXY_ADDON_VERSION: "v1.8.7-eksbuild.1"
```
Skarlso, line 219: @richardcase I shouldn't be changing this one but add a new config, right? 🤔 
And do I need a fully new suit like the one I added in eks_ipv6_test.go?

templates/cluster-template-eks-ipv6.yaml
```diff
@@ -0,0 +1,79 @@
+---
```
Skarlso, line 1: @richardcase I think I added the right e2e test config. Do I need to modify eks_test.go in any way to check for anything? :D

Also, check out e2e_eks_ipv6_conf.yaml. :))

Skarlso:
Test with the new way of defining extra environment variables to aws-node:

```
k get pods -A --kubeconfig test.kubeconfig -o wide
NAMESPACE     NAME                       READY   STATUS    RESTARTS   AGE   IP                                       NODE                                         NOMINATED NODE   READINESS GATES
kube-system   aws-node-snkpv             1/1     Running   0          48s   2600:1f14:38c:7e02:9036:9b37:f8da:ae0d   ip-10-0-117-206.us-west-2.compute.internal   <none>           <none>
kube-system   coredns-657694c6f4-j7qhf   1/1     Running   0          12m   2600:1f14:38c:7e02:4da::                 ip-10-0-117-206.us-west-2.compute.internal   <none>           <none>
kube-system   coredns-657694c6f4-p9wdp   1/1     Running   0          12m   2600:1f14:38c:7e02:4da::1                ip-10-0-117-206.us-west-2.compute.internal   <none>           <none>
kube-system   kube-proxy-cmp2f           1/1     Running   0          48s   2600:1f14:38c:7e02:9036:9b37:f8da:ae0d   ip-10-0-117-206.us-west-2.compute.internal   <none>           <none>
```

Works. :) 🎉

pkg/cloud/services/awsnode/cni.go
```diff
@@ -90,6 +90,7 @@ func (s *Service) ReconcileCNI(ctx context.Context) error {
 
 	if s.scope.SecondaryCidrBlock() == nil {
 		if len(envVars) > 0 {
+			s.scope.Info("adding environment properties to vpc-cni", "cluster-name", s.scope.Name(), "cluster-namespace", s.scope.Namespace())
```
Skarlso, line 93: So this happens continuously. I'm thinking about changing the loop to detect if there are no changes to be applied and then not do an update to the daemonset. What do you think @richardcase?

Skarlso: hmm, something is fishy here. I knew something was off in that loop. :D 

It keeps appending environment properties.

```yaml
 Environment:
      ADDITIONAL_ENI_TAGS:                    {}
      AWS_VPC_CNI_NODE_PORT_SUPPORT:          true
      AWS_VPC_ENI_MTU:                        9001
      AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER:     false
      AWS_VPC_K8S_CNI_EXTERNALSNAT:           false
      AWS_VPC_K8S_CNI_LOGLEVEL:               DEBUG
      AWS_VPC_K8S_CNI_LOG_FILE:               /host/var/log/aws-routed-eni/ipamd.log
      AWS_VPC_K8S_CNI_RANDOMIZESNAT:          prng
      AWS_VPC_K8S_CNI_VETHPREFIX:             eni
      AWS_VPC_K8S_PLUGIN_LOG_FILE:            /var/log/aws-routed-eni/plugin.log
      AWS_VPC_K8S_PLUGIN_LOG_LEVEL:           DEBUG
      DISABLE_INTROSPECTION:                  false
      DISABLE_METRICS:                        false
      DISABLE_NETWORK_RESOURCE_PROVISIONING:  false
      ENABLE_IPv4:                            false
      ENABLE_IPv6:                            true
      ENABLE_POD_ENI:                         false
      ENABLE_PREFIX_DELEGATION:               true
      MY_NODE_NAME:                            (v1:spec.nodeName)
      WARM_ENI_TARGET:                        1
      WARM_PREFIX_TARGET:                     1
      ENABLE_IPV6:                            true
      ENABLE_IPV4:                            false
      ENABLE_IPV4:                            false
      ENABLE_IPV4:                            false
```

Skarlso: Eh, the deletion is off because it doesn't reset the index and keeps looping on a slice which doesn't have the correct index any longer since an item has been removed. So it keeps adding the wrong properties.

Skarlso: Hmm, this is weird. It does do a `break` once it deleted stuff, but it is still off.

api/v1beta1/network_types.go
```diff
@@ -174,10 +174,28 @@ type VPCSpec struct {
 	// Defaults to 10.0.0.0/16.
 	CidrBlock string `json:"cidrBlock,omitempty"`
 
+	// EnableIPv6 requests an Amazon-provided IPv6 CIDR block with a /56 prefix length for
+	// the VPC. You cannot specify the range of IP addresses, or the size of the
+	// CIDR block.
+	// +optional
+	EnableIPv6 bool `json:"enableIPv6"`
```
richardcase, line 181: I'm wondering if we group  `EnableIPv6`, `IPv6CidrBlock`, `IPv6Pool` within an IPv6 specific struct that we use here?

So you end with something like this (just an illustration though):

```yaml
ipv6:
  enable: true
  cidrBlock: "XXXXXX"
  pool: "XXXXX"
```

docs/book/src/topics/eks/ipv6-enabled-cluster.md
```diff
@@ -0,0 +1,77 @@
...
+  network:
+    vpc:
+      enableIPv6: true
+  region: "${AWS_REGION}"
+  sshKeyName: "${AWS_SSH_KEY_NAME}"
+  version: "${KUBERNETES_VERSION}"
+  addons:
+    - name: "vpc-cni"
+      version: "v1.11.0-eksbuild.1"
+      conflictResolution: "overwrite" # this is important, otherwise environment property update will not work
```
richardcase, line 72: Seeing this has got me thinking. Do you think CAPA should default to `overwite`?

pkg/cloud/services/network/egress_only_gateways_test.go
```diff
@@ -0,0 +1,266 @@
+/*
+Copyright 2018 The Kubernetes Authors.
```
richardcase, line 2: nit
```suggestion
Copyright 2022 The Kubernetes Authors.
```

test/e2e/data/e2e_eks_ipv6_conf.yaml
```diff
@@ -0,0 +1,231 @@
+---
```
richardcase, line 1: I think we could probably use the existing EKS e2e conf file instead of creating a new one.  If we have a new conf file then we need to invoke the e2e tests separately.....which sometimes needs to be done for experimental features.

templates/cluster-template-eks-ipv6.yaml
richardcase: We probably need to either:

- Add a new test case to `eks_test.go`
- Add a new test file (i.e. `eks_ipv6_test.go`)

Either way, we would write a new test  that uses a IPV6 template. Lets chat via slack/zoom

richardcase (commented):
Fantastic work @Skarlso :+1: :tada: 

I had an initial quick pass through. But i will take a closer look this afternoon or tomorrow.

api/v1beta1/network_types.go
Skarlso: Good point! I think that would be cool. :)

docs/book/src/topics/eks/ipv6-enabled-cluster.md
Skarlso: I'm honestly not sure when you WOULDN'T want to set overwrite? So I think the answer is yes. But that might be a surprise for existing environments maybe? I don't know. :)

api/v1beta1/network_types.go
Skarlso: Uh that's quite a lot of moving around things. :D But it makes sense to group them. :)

Skarlso: Hmmmm... It makes some code a bit unwildely. Like, for example, take the Enable thing. If it's not defined then it defaults to `false`, right? Well if we group them, than the group will be a pointer of course. Which makes enable obsolete because we have to do a nil check for the group in any case. So the enable check then becomes a check if the group is not nil which can look something like this:

```go
if r.Spec.NetworkSpec.VPC.IPv6 != nil {
```

instead of

```go
if r.Spec.NetworkSpec.VPC.EnableIPv6 {
```

Which might not look good in some cases? ¯\\\_(ツ)\_/¯. What do you think @richardcase?

richardcase: Thats a good point, the fact that the `IPv6` is defined and not nil does indicate that the user wants a ipv6 cluster. Sometimes we provide a helper function on the scope to make the intent clearer. So something like:

```go
func (s *scope) IsIPV6Enabled() bool {
   return s.ControlPlane.Spec.NetworkSpec.VPC.IPv6 != nil
}
```

If we did want an explicit enable then you can always use the kubebuilder default comment. But as you point out its not needed if we group the options.

Skarlso: Good point! :) Will do that.

Skarlso: Ahh, we need the enable switch because you can't do this:
```
    vpc:
      ipv6:
```

Since the rest of the fields under ipv6 are all optionals. :smile: :smile: :smile: So I need something that the user puts in. But I think most of the times this will look like this:
```
    vpc:
      ipv6:
        enable: true
```

Because I don't think people use their own IPv6 pool that often. But I guess it's fine. :slightly_smiling_face: It looks a bit weird instead of:

```yaml
   vpc:
     enableIPv6: true
```

I'll run a poll :D

Skarlso: Actually, it can be something like:
```yaml
vpc:
  ipv6: {}
```
Which is fine, that's how GitHub actions work as well.

Skarlso: Done :)

test/e2e/data/e2e_eks_ipv6_conf.yaml
Skarlso: Let's talk about it when you get back. :)

richardcase: Discussed today

api/v1alpha3/awscluster_conversion.go
```diff
@@ -55,6 +57,29 @@ func (r *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
...
+		for k := range restored.Spec.NetworkSpec.Subnets {
+			if dst.Spec.NetworkSpec.Subnets[i].ID == restored.Spec.NetworkSpec.Subnets[k].ID {
+				dst.Spec.NetworkSpec.Subnets[i].IsIPv6 = restored.Spec.NetworkSpec.Subnets[i].IsIPv6
+				dst.Spec.NetworkSpec.Subnets[i].IPv6CidrBlock = restored.Spec.NetworkSpec.Subnets[i].IPv6CidrBlock
+				found = true
+				break
+			}
+		}
+		if !found {
+			return fmt.Errorf("subnet with id %s not found amongts restored subnets", dst.Spec.NetworkSpec.Subnets[i].ID)
```
pydctw, line 80: did you mean `amongst`?

api/v1beta1/network_types.go
```diff
@@ -222,13 +246,21 @@ type SubnetSpec struct {
...
 	// AvailabilityZone defines the availability zone to use for this subnet in the cluster's region.
 	AvailabilityZone string `json:"availabilityZone,omitempty"`
 
 	// IsPublic defines the subnet as a public subnet. A subnet is public when it is associated with a route table that has a route to an internet gateway.
 	// +optional
 	IsPublic bool `json:"isPublic"`
 
+	// IsIPv6 defines the subnet as an IPv6 subnet. A subnet is IPv6 when it is associated with a VPC that has IPv6 enabled.
+	// +optional
+	IsIPv6 bool `json:"isIpv6"`
```
pydctw, line 262: As this is an optional field
```suggestion
	IsIPv6 bool `json:"isIpv6"`
	IsIPv6 bool `json:"isIpv6,omitempty"`
```

Skarlso:
Wooot!!!!!!!!!!! Integration test passed for IPv6:

```
Ran 1 of 3 Specs in 1488.316 seconds
SUCCESS! -- 1 Passed | 0 Failed | 0 Pending | 2 Skipped
```

Skarlso:
- From office hours:
  - Add unmanaged test
  - Finish PlanUML diagram
    - Add a flow diagram of what happens when we request ipv6 vs BYOIPv6

templates/cluster-template-eks-ipv6.yaml
```diff
@@ -0,0 +1,87 @@
...
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: Cluster
+metadata:
+  name: "${CLUSTER_NAME}"
+spec:
+  clusterNetwork:
+    pods:
+      cidrBlocks: ["192.168.0.0/16"]
```
sedefsavas, line 9: Doesn't this need to be an IPv6 CIDR?

Skarlso: Good question!! That's actually being ignored. :D But good catch, I will document this behaviour. In case of ipv6 you aren't allowed to set pod cidrs. AWS doesn't allow it.

sedefsavas: You mean EKS?

Skarlso: In case of ipv6 we have to override the network config and pass in ipv6 family. This is done here:

```go
	var netConfig *eks.KubernetesNetworkConfigRequest
	if s.scope.VPC().IsIPv6Enabled() {
		netConfig = &eks.KubernetesNetworkConfigRequest{
			IpFamily: aws.String(eks.IpFamilyIpv6),
		}
	} else {
		netConfig, err = makeKubernetesNetworkConfig(s.scope.ServiceCidrs())
		if err != nil {
			return nil, errors.Wrap(err, "couldn't create Kubernetes network config for cluster")
		}
	}

```

During createCluster. This could be done a bit more friendly, like only removing any pod cidrs but leaving anything else that someone might have set up.

Skarlso: Found it!

```
If you specify ipv6, Kubernetes assigns Service
	// and Pod addresses from the unique local address range (fc00::/7). You can't
	// specify a custom IPv6 CIDR block.
```

I will add a validation where I'll disallow setting this field if ipv6 is enabled.

Thanks for reminding me about this @sedefsavas  :)) 👍

sedefsavas: ACK, thanks for the explanation.

We will probably need a little more complicated template for the unmanaged side, need some kubeadm settings for it as well.
Like CAPZ has here:
https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/2f36cf2ecbcb6bad3827bc5b9f8421a86f7efa21/templates/cluster-template-ipv6.yaml#L96

Skarlso: Ah, interesting. Thanks for pointing me to that. 👍

Skarlso: I think we discussed with @richardcase to support EKS clusters first, then unmanaged I think? But then some work went into doing bootstrapping and node handling that technically made it work for unmanaged as well with the cidr splitting. Why am I saying this? Because I think there might be other landmindes maybe for unmanaged clusters that I haven't even started looking at?

randomvariable:
@christianang would be useful to have extra eyes on this one.

docs/book/src/topics/eks/ipv6-enabled-cluster.md
```diff
@@ -0,0 +1,81 @@
+
+# IPv6 Enabled Cluster
+
+CAPA supports IPv6 enabled clusters. Dual stack clusters are not yet supported, but
+dual VPC, meaning both ipv6 and ipv4 are defined, is support and in fact, it's the
```
christianang, line 5:
```suggestion
dual VPC, meaning both ipv6 and ipv4 are defined, is supported and in fact, it's the
```

docs/book/src/topics/eks/ipv6-enabled-cluster.md
```diff
@@ -0,0 +1,81 @@
...
+
+```yaml
+kind: AWSManagedControlPlane
+apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+metadata:
+  name: "${CLUSTER_NAME}-control-plane"
+spec:
+  network:
+    vpc:
+      ipv6: {}
```
christianang, line 29: Given you say the vpc has to be dualstack should there be some sort of ipv4 block here?

Skarlso: You can define your ipv4 cidr as usual. They should work. This is dual stack in the sense that the vpc has both routes and not in the sense of kubernetes dual stack. :) So any setting that you add should work normally for IPv4.

christianang: Yeah I think that makes sense, but I was thinking more in terms of if defining ipv4 is required or not.

Since the vpc needs to be dualstack I was thinking it makes sense to see:
```
vpc:
  ipv6: {}
  ipv4: {}
```

which can be differentiated from what I imagine an ipv6-only vpc would be (which I understand isn't going to be supported at the moment):
```
vpc:
   ipv6: {}
```

But this is me just wondering if it is important to make that distinction explicitly, perhaps it doesn't matter too much. If it doesn't happy to just consider this thread resolved.

Skarlso: Yeah, I see what you're saying. However, IPv4 is set for you anyways internally. I think making it explicitly present in the VPC setting makes for poor readability. For ipv6 only this looks a lot better to me:

```yaml
vpc:
   ipv6:
   ipv6Only: true  
```

Which aligns more with the AWS settings anyways.

This is my opinion though. :)

christianang: Cool, happy to defer to what you think makes sense in the end.

Skarlso: It's a good question! :) I would like to ask the others what they think too. Some of it is behind the scenes. I'm not sure if it should be surfaced. But I defer to consensus too in the end. :)

christianang:
I'll add that I'm +1 on this PR. Can't give input into specifics related to AWS, but in general what I've seen in the changeset makes sense to me.

Skarlso:
Thank you! Right now, I'm working on getting unmanaged to work. There is a tiny bit of coding left on the security groups.

It's super challenging to test :D And it LOOKS like we are okay but need some additional hardship to get unmanaged working.

Honestly, adding an e2e for unmanaged would probably take another month. :D

Skarlso:
Great. These aren't failing locally. And I ran `make clean && make generate`.

Skarlso:
Recent modifications are still working with EKS and unmanaged:

```
➜  clusters clusterctl describe cluster managed-ipv6
NAME                                                                READY  SEVERITY  REASON  SINCE  MESSAGE
Cluster/managed-ipv6                                                True                     7m50s
├─ControlPlane - AWSManagedControlPlane/managed-ipv6-control-plane  True                     7m50s
└─Workers
  └─MachineDeployment/managed-ipv6-md-0                             True                     80s
    └─Machine/managed-ipv6-md-0-555b578464-qpfzs                    True                     3m50s
```

```
clusters k get pods -A -o wide --kubeconfig ipv6.kubeconfig
NAMESPACE     NAME                       READY   STATUS    RESTARTS        AGE     IP                                       NODE                                           NOMINATED NODE   READINESS GATES
kube-system   aws-node-knsrl             1/1     Running   1 (2m40s ago)   4m30s   2a05:d014:14a:5402:be25:fd0b:741e:3411   ip-10-0-124-86.eu-central-1.compute.internal   <none>           <none>
kube-system   coredns-78666889b9-pkwn7   1/1     Running   0               13m     2a05:d014:14a:5402:dcaa::                ip-10-0-124-86.eu-central-1.compute.internal   <none>           <none>
kube-system   coredns-78666889b9-tw8wk   1/1     Running   0               13m     2a05:d014:14a:5402:dcaa::1               ip-10-0-124-86.eu-central-1.compute.internal   <none>           <none>
kube-system   kube-proxy-s46xm           1/1     Running   0               4m30s   2a05:d014:14a:5402:be25:fd0b:741e:3411   ip-10-0-124-86.eu-central-1.compute.internal   <none>           <none>
```

Skarlso:
Yay, the original e2e-eks is passing too. I didn't mess up anything existing hopefully. :) 🤞

Skarlso:
I just successfully removed all my f*cking changes regarding the security groups and unmanaged clusters while doing the rebase. :( 😩

Skarlso:
> I just successfully removed all my f*cking changes regarding the security groups and unmanaged clusters while doing the rebase. :( 😩

Oh, thank all the GODS. I still had the changes on my old laptop...

And remember kids, always, *ALWAYS*, _*ALWAYS*_ do a git pull before pushing... :D

templates/cluster-template-ipv6.yaml
```diff
@@ -0,0 +1,121 @@
+---
+apiVersion: cluster.x-k8s.io/v1beta1
```
sedefsavas, line 2: Q: Which test is using this template?
I don't see it added to e2e_conf.yaml for unmanaged tests.

Skarlso: There are no unmanaged e2e tests for ipv6 because you can't make it IPv6 without manual mocking about.

Unless, you would want to implement installing a CNI, gathering the CIDR, passing that to the CNI config, updating the configmap, helm install calico, then terminate a pod and see if it comes back with ipv6. :D

I think that's a bit much. WDYT?

sedefsavas: You already tested it locally and things are working, do we have steps to make it work listed here in the doc?

> terminate a pod and see if it comes back with ipv6. :D

 Can't we start with IPv6? by passing a pre-defined CIDR to the CNI resource from the start?
Which configmap do we need to update?

Skarlso: >Can't we start with IPv6? by passing a pre-defined CIDR to the CNI resource from the start?

By the time we get to this point the cluster is already there with a couple of nodes and pods running. So like, kube-proxy and such will already exist.

I have not found a way to install CNI without pods already existing. If CNI is not there, then the nodes can't join the cluster. They will keep being pending. I haven't found a way to install CNI first, with it controlling all pods even, kube-proxy pods. Maybe there is a way with CAPA to do that?

Also, another problem is that we don't know beforehand which subnet they node will be in. But it seems to work with the VPC defined CIDR anyways for now. With the restriction that calico, for example, only supports length of 112 and above for CIDR passed to pods. Which is probably still plenty, but still...

Skarlso: BYOIP is possible in two ways. You either have one allocated by your ISP or your company, or, you create an AWS VPC with the setting `ipFamily` set to ipv6. In which case, AWS will give you an ipv6 CIDR.

If you have an existing CIDR you can provide that through the following settings:

```yaml
  clusterNetwork:
    pods:
      cidrBlocks: ["192.168.0.0/16", "2022:ff00::/56"]
    services:
      cidrBlocks: ["192.168.0.0/16", "2022:ff00::/56"]
...
spec:
  vpc:
    ipv6:
      ipv6CidrBlock: "2022:ff00::/56"
      ipv6PoolId: "pool-id"
```

Now that I see we don't need ipv6 prefix on the cidrblock anymore. :) heh.

If you have an existing network infrastructure with VPC and subnet configuration correctly laid out, you need to define the VPC ID and the subnets as you would normally when you bring your own infrastructure.

In that case, the CIDR is the VPC CIDR and you don't need to provide ipv6 details since CAPA will reconcile those.

I don't know how to do the rest for an unmanaged cluster. Like, CAPA doesn't install a CNI out of the box, right? If it does, how can I configure that?

sedefsavas: With ClusterResourceSet, once a single controlplane node comes up for the workload cluster, CNI is installed automatically.

sedefsavas: Even without existing network infra, with passing fd00:: a private CIDR, can this work>

Skarlso: Oooh! Interesting! How does that work? Which CNI provider does it install?

Skarlso: >Even without existing network infra, with passing fd00:: a private CIDR, can this work?

If the CNI can assign the right IP family and CIDR, then I suppose yes.

sedefsavas: Okay, I will test this and see with your branch.

ClusterResourceSet is not CNI specific, it just applies whatever objects user create as secret or configmap:

sedefsavas: https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20200220-cluster-resource-set.md

Skarlso: Ahha, interesting. So, the calico configMap needs the following settings:

```
--cluster-pool-ipv6-cidr="2a05:d014:852:f02::/112"
--cluster-pool-ipv6-cidr-size="128"
```

But the problem here is that `2a05:d014:852:f02::/112` is specific to a subnet. So maybe try the VPC CIDR block but change the 56 to 112. And that should then work. Perhaps... I'm not sure and the calico documentation is attrocious. :(

randomvariable:
> Why do we need to get the IPv6 CIDR, can't we use a pre-defined private CIDR as service/pod CIDRs?

In IPv6, the public/private distinction more or less goes away, there's just link-local addresses (starting with fe80) that every device gets and then to it is advertised routable IPv6 addresses (usually starting with 2002). All pods then share a prefix from the assigned CIDR.

There is a specific fc00 private network address space as well, but it's not that common afaik.

Skarlso:
>In IPv6, the public/private distinction more or less goes away, there's just link-local addresses (starting with fe80) that every device gets and then to it is advertised routable IPv6 addresses (usually starting with 2002). All pods then share a prefix from the assigned CIDR.

Yep, and using EKS, this is assigned automatically.

sedefsavas:
> > Why do we need to get the IPv6 CIDR, can't we use a pre-defined private CIDR as service/pod CIDRs?
> 
> In IPv6, the public/private distinction more or less goes away, there's just link-local addresses (starting with fe80) that every device gets and then to it is advertised routable IPv6 addresses (usually starting with 2002). All pods then share a prefix from the assigned CIDR.
> 
> There is a specific fc00 private network address space as well, but it's not that common afaik.

I was thinking something like ULA blocks, intention is not for it to be globally routable, so that we can set a random CIDR range without doing BYOI. But I guess this is not an option, need to have globally routable CIDR.

Looked at VPC creation, we can use AWS ipam pool, once CIDR is assigned, we can customize the CNI with that CIDR and apply to the cluster.

sedefsavas:
Looks like ULA is not a possiblity in AWS:
```
Amazon VPC doesn't support [unique local address](https://en.wikipedia.org/wiki/Unique_local_address) (ULA) CIDRs. All VPCs must have unique IPv6 CIDR. Two VPCs can’t have the same IPv6 CIDR range.
```
https://docs.aws.amazon.com/whitepapers/latest/ipv6-on-aws/planning-ipv6-adoption-in-the-aws-cloud-network.html

sedefsavas:
> Sure, yes. But IPAM pools for ipv6 require you to have a CIDR range already from somewhere, right?
right, it should be AWS-assigned IPv6 VPC then.
Trying to avoid BYOI to write the test easier and to provide a workaround for users that don't want to setup their networking.

Skarlso:
Hmm, I think I was working with an outdated calico installation before. Now, I have a much more refined one and I also found some documentation around setting up ipv6 finally here: https://projectcalico.docs.tigera.io/networking/ip-autodetection#change-the-autodetection-method

I'll try applying a calico resource with ClusterResourceSet and the latest calico manifest with some ipv6-related environment variables set up. Although I'm not 100% what they are yet. I have to do some more digging.

Skarlso:
Healthy cluster:
```
clusterctl describe cluster test-ipv6-unmanaged-2
NAME                                                                      READY  SEVERITY  REASON  SINCE  MESSAGE
Cluster/test-ipv6-unmanaged-2                                             True                     3m57s
├─ClusterInfrastructure - AWSCluster/test-ipv6-unmanaged-2                True                     6m9s
├─ControlPlane - KubeadmControlPlane/test-ipv6-unmanaged-2-control-plane  True                     3m57s
│ └─Machine/test-ipv6-unmanaged-2-control-plane-9hglf                     True                     5m49s
└─Workers
  └─MachineDeployment/test-ipv6-unmanaged-2-md-0                          True                     2m9s
    └─Machine/test-ipv6-unmanaged-2-md-0-9fd846f6d-dg28c                  True                     3m21s

```

And... drum rolls...

```
cluster k get pods -A --kubeconfig test.kubeconfig -o wide
NAMESPACE     NAME                                                                   READY   STATUS             RESTARTS        AGE     IP                                        NODE                                           NOMINATED NODE   READINESS GATES
kube-system   calico-kube-controllers-58f755f869-ndxwx                               0/1     CrashLoopBackOff   5 (2m41s ago)   6m41s   fd00:6541:c656:af57:1916:386e:7cfd:9a01   ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   calico-node-jm9s2                                                      1/1     Running            0               5m36s   10.0.82.174                               ip-10-0-82-174.eu-central-1.compute.internal   <none>           <none>
kube-system   calico-node-rthw2                                                      1/1     Running            0               6m42s   10.0.173.36                               ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   coredns-78fcd69978-5t54r                                               0/1     Running            0               6m57s   fd00:6541:c656:af57:1916:386e:7cfd:9a00   ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   coredns-78fcd69978-qfcj8                                               0/1     Running            0               6m57s   fd00:6541:c656:af57:1916:386e:7cfd:9a02   ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   etcd-ip-10-0-173-36.eu-central-1.compute.internal                      1/1     Running            0               6m51s   10.0.173.36                               ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   kube-apiserver-ip-10-0-173-36.eu-central-1.compute.internal            1/1     Running            0               6m51s   10.0.173.36                               ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   kube-controller-manager-ip-10-0-173-36.eu-central-1.compute.internal   1/1     Running            0               6m51s   10.0.173.36                               ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   kube-proxy-5zzq4                                                       1/1     Running            0               5m36s   10.0.82.174                               ip-10-0-82-174.eu-central-1.compute.internal   <none>           <none>
kube-system   kube-proxy-mqscf                                                       1/1     Running            0               6m56s   10.0.173.36                               ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
kube-system   kube-scheduler-ip-10-0-173-36.eu-central-1.compute.internal            1/1     Running            0               6m51s   10.0.173.36                               ip-10-0-173-36.eu-central-1.compute.internal   <none>           <none>
```

ALMOST :D

Skarlso:
Okay, so this IS working, but I'm still missing some kind of fine-tuning for calico. Because most of the nodes don't get ipv6 assigned on startup. Especially not the calico-node pods. I think I've set up all the environment properties correctly. I'll try setting ipv6 CIDR as well ( the doc said it can auto-detect that. but maybe not ).

Skarlso:
Oh!!!!!!!!!!!!! Calico is running in Dual Stack mode I think. And assigned ipv6 addresses as well.

Skarlso:
Although... all new pods will have ipv6 addresses and communicate correctly. @sedefsavas I can also read this in the calico guide:

>New pods will get IPv6 addresses, and can communicate with each other and the outside world over IPv6.

It DOES say that `New` pods will get ipv6 and not existing ones. And even though I used ClusterResourceSet, kube-proxy and the likes were already there with ipv4 addresses assigned. Did you have the time to do the testing you wanted by any chance? :) It's completely okay if not. Just wondering if you found something I didn't yet.

Skarlso:
So I gave this another spin and I ALMOST made it work, however, I'm facing a problem with unmanaged networking.

The bootstrap cloud-init function fails with the following:

```
Invalid value: "2a05:d014:5f3:6400::/64": specified service subnet is too large; for 128-bit addresses, the mask must be >= 108
```

If I specify anything smaller than the assigned subnet:

```
podSubnet: Invalid value: "2a05:d014:5f3:6400::/112": the size of pod subnet with mask 112 is smaller than the size of node subnet with mask 64
```

So... I have a chicken and egg problem. I need to figure something out without decreasing the size of the subnet.

Skarlso:
Okay, I think I traced this back to the following issue. It looks like the AWS bootstrap script for nodes needs to pass along some additional flags to the kubelet which can specify the cluster CIDR size.

I opened an issue with them to see what's going on. @richardcase or @sedefsavas. Do you know if there is a custom bootstrap script for nodes? I think there isn't right?

Alternatively, I suggest dropping unmanaged cluster support for now. In theory I believe it works and I'm not sure/don't think that there is any code-change left from CAPA's side. 🤔 Maybe after the issue is resolved we'll know more.

Until then I think we should issue a warning that unmanaged clusters will not work because of X. WDYT?

Skarlso:
I compared an EKS output and kubeconfig with that of an unmanaged output. And I can see that the node-ip passed to the kubelet is, of course, the full IP from the IPv6 pool. Which is possible since AWS already knows the assigned IP for the node so it's easy to set up. For unmanaged, our node didn't have this luxury. So when it starts the internal IP will not be ipv6 it will be ipv4.

I managed to get this working, but I had to literally hardcode the node-ip in order for it to properly pass along the IP to the node and kube-proxy. Which is not something you should/could do normally.

I remain that unmanaged should not be supported at this time. Though technically works, it requires a lot of manual interaction normally, unless you automate the cluster creation and the network creation.

Skarlso:
@sedefsavas @richardcase I updated the guide on how to install calico using a ClusterResourceSet and also updated the names for ipv6CidrBlock since it's already under ipv6 I just call it cidrBlock. And the how-to doc including Calico and manual steps. :)

Skarlso:
@sedefsavas 👋 

I think I added everything that we discussed during office hours.

- update to make it clear that it's only EKS
- add validation to prevent AWSCluster from using ipv6

Some things regarding dual stacks are already in the proposal.

If I left something out please don't hesitate to bring it in front. :)

sedefsavas (commented):
I will keep adding some comments, but they can be targeted in a follow up PR.
LGTM in general. We can focus on merging this asap to avoid any more merge conflicts.

cmd/clusterawsadm/cloudformation/bootstrap/cloud_provider_integration_node.go
```diff
@@ -43,8 +43,13 @@ func (t Template) cloudProviderNodeAwsPolicy() *iamv1.PolicyDocument {
 				Effect:   iamv1.EffectAllow,
 				Resource: iamv1.Resources{iamv1.Any},
 				Action: iamv1.Actions{
+					"ec2:AssignIpv6Addresses",
```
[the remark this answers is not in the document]
Skarlso: Agreed! 👍

api/v1alpha3/awscluster_conversion.go
```diff
@@ -55,6 +56,37 @@ func (r *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 	}
 
 	dst.Spec.S3Bucket = restored.Spec.S3Bucket
+	if restored.Spec.NetworkSpec.VPC.IPv6 != nil {
```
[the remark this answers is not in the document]
Skarlso: Ah, good spot. :)

api/v1beta1/awscluster_webhook.go
```diff
@@ -178,3 +179,11 @@ func (r *AWSCluster) Default() {
 func (r *AWSCluster) validateSSHKeyName() field.ErrorList {
 	return validateSSHKeyName(r.Spec.SSHKeyName)
 }
+
+func (r *AWSCluster) validateNetwork() field.ErrorList {
+	var allErrs field.ErrorList
+	if r.Spec.NetworkSpec.VPC.IsIPv6Enabled() {
```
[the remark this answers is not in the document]
Skarlso: Ah, hmmm, good question. My thinking here was, why would anyone want to set the ipv6 subnet field though if they can't set the CIDR in the first place? But it's still a good check to add. :)

Skarlso:
@sedefsavas Thanks, these are excellent suggestions, I'll apply them. :)

api/v1alpha4/awscluster_conversion.go
```diff
@@ -46,6 +49,37 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 	}
 
 	dst.Spec.S3Bucket = restored.Spec.S3Bucket
+	if restored.Spec.NetworkSpec.VPC.IPv6 != nil {
```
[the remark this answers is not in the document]
Skarlso: Err... if I don't do the conversion the conversion fuzzy test fails that it can be set in the other thing but not in this thing. How do I fix that? Do I just nil this field, or... ?

cmd/clusterawsadm/cloudformation/bootstrap/cloud_provider_integration_node.go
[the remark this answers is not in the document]
Skarlso: This does require some additional changes to the `AWSIAMConfigurationSpec` and then changes throughout the code to ensure that the new flag is set correctly. Do I see that right?

api/v1alpha4/awscluster_conversion.go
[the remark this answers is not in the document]
richardcase: As the NetworkSpec struct is used by `AWSCluster` & `AWSManagedControlPlane` the fuzzing tests will try and populate it. 

There are 2 ways we handle this for `AWSCluster`

- support roundtrip conversation with `AWSCluster` and restore the values...even though the validating webhooks won't allow setting ipv6.
- Add a fuzzer function for populating `NetworkSpec` when its used by `AWSCluster` so it always nils the IPv6 part

I'd say go with the first option so that when we do support IPv6 with AWSCluster we don't have to go and remove fuzzing functions....plus the first option is already written.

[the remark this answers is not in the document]
Skarlso: Yeah, that's largely what I understood as well. Also, the last point is a good point. We DO plan on supporting this in AWSCluster after I figure out why the heck the nodes aren't getting the right IP ( Oh did I mention that I figured out why unmanaged isn't working? It's because the nodes have ipv4 private addresses instead of ipv6 for some reason... ). :))

api/v1beta1/network_types.go
```diff
@@ -222,13 +246,21 @@ type SubnetSpec struct {
 	// CidrBlock is the CIDR block to be used when the provider creates a managed VPC.
 	CidrBlock string `json:"cidrBlock,omitempty"`
 
+	// IPv6CidrBlock is the IPv6 CIDR block to be used when the provider creates a managed VPC.
+	// A subnet can have an IPv4 and an IPv6 address.
```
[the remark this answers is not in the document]
Skarlso: Added comment to all fields and updated the AWSCluster webhook.

api/v1beta1/awscluster_webhook.go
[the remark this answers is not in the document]
Skarlso: Added check for the subnet fields too.

cmd/clusterawsadm/cloudformation/bootstrap/cloud_provider_integration_node.go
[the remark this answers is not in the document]
Skarlso: Added `clusterawsadm` flag and updated examples and the e2e test to use it. Hopefully correctly... :D

sedefsavas:
/test pull-cluster-api-provider-aws-e2e-clusterclass

/lgtm
/hold
 until tests pass

Skarlso:
/unhold

All passed 🎉 🎉 🎉

docs/book/src/topics/using-clusterawsadm-to-fulfill-prerequisites.md
```diff
@@ -106,7 +106,23 @@ spec:
   ...
 ```
 
+#### With EKS Support
```
richardcase, line 109:
```suggestion
#### With IPv6 Support
```

docs/book/src/topics/using-clusterawsadm-to-fulfill-prerequisites.md
```diff
@@ -106,7 +106,23 @@ spec:
...
+
+The pre-requisities for IPv6.... are enabled by default. However, if you want to use some of the optional features of EKS (see [here](eks/enabling.md) for more information on what these are) then you will need to enable these features via the configuration file. For example:
```
richardcase, line 111:
```suggestion
The pre-requisities for IPv6 are not enabled by default. If you want to create IPv6 enabled clusters using CAPA then you will need to enable these permissions via the configuration file. For example:
```

cmd/clusterawsadm/api/bootstrap/v1beta1/types.go
```diff
@@ -224,6 +231,10 @@ type AWSIAMConfigurationSpec struct {
 	// TODO: This field could be a pointer, but it seems it breaks setting default values?
 	// +optional
 	S3Buckets S3Buckets `json:"s3Buckets,omitempty"`
+
+	// IPv6 when enabled, will add IPv6 specific permissions.
+	// +optional
+	IPv6 *IPv6 `json:"ipv6,omitempty"`
```
richardcase, line 237: I'm wondering if we should enable IPv6 by default. If we do, then this turns into a way to disable the permissions instead of enabled them. wdyt?

docs/book/src/topics/using-clusterawsadm-to-fulfill-prerequisites.md
richardcase: And depending on a previous comment this may need to be changed.

cmd/clusterawsadm/api/bootstrap/v1beta1/types.go
richardcase: If it's not enabled by default, when do we change it to be the default in the future?

templates/cluster-template-ipv6.yaml
```diff
@@ -0,0 +1,126 @@
+---
```
richardcase, line 1: We should probably remove this template for the time being as we aren't supporting AWSCluster yet

test/e2e/shared/template.go
```diff
@@ -120,6 +120,9 @@ func newBootstrapTemplate(e2eCtx *E2EContext) *cfn_bootstrap.Template {
 	t.Spec.EKS.DefaultControlPlaneRole.Disable = false
 	t.Spec.EKS.ManagedMachinePool.Disable = false
 	t.Spec.S3Buckets.Enable = true
+	t.Spec.IPv6 = &bootstrapv1.IPv6{
```
richardcase, line 123: As we are enabling it by default in the e2e tests then perhaps we should enabling it by default in capa?

cmd/clusterawsadm/api/bootstrap/v1beta1/types.go
Skarlso: Enabling ipv6 by default is a non-goal, FWIW. It's experimental at best, right? People will need to keep testing it as I'm sure, some things will not work. :)

Skarlso: >when do we change it to be the default in the future?

Maybe half a year from now... ?

richardcase: If we think its experimental then we should also add a feature flag to fgo along with the opt-in permissions?

Skarlso: Isn't the fact that you explicitly have to set `vpc: ipv6{}` a kind of feature flag already? ( I don't know, just genuinely asking :) ).

docs/book/src/topics/using-clusterawsadm-to-fulfill-prerequisites.md
Skarlso: @Skarlso Update date this once discussions are concluded.

richardcase:
All tests passing so:

/lgtm
/approve

PR #3518, opened 2022-06-09
feat: external load balancer garbage collection

### LINKED ISSUE #1718: AWS NLB linger after they're orphaned
/kind bug

Workload clusters are able to create Network Load Balancers instead of Classic ELBs, you just add an annotation `service.beta.kubernetes.io/aws-load-balancer-type: nlb` ([read more](https://kubernetes.io/docs/concepts/services-networking/service/#aws-nlb-support)), we need to identify those Load Balancers by tag and delete them as well when the associated workload cluster is destroyed.

Referenced [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/d150541075065c52aea4640eaafc5c47aecdbcad/pkg/cloud/services/elb/loadbalancer.go#L128-L135) in a conversation on the topic.

NLB are deleted alongside the cluster.

**Environment:**

- Cluster-api-provider-aws version: v0.5.2

richardcase:
**What type of PR is this?**

/kind feature

This change includes a new experimental feature for garbage collecting external resources that have been created by workloads running in a workload/tenent cluster. This is specifically for the scenario where a cluster is deleted without deleting everything that runs within in.

Currently, if you create in the tenant cluster a service of type load balancer (with the NLB annotation) and delete the cluster without deleting the service the existence of the NLB will stop the cluster deleting successfully as the VPC cannot be deleted as its being used by the NLB.

This change works in 2 phases:

1. When you create a cluster:
  - The `AWSCluster` or `AWSManagedControlPlane` have a gc finalizer added to them which indicates the garbage collector should apply to this cluster. This can be disabled by adding an annotation
2. When you delete a cluster:
  - Normal deletion occurs
  - For the controllers of `AWSCluster` or `AWSManagedControlPlane`  in their `reconcileDelete` the garbage collector is called if the gc finalizer exists.

Fixes #1718 

> NOTE: this has a separate target for the e2e tests that cover this functionality. When we are happy with the functionality we can move it into the main tests.

A couple of questions / ideas:
- [x] Should we add a command to `clusterawsadm` to manually add/delete the GC annotation
- [x] If the feature is enabled all clusters at garbage collected, do we need a follow up that we can selectively GC clusters using a label select for clusters.
  - YES - this will be done as a follow-up 

**Release Note**

To use this feature you will need to enable it via the the feature flag (see the docs). Also, you will need to run clusterawsadm again to update the stack with a missing permission.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests


sedefsavas:
We might need to delete `deleteAWSCloudProviderELBs` method (depending on the order), it may cause Service deletion to fail.

richardcase:
> We might need to delete `deleteAWSCloudProviderELBs` method (depending on the order), it may cause Service deletion to fail.

Thats a good point. We may need to skip `deleteAWSCloudProviderELBs` if the feature is enabled...so we don't change the current behaviour. wdyt?

sedefsavas:
Sounds good to me, and in v1beta2 we might consider completely cleaning it, so that we don't provide a partial solution when the flag is not enabled.

test/e2e/suites/managed/externalresourcegc_test.go
```diff
@@ -0,0 +1,159 @@
...
+// EKS cluster external resource gc tests.
+var _ = ginkgo.Describe("[managed] [gc] EKS Cluster external resource GC tests", func() {
+	var (
+		namespace   *corev1.Namespace
+		ctx         context.Context
+		specName    = "eks-extresgc"
+		clusterName string
+	)
+
+	shared.ConditionalIt(runGCTests, "[managed] [gc] should cleanup a cluster that has ELB/NLB load balancers", func() {
```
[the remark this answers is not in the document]
richardcase: Yes...i was going to create a separate Prow job to run this as it requires enabling the feature flag. I had initially done this using a seperate e2e config file. But thinking about it i can just set the variable instead.

Thanks for making me thing @pydctw 👍 

I will change this......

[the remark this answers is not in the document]
pydctw: I was wondering about how/where you want to run the test. Thanks for the explanation and setting a variable makes sense to me.

richardcase:
I see what the error is here with the e2e. I need to go back to using a separate e2e config file for this test. This also means i will create a separate pre-submit job.

richardcase:
I think the verify failure was Cloudflare issues. Lets try again:

/test pull-cluster-api-provider-aws-verify

richardcase:
EKS e2e tests are passing

docs/book/src/topics/external-resource-gc.md
```diff
@@ -0,0 +1,29 @@
+# External Resource Garbage Collection
```
Ankitasw, line 1: It may be worth adding this doc link to https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure.html , wdyt?

richardcase: Good thinking....will change this :+1:

dlipovetsky:
Users expect this clean up, and everyone does it out of band, so this is a great improvement!

Sometimes, I need to clean up a workload cluster whose control plane is unavailable. In this case, the GC cannot list LoadBalancer-type Services on the workload cluster. I want to allow the deletion of the workload cluster, and so I need to disable the GC cleanup. Of course, this deletion will fail if there are ELB's for LoadBalancer-type Services, so it's my responsibility to discover and clean the ELBs up before allowing the deletion.

>  Should we add a command to clusterawsadm to manually add/delete the GC annotation

+1, because this allows the deletion of the workload cluster whose control plane is unavailable. Let's also document the annotation, so that users can use their own clients to remove the annotation.

richardcase:
Thanks @dlipovetsky for the comments. I have committed an additional change that includes:

- New `clusterawsadm` commands to enable/disable gc for a specific cluster
- Updates the gc docs with instructions on how to manually enable/disable gc.

docs/book/src/topics/external-resource-gc.md
richardcase: Added

richardcase:
I think we can ignore the apidiff error. @sedefsavas wdyt?

richardcase:
e2e tests are still passing after the recent change.

dlipovetsky:
I realize it is a bit late to question the design, but I observe that:
1. If the `aws.cluster.x-k8s.io/external-resource-gc` annotation exists and its value is `true`, GC is disabled. If the annotation does not exist, or its value is `false`, then GC is enabled. I think this annotation is an "opt out" mechanism, but that is not clear from the annotation name.
2. The same annotation that's used to enable/disable GC is also used to mark whether the GC is done with its work. That is, the value `true` means both "disabled" and "GC work is done for this object." Is it not worth the additional complexity to use a finalizer to mark the work, and an annotation to enable/disable the GC? I would prefer a finalizer, because its has a precise meaning: some controller needs to do work before the object can be deleted. I'm sorry if this was already discussed elsewhere, but I missed it.

richardcase:
@dlipovetsky - its never too late to question the design :smile: I appreciate the feedback and i did actually think about 2 separate annotations, 1 to mark a cluster as requiring gc and 1 to mark if it has been cleaned up. But decided for a simpler option.

But the fact that you picked up on this makes be realize the simpler option isn't ideal, and my doubts have been confirmed. I like your idea of using a finalizer :+1: 

I will make the changes now...

richardcase:
Whilst i do some more testing:

/hold

exp/controllers/externalresourcegc_controller.go
```diff
@@ -0,0 +1,190 @@
...
+		return reconcile.Result{}, nil
+	}
+
+	shouldGC, err := extScope.ShouldGarbageCollect()
+	if err != nil {
+		return reconcile.Result{}, fmt.Errorf("determining if cluster needs garbage collecting: %w", err)
+	}
+
+	if shouldGC {
+		extScope.Info("Deleting from workload/tenant cluster")
```
dlipovetsky, line 162:
```suggestion
		extScope.Info("Starting garbage collection of cluster resources")
```

exp/controllers/externalresourcegc_controller.go
```diff
@@ -0,0 +1,190 @@
...
+		res, err := wkSvc.ReconcileDelete(ctx)
+		if err != nil {
+			extScope.Error(err, "error deleting remote resources")
+			return reconcile.Result{}, fmt.Errorf("deleting remote resources: %w", err)
+		}
+		if !res.IsZero() {
+			return res, nil
+		}
+	} else {
+		extScope.Info("Infra cluster has GC finalizer but not deleting from workload/tenant cluster as annotation says don't garbage collect")
```
dlipovetsky, line 174:
```suggestion
		extScope.Info("Skipping garbage collection of cluster resources: infra cluster has garbage collection disabled")
```

dlipovetsky:
I observed that we use both the expanded form `garbage collect{ed,ion}` as well as the abbreviated form `gc` in user-facing error and log messages.

Using one form everywhere would make it easier for users to find related error and log messages.

Not a big deal, and something we can fix in a follow up PR.

richardcase:
> Thank you for considering, and then implementing the finalizer approach @richardcase. I find the mental model much easier to understand now: The garbage collection controller always applies its finalizer, and always removes it, but it decides whether to perform garbage collection based on the annotation.

I agree, having the 2 different indicators is a better mental model. I'm not sure that we need the finalizer as we could use 2 annotations but happy to keep the finalizer+annotation approach.

richardcase:
Removing the hold as i am testing more as part of the refactor

/unhold

richardcase:
e2e tests are passing.

richardcase:
@sedefsavas - i have created #3584 to enhance the e2e tests to cover ALBs. I thought it may be better to do this as a follow-up as this PR is already pretty big.

richardcase:
The build/test failures are weird as they run locally. Any ideas @sedefsavas ?

richardcase:
Ah, rebase from main and now i hit the issue:

image

That file was added as part of #3374 where the e2e passed.

richardcase:
Whilst i look at splitting this work into different PRs:

/hold

richardcase:
Closing this in favour of 4 smaller PRs:

/close

PR #3519, opened 2022-06-10
Make CAPA tooling work on arm64

### LINKED ISSUE #2930: Make CAPA tooling work on arm64
We should remove some hard-coded `amd64` to allow dev tooling to work on Darwin/arm64 following cluster-api.

See kubernetes-sigs/cluster-api#5547

Prajyot-Parab:
Signed-off-by: Prajyot-Parab <prajyot.parab2@ibm.com>

**What type of PR is this?**
/kind feature

- Make CAPA tooling work on arm64

Fixes #2930 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -93,7 +93,7 @@ BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD)
 TOOLCHAIN_IMAGE := toolchain
 
 TAG ?= dev
-ARCH ?= amd64
+ARCH ?= $(shell go env GOARCH)
 ALL_ARCH ?= amd64 arm arm64 ppc64le s390x
 
 # main controller
@@ -356,7 +356,7 @@ managers: ## Alias for manager-aws-infrastructure
 
 .PHONY: manager-aws-infrastructure
 manager-aws-infrastructure: ## Build manager binary
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
+	CGO_ENABLED=0 GOOS=linux GOARCH=${ARCH} go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
 
 ##@ test:
 
--- a/hack/ensure-kind.sh
+++ b/hack/ensure-kind.sh
@@ -20,18 +20,20 @@ set -o pipefail
 
 GOPATH_BIN="$(go env GOPATH)/bin/"
 MINIMUM_KIND_VERSION=v0.12.0
+GOARCH="$(go env GOARCH)"
+GOOS="$(go env GOOS)"
 
 # Ensure the kind tool exists and is a viable version, or installs it
 verify_kind_version() {
 
   # If kind is not available on the path, get it
   if ! [ -x "$(command -v kind)" ]; then
-    if [[ "${OSTYPE}" == "linux-gnu" ]]; then
+    if [ "$GOOS" == "linux" ] || [ "$GOOS" == "darwin" ]; then
       echo 'kind not found, installing'
       if ! [ -d "${GOPATH_BIN}" ]; then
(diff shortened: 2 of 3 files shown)


Ankitasw:
Thanks for the PR @Prajyot-Parab 🎉 
Was it possible for you to test it in darwin/arm64 machine? Unfortunately I don't have one otherwise would have tested.

Prajyot-Parab:
> Thanks for the PR @Prajyot-Parab 🎉
> Was it possible for you to test it in darwin/arm64 machine? Unfortunately I don't have one otherwise would have tested.

@Ankitasw I don't have access to a `darwin/arm64` machine at the moment, will update if I get hold of one by EOD today.

Ankitasw:
Thank you for verifying @Prajyot-Parab.
/lgtm
cc @sedefsavas @richardcase for approval.

Makefile
```diff
@@ -361,7 +361,7 @@ managers: ## Alias for manager-aws-infrastructure
 
 .PHONY: manager-aws-infrastructure
 manager-aws-infrastructure: ## Build manager binary
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
+	go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
```
richardcase, line 364: We have lost the `CGO_ENABLED=0` (i.e. disabling cgo). Was this intended?

hack/ensure-kind.sh
```diff
@@ -20,18 +20,20 @@ set -o pipefail
 
 GOPATH_BIN="$(go env GOPATH)/bin/"
 MINIMUM_KIND_VERSION=v0.12.0
+goarch="$(go env GOARCH)"
```
richardcase, line 23: nit: existing variables are all upper case

Prajyot-Parab: Fixed.

Makefile
Prajyot-Parab: Fixed.

Makefile
```diff
@@ -356,7 +356,7 @@ managers: ## Alias for manager-aws-infrastructure
 
 .PHONY: manager-aws-infrastructure
 manager-aws-infrastructure: ## Build manager binary
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
+	CGO_ENABLED=0 go build -ldflags "${LDFLAGS} -extldflags '-static'" -o $(BIN_DIR)/manager .
```
richardcase, line 359: Should we also be adding this:

```
GOOS=linux GOARCH=${ARCH}
```

Or something similar. Do we class the controller manager binary dev tooling? Probably not.

@sedefsavas @pydctw - wdyt?

Prajyot-Parab:
```
OS=$(go env GOOS)
GOOS=${OS} GOARCH=${ARCH}
```
@richardcase something like this should also work I believe

richardcase: That would work 👍 

(the reason in my example for hardcoding linux (`GOOS=linux`) is that we only support running the manager on linux so should we allow the manager to be built for windows or mac?)

Prajyot-Parab: @richardcase In that case it would make sense to have it as `GOOS=linux GOARCH=${ARCH}` since the support is only for linux.

richardcase: That was my thought, but for `clusterawsadm` its ok to compile for windows/mac/linux etc

Prajyot-Parab: Done.

Prajyot-Parab:
@Ankitasw I think we are good to move forward with this one, let me know incase of any issues.

richardcase:
/approve

@Ankitasw - will leave the lgtm to you if you are happy

PR #3524, opened 2022-06-13
Allow disabling vpc cni when setting secondary cidr block

mtougeron:
**What type of PR is this?**

/kind feature

cilium has an eni mode that requires a secondary cidr block on the vpc but does not use the vpc cni.

I looked for e2e tests but I didn't see anything specific related to this functionality. If I missed it please point me in the right direction and I'd be happy to update the code.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
+++ b/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
@@ -158,8 +158,7 @@ spec:
                   installed into the cluster. For clusters where you want to use an
                   alternate CNI this option provides a way to specify that the Amazon
                   VPC CNI should be deleted. You cannot set this to true if you are
-                  using the Amazon VPC CNI addon or if you have specified a secondary
-                  CIDR block.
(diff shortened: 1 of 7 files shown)


sedefsavas:
I don't see any regression issues with this PR for unamanaged clusters. LGTM, will defer to @richardcase.

mtougeron:
ah my bad, didn't regenerate the CRDs. Doing that now.

mtougeron:
Out of curiosity, once this is approved/merged, what's the release cycle like? Is it tied to cluster-api's or is it released independently? thx.

richardcase:
> Out of curiosity, once this is approved/merged, what's the release cycle like? Is it tied to cluster-api's or is it released independently? thx.

Our releases are mostly independent of cluster-api. However, there are times when we have to make sure we do a release after something in cluster-api is available.

richardcase:
Thanks for this @mtougeron . 

Also for your comment on there being no e2e tests that cover this area i have created #3544.

The existing e2e tests are passing and i see no issues with the code, so:

/approve

Issue #2355, opened 2021-04-16
v1beta2 API Changes


richardcase:
/kind refactor
/kind api-change
/area provider/eks
/priority important-soon
/milestone v0.7.0

**Describe the solution you'd like**
Refactor the API with the proposed changes for v1beta2/v1 including any field deprecations:

Changes:

- [ ] EKS changes (old doc: https://docs.google.com/document/d/1EeD2KjFfeKoD-o0LJTMMWh9GUy-15yG80S5Uoj7oKcA/edit)
- [ ] Remove `Equals` from `Tags`
- [ ] ARN field in AWSResourceReference.
- [ ] `AWSManagedMachinePool` field deprecations as a result of #3094 

**Environment:**

- Cluster-api-provider-aws version: 0.6.5

richardcase:
As this requires a change in contract we will have to wait for the next API bump (i.e. v1alpha5)

/milestone Next
/priority important-longterm

richardcase:
Changing this to be a general v1beta2/v1 API changes issue

Ankitasw:
ARN field in `AWSResourceReference`.
(couldn't edit the PR description, not enough permissions)

pydctw:
FailureDomain in `AWSMachine`.

From [CAPI book](https://cluster-api.sigs.k8s.io/developer/providers/machine-infrastructure.html?highlight=failuredomain#data-types)
> providers will be expected to remove the field in the next version that provides breaking API changes, favoring the value defined on Machine.Spec.FailureDomain instead.

richardcase:
Refactor the `disableVPCCNI` on the managed control plane so that the disable and extra args are in a struct.

richardcase:
Make sure we don't miss: #3526

richardcase:
Change the addons override to default to overwrite: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3513#discussion_r920090162

pydctw:
Add a required field to SubnetSpec to use as listMapKey in SSA: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3528

sedefsavas:
Remove `deviceName` from `rootVolume` in AWSMachinePool CR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3624

sedefsavas:
NLB support: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3088
[...]

PR #3526, opened 2022-06-13
Allow multiple security group filter matches

dlmather:
**What type of PR is this?**
/kind feature

Current behavior for additional security groups especially when using filters is somewhat cumbersome and mysterious. If a filter has no potential matches, it will silently blackhole *all* additional security groups due to returning up an error which is eventually silently dropped. Additionally, a choice was made to only return the first match from a specified filter. It would be nice to be able to instead match all security groups returned by a filter so that tags can be used to implicitly add security groups to AWSMachines in a more dynamic way. Finally, in the current setup, it is considered an error to have a filter match that returns no results... this forces the requirement that all security groups must be created before the creation of an AWSMachineTemplate using them. This creates a back-and-forth in Cluster bootstrap, since we need the VPC to exist before defining security groups, but then must circle back to retroactively add security group matchers, which in turn requires rolling all MachineDeployments that wish to use the SecurityGroups.

Fixes #

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

### DIFF
--- a/controllers/awsmachine_controller_unit_test.go
+++ b/controllers/awsmachine_controller_unit_test.go
@@ -914,7 +914,7 @@ func TestAWSMachineReconciler(t *testing.T) {
 					g.Expect(ms.AWSMachine.Finalizers).To(ContainElement(infrav1.MachineFinalizer))
 					expectConditions(g, ms.AWSMachine, []conditionAssertion{{infrav1.SecurityGroupsReadyCondition, corev1.ConditionFalse, clusterv1.ConditionSeverityError, infrav1.SecurityGroupsFailedReason}})
 				})
-				t.Run("Should return silently if ensureSecurityGroups fails to fetch additional security groups", func(t *testing.T) {
+				t.Run("Should fail if ensureSecurityGroups fails to fetch additional security groups", func(t *testing.T) {
 					g := NewWithT(t)
 					awsMachine := getAWSMachine()
 					setup(t, g, awsMachine)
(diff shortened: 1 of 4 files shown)


sedefsavas:
Failing to reconcile when additional SGs could not be found is a change of behaviour, so we can do this change in the v1beta2 version. It'd be good to track this here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2355

Since we already support list of additional SGs in other places, makes sense to add multiple additional SGs support to launchtemplates, could you file an issue and modify this PR to only cover that?

dlmather:
> Failing to reconcile when additional SGs could not be found is a change of behaviour
Just to clarify, we would *no longer* fail to reconcile when SGs couldn't be found under this change. As you say though, this is definitely a change of behavior from the current setup where failing to find SGs in any listed filter drops all AdditionalSecurityGroups.

Ankitasw:
cc @richardcase @sedefsavas for review/approval.

Ankitasw:
I think this PR is good to go.
@sedefsavas could you PTAL?

Ankitasw:
@dlmather could you please rebase the changes.

sedefsavas:
Based on[ this comment ](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3526#issuecomment-1154556309)above, let's wait until we open the main branch to v1beta2 changes before rebasing to avoid double efforts.

Ankitasw:
@dlmather could you rebase your PR, as the main branch is now open for the API changes.

Ankitasw:
@dlmather This looks good. please squash your commits so that I can merge.

Ankitasw:
@dlmather could you squash your commits so that we can merge this?

Ankitasw:
/lgtm
/approve
/hold 
until test passes

PR #3535, opened 2022-06-16
fix: Fail apidiff make target when git fails

### LINKED ISSUE #3443: apidiff target verify silently fails and prints errors in CI
/kind bug

CI prints:
```
./hack/verify-boilerplate.sh
fatal: ambiguous argument 'origin/main': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
```

https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/3442/pull-cluster-api-provider-aws-verify/1518584914981687296#

Also, the command fails and it's not making the job fail.

CI should work nicely without errors and CI to fail when some command execution fails.

Callisto13:
This is a fairly simple fix to ensure that when `git diff` fails on the
`make apidiff` target, the exit code is actually picked up by make.
Previously the exit code from `diff` was "swallowed" by the `if`.

```
$ cat Makefile
thing:
        @if (exit 1); then \
		echo foo; \
        fi
        @echo "still here"

$ make thing
still here
$ echo $?
0
```

Because this is a straightforward `if` and we don't actually care about
the output of that diff beyond that pipe, we can drop the `if` and just
run each thing one after the other. If the `diff` or the `grep` fail
the target will fail.

In the case of `$(APIDIFF_OLD_COMMIT)`, there is some annoying `make`
magic going on here.
The `shell` is expanded when the file is loaded and when the value is
used, but again the exit code is swallowed by the subshell.
After some messing about, this is the easiest way to get it to bail:

```
$ cat Makefile
APIDIFF_OLD_COMMIT ?= $(shell git rev-parse origin/foo)
$(info $(APIDIFF_OLD_COMMIT))
ifneq ($(.SHELLSTATUS),0)
  $(error failed)
endif

thing:
        @echo "won't actually run"

$ make thing
fatal: ambiguous argument 'origin/foo': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
origin/foo
Makefile:324: *** failed. Stop.
$ echo $?
2
```

This relies on make > 4.2. The `info` call directly after the
declaration is needed because until the variable is used, `.SHELLSTATUS`
is empty.

I know there are other places where we are doing `shell git ...` so I
can update those too if desired.

Closes #3443

**What type of PR is this?**

/kind bug

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


pydctw:
> [Keywords](https://help.github.com/articles/closing-issues-using-keywords) which can automatically close issues and at(@) or hashtag(#) mentions are not allowed in commit messages.

If I interpret the message correctly, you cannot use `@` in a commit message. I see two of them in the commit message, `@if` and `@echo`.

PR #3538, opened 2022-06-17
fix: Fail apidiff make target when git fails

### LINKED ISSUE #3443: apidiff target verify silently fails and prints errors in CI

Callisto13:
_Description has been updated to summarise discussion and eventual changes._

The `apidiff` target runs 2 commands (piped) in an `if` which swallows
the exit code, meaning it never fails.

What we want:
- exit 1 when `git diff` fails
- exit 0 when `grep` fails
- call `go-apidiff` when `git diff` and `grep` succeeds
- exit 1 when `go-apidiff` fails

This is honestly a complete pain to do in a Makefile.

One option was to move bits into another `.sh` script (not the
`scripts/ci-apidiff.sh` one as that would make the target pointless).

Another option was to fork the `go-apidiff` command so that we can make
it only run on the `api/` dir, since that is all we care about. While that option
is the easiest to run, we don't want to have to maintain another
thing.

So, if we can live with a little repetition, we can do the following:
- Check the `git diff` can run, exit 1 if not
- Run the `git diff` again, this time piping the successful command to
  `grep`
- If `grep` fails, then no worries, the target will roll on and exit 0
  like it always has.
- If `grep` succeeds then the `go-apidiff` tool is called and the target
  runs as intended.

------

In the case of `$(APIDIFF_OLD_COMMIT)`, there is some annoying `make`
magic going on here. Which I can't simply make fail since it will cause any
job which uses something in this Makefile to fail out of proximity.
The `shell` is expanded when the file is loaded meaning even targets
which do not care about the value will end up erroring (but not exiting)
when it fails. These are not errors which impact the target's run, but
they look annoying in CI.

Since this var is only used by `apidiff`, we can move it into that
target so it is only evaluated when specifically called.

**What type of PR is this?**

/kind bug

Fixes #3443 

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -317,13 +317,19 @@ verify-gen: generate ## Verify generated files
 		echo "generated files are out of date, run make generate"; exit 1; \
 	fi
 
-APIDIFF_OLD_COMMIT ?= $(shell git rev-parse origin/main)
-
 .PHONY: apidiff
+apidiff: APIDIFF_OLD_COMMIT ?= $(shell git rev-parse origin/main)
 apidiff: $(GO_APIDIFF) ## Check for API differences
-	@if (git --no-pager diff --name-only FETCH_HEAD | grep "api/"); then \
+	@$(call checkdiff) > /dev/null
+	@if ($(call checkdiff) | grep "api/"); then \
 		$(GO_APIDIFF) $(APIDIFF_OLD_COMMIT) --print-compatible; \
+	else
+		@echo "No changes to 'api/'. Nothing to do."
 	fi
+
+define checkdiff
+	git --no-pager diff --name-only FETCH_HEAD
+endef
 
 ##@ build:
 


hack/tools/Makefile
```diff
@@ -66,7 +66,7 @@ $(BIN_DIR)/restart.sh:
 
 GO_APIDIFF := $(BIN_DIR)/go-apidiff
 $(GO_APIDIFF): $(BIN_DIR) go.mod go.sum
-	go build -tags=tools -o $@ github.com/joelanford/go-apidiff
+	go build -tags=tools -o $@ github.com/warehouse-13/go-apidiff
```
Ankitasw, line 69: Could you please revert this change as it's using a fork of github.com/joelanford/go-apidiff repo

richardcase: The main issue is that we don't use `internal` folders and so we get lots of errors with apidiff if we don't try to filter on just the `api` folders.

As @Callisto13 has found out the apidiff tool we use assumes that you use `internal` folders and so a form is created that allows you to filter within the tool itself.

People have tried to filter by grep in the makefile but the existing solution swallows the error.

richardcase: The other option is that we do the work to move everything to an `internal` folder that isn't designed to be consumed outside CAPA. This was discussed at the recent office hours and there was reluctance to this.

Lets start another discussion on this.....

Ankitasw: Yeah so to use go-apidiff, only best solution left is to move everything to internal, but it would be good to explore more options.
The laziest solution suggested above is one such way to go I guess 🙂 .

Ankitasw: I am slightly not in favor of forking another repo, as it will be again a maintenance task added to the bucket of maintainers/reviewers.

richardcase: @Ankitasw - i agree on the fork 👍 In the long term i think we need to move things to `internal` but if this other option suggested by @Callisto13  works short term, then great. I wonder what happens if grep returns no matches in this example......

Callisto13: > I wonder what happens if grep returns no matches in this example......

Nothing. We don't want the grep to fail, since it will fail the whole job in CI (if there are no matches it means there are no relevant diffs so the job is a noop but does not fail). Since the grep is in the if, the exit code is swallowed. The if branch will only run if grep finds something.

Callisto13: updated to used the (refactored) solution I pasted in above

Issue #2701, opened 2021-08-24
Support spot instances for ManagedMachinePool


abhinavnagaraj:
/kind feature

**Describe the solution you'd like**
`AWSManagedMachinePool` should support spot instances. The SDK provides an option to select the `capacityType` for the node group. We should add the same in CAPA. 

**Anything else you would like to add:**
As a first step we can use the existing `instanceType` field with `capacityType=SPOT`. But the complete implementation should allow selection of multiple instance types.

PR #3365, opened 2022-03-29
docs: proposal for using luanch templates with machine pools

richardcase:
**What type of PR is this?**

/kind design

**What this PR does / why we need it**:

This proposal is for adding support for laucnh templates to
`AWSManagedMachinePool`.

Note: this is the successor to #3024 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #3094, opened 2022-01-22
AWSManagedMachinePool - Launch Template support

richardchen331:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Add launch template support to `AWSManagedMachinePool`.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2055

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Add launch template support to `AWSManagedMachinePool`.
```

PR #3569, opened 2022-06-30
eks: allow specifying multiple instance types for AWSManagedMachinePool

jashandeep-sohi:
**What type of PR is this?**
/kind feature

Adds a new field to `AWSManagedmachinePool` called `spec.instanceTypeList` that let's you specify a list of instance types that the underlying node group should use. This is primarily useful for `spec.capacityType: spot`, where specifying a set of instances increases your odds of finding an adequate instance, but there's nothing stopping you from using it with `spec.capacityType: onDemand`. More here https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html#managed-node-group-capacity-types

For backwards compatibility (and to avoid confusion), `spec.instanceType` is left as-is, but it's made mutually exclusive with `spec.instanceTypeList` so only one can be set.

#2701 had this as a request, but it was never added.

**Special notes for your reviewer**: I've named it `instanceTypeList` instead of `instanceTypes` to make it visually different from `instanceType` to avoid any confusion, but I'm not too strongly attached to that if you want to change it.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

### DIFF
--- a/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmanagedmachinepools.yaml
+++ b/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmanagedmachinepools.yaml
@@ -574,6 +574,13 @@ spec:
               instanceType:
                 description: InstanceType specifies the AWS instance type
                 type: string
+              instanceTypeList:
+                description: InstanceTypeList specifies a list of AWS instance types
+                  this pool can choose from. Mutually exclusive with instanceType.
+                items:
+                  type: string
+                maxItems: 20
+                type: array
               labels:
                 additionalProperties:
                   type: string
--- a/docs/book/src/topics/spot-instances.md
+++ b/docs/book/src/topics/spot-instances.md
@@ -47,6 +47,26 @@ spec:
   ...
 ```
 
+Multiple instance types can be provided via `instanceTypeList` to increase the
+number of Spot capacity pools available for allocating capacity from:
+```yaml
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+kind: AWSManagedMachinePool
+metadata:
+  name: ${CLUSTER_NAME}-pool-0
+spec:
+  capacityType: spot
+  instanceTypeList:
+    - "t3.small"
+    - "t3a.small"
+    - "t3.medium"
+    - "t3a.medium"
+    - "t2.medium"
+    - "m1.large"
+  ...
+```
+
+
 See [AWS doc](https://docs.aws.amazon.com/eks/latest/userguide/managed-node-groups.html) for more details.
 
 > **IMPORTANT NOTE**: The experimental feature `AWSMachinePool` does not support using spot instances as of now.
--- a/exp/api/v1alpha3/zz_generated.conversion.go
+++ b/exp/api/v1alpha3/zz_generated.conversion.go
@@ -783,6 +783,7 @@ func autoConvert_v1beta1_AWSManagedMachinePoolSpec_To_v1alpha3_AWSManagedMachine
 	// WARNING: in.Taints requires manual conversion: does not exist in peer-type
 	out.DiskSize = (*int32)(unsafe.Pointer(in.DiskSize))
 	out.InstanceType = (*string)(unsafe.Pointer(in.InstanceType))
+	// WARNING: in.InstanceTypeList requires manual conversion: does not exist in peer-type
 	out.Scaling = (*ManagedMachinePoolScaling)(unsafe.Pointer(in.Scaling))
 	out.RemoteAccess = (*ManagedRemoteAccess)(unsafe.Pointer(in.RemoteAccess))
 	out.ProviderIDList = *(*[]string)(unsafe.Pointer(&in.ProviderIDList))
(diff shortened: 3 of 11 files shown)


pydctw:
Hi @jashandeep-sohi, thanks for the PR. Agree that it will be good to support specifying multiple instance types for AWSMMP.

My initial thoughts
* How about changing the API in the next revision, [v1beta2](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2355), making `InstanceType` to be an array with a conversion webhook implemented? Having both `InstanceType` and `InstanceTypeList` can be confusing to a user and I think in the long run, we want to have only one field for specifying an instance type instead of having two fields and making them mutually exclusive with a webhook.
* There is a [proposal](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3365/files#diff-d6d8565c707995a7204010e1acade42238594c2bcf88d13a6ed7bd316369942eR117) to deprecate `InstanceType` field in favor of using a LaunchTemplate. @richardcase, thoughts on how the launch template implementation will impact the field?

jashandeep-sohi:
> * How about changing the API in the next revision, [v1beta2](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2355), making `InstanceType` to be an array with a conversion webhook implemented? Having both `InstanceType` and `InstanceTypeList` can be confusing to a user and I think in the long run, we want to have only one field for specifying an instance type instead of having two fields and making them mutually exclusive with a webhook.

Sounds good to me.

> * There is a [proposal](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3365/files#diff-d6d8565c707995a7204010e1acade42238594c2bcf88d13a6ed7bd316369942eR117) to deprecate InstanceType field in favor of using a LaunchTemplate. @richardcase, thoughts on how the launch template implementation will impact the field?

I see. This is definitely a better direction to go, however as far as I can tell it doesn't cover multiple instance types. Looks like in  `AWSMachinePool` that's done via `spec.mixedInstancesPolicy.overrides[].instanceType`. Perhaps something similar could be added to `AWSManagedMachinePool` as well to keep the APIs similar.

I just needed a quick and easy way to showcase that this controller could handle managed EKS nodes w/ multiple spot instance types in a prototype, so I was looking for the smallest diff.

richardcase:
This would be good to discuss at the office hours on Monday, wdyt?

pydctw:
> This would be good to discuss at the office hours on Monday, wdyt?

Sounds great. @jashandeep-sohi, there is a CAPA office hr on Monday 9am Pacific. We will discuss this issue and capture the notes (in case you cannot make it).

pydctw:
> I see. This is definitely a better direction to go, however as far as I can tell it doesn't cover multiple instance types. Looks like in `AWSMachinePool` that's done via `spec.mixedInstancesPolicy.overrides[].instanceType`. Perhaps something similar could be added to `AWSManagedMachinePool` as well to keep the APIs similar.
> 
> I just needed a quick and easy way to showcase that this controller could handle managed EKS nodes w/ multiple spot instance types in a prototype, so I was looking for the smallest diff.

@jashandeep-sohi, we had discussion during today's office hr. There are two options depending on timeline.

1) Wait for [AWSManagedMachinePool - Launch Template support #3094](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3094) to be merged. Add sth similar as `spec.mixedInstancesPolicy.overrides[].instanceType` as you suggested.
2) Merge this PR as is but deprecate the field when Launch Template PR is merged.

The first option is preferred as adding a field and deprecating soon after is not ideal but it may take time to merge launch template PR. 

Is the capability of supporting multiple instance types a blocker/urgent for your team? Based on your previous comment that it can wait for v1beta2 API, I didn't think so but want to confirm.

jashandeep-sohi:
> > I see. This is definitely a better direction to go, however as far as I can tell it doesn't cover multiple instance types. Looks like in `AWSMachinePool` that's done via `spec.mixedInstancesPolicy.overrides[].instanceType`. Perhaps something similar could be added to `AWSManagedMachinePool` as well to keep the APIs similar.
> > I just needed a quick and easy way to showcase that this controller could handle managed EKS nodes w/ multiple spot instance types in a prototype, so I was looking for the smallest diff.
> 
> @jashandeep-sohi, we had discussion during today's office hr. There are two options depending on timeline.
> 
>     1. Wait for [AWSManagedMachinePool - Launch Template support #3094](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3094) to be merged. Add sth similar as `spec.mixedInstancesPolicy.overrides[].instanceType` as you suggested.
> 
>     2. Merge this PR as is but deprecate the field when Launch Template PR is merged.
> 
> 
> The first option is preferred as adding a field and deprecating soon after is not ideal but it may take time to merge launch template PR.
> 
> Is the capability of supporting multiple instance types a blocker/urgent for your team? Based on your previous comment that it can wait for v1beta2 API, I didn't think so but want to confirm.

Thanks for the update! Yes, that's right, this is not exactly a blocker for us atm since we can just carry on using this patch internally. But I wouldn't want to do that for too long either. So it is a priority, but I agree that incorporating it via option 1 would be better in the long run.

/close

Issue #3414, opened 2022-04-13
Graduation of EventBridge Feature


sedefsavas:
This issue is to track graduation of the Eventbridge feature, which currently only watches instance state change events. 

There are other type of CloudWatch events that can be watched and acted upon such as ASG Scale in events and Spot Instance Termination Notices. 

Some use cases for these:

- In the current AWSMachinePool implementation, we don't have draining logic and with the use of lifecyclehooks, we can act on the instance termination events and drain the instances during rollout or scale-in. (see [graceful shutdown of MachinePools ](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3256)for details)
- In the near future we are planning to add support for[ spot instances with AWSMachinePool](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2523) and termination notices could give us enough time to drain spot instances before they shut down (notices give about 2 minutes in some scenarios)

In the scope of this effort, the existing implementation should be modified to accommodate different types of events and should be flexible enough to be used by several controllers. Also, for cases where events need to be exposed to the controllers (this will be the case probably for AWSMachinePool draining), some helper functions should make read/delete interaction with events possible.

richardcase:
@Ankitasw - shall we look at the geduation in the new year?

Ankitasw:
Yes, i started it a while ago, but got busy in focussing on E2E, my focus after E2E things would be this.

fiunchinho:
hey @Ankitasw , is there a way to track the progress on this? We are very interested on https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2574 , but it seems to be blocked by this issue. Thank you!

Ankitasw:
This is in our radar, would try to get this prioritized soon.

fiunchinho:
hey @Ankitasw , just checking in :slightly_smiling_face: . Any progress to share?

Issue #2794, opened 2021-09-24
AWSCluster can trigger re-enqueue of all machines when ingress rules fail to update, hitting AWS API limits


randomvariable:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

In a scenario where CAPA is trying to reconcile security groups and fails, this triggers all machines to be requeued, and then causes API rate limiting against the EC2 API.

**What did you expect to happen:**

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

We deliberately re-enqueue all machines when a AWSCluster is updated in unpaused state in:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/v0.6.4/controllers/awsmachine_controller.go#L239-L240

We wanted to do that so that we have fast reconciliation, however, it's definitely causing the API rate limit to hit quickly. There's a simple reproduction too:

Create a cluster with CAPA, and then change the description of one of the rules, but keep the rules otherwise the same. Or do the following:

![image](https://user-images.githubusercontent.com/1441070/134682762-ea0c2737-12a2-4e17-a530-3da7c347fc56.png)

**Environment:**

- Cluster-api-provider-aws version: v0.6.4
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`): 

/priority important-soon
/area networking

randomvariable:
We probably should discuss in next office hours.

sedefsavas:
@shivi28 is there any update on this bug?

shivi28:
Hey @sedefsavas, I am going to add debouncing logic and introduce a debouncing window in AWSCluster and AWSMachine reconcilers. Will raise a PR in this week

sedefsavas:
This issue related to:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1764

richardcase:
From office hours 2023-04-03:

- Do we care about re-reconciling if we don't hit api rate limits?
- We have exponential backoffs
- What about caching?
- Also related is #4192 

/triage accepted
/priority important-soon

Issue #1764, opened 2020-06-23
CAPA consuming a good portion of AWS API limits


sethp-nr:
When we recently shipped a tool that collected monitoring data from the same APIs, we noticed that our monitoring data had lots of holes in it from rate limiting. Via cloudtrail, we saw that CAPA was making ~80% of the API calls during one of these gaps, and ~90% of the calls being made in our account were `DescribeInstances` and `DescribeNetworkInterfaces`.

It looks like the current implementation of the reconciler operates on a single entity at a time – running at least one Describe or mutate call per AWS* object. This is an inefficient use of the API limits relative to batching these calls, so I'm here to propose that we introduce this notion to the various Service interfaces, starting with `DescribeNetworkInterfaces` and `DescribeInstances`.

An example implementation might look something like:

```
type ENIResponse struct {
	interfaces []*ec2.NetworkInterface
	err        error
}

type ENIRequest struct {
	instanceID string
	out        chan ENIResponse
}

func (s *Service) getInstanceENIs(instanceID string) ([]*ec2.NetworkInterface, error) {
	ch := make(chan ENIResponse)
	s.eniRequests <- ENIRequest{
		instanceID: instanceID,
		out:        ch,
	}
	ret := <-ch
	return ret.interfaces, ret.err
}

func (s *Service) runGetENIs() {
	for {
		items, timeout := []*string{}, time.After(1*time.Second)
		resp := map[string][]chan ENIResponse{}

		func() {
			for {
				select {
				case item := <-s.eniRequests:
					items = append(items, aws.String(item.instanceID))
					resp[item.instanceID] = append(resp[item.instanceID], item.out)
					if len(items) > 1 {
						return
					}
				case <-timeout:
					return
				}
			}
		}()

		if len(items) < 1 {
			continue
		}

		input := &ec2.DescribeNetworkInterfacesInput{
			Filters: []*ec2.Filter{
				{
					Name:   aws.String("attachment.instance-id"),
					Values: items,
				},
			},
		}

		output, err := s.scope.EC2.DescribeNetworkInterfaces(input)
		if err != nil {
			for _, chs := range resp {
				for _, ch := range chs {
					ch <- ENIResponse{err: err}
				}
			}
			continue
		}

		ifaces := make(map[string][]*ec2.NetworkInterface, len(items))
		for _, iface := range output.NetworkInterfaces {
			if iface == nil {
				continue
			}

			id := *iface.Attachment.AttachmentId
[...]

PR #3583, opened 2022-07-11
docs: ADR for graduating EventBridge in CAPA

Ankitasw:
**What type of PR is this?**
/kind documentation

This PR adds an ADR proposal to graduate EventBridge feature in CAPA, since there are design decisions required for the same. Once ADR is accepted, the actual work would start for this feature.

Fixes Part of #3414 

**Checklist**:

- [x] squashed commits
- [X] includes documentation
- [ ] adds unit tests


docs/adr/0005-Graduation of EventBridge.md
```diff
@@ -0,0 +1,59 @@
...
+* `EventBridgeEvent` would be populated based on the unmarshalled message received above.
+* While processing the message in `processMessage()`, AWSInstanceState controller processes below listed events based on the Source captured in above struct:
+   * Check EC2 state change events from `aws.ec2` source.
+   * Check ASG lifecycle hook events from `aws.autoscaling` source.
+   * Check spot instance termination notice events from `aws.ec2` source.
+   * Check scheduled events from `aws.health` source.
+* Based on the type of event, AWSInstanceState controller updates the label ec2-instance-state with the correct instance state.
+* Below helper functions in pkg/cloud/services/instancestate would be used by CAPA controllers:
+   * `GetEventFromSQS()`: This fetches the `EventBridgeEvent` details so as to act upon that event in the respective controller based on the use-case. As soon as the label `ec2-instance-state` is patched on the `AWSMachine`(we would need a watcher for this functionality), this func would be called by the CAPA controller to get the event details.
```
sedefsavas, line 58: We need to think about how to propagate this info to AWSMachinePool controller too. Since ASG instances do not have a matching AWSMachine as of now, there should be additional logic there too.

Ankitasw: Added.

docs/adr/0005-graduation of EventBridge.md
```diff
@@ -0,0 +1,57 @@
...
+### Current Design
+EventBridge looks for EC2 state change events for all the _AWSMachine_ based on `InstanceID` found in the SQS queue, and loads the state to `ec2-instance-state` label in that _AWSMachine_ in `processMessage()` func.
+
+Currently, EventBridge handles messages only from source `aws.ec2`.
+
+### Proposed Design
+* We would define `EventBridgeEvent` struct in `awsinstancestate_controller.go` to capture EventBridge event details:
+
+```go
+type EventBridgeEvent struct {
```
sedefsavas, line 33: Is there a `type` field associated with the events to show the source? 
I am assuming `detail-type` is not showing the source as aws.ec2 or aws.autoscaling.

AustinSiu: The `source` field will contain the context of what service sent the event, i.e., `aws.ec2` or `.aws.autoscaling`. 
The contents of `detail-type` can vary by service and type of event. A reference of events from AWS services can be found [here](https://docs.aws.amazon.com/eventbridge/latest/userguide/eb-service-event.html). For example, here's an EC2 Instance-terminate lifecycle event from EventBridge (from [EC2 Auto Scaling ref](https://docs.aws.amazon.com/autoscaling/ec2/userguide/ec2-auto-scaling-event-reference.html)):

```
{
  "version": "0",
  "id": "12345678-1234-1234-1234-123456789012",
  "detail-type": "EC2 Instance-terminate Lifecycle Action",
  "source": "aws.autoscaling",
  "account": "123456789012",
  "time": "yyyy-mm-ddThh:mm:ssZ",
  "region": "us-west-2",
  "resources": [
    "auto-scaling-group-arn"
  ],
  "detail": { 
    "LifecycleActionToken":"87654321-4321-4321-4321-210987654321", 
    "AutoScalingGroupName":"my-asg", 
    "LifecycleHookName":"my-lifecycle-hook", 
    "EC2InstanceId":"i-1234567890abcdef0", 
    "LifecycleTransition":"autoscaling:EC2_INSTANCE_TERMINATING", 
    "NotificationMetadata":"additional-info"
  } 
}
```

Ankitasw: @sedefsavas This detail is captured in line 49.

dlipovetsky:
_Slightly off-topic_

It looks like EventBridge provides CAPA controllers a way to "watch" any AWS resource. in the same way that the controllers already "watch" any Kubernetes API resource.

If that's right, then it looks like the Reconcile loop of each controller could be triggered by either a Kubernetes API or an AWS event.

A side effect would be fewer AWS API `Describe` calls, and in turn, a possible solution to CAPA's API rate limiting problem, e.g. #2794 and #1764.

docs/adr/0005-graduation of EventBridge.md
```diff
@@ -0,0 +1,57 @@
...
+* Date: TBD 
+* Authors: @Ankitasw
+* Deciders: @richardcase @sedefsavas
+
+## Context
+EventBridge is an eventbus provided by AWS to watch for the events generated by any SaaS application running on AWS, or any of the AWS services, that could be used by any other services.
+
+With respect to CAPA, EventBridge helps watch EC2 instance state change events using which _AWSMachine_ controller would trigger reconciliation and handle the lifecycle of the EC2 instances.
+
+There are other `CloudWatch` events that exists which can be watched upon and CAPA controllers can take necessary actions based on the events, for eg. ASG Scale-in events, Spot Instance Termination Notices, Scheduled Maintenance events etc.
```
richardcase, line 13:
```suggestion
In the future, there are other `CloudWatch` events that CAPA controllers might want to take action on based on the event (e.g.  ASG Scale-in events, Spot Instance Termination Notices, Scheduled Maintenance events).
```

docs/adr/0005-graduation of EventBridge.md
```diff
@@ -0,0 +1,57 @@
...
+
+* Status: [proposed]
+* Date: TBD 
+* Authors: @Ankitasw
+* Deciders: @richardcase @sedefsavas
+
+## Context
+EventBridge is an eventbus provided by AWS to watch for the events generated by any SaaS application running on AWS, or any of the AWS services, that could be used by any other services.
+
+With respect to CAPA, EventBridge helps watch EC2 instance state change events using which _AWSMachine_ controller would trigger reconciliation and handle the lifecycle of the EC2 instances.
```
richardcase, line 11:
```suggestion
With respect to CAPA, the  _AWSMachine_ controller uses EC2 instance state change events via EventBridge to  trigger reconciliation so that it can handle the lifecycle of the EC2 instances.
```

docs/adr/0005-graduation of EventBridge.md
```diff
@@ -0,0 +1,57 @@
...
+
+There are other `CloudWatch` events that exists which can be watched upon and CAPA controllers can take necessary actions based on the events, for eg. ASG Scale-in events, Spot Instance Termination Notices, Scheduled Maintenance events etc.
+
+## Decision
```
richardcase, line 15: An ADR is used to record a lightweight decision and the consequences of this decision. I'm struggling to understand what is the decision here from the description.

For me the decision is "We will graduate the EventBridge support out of experimental so that it is GA available and enabled by default in CAPA because we want to be able to use the different event types in the future."

docs/adr/0005-graduation of EventBridge.md
```diff
@@ -0,0 +1,57 @@
...
+This document describes the ways on how to accommodate watching different types of events and how controllers could utilize these events. The scenarios that would need modification in EventBridge in CAPA (in `AWSInstanceState` controller) are as below:
+* In the current _AWSMachinePool_ implementation, we don't have node draining logic. With the use of lifecycle hooks, CAPA controllers need to watch `ASG scale-in` events and act on them such that the EC2 instances shuts down gracefully.
+* We also plan to support spot instances with _AWSMachinePool_ which would require CAPA controllers to watch for `Spot Instance Termination Notices(SITN)` which would give enough time to drain spot instances gracefully before they shut down.
+* Handle scheduled maintenance events where the EC2 instance needs to be restarted because of reasons like hardware/software updates.
+
+The `AWSInstanceState` controller would accommodate different types of events and would be flexible enough to be used across CAPA controllers.
+
+## Consequences
+### Current Design
```
richardcase, line 24: I don't we should be covering "current design" in the consequences section. We really only need to cover the what will change as a result of this decision....so the stuff in the "proposed design" section.

Ankitasw: Shall we add this in decision section, or create a separate section?

docs/adr/0005-graduation of EventBridge.md
Ankitasw: Makes sense to me, Will add this in decision

docs/adr/0005-graduation of EventBridge.md
richardcase: I think there is too much detail overall. I think the decision section only needs to be a few sentances max.

richardcase:
> A side effect would be fewer AWS API `Describe` calls, and in turn, a possible solution to CAPA's API rate limiting problem, e.g. #2794 and #1764.

@dlipovetsky - good thinking. we should create an issue for this to investigate how we can change our controllers/services based on this.

docs/adr/0005-graduation of EventBridge.md
Ankitasw: Since this is a bigger change, I wanted to highlight design decision as well, so if I want to add the design part, can I add a separate section in ADR only? Otherwise if I directly raise PR, it would be too huge to understand. But since this was not that bigger feature change, I thought of adding this to ADR, wdyt @richardcase ?

Ankitasw: or else i create a separate google doc and just reference it in decision section.

Skarlso:
That ( fewer describe calls ) would be amazing!

docs/adr/0005-graduation of EventBridge.md
richardcase: In that case, yes let's add the detail in the "current design" to the context.

Ankitasw: and proposed design in decision section only?

richardcase: Proposed design in the consequences. Keep the decision small.

Ankitasw: sure, thanks for the inputs @richardcase 🙂

Ankitasw:
@sedefsavas @richardcase i have addressed all commentsz could you PTAL?

richardcase:
From my side:

/lgtm

Will leave it to Sedef to approve:

/assign sedefsavas

PR #3592, opened 2022-07-13
Use ginkgo.timeout for only ci conformance tests

sedefsavas:
/kind failing-test

After https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3586 is in, non-ci conformance tests started failing:
(https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/periodic-cluster-api-provider-aws-e2e-conformance/1547086357950631936)

The reason for this is only K8s >= 1.25 is using ginkgo v2, so we need to use ginkgo v1 flags until we bump all conformance test versions to >= 1.25.

As a follow up to this PR, I will create a test-infra PR to pass the new conformance-ci.yaml to be used in CI conformance test.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


CecileRobertMichon:
> As a follow up to this PR, I will create a test-infra PR to pass the new conformance-ci.yaml to be used in CI conformance test

could you instead use the k8s version value to determine which conformance config file to use?

Issue #3410, opened 2022-04-12
Deprecate `DisableVPCCNI` in `AWSManagedControlPlane`


richardcase:
/kind feature
/area provider/eks
/priority important-longterm
/triage accepted
/help

**Describe the solution you'd like**
With #3374 we are introducing a new struct for settings that relate to the VPC CNI  (a.k.a aws-node). We should add a `Disable` field to this and start deprecation of `DisableVPCCNI`.

Any code that relies on `DisableVPCCNI` should be changed to use a field on the managed control plane scope that can internally get the values from `DisableVPCCNI` or `VPCCNI.Disable`.

We will also probably need to update the admission controller to:

- report that `DisableVPCCNI` is deprecated
- check if the values of `DisableVPCCNI` & `VPCCNI.Disable` are set to something different

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

charlie-haley:
I'm happy to take a look at this once #3374 has been merged into master

/assign @charlie-haley

charlie-haley:
@richardcase do you also think it would make sense to do this https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3544 as part of the same PR?

richardcase:
> @richardcase do you also think it would make sense to do this #3544 as part of the same PR?

I'd be tempted to do them as 2 separate PRs. Only because the e2e tests can take a while to get working. But if you want to do them as a single PR then thats good as well. 

It would be great to improve the e2e coverage on the EKS side, so if you are keen in this area and if would be useful i could do a a brain dump on developing e2e tests in CAPA.

charlie-haley:
> > @richardcase do you also think it would make sense to do this #3544 as part of the same PR?
> 
> I'd be tempted to do them as 2 separate PRs. Only because the e2e tests can take a while to get working. But if you want to do them as a single PR then thats good as well.
> 
> It would be great to improve the e2e coverage on the EKS side, so if you are keen in this area and if would be useful i could do a a brain dump on developing e2e tests in CAPA.

Sounds good, I can take a look at the e2e tests once these API changes are in, I'll reach out when I start work on them, thanks!

PR #3596, opened 2022-07-13
[WIP] feat: deprecate DisableVPCCNI in AWSManagedControlPlane

charlie-haley:
**What type of PR is this?**
/kind deprecation

This PR deprecates the `disableVPCCNI` field on `AWSManagedControlPlane` in favour of `vpcCni.disable`.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3410

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controlplane/eks/api/v1beta1/awsmanagedcontrolplane_webhook.go
+++ b/controlplane/eks/api/v1beta1/awsmanagedcontrolplane_webhook.go
@@ -95,6 +95,7 @@ func (r *AWSManagedControlPlane) ValidateCreate() error {
 	allErrs = append(allErrs, r.validateSecondaryCIDR()...)
 	allErrs = append(allErrs, r.validateEKSAddons()...)
 	allErrs = append(allErrs, r.validateDisableVPCCNI()...)
+	allErrs = append(allErrs, r.validateVpcCni()...)
 	allErrs = append(allErrs, r.validateKubeProxy()...)
 	allErrs = append(allErrs, r.Spec.AdditionalTags.Validate()...)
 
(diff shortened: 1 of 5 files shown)


controlplane/eks/api/v1beta1/awsmanagedcontrolplane_webhook.go
```diff
@@ -337,6 +338,10 @@ func (r *AWSManagedControlPlane) validateDisableVPCCNI() field.ErrorList {
 	if r.Spec.DisableVPCCNI {
 		disableField := field.NewPath("spec", "disableVPCCNI")
 
+		if r.Spec.VpcCni.Disable != r.Spec.DisableVPCCNI {
```
charlie-haley, line 341: I'm struggling to think of the best way to handle this, there's a bit of a dilemna with being able to handle both the new field and the old field.

Currently, `disableVPCCNI` is a bool and it defaults to `false`, this means if the user configures their `AWSManagedControlPlane` like below, it will always error.
```yaml
  # disableVPCCNI: false
  vpcCni:
    disable: true
```

The only solution I can see to this is have the new `disable` field completely override `disableVPCCNI` if it's been set, any thoughts on this?

Skarlso: I would make it into a pointer. From the outside, nothing would change. But internally, you can then see if it was just left out or deliberately set to false. In case it was just left out, you can continue with the VpcCni value. WDYT?

Skarlso:
Hello 👋  :)

I don't see yet a warning for the deprecated field usage. But since this is a WIP I assume you'll just add that later? :D I'm referring to this comment from Richard:

>report that DisableVPCCNI is deprecated

Cheers! :)

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_webhook.go
charlie-haley: Sorry for the slow reply @Skarlso I've been away. I also thought about using a pointer, but even if we use a pointer we still have the problem of the default value of `disableVPCCNI` being `false`

charlie-haley:
> Hello wave :)
> 
> I don't see yet a warning for the deprecated field usage. But since this is a WIP I assume you'll just add that later? :D I'm referring to this comment from Richard:
> 
> > report that DisableVPCCNI is deprecated
> 
> Cheers! :)

I've added a deprecation notice to the description field on `disableVPCCNI`, do we need it anywhere other than that?

pydctw:
I know the issue description says what you have implemented but considering CAPA will introduce v1beta2 in coming months,

How about changing the API in v1beta2 and implement a conversion webhook rather than deprecate a field, `disableVPCCNI` and introduce a new field, `vpcCni.disable` with extra logic in validation webhook? I think the former is simpler.

Thoughts, @richardcase?

richardcase:
> How about changing the API in v1beta2 and implement a conversion webhook rather than deprecate a field, `disableVPCCNI`

As v1beta2 is imminent, I think we should go this route :+1:

Ankitasw:
/close
in favor of #3846

PR #3573, opened 2022-07-05
Renaming kubernetes.io/cluster tag to use EKS cluster name

knabben:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:
This PR changes the tag kubernetes.io/cluster/ to have the same EKS cluster name as the other objects, instead of the CAPI cluster name.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3572

**Special notes for your reviewer**:
1. Adding IAMAUTH mocker with mockgen
2. Increasing the coverage of EKS module for createCluster, this test will guarantee no regressions in the tag.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #3597, opened 2022-07-14
Fix the owner checking of a cluster

Skarlso:
**What type of PR is this?**

/kind bug

During this [3573](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3573) we changed the owner tag name. That made owner lookup fail on the next run here:

```go
		tagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.Name())
```

Because the key no longer was used with `Name()` this check returned a `nil` value. This PR fixes that.

Unfortunately, I don't know the code enough to ascertain if the "fix" introduced in that PR actually has more ramifications. This is the only one I could find for now that seems to be related to that change.

This was surfaced in test runs for this PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3568 which merged these changes after which the test started to fail.

Here is a successful run with this fix though for the e2e EKS test: https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/3568/pull-cluster-api-provider-aws-e2e-eks/1547472747523739648

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/eks/cluster.go
+++ b/pkg/cloud/services/eks/cluster.go
@@ -59,7 +59,7 @@ func (s *Service) reconcileCluster(ctx context.Context) error {
 			return errors.Wrap(err, "failed to create cluster")
 		}
 	} else {
-		tagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.Name())
+		tagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.KubernetesClusterName())
 		ownedTag := cluster.Tags[tagKey]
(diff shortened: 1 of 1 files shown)


richardcase:
/approve

until the the e2e passes:
/hold

richardcase:
@Skarlso / @Ankitasw - if you can unhold when the e2e passes.

/lgtm

PR #3609, opened 2022-07-21
feat: external load balancer garbage collection (part 1) - proposal 

richardcase:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:

This is the proposal for the new garbage collection functionality. It was written post implementation to make sure we captured the intent of the change and to aid with the review.

Even though its written post implementation it may merge first as the gc work has been split into smaller PRs to aid review.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Relates #1718 

**Special notes for your reviewer**:

This part 1 of 4 of changes (i.e. a stack) to implement the garbage collection. This is work to split up the original pr #3518 

**Checklist**:

- [x] squashed commits
- [x] includes documentation

PR #3610, opened 2022-07-21
feat: external load balancer garbage collection (part 2) - new gc service

richardcase:
**What type of PR is this?**

/kind feature

This introduces a new garbage collection service which will be used later by the controllers to do clean up of AWS resources for workload clusters that were created via the CCM. Initially, we support deleting load balancers (classic elb, nlb, alb), target groups and security groups.

The service is design to be called by the infra cluster reconcilers when a cluster is delete. The entry point is **ReconcileDelete** into the service.

When a cluster is deleted the cluster controllers will call **ReconcileDelete**. The first job is to determine if the workload cluster being deleted should be garbage collected. A cluster will be garbage collected if either of these are true:

- the `aws.cluster.x-k8s.io/external-resource-gc` annotation is absent
- the `aws.cluster.x-k8s.io/external-resource-gc `annotation exists and its value is set to true

If a cluster is to be garbage collected then we need to identify the AWS resources that where created by the CCM in the workload cluster. This is done by using the AWS resource tagging api and getting resources with the Kubernetes cluster owned label (i.e. `kubernetes.io/cluster/[CLUSTERNAME]`). The resources that are returned are converted and then **the list of them is passed to the defined cleanup functions in order**. The order of the functions matter because some AWS resources cannot be delete before others. For example, target groups can only be deleted after load balancers. The cleanup function will then iterate over the resources and decide if it wants to handle that resource and delete it from AWS.

**The service will be use by a later PR to the controllers and so is initially unused.**

Relates #1718 

This part 2 of 4 of changes (i.e. a stack) to implement the garbage collection. This is work to split up the original pr https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3518

**Checklist**:

- [x] squashed commits
- [x] adds unit tests


pkg/annotations/annotations.go
```diff
@@ -0,0 +1,85 @@
...
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+const (
+	// ExternalResourceGCAnnotation is the name of an annotation that indicates if
+	// external resources should be garbage collected for the cluster.
+	ExternalResourceGCAnnotation = "aws.cluster.x-k8s.io/external-resource-gc"
+)
```
dlipovetsky, line 27: I think this the annotation must be part of our public API. For example, [this annotation](https://github.com/kubernetes-sigs/cluster-api/blob/0571cc51ca53bb568dd7b62d3c66b62498550c16/api/v1beta1/common_types.go#L44) in the Cluster API core API.

We already have an experimental part of the API, where you have added the new finalizers. Should we move the annotation there?

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,85 @@
...
+// 		return true, nil
+// 	}
+
+// 	converted, err := strconv.ParseBool(status)
+// 	if err != nil {
+// 		return false, fmt.Errorf("converting annotation value to bool %s: %w", ExternalResourceGCAnnotation, err)
+// 	}
+
+// 	return converted, nil
+// }
```
dlipovetsky, line 85: These are commented out. Did you mean to add these in a later PR?

You can merge them in this PR, if you want, but I would ask to move these functions to separate file, e.g. `gc.go`, in this package.

pkg/cloud/services/gc/cleanup.go
```diff
@@ -0,0 +1,143 @@
...
+
+	return nil
+}
+
+func (s *Service) deleteResources(ctx context.Context) error {
+	s.scope.Info("deleting aws resources created by tenant cluster")
+
+	serviceTag := infrav1.ClusterAWSCloudProviderTagKey(s.scope.KubernetesClusterName())
+	awsInput := rgapi.GetResourcesInput{
+		TagFilters: []*rgapi.TagFilter{
```
dlipovetsky, line 91:
```suggestion
		ResourceTypeFilters: nil,
		TagFilters: []*rgapi.TagFilter{
```

I'm suggesting this because making it explicit helps me understand that this selects _all_ resources.

I also think that (in a subsequent PR) it would be helpful to factor out a common implementation of this useful, high-level concept of "find resources that are owned by the cluster." We have a similar implementation in one other place: https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/4d3a8c63fab99a3aa9c029394bce08cf343b9855/pkg/cloud/services/elb/loadbalancer.go#L593-L601

pkg/cloud/services/gc/ec2.go
```diff
@@ -0,0 +1,72 @@
...
+
+func (s *Service) deleteSecurityGroup(ctx context.Context, lbARN *arn.ARN, mapping *rgapi.ResourceTagMapping) error {
+	eksClusterName := getTagValue(eksClusterNameTag, mapping)
+	if eksClusterName != "" {
+		s.scope.V(2).Info("Security group created by EKS directly, skipping deletion", "cluster_name", eksClusterName)
+
+		return nil
+	}
+
+	//TODO: should we check for the security group name start with k8s-elb-
```
dlipovetsky, line 58: I find it confusing that the caller filters the set of resources to pass to the clean up function, e.g., https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3610/files#diff-ac45a93b8f5e87ab26c8de0dd3165846262b1cb89b0e318586f6ed2b33f61627R127, and then the clean up function filters the set further.

Can we make either the caller, or this function, responsible for all the filtering?

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -0,0 +1,149 @@
...
+	rgapi "github.com/aws/aws-sdk-go/service/resourcegroupstaggingapi"
+)
+
+func (s *Service) deleteElasticLoadbalancingResources(ctx context.Context, resources []*rgapi.ResourceTagMapping) error {
+	s.scope.V(2).Info("Deleting load balancers")
+	if err := s.deleteLoadBalancers(ctx, resources); err != nil {
+		return fmt.Errorf("deleting load balancers: %w", err)
+	}
+	if err := s.deleteTargetGroups(ctx, resources); err != nil {
+		return fmt.Errorf("deleting target groups: %w", err)
```
dlipovetsky, line 37: Do we need to confirm that the resources are gone, like we do [when we delete the API server ELB](https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/4d3a8c63fab99a3aa9c029394bce08cf343b9855/pkg/cloud/services/elb/loadbalancer.go#L176-L182)?

pkg/cloud/services/gc/options.go
```diff
@@ -0,0 +1,64 @@
...
+}
+
+// WithCleanupFuns is an option for specifying clean-up functions.
+func WithCleanupFuns(funcs map[string]ResourceCleanupFunc) ServiceOption {
+	return func(s *Service) {
+		for serviceName, fn := range funcs {
+			s.cleanupFuncs[serviceName] = fn
+		}
+	}
+}
```
dlipovetsky, line 64: Do we need to export these options? If we need them for tests, we can keep it private.

pkg/cloud/services/gc/cleanup_test.go
```diff
@@ -0,0 +1,654 @@
...
+		name         string
+		clusterScope cloud.ClusterScoper
+		elbMocks     func(m *mocks.MockELBAPIMockRecorder)
+		elbv2Mocks   func(m *mocks.MockELBV2APIMockRecorder)
+		rgAPIMocks   func(m *mocks.MockResourceGroupsTaggingAPIAPIMockRecorder)
+		ec2Mocks     func(m *mocks.MockEC2APIMockRecorder)
+		expectErr    bool
+	}{
+		{
+			name:         "eks with no Service load balances",
```
dlipovetsky, line 55:
```suggestion
			name:         "eks with no Service load balancers",
```

pkg/cloud/services/gc/cleanup_test.go
```diff
@@ -0,0 +1,654 @@
...
+					}, nil
+				})
+			},
+			elbMocks:   func(m *mocks.MockELBAPIMockRecorder) {},
+			elbv2Mocks: func(m *mocks.MockELBV2APIMockRecorder) {},
+			ec2Mocks:   func(m *mocks.MockEC2APIMockRecorder) {},
+			expectErr:  false,
+		},
+		{
+			name:         "ec2 cluster with no Service load balances",
```
dlipovetsky, line 77:
```suggestion
			name:         "ec2 cluster with no Service load balancers",
```

pkg/cloud/services/gc/cleanup_test.go
```diff
@@ -0,0 +1,654 @@
...
+			elbv2Mocks: func(m *mocks.MockELBV2APIMockRecorder) {
+				m.DeleteLoadBalancerWithContext(gomock.Any(), &elbv2.DeleteLoadBalancerInput{
+					LoadBalancerArn: aws.String("arn:aws:elasticloadbalancing:eu-west-2:1234567890:loadbalancer/app/aec24434cd2ce4630bd14a955413ee37"),
+				}).Return(&elbv2.DeleteLoadBalancerOutput{}, nil)
+			},
+			ec2Mocks:  func(m *mocks.MockEC2APIMockRecorder) {},
+			expectErr: false,
+		},
+		{
+			name:         "eks cluster full test",
```
dlipovetsky, line 403: Could you find a more descriptive name for this test?

pkg/cloud/services/gc/cleanup_test.go
```diff
@@ -0,0 +1,654 @@
...
+					}, nil
+				})
+			},
+			elbMocks:   func(m *mocks.MockELBAPIMockRecorder) {},
+			elbv2Mocks: func(m *mocks.MockELBV2APIMockRecorder) {},
+			ec2Mocks:   func(m *mocks.MockEC2APIMockRecorder) {},
+			expectErr:  false,
+		},
+		{
+			name:         "eks with non Service load balancer",
```
dlipovetsky, line 99:
```suggestion
			name:         "eks with non-Service load balancer",
```

pkg/cloud/services/gc/cleanup_test.go
```diff
@@ -0,0 +1,654 @@
...
+					}, nil
+				})
+			},
+			elbMocks:   func(m *mocks.MockELBAPIMockRecorder) {},
+			elbv2Mocks: func(m *mocks.MockELBV2APIMockRecorder) {},
+			ec2Mocks:   func(m *mocks.MockEC2APIMockRecorder) {},
+			expectErr:  false,
+		},
+		{
+			name:         "ec2 cluster with non Service load balancer",
```
dlipovetsky, line 131:
```suggestion
			name:         "ec2 cluster with non-Service load balancer",
```

pkg/cloud/services/gc/cleanup_test.go
dlipovetsky, line 99: If the LB is non-Service, does it matter what type (ALB, ELB, NLB) it is?

pkg/cloud/services/gc/cleanup_test.go
dlipovetsky, line 131: If the LB is non-Service, does it matter what type (ALB, ELB, NLB) it is?

dlipovetsky:
Really nice job, especially with all the new tests, which are complex to write! :tada: 

I looked at everything, except the `Reconcile` and `ReconcileDelete` in `pkg/cloud/services/gc/cleanup.go`, because I [asked](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3609#pullrequestreview-1050110612) about their design in the proposal PR.

pkg/annotations/annotations.go
richardcase: I agree, this has been moved within `exp/api`

pkg/annotations/annotations.go
richardcase: Ooops....they got left behind in a refactor :roll_eyes: Thanks

pkg/cloud/services/gc/cleanup.go
richardcase: Changed to add `ResourceTypeFilters: nil,`

I agree lets refactor the resource tagging into a common function in a later PR.

pkg/cloud/services/gc/ec2.go
richardcase: The caller is only filtering based on the resource type (i.e. `security-group`). I'm not sure this will be any clearer:

```go
if strings.HasPrefix(parsedARN.Resource, "security-group/") {
  eksClusterName := getTagValue(eksClusterNameTag, mapping)
  if eksClusterName != "" {
    s.scope.V(2).Info("Security group created by EKS directly, skipping deletion", "cluster_name", eksClusterName)
     continue
  }
  s.scope.V(2).Info("Deleting Security group", "arn", parsedARN.String())
  return s.deleteSecurityGroup(ctx, &parsedARN, res)
}
```

But happy to change if you want.

pkg/cloud/services/gc/loadbalancer.go
richardcase: I'm undecided. This will mean checking for every resource we deleted and so will be a bunch more api calls. In e2e testing i haven't seen any issues.

pkg/cloud/services/gc/options.go
richardcase: They are used in the tests only at the moment. Although as this is not an internal package it could be used by something else by import and these would be useful.

Will change to private.

pkg/cloud/services/gc/cleanup_test.go
richardcase: changed

pkg/cloud/services/gc/cleanup_test.go
richardcase: changed

pkg/cloud/services/gc/cleanup_test.go
richardcase: changed

richardcase: It shouldn't matter as it won't have the tags that we use for identification.

pkg/cloud/services/gc/cleanup_test.go
richardcase: It shouldn't matter as it won't have the tags that we use for identification.

richardcase:
> I looked at everything, except the `Reconcile` and `ReconcileDelete` in `pkg/cloud/services/gc/cleanup.go`, because I [asked](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3609#pullrequestreview-1050110612) about their design in the proposal PR.

@dlipovetsky - i have updated this in line with the latest changes from the proposal (which was updated based on your suggestions).

pkg/cloud/services/gc/cleanup_test.go
richardcase: Changed the name

richardcase:
Updated based on latest proposal changes....removing the extra finalizer we were using.

pkg/annotations/annotations.go
dlipovetsky: Thanks

exp/api/v1beta1/finalizers.go
```diff
@@ -0,0 +1,28 @@
...
+
+const (
+	// FargateProfileFinalizer allows the controller to clean up resources on delete.
+	FargateProfileFinalizer = "awsfargateprofile.infrastructure.cluster.x-k8s.io"
+
+	// MachinePoolFinalizer is the finalizer for the machine pool.
+	MachinePoolFinalizer = "awsmachinepool.infrastructure.cluster.x-k8s.io"
+
+	// ManagedMachinePoolFinalizer allows the controller to clean up resources on delete.
+	ManagedMachinePoolFinalizer = "awsmanagedmachinepools.infrastructure.cluster.x-k8s.io"
```
dlipovetsky, line 27: Just for reference: These constants existed before this PR. In this PR, they are moved into one file.

pkg/cloud/services/gc/ec2.go
dlipovetsky: I think what I am trying to say is this: if I had to call this function, I'm not how I would know that my input is correct.

Right now, the caller needs to filter resources before calling this function, but I think that's not enforced by the compiler.

For example, what happens if I call the function as it currently stands and pass a resource that is NOT a security group? Does `groupID` get an invalid value, and then the `DeleteSecurityGroupWithContext` call fails?

pkg/cloud/services/gc/loadbalancer.go
dlipovetsky: Good point.

The only reason to confirm a resource is gone before proceeding is if subsequent operations require the resource to be gone.

Let's say we do not confirm some LB is gone. We return from this function, and eventually finish garbage collection. Then, when CAPA begins deleting other resources, it may hit an unexpected DependencyViolation error. We expect this error to go away when the LB is eventually, but CAPA may have to requeue before that happens.

exp/api/v1beta1/finalizers.go
richardcase: Thats correct, they were moved as we added a new finalizer....which i have now removed. But it feels like a refactor to keep.

pkg/cloud/services/gc/options.go
```diff
@@ -0,0 +1,55 @@
...
+		s.resourceTaggingClient = client
+	}
+}
+
+// withEC2Client is an option for specifying a AWS EC2 Client.
+func withEC2Client(client ec2iface.EC2API) ServiceOption {
+	return func(s *Service) {
+		s.ec2Client = client
+	}
+}
```
dlipovetsky, line 55: This function is not called anywhere. Do we expect to call it?

The other functions are called in cleanup_test.go.

richardcase: Its called from the test file: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f55474951bfe224ce44c4dd34e836f6edbe60c8c/pkg/cloud/services/gc/cleanup_test.go#L555

pkg/cloud/services/gc/ec2.go
richardcase: Ah, i see what you are saying. I guess i am relying on the fact that the only exportable function is `ReconcileDelete` and from here we control what calls what. 

The current call to `deleteSecurityGroup` is protected with a check (i.e. `strings.HasPrefix(parsedARN.Resource, "security-group/") `). But what you are saying is that someone could come along later and make a change in the `gc` package somewhere and call `deleteSecurityGroup` without this check.

Its a bit late in the day for me...will ponder this tomorrow.

pkg/cloud/services/gc/options.go
dlipovetsky: (I'm sorry, I was reviewing the PR using an extension in vscode, and it was showing me stale data. My comment was about `WithCleanupFuncs`, which you already removed.)

pkg/cloud/services/gc/cleanup.go
```diff
@@ -0,0 +1,121 @@
...
+	shouldGC, err := strconv.ParseBool(val)
+	if err != nil {
+		return fmt.Errorf("converting value %s of annotation %s to bool: %w", val, expinfrav1.ExternalResourceGCAnnotation, err)
+	}
+
+	if shouldGC {
+		if err := s.deleteResources(ctx); err != nil {
+			return fmt.Errorf("deleting workload services of type load balancer: %w", err)
+		}
+	}
```
dlipovetsky, line 58: nit: Return early instead of nesting a conditional.

```suggestion
	if !shouldGC {
		return nil
	}
	
	if err := s.deleteResources(ctx); err != nil {
		return fmt.Errorf("deleting workload services of type load balancer: %w", err)
	}
```

pkg/cloud/services/gc/ec2.go
richardcase: @dlipovetsky - i have changed this now so that the clean up functions get passed the list of all the resources, they then decide to filter the resources and delete if required.

pkg/cloud/services/gc/loadbalancer.go
richardcase: The `ReconcileDelete` of the gc service can be called many times, so if we do go around the reconciliation look again for delete it won't be a problem.

pkg/cloud/services/gc/cleanup.go
richardcase: Changed.

richardcase:

And we will want all GC change in 1.5 so:

/cherry-pick release-1.5

pkg/cloud/services/gc/loadbalancer.go
Skarlso: Keep in mind that some things are deleted asynchronously and you usually have to pass in something like `Wait: true` in the options so it doesn't just mark it for deletion and then proceed. We want the call to wait I think, right?

pkg/annotations/annotations.go
```diff
@@ -0,0 +1,55 @@
...
+		return "", false
+	}
+
+	value, found = annotations[name]
+
+	return
+}
+
+// Has returns true if the supplied object has the supplied annotation.
+func Has(obj metav1.Object, name string) bool {
```
Skarlso, line 46: Nice, I almost suggested to use `return "" != annotations[name]` but remembered that annotations can exist without a value too.

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -0,0 +1,160 @@
...
+func (s *Service) deleteLoadBalancerV2(ctx context.Context, lbARN string) error {
+	input := elbv2.DeleteLoadBalancerInput{
+		LoadBalancerArn: aws.String(lbARN),
+	}
+
+	s.scope.V(2).Info("Deleting v2 load balancer", "arn", lbARN)
+	_, err := s.elbv2Client.DeleteLoadBalancerWithContext(ctx, &input)
+	if err != nil {
+		return fmt.Errorf("deleting v2 load balancer: %w", err)
+	}
```
Skarlso, line 129:
```suggestion
	if _, err := s.elbv2Client.DeleteLoadBalancerWithContext(ctx, &input); err != nil {
		return fmt.Errorf("deleting v2 load balancer: %w", err)
	}
```

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -0,0 +1,160 @@
...
+func (s *Service) deleteLoadBalancer(ctx context.Context, name string) error {
+	input := elb.DeleteLoadBalancerInput{
+		LoadBalancerName: aws.String(name),
+	}
+
+	s.scope.V(2).Info("Deleting classic load balancer", "name", name)
+	_, err := s.elbClient.DeleteLoadBalancerWithContext(ctx, &input)
+	if err != nil {
+		return fmt.Errorf("deleting classic load balancer: %w", err)
+	}
```
Skarlso, line 143:
```suggestion
	if _, err := s.elbClient.DeleteLoadBalancerWithContext(ctx, &input); err != nil {
		return fmt.Errorf("deleting classic load balancer: %w", err)
	}
```

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -0,0 +1,160 @@
...
+func (s *Service) deleteTargetGroup(ctx context.Context, targetGroupARN string) error {
+	input := elbv2.DeleteTargetGroupInput{
+		TargetGroupArn: aws.String(targetGroupARN),
+	}
+
+	s.scope.V(2).Info("Deleting target group", "arn", targetGroupARN)
+	_, err := s.elbv2Client.DeleteTargetGroupWithContext(ctx, &input)
+	if err != nil {
+		return fmt.Errorf("deleting target group: %w", err)
+	}
```
Skarlso, line 157:
```suggestion
	if _, err := s.elbv2Client.DeleteTargetGroupWithContext(ctx, &input); err != nil {
		return fmt.Errorf("deleting target group: %w", err)
	}
```

pkg/cloud/services/gc/loadbalancer.go
richardcase: Ah, i did this in other places i wonder why not in these functions, weird. Thanks @Skarlso . Changed.

pkg/cloud/services/gc/loadbalancer.go
richardcase: Changed

pkg/cloud/services/gc/loadbalancer.go
richardcase: Changed

pkg/cloud/services/gc/cleanup.go
```diff
@@ -0,0 +1,109 @@
...
+			ARN:  &parsedArn,
+			Tags: tags,
+		})
+	}
+
+	for _, cleanupFunc := range s.cleanupFuncs {
+		if deleteErr := cleanupFunc(ctx, resources); deleteErr != nil {
+			return fmt.Errorf("deleting resources: %w", deleteErr)
+		}
+	}
```
Skarlso, line 106: Optionally, this could be simplified here like this. Take the `s.cleanupFuncs` from this and have the type ( the slice type ) implement an interface which says `Execute(resources)`. That would internally loop on all its elements and call them with the same resources slice so we don't have to pass that slice around as many times as there are cleanup functions.

pkg/cloud/services/gc/ec2.go
```diff
@@ -0,0 +1,77 @@
...
+	"strings"
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
+)
+
+func (s *Service) deleteSecurityGroups(ctx context.Context, resources []*AWSResource) error {
+	for _, resource := range resources {
+		if !s.isSecurityGroupToDelete(resource) {
+			s.scope.V(4).Info("Resource not a security group for deletion", "arn", resource.ARN.String())
```
Skarlso, line 31: I wonder if this is relevant info here to display. As things are implemented, you can, by design, call this function with all of the resources and it would basically just filter resources that it needs. I think this log message will be just distraction when, by design, it can encounter other resources which it will just skip.

pkg/cloud/services/gc/ec2.go
```diff
@@ -0,0 +1,77 @@
...
+	s.scope.V(2).Info("Finished processing resources for security group deletion")
+
+	return nil
+}
+
+func (s *Service) isSecurityGroupToDelete(resource *AWSResource) bool {
+	//TODO: should we check for the security group name start with k8s-elb-
+
+	if resource.ARN.Service != ec2.ServiceName {
+		s.scope.V(5).Info("Resource is not for ec2 service", "arn", resource.ARN.String(), "check", "securitygroup")
```
Skarlso, line 49: Same as the comment here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3610/files#r932187278

pkg/cloud/services/gc/ec2.go
```diff
@@ -0,0 +1,77 @@
...
+		return false
+	}
+	if !strings.HasPrefix(resource.ARN.Resource, "security-group/") {
+		s.scope.V(5).Info("Resource is not a security group", "arn", resource.ARN.String(), "check", "securitygroup")
```
Skarlso, line 53: Same comment https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3610/files#r932187278

pkg/cloud/services/gc/ec2.go
```diff
@@ -0,0 +1,77 @@
...
+		return false
+	}
+	eksClusterName := resource.Tags[eksClusterNameTag] //nolint:ifshort
```
Skarlso, line 56: Why the nolint? :D if you don't use this value outside the if why not inline it?

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -0,0 +1,157 @@
...
+
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/elb"
+	"github.com/aws/aws-sdk-go/service/elbv2"
+)
+
+func (s *Service) deleteLoadBalancers(ctx context.Context, resources []*AWSResource) error {
+	for _, resource := range resources {
+		if !s.isLoadBalancerToDelete(resource) {
+			s.scope.V(4).Info("Resource not a load balancer for deletion", "arn", resource.ARN.String())
```
Skarlso, line 32: Same comment regarding not LB.

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -0,0 +1,157 @@
...
+			continue
+		}
+		if strings.HasPrefix(resource.ARN.Resource, "loadbalancer/") {
+			name := strings.ReplaceAll(resource.ARN.Resource, "loadbalancer/", "")
+			s.scope.V(5).Info("Deleting classic ELB for Service", "arn", resource.ARN.String(), "name", name)
+			if err := s.deleteLoadBalancer(ctx, name); err != nil {
+				return fmt.Errorf("deleting classic ELB: %w", err)
+			}
+			continue
+		}
```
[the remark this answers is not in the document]
Skarlso: So, something like:

```go
		switch {
		case strings.HasPrefix(resource.ARN.Resource, "loadbalancer/app/"),
			strings.HasPrefix(resource.ARN.Resource, "loadbalancer/net/"):
			s.scope.V(5).Info("Deleting ALB/NLB for Service", "arn", resource.ARN.String())
			if err := s.deleteLoadBalancerV2(ctx, resource.ARN.String()); err != nil {
				return fmt.Errorf("deleting ALB/NLB: %w", err)
			}
		case strings.HasPrefix(resource.ARN.Resource, "loadbalancer/"):
			name := strings.ReplaceAll(resource.ARN.Resource, "loadbalancer/", "")
			s.scope.V(5).Info("Deleting classic ELB for Service", "arn", resource.ARN.String(), "name", name)
			if err := s.deleteLoadBalancer(ctx, name); err != nil {
				return fmt.Errorf("deleting classic ELB: %w", err)
			}
		}
```

pkg/cloud/services/gc/ec2.go
richardcase: I agree that it could be a distraction and that's why I added the logs entries with verbosity of 4 & 5. If you are debugging a problem these will be helpful as you can see it was skipped and the arn.

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -0,0 +1,157 @@
...
+		if err := s.deleteTargetGroup(ctx, resource.ARN.String()); err != nil {
+			return fmt.Errorf("deleting target group %s: %w", name, err)
+		}
+	}
+	s.scope.V(2).Info("Finished processing resources for target group deletion")
+
+	return nil
+}
+
+func (s *Service) isLoadBalancerToDelete(resource *AWSResource) bool {
```
Skarlso, line 82: There is a lot of duplication in all of this doing the same thing slightly differently.

Like, the two delete functions can be one just called with a type.

The delegation is also only a single function with either the arn or the name set.

pkg/cloud/services/gc/cleanup.go
richardcase: I like this :+1:

pkg/cloud/services/gc/loadbalancer.go
Skarlso: What you could do is create two small functions saying `isLBv1` and `isLBv2` and than inside you can combine the two case checks from above. Replace those with a simple function call and you can actually combine the delegation saying something like:

```go
if isLBv2(arn) { input.LoadBalancerArn = aws.StringlbARN } else { input.LoadBalancerName = aws.String(lbARN) }
```

What do you think? :)))

Skarlso: the case above would just be: `case isLBv2(resource.ARN.Resource):`

Skarlso: isLoadBalancer and isTargetGroup can be combined:

```go
func (s *Service) isLoadBalancerToDelete(resource *AWSResource) bool {
	return s.checkResource(resource, "loadbalancer")
}

func (s *Service) isTargetGroupToDelete(resource *AWSResource) bool {
	return s.checkResource(resource, "targetgroup")
}


func (s *Service) checkResource(resource *AWSResource, prefix string) bool {
	if resource.ARN.Service != elb.ServiceName {
		s.scope.V(5).Info("Resource is not for "+prefix+" service", "arn", resource.ARN.String(), "check", prefix)
		return false
	}
	if !strings.HasPrefix(resource.ARN.Resource, prefix+"/") {
		s.scope.V(5).Info("Resource is not a load balancer", "arn", resource.ARN.String(), "check", prefix)
		return false
	}
	serviceName := resource.Tags[serviceNameTag]
	if serviceName == "" {
		s.scope.V(5).Info("Resource wasn't created for a Service via CCM", "arn", resource.ARN.String(), "check", prefix)
		return false
	}
	s.scope.V(5).Info("Resource is a load balancer to delete", "arn", resource.ARN.String(), "check", prefix, "servicename", serviceName)

	return true
}

```

pkg/cloud/services/gc/service.go
```diff
@@ -0,0 +1,76 @@
...
+	resourceTaggingClient resourcegroupstaggingapiiface.ResourceGroupsTaggingAPIAPI
+	ec2Client             ec2iface.EC2API
+	cleanupFuncs          []ResourceCleanupFunc
+}
+
+// NewService creates a new Service.
+func NewService(clusterScope cloud.ClusterScoper, opts ...ServiceOption) *Service {
+	svc := &Service{
+		scope:                 clusterScope,
+		elbClient:             scope.NewELBClient(clusterScope, clusterScope, clusterScope, clusterScope.InfraCluster()),
```
Skarlso, line 55: This looks a bit silly. :D Also, this entity implement WAAAY too many interfaces.

pkg/cloud/services/gc/loadbalancer.go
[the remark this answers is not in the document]
Skarlso: With the suggested isLBv2 function the first case could be further simplified. And then isLBv2 later can be used during deletion to determine which input value to set.

pkg/cloud/services/gc/ec2.go
richardcase: Yes there was reasoning behind that concious choice, a resource being skipped was debug level (4) but the actual reason why it was skipped was trace level (5). But i have changed to all 5.

pkg/cloud/services/gc/loadbalancer.go
richardcase: Agreed things can be deleted asynchronously. And it would be better to `Wait` after every delete. But if we wait, then it's at least 1 extra api call per resource we delete. We have struggled with the number of API calls and throttling in the past.

As it stands,  if we delete and continue with the delete reconciliation and something like a LB isn't deleted yet then we will get a `DependencyViolation` error when deleting the VPC. The reconcile delete will be retried and it will pass. 

@Skarlso @dlipovetsky - if you both strongly feel we should be waiting then i can add this is no problem.

pkg/cloud/services/gc/cleanup.go
richardcase: Changed.

pkg/cloud/services/gc/ec2.go
richardcase: Changed so they are all trace level (5).

pkg/cloud/services/gc/loadbalancer.go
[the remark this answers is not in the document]
richardcase: Changed to a switch. But I've kept them as separate cases in the switch so that we get different log messages.

pkg/cloud/services/gc/service.go
richardcase: Thats the convention used when creating services in CAPA.

pkg/cloud/services/gc/loadbalancer.go
richardcase: Changed this along these lines.

pkg/cloud/services/gc/loadbalancer.go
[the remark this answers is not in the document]
Skarlso: >Changed to a switch. But I've kept them as separate cases in the switch so that we get different log messages.

understandable-have-a-nice-day-1

pkg/cloud/services/gc/service.go
Skarlso: Yeah, no worries. I just found it a bit weird to pass in the same thing a couple of times since it implements a bunch of interfaces. Feels like it's an overloaded service. Or pass it in once, and then the client can sort out what it needs.

But this is not part of this code. :)

pkg/cloud/services/gc/loadbalancer.go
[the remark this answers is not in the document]
richardcase: :rofl:

Skarlso:
Great work! Really cool. :) I can't seem to resolve my comments. :( So sorry about them litering too much.

pkg/cloud/services/gc/service.go
richardcase: We should look at this again in the future and refactor.

richardcase: Did you want to capture it as an issue?

Skarlso: Yeah I'll create one! :)

richardcase:
> Great work! Really cool. :) I can't seem to resolve my comments. :( So sorry about them litering too much.

Thanks for the review @Skarlso - some great ideas on how to improve the implementation :bow:

pkg/cloud/services/gc/cleanup.go
dlipovetsky: A receiver on a type alias of a slice. Very neat! I don't think I've seen this before, so it looks a little exotic to my eyes, but it's only one step removed from a struct that has a slice field.

pkg/cloud/services/gc/ec2.go
dlipovetsky: I'm surprised. In my experience, if a nolint directive is not needed, the linter complains!

pkg/cloud/services/gc/loadbalancer.go
Skarlso: Nah, I'm fine honestly with not adding a bunch of API calls. My remark was more in the lines of "let's not forget this and revisit later". :)

sedefsavas:
Great work @richardcase.
/lgtm

richardcase:
With 3 different reviews lets merge. We can make changes if needed in the future.

/approve

PR #3613, opened 2022-07-22
🐛 Ensure that python and pip is installed for e2e and conformance tests

tobiasgiese:
**What type of PR is this?**

We identified in CAPO that with the latest kubekins-e2e image `gcr.io/k8s-staging-test-infra/kubekins-e2e:v20220720-b949db7485-master` the pip3 location `/usr/bin/pip3.7` does no longer exists.

```
 FileNotFoundError: [Errno 2] No such file or directory: '/usr/bin/pip3.7'
```

[pull-cluster-api-provider-openstack-e2e-test/1550462495938842624](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-openstack/1307/pull-cluster-api-provider-openstack-e2e-test/1550462495938842624)

In PR https://github.com/kubernetes-sigs/cluster-api-provider-openstack/pull/1307 we're changing this by setting a symlink. 

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

<sub>Tobias Giese <tobias.giese@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/LEGAL_IMPRINT.md)</sub>

### DIFF
--- a/scripts/ci-conformance.sh
+++ b/scripts/ci-conformance.sh
@@ -36,8 +36,13 @@ cleanup() {
 }
 trap cleanup EXIT
 
-#Install requests module explicitly for HTTP calls
-python3 -m pip install requests
+# Ensure that python3-pip is installed.
+apt update
+apt install -y python3-pip
+rm -rf /var/lib/apt/lists/*
+
+# Install/upgrade pip and requests module explicitly for HTTP calls.
+python3 -m pip install --upgrade pip requests
 
 # If BOSKOS_HOST is set then acquire an AWS account from Boskos.
 if [ -n "${BOSKOS_HOST:-}" ]; then
--- a/scripts/ci-e2e.sh
+++ b/scripts/ci-e2e.sh
@@ -47,8 +47,13 @@ cleanup() {
 }
 trap cleanup EXIT
 
-#Install requests module explicitly for HTTP calls
-python3 -m pip install requests
+# Ensure that python3-pip is installed.
+apt update
+apt install -y python3-pip
+rm -rf /var/lib/apt/lists/*
+
+# Install/upgrade pip and requests module explicitly for HTTP calls.
+python3 -m pip install --upgrade pip requests
 
 # If BOSKOS_HOST is set then acquire an AWS account from Boskos.
 if [ -n "${BOSKOS_HOST:-}" ]; then


scripts/ci-conformance.sh
```diff
@@ -36,7 +36,10 @@ cleanup() {
 }
 trap cleanup EXIT
 
-#Install requests module explicitly for HTTP calls
+# Set correct pip3.7 location for python3.
+ln -s /usr/local/bin/pip3.7 /usr/bin/pip3.7
```
Skarlso, line 40: This would blow up if it already exists though? 🤔  Isn't that the case for the current image?

tobiasgiese: `/usr/bin/pip3.7` does not exist in the current image, that's the point.

WDYT about

```suggestion
if ! test -f /usr/bin/pip3.7 && test -f /usr/local/bin/pip3.7; then
	ln -s /usr/local/bin/pip3.7 /usr/bin/pip3.7
fi
```

tobiasgiese: Or` we could easily install it via `apt install python3-requests -y`

Skarlso: I'm cool with the `if`! :)

scripts/ci-conformance.sh
```diff
@@ -36,7 +36,12 @@ cleanup() {
 }
 trap cleanup EXIT
 
-#Install requests module explicitly for HTTP calls
+# Set correct pip3.7 location for python3.
+if ! test -f /usr/bin/pip3.7 && test -f /usr/local/bin/pip3.7; then
+  ln -s /usr/local/bin/pip3.7 /usr/bin/pip3.7
```
Skarlso, line 41: Actually, one other question. Shouldn't this just be `pip3` like `python3`? I might be wrong, but isn't pip3 for all pip3+ things? So like, if we want a different version, we don't have to come and update that?

Skarlso: Of course, this is rather a mute point if pip3.7 is specifically required. :D

tobiasgiese: That's true, sure. But `python3 -m pip` is calling `/usr/bin/pip3.7`. Maybe we have to set the right version somewhere? I'm not that familiar with that Debian image

Skarlso: Yeah, I can see that. Interesting. Who and why moved Python away. :D That would be something that we would need to find out.

Skarlso: The last time it worked is `v20220624-1a63fdd9f2-master`.


2437cd785ea7 | v20220624-1a63fdd9f2-master | 1.2 GB | 24 Jun 2022 | 24 Jun 2022
-- | -- | -- | -- | --



So something happened between 24 Jun and 8th of July.

I'm not saying we shouldn't do this PR, but I would rather find out what happened then do this and then get bitten again once python3 refers to 3.8 if that makes sense?

tobiasgiese: I definitely agree with you. This is just a quick-fix to at least fix our CAPO conformance tests -- and maybe CAPA conformance tests.

Skarlso: Honestly, I think this is a bug in the image building process.

Skarlso: https://github.com/kubernetes/test-infra/commit/fa9fc389d33f92e9d41932d0c16dbe13ff8e9252

Added no-cache. That's the only pip related change and it somehow caused all of this.

Skarlso: Op! Found it! :D 

https://github.com/kubernetes/test-infra/pull/24967

Skarlso: Actually this hasn't been working for the infra image since 7th of June. This is the last working image: `v20220607-4b3fb50c36-test-infra`.

Skarlso:
For tracking the pip change https://github.com/kubernetes/test-infra/issues/26900

Skarlso:
/lgtm FWIW

I checked the images, indeed, pip has moved. Created a relevant issue in the sig project. Otherwise I'm okay with this so people are unblocked.

We can revisit it once we know more about the corresponding bug.

scripts/ci-conformance.sh
tobiasgiese: Thanks for finding out, really appreciate your help! :)

Skarlso:
Hmpf. So... They aren't taking any guarantees around python even being a thing. I'm wondering if this PR should just install Python and pip on it's own instead of doing a linking following this comment https://github.com/kubernetes/test-infra/issues/26900#issuecomment-1195163041.

Let's do a `python3 -m pip install --upgrade pip` and then `python3 -m pip install requests` will naturally work.

What do you think @tobiasgiese? Can you please do that change? Let's make sure our images have python and pip properly installed instead of relying on it to be there.

tobiasgiese:
> What do you think @tobiasgiese? Can you please do that change?

Sure, I've updated the PR. PTAL :)\

Thanks for clarifying with the sig testing team! 👍🏻

Skarlso:
Thank you! :)

Skarlso:
Cool, thanks for making the change Tobias! :) LGTM. :)

PR #3617, opened 2022-07-25
Bump github.com/golangci/golangci-lint from 1.45.2 to 1.47.2 in /hack/tools

Ankitasw:
**What type of PR is this?**
/area release

**What this PR does / why we need it**:
This PR does the following:
- Bumps github.com/golangci/golangci-lint from 1.45.2 to 1.47.2 in /hack/tools.
- Disable nonamedreturns, nosnakecase and exhaustruct linters.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:
- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #3632, opened 2022-07-29
[release-1.5] feat: external load balancer garbage collection (part 2) - new gc service

richardcase:
### This is a manual cherry-pick of #3610 

**What type of PR is this?**

/kind feature

This introduces a new garbage collection service which will be used later by the controllers to do clean up of AWS resources for workload clusters that were created via the CCM. Initially, we support deleting load balancers (classic elb, nlb, alb), target groups and security groups.

The service is design to be called by the infra cluster reconcilers when a cluster is delete. The entry point is **ReconcileDelete** into the service.

When a cluster is deleted the cluster controllers will call **ReconcileDelete**. The first job is to determine if the workload cluster being deleted should be garbage collected. A cluster will be garbage collected if either of these are true:

- the `aws.cluster.x-k8s.io/external-resource-gc` annotation is absent
- the `aws.cluster.x-k8s.io/external-resource-gc `annotation exists and its value is set to true

If a cluster is to be garbage collected then we need to identify the AWS resources that where created by the CCM in the workload cluster. This is done by using the AWS resource tagging api and getting resources with the Kubernetes cluster owned label (i.e. `kubernetes.io/cluster/[CLUSTERNAME]`). The resources that are returned are converted and then **the list of them is passed to the defined cleanup functions in order**. The order of the functions matter because some AWS resources cannot be delete before others. For example, target groups can only be deleted after load balancers. The cleanup function will then iterate over the resources and decide if it wants to handle that resource and delete it from AWS.

**The service will be use by a later PR to the controllers and so is initially unused.**

Relates #1718 

This part 2 of 4 of changes (i.e. a stack) to implement the garbage collection. This is work to split up the original pr https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3518

**Checklist**:

- [x] squashed commits
- [x] adds unit tests


sedefsavas:
We might need to backport a linter update: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3617

sedefsavas:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3637 should fix it.

Ankitasw:
@richardcase some issue with multitenancy test case 🤔 ?

richardcase:
All passing now :)

/approve

PR #3648, opened 2022-08-04
feat: external load balancer garbage collection (part 4) - e2e tests

### LINKED ISSUE #1718: AWS NLB linger after they're orphaned

richardcase:
**What type of PR is this?**

/kind feature

This change introduces new e2e tests that test the new garbage
collection feature. There are tests for both managed and unmanaged
clusters.

These are separate test suites whilst the GC feature is experimental as
we need to enable this feature just for our tests without effecting the
existing e2e tests.

When this feature moves out of experimental we will merge these tests
into the existing managed & unmanaged suites.

Fixes #1718 

This part 4 of 4 of changes (i.e. a stack) to implement the garbage collection. This is work to split up the original pr #3518. As this is the final part it will fix #1718. 

**Checklist**:

- [x] squashed commits


richardcase:
We want this in 1.5 so:

/cherry-pick release-1.5

Makefile
```diff
@@ -412,6 +412,14 @@ test-e2e: $(GINKGO) $(KIND) $(SSM_PLUGIN) $(KUSTOMIZE) generate-test-flavors e2e
 test-e2e-eks: generate-test-flavors  $(GINKGO) $(KIND) $(SSM_PLUGIN) $(KUSTOMIZE) e2e-image ## Run eks e2e tests
 	time $(GINKGO) -tags=e2e $(GINKGO_ARGS) ./test/e2e/suites/managed/... -- -config-path="$(E2E_EKS_CONF_PATH)" --source-template="$(EKS_SOURCE_TEMPLATE)" $(E2E_ARGS) $(EKS_E2E_ARGS)
 
+.PHONY: test-e2e-gc ## Run garbage collection e2e tests using clusterctl
+test-e2e-gc: generate-test-flavors  $(GINKGO) $(KIND) $(SSM_PLUGIN) $(KUSTOMIZE) e2e-image ## Run eks e2e tests
+	time $(GINKGO) -tags=e2e -focus="$(GINKGO_FOCUS)" -skip="$(GINKGO_SKIP)" $(GINKGO_ARGS) -p ./test/e2e/suites/gc_unmanaged/... -- -config-path="$(E2E_CONF_PATH)" $(E2E_ARGS)
+
+.PHONY: test-e2e-eks-gc ## Run EKS garbage collection e2e tests using clusterctl
+test-e2e-eks-gc: generate-test-flavors  $(GINKGO) $(KIND) $(SSM_PLUGIN) $(KUSTOMIZE) e2e-image ## Run eks e2e tests
+	time $(GINKGO) -tags=e2e -focus="$(GINKGO_FOCUS)" -skip="$(GINKGO_SKIP)" $(GINKGO_ARGS) ./test/e2e/suites/gc_managed/... -- -config-path="$(E2E_EKS_CONF_PATH)" --source-template="$(EKS_SOURCE_TEMPLATE)" $(E2E_ARGS) $(EKS_E2E_ARGS)
```
Skarlso, line 421: There is a way in CAPA to interchange EKS with managed from time to time. Like here, you call this `test-e2e-eks-gc` but the folder is called `gc_managed`. :D

I guess it's just better named like this because the other would be something like `gc_noneeks` which just doesn't sound nice. :D

scripts/ci-e2e-eks-gc.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+
+# Copyright 2019 The Kubernetes Authors.
```
Skarlso, line 3:
```suggestion
# Copyright 2022 The Kubernetes Authors.
```

scripts/ci-e2e-gc.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
+
+# Copyright 2019 The Kubernetes Authors.
```
Skarlso, line 3:
```suggestion
# Copyright 2022 The Kubernetes Authors.
```

scripts/ci-e2e-gc.sh
```diff
@@ -0,0 +1,111 @@
+#!/bin/bash
```
Skarlso, line 1: This is should be combined with the previous one and create a single script with a cli argument saying something like `{managed,unmanaged}`. WDYT?

test/e2e/shared/aws_helpers.go
```diff
@@ -0,0 +1,178 @@
...
+	}
+	descInput := &DescribeResourcesByTagsInput{
+		AWSSession: input.AWSSession,
+		Tags:       tags,
+	}
+
+	descOutput, err := DescribeResourcesByTags(*descInput)
+	if err != nil {
+		fmt.Fprintf(GinkgoWriter, "error querying resources by tags: %v\n", err)
+
```
Skarlso, line 102:
```suggestion
```

test/e2e/shared/aws_helpers.go
```diff
@@ -0,0 +1,178 @@
...
+		return nil, fmt.Errorf("describing resource tags: %w", err)
+	}
+
+	matchingARNs := []string{}
+	for i := range descOutput.ARNs {
+		resARN := descOutput.ARNs[i]
+		parsedArn, err := arn.Parse(resARN)
+		if err != nil {
+			fmt.Fprintf(GinkgoWriter, "error parsing arn %s: %v\n", resARN, err)
+
```
Skarlso, line 112:
```suggestion
```

scripts/ci-e2e-gc.sh
richardcase: These scripts are only temporary while the feature is experimental. Ultimately, they will be removed and these tests will be incorporated into the managed and unmanaged suites.

So i basically haven't spent anytime optimizing this.

Skarlso: ah, ok. makes sense then. Thanks.

scripts/ci-e2e-eks-gc.sh
richardcase: Changed

scripts/ci-e2e-gc.sh
richardcase: Changed

test/e2e/shared/aws_helpers.go
richardcase: Changed

test/e2e/shared/aws_helpers.go
richardcase: Changed

test/e2e/shared/aws_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+			fmt.Fprintf(GinkgoWriter, "error parsing arn %s: %v\n", resARN, err)
+			return nil, fmt.Errorf("parsing resource arn %s: %w", resARN, err)
+		}
+
+		if parsedArn.Service != "elasticloadbalancing" {
+			continue
+		}
+
+		switch input.Type {
+		case LoadBalancerTypeALB:
```
Skarlso, line 119: This is a bit weird. If you already have the type, is it necessary in the test to check the arn as well?
Wouldn't you make sure in the test that once you say it's an NLB it IS actually an NLB? :)

Or am I misunderstanding what this does?

test/e2e/shared/aws_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+
+type GetLoadBalancerARNsInput struct {
+	AWSSession       client.ConfigProvider
+	ServiceName      string
+	ServiceNamespace string
+	ClusterName      string
+	Type             LoadBalancerType
+}
+
+func GetLoadBalancerARNs(ctx context.Context, input GetLoadBalancerARNsInput) ([]string, error) {
```
Skarlso, line 86: I don't see context being used and this doesn't appear to implement any interfaces... soo... remove the context? :)

test/e2e/shared/aws_helpers.go
```diff
@@ -0,0 +1,176 @@
...
+
+	descOutput, err := DescribeResourcesByTags(*descInput)
+	if err != nil {
+		fmt.Fprintf(GinkgoWriter, "error querying resources by tags: %v\n", err)
+		return nil, fmt.Errorf("describing resource tags: %w", err)
+	}
+
+	matchingARNs := []string{}
+	for i := range descOutput.ARNs {
+		resARN := descOutput.ARNs[i]
```
Skarlso, line 107: Um, we aren't changing this is there a reason you require the `i` and this can't be just:
```go
for _, resARN := range descOutput {
```
?

test/e2e/shared/aws_helpers.go
richardcase: This function is used to get all the loadbalancer ARNs for a specific type of load balancer....which is supplied.  Using the resource tagging api like this means i only need 1 function.

test/e2e/shared/workload.go
```diff
@@ -0,0 +1,76 @@
...
+				return true
+			}
+		}
+		return false
+	}, intervals...).Should(BeTrue(), func() string { return DescribeFailedDeployment(input, deployment) })
+}
+
+// DescribeFailedDeployment returns detailed output to help debug a deployment failure in e2e.
+func DescribeFailedDeployment(input WaitForDeploymentsAvailableInput, deployment *appsv1.Deployment) string {
+	b := strings.Builder{}
```
Skarlso, line 67: Neat use of strings.Builder. :)

test/e2e/shared/aws_helpers.go
richardcase: Yep or use it.

richardcase: Removed it.

test/e2e/shared/aws_helpers.go
richardcase: No reason, changed.

Skarlso:
Great work, Richard! :)
/lgtm

Issue #3076, opened 2022-01-14
imageLookupFormat should support Kubernetes versions with build metadata, e.g. +fips


dkoshkin:
/kind feature

**Describe the solution you'd like**

I'm trying to use `imageLookupFormat: my-ami-{{.BaseOS}}-fips-release-?{{.K8sVersion}}-*` with a custom AMIs and a Kubernetes version that has some build metadata `v1.21.8+fips.0`.
AWS AMIs do not support `+` in the name, and when we build our AMIs using Packer we use [clean_resource_name](https://www.packer.io/docs/templates/legacy_json_templates/engine#clean_resource_name) to format the name, which replaces unsupported characters with a `-`.

So the AMI name ends up being something like `my-ami-rhel-8.2-1.21.8-fips.0-1641416145` but the AMI that is being searched for is `my-ami-rhel-8.2-?1.21.8+fips.0-*` (note the `+` vs `-`)

The code that is handling [this is here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1a61dfc39649749b8b8dfad8334909a388d675a7/pkg/cloud/services/ec2/ami.go#L73-L77), but its only handling an optional `v` prefix and no other changes.

**Anything else you would like to add:**

Originally asked in Slack https://kubernetes.slack.com/archives/CD6U2V71N/p1641571157011400 and discussed 01/10/2022 CAPA meeting.

**Environment:**

- Cluster-api-provider-aws version: https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/tag/v0.7.2
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sedefsavas:
There is a change in the AMI name, search logic specifically needs to handle `+` to `-` conversion. 
 With a generic logic, not sure if this specific case can be solved. 

[Runtime-extensions proposal ](https://github.com/kubernetes-sigs/cluster-api/issues/5175)in cluster-api might be useful to solve this issue: an extension can generate a replaced regex and then that regex is used in CAPA AMI search.

dlipovetsky:
For reference:
> Constraints: 3-128 alphanumeric characters, parentheses (()), square brackets ([]), spaces ( ), periods (.), slashes (/), dashes (-), single quotes ('), at-signs (@), or underscores(_)
> -- https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateImage.html

PR #3665, opened 2022-08-11
Add strings.ReplaceAll func for AMI name format template

dkoshkin:
**What type of PR is this?**
Introducing a mechanism to support Kubernetes versions with a build name, e.g. `v1.21.8+fips.0`, to be used in the `ImageLookupFormat`. The benefit of this approach is that its an opt-in feature, generic enough to be used for other substitutions and does not modify existing behavior.

Please see the unit test for a real usecase we have.

I hope the maintainers will consider this change, but I do understand that https://github.com/kubernetes-sigs/cluster-api/pull/6181 is actively being worked on.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3076

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/docs/book/src/topics/images/custom-amis.md
+++ b/docs/book/src/topics/images/custom-amis.md
@@ -23,6 +23,8 @@ See [Upgrading workload clusters][upgrading-workload-clusters] for more details.
 
 ## Creating a cluster from a custom image
 
+### Using the AMI ID
+
 To use a custom image, it needs to be referenced in an `ami:` section of your `AWSMachineTemplate`.
 
 ```yaml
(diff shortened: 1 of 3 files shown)


Ankitasw:
Hello @dkoshkin. Thanks for the PR.
I think it's better to discuss this in next office hours to understand the use case and it's criticality, and then get the community feedback. Based on that you could open an issue, or progress this PR itself.

sedefsavas:
This LGTM.
I created an AMI named `custom-amazon-2-1.24.1-build.0-1655360804`, it'd be good to add an e2e test for this.

Let me know if I need to change the name.
Also, would be good to add this to custom AMI documentation.

sedefsavas:
Btw the image is in us-west-1

dkoshkin:
Thanks @sedefsavas!

I just added the documentation but while writing I realized I need to think a little more if adding this complexity is worth it.
It may be better to ask users to set `ami.id` in this scenario.

sedefsavas:
ACK, we can go over it in the office hours if needed.

Ankitasw:
@dkoshkin would you be able to continue with this PR?

PR #3675, opened 2022-08-18
updating deprecated ENI_CONFIG_LABEL_DEF value

luthermonson:
**What type of PR is this?**
/kind bug
Since 1.17 this label is now [deprecated](https://kubernetes.io/docs/reference/labels-annotations-taints/#failure-domainbetakubernetesiozone). Switching the env to the new [label](https://kubernetes.io/docs/reference/labels-annotations-taints/#topologykubernetesioregion).

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/awsnode/cni.go
+++ b/pkg/cloud/services/awsnode/cni.go
@@ -168,7 +168,7 @@ func (s *Service) ReconcileCNI(ctx context.Context) error {
 				},
 				corev1.EnvVar{
 					Name:  "ENI_CONFIG_LABEL_DEF",
-					Value: "failure-domain.beta.kubernetes.io/zone",
+					Value: "topology.kubernetes.io/zone",
 				},
 			)
 		}


Skarlso:
Hmmm won't this negatively affect existing environments?

luthermonson:
you would prefer if it's already set to leave it as is and only do this in new instances?

luthermonson:
thought about it some more...

1st thought is this is actually [custom networking configuration](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) and it's turned on by default and it's probably safer to lock some of this stuff behind some boolean gate. there are also more env vars that are not implemented that im looking into.

2nd thought, eks is all experimental and this would be an easy release note + label update for anyone wanting to bump capa versions.

Skarlso:
Ah sorry I missed your answer. Eks is experimental that's true. However, it is quite in use.

So we must be vigilant in changing things that might affect existing clusters.

Skarlso:
For example, suppose you do this and run this on an existing cluster. In that case, I think it would reconcile the DaemonSet with the new environment property, which probably would trigger the recycling of the pods to apply the new environment property, which is an unexpected behaviour when doing a cluster API update.

Either that is fine to do and an accepted behaviour and well documented, OR, we don't do anything if the old value exists, but when creating a new cluster, we use the new value. And after a while, we can remove the check for the old value.

WDYT? @sedefsavas any objects or correction on my thinking? :)

Skarlso:
Oh! Also, I believe this will create two values with the same name.

Because the environment property list allows for that, unfortunately. It doesn't de-duplicate normally.

We should test this behaviour.

Skarlso:
When testing I actually discovered a different bug that the environment property changes weren't being applied correctly:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3677

Skarlso:
Also related, the values CAN have duplicate entries: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3678

So this change will not update a value but duplicate it.

luthermonson:
closing for a bigger PR with more stuff

PR #3677, opened 2022-08-19
Fix the update logic for aws-node daemonset environment properties

Skarlso:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

We apply environment property changes to aws-node. These changes are only applied to the loop variable, but the actual update doesn't update them correctly.

When I change the values, the update doesn't do anything.

Before:

- create node
```
Environment:
      ADDITIONAL_ENI_TAGS:                    {}
      AWS_VPC_CNI_NODE_PORT_SUPPORT:          true
      AWS_VPC_ENI_MTU:                        9001
      AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER:     false
      AWS_VPC_K8S_CNI_EXTERNALSNAT:           false
      AWS_VPC_K8S_CNI_LOGLEVEL:               DEBUG
      AWS_VPC_K8S_CNI_LOG_FILE:               /host/var/log/aws-routed-eni/ipamd.log
      AWS_VPC_K8S_CNI_RANDOMIZESNAT:          prng
      AWS_VPC_K8S_CNI_VETHPREFIX:             eni
      AWS_VPC_K8S_PLUGIN_LOG_FILE:            /var/log/aws-routed-eni/plugin.log
      AWS_VPC_K8S_PLUGIN_LOG_LEVEL:           DEBUG
      DISABLE_INTROSPECTION:                  false
      DISABLE_METRICS:                        false
      DISABLE_NETWORK_RESOURCE_PROVISIONING:  false
      ENABLE_IPv4:                            true
      ENABLE_IPv6:                            false
      ENABLE_POD_ENI:                         false
      ENABLE_PREFIX_DELEGATION:               false
      MY_NODE_NAME:                            (v1:spec.nodeName)
      WARM_ENI_TARGET:                        1
      WARM_PREFIX_TARGET:                     1
      AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG:     true
```
- change value
```
Environment:
      ADDITIONAL_ENI_TAGS:                    {}
      AWS_VPC_CNI_NODE_PORT_SUPPORT:          true
      AWS_VPC_ENI_MTU:                        9001
      AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER:     false
      AWS_VPC_K8S_CNI_EXTERNALSNAT:           false
      AWS_VPC_K8S_CNI_LOGLEVEL:               DEBUG
[...]

PR #3681, opened 2022-08-20
rework vpc cni envvar logic to drop the forced keys

### LINKED ISSUE #3678: aws-node environment properties can be duplicated
/kind bug

In the environment property update code in cni.go the environment properties can have duplicate entries if they exist already.

```
Environment:
      ADDITIONAL_ENI_TAGS:                    {}
      AWS_VPC_CNI_NODE_PORT_SUPPORT:          true
      AWS_VPC_ENI_MTU:                        9001
      AWS_VPC_K8S_CNI_CONFIGURE_RPFILTER:     false
      AWS_VPC_K8S_CNI_EXTERNALSNAT:           false
      AWS_VPC_K8S_CNI_LOGLEVEL:               DEBUG
      AWS_VPC_K8S_CNI_LOG_FILE:               /host/var/log/aws-routed-eni/ipamd.log
      AWS_VPC_K8S_CNI_RANDOMIZESNAT:          prng
      AWS_VPC_K8S_CNI_VETHPREFIX:             eni
      AWS_VPC_K8S_PLUGIN_LOG_FILE:            /var/log/aws-routed-eni/plugin.log
      AWS_VPC_K8S_PLUGIN_LOG_LEVEL:           DEBUG
      DISABLE_INTROSPECTION:                  false
      DISABLE_METRICS:                        false
      DISABLE_NETWORK_RESOURCE_PROVISIONING:  false
      ENABLE_IPv4:                            true
      ENABLE_IPv6:                            false
      ENABLE_POD_ENI:                         false
      ENABLE_PREFIX_DELEGATION:               false
      MY_NODE_NAME:                            (v1:spec.nodeName)
      WARM_ENI_TARGET:                        1
      WARM_PREFIX_TARGET:                     1
      SOME_KEY:                               some-value
      SOME_NAME:                              some-vale
      AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG:     true
      ENI_CONFIG_LABEL_DEF:                   failure-domain.beta.kubernetes.io/zone
      SOME_NAME:                              some-vale
```

Notice `SOME_NAME` is there twice.

De-duplication to happen to the environment properties.

**Environment:**

luthermonson:
**What type of PR is this?**
/kind bug

when adding vpc cni env it [forced envvars](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/awsnode/cni.go#L165-L172) for `aws-node` but we now know due to a [bug](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3677) were never being applied. This PR drops them entirely and pushes all the onus for VPC CNI Environment Variables to the user and makes CAPA less opinionated since these env vars are being added/dropped/deprecated constantly.

In addition, this Fixes #3678 because we aren't looping/assigning the env a second time and the dedupe [code above](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/awsnode/cni.go#L71-L75) is already fixing the daemonset container env properly.

Cleans up envvar assigning and makes CAPA less opinionated.

I merged the env var tests into one which does the tests twice once with secondary cidr and once without.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/docs/book/src/topics/eks/pod-networking.md
+++ b/docs/book/src/topics/eks/pod-networking.md
@@ -7,8 +7,26 @@ When creating a EKS cluster the Amazon VPC CNI will be used by default for Pod N
 ## Using the VPC CNI Addon
 You can use an explicit version of the Amazon VPC CNI by using the **vpc-cni** EKS addon. See the [addons](./addons.md) documentation for further details of how to use addons.
 
+## Using Custom VPC CNI Configuration
+If your use case demands [custom networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) VPC CNI configuration you might already be familiar with the [helm chart](https://github.com/aws/amazon-vpc-cni-k8s) which helps with the process. This gives you access to ENI Configs and you can set Environment Variables on the `aws-node` DaemonSet where the VPC CNI runs. CAPA is able to tune the same DaemonSet through Kubernetes.
 
-## Increase node pod limit
+The following example shows how to turn on custom network config and set a [label definition](https://github.com/aws/amazon-vpc-cni-k8s#eni_config_label_def).
+
+```yaml
+kind: AWSManagedControlPlane
+apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+metadata:
+  name: "capi-managed-test-control-plane"
+spec:
+  vpcCni:
+    env:
+    - name: AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG
+      value: "true" 
+    - name: ENABLE_PREFIX_DELEGATION
+      value: "true"
+```
+
+### Increase node pod limit
 You can increase the pod limit per-node as [per the upstream AWS documentation](https://aws.amazon.com/blogs/containers/amazon-vpc-cni-increases-pods-per-node-limits/). You'll need to enable the `vpc-cni` plugin addon on your EKS cluster as well as enable prefix assignment mode through the `ENABLE_PREFIX_DELEGATION` environment variable.
 
 ```yaml
@@ -19,6 +37,8 @@ metadata:
 spec:
   vpcCni:
     env:
+    - name: AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG
+      value: "true" 
     - name: ENABLE_PREFIX_DELEGATION
       value: "true"
   addons:
--- a/pkg/cloud/services/awsnode/cni.go
+++ b/pkg/cloud/services/awsnode/cni.go
@@ -72,7 +72,6 @@ func (s *Service) ReconcileCNI(ctx context.Context) error {
 		for i := range ds.Spec.Template.Spec.Containers {
 			container := &ds.Spec.Template.Spec.Containers[i]
 			if container.Name == "aws-node" {
-				container.Env = s.filterEnv(container.Env)
 				container.Env, needsUpdate = s.applyUserProvidedEnvironmentProperties(container.Env)
 			}
 		}
@@ -161,22 +160,7 @@ func (s *Service) ReconcileCNI(ctx context.Context) error {
 		}
 	}
 
-	s.scope.Info("updating containers", "cluster", klog.KRef(s.scope.Namespace(), s.scope.Name()))
-	for i := range ds.Spec.Template.Spec.Containers {
-		if ds.Spec.Template.Spec.Containers[i].Name == "aws-node" {
-			ds.Spec.Template.Spec.Containers[i].Env = append(s.filterEnv(ds.Spec.Template.Spec.Containers[i].Env),
-				corev1.EnvVar{
-					Name:  "AWS_VPC_K8S_CNI_CUSTOM_NETWORK_CFG",
-					Value: "true",
-				},
-				corev1.EnvVar{
-					Name:  "ENI_CONFIG_LABEL_DEF",
-					Value: "failure-domain.beta.kubernetes.io/zone",
-				},
-			)
-		}
-	}
-
+	s.scope.Info("updating containers", "cluster-name", s.scope.Name(), "cluster-namespace", s.scope.Namespace())
 	return remoteClient.Update(ctx, &ds, &client.UpdateOptions{})
 }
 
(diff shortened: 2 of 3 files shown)


Skarlso:
Sadly, according to the AWS documentation it is not something that you can drop.

These fields just happened to not used by too many people I assume, because you need custom networking for that and define SecondaryCidrBlock for it. ( Which I think is not something many people would be using ).

For more detail see here: https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html

luthermonson:
after talking with @Skarlso and a solutions architect we have internally i'd like to propose we still move forward with this PR and document that if you use a Secondary CIDR Block that you know what you're doing and you also set your VpcCni.Env to what you need. this makes this code simpler, accounts for dedupes and opens up the API to be more user defined as the env vars change. per what the [vpc cni helm chart](https://github.com/aws/amazon-vpc-cni-k8s#cni-configuration-variables) is doing they'll continue to use env vars to feature gate and this way CAPA just opens it up to user defined env and only reconciles based on what it's told rather than trying to infer and automatically set defaults.

Skarlso:
This is probably okay as these values were never really applied ever. And no one complained. 😂 I refee to others in this decision.

Ankitasw:
> this way CAPA just opens it up to user defined env and only reconciles based on what it's told rather than trying to infer and automatically set defaults.
@luthermonson I have not much idea of these env vars as I have not used it myself, but as you said CAPA is setting the defaults and not all of them, don't you think it is good to have the current implementation after @Skarlso 's [changes](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3677) that sets only default values? Do you still face issues after the above fix?

luthermonson:
It's not setting the defaults as one env var value is already deprecated. I don't think capa should have an opinion of which env vars are set and the user should decide per cluster. The only one I might be willing to concede on is the var to enable custom config but I don't like not giving the user the choice.

Skarlso:
@luthermonson There is no choice. Without it, it just doesn't work properly and might cause subtle bugs.

If CAPA doesn't set it and the user forgets about it, it might cause inconsistent behaviour.

luthermonson:
Just saying... it never worked. And the way it's written if you changed that topology key it overwrote your decision. I disagree that capa should have an opinion on these but I'll do whatever you want.

Are we saying let's keep the enable custom cni and drop the topology env var?

Skarlso:
I'm not saying it worked. :D I'm saying that technically it's a requirement. :) Regardless of whether this worked or not in the past. :) 

I'm fine with leaving them out if there is sufficient documentation explaining the need to define these values to the cni once extra networking is enabled.

Skarlso:
@Ankitasw To answer your question. This final part was messing with the deduplication list done a couple of steps before this. So removing the whole environment editing makes sure that deduplication works and stays like that since we don't mess with the environment values any more.

luthermonson:
lol fair, I'd prefer to add the docs, I don't think it's a stretch to the imagination that if you are looking to do custom cni you want complete control and that includes turning it off/on on demand. Your call tho

Skarlso:
>lol fair, I'd prefer to add the docs, I don't think it's a stretch to the imagination that if you are looking to do custom cni you want complete control and that includes turning it off/on on demand. Your call tho

I'm fine with this unless some else finds any problems with it. :)

luthermonson:
@Skarlso updated the documentation

docs/book/src/topics/eks/pod-networking.md
```diff
@@ -7,8 +7,26 @@ When creating a EKS cluster the Amazon VPC CNI will be used by default for Pod N
 ## Using the VPC CNI Addon
 You can use an explicit version of the Amazon VPC CNI by using the **vpc-cni** EKS addon. See the [addons](./addons.md) documentation for further details of how to use addons.
 
+## Using Custom VPC CNI Configuration
+If your use case demands [custom networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) VPC CNI configuration you might already be familiar with the [helm chart](https://github.com/aws/amazon-vpc-cni-k8s) which helps with the process. This gives you access to ENI Configs and you can set Environment Variables on the `aws-node` daemonset where the VPC CNI runs. The Cluster API Provider is able to tune the same daemonset through the API.
```
Skarlso, line 11:
```suggestion
If your use case demands [custom networking](https://docs.aws.amazon.com/eks/latest/userguide/cni-custom-network.html) VPC CNI configuration you might already be familiar with the [helm chart](https://github.com/aws/amazon-vpc-cni-k8s) which helps with the process. This gives you access to ENI Configs and you can set Environment Variables on the `aws-node` DaemonSet where the VPC CNI runs. CAPA is able to tune the same DaemonSet through Kubernetes.
```

docs/book/src/topics/eks/pod-networking.md
```diff
@@ -7,8 +7,26 @@ When creating a EKS cluster the Amazon VPC CNI will be used by default for Pod N
...
 
-## Increase node pod limit
+The following example shows how to turn on custom network config and set a [label definition](https://github.com/aws/amazon-vpc-cni-k8s#eni_config_label_def).
```
Skarlso, line 13: > [label definition]

Am I missing something here? These are environment properties that we are setting not labels. Right? :D

luthermonson: label definition is what they call that setting it's what they use to look up nodes by label and do whatever it is they do. that's why i linked to the helm chart docs to show people what im referencing.

Skarlso: Very weird stuff.

Skarlso:
/hold
until tests

docs/book/src/topics/eks/pod-networking.md
luthermonson: networkin' aint easy

luthermonson:
@sedefsavas @richardcase can we get this merged?

richardcase:
Even though this technically didn't work it's a change in behaviour, so we should do this as part of the bump v1beta2.

Whilst i check with @sedefsavas 

/hold

luthermonson:
@Skarlso i had to rebase this should be gtg again

Ankitasw:
@richardcase @Skarlso i thinks folks are ok to merge this, shall we go ahead?

Skarlso:
Yep I think it's good to go.

Issue #3594, opened 2022-07-13
For EKS, CAPI truncates the `version` field in AWSManagedControlPlane spec after apply


yogeek:
/kind bug

**What steps did you take and what happened:**

When setting k8s version for nodegroups in CAPI YAML manifest (`AWSManagedControlPlane.spec.template.spec.version`), EKS only allows MAJOR.MINOR (e.g. `1.22`) but ClusterAPI requires a "semantic version" MAJOR.MINOR.PATCH (e.g. 1.22.10)
- when setting 1.22 in the manifest, clusterAPI error is : `invalid: spec.template.spec.version: Invalid value: "v1.22": must be a valid semantic version`
- when setting 1.22.10, clusterAPI is ok **but** it modifies the final resources in K8S by truncating to `1.22` which causes a diff in gitops tool like ArgoCD (and if we ignore this diff, we will not be able to detect a version change later)

**What did you expect to happen:**

I expected ClusterAPI to not change a resource after it has been applied and to let version the same as declared in the initial manifest (in GIT). 

**Anything else you would like to add:**
This issue has already been discussed orally with @richardcase  

**Environment:**

- Cluster-api-provider-aws version: 1.4.1
- Kubernetes version: 1.22.9

sedefsavas:
MachinePool version looks like having a validation check for updates as well, so should not be changeable to v1.22 looking at this:
https://github.com/kubernetes-sigs/cluster-api/blob/ef7172a853344d3b047fb0ace3f349079f3eef1f/exp/api/v1beta1/machinepool_webhook.go#L149

Could you confirm if MachinePool.spec.template.spec.version is set to v1.22?

yogeek:
Hi @sedefsavas thanks for your help
Sorry, I mixed up resources and indeed, it is not `MachinePool` version which was truncated but `AWSManagedControlPlane` version
I am going to update the issue title accordingly

Here is the GIT manifest :
```
apiVersion: controlplane.cluster.x-k8s.io/v1beta1
kind: AWSManagedControlPlane
metadata:
  name: capi-eks-mycluster-control-plane
  namespace: mycluster-capi
spec:
  [...]
  version: v1.22.0
```
and the result after cluster api created the cluster /
```
apiVersion: controlplane.cluster.x-k8s.io/v1beta1
kind: AWSManagedControlPlane
metadata:
  creationTimestamp: '2022-07-18T10:18:10Z'
[...]

PR #3682, opened 2022-08-20
Relax the AWSManagedControlPlane version regex and remove normalising the version

Skarlso:
**What type of PR is this?**

/kind bug

Removes the normalisation we perform for the version given to the AWSManagedControlPlane object. EKS requires a major.minor version format and we did that by truncating any given version and having a regex validation for the version being major.minor.

This, however, lead to us editing the spec and causing diff in applied / existing configs.

This PR attempts to remove this normalisation and relax the regex, so existing configurations don't break ( normally they would break, because we no longer truncate a given version ).

However, everywhere internally where we DO use the version for various things like creating a cluster, it should already be normalised.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3594.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
+++ b/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
@@ -2700,7 +2700,7 @@ spec:
                   version number is supplied then the latest version of Kubernetes
                   that EKS supports will be used.
                 minLength: 2
-                pattern: ^v(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.?$
+                pattern: ^v?(0|[1-9][0-9]*)\.(0|[1-9][0-9]*)\.?(\.0|[1-9][0-9]*)?$
                 type: string
               vpcCni:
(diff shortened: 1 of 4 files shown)


Ankitasw:
> /ok-to-test

Since you a member of kubernetes-sigs, no need for the ok-to-test label, your changes would be tested automatically.

Skarlso:
@Ankitasw Ah! Okay, thank you! :) I thought it's always required. :D Cheers. :)

richardcase:
This will need to go into v1beta2 as its a change to the API (i.e. the format/validation). 

Tagging in #2355 so that we know this is linked to api changes.

richardcase:
Although thinking about it the regex is backwards compatible so if someone already has `v1.22` thats fine. And new clusters can additional use the `v1.22.2` format.

Skarlso:
>Although thinking about it the regex is backwards compatible so if someone already has v1.22 thats fine. And new clusters can additional use the v1.22.2 format.

Yep :nod:.

richardcase:
I'll leave this to @sedefsavas to approve:

/assign sedefsavas

PR #3683, opened 2022-08-22
allow annotation on MachinePool to set externally managed

### LINKED ISSUE #2022: AWSMachinePool does not work in combination with cluster-autoscaler
/kind bug

- Create a workload cluster with the experimental EKS Control Plane
- Create a MachinePool without specifying replicas and create associated AWSMachinePool resources.
- Install cluster-autoscaler and configure it to autoscale the autoscaling group managed by the AWSMachinePool
- Create a deployment that will trigger cluster-autoscaler to scale up the autoscaling group.

This caused the AWSMachinePool controller and the cluster-autoscaler to both attempt to manage DesiredInstances. The result being the DesiredInstances value on the ASG was alternating between the value cluster-autoscaler computed as needed and the value of `replicas: 1` coming from the defaulted value on the MachinePool object.

cluster-autoscaler should manage the `DesiredInstances` of the autoscaling group and the AWSMachinePool controller should ignore changes to the `replicas` and `DesiredInstances` fields.

Even though `replicas` is an optional field on the MachinePool resource, it defaults to 1, so leaving the value unspecified is insufficient to prevent the AWSMachinePool controller from attempting to managed the DesiredInstance count.

**Environment:**

- Cluster-api-provider-aws version: Commit: 3338cd4
- Kubernetes version: (use `kubectl version`): v.1.17.9
- OS (e.g. from `/etc/os-release`): Amazon Linux 2

mweibel:
**What type of PR is this?**
/kind feature

As a cluster operator, I can mark a MachinePool with the annotation `cluster.x-k8s.io/replicas-managed-by-autoscaler` to indicate it should not enforce the `MachinePool.replicas` field but instead reconcile it to match the ASGs desiredCapacity field. This allows me to use an external autoscaler which does not come into conflict with the MachinePool reconciliation loop. 

Fixes #2022

The annotation is the same as in kubernetes-sigs/cluster-api-provider-azure#2444.

Unfortunately, I did not yet manage to make the added test in `awsmachinepool_controller_test.go` pass. This is also why the PR is marked as WIP/Draft. The test fails with:

```
I0822 10:22:31.013336   20405 awsmachinepool_controller.go:256]  "msg"="Setting MachinePool replicas to ASG DesiredCapacity" "external"=1 "local"=0
    awsmachinepool_controller_test.go:313: 
        Expected
            <errors.aggregate | len:1, cap:1>: [
                <*meta.NoKindMatchError | 0x14000263ec0>{
                    GroupKind: {
                        Group: "cluster.x-k8s.io",
                        Kind: "MachinePool",
                    },
                    SearchedVersions: ["v1beta1"],
                },
            ]
        to be nil
```

I believe the issue is that the API client does not know about the MachinePool CRD. I added a fake machinepool CRD just like the one for Machines/Clusters but it did not resolve the issue. Would be great if somebody could have a look at it. 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/docs/book/src/topics/machinepools.md
+++ b/docs/book/src/topics/machinepools.md
@@ -105,3 +105,46 @@ spec:
       kubeletExtraArgs:
         cloud-provider: aws
 ```
+
+## Autoscaling
+
+[`cluster-autoscaler`](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) can be used to scale MachinePools up and down.
+Two providers are possible to use with CAPA MachinePools: `clusterapi`, or `aws`.
+
+If the `AWS` autoscaler provider is used, each MachinePool needs to have an annotation set to prevent scale up/down races between
+cluster-autoscaler and cluster-api. Example:
+
+```yaml
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: MachinePool
+metadata:
+  name: capa-mp-0
+  annotations:
+    cluster.x-k8s.io/replicas-managed-by: "external-autoscaler"
+spec:
+  clusterName: capa
+  replicas: 2
+  template:
+    spec:
+      bootstrap:
+        configRef:
+          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
+          kind: KubeadmConfig
+          name: capa-mp-0
+      clusterName: capa
+      infrastructureRef:
+        apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
+        kind: AWSMachinePool
+        name: capa-mp-0
+      version: v1.24.0
+```
+
+When using GitOps, make sure to ignore differences in `spec.replicas` on MachinePools. Example when using ArgoCD:
+
+```yaml
+  ignoreDifferences:
+    - group: cluster.x-k8s.io
+      kind: MachinePool
+      jsonPointers:
+        - /spec/replicas
+```
--- a/pkg/cloud/scope/machinepool.go
+++ b/pkg/cloud/scope/machinepool.go
@@ -43,11 +43,27 @@ import (
 	"sigs.k8s.io/cluster-api/util/patch"
 )
 
+const (
+	// ReplicasManagedByAnnotation is an annotation that indicates external (non-Cluster API) management of infra scaling.
+	// The practical effect of this is that the capi "replica" count is derived from the number of observed infra machines,
+	// instead of being a source of truth for eventual consistency.
+	//
+	// N.B. this is to be replaced by a direct reference to CAPI once https://github.com/kubernetes-sigs/cluster-api/pull/7107 is meged.
+	ReplicasManagedByAnnotation = "cluster.x-k8s.io/replicas-managed-by"
+
+	// ExternalAutoscalerReplicasManagedByAnnotationValue is used with the "cluster.x-k8s.io/replicas-managed-by" annotation
+	// to indicate an external autoscaler enforces replica count.
+	//
+	// N.B. this is to be replaced by a direct reference to CAPI once https://github.com/kubernetes-sigs/cluster-api/pull/7107 is meged.
+	ExternalAutoscalerReplicasManagedByAnnotationValue = "external-autoscaler"
+)
+
 // MachinePoolScope defines a scope defined around a machine and its cluster.
 type MachinePoolScope struct {
 	logr.Logger
 	client.Client
-	patchHelper *patch.Helper
+	patchHelper                *patch.Helper
+	capiMachinePoolPatchHelper *patch.Helper
 
 	Cluster        *clusterv1.Cluster
 	MachinePool    *expclusterv1.MachinePool
@@ -98,15 +114,20 @@ func NewMachinePoolScope(params MachinePoolScopeParams) (*MachinePoolScope, erro
 		params.Logger = &log
 	}
 
-	helper, err := patch.NewHelper(params.AWSMachinePool, params.Client)
+	ampHelper, err := patch.NewHelper(params.AWSMachinePool, params.Client)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to init patch helper")
+		return nil, errors.Wrap(err, "failed to init AWSMachinePool patch helper")
+	}
+	mpHelper, err := patch.NewHelper(params.MachinePool, params.Client)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to init MachinePool patch helper")
 	}
 
 	return &MachinePoolScope{
-		Logger:      *params.Logger,
-		Client:      params.Client,
-		patchHelper: helper,
+		Logger:                     *params.Logger,
+		Client:                     params.Client,
+		patchHelper:                ampHelper,
+		capiMachinePoolPatchHelper: mpHelper,
 
 		Cluster:        params.Cluster,
 		MachinePool:    params.MachinePool,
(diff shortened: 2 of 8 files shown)


sedefsavas:
There is a related PR to this in CAPI:
https://github.com/kubernetes-sigs/cluster-api/pull/7107

mweibel:
@sedefsavas yes saw that one, thanks. Once that is merged and released, this code could reuse the const from CAPI. 

I still didn't figure out the issue with the failing test. @richardcase had a look at it IIRC but not sure what the state is.

mweibel:
I'm trying every now and then when I find some time to figure out the reason why the MachinePool Kind is not found. I'd love some help on this topic :)

richardcase:
I had a quick look at this and i also couldn't work out why it was failing.

controllers/suite_test.go
```diff
@@ -45,6 +47,8 @@ func TestMain(m *testing.M) {
 func setup() {
 	utilruntime.Must(infrav1.AddToScheme(scheme.Scheme))
 	utilruntime.Must(clusterv1.AddToScheme(scheme.Scheme))
+	utilruntime.Must(expclusterv1.AddToScheme(scheme))
```
Skarlso, line 50: Shouldn't this be `scheme.Scheme`? Maybe that's what's causing issues.

controllers/suite_test.go
```diff
@@ -45,6 +47,8 @@ func TestMain(m *testing.M) {
...
+	utilruntime.Must(expinfrav1.AddToScheme(scheme.Scheme))
```
Skarlso, line 51: Or this one be also `scheme`. :D I'm actually unsure. :D

controllers/suite_test.go
Skarlso, line 50: Yep, it's this. This is actually a compile time failure for me if I check this out.

richardcase: I have it as `scheme.Scheme` and i get the `meta.NoKindMatchError` error. I'm stumped

richardcase:
@mweibel - i've got to the bottom of the `NoKindMatchError` error you are seeing. I am going to make a small PR later today. Once that merges you will need to rebase.

richardcase:
That change has merged. So if you rebase you should be able to get past the `NoKindMatchError` error.

mweibel:
@richardcase oh - that change makes sense :+1: thanks for investigating and fixing it! Will rebase as soon as I'm able (currently on vacation)

controllers/suite_test.go
mweibel: possibly - sorry, my second commit was actually just WIP garbage commit where I tried to find the cause. Will revert those unnecessary changes.

pkg/cloud/scope/machinepool.go
```diff
@@ -43,11 +43,14 @@ import (
...
 
+const ReplicasManagedByAutoscalerAnnotation = "cluster.x-k8s.io/replicas-managed-by-autoscaler"
+
 // MachinePoolScope defines a scope defined around a machine and its cluster.
 type MachinePoolScope struct {
 	logr.Logger
 	client.Client
-	patchHelper *patch.Helper
+	patchHelper            *patch.Helper
+	MachinePoolPatchHelper *patch.Helper
```
richardcase, line 53: Ideally we'd want to not export this and provide a `PatchMachinePool` function on the scope. But i'm not so concerned about this

richardcase:
e2e tests are passing :+1:

Skarlso:
Ah, that's a good catch. We actually ran into a similar problem with one of the issues. That was patching some template which ended up being out of sync with what you basically had locally. So once you applied again, the changes were lost.

Interesting thought. I don't see an option out of that other than having an unmanaged field perhaps.

mweibel:
@richardcase that's a good catch indeed. We're using ArgoCD where you can ignore differences. A user with a GitOps solution would need to do that:
```yaml
  ignoreDifferences:
    - group: cluster.x-k8s.io
      kind: MachinePool
      jsonPointers:
        - /spec/replicas
```

I agree that this is not beautiful. One way this could be different when CAPI would provide e.g. a `status.replicas` field which overrules `spec.replicas` field in case it's set, so `spec.replicas` would only be considered for the first reconciliation. Not sure if that's a good approach though.

pkg/cloud/scope/machinepool.go
mweibel: I wondered if I should do it that way or provide a specific function on the scope. I don't mind changing this.

dthorsen:
An even better option is to completely omit the replicas field from your manifest.

richardcase:
> An even better option is to completely omit the replicas field from your manifest.

@dthorsen - i think thats probably the best solution. 

What do you think to the following:

- Change the MachinePool update logic that in the `AWSMachinePool` reconciler so that it only updates the replicas if its not nil
- Add a documentation for externally managed replicas that includes:
  - The option/recommendation of not setting replicas on MachinePool when using externally managed
  - The sample above for ignoring the field with argo

Skarlso:
> An even better option is to completely omit the replicas field from your manifest.

Wouldn't that still fail? Because it's nil locally, but set upstream?

richardcase:
> Wouldn't that still fail? Because it's nil locally, but set upstream?

Its omit it from the upstream `MachinePool`

pkg/cloud/scope/machinepool.go
```diff
@@ -43,11 +43,14 @@ import (
 	"sigs.k8s.io/cluster-api/util/patch"
 )
 
+const ReplicasManagedByAutoscalerAnnotation = "cluster.x-k8s.io/replicas-managed-by-autoscaler"
```
AverageMarcus, line 46: Am I correct that we'd import this value from CAPI once [this PR](https://github.com/kubernetes-sigs/cluster-api/pull/7107) is merged rather than hardcoding it in here?

mweibel:
> > Wouldn't that still fail? Because it's nil locally, but set upstream?
> 
> Its omit it from the upstream `MachinePool`

@richardcase looking at the webhook code, I believe this wouldn't work since the webhook would set it anyway:
https://github.com/kubernetes-sigs/cluster-api/blob/main/exp/api/v1beta1/machinepool_webhook.go#L54..L56

or do I miss something there?

pkg/cloud/scope/machinepool.go
mweibel: yes, I think that would make sense.

pkg/cloud/scope/machinepool.go
mweibel: adjusted :+1:

mweibel:
I adjusted the patch helper so it's a function on the MachinePool scope instead of exposing the patch helper itself. 

@richardcase how to handle MachinePool.replicas: I don't believe there is currently a better option. I added some documentation which might help in this case. 

Would be great to have another review on this :+1:

docs/book/src/topics/machinepools.md
```diff
@@ -104,4 +104,47 @@ spec:
...
       kubeletExtraArgs:
         cloud-provider: aws
-```
\ No newline at end of file
+```
+
+## Autoscaling
+
+[`cluster-autoscaler`](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler) can be used to scale MachinePools up and down.
+Two providers are possible to use with CAPZ MachinePools: `clusterapi`, or `aws`.
```
AverageMarcus, line 112:
```suggestion
Two providers are possible to use with CAPA MachinePools: `clusterapi`, or `aws`.
```

AverageMarcus (changes requested):

docs/book/src/topics/machinepools.md
Ankitasw: Good catch 😄

mweibel: oh 😄 sorry about that one! good catch indeed.

mweibel:
docs typo is fixed 👍

AverageMarcus (approved):

mweibel:
FTR this PR has been updated with the updated annotation key/value from kubernetes-sigs/cluster-api#7107.

Issue #3679, opened 2022-08-19
Support autoscaling from zero


com6056:
/kind feature

**Describe the solution you'd like**
[A clear and concise description of what you want to happen.]

Add support for autoscaling from zero defined in https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210310-opt-in-autoscaling-from-zero.md

Just implemented upstream in https://github.com/kubernetes/autoscaler/pull/4840

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

Skarlso:
/triage accepted

I saw this and wanted to do it! :)

Skarlso:
Ping @richardcase, @sedefsavas and @Ankitasw for second opinion. :)

Skarlso:
Also, now I'm wondering, when and how do we usually do triage-es?

Skarlso:
My understanding from reading this is as follows:

- Create a `Status` field in AWSMachineTemplate and reconcile it
- Based on those templates, create annotations on the machine deployment following the names outlined within
- ?
- Profit.

I'm not sure what else is required. Do we just have to add the annotations based on our spec fields and reconcile them back into the status field? Is that it? :)

Skarlso:
And, ofc, preventing CAPA itself from creating a machine instance where it expects one to be, if these fields are defined.

Skarlso:
Aha
```go
	infraref, found, err := unstructured.NestedStringMap(r.unstructured.Object, "spec", "template", "spec", "infrastructureRef")
```

Skarlso:
Answer from cluster-api channel:

Yes, it's enough to define the status. 
Reference implementation can be found [here](https://github.com/kubernetes-sigs/cluster-api-provider-kubemark/pull/30).

PR #3684, opened 2022-08-22
Add scaling from 0 support

Skarlso:
**What type of PR is this?**

/kind feature

Adds scaling from 0 to CAPA.

Proposal -> https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210310-opt-in-autoscaling-from-zero.md#upgrade-strategy
Reference implementation -> https://github.com/kubernetes-sigs/cluster-api-provider-kubemark/pull/30/files

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3679

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests
  ~~- I will try adding one!~~
    - I decided not to considering how convoluted the setup gets


Skarlso:
I'm reluctant to auto-generate anything. That would be magic and implicit. Or do you mean, that a user should set something like, enable scale from 0 and that should autofill some data?

com6056:
> I'm reluctant to auto-generate anything. That would be magic and implicit. Or do you mean, that a user should set something like, enable scale from 0 and that should autofill some data?

Ya, the user should be able to enable this feature and by default we could populate the resource quantities with the values provided by the AWS API. Users could override that if they want, but seems like a pretty nice convenience and most likely what most folks would need and remove a bunch of unnecessary hardcoded resource quantities.

Skarlso:
>Ya, the user should be able to enable this feature and by default we could populate the resource quantities with the values provided by the AWS API. Users could override that if they want, but seems like a pretty nice convenience and most likely what most folks would need, and at the same time remove the need for a bunch of unnecessary hardcoded resource quantities that we can just get programmatically.

This would cause a diff between the applied and existing templates which a tool like Flux would pick up on and try constantly reconciling. We can't just change the spec and add values to it in flight.

>cluster-autoscaler already exposes the annotations that let you specify everything manually, so just passing this through as a field in the MachineTemplate isn't adding much benefit.

I'm not just passing through as a field, right? This is the first option in the proposal, which CAN be overwritten via annotations.

com6056:
> >Ya, the user should be able to enable this feature and by default we could populate the resource quantities with the values provided by the AWS API. Users could override that if they want, but seems like a pretty nice convenience and most likely what most folks would need, and at the same time remove the need for a bunch of unnecessary hardcoded resource quantities that we can just get programmatically.
> 
> 
> 
> This would cause a diff between the applied and existing templates which a tool like Flux would pick up on and try constantly reconciling. We can't just change the spec and add values to it in flight. Even if you don't change it, the diff will be there in the end applied Object. Which, upon re-applying, would again configure the resource.
> 
> 
> 
> >cluster-autoscaler already exposes the annotations that let you specify everything manually, so just passing this through as a field in the MachineTemplate isn't adding much benefit.
> 
> 
> 
> I'm not just passing through as a field, right? This is the first option in the proposal, which CAN be overwritten via annotations.

I believe the proposal is loosely suggesting that we use the information available in our infrastructure provider to power these fields.

> Each infrastructure provider will be responsible for determining the best implementation for adding the status field based on the information available on their platform.

If this feature is opt-in, the diff created will be expected and after applied once, won't come up again.

Skarlso:
>I believe the proposal is loosely suggesting that we use the information available in our infrastructure provider to power these fields.

It is a loose suggestion in what terms? You either provide a status field or use annotations. There are no suggestions here.

>If this feature is opt-in, the diff created will be expected and after applied once, won't come up again.

So, whenever you have an edit in a template and apply it and edit it before you apply it, than there is a diff between the real template and what has been applied. And even if you do the same all the time, there will be a `configured` event instead of `unchanged` and some tools like flux will pick up on diffs when applying said thing to the cluster over and over again unless there's no change.

But I digress. Regardless, I am against magic. This can be a very simplistic PR in which we provide people with the option to define these fields and it should work from there on. Anything implicit is magic, and a convenience feature which can be added later if there is demand for it.

This is my opinion. I refer to the judgment of my peers @richardcase and @sedefsavas.

Skarlso:
To be more clear. I'm not completely against doing that. I'm just saying it will take some more time and more code which will be harder to review and get in. I'm suggesting doing that in smaller increments. :) If that makes sense.

com6056:
> To be more clear. I'm not completely against doing that. I'm just saying it will take some more time and more code which will be harder to review and get in. I'm suggesting doing that in smaller increments. :) If that makes sense.

Oh ya agreed! Sorry didn't mean to imply this PR wasn't good, just suggesting what a good end goal would be. Seems like we're on the same page though 🙌 Thanks so much for tackling this first pass!

Skarlso:
No problem! I first, like to get this working at all. :D I'm testing it right now, but I'm yet to understand under what circumstances and how this would work.... Would I need to set the replica count to 0? But that makes no sense because then we don't get nodes. :D 

Anyways. I'll figure it out first, then I can make sure I understand how to improve it. :) 🙇

Skarlso:
IITTTT WORKIIINGGG

```
➜  cluster k get nodes --kubeconfig workload.kubeconfig
No resources found
➜  cluster k get nodes --kubeconfig workload.kubeconfig
NAME                                           STATUS   ROLES    AGE   VERSION
ip-10-0-81-112.eu-central-1.compute.internal   Ready    <none>   46s   v1.22.12-eks-ba74326
```

I will write up testing and usage guide. :D

Skarlso:
Booom more nodes as there are a lot of pods:

```
k get nodes --kubeconfig workload.kubeconfig
NAME                                            STATUS   ROLES    AGE     VERSION
ip-10-0-112-36.eu-central-1.compute.internal    Ready    <none>   3m12s   v1.22.12-eks-ba74326
ip-10-0-118-146.eu-central-1.compute.internal   Ready    <none>   2m23s   v1.22.12-eks-ba74326
ip-10-0-120-50.eu-central-1.compute.internal    Ready    <none>   2m4s    v1.22.12-eks-ba74326
ip-10-0-81-112.eu-central-1.compute.internal    Ready    <none>   4m24s   v1.22.12-eks-ba74326
ip-10-0-97-82.eu-central-1.compute.internal     Ready    <none>   2m24s   v1.22.12-eks-ba74326
```

Skarlso:
Scaling down to zero also worked!!!!!!! :)))))

```
k get nodes --kubeconfig workload.kubeconfig
No resources found

```

Skarlso:
I'll write up a book on how to scale to 0 using the cluster-autoscaler.

Skarlso:
The first pass is now done for this. The next pass can take on the above-discussed convenience feature.
Adding @richardcase for interest. :)

controllers/awsmachine_controller.go
```diff
@@ -137,6 +137,8 @@ func (r *AWSMachineReconciler) getObjectStoreService(scope scope.S3Scope) servic
 // +kubebuilder:rbac:groups="",resources=secrets;,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=namespaces,verbs=get;list;watch
 // +kubebuilder:rbac:groups="",resources=events,verbs=get;list;watch;create;update;patch
+// +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=kubemarkmachinetemplates,verbs=get;list;watch;create;update;patch;delete
```
sedefsavas, line 140: Why do we need these permissions?

Skarlso: Yeah, this is a requirement for the autoscaler indeed. I guess we can do that, but that just adds an extra step for the user which they will need to automate instead of us doing this for them which results in a better UX IMHO.

com6056: I think this should be `awsmachinetemplates` instead of `kubemarkmachinetemplates`

Skarlso: Hah, that's actually a very good point. :D

com6056: Pretty sure it is `cluster-autoscaler` that needs this permission, not `cluster-api-provider-aws`, so don't think this is needed here (unless I'm misunderstanding what this does)?

sedefsavas: If these permissions are not needed here, then after this is removed, LGTM, we can even try to add this feature to the upcoming v1beta1 release as the changes are not big.

Skarlso: Sounds good. I'll remove it!

Skarlso: So, the reason why I added this is that the reference implementation was also adding these RBAC rules and the proposal states that the service account associated with the cluster-autoscaler must have these permissions.

I'll update the document that this now something that the user has to do.

Skarlso: Ah! I think we implicitly already allow this through various `+kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io` rbacs. :D That's why it was working regardless. Anyways, I updated the docs and the code. Thanks you two. :)

richardcase:
@Skarlso - is there an easy way we can get this covered by an e2e test?

Skarlso:
> @Skarlso - is there an easy way we can get this covered by an e2e test?

I thought about this... is there a way to install cluster-autoscaler automatically? If that is done, maybe I can do it. But it won't be nice. :D

richardcase:
> I thought about this... is there a way to install cluster-autoscaler automatically? If that is done, maybe I can do it. But it won't be nice or easy. :D

We could do that via CRS or after the cluster has been created. We have some examples of this.

Shall we create an issue to follow-up on adding an e2e test?

Skarlso:
>Shall we create an issue to follow-up on adding an e2e test?

Yeah, sounds good. This is the base support anyways. A further ticket will be about what @com6056 is suggesting. So I'm okay with a follow up.

richardcase:
Created #3704 to follow-up creating a e2e test for this.

richardcase:
/lgtm

For final approval:

/assign sedefsavas

PR #3688, opened 2022-08-23
Allow for Self-Managed VPC with a Secondary Subnet for Pods

luthermonson:
**What type of PR is this?**
/kind bug

This is quite the logic puzzle so apologies in advance and try and bear with me. I found a situation when trying to configure the VPC/CNI you are unable to generate ENIConfigs for secondary subnets for the pods. CAPA is making the assumption that if you want a secondary subnet you are using SecondaryCIDRBlock. 

This is the core logic
* ENIConfig creation is gated behind a not nil SecondaryCIDRBlock
* ENIConfig creation only creates ENIConfigs for subnets tagged `sigs.k8s.io/cluster-api-provider-aws/association=secondary`
* subnet creation is gated behind a VPC ID being not present

The only scenario you can get ENIConfigs for an unmanaged VPC is by setting the VPC ID, setting SecondaryCIDRBlock (which is ignored) to one of the two allowed ranges and tagging your subnets.

To fix I propose we only make ENIConfigs if we have secondary subnets. This will leave the managed VPCs with the same logic around SecondaryCIDRBlock while allowing unmanaged VPC users to tag and choose which subnets CAPA will create ENIConfigs.

Fixes #

This PR contains a cherry-pick of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3681 because it has the updated networking docs so only look at 946c0b90dfdf69732209bf3af2e245f9d95f8e05. It also seems I wrote wrote the cidr parser test for a /16 instead of a /10

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/internal/cidr/cidr_test.go
+++ b/pkg/internal/cidr/cidr_test.go
@@ -34,20 +34,20 @@ func TestSplitIntoSubnetsIPv4(t *testing.T) {
 		{
 			// https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-eks-now-supports-additional-vpc-cidr-blocks/
 			name:        "default secondary cidr block configuration with primary cidr",
-			cidrblock:   "100.64.0.0/16",
+			cidrblock:   "100.64.0.0/10",
 			subnetcount: 3,
 			expected: []*net.IPNet{
 				{
 					IP:   net.IPv4(100, 64, 0, 0).To4(),
-					Mask: net.IPv4Mask(255, 255, 192, 0),
+					Mask: net.IPv4Mask(255, 240, 0, 0),
 				},
 				{
-					IP:   net.IPv4(100, 64, 64, 0).To4(),
-					Mask: net.IPv4Mask(255, 255, 192, 0),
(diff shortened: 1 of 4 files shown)


pkg/internal/cidr/cidr_test.go
```diff
@@ -34,20 +34,20 @@ func TestSplitIntoSubnetsIPv4(t *testing.T) {
 		{
 			// https://aws.amazon.com/about-aws/whats-new/2018/10/amazon-eks-now-supports-additional-vpc-cidr-blocks/
 			name:        "default secondary cidr block configuration with primary cidr",
-			cidrblock:   "100.64.0.0/16",
+			cidrblock:   "100.64.0.0/10",
```
Skarlso, line 37: Neat. Nice find.... Technically, those values could be anything. But it's nice to reflect the documentation.

lmasiero (approved):

Ankitasw:
@luthermonson could you please fix unit tests `TestReconcileCniVpcCniValues` ?

luthermonson:
@Ankitasw all done

PR #3693, opened 2022-08-26
Bump golangci-lint to v1.49.0

Prajyot-Parab:
Signed-off-by: Prajyot-Parab <prajyot.parab2@ibm.com>

**What type of PR is this?**
/kind cleanup

- Bump golangci-lint to v1.49.0
- Fix lint error for usage of net/http (gosec)

The following linters are deprecated in golangci-lint:

structcheck - Finds unused struct fields
deadcode - Finds unused code
varcheck - Finds unused global variables and constants
ifshort - Go linter that checks if your code uses short syntax for if-statements whenever possible. (Buggy)

Remove these linters as they're no longer maintained. (Ref -https://github.com/golangci/golangci-lint/issues/1841)

+ disable interfacebloat linter to avoid setting max limit on number of methods in an interface.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


main.go
```diff
@@ -130,7 +130,14 @@ func main() {
 	if profilerAddress != "" {
 		setupLog.Info("Profiler listening for requests", "profiler-address", profilerAddress)
 		go func() {
-			setupLog.Error(http.ListenAndServe(profilerAddress, nil), "listen and serve error")
+			server := &http.Server{
```
Skarlso, line 133: Please remove this code change or change. You are bumping the golangci-lint version. To that effect, there should be no surprise code-changes.

If you have a refactoring in mind, I suggest opening a separate issue for the refactoring alone so we can review and test it separately. :)

.golangci.yml
```diff
@@ -56,9 +61,6 @@ linters-settings:
     rules:
       - name: exported
         severity: warning
-  ifshort:
```
Ankitasw, line 59: this removal is not necessary.

Skarlso: `ifshort` has been deprecated and is considered buggy apparently. And since it's removed above, I suspect, that this setting no longer makes sense?

Ankitasw: Oh i missed to see ifshort in disabled linters list, thanks 😄

main.go
Prajyot-Parab: Noted, will fix it as part of a separate PR.

Prajyot-Parab: @Skarlso I can see failure in test prow job after removing above piece of code, as it was part of v1.49.0 compatibility required refactoring.

Prajyot-Parab: Any suggestions on how to move forward on this one?

Skarlso: Oh wow. What the heck. :D I'll take a look later on.

Skarlso: @Prajyot-Parab Thanks, this didn't occur to me reading the diff and the description! I see.

How did you do the update? If I do a `go get -u github.com/golangci/golangci-lint` I get the new version but I don't get this error locally.

Skarlso: Never mind, that actually fails because go mod tidy wants to remove golangci-lint. :D So, what did you do to update it? :D

Prajyot-Parab:
```
			server := &http.Server{
 				Addr:              profilerAddress,
 				ReadHeaderTimeout: 3 * time.Second,
 			}

 			err := server.ListenAndServe()
 			if err != nil {
 				setupLog.Error(err, "listen and serve error")
 			}
```
 			
 This patch should fix it. Should I go ahead and re-add it? (or are there any concerns from your end)

Prajyot-Parab: > Never mind, that actually fails because go mod tidy wants to remove golangci-lint. :D So, what did you do to update it? :D

`make lint` should rebuild the latest version binary and use it.

Skarlso: Hang on a sec. :) The question now is, how did you update golangci-lint? And why does that result in a need for updating a version of the http.Server? What is the connection there?

Skarlso: That doesn't make sense to me. Why would `make lint` update golangci-lint? It just runs lint.

Prajyot-Parab: 1. I bumped the version in go.mod of hack tools (golangci-lint)
2. I ran `make lint` which would use 1.49.0
3. which resulted in this new linting error -
```
main.go:133:19: G114: Use of net/http serve function that has no support for setting timeouts (gosec)
                        setupLog.Error(http.ListenAndServe(profilerAddress, nil), "listen and serve error")
```
4. Added above patch to fix it.
5. Ran `make lint `again no lint issue found.

Skarlso: For the future, please describe "seemingly" unrelated code changes. I know you wrote this:

>Fix lint error for usage of net/http (gosec)

But this doesn't explain the why. :)

And for me, I should rather ask what prompted a code change. :D Instead of asking to remove it. :)))

Prajyot-Parab: Done.

Skarlso: Thanks! :)) That was a journey. :D

dlipovetsky: _nit, non-blocking_: I think it would be nice to put this change a separate commit, and have the message explain why it's needed, i.e. something like "Use timeouts to avoid leaking file descriptors. Explained in https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/. Caught by the gosec linter rule introduced in https://github.com/securego/gosec/commit/0c8e63ed86f8fc2da42787e0a653114490078d6c"

Skarlso: But the build fails if you update the golangci-lint. 😂

PR #3706, opened 2022-09-05
refactor: add capi crds to testenv

richardcase:
**What type of PR is this?**

/kind support

Changed the testenv creation so that it explicitly adds the CRDs from
CAPI. This has been copied for CAPZ :) By making this change it caused
some tests to fail as the `Machine` definitions weren't valid and so
these have been updated.

Relates #3683 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/test/helpers/envtest.go
+++ b/test/helpers/envtest.go
@@ -19,10 +19,12 @@ package helpers
 import (
 	"context"
 	"fmt"
+	"go/build"
 	"net"
 	"os"
 	"path"
 	"path/filepath"
+	"regexp"
 	goruntime "runtime"
 	"strconv"
 	"strings"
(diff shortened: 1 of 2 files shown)


test/helpers/envtest.go
```diff
@@ -278,3 +280,32 @@ func (t *TestEnvironment) Stop() error {
...
+			clusterAPIVersion = matches[2]
+		}
+	}
+
+	if clusterAPIVersion == "" {
+		return ""
+	}
+
+	gopath := envOr("GOPATH", build.Default.GOPATH)
+	return filepath.Join(gopath, "pkg", "mod", "sigs.k8s.io", fmt.Sprintf("cluster-api@v%s", clusterAPIVersion), "config", "crd", "bases")
```
Skarlso, line 303: I suggest taking a look at [runtime.Caller](https://pkg.go.dev/runtime#Caller). I think it returns just what you need here.

Skarlso: Although that might not help since you need the stuff from config crd bases. And you would have to manipulate the path nevertheless. But maybe it helps you eliminate the need for GOPATH check. 🤷  I leave it up to you. :D

richardcase: This copied from CAPZ...i didn't change it. But we do already use `runtime.Caller` to get the root: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/e0fc4de3bc5f4008b214a8cab6c72a5fe51c6960/test/helpers/envtest.go#L85

Skarlso: Oh nice. :D ok. :)

Issue #3698, opened 2022-09-01
v1.5.0 fails to reconcile EKS cluster when cluster name is not same as eks control plane name


Jacobious52:
/kind bug

**What steps did you take and what happened:**

Starting in 1.5.0 with https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3598/files# if the EKS controlplane name is different the cluster name, reconciliation fails with `checking owner of bike is bike-cluster: **%!w(<nil>)**`

For an example:

Cluster object is called `bike-cluster`
```yaml
apiVersion: cluster.x-k8s.io/v1beta1
kind: Cluster
metadata:
  name: bike-cluster
```

eksClusterName is `bike`
```yaml
apiVersion: controlplane.cluster.x-k8s.io/v1beta1
kind: AWSManagedControlPlane
metadata:
  name: bike-control-plane
spec:
  eksClusterName: bike
```

This causes reconciliation to fail for the cluster
```
E0831 06:42:46.562913       1 controller.go:317] controller/awsmanagedcontrolplane "msg"="Reconciler error" "error"="failed to reconcile control plane for AWSManagedControlPlane clusters/bike-control-plane: checking owner of bike is bike-cluster: %!w(<nil>)" "name"="bike-control-plane" "namespace"="clusters" "reconciler group"="controlplane.cluster.x-k8s.io" "reconciler kind"="AWSManagedControlPlane"
```

A temporary fix for end users is to manually add the cluster tag to the control plane resource, so that 
```
  additionalTags:
    "kubernetes.io/cluster/bike": "owned"
```

So that the control plane has both Cluster.Metadata.Name and AWSControlPlane.Spec.EKSClusterName`

![image](https://user-images.githubusercontent.com/12645162/187810315-d021d6c2-331a-4f39-a8ce-51bc3d55d03d.png)

**What did you expect to happen:**

The controller to either use Cluster Name or EKSClusterName consistently. 

**Anything else you would like to add:**

There is also a confusion about which name is correct to use in some places like
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3b32ca01169ab2cc1510398dd82b0669b9002db8/pkg/cloud/services/iamauth/reconcile.go#L33
and 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3b32ca01169ab2cc1510398dd82b0669b9002db8/pkg/cloud/services/iamauth/reconcile.go#L80

```
[...]

PR #3713, opened 2022-09-07
A small fix to the logging output of CAPA regarding cluster names

Skarlso:
**What type of PR is this?**

/kind bug

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3698

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


Ankitasw:
I think [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3573) it was changed sometime back, but not sure if it was tested.
Good catch though @Skarlso 
/lgtm

sedefsavas:
This is a critical bug not just a logging fix, right?
Do we want to cherry pick this?

sedefsavas:
So, here the tag with correct EKS cluster name is added and in the issue, it is also said this fixes the issue when manually done.
What am I missing?

Skarlso:
I'm not sure because `createCluster` is called like this:

```go
cluster, err = s.createCluster(eksClusterName)
```

Where `eksClusterName` is defined a few lines above as:
```go
eksClusterName := s.scope.KubernetesClusterName()
```

So technically, unless the cluster name in the scope somehow changed, I don't think this should have fixed the issue. So either this is just a convenient logging fix for the output and usage of an already declared function parameter, or it actually "fixed" a subtle bug where the scope cluster changed somehow between two function calls on the same scope. Which is weird. :)

Skarlso:
As a sidenote... It did fix _something_ because the test had to be changed to include the proper tag which was missed previously.

Skarlso:
@richardcase The test passed. :)))

PR #3720, opened 2022-09-12
v1beta2 APIs

sedefsavas:
**What this PR does / why we need it**:
This PR adds new v1beta2 API version and it is the exact copy of v1beta1.
After this PR is merged, v1beta2 changes will come as follow up PRs.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #3735, opened 2022-09-14
normalize oidc configs to string values for comparison

### LINKED ISSUE #3734: EKS Identity Provider Attach/Detach Loop
/kind bug

When adding an OIDC Provider you end up with something that looks like this in the EKS cluster 

Screen Shot 2022-09-02 at 7 05 59 AM (1)

I know the issue and working on a fix already but writing an issue to document the behavior for posterity. What you will see is the reconcile loops will alternate from attaching and detaching the identity provider in about every 20min.  This is coming from a comparison in the identiy provider plan on the awsmanagedcontrolplane scope vs. the upstream (current vs desired). CAPA is converting these to a local [type](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/eks/identityprovider/types.go#L27-L39) and running a comparison against the two. Unfortunately, a pointer comparison of a nil pointer and an empty string ends up happening due the results coming from two different api sources.

This pic shows a debugger showing the plan Create where the disassociation happens with GroupPrefix being nil in one and a pointer to an empty string in the other.

e0adc92a1598dfb589b2d192a00e3129

and here is an example of this situation and the cmp.Equal call returning false to prove my theory: https://go.dev/play/p/8iWiT1IqNvj

I'm going to try and remove the string pointers from the local struct and convert everything into their values and make nil string pointers into empty strings.

Shouldn't infinitely switch between associating and disassociating the auth provider.

I have an active test where this is failing and will show proof it goes away if this all works out.

**Environment:**

luthermonson:
**What type of PR is this?**
/kind bug

Normalizes OIDC configs into string values for easier comparison for reconciliation.

Fixes #3734 

I have a test running with a debugger right now which will prove this fixed the issue so this is just ready for review and ill post screenshots of it being fixed when I have them.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/eks/identity_provider.go
+++ b/pkg/cloud/services/eks/identity_provider.go
@@ -49,14 +49,20 @@ func (s *Service) reconcileIdentityProvider(ctx context.Context) error {
 	}
 
 	s.scope.V(2).Info("creating oidc provider plan", "desired", desired, "current", current)
+	procedures, err := identityprovider.
+		NewPlan(clusterName, current, desired, s.EKSClient, s.scope.Logger).
+		Create(ctx)
 
-	identityProviderPlan := identityprovider.NewPlan(clusterName, current, desired, s.EKSClient, s.scope.Logger)
-
-	procedures, err := identityProviderPlan.Create(ctx)
 	if err != nil {
 		s.scope.Error(err, "failed creating eks identity provider plan")
 		return fmt.Errorf("creating eks identity provider plan: %w", err)
 	}
+
+	// nothing will be done, we can leave
+	if len(procedures) == 0 {
+		return nil
+	}
+
 	s.scope.V(2).Info("computed EKS identity provider plan", "numprocs", len(procedures))
 
 	// Perform required operations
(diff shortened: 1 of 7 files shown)


luthermonson:
Confirmed, this now returns true in the IsEqual func

600503d02dd63e11921f556faea41f7b

luthermonson:
oh... fixing the tests sorry about that

luthermonson:
pic of it working

600503d02dd63e11921f556faea41f7b

sedefsavas:
/hold
until https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3720 is merged.
All upcoming PRs need to be merged once v1beta2 API types are in.

luthermonson:
the idp code was also patching the control plane object every reconcile loop. added additional logic in this commit https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3735/commits/533002401b62bc0ae757a3c5bb18c0f81107be53 to get out sooner if there are no procedures or if nothing changed on the idp status.

luthermonson:
the apidiff failure is a false positive, the OidcIdentityProviderConfig is an internal struct and never used in a crd and i'd like to propose we ignore it

Skarlso:
I concur.
/lgtm

pkg/cloud/services/eks/identity_provider.go
```diff
@@ -49,14 +49,20 @@ func (s *Service) reconcileIdentityProvider(ctx context.Context) error {
...
-	identityProviderPlan := identityprovider.NewPlan(clusterName, current, desired, s.EKSClient, s.scope.Logger)
-
-	procedures, err := identityProviderPlan.Create(ctx)
 	if err != nil {
 		s.scope.Error(err, "failed creating eks identity provider plan")
 		return fmt.Errorf("creating eks identity provider plan: %w", err)
 	}
+
+	// nothing will be done, we can leave
+	if len(procedures) == 0 {
```
Skarlso, line 62: This change here introduces a change in behaviour. Which might be fine, I'm just thinking aloud on this. Previously, we constantly reconciled the status even if it was the same. Which makes no sense. Why update it if the status is the same? I think this should be okay. Though I really would like to hear @dlipovetsky's opinion. :)

luthermonson: ya i added this for exactly that reason... no point to update if everything is the same. i spent a lot of time in this code >.<

luthermonson:
@lmasiero this wont pass the apidiff check. it will need to get merged with it failed, it's not actually changing any real APIs it's just an internal struct used to reconcile the oidc config

richardcase:
After review and e2e tests passing i think this is good to go. We will ignore the apidiff failure. So:

/approve

PR #3747, opened 2022-09-25
Provideridlist fix

sedefsavas:
**What type of PR is this?**

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


pkg/cloud/services/eks/nodegroup.go
```diff
@@ -591,12 +590,7 @@ func (s *NodegroupService) setStatus(ng *eks.Nodegroup) error {
 		for _, group := range groups.AutoScalingGroups {
 			replicas += int32(len(group.Instances))
 			for _, instance := range group.Instances {
-				id, err := noderefutil.NewProviderID(fmt.Sprintf("aws://%s/%s", *instance.AvailabilityZone, *instance.InstanceId))
-				if err != nil {
-					s.Error(err, "couldn't create provider ID for instance", "id", *instance.InstanceId)
-					continue
-				}
-				providerIDList = append(providerIDList, id.String())
+				providerIDList = append(providerIDList, fmt.Sprintf("aws:///%s/%s", *instance.AvailabilityZone, *instance.InstanceId))
```
richardcase, line 593: I've been thinking about this over the weekend (including the fix from the end of last week) and the `///` doesn't feel right to me. I'm thinking that we need to change to `aws://%s/%s`

sedefsavas: https://github.com/kubernetes/cloud-provider-aws/blob/v1.25.0/pkg/providers/v1/instances.go#L53

CCM sets it like this.

sedefsavas: cc @yastij

sedefsavas: How about deciding this after merging this PR as all tests passed here? We can do a beta cut after this is merged and continue improving?

Ankitasw:
FYI i have given LGTM for machinepool providerID changes, as I am the co-author of this PR, I would let @richardcase @Skarlso to also provide LGTM from their end.

Skarlso:
Ah thanks for the reminder that I still have to review this one. 🤣

main.go
```diff
@@ -114,7 +115,7 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
-	if err := logOptions.ValidateAndApply(nil); err != nil {
+	if err := logsv1.ValidateAndApply(logOptions, nil); err != nil {
```
Skarlso, line 118: Ops, thanks for the fix!

pkg/cloud/services/eks/nodegroup.go
Skarlso, line 593: Hmm, I would prefer using safe dereferences here. If, for whatever reason, AZ or InstanceID is nil, this will panic.

go.mod
```diff
@@ -28,17 +28,17 @@ require (
 	golang.org/x/crypto v0.0.0-20220411220226-7b82a4e95df4
 	golang.org/x/text v0.3.7
 	gopkg.in/yaml.v2 v2.4.0
-	k8s.io/api v0.24.2
-	k8s.io/apiextensions-apiserver v0.24.2
-	k8s.io/apimachinery v0.24.2
+	k8s.io/api v0.25.0
```
Skarlso, line 31: I gather this is not a problem, right? ( It's a minor version jump which might cause some issues with dependent libraries ).

I guess if all tests passed this shouldn't be a problem.

pkg/cloud/services/eks/nodegroup.go
richardcase: Yes lets decide afterwards. 

I think some providers used the triple slash as they didn't naturally have separate segements...maybe?

Skarlso:
LGTM me too, just two questions. :)

pkg/cloud/services/eks/nodegroup.go
Ankitasw: In any case, these two fields should not be empty as these are required parameters.
Do you think adding validations would help here? I don't strongly feel the need for validation, but I am ok either ways 😄

Skarlso: Oh, I just thought something like using `aws.String` instead of dereferencing. Nothing too fancy. :)

Skarlso: Just to prevent a possible panic. An error is much preferred I believe? :) Because a panic stops events from happening such as deletion for example.

Ankitasw: Oh so sorry i misunderstood it as validity check for dereferencing. Yes you are right, we could use aws library functions for dereferencing.
cc @sedefsavas could you please take care of this comment?

sedefsavas: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3747#discussion_r979633728

Can we target this after merging this PR? We might change the providerID string as per Richard's suggestion.

go.mod
sedefsavas: I agree since we haven't bumped our test versions yet, it should be 0.24.x and CAPI 1.2.2 is using it.

pkg/cloud/services/eks/nodegroup.go
sedefsavas: Sounds good, also whatever change we do here, we need to do on the unmanaged side as well.

go.mod
sedefsavas: Since we are using this `k8s.io/component-base/logs/api/v1`, it pulls v0.25.x k8s.io dependencies.
Reverted back to 0.25.x.

Ankitasw:
@sedefsavas @richardcase  we also need providerIDList change for managed node pools in release 1.6, isn't it?

Issue #3530, opened 2022-06-15
Tasks for adopting CAPI's Server Side Apply


pydctw:
This issue is **tracking** the list of tasks to make CAPI's SSA (Server Side Apply) to work with CAPA. 

## Why do we need this?
CAPA's `spec.network.subnets` is coauthored by CAPI and CAPA controllers when using ClusterClass. To properly manage these coauthoring slices and prevent them continuously getting patched by both controllers, CAPI is now using [Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/).
* [x] Issue: https://github.com/kubernetes-sigs/cluster-api/issues/6320
* [x] Solution: https://github.com/kubernetes-sigs/cluster-api/pull/6495

## Changes Required in CAPA
* [ ] https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3531

The following issues require **v1beta2 API version bump** as a pre-requiste.
* [ ] https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3528
* [ ] https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3536

## Other Issues to Follow
* [x] https://github.com/kubernetes-sigs/controller-tools/pull/692
  * This is needed to properly generate CRD manifests with the list markers. Currently, we are using a hack to overcome this issue.
* [x] https://github.com/kubernetes-sigs/cluster-api/issues/6650
  * There is an issue with controller metadata in logging. The log prints out a wrong controller type and kind.

## CAPA issues that will be resolved
* [ ] https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3399
* [ ] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3397

pydctw:
## PoC
While waiting for controller-tools and listMapKey issues to be resolved, did an *Initial PoC* with test purpose CRDs. This required some hacks so the result needs to be confirmed when all the tasks listed in the issue is completed.

#### Hacks
* Used []SubnetSpec, a slice, as a type for Subnets for CRD manifest generation.
```yaml
// +optional
// +listType=map
// +listMapKey=id
Subnets []SubnetSpec `json:"subnets,omitempty"`
```
* Made `subnet.id` as a required field in CRD to use as a listMapKey.

#### Scenario: BYO Infra Case
AWSClusterTemplate in ClusterClass
```yaml
[...]

Issue #3528, opened 2022-06-14
Add a required field to SubnetSpec to use as listMapKey in SSA


pydctw:
/kind feature

**Describe the solution you'd like**
While validating a fix for co-authored slice issue in ClusterClass, 
* Issue: https://github.com/kubernetes-sigs/cluster-api/issues/6320
* Fix: https://github.com/kubernetes-sigs/cluster-api/pull/6495
* Doc: [Server-Side Apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/)

It was discovered that we need a required field in [SubnetSpec](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta1/network_types.go#L218) that can be used as a listMapKey with Server Side Apply. This fields needs to be unique in the list.

```yaml
// +listType=map
// +listMapKey=some-unique-key
Subnets Subnets `json:"subnets,omitempty"
```

I tried to use `id` as a key but got the following error as it is not a required field.
```sh
$ kubectl apply -f config/crd/bases/infrastructure.cluster.x-k8s.io_awsclusters.yaml
The CustomResourceDefinition "awsclusters.infrastructure.cluster.x-k8s.io" is invalid: spec.versions[2].schema.openAPIV3Schema.properties[spec].properties[network].properties[subnets].items.properties[id].default: Required value: this property is in x-kubernetes-list-map-keys, so it must have a default or be a required property
```

All the existing fields in SubnetSpec are optional, hence cannot be used as a listMapKey.

**Anything else you would like to add:**
My initial proposal is to add `name` as a required field similar to CAPZ. This needs more investigation.

Feel free to suggest other ideas!

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

sedefsavas:
/triage accepted
/priority important-soon
/milestone v1.5.0

Need to make sure whatever option we choose for this, the existing clusters+manifests used by users keep working. Making a field required might break things.
Will take a look more.

pydctw:
@sedefsavas, @yastij and I experimented adding `uuid` field to SubnetSpec with a default with the intention of modifying it in a defaulting webhook.

```go
// SubnetSpec configures an AWS Subnet.
type SubnetSpec struct {
	// +kubebuilder:default=foo
	UUID string `json:"uuid"`

[...]

Issue #3536, opened 2022-06-16
Add markers defining merge strategy for Server Side Apply to Subnet spec


pydctw:
/kind feature

**Describe the solution you'd like**
`listType` and `listMapKey` need to be added to Subnets field to define a merge strategy for Server Side Apply.
```go
// +listType=map
// +listMapKey=some-unique-key
Subnets Subnets `json:"subnets,omitempty"
```

See [markers defining merge strategy for server-side apply](https://kubernetes.io/docs/reference/using-api/server-side-apply/#merge-strategy).

Issue #1484, opened 2020-01-15
VPC and Networking Topologies


randomvariable:
/kind feature

**Describe the solution you'd like**
There are differing options on how to run Kubernetes clusters in AWS, these include:

* IPv6 vs IPv4
* NAT gateways for internet connectivity, vs. DirectConnect vs. internal only
* A plethora of CNIs
* Multiple load balancer implementation for Kubernetes services

How does a user figure out what one to use, and how best can we enable them?

Related issues include:
#931, #1208 , #1158 , #1062 

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

This also relates to the UX of clusterctl v2

randomvariable:
@fabriziopandini if you have any thoughts on this, would be appreciated.

fabriziopandini:
@randomvariable I have something similar in the radar, but I dubt we can workout details during this iteration
rif https://github.com/kubernetes-sigs/cluster-api/issues/1729
> - [ ] Prototype interactive mode for allowing users to set "on-the-fly" the variables to be injected in the yaml for providers components or providers templates
> - [ ] Prototype a pluggable template system (vs supporting only variables substitution)

In order to ^^^, it should be defined a way for clusterctl to interact with each provider while creating the cluster template, because the providers are the owner of the knowledge of what can be configured or not, and considering this should be accepted by each provider, I assume this requires a CAEP

richardcase:
Also relates to #1643 and #1323

randomvariable:
@richardcase Would be useful to get the requirements for EKS down. I would have thought we can make EKS work with the existing topology.

richardcase:
@randomvariable  - i'll start documenting the requirments. The current default topology doesn't work as the 2 subnets it creates (1 public & 1 private) are in a single az and EKS requires subnets in at least 2 azs.

richardcase:
https://docs.aws.amazon.com/eks/latest/userguide/network_reqs.html

moensch:
I've been asked to document our specific topology in here (ref: https://kubernetes.slack.com/archives/CD6U2V71N/p1588804729193200)

Our starting point in the account is that a Direct Connect Gateway (DXG) exists. What we would need CAPA be able to do is:

1. Accept DXG ID as input (probably as part of the `AWSCluster NetworkSpec`?)
2. Ability to create a Virtual Private Gateway/VPN Gateway/VGW (it has many names)
[...]

PR #3748, opened 2022-09-25
Make subnet spec id field required for SSA to work with CC

sedefsavas:
This PR makes the id field in the subnet spec required to solve SSA coauthoring issue in https://github.com/kubernetes-sigs/cluster-api/issues/6320.

Before this PR, some fields in network.subnets list could be set, although the infrastructure was managed by CAPA and this was unintentional, just because there were no webhook checks in place.

After this PR, the only 2 possible cases we support are:

- User can set any/all fields in subnet spec when using external infrastructure as long as id of the subnet is provided.
- User cannot set any field in subnet spec if infrastructure is managed by CAPA.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3530
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3528

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


Skarlso:
@sedefsavas Hi 👋 For my sake, would you mind clarifying the abbreviations in the title, please? :) Is CC `ClusterClass`?

sedefsavas:
@Skarlso I added the issues that this PR solves.
SSA -> Server side apply
CC-> ClusterClass

Skarlso:
@sedefsavas This is now ready for review, right? :)

sedefsavas:
> @sedefsavas This is now ready for review, right? :)
Yes, ready. Passed ClusterClass tests.
Let me run the regular tests too.

Skarlso:
Awesome, I shall take a crack at it. :)

api/v1beta1/awscluster_conversion.go
```diff
@@ -1,5 +1,5 @@
 /*
-Copyright 2021 The Kubernetes Authors.
+Copyright 2022 The Kubernetes Authors.
```
Skarlso, line 2: You haven't touched these files, why change these particular ones? There are MANY others with `Copyright 2021 The Kubernetes Authors`. I think either don't change these or change all of them, right? :)

api/v1beta2/network_types.go
```diff
@@ -242,7 +242,7 @@ func (v *VPCSpec) IsIPv6Enabled() bool {
 // SubnetSpec configures an AWS Subnet.
 type SubnetSpec struct {
 	// ID defines a unique identifier to reference this resource.
-	ID string `json:"id,omitempty"`
+	ID string `json:"id"`
```
Skarlso, line 245: So, what happens in case of existing clusters is that their ID will be set to a default empty string, right?
Does that affect the cluster at all?

test/e2e/data/infrastructure-aws/kustomize_sources/external-vpc-clusterclass/byo-infra-variables.yaml
```diff
@@ -0,0 +1,28 @@
+- op: add
```
Skarlso, line 1: So... (and this is probably my lack of knowledge in this area, but please bare with me) They way these different from currently configurable networking, like just plain adding your VPC ID into the cluster config, is that these values will now be managed fields and used in SSA. So the user doesn't have to change things in a config file, but rather apply new patch values? Did I get that right?

api/v1beta2/network_types.go
sedefsavas: This means if a subnet spec is defined, it should have the id field.

test/e2e/data/infrastructure-aws/kustomize_sources/external-vpc-clusterclass/byo-infra-variables.yaml
sedefsavas: Yes, this particular scenario tests BYO VPC and subnets which was not working before SSA and id field being compulsory.

api/v1beta2/network_types.go
sedefsavas: https://github.com/kubernetes-sigs/cluster-api/issues/6320

Skarlso: Yeah, I saw this one. Just was wondering about existing configs since now it will not just omit the value but update it to an empty string.

sedefsavas:
Clusterclass and EKS tests passed.

On the unmanaged side, only multi-workload test failed because it is no longer a supported workflow (see the explanation in PR definition). Removed the test.

In summary, if the infrastructure is not external (i.e., id is not set), then the fields below cannot be filled (as we have in the multi-workload test).
```
  network:
    subnets:
      - availabilityZone: "${AWS_AVAILABILITY_ZONE_1}"
        cidrBlock: "10.0.0.0/24"

```

sedefsavas:
All e2e tests passed before, the new commit only removed the multi-az test, which was failing due to its use case became unsupported with this PR. So, not running them again.

api/v1beta2/network_types.go
sedefsavas: For existing configs (by this I mean existing clusters), even though initially id field was empty, after the creation, ids are filled by CAPA controllers, so no existing cluster should have their id fields empty during upgrading to v1beta2.

If you are asking about using an existing template, they will no longer work for creating new clusters. 
This is a breaking change for existing templates that use the unsupported use cases and the reason why we needed a new API version is because of this. We will need to document this properly with the v1beta2 release.

richardcase:
> Clusterclass and EKS tests passed.
> 
> On the unmanaged side, only multi-workload test failed because it is no longer a supported workflow (see the explanation in PR definition). Removed the test.
> 
> In summary, if the infrastructure is not external (i.e., id is not set), then the fields below cannot be filled (as we have in the multi-workload test).
> 
> ```
>   network:
>     subnets:
>       - availabilityZone: "${AWS_AVAILABILITY_ZONE_1}"
>         cidrBlock: "10.0.0.0/24"
> ```

@sedefsavas - sorry but are you saying that the previous logic of only setting az/cidrblock (and not id) didn't work previously? If it did work this seems quite a change in behaviour.

richardcase:
Based on a conversation in [slack](https://kubernetes.slack.com/archives/CD6U2V71N/p1664272448507439) with @sedefsavas:

- We'll mark this change in behaviour as a breaking change in the release notes
- We'll try to come up with an alternative that people can use (perhaps explicitly specifying failure domains or something similar)

Not from the conversation, but it would be good to [re-]add an e2e test that explicitly selects a subset of AZs

So based on this:

/approve

Ankitasw:
Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3536

anmolsachan:
> Based on a conversation in [slack](https://kubernetes.slack.com/archives/CD6U2V71N/p1664272448507439) with @sedefsavas:
> 
> * We'll mark this change in behaviour as a breaking change in the release notes
> * We'll try to come up with an alternative that people can use (perhaps explicitly specifying failure domains or something similar)
> 
> Not from the conversation, but it would be good to [re-]add an e2e test that explicitly selects a subset of AZs
> 
> So based on this:
> 
> /approve

@richardcase @Ankitasw  Hi,

Although I understand the intent behind this change, it takes away the following functionalities from the user.

1. Ability to create custom VPC using CAPI. One of the primary intents of using CAPI is to leave the responsibility of Infrastructure creation to CAPI. With this change, if I have to create 100 clusters with my custom VPC configuration, I will have to now provision VPCs using some other tool like terraform etc. 

2. Ability to customize the subnets - Count of public and private subnets, Subnet ranges.

Example:

```
network:
    subnets:
      - cidrBlock: 10.0.32.0/20
        isPublic: true
        availabilityZone: us-west-2c
      - cidrBlock: 10.0.160.0/20
        isPublic: false
        availabilityZone: us-west-2c
      - cidrBlock: 10.0.0.0/20
        isPublic: true
        availabilityZone: us-west-2a
      - cidrBlock: 10.0.128.0/20
        isPublic: false
        availabilityZone: us-west-2a
      - cidrBlock: 10.0.176.0/20
        isPublic: false
        availabilityZone: us-west-2a
      - cidrBlock: 10.0.16.0/20
        isPublic: true
        availabilityZone: us-west-2b
      - cidrBlock: 10.0.144.0/20
        isPublic: false
        availabilityZone: us-west-2b
      - cidrBlock: 10.0.192.0/20
        isPublic: false
        availabilityZone: us-west-2b
```

3. Availability Zone selection. If I want CAPI to create the AWS VPC, this would also now stop me to choose the availability zones reliably in other APIs like :
AwsMachinePool: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b7df7032c3efa43eeeab5fb126f1aadbf4a18f36/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmachinepools.yaml#L69
AwsManagedMachinePool: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b7df7032c3efa43eeeab5fb126f1aadbf4a18f36/config/crd/bases/infrastructure.cluster.x-k8s.io_awsmanagedmachinepools.yaml#L74

Would there be a possibility to provide solutions to the above?

richardcase:
@MarcusNoble -  Coming up with the alternative did not happen yet and in all honesty it was forgotten about. The problem is that there was significant pressure from certain areas to get ClusterClass support into CAPA and the lack of a key on the subnets caused issues.

The specifying of subnets has been an area of pain for as long as I can remember, as the original implementation wasn't great.

In the slack discussion on this i did raise it was a change in behaviour but was told its undocumented and is a side effect. In hindsight, we should've been more diligent on this and not bowed to pressure to get clusterclass support out.

If we plan to change this then it will be a breaking API change and so we will need to bump the API version / CAPA version number. 

We should take this opportunity to rethink how we handle the networking side of CAPA and once and for all deprecate the old subnet code which has been problematic for years. Lets add this to: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1484 (and also create a specific issue for this problem).

richardcase:
This change has also highlighted that we didn't include this change in the release notes, which is a good example as to why we should go back to using the Prow release notes plugin.

PR #3341, opened 2022-03-23
WIP feat: add support for pre/post commands on EKS node bootstrap 

richardcase:
**What type of PR is this?**

/kind feature
/kind api-change

**What this PR does / why we need it**:

Enable the ability to run custom commnads pre/post bootstrapping of a node into an EKS cluster

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2771 

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
NONE
```

PR #3757, opened 2022-09-27
Enhance the EKSConfigTemplate and add configuration for files, mounts, users, ntp, etc for CAPI feature parity

### LINKED ISSUE #3756: Enhance cloud-init to use the data-format and add missing capabilities for parity w/CAPI
/kind feature

The EKSConfigTemplate currently uses a [bash-style cloud-init user data](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/bootstrap/eks/internal/userdata/node.go#L28). However, [cloud-init supports several user data formats](https://cloudinit.readthedocs.io/en/latest/topics/format.html), including a YAML-based document which has exposes more capabilities than can easily be accomplished with bash. CAPI's KubeadmConfigTemplate uses the YAML data format of cloud-init, which allows it to have additional capabilities in disk, mount, and filesystem layouts, amongst other things. CAPA should adopt the same YAML-based cloud-init format and expose the [same set of capabilities](https://github.com/kubernetes-sigs/cluster-api/blob/main/bootstrap/kubeadm/api/v1beta1/kubeadmconfig_types.go#L54-L78) as the KubeadmConfigTemplate in regards to cloud-init. The feature parity makes it easier for users coming from a self-managed clusters to switch and adopt EKS clusters.

Relevant CAPI cloudinit logic: https://github.com/kubernetes-sigs/cluster-api/blob/main/bootstrap/kubeadm/internal/cloudinit/cloudinit.go

Slack: https://kubernetes.slack.com/archives/CD6U2V71N/p1664301807063759

-   Cluster-api-provider-aws version: 1.5.0
-   Kubernetes version: (use kubectl version): 
-   OS (e.g. from /etc/os-release):

cnmcavoy:
**What type of PR is this?**
/kind feature

Enhances the EKSConfigTemplate cloud-init to expose many more userdata features when bootstrapping EKS nodes in a EKS managed cluster. CAPI already exposes these configuration in the KubeadmConfigTemplate. Feature parity makes it easier for users coming from a self-managed clusters to switch and adopt EKS clusters.

Enhances the EKSConfigTemplate cloud-init, which currently uses a [bash-style cloud-init user data](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/bootstrap/eks/internal/userdata/node.go#L28) to the [YAML-based cloud-init data format](https://cloudinit.readthedocs.io/en/latest/topics/format.html), which has exposes more capabilities than can easily be accomplished with bash. CAPI's KubeadmConfigTemplate uses the YAML data format of cloud-init, which allows it to have additional capabilities in disk, mount, and filesystem layouts, amongst other things. This PR adds the same capabilities to CAPA  and exposes them as configuration in the EKSConfigTemplate. 

Fixes #3756

This PR was rebased off of an existing PR, https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3341, which exposed more commands to the user, but did not switch the cloud-init away from a bash format.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests


bootstrap/eks/api/v1beta1/conversion.go
```diff
@@ -93,3 +93,8 @@ func (r *EKSConfigTemplateList) ConvertFrom(srcRaw conversion.Hub) error {
 
 	return Convert_v1beta2_EKSConfigTemplateList_To_v1beta1_EKSConfigTemplateList(src, r, nil)
 }
+
```
Skarlso, line 96: I can't comment on the line above, but what's missing here is to convert these values:
```
// WARNING: in.PreBootstrapCommands requires manual conversion: does not exist in peer-type
	// WARNING: in.PostBootstrapCommands requires manual conversion: does not exist in peer-type
	// WARNING: in.BootstrapCommandOverride requires manual conversion: does not exist in peer-type
	// WARNING: in.Files requires manual conversion: does not exist in peer-type
	// WARNING: in.DiskSetup requires manual conversion: does not exist in peer-type
	// WARNING: in.Mounts requires manual conversion: does not exist in peer-type
	// WARNING: in.Users requires manual conversion: does not exist in peer-type
	// WARNING: in.NTP requires manual conversion: does not exist in peer-type
```

These should be done manually in the above ConvertTo and ConvertFrom methods.

For example, like here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3741/files#diff-a86bb4e63de0ea453a7d13e98a2d9b8b7de1181d7bfa542a67e76e7d711678baR36

bootstrap/eks/api/v1beta1/conversion_test.go
```diff
@@ -1,46 +0,0 @@
-/*
```
Skarlso, line 1: Why was this test deleted? 🤔

bootstrap/eks/internal/userdata/commands.go
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

bootstrap/eks/internal/userdata/disk_setup.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

bootstrap/eks/internal/userdata/files.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

bootstrap/eks/internal/userdata/fs_setup.go
```diff
@@ -0,0 +1,44 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

bootstrap/eks/internal/userdata/mounts.go
```diff
@@ -0,0 +1,29 @@
+/*
+Copyright 2020 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

bootstrap/eks/internal/userdata/ntp.go
```diff
@@ -0,0 +1,32 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

bootstrap/eks/internal/userdata/users.go
```diff
@@ -0,0 +1,60 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

bootstrap/eks/internal/userdata/utils.go
```diff
@@ -0,0 +1,34 @@
+/*
+Copyright 2019 The Kubernetes Authors.
```
Skarlso, line 2:
```suggestion
Copyright 2022 The Kubernetes Authors.
```

Skarlso (changes requested):
Just a couple little changes. This looks okay for me so far. Can you please add an e2e test for this? I can help you with how to write it. :)

Alternatively, you can create a follow up ticket to write an e2e test, in that case, please include a manual run and it's output showing that it actually works.

bootstrap/eks/api/v1beta1/conversion.go
cnmcavoy: Thanks for providing an example, I was able to fill in the ConvertTo and ConvertFrom and restore the conversion_test that I deleted to make the tests pass earlier.

Are there any additional docs on this conversion behavior? I still feel like there's a bit happening under the covers in the conversion that I don't follow.

bootstrap/eks/api/v1beta1/conversion_test.go
cnmcavoy: It's restored, see https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3757#discussion_r990466679

cnmcavoy:
> Just a couple little changes. This looks okay for me so far. Can you please add an e2e test for this? I can help you with how to write it. :)

I'll gladly write an e2e test for this, but I'm not familiar with how the e2e testing works here in general so I'd need some hand holding in testing locally and writing them. To not block this, I lean towards adding the e2e test in a separate ticket.

> 
> Alternatively, you can create a follow up ticket to write an e2e test, in that case, please include a manual run and it's output showing that it actually works.

What kind of output are you looking for? The controller doesn't log anything new, but here's an example EKSConfigTemplate which I successfully used with a MachineDeployment to create EKS workers:
```
apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
kind: EKSConfigTemplate
metadata:
  annotations:
  creationTimestamp: "2022-09-28T19:09:35Z"
  generation: 1
  name: rad-worker-az1-bd8c2fb9
  namespace: capi-ekscmhdev
  ownerReferences:
  - apiVersion: cluster.x-k8s.io/v1beta1
    kind: Cluster
    name: ekscmhdev
    uid: 442b7e5e-85b6-43fb-a2b8-6b6f27dd4f03
  resourceVersion: "172425"
  uid: 9daddfce-2505-4972-9bdf-7af2c745184a
spec:
  template:
    spec:
      containerRuntime: containerd
      diskSetup:
        filesystems:
        - device: /dev/sdb
          filesystem: ext4
          label: RAD
        partitions:
        - device: /dev/sdb
          layout: true
      files:
      - content: |
          vm.max_map_count=262144
        path: /etc/sysctl.d/90-vm-max-map-count.conf
      - content: |
          fs.inotify.max_user_instances=256
        path: /etc/sysctl.d/91-fs-inotify.conf
      kubeletExtraArgs:
        cloud-provider: aws
        event-qps: "10"
        eviction-hard: memory.available<500Mi,nodefs.available<10%
        kube-reserved: cpu=500m,memory=2Gi,ephemeral-storage=1Gi
        node-labels: role.node.kubernetes.io/worker=true,role.node.kubernetes.io/rad-worker=true,network.node.indeed.com/pods=flat,node.indeed.com/node-group=rad-worker-az1
        protect-kernel-defaults: "true"
        register-with-taints: role.node.kubernetes.io/rad-worker=true:NoSchedule
        system-reserved: cpu=500m,memory=1Gi,ephemeral-storage=1Gi
      mounts:
      - - LABEL=RAD
        - /var/lucy
        - ext4
        - defaults
      preBootstrapCommands:
      - 'TOKEN=$(curl -X PUT "http://169.254.169.254/latest/api/token" -H "X-aws-ec2-metadata-token-ttl-seconds:
        21600" 2>/tmp/token)'
      - 'INSTANCE_ID=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/instance-id
        2>/tmp/instance)'
      - 'REGION=$(curl -H "X-aws-ec2-metadata-token: $TOKEN" -v http://169.254.169.254/latest/meta-data/placement/region
        2>/tmp/region)'
      - aws ec2 modify-instance-metadata-options --http-put-response-hop-limit 2 --region
        ${REGION} --instance-id ${INSTANCE_ID}
      - chown 1505:1505 /var/lucy
      - chmod 2755 /var/lucy
      - mkdir -p /var/lucy/_rad/headwater2
      - mkdir -p /var/lucy/_rad/delta2
      - mkdir -p /var/lucy/_rad/sandboxes
      - chown -R 1505:1505 /var/lucy/_rad
      - chmod 2775 /var/lucy/_rad/sandboxes
      - sudo systemctl restart systemd-sysctl
```

bootstrap/eks/api/v1beta1/conversion.go
Skarlso: Basically, they are converted to annotations and then those annotations are restored to actual values. :)

Yeah, the conversion Fuzzy test makes sure that the conversion ( when people upgrade for example ) work in these scenarios.

Skarlso:
For output, I'm looking for something like the actual set of values applied to the cluster and the node. So I would assume a describe of the Node would show the set values.

cnmcavoy:
> For output, I'm looking for something like the actual set of values applied to the cluster and the node. So I would assume a describe of the Node would show the set values.

I dumped capi/capa resources and nodes: https://gist.github.com/cnmcavoy/52457c07ccdf07b388bf62a731c72e27

Skarlso:
Thanks, @cnmcavoy. I'll take a good look at that output. :)

Skarlso:
@cnmcavoy So looking at this output, I have no idea what values you set. :)) Can you also please share your config file?

cnmcavoy:
> @cnmcavoy So looking at this output, I have no idea what values you set. :)) Can you also please share your config file?

Not sure what config file you are referencing. We use a custom helm chart to generate CAPI resources for each cluster, since we started using CAPI in the 0.3 days. I created a gist that also shows the output of the chart, which are effectively the inputs to CAPI: https://gist.github.com/cnmcavoy/b93420d14fc38cfe6bb080fcaac92032

Skarlso:
Okay so I might be misunderstanding something. But I thought that with this pr you are enhancing the ability to configure bootstrapping nodes through making the bootstrap script configurable. Adding pre and post bootstrap scripts as well. What I'm looking for is using this ability to set something to see that it works.

bootstrap/eks/api/v1beta2/eksconfig_types.go
```diff
@@ -87,6 +110,207 @@ type EKSConfigStatus struct {
...
+	// Label specifies the file system label to be used. If set to None, no label is used.
+	Label string `json:"label"`
+	// Partition specifies the partition to use. The valid options are: "auto|any", "auto", "any", "none", and <NUM>, where NUM is the actual partition number.
+	// +optional
+	Partition *string `json:"partition,omitempty"`
+	// Overwrite defines whether or not to overwrite any existing filesystem.
+	// If true, any pre-existing file system will be destroyed. Use with Caution.
+	// +optional
+	Overwrite *bool `json:"overwrite,omitempty"`
+	// ReplaceFS is a special directive, used for Microsoft Azure that instructs cloud-init to replace a file system of <FS_TYPE>.
```
Skarlso, line 302: I don't think this is needed here, right? :) We are specifically an AWS provider.

Skarlso:
/lgtm
With one remark.

@richardcase If you still want to take do a read on this. :)

PR #3758, opened 2022-09-28
Downgrade K8s dependency version to v0.24

sedefsavas:
/kind bug

Using k8s v0.25.x dependencies requires go version 1.19, this pr downgrades the k8s dependency versions to v0.24.2 which is same with CAPI v1.2.2.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


sedefsavas:
cc @Skarlso

main.go
```diff
@@ -115,7 +114,7 @@ func main() {
 	pflag.CommandLine.AddGoFlagSet(flag.CommandLine)
 	pflag.Parse()
 
-	if err := logsv1.ValidateAndApply(logOptions, nil); err != nil {
+	if err := logOptions.ValidateAndApply(nil); err != nil {
```
Skarlso, line 117: Is this right? I thought I made a mistake here and the fix was to use that logv1 package instead.

sedefsavas: Yes, logv1 package is in v0.25 and it was causing the bump.

sedefsavas: When that line is not removed, go modules pull k8s.io v0.25, you can see this locally.

Skarlso: Ah, okay. :)

PR #3783, opened 2022-10-18
fix: wrong EKS tag when upgrading older clusters

dkoshkin:
**What type of PR is this?**
/kind bug

After https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3573, the AWS tag on a EKS cluster was changed. This was a good change to make everything more consistent, but it broke upgrading EKS clusters that were created with CAPA < v1.5.0.

In this PR, both the new and the previous tag formats will be checked and only fail if both are nil.

I also took the opportunity to clean up the error message, here is what it looked like before. It wasn't very clear what it was looking for and printed a nil error.
```
    - lastTransitionTime: "2022-09-22T10:01:30Z"
      message: 'checking owner of default_e2e-eks-upgrade-test-3618755-control-plane
        is e2e-eks-upgrade-test-3618755: %!!(MISSING)w(<nil>)'
```

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/eks/cluster.go
+++ b/pkg/cloud/services/eks/cluster.go
@@ -69,7 +69,7 @@ func (s *Service) reconcileCluster(ctx context.Context) error {
 		oldOwnedTag := cluster.Tags[oldTagKey]
 
 		if ownedTag == nil && oldOwnedTag == nil {
-			return fmt.Errorf("cluster is not tagged with neither %s nor %s", tagKey, oldTagKey)
+			return fmt.Errorf("EKS cluster resource %q must have a tag with key %q or %q", eksClusterName, oldTagKey, tagKey)
 		}
 
 		s.scope.V(2).Info("Found owned EKS cluster in AWS", "cluster", klog.KRef("", eksClusterName))
--- a/pkg/cloud/services/eks/cluster.go
+++ b/pkg/cloud/services/eks/cluster.go
@@ -62,8 +62,14 @@ func (s *Service) reconcileCluster(ctx context.Context) error {
 	} else {
 		tagKey := infrav1.ClusterAWSCloudProviderTagKey(eksClusterName)
 		ownedTag := cluster.Tags[tagKey]
-		if ownedTag == nil {
-			return fmt.Errorf("checking owner of %s is %s: %w", s.scope.KubernetesClusterName(), eksClusterName, err)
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/eks/cluster.go
```diff
@@ -62,8 +62,14 @@ func (s *Service) reconcileCluster(ctx context.Context) error {
...
-		if ownedTag == nil {
-			return fmt.Errorf("checking owner of %s is %s: %w", s.scope.KubernetesClusterName(), eksClusterName, err)
+		// Prior to https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3573,
+		// Clusters were tagged using s.scope.Name()
+		// To support upgrading older clusters, check for both tags
+		oldTagKey := infrav1.ClusterAWSCloudProviderTagKey(s.scope.Name())
+		oldOwnedTag := cluster.Tags[oldTagKey]
+
+		if ownedTag == nil && oldOwnedTag == nil {
+			return fmt.Errorf("cluster is not tagged with neither %s nor %s", tagKey, oldTagKey)
```
dlipovetsky, line 72: Let's use the error message to explain what _should_ be true, since that may be enough information for the user to address the error.

Let's also mention the EKS cluster resource by name, so that the user can look it up using the AWS CLI or Console.

```suggestion
			return fmt.Errorf("EKS cluster resource %q must have a tag with key %q or %q", eksClusterName, oldTagKey, tagKey)
```

dlipovetsky:
We test Kubernetes version upgrades using the same CAPA version for [managed clusters](https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/ad99a1d982c9e938ac22905926a5f7c82e1603d6/test/e2e/suites/managed/upgrade_test.go), but I do not think we test CAPA upgrades themselves.

It would be nice to think of lighter-weight test that could catch this type of bug, but testing a CAPA upgrade would definitely work.

faiq:
This is a really nice change. Should we consider removing the old tag as well and potentially getting rid of this extra code moving forward?

dlipovetsky:
/triage accepted

As far as I understand it, v1.5.0 and v2.0-beta are unable to reconcile an EKS cluster created with an earlier CAPA version, e.g., v1.4.x.

/priority critical-urgent

dlipovetsky:
I'll look into the release-1.5 backport

PR #3785, opened 2022-10-19
Add /v2 to go module name

sbueringer:
**What type of PR is this?**

/kind cleanup

This PR renames the go module to be compliant with go modules version v2 or higher.

I assume `repo: sigs.k8s.io/cluster-api-provider-aws` in PROJECT files should not be changed

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


sbueringer:
cc @richardcase @yastij @srm09

Skarlso:
Hmm. This will break anyone's existing code today that requires any kind of capa library or capa code.

Personally, I would prefer the normal way of having a subfolder called v2. This would allow all tooling to still work and also allows CAPA to be backwards compatible and not break any existing imports from other possible projects.

[This](https://go.dev/blog/v2-go-modules) post outlines that approach.

That said... that path is rarely followed. And possibly very few projects require code from CAPA. So... On second thought, maybe this is fine. :)

sbueringer:
Yup. I think it's also fine as a major version bump is usually considered breaking. For the import paths themselves it's still relatively simple to adapt.

Ankitasw:
/lgtm
Need another 👀 for approval. cc @richcase @Skarlso

yastij (commented):
/lgtm

Skarlso:
Tests passed, and this looks good to me too.
/approve

PR #3798, opened 2022-10-27
fix: remove check for root volume device name on create

### LINKED ISSUE #3624: `clusterctl move` not compatible with `AWSMachinePools`
/kind bug

1. Spin up a bootstrap cluster in Kind
2. Create a new target cluster with at least one `AWSMachinePool` defined
3. Wait for target cluster to be created and ready
4. Perform a pivot of the cluster using `clusterctl move` so the target cluster is self-managing.
5. The following error will be reported:
```
Performing move...
Discovering Cluster API objects
Moving Cluster API objects Clusters=1
Moving Cluster API objects ClusterClasses=0
Creating objects in the target cluster
Error: [action failed after 10 attempts: error creating "infrastructure.cluster.x-k8s.io/v1beta1, Kind=AWSMachinePool" default/golem-def00a: admission webhook "validation.awsmachinepool.infrastructure.cluster.x-k8s.io" denied the request: AWSMachinePool.infrastructure.cluster.x-k8s.io "golem-def00a" is invalid: spec.awsLaunchTemplate.rootVolume.deviceName: Forbidden: root volume shouldn't have device name, action failed after 10 attempts: error creating "infrastructure.cluster.x-k8s.io/v1beta1, Kind=AWSMachinePool" default/golem-def00b: admission webhook "validation.awsmachinepool.infrastructure.cluster.x-k8s.io" denied the request: AWSMachinePool.infrastructure.cluster.x-k8s.io "golem-def00b" is invalid: spec.awsLaunchTemplate.rootVolume.deviceName: Forbidden: root volume shouldn't have device name, action failed after 10 attempts: error creating "infrastructure.cluster.x-k8s.io/v1beta1, Kind=AWSMachinePool" default/golem-def00c: admission webhook "validation.awsmachinepool.infrastructure.cluster.x-k8s.io" denied the request: AWSMachinePool.infrastructure.cluster.x-k8s.io "golem-def00c" is invalid: spec.awsLaunchTemplate.rootVolume.deviceName: Forbidden: root volume shouldn't have device name]
```

All resources moved to the target cluster successfully.

The `rootVolume.deviceName` is initially not provided when first creating the cluster resources in the bootstrap cluster. Once the AWS Launch Template has been created the details of the root volume are retrieved and the `deviceName` value is populated on the `AWSMachinePool` resource(s). When it comes to moving to the new cluster, the property remains populated and is then blocked by the admission webhook, preventing the move completing.

This value only seems to be used during the initial setup of the Launch Template and as far as I can see is never referenced by anything else after that. Manually removing the `deviceName` property from the `AWSMachinePool` resources allows the move to be performed but the value is never re-populated again as it's only fetched when initially creating the Launch Template.

Also discussed on Slack: https://kubernetes.slack.com/archives/CD6U2V71N/p1658902259480619

**Environment:**

- Cluster-api-provider-aws version: `v1.4.1`
- Cluster-api version: `v1.1.5`
- clusterctl version: `v1.2.0`
- Kubernetes version: (use `kubectl version`): `v1.21.1`
- OS (e.g. from `/etc/os-release`): `Ubuntu`

AverageMarcus:
**What type of PR is this?**

/kind bug

`clusterctl move` currently fails if trying to move an `AWSMachinePool` that have it's `rootVolume` already populated.

Fixes #3624

From what I can tell, the RootVolume device name is actually retrieved from the AMI image and the value in the `Spec` isn't used (only populated after creating).

See: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7440d3ed110540432a7aed1ea165aaea79157a9c/pkg/cloud/services/ec2/instances.go#L926-L942

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/api/v1beta2/awsmachine_webhook.go
+++ b/api/v1beta2/awsmachine_webhook.go
@@ -193,7 +193,7 @@ func (r *AWSMachine) validateRootVolume() field.ErrorList {
 	}
 
 	if r.Spec.RootVolume.DeviceName != "" {
-		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.rootVolume.deviceName"), "root volume shouldn't have device name"))
+		log.Info("root volume shouldn't have a device name (this can be ignored if performing a `clusterctl move`)")
 	}
 
 	return allErrs
--- a/api/v1beta2/awsmachine_webhook_test.go
+++ b/api/v1beta2/awsmachine_webhook_test.go
@@ -80,16 +80,18 @@ func TestAWSMachine_Create(t *testing.T) {
 			wantErr: true,
 		},
 		{
-			name: "ensure root volume has no device name",
+			name: "ensure root volume with device name works (for clusterctl move)",
 			machine: &AWSMachine{
 				Spec: AWSMachineSpec{
 					RootVolume: &Volume{
 						DeviceName: "name",
+						Type:       "gp2",
+						Size:       *aws.Int64(8),
 					},
 					InstanceType: "test",
 				},
 			},
-			wantErr: true,
+			wantErr: false,
 		},
 		{
 			name: "ensure non root volume have device names",
--- a/api/v1beta2/awsmachinetemplate_webhook.go
+++ b/api/v1beta2/awsmachinetemplate_webhook.go
@@ -68,7 +68,7 @@ func (r *AWSMachineTemplate) validateRootVolume() field.ErrorList {
 	}
 
 	if spec.RootVolume.DeviceName != "" {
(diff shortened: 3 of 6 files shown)


api/v1beta2/awsmachine_webhook.go
```diff
@@ -192,10 +192,6 @@ func (r *AWSMachine) validateRootVolume() field.ErrorList {
 		}
 	}
 
-	if r.Spec.RootVolume.DeviceName != "" {
```
Ankitasw, line 195: How about we log this info and not return error?

AverageMarcus: Good idea!  Updated :)

PR #3805, opened 2022-10-31
Do not return error if secret does not exist

enxebre:
**What type of PR is this?**
The non existence of the bootstrap secret should not prevent an awsMachine from being deleted

Fixes #

This change in behaviour was most likely introduced during the ignition support introduction 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2271/files#diff-4f8240f4e7ff07dcfa9627342b62772c1fe6ae0943021abe53fbb966db88886e

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2271/files#diff-0b559bbd149f0e6d54d789235423f66fa8906fdc6ee9c99b9e85db912912011e

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
deleteIgnitionBootstrapDataFromS3 do not return error if secret is not found
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -763,7 +763,7 @@ func (r *AWSMachineReconciler) deleteIgnitionBootstrapDataFromS3(machineScope *s
 	}
 
 	_, userDataFormat, err := machineScope.GetRawBootstrapDataWithFormat()
-	if err != nil {
+	if err != nil && !apierrors.IsNotFound(err) {
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedGetBootstrapData", err.Error())
 		return err
 	}


pkg/cloud/scope/machine.go
```diff
@@ -275,6 +276,10 @@ func (m *MachineScope) GetRawBootstrapDataWithFormat() ([]byte, string, error) {
 	secret := &corev1.Secret{}
 	key := types.NamespacedName{Namespace: m.Namespace(), Name: *m.Machine.Spec.Bootstrap.DataSecretName}
 	if err := m.client.Get(context.TODO(), key, secret); err != nil {
+		if apierrors.IsNotFound(err) {
```
Ankitasw, line 279: @enxebre is it possible to add a unit test for the same?

dlipovetsky:
I agree with the goal of the PR.

However, `GetRawBootstrapDataWithFormat` is also called by [`resolveUserData`](https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/7160f5ee4fbb27c3b922f62445f09ced9a4615c3/controllers/awsmachine_controller.go#L654), which is called  by [`createInstance`](https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/7160f5ee4fbb27c3b922f62445f09ced9a4615c3/controllers/awsmachine_controller.go#L640), and I'm sure we expect an error there if the data isn't found.

Moreover, we want to return an error to indicate that the first return parameter should not be used.

@enxebre, what if this function continue to return an error, and any caller that wants to ignore the NotFound error can check for it?

pkg/cloud/scope/machine.go
alexander-demicev: maybe also add a debug log?

enxebre:
@dlipovetsky update PTAL.

enxebre:
cc @Ankitasw

pkg/cloud/scope/machine.go
Ankitasw: @enxebre can these comments be taken care of?

Ankitasw:
@enxebre could you please take care of open comment?

Ankitasw:
@enxebre would you be able to continue with this PR?

enxebre:
All comments have been addressed via https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3805#issuecomment-1297454230

PR #3467, opened 2022-05-12
Fix typo in AWSManagedMachinePool UpdateConfig

phoban01:
**What type of PR is this?**

/kind cleanup

**What this PR does / why we need it**:

I noticed a small typo in the spelling of "percentage" in the struct tag for ``UpdateConfig.MaxUnavailablePercentage`.  

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

PR #3830, opened 2022-11-08
Fix typo in UpdateConfig field, from maxUnavailablePrecentage to maxUnavailablePercentage

dlipovetsky:
**What type of PR is this?**
/kind cleanup

This is the change in #3467, but applied to v1beta2, instead of v1beta1. The typo should not be fixed in v1beta1, as it is a breaking change.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


dlipovetsky:
We also need to generate the ["CRD reference"](https://cluster-api-aws.sigs.k8s.io/crd/index.html) in the book. However, that change has a larger scope, because we haven't re-generated it in some time. Therefore I'd like to submit a separate PR with those changes. WDYT?

PR #3833, opened 2022-11-09
Make overwrite the default behaviour for conflict resolution in addons

Skarlso:
**What type of PR is this?**

/kind feature

After the IPv6 discussion there was this comment from Richard: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3513#discussion_r917737084

There is no scenario in which you wouldn't want to overwrite, so we set it up as default behaviour.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/eks/addons.go
+++ b/pkg/cloud/services/eks/addons.go
@@ -208,12 +208,8 @@ func (s *Service) translateAPIToAddon(addons []ekscontrolplanev1.Addon) []*eksad
 }
 
 func convertConflictResolution(conflict ekscontrolplanev1.AddonResolution) *string {
-	switch conflict {
-	case ekscontrolplanev1.AddonResolutionNone:
+	if conflict == ekscontrolplanev1.AddonResolutionNone {
 		return aws.String(eks.ResolveConflictsNone)
-	case ekscontrolplanev1.AddonResolutionOverwrite:
-		return aws.String(eks.ResolveConflictsOverwrite)
-	default:
-		return nil
 	}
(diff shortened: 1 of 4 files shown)


Ankitasw:
cc @richardcase for approval

pkg/cloud/services/eks/addons.go
```diff
@@ -214,6 +214,6 @@ func convertConflictResolution(conflict ekscontrolplanev1.AddonResolution) *stri
 	case ekscontrolplanev1.AddonResolutionOverwrite:
 		return aws.String(eks.ResolveConflictsOverwrite)
 	default:
-		return nil
+		return aws.String(eks.ResolveConflictsOverwrite)
```
AverageMarcus, line 217: Is it worth simplifying this function?

```
func convertConflictResolution(conflict ekscontrolplanev1.AddonResolution) *string {
	if conflict == ekscontrolplanev1.AddonResolutionNone {
		return aws.String(eks.ResolveConflictsNone)
	}
	return aws.String(eks.ResolveConflictsOverwrite)
}
```

Skarlso: Sure why not. I don't think there will ever be a third option. So it makes sense to simplify it.

AverageMarcus: Yeah that's what I was thinking but wasn't 100% sure if more options could be possible. Either way, code change would be needed if they ever are so can change it back then if needed.

PR #3836, opened 2022-11-10
chore: bump capi to 1.2.5

richardcase:
Signed-off-by: Richard Case <richard.case@outlook.com>

**What type of PR is this?**

/kind cleanup

Bumps CAPI to 1.2.5 as part of the 2.0.0 release

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


Ankitasw:
I guess we should also bump controller-runtime with this bump.

richardcase:
> I guess we should also bump controller-runtime to 0.13.1 and gomega to 0.24.x with this bump.

I don't think we should upgrade controller-runtime as part of this, we are inline with CAPI 1.2.5 with the version we have. Plus, it's bumping a minor version number and this may require changes.

Let's keep this as a pure capi version bump (and indirect deps)

PR #3843, opened 2022-11-11
fix: Replace misleading wording used for security group overrides

dlipovetsky:
The wording "overridden" is misleading. In every case where the wording is used to reference a security group, the security group is being overridden. Instead, it is the security group provided as an override.

**What type of PR is this?**
/kind cleanup

The wording "overridden" is misleading. In every case where the wording is used to reference a security group, the security group is being overridden. Instead, it is the security group provided as an override.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/scope/managedcontrolplane.go
+++ b/pkg/cloud/scope/managedcontrolplane.go
@@ -193,7 +193,7 @@ func (s *ManagedControlPlaneScope) SecondaryCidrBlock() *string {
 	return s.ControlPlane.Spec.SecondaryCidrBlock
 }
 
-// SecurityGroupOverrides returns the security groups that are overridden in the ControlPlane spec.
+// SecurityGroupOverrides returns the security groups that are overrides in the ControlPlane spec.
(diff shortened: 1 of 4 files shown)


pkg/cloud/scope/managedcontrolplane.go
```diff
@@ -193,7 +193,7 @@ func (s *ManagedControlPlaneScope) SecondaryCidrBlock() *string {
 	return s.ControlPlane.Spec.SecondaryCidrBlock
 }
 
-// SecurityGroupOverrides returns the security groups that are overridden in the ControlPlane spec.
+// SecurityGroupOverrides returns the security groups that are overrides in the ControlPlane spec.
```
Ankitasw, line 196:
```suggestion
// SecurityGroupOverrides returns the security groups that are used as overrides in the ControlPlane spec.
```

Ankitasw (commented):
Small nit, Otherwise LGTM

pkg/cloud/scope/managedcontrolplane.go
Ankitasw: Can be taken up in next PR

PR #2715, opened 2021-08-26
Add AWS cloud provider tags to pre-existing infrastructure

pydctw:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Support writing AWS cloud provider tags on pre-existing infrastructure

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2584

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests
- [ ] adds or updates e2e tests

**Release note:**
```release-note
Support adding AWS cloud provider tags to pre-existing infrastructure
`kubernetes.io/cluster/<name> = shared` and `kubernetes.io/role/elb = 1` for public subnets
`kubernetes.io/cluster/<name> = shared` and`kubernetes.io/role/internal-elb = 1` for private subnets
```

PR #3844, opened 2022-11-11
feat: Tag for CCM the security group provided as an override for the "sg-lb" security group role

### LINKED ISSUE #3481: Unmanaged SecurityGroups should also get tagged for CCM
/kind feature

When using an unmanaged `SecurityGroup` for `[spec.network.securityGroupOverrides.lb](http://spec.network.securitygroupoverrides.lb/)` CAPA should tag it with `kubernetes.io/cluster/<cluster-name>` so that CCM can create ELBs for `LoadBalancer` `Services` without requiring users to tag the `SecurityGroup` themselves.

When an instance has multiple SecurityGroups attached, CCM requires 1 of them to be tagged https://github.com/kubernetes/cloud-provider-aws/blob/8d2f0fd2b1b574bde3239a344bd0a9a4f244cdb0/pkg/providers/v1/aws.go#L4479-L4485

Currently managed `SecurityGroups` are already tagged with the required `kubernetes.io/cluster` tag. **Unmanaged** Subnets are also already tagged. It would not be a stretch for CAPA to also tag the SecurityGroup.

[Originally discussed in Slack.](https://kubernetes.slack.com/archives/CD6U2V71N/p1652814232505389)

**Environment:**

[v1.4.1](https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/tag/v1.4.1)
v1.22.8

dlipovetsky:
**What type of PR is this?**
/kind feature 

Tag for CCM the security group provided as an override for the "sg-lb" security group role. This tag is required for the CCM to choose the subnet to deploy an ELB. Note that CAPA already adds this tag to an existing subnet (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2715).

Fixes #3481

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/securitygroup/securitygroups.go
+++ b/pkg/cloud/services/securitygroup/securitygroups.go
@@ -119,18 +119,16 @@ func (s *Service) ReconcileSecurityGroups() error {
 			continue
 		}
 
-		if !s.securityGroupIsAnOverride(existing.ID) {
-			// Make sure tags are up to date.
-			if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
-				buildParams := s.getSecurityGroupTagParams(existing.Name, existing.ID, role)
-				tagsBuilder := tags.New(&buildParams, tags.WithEC2(s.EC2Client))
-				if err := tagsBuilder.Ensure(existing.Tags); err != nil {
-					return false, err
-				}
-				return true, nil
-			}, awserrors.GroupNotFound); err != nil {
-				return errors.Wrapf(err, "failed to ensure tags on security group %q", existing.ID)
+		// Make sure tags are up to date.
+		if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+			buildParams := s.getSecurityGroupTagParams(existing.Name, existing.ID, role)
+			tagsBuilder := tags.New(&buildParams, tags.WithEC2(s.EC2Client))
+			if err := tagsBuilder.Ensure(existing.Tags); err != nil {
+				return false, err
 			}
+			return true, nil
+		}, awserrors.GroupNotFound); err != nil {
+			return errors.Wrapf(err, "failed to ensure tags on security group %q", existing.ID)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


dlipovetsky:
If CAPA applies the CCM tag to a user-owned security group on cluster create, should it remove the tag on cluster delete?

If it should remove the tag, then this PR is not complete, because it doesn't remove the tag.

dlipovetsky:
I rebased on main because #3843 merged.

dlipovetsky:
> > If CAPA applies the CCM tag to a user-owned security group on cluster create, should it remove the tag on cluster delete?
> 
> I would say it should undo the action it took on delete.

Would you say the same for subnets and route tables, then? Because we currently do not undo the action for subnets.


> But, do you think we can have multiple CAPA clusters using the same SG in this scenario?

I suppose it is possible, although I'm not sure it would be a good idea. The `lb` SG has one inbound rule for every ELB backing a LoadBalancer Service. It has one outbound rule that allows all traffic.

dlipovetsky:
/hold cancel

#3843 merged.

richardcase:
> Would you say the same for subnets and route tables, then? Because we currently do not undo the action for subnets.

As a guiding principle it would be good to undo changes we have made, so yes. But I'm wondering why didn't do that in the first place. Was it because we just didn't think of it or was there a specific reason.

I agree it does need further discussion. Do you think we delay v2.0.0 until we decide this, or can we push this change to the next API version bump to give us more time?

Ankitasw:
To not delay the v2.0.0 release, can we somehow document this and put this PR on hold for next API version?

dlipovetsky:
This PR does not remove the CCM tag from user-owned resources. As I explain in https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3854, not removing the tag on cluster delete prevents use of the resource in future clusters.

CAPA needs to either remove the CCM tag on cluster delete, or not apply it at all.

I'm going to close this PR. We can open a new one, if needed, once we have consensus on #3854.

PR #3871, opened 2022-11-22
Update LoadBalancerReadyCondition on deletion

### LINKED ISSUE #3870: LoadBalancerReady condition stuck in `Deleting`
/kind bug

1. Create AWSCluster WC
1. Wait for cluster to become ready
1. Delete the cluster
1. Notice that the LoadBalancerReadyCondition never becomes `Deleted`

_NOTE:_ The finalizer is eventually removed, but the condition is stuck in `Deleting`.

The condition to eventually become `Deleted`.

**Environment:**

- Cluster-api-provider-aws version: v1.5.2
- Kubernetes version: (use `kubectl version`): 1.23.13
- OS (e.g. from `/etc/os-release`): ubuntu 22.04

CC @fiunchinho

mnitchev:
**What type of PR is this?**

/kind bug

The LoadBalancerReadyCondition is sometimes explicitly patched and other times it only updates the awsCluster object. Furthermore the condition won't be patched by the patch helper in the AWSClusterReconciler, because it's not in the `patch.WithOwnedConditions` list. This change puts the condition in the list and also updates the condition reason to `Deleted` when the LB is not found. Without setting the condition in that case, the reconciler will set condition to `Deleting` and it will never go back to `Deleted`.

Fixes #3870

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

CC @fiunchinho

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -38,13 +38,6 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
-	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
-	"sigs.k8s.io/cluster-api/util"
-	capiannotations "sigs.k8s.io/cluster-api/util/annotations"
-	"sigs.k8s.io/cluster-api/util/conditions"
-	"sigs.k8s.io/cluster-api/util/patch"
-	"sigs.k8s.io/cluster-api/util/predicates"
-
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/feature"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope"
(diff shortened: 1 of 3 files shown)


Ankitasw:
/hold
until [this](https://kubernetes.slack.com/archives/CD6U2V71N/p1669226823932389) discussion is concluded

AverageMarcus:
@Ankitasw what needs to be concluded with the discussion on Slack?

Is there anything that's blocking this PR from moving ahead?

We've been running with these changes included for a couple weeks now (based on the 1.5 branch) and it's been working as expected.

controllers/awscluster_controller.go
```diff
@@ -171,6 +171,7 @@ func (r *AWSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 			patch.WithOwnedConditions{Conditions: []clusterv1.ConditionType{
 				infrav1.PrincipalCredentialRetrievedCondition,
 				infrav1.PrincipalUsageAllowedCondition,
+				infrav1.LoadBalancerReadyCondition,
```
Ankitasw, line 174: Is this change really needed?
Does setting `LoadBalancerReadyCondition` as done in loadbalancer.go would not suffice?

mnitchev: Our observation was that it didn't work without having the condition in that list. My understanding is that the patch helper will ignore conditions not owned by the controller. At least that's the impression I'm left with by the documentation [here](https://github.com/kubernetes-sigs/cluster-api/blob/333341b1f72e9ac2fdb40b02b0f114054c68cd5b/util/patch/options.go#L60-L64).

I'll double check, since it's been a while since I submitted this change.

Ankitasw: cc @richardcase if you have any opinions

richardcase:
This change seems reasonable to me. Lets check the e2e pass and then merge:

richardcase:
Until the e2e pass:
/hold

richardcase:
/approve

@mnitchev @AverageMarcus - feel free to unhold when the e2e passes.

AverageMarcus:
E2E passed

/hold cancel

PR #3926, opened 2022-12-13
capa fix hardcoded role arn for aws iam authenticator

AmitSahastra:
## **What is the purpose of the pull request**
ARN hardcoded with ":aws:" partition do not work with "us-gov" cloud.

## **Implementation**
Issue: For few roles/policies arn partition is hardcoded to "aws". While working with "aws-us-gov" cloud partition this does not work and cluster creation failed.

Fix: Check for aws cloud "partition" or "region" to derive correct arn and use it.

/kind bug

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


pkg/cloud/services/s3/s3.go
```diff
@@ -242,10 +242,10 @@ func (s *Service) bucketPolicy(bucketName string) (string, error) {
 			Sid:    "control-plane",
 			Effect: iam.EffectAllow,
 			Principal: map[iam.PrincipalType]iam.PrincipalID{
-				iam.PrincipalAWS: []string{fmt.Sprintf("arn:aws:iam::%s:role/%s", *accountID.Account, bucket.ControlPlaneIAMInstanceProfile)},
+				iam.PrincipalAWS: []string{fmt.Sprintf("arn:*:iam::%s:role/%s", *accountID.Account, bucket.ControlPlaneIAMInstanceProfile)},
```
Skarlso, line 245: Shouldn't this be specific? I'm not sure using `*` is appropriate. Maybe you don't want to have other partitions to have access.

sadysnaat: @Skarlso I think we should make this all over the places where we are referring the policies 

in

clusterawsadm bootstrap iam print-policy --document AWSIAMManagedPolicyControllersEKS

even if we provide aws partition we see only replace partition at one place rest all permission 

```json 
{
  "Version": "2012-10-17",
  "Statement": [
    {
      "Effect": "Allow",
      "Action": [
        "ssm:GetParameter"
      ],
      "Resource": [
        "arn:*:ssm:*:*:parameter/aws/service/eks/optimized-ami/*"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "iam:CreateServiceLinkedRole"
      ],
      "Resource": [
        "arn:*:iam::*:role/aws-service-role/eks.amazonaws.com/AWSServiceRoleForAmazonEKS"
      ],
      "Condition": {
        "StringLike": {
          "iam:AWSServiceName": "eks.amazonaws.com"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "iam:CreateServiceLinkedRole"
      ],
      "Resource": [
        "arn:*:iam::*:role/aws-service-role/eks-nodegroup.amazonaws.com/AWSServiceRoleForAmazonEKSNodegroup"
      ],
      "Condition": {
        "StringLike": {
          "iam:AWSServiceName": "eks-nodegroup.amazonaws.com"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "iam:CreateServiceLinkedRole"
      ],
      "Resource": [
        "arn:aws:iam::*:role/aws-service-role/eks-fargate-pods.amazonaws.com/AWSServiceRoleForAmazonEKSForFargate"
      ],
      "Condition": {
        "StringLike": {
          "iam:AWSServiceName": "eks-fargate.amazonaws.com"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "iam:GetRole",
        "iam:ListAttachedRolePolicies"
      ],
      "Resource": [
        "arn:*:iam::*:role/*"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "iam:GetPolicy"
      ],
      "Resource": [
        "arn:aws-us-gov:iam::aws:policy/AmazonEKSClusterPolicy"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "eks:DescribeCluster",
        "eks:ListClusters",
        "eks:CreateCluster",
        "eks:TagResource",
        "eks:UpdateClusterVersion",
        "eks:DeleteCluster",
        "eks:UpdateClusterConfig",
        "eks:UntagResource",
        "eks:UpdateNodegroupVersion",
        "eks:DescribeNodegroup",
        "eks:DeleteNodegroup",
        "eks:UpdateNodegroupConfig",
        "eks:CreateNodegroup",
        "eks:AssociateEncryptionConfig",
        "eks:ListIdentityProviderConfigs",
        "eks:AssociateIdentityProviderConfig",
        "eks:DescribeIdentityProviderConfig",
        "eks:DisassociateIdentityProviderConfig"
      ],
      "Resource": [
        "arn:*:eks:*:*:cluster/*",
        "arn:*:eks:*:*:nodegroup/*/*/*"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "ec2:AssociateVpcCidrBlock",
        "ec2:DisassociateVpcCidrBlock",
        "eks:ListAddons",
        "eks:CreateAddon",
        "eks:DescribeAddonVersions",
        "eks:DescribeAddon",
        "eks:DeleteAddon",
        "eks:UpdateAddon",
        "eks:TagResource",
        "eks:DescribeFargateProfile",
        "eks:CreateFargateProfile",
        "eks:DeleteFargateProfile"
      ],
      "Resource": [
        "*"
      ]
    },
    {
      "Effect": "Allow",
      "Action": [
        "iam:PassRole"
      ],
      "Resource": [
        "*"
      ],
      "Condition": {
        "StringEquals": {
          "iam:PassedToService": "eks.amazonaws.com"
        }
      }
    },
    {
      "Effect": "Allow",
      "Action": [
        "kms:CreateGrant",
        "kms:DescribeKey"
      ],
      "Resource": [
        "*"
      ],
      "Condition": {
        "ForAnyValue:StringLike": {
          "kms:ResourceAliases": "alias/cluster-api-provider-aws-*"
        }
      }
    }
  ]
}
```

AmitSahastra: Pr is updated to look for partition and generate separate arn

Skarlso:
/lgtm

@sadysnaat Are you happy with this too? :)

pkg/cloud/scope/managedcontrolplane.go
```diff
@@ -401,3 +403,14 @@ func (s *ManagedControlPlaneScope) ServiceCidrs() *clusterv1.NetworkRanges {
...
+// Partition returns the cluster partition.
+func (s *ManagedControlPlaneScope) Partition() string {
+	var partition string
+	if strings.Contains(s.Region(), v1beta1.DefaultPartitionNameUSGov) {
+		partition = v1beta1.DefaultPartitionNameUSGov
+	} else {
+		partition = v1beta1.DefaultPartitionName
+	}
+	return partition
+}
```
Skarlso, line 416:
```suggestion
func (s *ManagedControlPlaneScope) Partition() string {
	partition := v1beta1.DefaultPartitionName
	if strings.Contains(s.Region(), v1beta1.DefaultPartitionNameUSGov) {
		partition = v1beta1.DefaultPartitionNameUSGov
	}

	return partition
}
```

pkg/cloud/services/ec2/ami.go
```diff
@@ -198,10 +200,6 @@ func GetLatestImage(imgs []*ec2.Image) (*ec2.Image, error) {
 func (s *Service) defaultBastionAMILookup() (string, error) {
 	describeImageInput := &ec2.DescribeImagesInput{
 		Filters: []*ec2.Filter{
-			{
```
Skarlso, line 201: Why was this removed? 🤔

pkg/cloud/services/ec2/ami.go
```diff
@@ -220,6 +218,21 @@ func (s *Service) defaultBastionAMILookup() (string, error) {
...
+			Values: []*string{aws.String(ubuntuOwnerIDUsGov)},
+		}
+		describeImageInput.Filters = append(describeImageInput.Filters, filter)
+	} else {
+		filter := &ec2.Filter{
+			Name:   aws.String("owner-id"),
+			Values: []*string{aws.String(ubuntuOwnerID)},
+		}
+		describeImageInput.Filters = append(describeImageInput.Filters, filter)
+	}
```
Skarlso, line 234:
```suggestion
        ownerID := ubuntuOwnerID
	if strings.Contains(s.scope.Partition(), defaultUsGovPartitionName) {
                ownerID := ubuntuOwnerIDUsGov
	}

	filter := &ec2.Filter{
		Name:   aws.String("owner-id"),
		Values: []*string{aws.String(ownerID)},
	}
	describeImageInput.Filters = append(describeImageInput.Filters, filter)
```

pkg/cloud/services/ec2/bastion.go
```diff
@@ -35,7 +35,8 @@ import (
 )
 
 const (
-	defaultSSHKeyName = "default"
+	defaultSSHKeyName         = "default"
+	defaultUsGovPartitionName = "us-gov"
```
Skarlso, line 39: You already have this declared in `cmd/clusterawsadm/api/bootstrap/v1beta1/defaults.go`

pkg/cloud/services/ec2/bastion_test.go
```diff
@@ -327,6 +323,10 @@ func TestServiceReconcileBastion(t *testing.T) {
 						Name:   aws.String("description"),
 						Values: aws.StringSlice([]string{ubuntuImageDescription}),
 					},
+					{
+						Name:   aws.String("owner-id"),
+						Values: aws.StringSlice([]string{ubuntuOwnerID}),
+					},
```
Skarlso, line 329: Please add a test so we see that he right owner is applied if necessary.

Skarlso:
Generally, it's okay; I have one question. Most changes restrict the default partition and only allow the government partition to work.

Is that the intention?

Skarlso:
Ops sorry I probably fumbled one of the suggestions.

pkg/cloud/services/eks/roles.go
```diff
@@ -93,8 +111,15 @@ func (s *Service) reconcileControlPlaneIAMRole() error {
...
-	policies := []*string{
-		aws.String("arn:aws:iam::aws:policy/AmazonEKSClusterPolicy"),
+	var policies []*string
+	if strings.Contains(s.scope.ControlPlane.Spec.Region, v1beta1.DefaultPartitionNameUSGov) {
+		policies = []*string{
+			aws.String("arn:aws-us-gov:iam::aws:policy/AmazonEKSClusterPolicy"),
+		}
+	} else {
+		policies = []*string{
+			aws.String("arn:aws:iam::aws:policy/AmazonEKSClusterPolicy"),
```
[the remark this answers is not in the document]
AmitSahastra: That's one reason why in this PR I only addressed resources specific to given regions/partitions like "aws" or "aws-us-gov" and tried to limit this PRs scope to only addressing hardcoded portion issues.

pkg/cloud/services/ec2/bastion_test.go
AmitSahastra: A test case has been added for us-gov partition.

pkg/cloud/services/ec2/bastion.go
AmitSahastra: Thanks for pointing it out, removed duplicate.

AmitSahastra:
> Generally, it's okay; I have one question. Most changes restrict the default partition and only allow the government partition to work.
> 
> Is that the intention?

The intention was to make sure the provider works for the specific partition. 

Due to the hardcoded "aws" partition, there was an issue with resolving regions, ami, and images that are different for different partitions, making sure we look dynamically for region/partition so it works for both.

Avoided going for pain partition here so as to not expose resources meant for specific partition to be exposed to others.

pkg/cloud/services/eks/roles.go
```diff
@@ -52,6 +54,22 @@ func FargateRolePolicies() []string {
 	}
 }
 
+// NodegroupRolePoliciesAWSUSGov gives the policies required for a nodegroup role.
+func NodegroupRolePoliciesAWSUSGov() []string {
```
sadysnaat, line 58: we should refactor and reuse  NodegroupRolePolicies 

as they are exactly same and only difference is partition 

maybe a common function that takes partition as argument

pkg/cloud/services/eks/roles.go
```diff
@@ -203,7 +228,12 @@ func (s *NodegroupService) reconcileNodegroupIAMRole() error {
 		return errors.Wrapf(err, "error ensuring tags and policy document are set on node role")
 	}
 
-	policies := NodegroupRolePolicies()
+	var policies []string
+	if strings.Contains(s.scope.ControlPlane.Spec.Region, v1beta1.DefaultPartitionNameUSGov) {
```
sadysnaat, line 232: Instead of determining this partition all the time using if else we should do it one time and store the result 

and call functions with partition

sadysnaat:
> /lgtm
> 
> @sadysnaat Are you happy with this too? :)

I think we need to refactor the duplicate code and avoid deducing partition using if else all the time 

we should do this once and store in service.

pkg/cloud/services/eks/roles.go
AmitSahastra: Have optimised this code

pkg/cloud/services/eks/roles.go
AmitSahastra: Currently 

> > /lgtm
> > @sadysnaat Are you happy with this too? :)
> 
> I think we need to refactor the duplicate code and avoid deducing partition using if else all the time
> 
> we should do this once and store in service.

Pls take look at the latest code, have made code changes to have util func to get a portion from the region, do we want to update service/crd to store partition ??

AmitSahastra:
> > /lgtm
> > @sadysnaat Are you happy with this too? :)
> 
> I think we need to refactor the duplicate code and avoid deducing partition using if else all the time
> 
> we should do this once and store in service.

Code refactoring is done. Do you have a suggestion on where to store the derived partition in CRD or status or in the scope objects?

cmd/clusterawsadm/api/bootstrap/v1beta1/defaults.go
```diff
@@ -31,6 +31,8 @@ const (
 	DefaultStackName = "cluster-api-provider-aws-sigs-k8s-io"
 	// DefaultPartitionName is the default security partition for AWS ARNs.
 	DefaultPartitionName = "aws"
+	// DefaultPartitionNameUSGov is the default security partition for AWS ARNs.
+	DefaultPartitionNameUSGov = "aws-us-gov"
```
sadysnaat, line 35:
```suggestion
	PartitionNameUSGov = "aws-us-gov"
```

sadysnaat: We can remove Default from DefaultPartitionNameUSGov

sadysnaat (changes requested):

cmd/clusterawsadm/api/bootstrap/v1beta1/defaults.go
AmitSahastra: done

AmitSahastra:
[Skarlso](https://github.com/Skarlso) pls provide the review on the latest code.

Skarlso:
Will take a look today.

Skarlso:
Looks good to me. Let's talk about where to store this @sadysnaat.

sadysnaat:
> Looks good to me. Let's talk about where to store this @sadysnaat.

Would it make sense to have a field in spec to store this info?

Skarlso:
Hm. First, I thought that it should be in the status since it's dynamic. But that also means that we lose this information if the CR is deleted for whatever reason. So, the spec might be a better place indeed.

Ankitasw:
Looks like this is breaking API change in CAPA EKS API.

sadysnaat:
@Ankitasw 

api changes are compatible 

```
sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta2
  Compatible changes:
  - AWSManagedControlPlaneSpec.Partition: added

sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta1
  Compatible changes:
  - AWSManagedControlPlaneSpec.Partition: added
```


and 


```
sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta1
  Compatible changes:
  - AWSClusterSpec.Partition: added

sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2
  Compatible changes:
  - AWSClusterSpec.Partition: added
```


I think by grepping the /api in makefile 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/ef5baabb770749c300c2bc893585b6f2d96d07ee/Makefile#L321


we are expecting only api changes but here other changes are present too 

```
sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/eks
  Incompatible changes:
  - FargateRolePolicies: changed from func() []string to func(string) []string
  - NodegroupRolePolicies: changed from func() []string to func(string) []string
  
  sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope
  Incompatible changes:
  - ClusterScoper.Partition: added
  Compatible changes:
  - (*ClusterScope).Partition: added
  - (*FargateProfileScope).Partition: added
  - (*ManagedControlPlaneScope).Partition: added
  - (*ManagedMachinePoolScope).Partition: added
```

are valid signature and field changes that are done, please let me know if my understanding is not correct.

As this build test was passing with same signature changes before /api changes were made 
job success before api changes 
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/3926/pull-cluster-api-provider-aws-apidiff-main/1605065395755028480

same job failed after api changes 
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/3926/pull-cluster-api-provider-aws-apidiff-main/1605910892107534336

Skarlso:
v1beta1 should not be changed. New entries should only be added to the latest version and conversion function should be written for the new field to be translated from old version to new version.

Skarlso:
Take a look at this pr for an example https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3804/files#diff-010a3c3002f2104a2dc44482ef65502bfe78bac5eef87e8356d588966b919359

Or any pr which touches conversion.go. :)

AmitSahastra:
> Take a look at this pr for an example https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3804/files#diff-010a3c3002f2104a2dc44482ef65502bfe78bac5eef87e8356d588966b919359
> 
> Or any pr which touches conversion.go. :)

Thanks ! was able to fix coversion.go but the test failed so didn't commit changes.

> Take a look at this pr for an example https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3804/files#diff-010a3c3002f2104a2dc44482ef65502bfe78bac5eef87e8356d588966b919359
> 
> Or any pr which touches conversion.go. :)

Thanks! was stuck at the hub-spoke conversion.
This pr needs a good amount of squashing :)

api/v1beta1/awscluster_conversion.go
```diff
@@ -17,40 +17,41 @@ limitations under the License.
 package v1beta1
 
 import (
-	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	infrav2 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
```
Skarlso, line 20: This is wrong sadly. 🤣 I'll include a sample of how this needs to look like. I know the conversion stuff is pretty complex so it's hard to follow what to do with this thing.

Skarlso:
@sadysnaat @AmitSahastra So sorry, but on second thought, I don't think we need to store this info for three reasons:

- It's derived automatically anyways
- We didn't store it in the first place
- If we DO store it, that will generate a diff of templates in tools such as terraform or other diff tracking appliances between the local template and the applied one, since it's changes after it's applied.

I think that last was in the best counterargument as it can cause confusion locally with users.

I suggest to revert the spec part the rest is fine to merge. WDYT?

sadysnaat:
Hi @Skarlso 

my points 

> It's derived automatically, anyways

yes but if we store if we need not to derive it again and again like we do for https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f63dcb936233a0e6c3ba318ab677bafd4de2f5b7/api/v1beta1/network_types.go#L149
where we backfill networking info 


> We didn't store it in the first place

that's because this flow was broken and didn't need derivation, for older versions capa would fail to provision bastion host 


> If we DO store it, that will generate a diff of templates in tools such as terraform or other diff tracking appliances between the local template and the applied one since it changes after it's applied.

I think we do it all the time one mentioned in networkspec other place is https://github.com/kubernetes-sigs/cluster-api/blob/0587101abbc0d558af4f2715b0be536b825742f6/api/v1beta1/cluster_types.go#L53
where endpoint is filled in after cluster provisioning has begun. 

I am not sure how tools handle this but if you have reference to any repo which gets affected would like to see how clusterapi.cluster.endpoint changes are handled there

Skarlso:
Right. We do it all the time but it's not necessarily good. 😁

Okay then. Fine with me.

AmitSahastra:
> Right. We do it all the time but it's not necessarily good. 😁
> 
> Okay then. Fine with me.

@Skarlso @sadysnaat can you provide final review on this.
any action item pending from my side ?

Skarlso:
Okay, if we are doing this, than it's incorrect that v1beta1 has changes in it and the conversion is incorrect too. I'll try to show you later how it can be done.

api/v1beta1/awscluster_conversion.go
Skarlso: I can't read for shit. I thought this renamed the import 🤦 This is actually correct.

Skarlso:
It looks like this needs to be reverted, since this has been merged the periodic e2e unmanaged S3 thing fails.

Skarlso:
@AmitSahastra Could you open this again please in a separate PR so we can run other tests on it?

We need to find out what went wrong. :)

AmitSahastra:
> @AmitSahastra Could you open this again please in a separate PR so we can run other tests on it?
> 
> We need to find out what went wrong. :)

Saw this comment today. have opened new PR https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4010.
Pls take a look.

Skarlso:
Thanks!

PR #3939, opened 2022-12-20
Fix AWS Cluster Tags Deletion & Re-addition issue

vishu2498:
**What type of PR is this?** Bug Fix PR

/kind bug

**What this PR does / why we need it**: This PR fixes the AWS tags deletion and re-addition issue. So, when the cluster is deleted, the subnet will no longer have the tag of that cluster and will also not re-populate those tags while restarting the pod/deployment of controller.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/network/subnets.go
+++ b/pkg/cloud/services/network/subnets.go
@@ -119,6 +119,7 @@ func (s *Service) reconcileSubnets() error {
 			if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
 				buildParams := s.getSubnetTagParams(unmanagedVPC, existingSubnet.ID, existingSubnet.IsPublic, existingSubnet.AvailabilityZone, subnetTags)
 				tagsBuilder := tags.New(&buildParams, tags.WithEC2(s.EC2Client))
+				s.removeNonClusterSharedTags(subnetTags, existingSubnet.Tags)
 				if err := tagsBuilder.Ensure(existingSubnet.Tags); err != nil {
 					return false, err
 				}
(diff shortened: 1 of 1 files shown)


Ankitasw:
Hey @vishu2498 thanks for the PR.
Could you please first raise an issue for the same, so that we could discuss there about root cause.

vishu2498:
Hey @Ankitasw .
This is the issue: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3947

Skarlso:
So, I have been thinking about this and mapping things out on paper. I firmly believe that we shouldn't tag any unmanaged objects. Be that subnet or VPC or whatever. It just generates way too many problems. We will document what kind of tags and labels unmanaged clusters and network topologies should have, and that's it.

It's a burden on the cluster, managing tags all over the place. If you wish to reuse an environment, I opt for the user to clean it up. CAPA doing anything is just prone to a disaster where it deletes something it shouldn't have.

vishu2498:
Hey @Skarlso 
I understand your point of view. However, the way we are leveraging AWS cluster-api in our product, this issue kept occurring and we fixed it just like it has been fixed in this PR.

Yeah it causes a hurdle for CAPA contributors to manage all these things. Please let me know your thoughts and what strategy you are willing to go ahead with.

Ankitasw:
@vishu2498 I agree with @Skarlso. We don't want to manage tagging in unmanaged infra. Apologies for the incovenience it's causing in your env. But I would suggest manual cleanup for the tags in your env as of now, and we could remove the tagging of unmanaged infra such that we dont face such issues, as this issue has been observed by users multiple times. Since this PR is handling removal of tags, and we dont even want to support tagging in first place, I would recommend closing this PR, and if you are interested, try a new PR with removing the tagging during reconcile of unmanaged infra.

PR #3967, opened 2023-01-09
Apply #3871 "Update LoadBalancerReadyCondition on deletion" to release-1.5

mnitchev:
The LoadBalancerReadyCondition is sometimes explicitly patched and other times it only updates the awsCluster object. Furthermore the condition won't be patched by the patch helper in the AWSClusterReconciler, because it's not in the `patch.WithOwnedConditions` list. This change puts the condition in the list and also updates the condition reason to `Deleted` when the LB is not found. Without setting the condition in that case, the reconciler will set condition to `Deleting` and it will never go back to `Deleted`.

Co-authored-by: Jose Armesto <github@armesto.net>

**What type of PR is this?**
/kind bug

This is #3871, but for the release-1.5. The original PR could not be backported due to conflicts.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -38,6 +38,13 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util"
+	capiannotations "sigs.k8s.io/cluster-api/util/annotations"
+	"sigs.k8s.io/cluster-api/util/conditions"
+	"sigs.k8s.io/cluster-api/util/patch"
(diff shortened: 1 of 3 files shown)


AverageMarcus:
cc @richardcase 

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3871 couldn't be cherry picked due to changes in type imports (I think) conflicting. This PR adds the same logic to the 1.5 codebase.

dlipovetsky:
The import grouping needs to be fixed: https://github.com/dlipovetsky/cluster-api-provider-aws/commit/feb6e3236d0eacd003b82a5864d118a562f1d434

I also see the linter complain about a file that isn't touched by this PR: https://github.com/dlipovetsky/cluster-api-provider-aws/commit/2830053ee0f132475b470d1bcdba83057b7baa8c

PR #3978, opened 2023-01-12
Fail creation of machine pool if no subnets matching filters found

### LINKED ISSUE #3977: AWSMachinePool subnet filters ignored if no matches found
/kind bug

Create an `AWSMachinePool` CR with a subnet filter that doesn't match any existing subnet in the account. 
E.g.

```yeaml
subnet:
  filters:
    - name: tag:subnet-role
      values: ["non-existent"]
```

When no subnets are found to match the provided filter the [SubnetIDs](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7411875ed5505b2c3841ae142ffba651b16c93d8/pkg/cloud/services/autoscaling/autoscalinggroup.go#L500) function in the autoscaling service will use an empty array for `subnetIDs` (rather than populating it with the IDs of any subnets found to match the filter). This empty array is then passed on to the [SubnetIDs](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7411875ed5505b2c3841ae142ffba651b16c93d8/pkg/cloud/scope/machinepool.go#L294) function on the scope which mistakenly treats the empty array of subnets IDs as no desired subnets were provided on the `AWSMachinePool` CR. This will then fall back to fetching any subnet that matches the provided availability zone (or the subnets of the `AWSCluster` if no AZs provided).

If no matching subnet is found to match the provided filters an error should be shown instead to avoid mistakenly placing instances into incorrect subnets.

* More discussion can be found in this [Slack thread](https://kubernetes.slack.com/archives/CD6U2V71N/p1673531348507519)
* `AWSMachine` CRs respect the filter and will display an error if no matching subnet is found.
* Proposed solution: have the `SubnetIDs` function of the autoscaling service return an error if `inputFilters` are provided but `subnetIDs` is empty at the end of the function.

**Environment:**

- Cluster-api-provider-aws version: >= 1.5

AverageMarcus:
**What type of PR is this?**

/kind bug

Currently if when using subnet filters with an AWSMachinePool no subnets are found to match the code will fallback to using any matching the AZs specified on the CR. This is unexpected and in contrast to how AWSMachine is handled in the same situation. 
This PR causes an error to be raised if the filters provided result in no subnets matching.

Fixes #3977

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/pkg/cloud/services/autoscaling/autoscalinggroup.go
+++ b/pkg/cloud/services/autoscaling/autoscalinggroup.go
@@ -526,6 +526,12 @@ func (s *Service) SubnetIDs(scope *scope.MachinePoolScope) ([]string, error) {
 		for _, subnet := range out.Subnets {
 			subnetIDs = append(subnetIDs, *subnet.SubnetId)
 		}
+
+		if len(subnetIDs) == 0 {
+			errMessage := fmt.Sprintf("failed to create ASG %q, no subnets available matching criteria %q", scope.Name(), inputFilters)
+			record.Warnf(scope.AWSMachinePool, "FailedCreate", errMessage)
+			return subnetIDs, awserrors.NewFailedDependency(errMessage)
+		}
 	}
 
 	return scope.SubnetIDs(subnetIDs)
--- a/pkg/cloud/services/autoscaling/autoscalinggroup_test.go
+++ b/pkg/cloud/services/autoscaling/autoscalinggroup_test.go
@@ -644,6 +644,26 @@ func TestServiceUpdateASGWithSubnetFilters(t *testing.T) {
 			},
 		},
 		{
+			name:            "should return an error if no matching subnets found",
+			machinePoolName: "update-asg-fail",
+			wantErr:         true,
+			awsResourceReference: []infrav1.AWSResourceReference{
+				{
+					Filters: []infrav1.Filter{
+						{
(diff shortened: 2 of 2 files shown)


pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -526,6 +526,12 @@ func (s *Service) SubnetIDs(scope *scope.MachinePoolScope) ([]string, error) {
 		for _, subnet := range out.Subnets {
 			subnetIDs = append(subnetIDs, *subnet.SubnetId)
 		}
+
+		if len(subnetIDs) == 0 {
```
Skarlso, line 530: Hm. So what happens if in the above there are filters and IDs defined? It might be that the list is not empty so the filter might not return anything, but the list still has a subnet ID in it.

I imagine we want that to proceed though and not care about the filter not working, right?

AverageMarcus: Yeah in that scenario I'd expect it just to make use of the subnets with the matching IDs.

Skarlso:
/lgtm
/approve
/hold

until tests

@AverageMarcus you can unhold if tests passed :)

PR #3981, opened 2023-01-14
Increase CloudFormation stack creation timeout

Skarlso:
**What type of PR is this?**

/kind bug

It looks like the cloudformation stack took a while to get ready. More than 2 minutes in fact.
There were no particular failures in the stack so I'm assuming this is just AWS being slower.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests


Skarlso:
Hmm, actually, I see some creation faileds in the log.

Skarlso:
Yeah, it cloudn't create the CF stack. And the dump failed. I'll take a look at that instead.
/close

PR #3997, opened 2023-01-20
Support specifying subnet filters for API ELB subnets

### LINKED ISSUE #3979: `ControlPlaneLoadBalancer` subnet filters
/kind bug
/kind cleanup

Currently it's possible to create and assign subnets to almost all resources handled by CAPA except for the control plane API server load balancer. The subnet to use can be specified on the `AWSCluster` CR but only by using the subnet IDs rather than being able to use filters like we can for other resources. This means that to use specific subnets for the load balancer they need to be created in advance without the help of CAPA instead of being able to specify them within the `NetworkSpec` of an `AWSCluster` CR.

When no subnet IDs are provided CAPA falls back to using the first subnet in each AZ associated with the `AWSCluster`.

The ability to have CAPA both create the subnets and associate them with the load balancer.

The ideal approach in my opinion would be to use the same approach as on `AWSMachine` and `AWSMachinePool` where its possible to specify subnets ID and/or filters to use to lookup the subnets. This would allow the subnets to be created with CAPA and then use those to associate with the load balancer based on, for example, specific AWS tags added to the subnets.

* Related Slack discussion: https://kubernetes.slack.com/archives/CD6U2V71N/p1673600628090409
* Unfortunately an ideal solution would result in a breaking API change as the [`Subnets` property](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7411875ed5505b2c3841ae142ffba651b16c93d8/api/v1beta2/awscluster_types.go#L194) is a string array where it would be best to be a `infrav1.AWSResourceReference` array to match other resources.

**Environment:**

AverageMarcus:
**What type of PR is this?**

/kind api-change

The subnets for the control plane load balancer can currently only be specified by providing their subnet ID. This is only possible when the subnets are created ahead of the cluster and running in VPC-unmanaged mode.

This change introduces a new `SubnetSpec` field (that would ideally just be called `Subnets` in the next API version) that allows providing subnet filters like we already have on AWSMachine and AWSMachinePool. This means we can create the subnets in CAPA (by defining them on the NetworkSpec of the AWSCluster CR) and still specify which to use for the control plane ELB.

Fixes #3979

At this point, I'd like to just get some thoughts on the approach as I'm not 100% sure how API changes are handled.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/api/v1beta1/zz_generated.conversion.go
+++ b/api/v1beta1/zz_generated.conversion.go
@@ -1200,6 +1200,7 @@ func autoConvert_v1beta2_AWSLoadBalancerSpec_To_v1beta1_AWSLoadBalancerSpec(in *
 	out.Scheme = (*ClassicELBScheme)(unsafe.Pointer(in.Scheme))
 	out.CrossZoneLoadBalancing = in.CrossZoneLoadBalancing
 	out.Subnets = *(*[]string)(unsafe.Pointer(&in.Subnets))
+	// WARNING: in.SubnetFilter requires manual conversion: does not exist in peer-type
 	out.HealthCheckProtocol = (*ClassicELBProtocol)(unsafe.Pointer(in.HealthCheckProtocol))
 	out.AdditionalSecurityGroups = *(*[]string)(unsafe.Pointer(&in.AdditionalSecurityGroups))
 	// WARNING: in.LoadBalancerType requires manual conversion: does not exist in peer-type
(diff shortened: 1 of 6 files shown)


Ankitasw:
@AverageMarcus would you be able to continue with this PR?

AverageMarcus:
(Whoops, accidentally closed PR)

@Ankitasw sorry, I've been unable to do any work on this lately (and my communication has been terrible, I'm sorry)

I'm going to discuss it with my colleagues and see if one of those is able to finish it off for me as I've since moved teams so it's not a priority for me right now.

PR #4000, opened 2023-01-23
fix(#3980): save the API ELB AZ to NetworkStatus

### LINKED ISSUE #3980: CAPA should autodiscover unmanaged LB subnets to enable multi-AZ control planes
/kind feature

CAPA allows users to BYO infrastructure, including ELBs. The algorithm used to select which AZ should be used by control planes appears to be as follows:

1. Fetch the subnet information from the AWSCluster spec and filter it for all of the private subnets
2. Compare the AZs in which the API server ELB lives to the subnet AZs
3. Every match is then marked in the AWSCluster status as a failureDomain capable of housing a control plane, i.e.
```
failureDomain:
    az-name:
        controlPlane: true
```

The above is deduced from https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c0bd2df52fb82e02bd293e3d705b4f363ff20941/controllers/awscluster_controller.go#L343-L355

The subnet spec in the AWSClusterSpec allows the user to set the `isPublic` and `availabilityZone` fields; however, there is no place to specify the AZs in which the ELB lives. Instead, that is autodiscovered.

The algorithm for determining what AZs in which the ELB lives is as follows:

1. Fetch the spec from the NetworkSpec.APIServerELB
2. Use the named stored in the spec to describe the ELB
3. Check if the ELB is managed by CAPA
    1. If the ELB is managed, discover the AZs used by the ELB and copy that information into NetworkStatus
    2. If the ELB is unmanaged, continue

The above was ascertained from the following places:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7549bed37fb8c59c51c7f014a541df11e4585bc1/pkg/cloud/services/elb/loadbalancer.go#L402-L465
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7549bed37fb8c59c51c7f014a541df11e4585bc1/pkg/cloud/services/elb/loadbalancer.go#L446-L448

In essence, I think there's a bit of a bug here. My naive reading of the code seems to indicate that [these three lines](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7549bed37fb8c59c51c7f014a541df11e4585bc1/pkg/cloud/services/elb/loadbalancer.go#L446-L448) should live on line 463. Since the NetworkStatus block already handles an empty array just fine, the only possible states for the APIServerELB's AZ array should be containing a list of AZs or empty so I don't see this change introducing any bugs. In fact, it almost looks like this was the actual intended behavior.

**Environment:**

- Kubernetes version: (use `kubectl version`): v1.25.4

thefirstofthe300:
**What type of PR is this?**
/kind bug

In the case of a BYO load balancer, CAPA does not discover what AZs the LoadBalancer is a part of automatically. A downstream consequence is that control plane nodes are unable to target multiple AZs for placement, leading to all nodes being placed in a single AZ. Instead, CAPA should always be able to determine the AZs in which a load balancer is available and select the proper AZs accordingly.

Fixes #3980

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -443,9 +443,6 @@ func (s *Service) reconcileClassicLoadBalancer() error {
 				return errors.Wrapf(err, "failed to attach apiserver load balancer %q to subnets", apiELB.Name)
 			}
 		}
-		if len(apiELB.AvailabilityZones) != len(spec.AvailabilityZones) {
-			apiELB.AvailabilityZones = spec.AvailabilityZones
-		}
 
 		// Reconcile the security groups from the spec and the ones currently attached to the load balancer
 		if !sets.NewString(apiELB.SecurityGroupIDs...).Equal(sets.NewString(spec.SecurityGroupIDs...)) {
@@ -459,6 +456,10 @@ func (s *Service) reconcileClassicLoadBalancer() error {
 		}
 	} else {
 		s.scope.Trace("Unmanaged control plane load balancer, skipping load balancer configuration", "api-server-elb", apiELB)
+	}
+
+	if len(apiELB.AvailabilityZones) != len(spec.AvailabilityZones) {
+		apiELB.AvailabilityZones = spec.AvailabilityZones
 	}
 
 	// TODO(vincepri): check if anything has changed and reconcile as necessary.


AverageMarcus (commented):
/ok-to-test

AverageMarcus (commented):
/lgtm

PR #4007, opened 2023-01-26
Fix awsmanagedcontrolplane doesn’t get reconciled

kahun:
**What type of PR is this?**

/kind bug

Provider infrastructure is not ready when AWS Managed Control Plane is reconciled for the first time and the cluster creation get stuck for 15 minutes until the reconciliation is forced.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release Notes**:
```
Fix AWSManagedControlPlane reconciliation
```

### DIFF
--- a/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
+++ b/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
@@ -92,6 +92,7 @@ type AWSManagedControlPlaneReconciler struct {
 	WatchFilterValue      string
 	ExternalResourceGC    bool
 	AlternativeGCStrategy bool
+	WaitInfraPeriod       time.Duration
 }
 
 // SetupWithManager is used to setup the controller.
@@ -236,7 +237,7 @@ func (r *AWSManagedControlPlaneReconciler) reconcileNormal(ctx context.Context, 
 		// Wait for the cluster infrastructure to be ready before creating machines
 		if !managedScope.Cluster.Status.InfrastructureReady {
 			managedScope.Info("Cluster infrastructure is not ready yet")
-			return ctrl.Result{}, nil
+			return ctrl.Result{RequeueAfter: r.WaitInfraPeriod}, nil
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
```diff
@@ -232,7 +232,7 @@ func (r *AWSManagedControlPlaneReconciler) reconcileNormal(ctx context.Context,
 		// Wait for the cluster infrastructure to be ready before creating machines
 		if !managedScope.Cluster.Status.InfrastructureReady {
 			managedScope.Info("Cluster infrastructure is not ready yet")
-			return ctrl.Result{}, nil
+			return ctrl.Result{RequeueAfter: 5 * time.Second}, nil
```
richardcase, line 235: I think it's reasonable to requeue and not wait for the resync period.

Shall we make this duration configurable via a flag with 5 mins as the default?

kahun: Hey @richardcase, thanks for the comment, do you mean pass the flag as container argument?

kahun: I've added the flag, please review it and give me feedback.

Also, I think is better to set 1 minute as default, in all my tests, Infrastructure become ready between 30 seconds and one minute

kahun:
Thanks for the ok-to-test.

The error doesn't seem related to the changes.

What do you think about lowering the default to 1 minute?

controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
```diff
@@ -236,7 +237,7 @@ func (r *AWSManagedControlPlaneReconciler) reconcileNormal(ctx context.Context,
 		// Wait for the cluster infrastructure to be ready before creating machines
 		if !managedScope.Cluster.Status.InfrastructureReady {
 			managedScope.Info("Cluster infrastructure is not ready yet")
-			return ctrl.Result{}, nil
```
Skarlso, line 239: nice catch

PR #4023, opened 2023-01-31
Fix `make test` for Apple M1 (darwin-arm64), fail if kubebuilder assets cannot be found

AndiDog:
**What type of PR is this?**

/kind failing-test

- Fail if the specified Kubernetes version binaries cannot be found. Otherwise the test starts and tries to use system-installed binaries by mistake, which it typically wouldn't find. The actual download error is then hidden at the beginning of the test output.
- Bump envtest to Kubernetes 1.24 because then it will work for [kubebuilder on Apple M1 as of this comment](https://github.com/kubernetes-sigs/kubebuilder/issues/2664#issuecomment-1183351948). With this, I found that the `--arch amd64` workaround isn't needed anymore (I'm on Apple M1 and it's working), so I removed the OS/arch distinction.
- Use latest envtest commit because the Makefile used a version from 2021

n/a

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -131,8 +131,8 @@ E2E_SKIP_EKS_UPGRADE ?= "false"
 EKS_SOURCE_TEMPLATE ?= eks/cluster-template-eks-control-plane-only.yaml
 
 # set up `setup-envtest` to install kubebuilder dependency
-export KUBEBUILDER_ENVTEST_KUBERNETES_VERSION ?= 1.23.3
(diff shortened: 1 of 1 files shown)


AverageMarcus:
/lgtm

(assuming tests pass)

Ankitasw:
Actually since this problem arises only in `go test`, it must have been tested by `make test` target, so

/unhold

PR #4025, opened 2023-01-31
Move defaulting before creation of patch helper so that no differences will be detected unnecessarily

AndiDog:
**What type of PR is this?**

/kind bug

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2090/files had introduced defaulting of `AWSCluster` values for valid reasons. However, `.Default()` changes field values and therefore `clusterScope.Close()` would detect differences and send a patch to the Kubernetes API – in every request to the reconciler. Since other controllers may also listen to changes (such as these patch operations) on `AWSCluster` objects, this can lead to a ping-pong situation where the controller keeps getting triggered despite no semantic change is happening. By defaulting the object fields before creating the `clusterScope` helper, we can avoid detecting the defaulting as difference, resulting in no unnecessary patch operations.

n/a

I couldn't find an easy way to write a test for this. Please suggest and I'll do the effort.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -142,6 +142,13 @@ func (r *AWSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return reconcile.Result{}, err
 	}
 
+	// CNI related security groups gets deleted from the AWSClusters created prior to networkSpec.cni defaulting (5.5) after upgrading controllers.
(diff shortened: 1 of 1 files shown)


AverageMarcus:
/lgtm

(assuming tests pass)

PR #4029, opened 2023-02-01
Default value of load balancer type since code does not handle an empty string

AndiDog:
**What type of PR is this?**

/kind bug

On upgrade to CAPA ~ v2.0.2, one of our v1beta1 `AWSCluster` objects was converted to this in v1beta2:

```
$ k get awscluster [...] -o yaml | grep -A4 controlPlaneLoadBalancer
  controlPlaneLoadBalancer:
    crossZoneLoadBalancing: false
    scheme: internet-facing
  identityRef:
    kind: AWSClusterRoleIdentity
```

The field `spec.controlPlaneLoadBalancer.loadBalancerType` is not set. Not sure how that could happen on conversion to v1beta2. Maybe because the v1beta2 CRDs were installed later due to a mistake, and I already had the matching CAPA controller version running, then the defaulting of values failed. Or it's just expected behavior, e.g. due to the conversion `restoreControlPlaneLoadBalancer`.

Anyway... this is a corner case that can and will happen. The controller does not handle `""` in this enum field, and therefore wrongly assumes the network load balancer (v2) type instead of defaulting to classic load balancer as desired:

```
func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.MachineScope, elbScope scope.ELBScope, i *infrav1.Instance) error {
	// [...]
	if elbScope.ControlPlaneLoadBalancer().LoadBalancerType == infrav1.LoadBalancerTypeClassic {
		// [...] attach instance to classic LB [...] 
	}
	
	// [...] else attach instance to network load balancer (v2) [...] 
}
```

We should therefore explicitly default the relevant fields. The test uses a fake client to simulate this case. A real client would default the field to `"classic"` on create/update and therefore wouldn't allow testing this. (see [Slack thread](https://kubernetes.slack.com/archives/CD6U2V71N/p1675248584015329) – @Skarlso)

n/a

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests


controllers/awsmachine_controller.go
```diff
@@ -822,12 +822,12 @@ func (r *AWSMachineReconciler) registerInstanceToClassicLB(machineScope *scope.M
 
 	if err := elbsvc.RegisterInstanceWithAPIServerELB(i); err != nil {
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedAttachControlPlaneELB",
-			"Failed to register control plane instance %q with load balancer: %v", i.ID, err)
+			"Failed to register control plane instance %q with classic load balancer: %v", i.ID, err)
```
Skarlso, line 825: This is delibarately not mentioning `classic` because once the "classic" load balancer expires, we don't have to go around and change all the log messages. :) Besides, from the user's perspective, it shouldn't matter. It's just a load balancer.

controllers/awsmachine_controller.go
```diff
@@ -836,8 +836,8 @@ func (r *AWSMachineReconciler) registerInstanceToV2LB(machineScope *scope.Machin
 	_, registered, err := elbsvc.IsInstanceRegisteredWithAPIServerLB(instance)
 	if err != nil {
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedAttachControlPlaneELB",
-			"Failed to register control plane instance %q with load balancer: failed to determine registration status: %v", instance.ID, err)
-		return errors.Wrapf(err, "could not register control plane instance %q with load balancer - error determining registration status", instance.ID)
+			"Failed to register control plane instance %q with network load balancer: failed to determine registration status: %v", instance.ID, err)
```
Skarlso, line 839: Same here, since there are a multitude of load balancer types, this is a generic message.

controllers/awsmachine_controller.go
```diff
@@ -1025,6 +1025,12 @@ func (r *AWSMachineReconciler) getInfraCluster(ctx context.Context, log *logger.
 		return nil, nil //nolint:nilerr
 	}
 
+	// Set default for `ControlPlaneLoadBalancer.LoadBalancerType` since otherwise value might
+	// be an empty string (which we don't handle).
+	// The defaulting must happen before `NewClusterScope` is called since otherwise we keep detecting
+	// differences that result in patch operations.
+	awsCluster.Default()
```
Skarlso, line 1032: This is interesting. I wonder if this would potentially overwrite anything. But I believe it has a couple nil checks in it, right?

controllers/awsmachine_controller_unit_test.go
```diff
@@ -2367,6 +2371,220 @@ func TestAWSMachineReconcilerReconcile(t *testing.T) {
 	}
 }
 
+func TestAWSMachineReconcilerReconcileWithoutLoadBalancerType(t *testing.T) {
```
Skarlso, line 2374: Nice test. Well done. :)

controllers/awsmachine_controller.go
AndiDog: Right, the v2 LB could be of different type. I made the wrong assumption here.

My point here was to clearly distinguish the log messages, since that would've led me into the right direction earlier, namely that the controller tried to attach to a v2 LB, not a classic LB. In general, it's much easier to have unique error messages. So can we keep at least the word "classic" inside `registerInstanceToClassicLB` since that one is specifically for that load balancer type!?

controllers/awsmachine_controller.go
AndiDog: See `SetObjectDefaults_AWSCluster` and the functions it calls. I considered it very safe.

Skarlso: I meant this in the end:
```go
func SetDefaults_AWSClusterSpec(s *AWSClusterSpec) { //nolint:golint,stylecheck
	if s.IdentityRef == nil {
		s.IdentityRef = &AWSIdentityReference{
			Kind: ControllerIdentityKind,
			Name: AWSClusterControllerIdentityName,
		}
	}
	if s.ControlPlaneLoadBalancer == nil {
		s.ControlPlaneLoadBalancer = &AWSLoadBalancerSpec{
			Scheme: &ELBSchemeInternetFacing,
		}
	}
	if s.ControlPlaneLoadBalancer.LoadBalancerType == "" {
		s.ControlPlaneLoadBalancer.LoadBalancerType = LoadBalancerTypeClassic
	}
}
```

Which will eventually set things up. I just didn't remember at the top of my head if it contained any checks to not overwrite existing things. :)

luthermonson: random question as I just did this same debugging... @AndiDog i found out i was using the v2.0.2 crd validations which come with `clusterctl init --infrastructure=aws` with latest code which uses the LB type. the value wasn't persisting due to the validations and cluster controller's defer to patch the defaults into the cluster object https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awscluster_controller.go#L202-L206 were being ignored. can you `kubectl describe crd awscluster` and see if your validations in there for v1beta2 don't have the loadBalancerType field? you might just be running into the same thing I was and we dont have to put the defaults to awscluster in a machine controller.

AndiDog: Hmm `loadBalancerType` was not part of v2.0.2 CRDs but was added later. I have the CRDs without that field.

```
$ git diff v2.0.2..7e87185a82068dc22d2eb7b03362639829c85aed -- config/crd | grep "loadBalancerType:"
+                      loadBalancerType:
+                      loadBalancerType:
+                  loadBalancerType:
+                      loadBalancerType:
+                          loadBalancerType:
```

@luthermonson so what would happen if I upgrade to the very latest CRDs? Would my already-stored, somewhat-invalid object get fixed? Or would the controller fail to `Get` the object because it's now invalid (CRD tells me that `loadBalancerType` has no default i.e. is required).

luthermonson: the problem is running v2.0.2 certs but with latest code from main. if you run v2.0.2 with the v2.0.2 released code everything is fine or if you run the latest main with the same crds in main it should be fine. if you upgrade to the latest CRDs the defer patch will set loadBalancerType (instead of silent error) and you will not see the behavior where it boots and creates a classic load balancer and tries to assign the control plane node to a v2 load balancer.

this is a calamity of errors... 
1) patch should return an error to the logs to say "loadBalancerType" has errored, instead it eats the validation error. this would help devs figure out what is wrong faster, i had to turn on a debugger and step through everything to get to this idea
2) the developer guides dont tell developers how to install the latest CRDs and that should be fixed, most will do the `clusterctl init` method from the quickstart and not realized they locked themselves to a release crd, might be interesting if we could get `clusterctl init --infrastructure=aws:main@latest or something but we'd have to checkin an infrastructure-components.yaml whenever we update the CRDs
3) i dont feel the machine controller should be in the business of setting the cluster defaults so unfortunately, I dont think this PR should be merged
4) i also dont think the cluster controller should be in the business of setting his own defaults when we have webhooks already...

possible better fixes
1) have the mutating webhook set the default for "classic" for the lb type instead of relying on the cluster controller to get to it, it should also validate it's nlb/elb/alb too in the validating webhook
2) flip logic in reconcileLBAttachment where classic is the default and the if check is for the v2 balancers, this will default classic if it's an empty string or anything else and your tests will work

Skarlso: I tend to agree with Luther on this. 🤔  I have to think it through. :)

AndiDog: Very good explanations, thanks. @luthermonson since I'm getting to know the code base right now, where exactly did you find that patching swallows errors? I anyway need to fix the CRD version choice on our side and will therefore test the controller against latest CRDs soon. Until we have some ideas, we can just leave this PR pending. Probably it won't be necessary (I didn't test in prod 🤓).

AndiDog: If you agree with the above proposal, here's the PR for it: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4033

Ankitasw:
@AndiDog can you close this PR since you have another PR for the same?

controllers/awsmachine_controller.go
luthermonson: you dont like the idea of the mutating webhook setting the default and the validating webhook checking that the value matches one of the lb types? i think that would nip all the corner cases in the bud

AndiDog:
Superseded by https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4033

controllers/awsmachine_controller.go
AndiDog: Webhooks only run on create/update, so that wouldn't solve my use case. And shouldn't the hook already validate to some degree because of `// +kubebuilder:validation:Enum:=classic;elb;alb;nlb`? (Sorry, didn't investigate that.)

PR #4030, opened 2023-02-01
Ensure tags on managed VPCs

Skarlso:
**What type of PR is this?**

/kind bug

Ensure tags on managed VPCs. If `AdditionalTags` change, they should be applied to the VPC.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4031

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/pkg/cloud/services/network/vpc.go
+++ b/pkg/cloud/services/network/vpc.go
@@ -66,6 +66,20 @@ func (s *Service) reconcileVPC() error {
 			return nil
 		}
 
+		// Make sure tags are up-to-date.
+		// **Only** do this for managed VPCs. Make sure this logic is below the above `vpc.IsUnmanaged` check.
+		if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+			buildParams := s.getVPCTagParams(s.scope.VPC().ID)
+			tagsBuilder := tags.New(&buildParams, tags.WithEC2(s.EC2Client))
+			if err := tagsBuilder.Ensure(s.scope.VPC().Tags); err != nil {
+				return false, err
+			}
+			return true, nil
+		}, awserrors.VPCNotFound); err != nil {
+			record.Warnf(s.scope.InfraCluster(), "FailedTagVPC", "Failed ensure managed VPC %q: %v", s.scope.VPC().ID, err)
+			return errors.Wrapf(err, "failed to ensure tags on vpc %q", s.scope.VPC().ID)
+		}
+
 		// if the VPC is managed, make managed sure attributes are configured.
 		if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
 			if err := s.ensureManagedVPCAttributes(vpc); err != nil {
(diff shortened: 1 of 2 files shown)


Skarlso:
Manual test incoming...

Skarlso:
Created a cluster with the following Managed cluster setting:

```yaml
---
kind: AWSManagedControlPlane
metadata:
  name: "managed-cluster-control-plane"
spec:
  additionalTags:
    testtag: update-this
  region: "eu-central-1"

```

This produced this VPC:

Screenshot 2023-02-01 at 17 32 46

Updated the tag to this:

```yaml
---
kind: AWSManagedControlPlane
metadata:
  name: "managed-cluster-control-plane"
spec:
  additionalTags:
    testtag: update-this
    updated: this-is-new
  region: "eu-central-1"

```

Eventually reconciled the tag correctly on the VPC:

Screenshot 2023-02-01 at 17 47 16


I see no errors in the controller.

pkg/cloud/services/network/vpc.go
```diff
@@ -66,6 +66,20 @@ func (s *Service) reconcileVPC() error {
 			return nil
 		}
 
+		// Make sure tags are up-to-date.
+		// **Only** do this for managed VPCs. Make sure this logic is below the above `vpc.IsUnmanaged` check.
```
Ankitasw, line 70: Liking the disclaimer 👏

Ankitasw, line 70: Shall we not apply additionalTags to the unmanaged VPC? AdditionalTags are user defined tags that could be applied to managed/unmanaged VPC right?
Please don't mind if I don't have any prior context.

Skarlso: Hmmm. I kind of want to avoid tagging unmanaged resources. This doesn't discriminate between user defined tags either. Because of the previous tags assignment. If any user adds a tag on the vpc itself this will nuke it and override the whole thing with what's set on the cluster and the additional tags. I think... :D but let me make sure about that assumption.

Skarlso: Ok, so this actually worked since there is a describe in front of this where it gets the tag and then does a diff. I'll push and run some tests, hopefully nothing will break. :D

Skarlso: Nooooo no no no this actually didn't work at all. It completely messed up the tags on the unmanaged VPC. :D So reverting to the previous one. :)

Ankitasw: But is there a way we can somehow do this for unmanaged VPC as well, as additionalTags can be applied to both managed/unmanaged VPC

Skarlso: I don't see how. I don't think we should manage tags on an existing VPC. Additional Tags are no exception. If the user defines a VPC we shouldn't touch it even if there are additional tags defined. It just means trouble. And when I tried it it got into all sorts of problems when I started changing tags on the existing VPC that CAPA should not have managed.

Ankitasw: Hmm makes sense to me. 🙂 
We can see it later if some request comes regarding this.

Skarlso: Sounds good! :)

PR #4033, opened 2023-02-02
Ensure empty `loadBalancerType` field value is handled correctly

AndiDog:
**What type of PR is this?**

/kind bug

If outdated CRD versions (in my case: v2.0.2) were used against a newer controller (`main` branch), the `AWSCluster` object could be stored without a default value for `AWSCluster.Spec.ControlPlaneLoadBalancer.LoadBalancerType`. The machine controller handles that incorrectly: it tries to register instances with a v2 LB, but should default to a classic LB.

The previous fix attempt is in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4029, but since we found the root cause, I'd like to suggest _this_ PR's solution instead since `awsCluster.Default()` is an undesired workaround.

n/a

Summoning previous reviewers from the other PR:

/assign @Skarlso 
/assign @luthermonson 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -798,22 +798,31 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 		return r.deregisterInstanceFromV2LB(machineScope, elbsvc, i)
 	}
 
-	// This changes the flow because previously it didn't care about this part.
-	if elbScope.ControlPlaneLoadBalancer().LoadBalancerType == infrav1.LoadBalancerTypeClassic {
+	switch elbScope.ControlPlaneLoadBalancer().LoadBalancerType {
+	case infrav1.LoadBalancerTypeClassic:
+		fallthrough
+	case "":
 		machineScope.Debug("registering to classic load balancer")
 		return r.registerInstanceToClassicLB(machineScope, elbsvc, i)
+
+	case infrav1.LoadBalancerTypeELB:
+		fallthrough
+	case infrav1.LoadBalancerTypeALB:
+		fallthrough
+	case infrav1.LoadBalancerTypeNLB:
+		machineScope.Debug("registering to v2 load balancer")
+		return r.registerInstanceToV2LB(machineScope, elbsvc, i)
 	}
 
-	machineScope.Debug("registering to v2 load balancer")
-	return r.registerInstanceToV2LB(machineScope, elbsvc, i)
+	return errors.Errorf("unknown load balancer type %q", elbScope.ControlPlaneLoadBalancer().LoadBalancerType)
 }
 
 func (r *AWSMachineReconciler) registerInstanceToClassicLB(machineScope *scope.MachineScope, elbsvc services.ELBInterface, i *infrav1.Instance) error {
 	registered, err := elbsvc.IsInstanceRegisteredWithAPIServerELB(i)
 	if err != nil {
 		r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedAttachControlPlaneELB",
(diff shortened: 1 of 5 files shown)


luthermonson:
answering your last comments from the previous PR here so we keep them in one place...

the kubebuilder tags just help generate those openapi schemas on the CRDs... and you make a good point there is something technically missing from the loadBalancerType which is...

```
// +kubebuilder:default=classic
```

maybe add that to your PR and run go generate and see if that takes care of most of issues? not sure i understand what your use case is that is different from an create/update, if you are migrating from v1beta1 to v1beta2 you're still performing updates on the custom resources and they'll go through validation/mutating webhooks... there is also an upgrade that capa covers and you could add [conversions](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta1/awscluster_conversion.go)

AndiDog:
> the kubebuilder tags just help generate those openapi schemas on the CRDs... and you make a good point there is something technically missing from the loadBalancerType which is...
> 
> ```
> // +kubebuilder:default=classic
> ```
> 
> maybe add that to your PR and run go generate and see if that takes care of most of issues? not sure i understand what your use case is that is different from a create/update, if you are migrating from v1beta1 to v1beta2 you're still performing updates on the custom resources and they'll go through validation/mutating webhooks... there is also an upgrade that capa covers and you could add [conversions](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta1/awscluster_conversion.go)

Added the kubebuilder annotation. Good hint!

My (mis)use case is when stuff already went wrong, so the invalid object has already been stored with an empty string field. The switch statement now explicitly handles that, and errors loudly on unsupported values. I think that's a very good solution (from idea 2 of your earlier comments) and not even a workaround.

controllers/awsmachine_controller.go
```diff
@@ -798,22 +798,31 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 		return r.deregisterInstanceFromV2LB(machineScope, elbsvc, i)
 	}
 
-	// This changes the flow because previously it didn't care about this part.
-	if elbScope.ControlPlaneLoadBalancer().LoadBalancerType == infrav1.LoadBalancerTypeClassic {
+	switch elbScope.ControlPlaneLoadBalancer().LoadBalancerType {
+	case infrav1.LoadBalancerTypeClassic:
+		fallthrough
+	case "":
```
Skarlso, line 804: With the default now set through kube is this even possible?

AndiDog: I've documented the corner case in the test, and tried to describe it in the PR summary. What Luther helped me find out is this: I had v2.0.2 CRDs installed, but v2.0.2+X controller running. This led to an `AWSCluster` being stored in etcd without having a value for `loadBalancerType` (i.e. empty string). Since a mutating webhook only runs on create/update, but the controller has no reason to perform such a change, the controller still runs with the object with the empty field value. The solution is therefore easy: support that here. Very likely, only few people will even hit this, but the chances exist. Having the wrong CRD versions installed doesn't even have to be a human error – it could even happen when some "crd-install" job technically fails but the new controller version starts up anyway. It's just a risk that we should cover.

luthermonson:
love it 👍 from me

controllers/awsmachine_controller.go
Skarlso: Right. I just thought that the annotation might have fixed that. But I see now. :)

Skarlso:
/lgtm
/hold
/approve

hold until tests

Skarlso:
Yep, sadly, this looks like a releated failure.

AndiDog:

Trying again since the cluster readiness took 30 minutes as of the logs.

PR #4005, opened 2023-01-26
Move all E2E test templates to use external CCM and CSI

wyike:
**What type of PR is this?**

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2941 

**Special notes for your reviewer**:

Remaining work:
1. bump CCM and CSI version
2. Update templates under test/e2e/data/infrastructure-aws/withoutclusterclass/e2e_test_templates and test/e2e/data/infrastructure-aws/withclusterclass/e2e_test_templates, after this PR merged
3. Note: one testcase using capi-upgrades https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/test/e2e/data/infrastructure-aws/capi-upgrades/v1beta1 is still using in tree ccm as the upgrade-from cluster template. (I cannot recall the exact test case name, but I remember there is such one)
4. Didn't take care `PDescribe` testcases

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [x] adds or updates e2e tests

PR #4051, opened 2023-02-08
Revert "Move all E2E test templates to use external CCM and CSI"

Ankitasw:
Reverts kubernetes-sigs/cluster-api-provider-aws#4005

Testing if E2E failures has something to do with this change.


Ankitasw:
cc @wyike
We are facing E2E failures much more frequently after this was merged. I am trying to test if this was actually the culprit, and then find the root cause for the same.

Ankitasw:
pull-cluster-api-provider-aws-e2e-blocking

PR #3208, opened 2022-02-12
Allow arm64 image lookup based on instance type

khoavo-hatch:
**What type of PR is this?**

This is a bug fix PR to support looking up AMIs for arm64 machines.

/kind bug

**What this PR does / why we need it**:

Currently the logic for looking up AMIs does not take CPU architecture into account:
- The default AMI lookup hard-codes the architecture to x86_64.
- The EKS-optimized image lookup also only fetch the SSM parameter for x86 images.

So this PR adds logic to take into account the instance type's supported cpu architecture, so we will lookup images for the correct architecture.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3121

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- includes documentation (this is kind of a like a bug fix, so I don't think it needs docs)
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Action required: Added ability to look up x86_64 or arm64 images based on the instance type (#3121). Please rerun `clusterawsadm bootstrap iam create-cloudformation-stack` to give the controller permission to describe instance types.
```

PR #4054, opened 2023-02-08
✨ support arm64 AMI lookup based on instance type

### LINKED ISSUE #3121: Support looking up EKS optimized image for ARM
/kind bug

When creating an EKS based cluster if we don't specify any of the image properties we [look up the default EKS optimized  AMIs using SSM]. 

We only support looking up the base AL2 [EKS ami](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/50af6ec5ce7a183b67d23a187e687c42fef24176/pkg/cloud/services/ec2/ami.go#L59) and the [GPU optimized AMI](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/50af6ec5ce7a183b67d23a187e687c42fef24176/pkg/cloud/services/ec2/ami.go#L62).

To support looking up EKS optimized images for ARM we need to use the following format when querying SSM:

`/aws/service/eks/optimized-ami/%s/amazon-linux-2-arm64/recommended/image_id`

This will require that we know we are using ARM based instances when looking up the images.

I would expect to be able to use EKS optimized ARM images without having to explicitly specify the image to look up.
 
Created as a result of a [discussion](https://kubernetes.slack.com/archives/CD6U2V71N/p1643608696167349) on the community slack.

**Environment:**

- Cluster-api-provider-aws version: 1.2.0

charlie-haley:
**What type of PR is this?**
/kind feature 

This PR is a continuation of the work done in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3208

Fixes #3121

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

### DIFF
--- a/cmd/clusterawsadm/ami/copy.go
+++ b/cmd/clusterawsadm/ami/copy.go
@@ -57,7 +57,7 @@ func Copy(input CopyInput) (*amiv1.AWSAMI, error) {
 	}
 	ec2Client := ec2.New(sourceSession)
 
-	image, err := ec2service.DefaultAMILookup(ec2Client, input.OwnerID, input.OperatingSystem, input.KubernetesVersion, "")
+	image, err := ec2service.DefaultAMILookup(ec2Client, input.OwnerID, input.OperatingSystem, input.KubernetesVersion, ec2service.Amd64ArchitectureTag, "")
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/cloud/services/ec2/ami.go
+++ b/pkg/cloud/services/ec2/ami.go
@@ -36,6 +36,12 @@ import (
 )
 
 const (
+	// Amd64ArchitectureTag is the reference AWS uses for amd64 architecture images.
+	Amd64ArchitectureTag = "x86_64"
+
+	// Arm64ArchitectureTag is the reference AWS uses for arm64 architecture images.
+	Arm64ArchitectureTag = "arm64"
+
 	// DefaultMachineAMIOwnerID is a heptio/VMware owned account. Please see:
 	// https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/487
 	DefaultMachineAMIOwnerID = "258751437250"
@@ -64,6 +70,9 @@ const (
 
 	// EKS AMI ID SSM Parameter name.
 	eksAmiSSMParameterFormat = "/aws/service/eks/optimized-ami/%s/amazon-linux-2/recommended/image_id"
+
+	// EKS ARM64 AMI ID SSM Parameter name.
+	eksARM64AmiSSMParameterFormat = "/aws/service/eks/optimized-ami/%s/amazon-linux-2-arm64/recommended/image_id"
 
 	// EKS GPU AMI ID SSM Parameter name.
 	eksGPUAmiSSMParameterFormat = "/aws/service/eks/optimized-ami/%s/amazon-linux-2-gpu/recommended/image_id"
@@ -94,8 +103,50 @@ func GenerateAmiName(amiNameFormat, baseOS, kubernetesVersion string) (string, e
 	return templateBytes.String(), nil
 }
 
+// Determine architecture based on instance type.
+func (s *Service) pickArchitectureForInstanceType(instanceType string) (string, error) {
+	descInstanceTypeInput := &ec2.DescribeInstanceTypesInput{
+		InstanceTypes: []*string{&instanceType},
+	}
+	describeInstanceTypeResult, err := s.EC2Client.DescribeInstanceTypes(descInstanceTypeInput)
+	if err != nil {
+		return "", err
+	}
+
+	if len(describeInstanceTypeResult.InstanceTypes) == 0 {
+		return "", fmt.Errorf("instance type result empty for type %q", instanceType)
+	}
+
+	supportedArchs := describeInstanceTypeResult.InstanceTypes[0].ProcessorInfo.SupportedArchitectures
+
+	logger := s.scope.GetLogger().WithValues("instance type", instanceType, "supported architectures", supportedArchs)
+	logger.Info("Obtained a list of supported architectures for instance type")
+
+	// Loop over every supported architecture for the instance type
+	architecture := ""
+archCheck:
+	for _, a := range supportedArchs {
+		switch *a {
+		case Amd64ArchitectureTag:
+			architecture = *a
+			break archCheck
+		case Arm64ArchitectureTag:
+			architecture = *a
+			break archCheck
+		}
+	}
+
+	if architecture == "" {
+		return "", fmt.Errorf("unable to find preferred architecture for instance type %q", instanceType)
+	}
+
+	logger.Info("Chosen architecture", "architecture", architecture)
+
+	return architecture, nil
+}
+
 // DefaultAMILookup will do a default AMI lookup.
-func DefaultAMILookup(ec2Client ec2iface.EC2API, ownerID, baseOS, kubernetesVersion, amiNameFormat string) (*ec2.Image, error) {
+func DefaultAMILookup(ec2Client ec2iface.EC2API, ownerID, baseOS, kubernetesVersion, architecture, amiNameFormat string) (*ec2.Image, error) {
 	if amiNameFormat == "" {
 		amiNameFormat = DefaultAmiNameFormat
 	}
@@ -122,7 +173,7 @@ func DefaultAMILookup(ec2Client ec2iface.EC2API, ownerID, baseOS, kubernetesVers
 			},
 			{
 				Name:   aws.String("architecture"),
-				Values: []*string{aws.String("x86_64")},
+				Values: []*string{aws.String(architecture)},
 			},
 			{
 				Name:   aws.String("state"),
@@ -151,10 +202,10 @@ func DefaultAMILookup(ec2Client ec2iface.EC2API, ownerID, baseOS, kubernetesVers
 }
 
 // defaultAMIIDLookup returns the default AMI based on region.
-func (s *Service) defaultAMIIDLookup(amiNameFormat, ownerID, baseOS, kubernetesVersion string) (string, error) {
-	latestImage, err := DefaultAMILookup(s.EC2Client, ownerID, baseOS, kubernetesVersion, amiNameFormat)
+func (s *Service) defaultAMIIDLookup(amiNameFormat, ownerID, baseOS, architecture, kubernetesVersion string) (string, error) {
+	latestImage, err := DefaultAMILookup(s.EC2Client, ownerID, baseOS, kubernetesVersion, architecture, amiNameFormat)
 	if err != nil {
-		record.Eventf(s.scope.InfraCluster(), "FailedDescribeImages", "Failed to find ami for OS=%s and Kubernetes-version=%s: %v", baseOS, kubernetesVersion, err)
+		record.Eventf(s.scope.InfraCluster(), "FailedDescribeImages", "Failed to find ami for OS=%s, Architecture=%s and Kubernetes-version=%s: %v", baseOS, architecture, kubernetesVersion, err)
 		return "", errors.Wrapf(err, "failed to find ami")
 	}
 
@@ -234,7 +285,7 @@ func (s *Service) defaultBastionAMILookup() (string, error) {
 	return *latestImage.ImageId, nil
 }
 
-func (s *Service) eksAMILookup(kubernetesVersion string, amiType *infrav1.EKSAMILookupType) (string, error) {
+func (s *Service) eksAMILookup(kubernetesVersion string, architecture string, amiType *infrav1.EKSAMILookupType) (string, error) {
 	// format ssm parameter path properly
 	formattedVersion, err := formatVersionForEKS(kubernetesVersion)
 	if err != nil {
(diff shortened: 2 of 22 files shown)


cmd/clusterawsadm/ami/copy.go
```diff
@@ -57,7 +57,7 @@ func Copy(input CopyInput) (*amiv1.AWSAMI, error) {
 	}
 	ec2Client := ec2.New(sourceSession)
 
-	image, err := ec2service.DefaultAMILookup(ec2Client, input.OwnerID, input.OperatingSystem, input.KubernetesVersion, "")
+	image, err := ec2service.DefaultAMILookup(ec2Client, input.OwnerID, input.OperatingSystem, input.KubernetesVersion, "x86_64", "")
```
Ankitasw, line 60: In the func definition, the architecture param is added at no.4 param, but here it's been passed at no.5.

pkg/cloud/services/ec2/ami.go
```diff
@@ -94,8 +97,51 @@ func GenerateAmiName(amiNameFormat, baseOS, kubernetesVersion string) (string, e
...
+
+	supportedArchs := describeInstanceTypeResult.InstanceTypes[0].ProcessorInfo.SupportedArchitectures
+
+	logger := s.scope.GetLogger().WithValues("instance type", instanceType, "supported architectures", supportedArchs)
+	logger.Info("Obtained a list of supported architectures for instance type")
+
+	architecture := ""
+	switch len(supportedArchs) {
+	case 1:
+		architecture = *(supportedArchs[0])
```
Ankitasw, line 127: What would be `supportedArchs[0]` in this case? why is it distinguished based on the length of `supportedArchs`?

charlie-haley: This was something I took from the old PR, I've just pushed a commit to change how this is handled. Instead of doing this elaborate check I've made it more explicit and easier to read.
This does mean if in the future more architecture types are added (maybe `riscv64`?) we will have to manually update this switch statement, which I don't think is a big deal.

A little context on why it's checking the length - this exists is because some instance types have multiple supported architectures. The only notable case I could find is where `i386` and `x86_64` are both supported, it also only seems to be old `t2.` instance types that have this issue.

Ankitasw: Ok great that makes sense to me.

Ankitasw:
@charlie-haley I think the test is failing continuously in every run, we have to debug. If you are not able to, I would do that later today/tomorrow.

Ankitasw:
@charlie-haley Could you rebase your PR.
Also we are getting below error which may be arising as AMI is not populated correctly.

> I0228 13:16:58.758890       1 logger.go:71] "Looking for existing LaunchTemplates" controller="awsmanagedmachinepool" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSManagedMachinePool" AWSManagedMachinePool="cluster-zhvhyf/cluster-qwnn1j-pool-lt-0" namespace="cluster-zhvhyf" name="cluster-qwnn1j-pool-lt-0" reconcileID=1de0c1c2-ab03-48d4-819d-243081f2a953 MachinePool="cluster-zhvhyf/cluster-qwnn1j-pool-lt-0" cluster="cluster-zhvhyf/cluster-qwnn1j"
E0228 13:16:59.371199       1 logger.go:83] "identity: failed to reconcile launch template" err=<
	InvalidInstanceType: The following supplied instance types do not exist: []
		status code: 400, request id: 7f13aa1f-7b58-45b7-98fb-e2949636fc48
 > identityKey="/e2e-account"
I0228 13:16:59.371286       1 recorder.go:103] "events: Failed to reconcile launch template: InvalidInstanceType: The following supplied instance types do not exist: []\n\tstatus code: 400, request id: 7f13aa1f-7b58-45b7-98fb-e2949636fc48" type="Warning" object={Kind:AWSManagedMachinePool Namespace:cluster-zhvhyf Name:cluster-qwnn1j-pool-lt-0 UID:28c2a31f-247c-485d-866f-3ad8fe8704a7 APIVersion:infrastructure.cluster.x-k8s.io/v1beta2 ResourceVersion:15291 FieldPath:} reason="FailedLaunchTemplateReconcile"

pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -749,13 +750,35 @@ func (s *Service) DiscoverLaunchTemplateAMI(scope scope.LaunchTemplateScope) (*s
 		imageLookupBaseOS = scope.GetEC2Scope().ImageLookupBaseOS()
 	}
 
+	instanceType := lt.InstanceType
+	// If instance type is not specified on a launch template, we can safely assume the instance type will be a `t3.medium`.
```
Ankitasw, line 754: Since this should be input from users, we could fix this in the E2E test templates/cluster configs rather than hardcoding here.

pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -749,13 +750,35 @@ func (s *Service) DiscoverLaunchTemplateAMI(scope scope.LaunchTemplateScope) (*s
...
+	// As specified in the AWS docs https://docs.aws.amazon.com/eks/latest/userguide/launch-templates.html.
+	if instanceType == "" {
+		instanceType = "t3.medium"
+	}
```
charlie-haley, line 758: I found the issue with the AMI population @Ankitasw, it looks like when running the tests with a launch template, the `instanceType` field is left empty, this is because AWS automatically populates this with `t3.medium` on their end when using a launch template.

I've added a check here to just assume the instanceType is a `t3.medium` in this case.

pkg/cloud/services/ec2/launchtemplate.go
charlie-haley: I'll move my comment over here as it'll be easier:

When running the tests with a launch template, the instanceType field is left empty, this is because AWS automatically populates this with t3.medium on their end when using a launch template.

I've added a check here to just assume the instanceType is a t3.medium in this case.

Obviously this is hardcoded and AWS could change this value at some point, do you think instead of assuming the instance type, we should just default the architecture x86_64 if the instance type is empty instead?

Ankitasw: > this is because AWS automatically populates this with t3.medium on their end when using a launch template.

Is this always set to m3.medium? Is it not taken as inout from user in launch template?

charlie-haley: According to AWS it's always set to `t3.medium`, unless a user specifies an instance type in their launch template, then it's overridden. I'll push a change to default the architecture instead as that was the previous behavior

Ankitasw:
/lgtm

Ankitasw:
cc @Skarlso @dlipovetsky for another review.

Ankitasw:
@richardcase would you be able to take a look at this, needed another pair of 👀 for review.

pkg/cloud/services/ec2/ami.go
```diff
@@ -94,8 +103,43 @@ func GenerateAmiName(amiNameFormat, baseOS, kubernetesVersion string) (string, e
...
+	}
+
+	supportedArchs := describeInstanceTypeResult.InstanceTypes[0].ProcessorInfo.SupportedArchitectures
+
+	logger := s.scope.GetLogger().WithValues("instance type", instanceType, "supported architectures", supportedArchs)
+	logger.Info("Obtained a list of supported architectures for instance type")
+
+	// Loop over every supported architecture for the instance type
+	architecture := ""
+	for _, a := range supportedArchs {
```
Skarlso, line 123: Once it's found we should `break`.

Skarlso: I'm... not even sure this is correct, right? This looks weird. This means if ARM is the last thing, it will choose that. But if it happens to come before AMD then and might be selected last, right? This will select whatever comes last in the list.

pkg/cloud/services/ec2/ami.go
```diff
@@ -151,8 +195,8 @@ func DefaultAMILookup(ec2Client ec2iface.EC2API, ownerID, baseOS, kubernetesVers
 }
 
 // defaultAMIIDLookup returns the default AMI based on region.
-func (s *Service) defaultAMIIDLookup(amiNameFormat, ownerID, baseOS, kubernetesVersion string) (string, error) {
-	latestImage, err := DefaultAMILookup(s.EC2Client, ownerID, baseOS, kubernetesVersion, amiNameFormat)
+func (s *Service) defaultAMIIDLookup(amiNameFormat, ownerID, baseOS, architecture, kubernetesVersion string) (string, error) {
+	latestImage, err := DefaultAMILookup(s.EC2Client, ownerID, baseOS, kubernetesVersion, architecture, amiNameFormat)
 	if err != nil {
 		record.Eventf(s.scope.InfraCluster(), "FailedDescribeImages", "Failed to find ami for OS=%s and Kubernetes-version=%s: %v", baseOS, kubernetesVersion, err)
```
Skarlso, line 201: Add architecture to the log.

pkg/cloud/services/ec2/ami.go
```diff
@@ -251,7 +295,14 @@ func (s *Service) eksAMILookup(kubernetesVersion string, amiType *infrav1.EKSAMI
...
 		paramName = fmt.Sprintf(eksGPUAmiSSMParameterFormat, formattedVersion)
 	default:
-		paramName = fmt.Sprintf(eksAmiSSMParameterFormat, formattedVersion)
+		switch architecture {
+		case Arm64ArchitectureTag:
+			paramName = fmt.Sprintf(eksARM64AmiSSMParameterFormat, formattedVersion)
+		case Amd64ArchitectureTag:
+			paramName = fmt.Sprintf(eksAmiSSMParameterFormat, formattedVersion)
+		default:
+			return "", errors.Errorf("cannot look up eks-optimized image for architecture %q", architecture)
```
Skarlso, line 304: Please use `fmt.Errorf`?

pkg/cloud/services/ec2/ami_test.go
```diff
@@ -54,6 +55,7 @@ func TestDefaultAMILookup(t *testing.T) {
 			args: args{
 				ownerID:           "ownerID",
 				baseOS:            "baseOS",
+				architecture:      "x86_64",
```
Skarlso, line 58: Please add `arm` tests.

pkg/cloud/services/ec2/launchtemplate_test.go
```diff
@@ -1171,6 +1171,7 @@ func TestDiscoverLaunchTemplateAMI(t *testing.T) {
 				ImageLookupFormat: "ilf",
```
Skarlso, line 1171: Please add arm tests.

pkg/cloud/services/ec2/ami.go
charlie-haley: There was some previous conversation around this logic https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4054#pullrequestreview-1315802219

The value of `supportedArchs` is almost always a single architecture, e.g `amd64`. In some cases there's multiple architectures like in the case of some small `t2` instances, they have both `amd64` and `i386` - if we ever encounter that we only want to select `amd64` anyway. An instance should never be able to support both `amd64` and `arm64` so we shouldn't encounter any issues from doing the checks like this.

As mentioned in the previous comment, it means if AWS decide to add more architectures like `riscv64` we'll have to update the switch statement.

Ankitasw:
@Skarlso could you please take another look, so that we can close this one?

pkg/cloud/services/ec2/ami.go
```diff
@@ -94,8 +103,46 @@ func GenerateAmiName(amiNameFormat, baseOS, kubernetesVersion string) (string, e
...
+func (s *Service) pickArchitectureForInstanceType(instanceType string) (string, error) {
+	descInstanceTypeInput := &ec2.DescribeInstanceTypesInput{
+		InstanceTypes: []*string{&instanceType},
+	}
+	describeInstanceTypeResult, err := s.EC2Client.DescribeInstanceTypes(descInstanceTypeInput)
+	if err != nil {
+		return "", err
+	}
+
+	supportedArchs := describeInstanceTypeResult.InstanceTypes[0].ProcessorInfo.SupportedArchitectures
```
Skarlso, line 116: Do a len check before accessing the first element.

pkg/cloud/services/ec2/ami.go
```diff
@@ -94,8 +103,46 @@ func GenerateAmiName(amiNameFormat, baseOS, kubernetesVersion string) (string, e
...
+
+	logger := s.scope.GetLogger().WithValues("instance type", instanceType, "supported architectures", supportedArchs)
+	logger.Info("Obtained a list of supported architectures for instance type")
+
+	// Loop over every supported architecture for the instance type
+	architecture := ""
+archCheck:
+	for _, a := range supportedArchs {
```
Skarlso, line 124: Okay so sorry but I'm still struggling to understand how this switch works. 🤣🤣 so in supported architecture if the first item is amd we will just accept that but if the first item is arn then we will just accept THAT! Like how is this determinate in any way or how is this a "choice"? Sorry I just might not know the api here very well.

Skarlso (commented):
Just two more things. 😂

pkg/cloud/services/ec2/ami.go
Skarlso: Yeah, this is what I was expecting. However... This means that if both arm and amd are available, there is no choice, right? It will return whichever is first.

Skarlso: Okay then. :)

Skarlso: In that case, please just add the len check, than it's good to go. :)

mloiseleur:
Thanks @charlie-haley !

PR #4059, opened 2023-02-10
fix: cleanup AWS CloudFormation stack in Test environment

Skarlso:
**What type of PR is this?**

/kind bug

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

### DIFF
--- a/cmd/clusterawsadm/cmd/bootstrap/iam/cloudformation.go
+++ b/cmd/clusterawsadm/cmd/bootstrap/iam/cloudformation.go
@@ -109,7 +109,7 @@ func createCloudFormationStackCmd() *cobra.Command {
 
 			cfnSvc := cloudformation.NewService(cfn.New(sess))
 
-			err = cfnSvc.ReconcileBootstrapStack(t.Spec.StackName, *t.RenderCloudFormation(), t.Spec.StackTags)
+			err = cfnSvc.ReconcileBootstrapStack(t.Spec.StackName, *t.RenderCloudFormation(), t.Spec.StackTags, false)
 			if err != nil {
 				fmt.Printf("Error: %v\n", err)
 				return err
--- a/test/e2e/shared/aws.go
+++ b/test/e2e/shared/aws.go
@@ -425,7 +425,9 @@ func createCloudFormationStack(prov client.ConfigProvider, t *cfn_bootstrap.Temp
 		deleteMultitenancyRoles(prov)
 		if aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackFailed ||
 			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackComplete ||
-			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackInProgress {
+			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackInProgress ||
+			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusCreateFailed ||
+			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusDeleteFailed {
 			// If cloudformation stack creation fails due to resources that already exist, stack stays in rollback status and must be manually deleted.
 			// Delete resources that failed because they already exists.
 			deleteResourcesInCloudFormation(prov, t)
--- a/test/e2e/shared/aws.go
+++ b/test/e2e/shared/aws.go
@@ -410,13 +410,16 @@ func createCloudFormationStack(prov client.ConfigProvider, t *cfn_bootstrap.Temp
 	cfnSvc := cloudformation.NewService(CFN)
 
 	Eventually(func() bool {
-		err := cfnSvc.ReconcileBootstrapStack(t.Spec.StackName, *renderCustomCloudFormation(t), tags)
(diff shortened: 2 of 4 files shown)


test/e2e/shared/aws.go
```diff
@@ -410,13 +410,16 @@ func createCloudFormationStack(prov client.ConfigProvider, t *cfn_bootstrap.Temp
 	cfnSvc := cloudformation.NewService(CFN)
 
 	Eventually(func() bool {
```
Skarlso, line 412: This whole eventually is a wonkey. The first event that it sees is this:
```
   STEP: Loading the e2e test configuration from "/home/prow/go/src/sigs.k8s.io/cluster-api-provider-aws/test/e2e/data/e2e_eks_conf.yaml" @ 02/10/23 05:21:14.18
  STEP: Getting an AWS IAM session - from environment @ 02/10/23 05:21:14.181
  STEP: Creating a bootstrap AWSIAMConfiguration @ 02/10/23 05:21:14.182
  STEP: Creating AWS CloudFormation stack for AWS IAM resources: stack-name=cluster-api-provider-aws-sigs-k8s-io @ 02/10/23 05:21:14.188 
```

That suggests that the CF stack is already there. Yet there is no already exists error. So the CF stack must be there from a previous test run somehow and it somehow doesn't find it again or, immediately the first event it got is that. The problem is that after that it stays in RollBack. So instead of that, we Delete it and we don't roll back in a test environment. Hopefully that will help.

test/e2e/shared/aws.go
```diff
@@ -445,17 +448,23 @@ func deleteResourcesInCloudFormation(prov client.ConfigProvider, t *cfn_bootstra
...
+			By(fmt.Sprintf("cleanup for role with name '%s'", role.RoleName))
 			Eventually(func(gomega Gomega) bool {
 				_, err := iamSvc.DeleteRole(&iam.DeleteRoleInput{RoleName: aws.String(role.RoleName)})
 				return awserrors.IsNotFound(err) || err == nil
 			}, 5*time.Minute, 5*time.Second).Should(BeTrue())
 		}
 		if val.AWSCloudFormationType() == "AWS::IAM::InstanceProfile" {
 			profile := val.(*cfn_iam.InstanceProfile)
-			_, _ = iamSvc.DeleteInstanceProfile(&iam.DeleteInstanceProfileInput{InstanceProfileName: aws.String(profile.InstanceProfileName)})
+			By(fmt.Sprintf("cleanup for profile with name '%s'", profile.InstanceProfileName))
```
Skarlso, line 463: I also added cleanup for these guys. Let's see if that helps.

Skarlso: Since the thing it complains about existing is actually an `ManagedPolicy` and not an IAM Role.

test/e2e/shared/aws.go
Skarlso: BTW, create will NOT throw an error if it fails afterwards. Meaning, we also need to check what the status is and what the events are, which this eventually doesn't seem to care about. It describes events, but doesn't check if any of them is a ROLLBACK or a FAILED or anything....

Skarlso:

Well, at least the CF part worked. :D

cmd/clusterawsadm/cmd/bootstrap/iam/cloudformation.go
```diff
@@ -109,7 +109,7 @@ func createCloudFormationStackCmd() *cobra.Command {
 
 			cfnSvc := cloudformation.NewService(cfn.New(sess))
 
-			err = cfnSvc.ReconcileBootstrapStack(t.Spec.StackName, *t.RenderCloudFormation(), t.Spec.StackTags)
+			err = cfnSvc.ReconcileBootstrapStack(t.Spec.StackName, *t.RenderCloudFormation(), t.Spec.StackTags, false)
```
dlipovetsky, line 112: Note to self: The is backward-compatible.

test/e2e/shared/aws.go
```diff
@@ -410,19 +410,24 @@ func createCloudFormationStack(prov client.ConfigProvider, t *cfn_bootstrap.Temp
...
+
 	stack, err := CFN.DescribeStacks(&cfn.DescribeStacksInput{StackName: aws.String(t.Spec.StackName)})
 	if err == nil && len(stack.Stacks) > 0 {
 		deleteMultitenancyRoles(prov)
 		if aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackFailed ||
 			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackComplete ||
-			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackInProgress {
+			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusRollbackInProgress ||
+			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusCreateFailed ||
+			aws.StringValue(stack.Stacks[0].StackStatus) == cfn.StackStatusDeleteFailed {
```
dlipovetsky, line 430: If the delete failed, we try deleting again? Do we expect it work the second time?

Skarlso: Maybe yes. This is aws... but I can remove these. 😁

dlipovetsky: Let's revisit after merging.

PR #4092, opened 2023-02-21
Modify AWSMachine reconciliation behavior to terminate and create instances without blocking

cnmcavoy:
**What type of PR is this?**
/kind bug

The existing AWSMachine reconciliation loop blocks on terminating an ec2 instance and waits for it to confirm the termination with AWS, as well as blocking on ec2 instance creation and waits up to one minute, for no discernible reason I could detect. This PR removes the blocking behavior, proceeding with reconciliation immediately after the AWS sdk confirms success on the API call. 

As a result, this change greatly speeds up large machine deployment rollouts. Previously, very large Machine Deployments (> 100 machines) would take several hours for the machine sets to provision new nodes and tear down machines. This was mostly a result of the reconciliation blocking for each instance creation and delete, and while blocked, prevented additional concurrent work from proceeding, as there is a fixed number of concurrent AWSMachine reconcile workers allowed (controlled by `--awsmachine-concurrency`).

We tested this change in one of our clusters - timing it with 600 nodes, split into 3 node groups (really, 1 node group, duplicated for 3 az's in us-east-2). We were testing with t3.small instances which have just enough CPU to run kubelet and the daemonset pods we deploy. 

We captured metrics from both the CAPI and CAPA controllers as well as metrics about the state of the CAPI resources in the kubernetes cluster. The most meaningful thing to note in the metrics is the time scale. The rollout took ~2 hours to complete on the mainline version and < 45 minutes on this branch. The CAPA unfinished workqueue metric is also significantly different before and after this change.

This is metrics without the change (running on capi v1.3.2, capa v2.0.2 official release):

Screenshot 2023-02-23 at 10-30-51 Kubernetes Cluster Autoscaling Datadog

This is the metrics from the same cluster after this change (running on capi v1.3.2, this PR branch):

Screenshot 2023-02-23 at 10-31-38 Kubernetes Cluster Autoscaling Datadog

Fixes #

Split into two commits for review, first commit is the modified behavior. The second commit is updating the e2e test sites for the new reconciliation behavior.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Modifies the AWSMachine reconciliation to no longer block on ec2 termination or creation, resulting in faster rollouts for large Machine Deployments.
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
 	ignTypes "github.com/flatcar/ignition/config/v2_3/types"
@@ -340,8 +341,14 @@ func (r *AWSMachineReconciler) reconcileDelete(machineScope *scope.MachineScope,
 	// This decision is based on the ec2-instance-lifecycle graph at
 	// https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/ec2-instance-lifecycle.html
 	switch instance.State {
-	case infrav1.InstanceStateShuttingDown, infrav1.InstanceStateTerminated:
+	case infrav1.InstanceStateShuttingDown:
 		machineScope.Info("EC2 instance is shutting down or already terminated", "instance-id", instance.ID)
+		// requeue reconciliation until we observe termination (or the instance can no longer be looked up)
+		return ctrl.Result{RequeueAfter: time.Minute}, nil
+	case infrav1.InstanceStateTerminated:
+		machineScope.Info("EC2 instance terminated successfully", "instance-id", instance.ID)
+		controllerutil.RemoveFinalizer(machineScope.AWSMachine, infrav1.MachineFinalizer)
+		return ctrl.Result{}, nil
 	default:
 		machineScope.Info("Terminating EC2 instance", "instance-id", instance.ID)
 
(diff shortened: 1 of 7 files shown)


controllers/awsmachine_controller.go
```diff
@@ -499,6 +498,7 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 			conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceProvisionFailedReason, clusterv1.ConditionSeverityError, err.Error())
 			return ctrl.Result{}, err
 		}
+		//return ctrl.Result{RequeueAfter: time.Minute}, nil
```
javanthropus, line 501: Why was this commented out when updating the tests?

cnmcavoy: Good question. I previously tested this controller with that uncommented - however after reviewing the code and tests today, I don't believe it's necessary, we can just fall through and continue with reconciliation instead of exiting early and re-queuing. I need to confirm it's not behavior we actually need, I'm going to do some more tests with large Machine Deployments tomorrow.

Hence the WIP status.

cnmcavoy: Removed, this was not necessary in my tests today.

Ankitasw (commented):
@cnmcavoy AFAIR, the blocking call was used such that instance termination happens successfully. As the operation takes some time to finish, hence is the blocking call. What would happen if instance termination doesn't finish at that point, and happen to be at failed state in the later point of time(due to some xyz reason) but we already set the conditions as deleted, but the instance is left behind and may fail the delete operation?
_Thinking out loud here, about the downside of removing blocking calls_

cnmcavoy:
@Ankitasw 

Thanks for the additional context. If the intent is to ensure that the ec2 instance associated with the AWSMachine is in a terminated state (or absent entirely) we don't need to block, one option could be to requeue the AWSMachine to be re-reconciled in a later reconciliation and repeat this until the cloud provider state matches. Blocking still doesn't seem to be necessary to satisfy that requirement. 

I can update this PR to include covering that edge case and requeue the AWSMachine after `TerminateInstance` until the lookup for the instance no longer succeeds.

Ankitasw:
> I can update this PR to include covering that edge case and requeue the AWSMachine after TerminateInstance until the lookup for the instance no longer succeeds.

This sounds good to me, I will take a look at the PR update. Thanks @cnmcavoy

Ankitasw:
/lgtm
/approve

We can ignore that error, as its failing sometimes.

PR #4094, opened 2023-02-22
Add IRSA Support to Self Managed Clusters

### LINKED ISSUE #3560: Support for non-EKS IRSA
/kind feature

IRSA appears to have been complete for EKS clusters in #2054. My company currently uses non-EKS clusters in combination with IRSA. I'd like the ability to install IRSA using the provider. The main piece that is tricky to coordinate is the certificate authority used to sign and validate the JWTs. If these pieces could be automated as part of the ignition config (we also use Flatcar), our burden of installation/maintenance would be greatly decreased. 

**Environment:**

- Cluster-api-provider-aws version: v1.4
- Kubernetes version: (use `kubectl version`): v1.24
- OS (e.g. from `/etc/os-release`): Flatcar Container Linux by Kinvolk 3139.2.3 (Oklo)

luthermonson:
**What type of PR is this?** 
/kind feature

**What this PR does / why we need it**: Adds IRSA (oidc provider/pod-identity-webhook) functionality to an IAAS cluster. This will bring the IAAS clusters inline in functionality with Managed clusters which provide most of this functionality out of the box.

**How this was Accomplished**:
In CAPA kubeadm already builds in serving certs and since kube 1.21 there are two endpoints in the apiserver which make it an oidc provider, these are `/.well-known/openid-configuration` and `/openid/v1/jwks`. One can configure the openid responses through a couple kube apiserver parameters which is managed and ran as a static pod. To accomplish any modifications the preferred method to add additional extra args is actually through kubeadm [patches](https://kubernetes.io/docs/setup/production-environment/tools/kubeadm/control-plane-flags/#patches) as per a trail of [issues](https://github.com/kubernetes/kubeadm/issues/1601) which show an api design flaw in the map[string]string used for ExtraArgs.

Once the apiserver end points are running they need to be served publicaly on port 443, since CAPA already uses an s3 bucket for ignition configs I repurposed it to host the contents of the two files and configured the API server accordingly. Then all that has to be done is an oidc provider is created in IAM which points to the S3 bucket and we install the pod-identity-webhook provided by AWS.

The EKS implementation only creates a configmap called `boilerplate-oidc-trust-policy`. You are expected to copy the contents of this configmap, edit in your details for the name/namespace of the service account and create a new role with that trust policy while assigning additional roles you with the account to use. There is an [example](https://gist.github.com/luthermonson/8e813cfe792cb5f0a8c038771745714f) I use to test IRSA which a role will need to be created and given both the trust policy and assigned addtional roles and it will list your s3 buckets confirming the oidc provider and IRSA are functioning.

Extra details to look for to understand how this actually works, the pod-identity-webhook adds a mutating webhook for all pod creation. If the pod has a service account and it contains an annotation like `eks.amazonaws.com/role-arn: arn:aws:iam::<accountID>:role/test` it will know to generate an STS token and inject it directly into the pod via Env variables.

Fixes #3560

Adding a new ReconcileOIDCProvider to the AWSCluster reconciliation loop.
- Created a new IAM service, in the future the same logic for EKS could be combined as previously the oidc code was buried in the EKS service. Details on the reconciler can be found in [comments](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4094/files#diff-a41f3e76f5a70a6f449fecf5766e35a00299daf53c97fa84838d557f8c4766d7R44-R53).
- Updated the S3 service to take keys instead of Machines and added a CreatePublic for the JWKS/OIDC config
- Exposed a ManagementClient and RemoteClient for both cluster types and exported Client.
- Moved OIDCProvider status type to v1beta2 and migrated out of the EKS API to make one type both clusters can reference a single type.
- This cyclomatic showed up during testing so I refactored reconcileNormal
```
controllers/awsmachine_controller.go:435:1: cyclomatic complexity 38 of func `(*AWSMachineReconciler).reconcileNormal` is high (> 30) (gocyclo)
func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope *scope.MachineScope, clusterScope cloud.ClusterScoper, ec2Scope scope.EC2Scope, elbScope scope.ELBScope, objectStoreScope scope.S3Scope) (ctrl.Result, error) {
```

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:
```release-note
adding IRSA support for self-hosted clusters
```


Ankitasw:
@richardcase @Skarlso since this is a significant change, wouldn't it be good if we have proposal in place before actual implementation?

luthermonson:
I can write one of you want one

richardcase:
It's a fair size change, but not massive. We have made more considerable changes without a proposal.

But as this is related to IRSA and credentials it might be a good idea for a proposal to be written. If you were willing to do a proposal @luthermonson that would be super useful. But also appreciate that this is more work for you.

Skarlso:
@luthermonson Usually, proposals are an independent PR.

luthermonson:
found a bunch more stuff while testing this. will be rebasing and adding the things I worked on for our capa fork and get them back over here as well as cleaning up some superfluous things in this PR. soon as my head is screwed on straight with the code ill write the proposal and you guys can go over this and confirm we agree on what the feature actually does and if the code reflects that.

Skarlso:
@luthermonson can you ping me later on the proposal PR please? :) Cheers.

luthermonson:
> @luthermonson can you ping me later on the proposal PR please? :) Cheers.

no time to write yet... im trying

Skarlso:
No worries, mate. You've got time.

DavidSpek:
@luthermonson @Skarlso Is this PR still active and what needs to be done for it to get merged?

Skarlso:
@DavidSpek I might take this up and try finishing it if @luthermonson is out.

I got a bit burned out and removed myself from CAPA and open source for a little bit. But I'm back now, so I'll try and help kick this over the finishline.

Skarlso:
Hopefully, my rebase was successful.. We'll find out in a sec.. :D

Skarlso:
I'll go over the whole PR and do some cleanup.

Luther already tested this thing extensively as far as I followed it. So the code is more or less complete.

DavidSpek:
Awesome, thanks for having a look at it again.

controllers/awsmachine_controller.go
```diff
@@ -639,6 +648,45 @@ func (r *AWSMachineReconciler) reconcileOperationalState(ec2svc services.EC2Inte
...
+		if oneReady {
+			awsCluster := &infrav1.AWSCluster{}
+			key := types.NamespacedName{Namespace: machineScope.InfraCluster.Namespace(), Name: machineScope.InfraCluster.Name()}
+			if err := r.Client.Get(ctx, key, awsCluster); err != nil {
+				return err
+			}
+			anno := awsCluster.GetAnnotations()
+			anno[scope.KubeconfigReadyAnnotation] = "true"
+			awsCluster.SetAnnotations(anno)
+			if err := r.Client.Update(ctx, awsCluster); err != nil {
```
Skarlso, line 682: @richardcase Is this the right way to update the annotation on aws Cluster object? Don't we have to use

```
func (s *ClusterScope) PatchObject() error {
```

... or something similar? Although if I'm not mistaken the patcher can't update annotations just the spec and status or something like that... Maybe I'm mistaken.

pkg/cloud/services/iam/podidentitywebhook.go
```diff
@@ -0,0 +1,340 @@
...
+								},
+							},
+						},
+					},
+				},
+			},
+		},
+	}
+
+	return remoteClient.Create(ctx, deployment)
```
Skarlso, line 218: I think this should be a create or update or something like that if this is really a reconciliation process. Right?

pkg/cloud/services/iam/podidentitywebhook.go
```diff
@@ -0,0 +1,340 @@
...
+						},
+					},
+				},
+				SideEffects:             &none,
+				AdmissionReviewVersions: []string{"v1beta1"},
+			},
+		},
+	}
+
+	return remoteClient.Create(ctx, mwh)
```
Skarlso, line 273: Same here

Skarlso:
Did a first pass-through of the code. I just have two questions I'm going to find answers for, then I'm gonna read it again and try some tests.

Skarlso:
Uh, these conflicts are killing me. :D Would be nice to get this in or reviewed. :)

luthermonson:
> @DavidSpek I might take this up and try finishing it if @luthermonson is out.

i'm not quite dead yet

pkg/cloud/services/iam/podidentitywebhook.go
luthermonson: most of these reconcilers basically make sure the thing exists or creates it... there is no check if anything changed and to fix it via an update back to what we need... that makes it a bit more complex but if we want that on all of these i can try one and see if we like the direction and then implement the rest

Skarlso: Sounds good!

pavansokkenagaraj:
Hi @luthermonson, Please correct me if I am wrong,

If the pod-identity-webhook is situated on a worker node, and that node experiences downtime or is rescheduled to another node, the subsequent rescheduling of new pods on the affected node results in both the pods and the pod-identity-webhook being scheduled simultaneously.

This simultaneous rescheduling disrupts the injection of values from the pod-identity-webhook into the pods, leading to an absence of the expected injected values.

To address this issue, several potential solutions can be considered:

1. **Schedule pod-identity-webhook in the control plane:**
   Ensure that the pod-identity-webhook is scheduled within the control plane, preventing disruptions caused by worker node rescheduling.

2. **Implement a PodDistributionBudget with min.available set to 1:**
   Utilize a PodDistributionBudget to guarantee a minimum number of available pods (set to 1) even during rescheduling scenarios, maintaining continuous operation.

3. **Establish a PriorityClass:**
   Employ a PriorityClass to assign priority to the pod-identity-webhook, influencing scheduling decisions and reducing the likelihood of simultaneous rescheduling.

4. **Configure a MutatingWebhookConfiguration with failurePolicy set to fail (or make it optional):**
   Define a MutatingWebhookConfiguration with a failurePolicy set to "fail" or consider making it optional. This helps handle errors more explicitly during webhook execution.

Additionally, it's crucial to address the lack of updates on pod-identity-webhook resources to prevent upgrade issues when introducing new features during capi-aws upgrades. Regularly update the pod-identity-webhook resources to ensure compatibility with the latest capabilities and avoid potential complications during upgrades.

richardcase:
@luthermonson - feel free to ping me if i can help get this in.

pkg/cloud/services/iam/reconcilers.go
```diff
@@ -0,0 +1,180 @@
...
+	"k8s.io/apimachinery/pkg/types"
+
+	"sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	"sigs.k8s.io/cluster-api-provider-aws/v2/cmd/clusterawsadm/converters"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+// reconcilePodIdentityWebhook generates certs and starts the webhook in the workload cluster
+// https://github.com/aws/amazon-eks-pod-identity-webhook
+// 1. generate webhook certs via cert-manager in the management cluster
```
pavansokkenagaraj, line 24: @luthermonson 
What is the reasoning behind creating certs resources in the management cluster?

Skarlso: Security. You don't want unencrypted traffic to an HTTP webhook endpoint.

luthermonson: ehn, i think i know what he's asking... @pavansokkenagaraj capa does not automatically install cert-manager into the downstream cluster. for capi projects the management cluster gets cert-manager automatically and is the only place we can generate them which is why it's done there and moved

richardcase:
@luthermonson - sorry i need to get back to this.

joshfrench:
Is there anything an interested observer can do to help get this over the finish line?

sl1pm4t:
While testing this today, I found the CAPA controller needed these additional permissions:

```yaml
- apiGroups:
  - cert-manager.io
  resources:
  - certificates
  - issuers
  verbs:
  - create
  - get
  - list
  - delete
- apiGroups:
    - controlplane.cluster.x-k8s.io
  resources:
    - kubeadmcontrolplanes
  verbs:
    - get
    - list
    - patch
    - update
```

sl1pm4t:
Also while testing this PR, and upon trying to delete the cluster I found the deletion was getting blocked because CAPA controller was throwing error if the OIDC provider didn't exist / had already been deleted, so I added this to gracefully handle the `NoSuchEntity` error condition from AWS API:

https://github.com/sl1pm4t/cluster-api-provider-aws/commit/5956c872f7cd42754fade4199f557685d3e7518a

richardcase:
It will be a push to get this in but lets add this so we are tracking it:

/milestone v2.6.0

pkg/cloud/services/iam/reconcilers.go
```diff
@@ -0,0 +1,180 @@
...
+	if err != nil {
+		return err
+	}
+
+	// switch it to kube-system and move it to the remote cluster
+	if err := reconcileCertificateSecret(ctx, certSecret, remoteClient); err != nil {
+		return err
+	}
+
+	if err := reconcilePodIdentityWebhookComponents(ctx, s.scope.Namespace(), certSecret, remoteClient); err != nil {
```
sl1pm4t, line 50: The use of `s.scope.Namespace()` here means that CAPA will attempt to install the pod-identity-webhook into whatever namespace the cluster resources where created in, on the management cluster.

This works for the simplest case of installing everything into `default`, but if the cluster resources are created in something else like a namespace called `capi-<cluster-name>` then this will fail, because the namespace does not exist in the remote cluster. 

IMO - the pod-identity-webhook should be deployed into the `kube-system` namespace.

sl1pm4t: I've pushed a commit [here](https://github.com/sl1pm4t/cluster-api-provider-aws/commit/a2b8fb94338a8fc600ae233b02da0a374de1e09f) that:
* hard codes the installation namespace of pod-identity-webhook to be `kube-system`
* fixes the certificates to reflect the new namespace
* fixes an issue that could prevent a successful reconcile if the webhook certificate secret already existed in the remote cluster (from a previous reconcile loop).
* bumps the `pod-identity-webhook` image version to the latest (v0.5.5)

luthermonson: this was based on the install examples of the podid wh... https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/deploy/deployment-base.yaml#L5 trying to keep the install as similar to upstream as possible which does install it into default. the rig I used to work on made that namespace for you in the workload cluster (we created and pivoted and put everything there) so I never caught that bug,

I'm ok with hardcoding to `kube-system` (which is probably the best place for it) but just like to make a case for `default` being where the AWS/EKS devs intended self-install users to put it

sl1pm4t: I've always thought it was strange their examples deployed to `default`. From an permissions point of view, it's easy to protect access to it in `kube-system` but often users would have permission in `default`. 

I note that kOps also hard codes it to `kube-system`: https://github.com/kubernetes/kops/blob/41480d95a96edad247ecd777b2790ecda33f6a85/upup/models/cloudup/resources/addons/eks-pod-identity-webhook.addons.k8s.io/k8s-1.16.yaml.template#L69

damdo:
@luthermonson this will need rebasing

sl1pm4t:
@damdo - with agreement from @luthermonson I have continued the work on this PR (including rebase) over in #5109

damdo:
Thanks for the heads up @sl1pm4t 
I suppose we can close this PR in favor of #5109 then @luthermonson ?

richardcase:
As #5109 is the new PR:

/close

PR #4095, opened 2023-02-23
cleaning up unnecessary object patches

### LINKED ISSUE #4093: Reduce PatchObject Calls in Controllers
/kind bug

Stepping through an AWSCluster Reconcile code and you'll find 3 distinct Patch calls made to object... it's possible this is a NOOP when there is no diff but it's also a waste of resources to even calculate so I have two suggestions for improvement.

1) this defer (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awscluster_controller.go#L174-L186) could be removed entirely if the patch conditions were added to this call here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f4094499aa9b07da0a16b68c72f7c586359b630e/pkg/cloud/scope/cluster.go#L251-L262 which is called from the already existing defer added in 2019: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awscluster_controller.go#L203
2) this Patch (https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awscluster_controller.go#L277-L281) is being called every reconcile when it's only ever done the first time the AWSCuster comes through the reconciler. the logic could be improved to check the AWSCluster if it had the reconciler and to skip the patch... this is done in multiple places in other controllers (e.g. AWSMachine) so this could be a decent improvement across the whole project.

Call Patch less

I will be making a PR to try and fix this.

luthermonson:
**What type of PR is this?**
/kind bug

- The AWSCluster reconciler has an additional defer which included a patch in it which is ignoring the existing existing Close happening in a defer about 10 lines further down. During reconciliation you'll get a double Patch call and the only difference in the two was the list of conditions. I merged the conditions and deleted the unneeded Patch.
- All finalizers are being added by `controllerutil.AddFinalizer` which returns a bool if the finalizer was actually added to the object. It will return true on the first reconciliation and the patch will be called and false from then on, wrapping the AddFinalizer call with an if and gating the Patch will avoid the patch logic when it's obviously not necessary.

Fixes #4093

yo dog i got a patch for your patch when you want patches!

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Remove extra object patching from AWSCluster to decrease the overwhelming number of kube API calls.
```

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -55,7 +55,6 @@ import (
 	"sigs.k8s.io/cluster-api/util"
 	capiannotations "sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/conditions"
-	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/predicates"
 )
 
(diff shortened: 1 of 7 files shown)


Skarlso:
Nice catch and good use of AddFinalizer's return value.

Ankitasw:
@luthermonson please make sure to update release notes, otherwise PR won't merge.

Skarlso:
Updated the release notes for Luther as he's in a different TZ. :)

luthermonson:
@Skarlso thanks friendo

PR #4104, opened 2023-03-01
add explicit securityContexts to the controller

chrischdi:
**What type of PR is this?**

/kind feature

This does not really change the configuration, it just makes it explicit and enforce the defaults, except for the seccompPolicy which changes from Unconfined to RuntimeDefault. Syscalls filtered by RuntimeDefault policy are 95% namespaced and require capabilities (which we drop) in the first place, so no practical change there either.

This allows to be compatible to the restricted pod security admission profile.

Fixes #

Prior art:
* https://github.com/kubernetes-sigs/cluster-api/pull/7831

**Checklist**:

- [X] squashed commits

**Release note**:

```release-note
Explicitly set securityContexts in the manifests to comply with the restricted pod security admission profile.
```

### DIFF
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -37,6 +37,17 @@ spec:
           httpGet:
             path: /healthz
             port: healthz
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+          runAsUser: 65532
+          runAsGroup: 65532
+      securityContext:
+        runAsNonRoot: true
+        seccompProfile:
+          type: RuntimeDefault
       terminationGracePeriodSeconds: 10
       tolerations:
         - effect: NoSchedule


Ankitasw:
@chrischdi  could you wait for some more time before triggering again, I think it would fail continuously for 5-10 min.

chrischdi:
let's try again:



chrischdi:
And also

dlipovetsky:
@chrischdi Thanks for this!

My only concern is whether this securityContext is compatible with all the Kubernetes versions CAPA v2 is guaranteed to work with.

I see that the same concerns are raised in https://github.com/kubernetes-sigs/cluster-api/pull/7831#issuecomment-1369687785.

I would appreciate it if you could briefly address this.

chrischdi:
> @chrischdi Thanks for this!
> 
> My only concern is whether this securityContext is compatible with all the Kubernetes versions CAPA v2 is guaranteed to work with.
> 
> I see that the same concerns are raised in [kubernetes-sigs/cluster-api#7831 (comment)](https://github.com/kubernetes-sigs/cluster-api/pull/7831#issuecomment-1369687785).
> 
> I would appreciate it if you could briefly address this.

Fair ask, I will take a look at this 👍

chrischdi:
I explored the following:
* Create Kind Cluster (`kind create cluster --image <node-image>`)
* Create overwrite configuration to the `make staging-manifest` output of (copy over `out/infrastructure-components.yaml` to `$HOME/.cluster-api/overrides/infrastructure-aws/v2.0.2`, had to remove the `ExternalResourceGC` feature gate because its not part of v2.0.2)
* Run `clusterctl init --infrastructure aws`
* See if `capa-controller-manager` pod in `capa-system` is running

I did that using the following kind/kubernetes versions:
* v1.19 (using clusterctl `v1.1.6`)
* v1.20 (using clusterctl `v1.3.4`)

@dlipovetsky : could you maybe help me out to outline which the lowest compatible kubernetes version is and or if this good enough? The [book](https://cluster-api-aws.sigs.k8s.io/topics/reference/versions.html) is not accurate about this IMHO.

* When reading it, the lowest border may be CAPI v1.0 (which is EOL already since 2022-02-02).
  * Via the table [Compatibility with Kubernetes Versions](https://cluster-api-aws.sigs.k8s.io/topics/reference/versions.html#compatibility-with-kubernetes-versions)
* CAPI v1.0 required a management cluster in v1.19 or greater

chrischdi:
Sorry to bother you @dlipovetsky , I'd really appreciate if you could take another look at the above 🙂

config/manager/manager.yaml
```diff
@@ -37,6 +37,17 @@ spec:
...
             port: healthz
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+          runAsUser: 65532
+          runAsGroup: 65532
+      securityContext:
+        runAsNonRoot: true
```
Skarlso, line 48: Sadly, this will immediately kill the nice Tilt setup we have for controller testing because Tilt requires root access to the deployment.

Before we merge this, the Tiltfile somehow has to be made aware.

In fact, can you please test the tilt setup under cluster-api?

Ankitasw: Nice catch @Skarlso

Skarlso: So I had a quick look at the CAPI Tiltfile. Indeed it would break the deployment. But I would test this nevertheless. Maybe there is magic I'm missing, but most likely not.

This is the thing that needs to be added in order to change the deployment in development:

```python
objects = decode_yaml_stream(install)
for o in objects:
    if o.get('kind') == 'Deployment' and o.get('metadata').get('name') == '<bla>':
        o['spec']['template']['spec']['securityContext']['runAsNonRoot'] = False
        break
```

Or something like that. :) Which means, in order to get this in, we require a CAPI PR that fixes the Tiltfile first.

chrischdi: Thanks for  the awesome feedback!

I thought this was already covered by 
- https://github.com/kubernetes-sigs/cluster-api/pull/7846
But I will take a look into it tomorrow and also try if capa runs with tilt or requires an PR (and file that PR if so :-) )

Skarlso: My god I completely forgot this existed. Why do they have ALL of that. :D :D

Skarlso: To speed up start time. Interesting.

chrischdi: I will check that out tomorrow and run capa with tilt to ensure I don't miss something 👍

chrischdi: So to confirm: it works with the above linked change.

However: using CAPI v1.3 + tilt will not work, because the change was not cherry-picked to CAPI v1.3.

CAPI v1.4 or main works just fine.

Skarlso: Uh, what a bother. But okay, this does not concern this change. I'm wondering however, how this affects upgrade procedures. I'm very well-versed in that area. What should we look out for when changing existing deployments?

Ankitasw: @Skarlso we could bring this up with CAPI folks.

richardcase:
This is a recommendation in the [CAPI v1.3->1.4 upgrade guide](https://main.cluster-api.sigs.k8s.io/developer/providers/v1.3-to-v1.4.html). 

So will have to wait until we upgrade main to v1.4.0 (testing in #4154 ), until then:

/hold

Ankitasw:
@chrischdi Since CAPI bump PR is merged now, would you be able to rebase this PR, so we could run E2E tests and see if everything is fine?

chrischdi:
Thanks for the heads up 🙂 

Rebased the PR.

PR #4112, opened 2023-03-02
Add instance details to AWSMachine, enables more than one instance type for instance creation

### LINKED ISSUE #4118: Enhance the AWSMachineTemplate to support multiple instance types, enable mixed instances in a MachineDeployment
/kind feature

Enhance the AWSMachineTemplate to support multiple instance types, specifically an ordered of instances and spot market options to use when creating the ec2 instance to back an AWSMachine, in order to handle provisioning failures when the desired instance type is unavailable. The AWSMachine should provision the ec2 instance with the type and spot options of the first item of the list, and only walk the list and use the other configurations if there is insufficient capacity to provision.

User stories:
* As a user, I want to be able to create a MachineDeployment which can be satisfied with multiple instance types of the same capacity but different families. For example, I want to create a node group for m6a.32xlarge or m6i.32xlarge instances, and I want to try creating m6a instances if they are available, but if they are unavailable, use m6i's.

* Also, as a user I want to be able to create a MachineDeployment which can use spot instances, but fall back to on-demand instances when the spot market is empty. For example, I want to create a node group for t3.large, but specify spot options to get a better price, and then fall back to on-demand t3.large's when the spot market is empty.

This would enable mixed instance capacities, e.g a m6a.2xlarge and m6a.4xlarge in the same MachineDeployment, however this kind of configuration is unsupported by the cluster autoscaler, so we may need to improve documentation to state that mixed capacities are unsupported in CAPI as well.

**Environment:**

- Kubernetes version: (use `kubectl version`):  v1.24
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

cnmcavoy:
**What type of PR is this?**
/kind feature

The implementation adds `instanceDetails`, a list of instance types and spot market options, which is an ordered list to the AWSMachine spec. The list is walked when trying to launch instances, using the first item in the list as the highest priority instance. 

If lower priority instance is used to launch an AWSMachine, and capacity to the higher priority instance type becomes available later, no attempt is made to remediate and move the AWSMachine to the higher priority instance type. 

Fixes #4118

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Adds a new field, `instanceDetails` to the AWSMachineSpec which allows a list of instance types and spot options to satisfy an AWSMachine creation.
```


api/v1beta2/awsmachine_webhook.go
```diff
@@ -113,6 +113,20 @@ func (r *AWSMachine) ValidateUpdate(old runtime.Object) error {
...
+		instanceDetails := map[string]interface{}{
+			"instanceType": oldAWSMachineSpec["instanceType"],
+		}
+		if _, ok := oldAWSMachineSpec["spotMarketOptions"]; ok {
+			instanceDetails["spotMarketOptions"] = oldAWSMachineSpec["spotMarketOptions"]
+		}
+		oldAWSMachineSpec["instanceDetails"] = []interface{}{instanceDetails}
+	}
+
 	if !cmp.Equal(oldAWSMachineSpec, newAWSMachineSpec) {
```
Skarlso, line 130: Errr, correct me if I'm wrong, but if the previous thing changes something in the machine spec because of the re-order, shouldn't this thing fail here?

api/v1beta2/awsmachinetemplate_webhook.go
```diff
@@ -226,6 +230,12 @@ func (r *AWSMachineTemplateWebhook) ValidateUpdate(ctx context.Context, oldRaw r
 
 	var allErrs field.ErrorList
 
+	// allow limited changes to instanceDetails if we are defaulting and the field was missing
```
Skarlso, line 233: I feel like I saw this sentence already... :D

api/v1beta2/awsmachinetemplate_webhook_test.go
```diff
@@ -128,7 +158,6 @@ func TestAWSMachineTemplateValidateUpdate(t *testing.T) {
 				Spec: AWSMachineTemplateSpec{
 					Template: AWSMachineTemplateResource{
 						Spec: AWSMachineSpec{
-							CloudInit:    CloudInit{},
```
Skarlso, line 131: Huh..? This isn't needed?

api/v1beta2/awsmachine_types.go
```diff
@@ -185,6 +193,18 @@ type CloudInit struct {
...
+// AWSInstanceDetails describes an instance details for creating a AWSMachine.
+type AWSInstanceDetails struct {
+	// InstanceType is the type of instance to create. Example: m4.xlarge
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength:=2
+	InstanceType string `json:"instanceType"`
+
+	// SpotMarketOptions allows users to configure instances to be run using AWS Spot instances.
+	// +optional
+	SpotMarketOptions *SpotMarketOptions `json:"spotMarketOptions,omitempty"`
```
faiq, line 205: Do you think `SpotMarketOptions` here is redundant with the one already defined in the machine spec? Is the idea that each `InstanceType` would have its own `SpotMarketOptions`. Some clarification here would be nice.

PR #4117, opened 2023-03-03
Add missing configuration to enable the awsmachinetemplate validating webhook

### LINKED ISSUE #4116: AWSMachineTemplate validation webhook is not configured, and its unit tests are broken
/kind bug

The AWSMachineTemplate validation webhook is not configured as expected, and so it is never used at runtime. In addition, because the configuration is missing, envtest fails to hit the code path for validation and unit tests for the AWSMachineTemplate webhook effectively never run. Also, when you restore the missing configuration, the tests are broken and fail to pass.

The AWSMachineTemplate validation webhook should be configured for the capa controller manager and the tests should pass.

The missing webhook configuration should be inserted here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/config/webhook/manifests.yaml#L410 and the correct config is:
```
- admissionReviewVersions:
    - v1
    - v1beta1
  clientConfig:
    service:
      name: webhook-service
      namespace: system
      path: /validate-infrastructure-cluster-x-k8s-io-v1beta2-awsmachinetemplate
  failurePolicy: Fail
  matchPolicy: Equivalent
  name: validation.awsmachinetemplate.infrastructure.cluster.x-k8s.io
  rules:
    - apiGroups:
        - infrastructure.cluster.x-k8s.io
      apiVersions:
        - v1beta2
      operations:
        - CREATE
        - UPDATE
      resources:
        - awsmachinetemplates
  sideEffects: None
  ```

**Environment:** N/A

### PR DESCRIPTION
**Pull request by cnmcavoy**
**What type of PR is this?**
/kind bug

Enables the AWSMachineTemplate validating webhook. Unclear why it was not previously configured.

~~The CloudInit behavior was removed, as it was copied from the v1alpha4 resources in [c09f55f8f2249d9fde6d7878a6595bf1c4672ff2](https://github.com/kubernetes-sigs/cluster-api-provider-aws/commit/c09f55f8f2249d9fde6d7878a6595bf1c4672ff2#diff-d85ef380a29a05566d8f529b89c7746d5ce4b257400061fe52be4a9ef0647668R65-R68) and kept being dragged forward, but the comment indicates it was only needed for one resource version upgrade, so I _think_ it's no longer necessary.~~

Fixes #4116

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fixed bug which resulted in the AWSMachineTemplate validating webhook not being enabled as expected.
```

### DIFF
--- a/api/v1beta2/awsmachinetemplate_webhook.go
+++ b/api/v1beta2/awsmachinetemplate_webhook.go
@@ -40,6 +40,7 @@ func (r *AWSMachineTemplateWebhook) SetupWebhookWithManager(mgr ctrl.Manager) er
 }
 
 // AWSMachineTemplateWebhook implements a custom validation webhook for AWSMachineTemplate.
+// Note: we use a custom validator to access the request context for SSA of AWSMachineTemplate.
 // +kubebuilder:object:generate=false
 type AWSMachineTemplateWebhook struct{}
 
@@ -224,11 +225,6 @@ func (r *AWSMachineTemplateWebhook) ValidateUpdate(ctx context.Context, oldRaw r
 	}
 
 	var allErrs field.ErrorList
-
-	// Allow setting of cloudInit.secureSecretsBackend to "secrets-manager" only to handle v1alpha3 upgrade
-	if oldAWSMachineTemplate.Spec.Template.Spec.CloudInit.SecureSecretsBackend == "" && newAWSMachineTemplate.Spec.Template.Spec.CloudInit.SecureSecretsBackend == SecretBackendSecretsManager {
-		newAWSMachineTemplate.Spec.Template.Spec.CloudInit.SecureSecretsBackend = ""
-	}
 
 	if !topology.ShouldSkipImmutabilityChecks(req, newAWSMachineTemplate) && !cmp.Equal(newAWSMachineTemplate.Spec, oldAWSMachineTemplate.Spec) {
 		allErrs = append(allErrs, field.Invalid(field.NewPath("spec"), newAWSMachineTemplate, "AWSMachineTemplate.Spec is immutable"))
--- a/api/v1beta2/awsmachinetemplate_webhook.go
+++ b/api/v1beta2/awsmachinetemplate_webhook.go
@@ -43,7 +43,8 @@ func (r *AWSMachineTemplateWebhook) SetupWebhookWithManager(mgr ctrl.Manager) er
 // +kubebuilder:object:generate=false
 type AWSMachineTemplateWebhook struct{}
 
-// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta2-awsmachinetemplate,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=awsmachinetemplates,versions=v1beta2,name=validation.awsmachinetemplate.infrastructure.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta2-awsmachinetemplate,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=awsmachinetemplates,versions=v1beta2,name=validation.awsmachinetemplate.infrastructure.cluster.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+
 var _ webhook.CustomValidator = &AWSMachineTemplateWebhook{}
 
 func (r *AWSMachineTemplate) validateRootVolume() field.ErrorList {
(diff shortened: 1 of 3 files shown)


api/v1beta2/awsmachinetemplate_webhook.go
```diff
@@ -224,11 +224,6 @@ func (r *AWSMachineTemplateWebhook) ValidateUpdate(ctx context.Context, oldRaw r
 
 	var allErrs field.ErrorList
 
-	// Allow setting of cloudInit.secureSecretsBackend to "secrets-manager" only to handle v1beta2 upgrade
```
Ankitasw, line 227: This has to be reverted, as upgrade E2E tests would fail

cnmcavoy: `make generate` removes the changes in the manifest.yaml, disabling the validating webhook, which also seems incorrect.

Ankitasw: Without `make generate` being successful, it wont work.
Not sure why its getting deleted with generate target. 
I will check this, meanwhile tagging @Skarlso @richardcase if they know about it.

Ankitasw: If you look at the other *webhook.go files, you would notice that only `AWSMachineTemplate` webhook is defined as customvalidator with kubebuilder tags. We should make CustomValidator as Validator, something like:

```go
func (r *AWSMachineTemplate) SetupWebhookWithManager(mgr ctrl.Manager) error {
	return ctrl.NewWebhookManagedBy(mgr).
		WithValidator(r).
		Complete()
}

// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta2-awsmachinetemplate,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=awsmachinetemplates,versions=v1beta2,name=validation.awsmachinetemplate.infrastructure.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
var _ webhook.Validator = &AWSMachineTemplate{}

```
But this is just a difference I observed which could be the reason its missed out of manifest. Could you give it a try?

cnmcavoy: `AWSMachineTemplate` doesn't implement the Validator interface though, so that won't work either unless we rewrite it. It uses a `CustomValidator` which I am now suspecting is the cause of all this.

Ankitasw: Yeah but we can implement the missing interfaces, isn't it? Wdyt? I think that will maintain consistency and also would solve our problem.

cnmcavoy: I will try swapping to a regular Validator, but I lack context on why this used a CustomValidator. I don't know if there was a reason behind the difference here.

Ankitasw: Yes exactly, I think this never worked coz its not there in manifest.
Maybe even I lack the context on why CustomValidator is used from the beginning, but it should not matter I guess. We anyways have to make it work as this is a bug and we should have the validation for AWSMachineTemplate.

cc @richardcase if he knows motivation behind the customvalidator.

cnmcavoy: Refactoring to use the `webhook.Validator` interface fixed the kubebuilder manifest generation. I still do not know how to address the test failures, given that the test descriptions as I understand them disagree with the code.

Ankitasw: I totally missed the tests part, I will take a look. Thanks for making the changes @cnmcavoy

Ankitasw: But one thing I am not sure of, that if this is breaking change, then we might have to hold onto this PR till next api version change.

cnmcavoy: The API break is around the removal of the `AWSMachineTemplateWebhook` struct. How strict are the api-breaking rules? I'd be shocked if removing that type mattered to any API consumer.

However, if we must preserve 1:1 compatibility,  we could leave that type and wire it into the validator for backwards compat.

cnmcavoy: I believe I satisfied the api-break by re-adding the removed types and deprecating them all. That should remove that blocker.

Ankitasw: It is used internally in CAPA but there might be consumers who would directly use APIs in their projects, so it would be better if we could leave the type and wire it for backward compatibility.

Ankitasw: @richardcase @Skarlso would also need your opinion on this.

Ankitasw: I am sorry to make this ping pong but looks like CustomValidator has been introduced for SSA feature. But I am not sure how to make this correct in the manifest.
I will give it another look once again before you make any new changes.

Ankitasw: See [this](https://cluster-api.sigs.k8s.io/developer/providers/webhooks.html#validating-webhooks) for details.

Ankitasw: So that would mean we just have to debug our generate target, but rest all code should remain same.

cnmcavoy: I figured out the issue with `make generate`. The kubebuilder comment needs a line of blank whitespace following it. Inserting a blank line after L46 of `awsmachinetemplates.go` causes the manifest to be added to the yaml.

I'm not sure why the whitespace matters to kubebuilder though.

cnmcavoy: I reverted this back to where it was before this thread sent us on the wrong path, but with the addition of the extra blank line.

The test failures still need to be resolved.

Ankitasw: Based on the comment and the previous git history, I see that I might have changed `SecretBackendSecretsManager` to `SecretBackendSSMParameterStore` unintentionally. Could you change it back to SecretBackendSecretsManager in the if condition and see if it works

Ankitasw: Now that I think about it again, I think you were right initially that this check doesn't affect us anymore. Let me dig out the original issue for this check.

cnmcavoy: Thanks for digging up that history @Ankitasw, that explains much. I updated the PR to "fix" the behavior as you documented, then added a followup commit to remove the defaulting behavior and tests.

Assuming the e2e passes, this should be ready to review / merge.

PR #4129, opened 2023-03-07
WIP: Gc functions

wyike:
**What type of PR is this?**

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```


pkg/cloud/services/gc/options.go
```diff
@@ -53,3 +53,14 @@ func withEC2Client(client ec2iface.EC2API) ServiceOption {
 		s.ec2Client = client
 	}
 }
+
+func WithGcStrategy(alternativeGCStrategy bool) ServiceOption {
```
dlipovetsky, line 57: nit

```suggestion
func WithGCStrategy(alternativeGCStrategy bool) ServiceOption { 
```

pkg/cloud/services/gc/utils.go
```diff
@@ -0,0 +1,56 @@
...
+
+	resource := &AWSResource{
+		ARN:  &parsedArn,
+		Tags: resourceTags,
+	}
+
+	return resource, nil
+}
+
+// chunkResources is similar to https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/elb/loadbalancer.go#L1488.
```
dlipovetsky, line 45: Thanks for leaving this comment. Could you also add a similar one to https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/elb/loadbalancer.go#L1488?

If someone needs to copy either one of these in the future, they'll know there will be 3 places where such a function is need, and will decide to factor it out into a separate package.

dlipovetsky:
I think "alternative" is not descriptive enough for a strategy name. I think there are two strategies here, e.g. `ResourceTaggingAPIStrategy` (existing) and `MultipleAPIStrategy` (introduced in this PR).

I'm not sure how to use a more descriptive strategy name with a feature gate, though.

pkg/cloud/services/gc/utils.go
```diff
@@ -0,0 +1,56 @@
...
+
+	// maxDescribeTagsRequest is the maximum number of resources for the DescribeTags API call
+	// see: https://docs.aws.amazon.com/elasticloadbalancing/latest/APIReference/API_DescribeTags.html.
+	maxDescribeTagsRequest = 20
+)
+
+// composeArn composes a resource arn with correct service and resource, but fake partition, region and account.
+func composeArn(service, resource string) string {
+	return "arn:" + fakePartition + ":" + service + ":" + fakeRegion + ":" + fakeAccount + ":" + resource
+}
```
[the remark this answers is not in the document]
wyike: I reuse the cleanup functions as before https://github.com/wyike/cluster-api-provider-aws/blob/gc_functions/pkg/cloud/services/gc/service.go#L64-L70 for alternative gc strategy.
These cleanups for example https://github.com/wyike/cluster-api-provider-aws/blob/gc_functions/pkg/cloud/services/gc/loadbalancer.go#L39 receives *AWSResource as a parameter and strip the prefixes in the arn and get the final resource ID.  However with elb client, the elb v1 client describe result won't show elb or security group arns, but names only (https://stackoverflow.com/questions/50134881/get-elb-arn-as-an-output). So I come up with an idea to compose a fake arn but with the real service type and resource prefix with the resource name(id) to let cleanup functions to handle correctly.

dlipovetsky:
Thank you for this feature, @wyike! :bowing_man:

pkg/cloud/services/gc/options.go
wyike: Thanks! Done.

pkg/cloud/services/gc/utils.go
wyike: OK. Added!

wyike:
> I think "alternative" is not descriptive enough for a strategy name. I think there are two strategies here, e.g. `ResourceTaggingAPIStrategy` (existing) and `MultipleAPIStrategy` (introduced in this PR).
> 
> I'm not sure how to use a more descriptive strategy name with a feature gate, though. The `--feature-gates` only accepts `KEY=true|false` values.
> 
> What about `ExternalResourceGCMultipleAPIStrategy` as the feature name? It's long, but descriptive.

Thanks Daniel for the inputs! It's really not easy to think of one most suitable name.  Richard has also given a lot of help on the naming.  cc@richardcase for further comments.

pkg/cloud/services/gc/utils.go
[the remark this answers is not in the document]
Skarlso: I think what Daniel is suggesting is, just to name it appropriately. Something like `composeFakeArn` and then an explanation for the existence of this function as a comment?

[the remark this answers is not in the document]
wyike: Ah.. yes.. thanks @Skarlso for reminder. Let me add more for it

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -1485,6 +1485,7 @@ func fromSDKTypeToLB(v *elbv2.LoadBalancer, attrs []*elbv2.LoadBalancerAttribute
 	return res
 }
 
+// chunkELBs is similar to https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/gc/utils.go#L46.
```
Skarlso, line 1488: Since this is the same project we should probably point to the relative path of the file?
Also, maybe since lines can change, copy at least the function name here or something so we find what this was pointing to in the future. :)

pkg/cloud/services/gc/cleanup.go
```diff
@@ -62,9 +62,25 @@ func (s *Service) ReconcileDelete(ctx context.Context) error {
 }
 
 func (s *Service) deleteResources(ctx context.Context) error {
-	s.scope.Info("deleting aws resources created by tenant cluster")
+	s.scope.Info("deleting aws resources created by tenant cluster", "cluster", s.scope.InfraClusterName())
+
+	resources, err := s.collectFuncs.Execute(ctx)
+	if err != nil {
+		return err
```
Skarlso, line 69:
```suggestion
		return fmt.Errrorf("failed to collect resources: %w", err)
```

pkg/cloud/services/gc/cleanup.go
```diff
@@ -77,7 +93,7 @@ func (s *Service) deleteResources(ctx context.Context) error {
 
 	awsOutput, err := s.resourceTaggingClient.GetResourcesWithContext(ctx, &awsInput)
 	if err != nil {
-		return fmt.Errorf("getting tagged resources: %w", err)
+		return nil, err
```
Skarlso, line 96:
```suggestion
		return nil, fmt.Errorf("getting tagged resources: %w", err)
```

pkg/cloud/services/gc/ec2.go
```diff
@@ -67,3 +70,33 @@ func (s *Service) deleteSecurityGroup(ctx context.Context, securityGroupID strin
...
+		for _, group := range out.SecurityGroups {
+			if group != nil {
+				arn := composeFakeArn(sgService, sgResourcePrefix+*group.GroupId)
+				resource, err := composeAWSResource(aws.String(arn), converters.TagsToMap(group.Tags))
+				if err != nil {
+					s.scope.Error(err, "compose aws resource error: %v")
+					return false
+				}
+				resources = append(resources, resource)
+			}
```
Skarlso, line 93:
```suggestion
			if group == nil {
				continue
			}

			arn := composeFakeArn(sgService, sgResourcePrefix+*group.GroupId)
			resource, err := composeAWSResource(aws.String(arn), converters.TagsToMap(group.Tags))
			if err != nil {
				s.scope.Error(err, "compose aws resource error: %v")
				return false
			}
			resources = append(resources, resource)

```

pkg/cloud/services/gc/ec2.go
```diff
@@ -67,3 +70,33 @@ func (s *Service) deleteSecurityGroup(ctx context.Context, securityGroupID strin
...
+
+	resources := []*AWSResource{}
+	err := s.ec2Client.DescribeSecurityGroupsPages(input, func(out *ec2.DescribeSecurityGroupsOutput, last bool) bool {
+		for _, group := range out.SecurityGroups {
+			if group != nil {
+				arn := composeFakeArn(sgService, sgResourcePrefix+*group.GroupId)
+				resource, err := composeAWSResource(aws.String(arn), converters.TagsToMap(group.Tags))
+				if err != nil {
+					s.scope.Error(err, "compose aws resource error: %v")
+					return false
```
Skarlso, line 90: Are you sure you want to stop the whole iteration here instead of just skipping this one resource on an error? So `continue`?

pkg/cloud/services/gc/ec2.go
```diff
@@ -67,3 +70,33 @@ func (s *Service) deleteSecurityGroup(ctx context.Context, securityGroupID strin
...
+				}
+				resources = append(resources, resource)
+			}
+		}
+		return true
+	})
+	if err != nil {
+		return nil, err
```
Skarlso, line 98:
```suggestion
		return nil, fmt.Errorf("failed list security groups: %w", err)
```

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -128,3 +131,138 @@ func (s *Service) deleteTargetGroup(ctx context.Context, targetGroupARN string)
...
+func (s *Service) describeLoadBalancers(ctx context.Context) ([]string, error) {
+	var names []string
+	err := s.elbClient.DescribeLoadBalancersPagesWithContext(ctx, &elb.DescribeLoadBalancersInput{}, func(r *elb.DescribeLoadBalancersOutput, last bool) bool {
+		for _, lb := range r.LoadBalancerDescriptions {
+			names = append(names, *lb.LoadBalancerName)
+		}
+		return true
+	})
+	if err != nil {
+		return nil, err
```
Skarlso, line 145:
```suggestion
		return nil, fmt.Errorf("failed to describe load balancers: %w", err)
```

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -128,3 +131,138 @@ func (s *Service) deleteTargetGroup(ctx context.Context, targetGroupARN string)
...
+func (s *Service) describeLoadBalancersV2(ctx context.Context) ([]string, error) {
+	var arns []string
+	err := s.elbv2Client.DescribeLoadBalancersPagesWithContext(ctx, &elbv2.DescribeLoadBalancersInput{}, func(r *elbv2.DescribeLoadBalancersOutput, last bool) bool {
+		for _, lb := range r.LoadBalancers {
+			arns = append(arns, *lb.LoadBalancerArn)
+		}
+		return true
+	})
+	if err != nil {
+		return nil, err
```
Skarlso, line 161:
```suggestion
		return nil, fmt.Errorf("failed to describe load balancer v2s: %w", err)
```

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -128,3 +131,138 @@ func (s *Service) deleteTargetGroup(ctx context.Context, targetGroupARN string)
...
+	}
+
+	return arns, nil
+}
+
+func (s *Service) describeTargetgroups(ctx context.Context) ([]string, error) {
+	groups, err := s.elbv2Client.DescribeTargetGroupsWithContext(ctx, &elbv2.DescribeTargetGroupsInput{})
+	if err != nil {
+		return nil, err
```
Skarlso, line 170:
```suggestion
		return nil, fmt.Errorf("failed to describe target groups: %w", err)
```

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -128,3 +131,138 @@ func (s *Service) deleteTargetGroup(ctx context.Context, targetGroupARN string)
...
+	}
+
+	targetGroups := []string{}
+	for _, group := range groups.TargetGroups {
+		if err != nil {
+			return nil, err
+		}
```
Skarlso, line 177:
```suggestion
```

This just copy paste mistake I assume :)

pkg/cloud/services/gc/loadbalancer.go
```diff
@@ -128,3 +131,138 @@ func (s *Service) deleteTargetGroup(ctx context.Context, targetGroupARN string)
...
+		targetGroups = append(targetGroups, *group.TargetGroupArn)
+	}
+	return targetGroups, nil
+}
+
+// / getProviderOwnedLoadBalancers gets cloud provider created LB(ELB) for this cluster, filtering by tag: kubernetes.io/cluster/<cluster-name>:owned.
+func (s *Service) getProviderOwnedLoadBalancers(ctx context.Context) ([]*AWSResource, error) {
+	names, err := s.describeLoadBalancers(ctx)
+	if err != nil {
+		return nil, err
```
Skarlso, line 187: Please wrap all your errors. :D

pkg/cloud/services/gc/service.go
```diff
@@ -88,3 +106,23 @@ func (fn *ResourceCleanupFuncs) Execute(ctx context.Context, resources []*AWSRes
...
+
+// ResourceCollectFunc is a function type to collect resources for a specific AWS service type.
+type ResourceCollectFunc func(ctx context.Context) ([]*AWSResource, error)
+
+// ResourceCollectFuncs is a collection of ResourceCollectFunc.
+type ResourceCollectFuncs []ResourceCollectFunc
+
+// Execute will execute all the defined collect functions against the aws resources.
+func (fn *ResourceCollectFuncs) Execute(ctx context.Context) ([]*AWSResource, error) {
+	resources := []*AWSResource{}
```
Skarlso, line 118:
```suggestion
	var resources []*AWSResource
```

pkg/cloud/services/gc/service.go
```diff
@@ -88,3 +106,23 @@ func (fn *ResourceCleanupFuncs) Execute(ctx context.Context, resources []*AWSRes
...
 }
+
+// ResourceCollectFunc is a function type to collect resources for a specific AWS service type.
+type ResourceCollectFunc func(ctx context.Context) ([]*AWSResource, error)
+
+// ResourceCollectFuncs is a collection of ResourceCollectFunc.
+type ResourceCollectFuncs []ResourceCollectFunc
+
+// Execute will execute all the defined collect functions against the aws resources.
+func (fn *ResourceCollectFuncs) Execute(ctx context.Context) ([]*AWSResource, error) {
```
Skarlso, line 117: Since we aren't modifying the custom slice type, you don't need a pointer receiver here. Incidentally, this will also fix having to dereference the type in the below range loop. Same goes for the other type. :)

pkg/cloud/services/gc/utils.go
Skarlso, line 1: Let's give this a more descriptive name, please. Something like compose.go or partition_handler.go...

Skarlso: I don't really prefer utils for anything because it doesn't convey any meaning or hint of the content inside. :)

pkg/cloud/services/gc/loadbalancer.go
richardcase, line 145: Would be good to wrap errors in these new "collect" functions to get some extra context.

pkg/cloud/services/elb/loadbalancer.go
wyike: How about something like:
# chunkELBs is similar to chunkResources in pkg/cloud/services/gc

pkg/cloud/services/gc/ec2.go
wyike: How about I remove `if group == nil`?
I come up that the describe result out.SecurityGroups should be have values or empty and the values (the group) should have no chance to be nil. 
I refer to https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/securitygroup/securitygroups.go#L330 before.

pkg/cloud/services/gc/ec2.go
wyike: Yeah! `continue`  will be better!

pkg/cloud/services/gc/service.go
wyike: Thanks @Skarlso for the valuable comment.
I'll update it as well as previous existing `fn *ResourceCleanupFuncs` https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/gc/service.go#L82 . Is this oK?

pkg/cloud/services/gc/ec2.go
Skarlso: Yep, agreed. I was thinking about why it was there. :D But I agree with you. Removing it should be okay. :)

pkg/cloud/services/gc/service.go
Skarlso: Yep 👍

pkg/cloud/services/gc/utils.go
Skarlso: If you really hate it, you can leave it as util.. It's just so... ew. :D :D

Skarlso:
Neat, thanks a really great job here. :)

/lgtm

From me. I leave the final approve to @richardcase as he has the most context in this. If I understood the referenced issue well I think this is okay now. :)

pkg/cloud/services/gc/utils.go
wyike: I moved `chunkResources` to gc/loadbalancer.go since it is only used by that file and renamed `utils.go` to `compose.go`.

How is this?

wyike:
Thanks @Skarlso , addressed the last comment https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4129#discussion_r1136526817.
I also did another round of testing again today. Wait @richardcase  for the final review.

pkg/cloud/services/gc/utils.go
Skarlso: Nice! :) Thank you, @wyike. That sounds awesome. :)

wyike:
Squash commits to one locally before merging, so pushing a new.

richardcase:
Adding this back after squash (thanks @Skarlso ):

/lgtm

PR #4130, opened 2023-03-07
Remove CAPA tagging from unmanaged network resources

### LINKED ISSUE #3692: Static (Unmanaged) VPC Placement has Messy Tag Handling
/kind bug

Create two clusters with the same static VPC/Subnets/NAT Gateways
 - CAPA will reconcile both clusters and only the values of the most recently reconciled cluster will apply leading to them playing stomp out every reconciliation loop. This applies to AdditionalTags too which could get dicey if you think of something like a billing tag where you get them flipping the value back and forth.
 - When you delete these clusters there is code in delete VPC/Subnet/NAT Gateways to return nil immediately as CAPA didn't create them, this leaves the tags all in place and clutters up the VPC.

Since CAPA was told to not manage the VPC I don't think we should do the tag reconciliation if the user is trying to manage it themselves.

I would like to add code to the reconciliation loops which looks like this code which will not reconcile tags if the VPC is unmanaged. Any feedback on that is appreciated.

**Environment:**

### LINKED ISSUE #3919: CAPA is modifiying tags of user owned VPCs, resulting into reconilcation loops
/kind bug

Creating a new cluster with custom vpc (and ipv6 enabled).
We deploy our VPC via terraform and it should be used my multiple clusters, each cluster has its own dedicated subnets. 
However, we see that capa is modifying existing tags (and even removes some) from the several vpc resources, even if its not meant to be managed at all by capa like the `Egress Only Internet Gateway`. 

Even worse, it also ends up in reconciliation loops as it reads (preview tags) and updates the tags all the time.

Do not touch resources what are not part of its business.

Screenshot 2022-12-06 at 16-07-59 run-niZ3DRMr54ndv1wo Runs networking-aws-eu-west-2 Traefik Terraform Cloud

**Environment:**

- Cluster-api-provider-aws version: 2.0.2
- Kubernetes version: (use `kubectl version`):  1.24.7

### LINKED ISSUE #3948: Proposal: Tagging of unmanaged network items should not be done by CAPA
/kind feature

At the moment, CAPA manages tags on unmanaged network items. This should be avoided. They cause several issues when they aren't removed, and they cause several problems when they are added to the wrong subnet, or many cases when security groups can't be deleted, etc.

Cleaning them up is a pain, and we don't want to end up causing a potential disaster when we delete the wrong tag.

I propose that we shouldn't manage tags on unmanaged items. This is partially already being done, but some tags are added nevertheless. These should be documented and left up to the user to do.

**Environment:**

Ankitasw:
**What type of PR is this?**
/kind feature

This PR removed tagging from unmanaged network resources such that the unmanaged resource tagging is taken care by the users themselves.

Fixes #3948 
Fixes #3919 
Fixes #3692 

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
action required
If the **TagUnmanagedNetworkResources** feature flag is set to false then CAPA tagging of unmanaged network resources will be disabled.
```

### DIFF
--- a/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
+++ b/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
@@ -87,12 +87,13 @@ type AWSManagedControlPlaneReconciler struct {
 	Recorder  record.EventRecorder
 	Endpoints []scope.ServiceEndpoint
 
-	EnableIAM             bool
-	AllowAdditionalRoles  bool
-	WatchFilterValue      string
-	ExternalResourceGC    bool
-	AlternativeGCStrategy bool
-	WaitInfraPeriod       time.Duration
+	EnableIAM                    bool
+	AllowAdditionalRoles         bool
+	WatchFilterValue             string
+	ExternalResourceGC           bool
+	AlternativeGCStrategy        bool
+	WaitInfraPeriod              time.Duration
+	TagUnmanagedNetworkResources bool
 }
 
 // SetupWithManager is used to setup the controller.
@@ -173,13 +174,14 @@ func (r *AWSManagedControlPlaneReconciler) Reconcile(ctx context.Context, req ct
 	}
 
 	managedScope, err := scope.NewManagedControlPlaneScope(scope.ManagedControlPlaneScopeParams{
-		Client:               r.Client,
-		Cluster:              cluster,
-		ControlPlane:         awsControlPlane,
-		ControllerName:       strings.ToLower(awsManagedControlPlaneKind),
-		EnableIAM:            r.EnableIAM,
-		AllowAdditionalRoles: r.AllowAdditionalRoles,
-		Endpoints:            r.Endpoints,
+		Client:                       r.Client,
+		Cluster:                      cluster,
+		ControlPlane:                 awsControlPlane,
+		ControllerName:               strings.ToLower(awsManagedControlPlaneKind),
+		EnableIAM:                    r.EnableIAM,
+		AllowAdditionalRoles:         r.AllowAdditionalRoles,
+		Endpoints:                    r.Endpoints,
+		TagUnmanagedNetworkResources: r.TagUnmanagedNetworkResources,
 	})
 	if err != nil {
 		return reconcile.Result{}, fmt.Errorf("failed to create scope: %w", err)
--- a/pkg/cloud/services/network/subnets.go
+++ b/pkg/cloud/services/network/subnets.go
@@ -53,12 +53,6 @@ func (s *Service) reconcileSubnets() error {
 		s.scope.SetSubnets(subnets)
 	}()
 
-	// Describe subnets in the vpc.
-	existing, err := s.describeVpcSubnets()
-	if err != nil {
-		return err
-	}
-
 	unmanagedVPC := s.scope.VPC().IsUnmanaged(s.scope.Name())
 
 	if len(subnets) == 0 {
@@ -70,6 +64,7 @@ func (s *Service) reconcileSubnets() error {
 		}
 		// If we a managed VPC and have no subnets then create subnets. There will be 1 public and 1 private subnet
 		// for each az in a region up to a maximum of 3 azs
+		var err error
 		s.scope.Info("no subnets specified, setting defaults")
 		subnets, err = s.getDefaultSubnets()
 		if err != nil {
@@ -81,6 +76,12 @@ func (s *Service) reconcileSubnets() error {
 			s.scope.Error(err, "failed to patch object to save subnets")
 			return err
 		}
+	}
+
+	// Describe subnets in the vpc.
+	existing, err := s.describeVpcSubnets()
+	if err != nil {
+		return err
 	}
 
 	if s.scope.SecondaryCidrBlock() != nil {
--- a/pkg/cloud/services/network/subnets.go
+++ b/pkg/cloud/services/network/subnets.go
@@ -53,6 +53,19 @@ func (s *Service) reconcileSubnets() error {
 		s.scope.SetSubnets(subnets)
 	}()
 
+	var (
+		err      error
+		existing infrav1.Subnets
+	)
+
+	// Describing the VPC Subnets tags the resources.
+	if s.scope.TagUnmanagedNetworkResources() {
+		// Describe subnets in the vpc.
+		if existing, err = s.describeVpcSubnets(); err != nil {
+			return err
+		}
+	}
+
 	unmanagedVPC := s.scope.VPC().IsUnmanaged(s.scope.Name())
 
 	if len(subnets) == 0 {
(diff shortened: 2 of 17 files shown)


Ankitasw:
This is WIP as I have to test scenarios manually, but I would be open for suggestions at this point for early feedback.

Skarlso:
So far so good. The vpc looks okay I think.

Ankitasw:
PR is tested now, and ready for review.

Ankitasw:
New commit contains only doc changes, hence not running e2e tests again.
PR is ready for review now.

richardcase:
@Ankitasw - this generally looks good to me. However, this is changing the documented behaviour as we originally stated this:

> **Note**: The subnet tagging above is taken care by the CAPA controllers but additionalTags provided by users won't be propagated to the unmanaged VPC subnets.

With a change in behaviour that could potentially break users we should probably do this as part of a version bump. wdyt?

Ankitasw:
Yes sure, but I think this issue was raised by @wyike as some customer env is breaking , so do you think we could somehow make a new release with this change?

richardcase:
> @richardcase Yes sure, but I think this issue was raised by @wyike as some customer env is breaking and they would need to provide fix soon, so do you think we could somehow make a new release with this change?

Lets chat via zoom to work out a solution.

wyike:
> @richardcase Yes sure, but I think this issue was raised by @wyike as some customer env is breaking and they would need to provide fix soon, so do you think we could somehow make a new release with this change?

One info is I gave them the RCA and come up with an workaround to them (to remove staled tags on both aws server and the managed cluster awscluster spec). So far I haven't heard response from them. Maybe they will be no urgent.

Ankitasw:
> One info is I gave them the RCA and come up with an workaround to them (to remove staled tags on both aws server and the managed cluster awscluster spec, so that managed cluster reconcile won't re-tag other cluster tags to the shared subnets continuously). So far I haven't heard response from them. Maybe they can be good with workaround and no urgent.

Ok thanks @wyike , then we can hold this PR off for next version bump. Although we have not planned it yet. But please share your views @richardcase if any.

pkg/cloud/services/network/subnets.go
```diff
@@ -83,6 +78,12 @@ func (s *Service) reconcileSubnets() error {
 		}
 	}
 
+	// Describe subnets in the vpc.
+	existing, err := s.describeVpcSubnets()
```
wyike, line 82: curious what would be the difference if moving from L57 to L82 :)

wyike: For subnets tags of the managed cluster and unmanaged cluster, I didn't see the difference and even from the testing, not sure if I understand wrong.
What's our expectations? :)

Skarlso:
I think we should also remove tagging from security groups and the likes, right?

Ankitasw:
> I think we should also remove tagging from security groups and the likes, right?

Oops, I thought that's already handled, but thanks, I would take a look.

Ankitasw:
@richardcase that sounds good to me 👍

Skarlso:
@Ankitasw Uh, aren't additional security groups defined by the user explicitly in the template? I might be wrong here. So I guess just filter them out in that case? Whichever SG is in the additionally defined sgs leave those alone?

Ankitasw:
> @Ankitasw Uh, aren't additional security groups defined by the user explicitly in the template? I might be wrong here. So I guess just filter them out in that case? Whichever SG is in the additionally defined sgs leave those alone?

Oh you are right, I didn't mean additional security groups, I was talking about security groups created by CAPA.
This makes sense, I will also take care of removing tagging from additional security groups.

pkg/cloud/services/network/subnets.go
Ankitasw: `describeVpcSubnets` actually tags the existing subnets. But you are right, only moving this wont help.
I will take a look at subnet tagging again.

Ankitasw:
@Skarlso another though, we dont add tags to additionalSecurityGroups but I might be wrong, could you point the saem in the code , so that my confusion is cleared 😓

Skarlso:
@Ankitasw Ah, we tag security group overrides. Maybe that's something we shouldn't do? 🤔 Overrides are optional and provided by a user, right?

Skarlso:
Ah but we only tag if it's NOT an override.

```go
		if !s.securityGroupIsAnOverride(existing.ID) {
			// Make sure tags are up to date.
			if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
				buildParams := s.getSecurityGroupTagParams(existing.Name, existing.ID, role)
				tagsBuilder := tags.New(&buildParams, tags.WithEC2(s.EC2Client))
				if err := tagsBuilder.Ensure(existing.Tags); err != nil {
					return false, err
				}
				return true, nil
			}, awserrors.GroupNotFound); err != nil {
				return errors.Wrapf(err, "failed to ensure tags on security group %q", existing.ID)
			}
		}
```

So maybe you're right that this is fine. :D

Skarlso:
Okay, let's say we are fine then, and hopefully we don't mess with user security groups hehehe.

pkg/cloud/services/network/subnets.go
Ankitasw: Refer slack thread [here](https://kubernetes.slack.com/archives/CD6U2V71N/p1678252002376529).

Skarlso:
Ah you already ran them GitHub just collapsed those.

/unholy

Skarlso:
Unholy. Thanks autocorrect... 🤦 

/unhold

Ankitasw:
@Skarlso I think I need to add feature gate as well.

Skarlso:
Oh snap. Sorry I thought that already happened for some reason.🤣🤣🤣

Ankitasw:
no its on me 😄 I should have put WIP.

Skarlso:
/assign

@Ankitasw If you don't mind, I will add the flag to get this in for the release.

Ankitasw:
@Skarlso hey thanks for doing it, didn't get time to do it 😕

Skarlso:
Don't worry about it. 😊😊 you did the hard part. 😁😁

Skarlso:
That's the remediation test ( which is expected ).

controlplane/eks/controllers/awsmanagedcontrolplane_controller.go
```diff
@@ -87,12 +87,13 @@ type AWSManagedControlPlaneReconciler struct {
...
-	ExternalResourceGC    bool
-	AlternativeGCStrategy bool
-	WaitInfraPeriod       time.Duration
+	EnableIAM                    bool
+	AllowAdditionalRoles         bool
+	WatchFilterValue             string
+	ExternalResourceGC           bool
+	AlternativeGCStrategy        bool
+	WaitInfraPeriod              time.Duration
+	TagUnmanagedNetworkResources bool
```
richardcase, line 96: For a different PR, we should think about grouping these setting into a struct

richardcase (commented):
Based on the fact that the new behaviour is not enabled by default i think this looks good. I have also updated the release note.

/lgtm

PR #4135, opened 2023-03-09
Set ASG DesiredCapacity value only if MachinePool replicas is between min and max size of the AWSMachinePool

Fedosin:
**What type of PR is this?**
/kind bug

Now if we don't define replicas value in MachinePool spec, it is [defaulted](https://github.com/kubernetes-sigs/cluster-api/blob/main/exp/api/v1beta1/machinepool_webhook.go#L54-L56) to 1 by the webhook. Unfortunately, if minimum size of AWSMachinePool is greater than that, ASG creation fails with 400 error from AWS.

To prevent this, we ensure that the desired number of replicas for ASG is set only if MachinePool replicas is between min and max size of the AWSMachinePool, otherwise we return an error.

For clusters with externally managed replicas we can ignore this problem and do not set desired replicas during ASG creation at all. In this case MachinePool replicas value will be defaulted to AWSMachinePool MinSize by AWS and then updated by external autoscaler to the right value.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Set ASG DesiredCapacity value only if MachinePool replicas is between min and max size of the AWSMachinePool.
```

### DIFF
--- a/pkg/cloud/services/autoscaling/autoscalinggroup.go
+++ b/pkg/cloud/services/autoscaling/autoscalinggroup.go
@@ -157,45 +157,45 @@ func (s *Service) GetASGByName(scope *scope.MachinePoolScope) (*expinfrav1.AutoS
 }
 
 // CreateASG runs an autoscaling group.
-func (s *Service) CreateASG(scope *scope.MachinePoolScope) (*expinfrav1.AutoScalingGroup, error) {
-	subnets, err := s.SubnetIDs(scope)
+func (s *Service) CreateASG(machinePoolScope *scope.MachinePoolScope) (*expinfrav1.AutoScalingGroup, error) {
+	subnets, err := s.SubnetIDs(machinePoolScope)
 	if err != nil {
 		return nil, fmt.Errorf("getting subnets for ASG: %w", err)
 	}
 
 	input := &expinfrav1.AutoScalingGroup{
-		Name:                 scope.Name(),
-		MaxSize:              scope.AWSMachinePool.Spec.MaxSize,
-		MinSize:              scope.AWSMachinePool.Spec.MinSize,
+		Name:                 machinePoolScope.Name(),
+		MaxSize:              machinePoolScope.AWSMachinePool.Spec.MaxSize,
+		MinSize:              machinePoolScope.AWSMachinePool.Spec.MinSize,
 		Subnets:              subnets,
-		DefaultCoolDown:      scope.AWSMachinePool.Spec.DefaultCoolDown,
-		CapacityRebalance:    scope.AWSMachinePool.Spec.CapacityRebalance,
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -157,45 +157,61 @@ func (s *Service) GetASGByName(scope *scope.MachinePoolScope) (*expinfrav1.AutoS
...
 	}
 
-	if scope.MachinePool.Spec.Replicas != nil {
-		input.DesiredCapacity = scope.MachinePool.Spec.Replicas
+	if machinePoolScope.MachinePool.Spec.Replicas != nil {
+		input.DesiredCapacity = machinePoolScope.MachinePool.Spec.Replicas
+	}
+
+	if scope.ReplicasExternallyManaged(machinePoolScope.MachinePool) {
+		if input.DesiredCapacity != nil && *input.DesiredCapacity > machinePoolScope.AWSMachinePool.Spec.MaxSize {
```
Ankitasw, line 181: Good find.
But why this validation only needed for externally managed replicas and not for CAPI managed?

Fedosin: Hi Ankita!

That's a great question! The reason why we only need validation for externally managed replicas is that they are controlled by an external tool, whereas with CAPI managed replicas, users define them manually.

When the initial number of replicas is incorrect for externally managed replicas, CAPA will automatically update it to the correct value after start [here](https://github.com/newrelic-forks/cluster-api-provider-aws/blob/main/exp/controllers/awsmachinepool_controller.go#L277-L288). However, when the number of replicas is incorrect for CAPI managed replicas, the MachinePool will remain out of sync until the user manually updates it. This can lead to inconsistencies in the cluster, which is why we need to report an error.

Ankitasw:
Overall it looks good so ...

/lgtm


cc @Skarlso @richardcase for another review.

richardcase:
If the machinepool isn't externally managed what should we do? For example if MP.Replicas is defaulted to 1 and the AWMP.Min = 2, should we error or have similar logic?

@Ankitasw @Skarlso - wdyt?

Skarlso:
Huh I have to understand first what's going on here to answer that question. :D

Ankitasw:
> If the machinepool isn't externally managed what should we do? For example if MP.Replicas is defaulted to 1 and the AWMP.Min = 2, should we error or have similar logic?
> 
> @Ankitasw @Skarlso - wdyt?

I was thinking the same, on how to cover this for capi managed replicas, but didn't get any ideas.
I think we could have similar logic right? What would happen if we remove the if check(thinking out loud)

richardcase:
Shall we discuss on Monday?

richardcase (commented):
/lgtm

Issue #4134, opened 2023-03-08
AWSManagedMachinePool does not work in combination with cluster-autoscaler


ionutbalutoiu:
/kind bug

**What steps did you take and what happened:**

The use case is similar with the one described in issue https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2022. The only differences being:
* Usage of `AWSManagedMachinePool` resource instead of `AWSMachinePool` resource.
* The `MachinePool` resource is annotated with `cluster.x-k8s.io/replicas-managed-by: "external-autoscaler"` as described [here](https://cluster-api-aws.sigs.k8s.io/topics/machinepools.html#autoscaling) (because I'm using `AWS` autoscaler provider).

**What did you expect to happen:**

cluster-autoscaler should manage the `DesiredInstances` of the autoscaling group and the `AWSManagedMachinePool` controller should ignore changes to the `replicas` and `DesiredInstances`.

The `AWSManagedMachinePool` controller should respect the annotation for externally managed autoscaler.

**Anything else you would like to add:**

This issue is already fixed with `AWSMachinePool` controller via PR https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3683. 

All we need to do is to make sure that the `AWSManagedMachinePool` controller respects the annotation for externally managed autoscaler.
At the moment, there is no knowledge of the new annotation in the `AWSManagedMachinePool` controller.

In my scenario (using `v2.0.2` release), the autoscaler would adjust the `DesiredInstances` of the autoscaling group, and this  code path would cause the `AWSManagedMachinePool` controller to override those changes:

 * https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/28bc9b8756d6d7f73038604038ddb3ccf4b22396/exp/controllers/awsmanagedmachinepool_controller.go#L237
 * https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/28bc9b8756d6d7f73038604038ddb3ccf4b22396/pkg/cloud/services/eks/eks.go#L108
 * https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/28bc9b8756d6d7f73038604038ddb3ccf4b22396/pkg/cloud/services/eks/nodegroup.go#L542
[...]

PR #4137, opened 2023-03-12
Allow external autoscaler for EKS managed node groups

ionutbalutoiu:
**What type of PR is this?**
/kind bug

The `AWSManagedMachinePool` controller should respect the annotation for externally managed autoscaler.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4134

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

```release-note
The `AWSManagedMachinePool` controller should respect the annotation for the externally managed autoscaler.
```

### DIFF
--- a/exp/controllers/awsmanagedmachinepool_controller.go
+++ b/exp/controllers/awsmanagedmachinepool_controller.go
@@ -85,7 +85,7 @@ func (r *AWSManagedMachinePoolReconciler) SetupWithManager(ctx context.Context, 
 		Complete(r)
 }
 
-// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools;machinepools/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools;machinepools/status,verbs=get;list;watch;patch
 // +kubebuilder:rbac:groups=core,resources=events,verbs=get;list;watch;create;patch
 // +kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=awsmanagedcontrolplanes;awsmanagedcontrolplanes/status,verbs=get;list;watch
 // +kubebuilder:rbac:groups=infrastructure.cluster.x-k8s.io,resources=awsmanagedmachinepools,verbs=get;list;watch;update;patch;delete
@@ -193,7 +193,7 @@ func (r *AWSManagedMachinePoolReconciler) Reconcile(ctx context.Context, req ctr
 }
 
 func (r *AWSManagedMachinePoolReconciler) reconcileNormal(
-	_ context.Context,
+	ctx context.Context,
 	machinePoolScope *scope.ManagedMachinePoolScope,
 	ec2Scope scope.EC2Scope,
 ) (ctrl.Result, error) {
@@ -235,7 +235,7 @@ func (r *AWSManagedMachinePoolReconciler) reconcileNormal(
 		conditions.MarkTrue(machinePoolScope.ManagedMachinePool, expinfrav1.LaunchTemplateReadyCondition)
 	}
 
-	if err := ekssvc.ReconcilePool(); err != nil {
+	if err := ekssvc.ReconcilePool(ctx); err != nil {
 		return reconcile.Result{}, errors.Wrapf(err, "failed to reconcile machine pool for AWSManagedMachinePool %s/%s", machinePoolScope.ManagedMachinePool.Namespace, machinePoolScope.ManagedMachinePool.Name)
 	}
 
--- a/pkg/cloud/scope/managednodegroup.go
+++ b/pkg/cloud/scope/managednodegroup.go
@@ -88,20 +88,26 @@ func NewManagedMachinePoolScope(params ManagedMachinePoolScopeParams) (*ManagedM
 		return nil, errors.Errorf("failed to create aws session: %v", err)
 	}
 
-	helper, err := patch.NewHelper(params.ManagedMachinePool, params.Client)
+	ammpHelper, err := patch.NewHelper(params.ManagedMachinePool, params.Client)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to init patch helper")
+		return nil, errors.Wrap(err, "failed to init AWSManagedMachinePool patch helper")
+	}
+	mpHelper, err := patch.NewHelper(params.MachinePool, params.Client)
+	if err != nil {
+		return nil, errors.Wrap(err, "failed to init MachinePool patch helper")
 	}
 
 	return &ManagedMachinePoolScope{
-		Logger:               *params.Logger,
-		Client:               params.Client,
+		Logger:                     *params.Logger,
+		Client:                     params.Client,
+		patchHelper:                ammpHelper,
+		capiMachinePoolPatchHelper: mpHelper,
+
 		Cluster:              params.Cluster,
 		ControlPlane:         params.ControlPlane,
 		ManagedMachinePool:   params.ManagedMachinePool,
 		MachinePool:          params.MachinePool,
 		EC2Scope:             params.InfraCluster,
-		patchHelper:          helper,
 		session:              session,
 		serviceLimiters:      serviceLimiters,
 		controllerName:       params.ControllerName,
@@ -114,7 +120,8 @@ func NewManagedMachinePoolScope(params ManagedMachinePoolScopeParams) (*ManagedM
 type ManagedMachinePoolScope struct {
 	logger.Logger
 	client.Client
-	patchHelper *patch.Helper
+	patchHelper                *patch.Helper
+	capiMachinePoolPatchHelper *patch.Helper
 
 	Cluster            *clusterv1.Cluster
 	ControlPlane       *ekscontrolplanev1.AWSManagedControlPlane
@@ -255,6 +262,14 @@ func (s *ManagedMachinePoolScope) PatchObject() error {
 			expinfrav1.EKSNodegroupReadyCondition,
 			expinfrav1.IAMNodegroupRolesReadyCondition,
 		}})
+}
+
+// PatchCAPIMachinePoolObject persists the capi machinepool configuration and status.
+func (s *ManagedMachinePoolScope) PatchCAPIMachinePoolObject(ctx context.Context) error {
+	return s.capiMachinePoolPatchHelper.Patch(
+		ctx,
+		s.MachinePool,
+	)
 }
 
 // Close closes the current scope persisting the control plane configuration and status.
--- a/pkg/cloud/services/eks/eks.go
+++ b/pkg/cloud/services/eks/eks.go
@@ -90,7 +90,7 @@ func (s *Service) DeleteControlPlane() (err error) {
 }
 
 // ReconcilePool is the entrypoint for ManagedMachinePool reconciliation.
-func (s *NodegroupService) ReconcilePool() error {
+func (s *NodegroupService) ReconcilePool(ctx context.Context) error {
 	s.scope.Debug("Reconciling EKS nodegroup")
 
 	if err := s.reconcileNodegroupIAMRole(); err != nil {
@@ -105,7 +105,7 @@ func (s *NodegroupService) ReconcilePool() error {
 	}
 	conditions.MarkTrue(s.scope.ManagedMachinePool, expinfrav1.IAMNodegroupRolesReadyCondition)
 
-	if err := s.reconcileNodegroup(); err != nil {
+	if err := s.reconcileNodegroup(ctx); err != nil {
 		conditions.MarkFalse(
 			s.scope.ManagedMachinePool,
 			expinfrav1.EKSNodegroupReadyCondition,
--- a/pkg/cloud/services/eks/nodegroup.go
+++ b/pkg/cloud/services/eks/nodegroup.go
@@ -17,6 +17,7 @@ limitations under the License.
 package eks
 
 import (
+	"context"
 	"fmt"
 
 	"github.com/aws/aws-sdk-go/aws"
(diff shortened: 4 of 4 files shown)


Skarlso:
Uh... :D

exp/controllers/awsmanagedmachinepool_controller.go
```diff
@@ -85,7 +85,7 @@ func (r *AWSManagedMachinePoolReconciler) SetupWithManager(ctx context.Context,
 		Complete(r)
 }
 
-// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools;machinepools/status,verbs=get;list;watch
+// +kubebuilder:rbac:groups=cluster.x-k8s.io,resources=machinepools;machinepools/status,verbs=get;list;watch;patch
```
Skarlso, line 88: I don't see this change in the generated RBAC.

Ankitasw:
@ionutbalutoiu I think we need a force push to try triggering the tests again.

ionutbalutoiu:
> @ionutbalutoiu I think we need a force push to try triggering the tests again.

I did a force push with the same codebase in this PR.

Ankitasw:
Didn't work 😞

ionutbalutoiu:
> Didn't work 😞

Perhaps the CI system is not working now.

Let me know if I need to do anything else to re-run tests.

exp/controllers/awsmanagedmachinepool_controller.go
Skarlso: Ah okay cool

ionutbalutoiu:
> @ionutbalutoiu: The following test **failed**, say `/retest` to rerun all failed tests or `/retest-required` to rerun all mandatory failed tests:
> 
> Test name	Commit	Details	Required	Rerun command
> pull-cluster-api-provider-aws-test	[61f7282](https://github.com/kubernetes-sigs/cluster-api-provider-aws/commit/61f7282265eabd370dc7fcbb4658f4452e7c6d8d)	[link](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/4137/pull-cluster-api-provider-aws-test/1635503129061494784)	true	`/test pull-cluster-api-provider-aws-test`
> [Full PR test history](https://prow.k8s.io/pr-history?org=kubernetes-sigs&repo=cluster-api-provider-aws&pr=4137). [Your PR dashboard](https://prow.k8s.io/pr?query=is%3Apr%20state%3Aopen%20author%3Aionutbalutoiu). Please help us cut down on flakes by [linking to](https://git.k8s.io/community/contributors/devel/sig-testing/flaky-tests.md#filing-issues-for-flaky-tests) an [open issue](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues?q=is:issue+is:open) when you hit one in your PR.

Fixed this `lint` failure. Also, rebase the PR against latest upstream `main` code.

Skarlso:
@ionutbalutoiu Do you have an example output using this PR? A manual test run with steps?

ionutbalutoiu:
> @ionutbalutoiu Do you have an example output using this PR? A manual test run with steps?

Yes. Here are the steps:

* Deploy an EKS cluster using the manifests examples given in this comment:
  https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4134#issuecomment-1465316610
* Deploy `cluster-autoscaler` using `AWS` provider using the instructions from:
  https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md
* Build a container image with this patch using `make docker-build-amd64`
* Patch the existing `capa-controller-manager` image with:
  ```
  kubectl patch deployment -n capa-system capa-controller-manager -p '{"spec":{"template":{"spec":{"containers":[{"name": "manager", "image":"docker.io/ionutbalutoiu/cluster-api-aws-controller:v2.0.2-patched", "imagePullPolicy": "Always"}]}}}}'
  ```

  NOTE: I built the image `ionutbalutoiu/cluster-api-aws-controller:v2.0.2-patched`, which is `v2.0.2` codebase with this PR cherry-picked (because we can cleanly back-port this fix to `release-2.0` branch).

* Deploy multiple pods, so that `cluster-autoscaler` will adjust the nodegroup ASG `DesiredCapacity` in order to add more nodes to the cluster.

  NOTE: The nodegroup size `m5.large` in the above example manifest allows maximum `29` pods per Kubernetes node. If you deploy more than `29` pods, the `cluster-autoscaler` will add more nodes.

* Using this PR, the `cluster-api-provider-aws` will never override the `DesiredCapacity` as described in the original issue https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4134.

  Also, we will see the following `cluster-api-provider-aws` log message:
  https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4137/files#diff-ed6ead8e5e6ab86b4693e189f4568484aeda9b9c660e21f70f08fab9f5da13e3R540-R542

  This means that CAPA manager acknowledges the new `DesiredCapacity` and updates the `MachinePool` spec accordingly.

Skarlso:
Sorry, let me be more specific. Since we don't have an E2E test case, please showcase that this PR works as expected. :)

ionutbalutoiu:
> Sorry, let me be more specific. Since we don't have an E2E test case, please showcase that this PR works as expected. :)

So, using the above instructions and [cluster spec](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4134#issuecomment-1465316610), I deployed:
* 1x EKS managed control-plane
* 1x EKS managed node group (minimum size 1 and maximum size 5) using `m5.large` size for the EC2 VMs in the managed ASG.

After deployment finishes, I end up with a single machine registered to the EKS control-plane, since the cluster-api AWS provider sets the EKS node group desired size to 1, by default.

I configure the `cluster-autoscaler` with `AWS` provider using these [instructions](https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/aws/README.md). The purpose of this component is to adjust the EKS managed node group size, if pods cannot be scheduled on existing Kubernetes nodes (for example: max pods per node was reached).

Before using the patched image with this PR applied, let's see what happens if we use the current stable CAPA image:
```
registry.k8s.io/cluster-api-aws/cluster-api-aws-controller:v2.0.2
```

I also set a higher log verbosity (`--v=5`) to see more CAPA manager logs.

Since I'm using `m5.large` size, each node is allowed maximum `29` pods. The max pods value can be checked in the CLI with:
```bash
$ kubectl get node NODE_NAME -o jsonpath='{.status.allocatable.pods}'
29
```

Let's simulate some load now. I'll be using this simple Nginx deployment with 30 replicas:
```yaml
apiVersion: apps/v1
kind: Deployment
metadata:
  name: nginx-deployment
  labels:
    app: nginx
spec:
  replicas: 30
  selector:
    matchLabels:
      app: nginx
  template:
    metadata:
      labels:
        app: nginx
    spec:
      containers:
      - name: nginx
        image: nginx:stable
        ports:
        - containerPort: 80
```

The max pods per node limit is reached, `cluster-autoscaler` detects this, and it increases the node group size:
```
I0320 12:15:47.412417       1 scale_up.go:481] Estimated 1 nodes needed in eks-default_eks-cluster-pool-0-22c3373b-951b-7fe0-e454-4e90c233ca84
I0320 12:15:47.412442       1 scale_up.go:601] Final scale-up plan: [{eks-default_eks-cluster-pool-0-22c3373b-951b-7fe0-e454-4e90c233ca84 1->2 (max: 5)}]
I0320 12:15:47.412469       1 scale_up.go:700] Scale-up: setting group eks-default_eks-cluster-pool-0-22c3373b-951b-7fe0-e454-4e90c233ca84 size to 2
I0320 12:15:47.412504       1 auto_scaling_groups.go:248] Setting asg eks-default_eks-cluster-pool-0-22c3373b-951b-7fe0-e454-4e90c233ca84 size to 2
```

An extra Kubernetes node is registered to the cluster, and all the pods are `Running`.

However, on the next reconciliation loop, the CAPA manager will set the EKS node group desired size back to 1:
```
I0320 12:33:02.731379       1 logger.go:72] "identity: Nodegroup has no desired size or differs from replicas, updating scaling configuration" identityKey="/default" nodegroup=0xc000b67aa0
```

This is not expected, since I'm using `external-autoscaler` annotation as described in the [documentation](https://cluster-api-aws.sigs.k8s.io/topics/machinepools.html#autoscaling).

Now, cleanup the Nginx deployment, scale down the EKS node group back to 1, and test the same scenario with the patched CAPA image `docker.io/ionutbalutoiu/cluster-api-aws-controller:v2.0.2-patched` (built with this PR applied on top of `v2.0.2` release).

We will simulate the same `cluster-autoscaler` scale-up with the above Nginx deployment, and when the next `AWSManagedMachinePool` reconciliation loop happens, the CAPA manager will not override the EKS node group desired size.

This log message confirms that the CAPA manager respects the `external-autoscaler` annotation, and it merely adjusts the `MachinePool.Spec.Replicas` value, as expected:
```
I0320 13:01:09.563279       1 logger.go:68] "identity: Setting MachinePool replicas to node group DesiredCapacity" identityKey="/default" local=1 external=2
```

No matter how many times the `cluster-autoscaler` changes the EKS node group desired size, the CAPA manager will not override the node group desired size, and it will only adjust the `MachinePool.Spec.Replicas` value.

This confirms that this PR is working as expected, and it fixes the underlying issue.

richardcase:
> @richardcase Do you see any problems with a potential race where the desired capacity keeps getting overridden then set again then overridden...? I don't think I do, but I would love to have a second opinion on that.

I think there may be a problem with the value flip/flopping. And even more so if you are using GitOps and you set the `MachinePool.Spec.Replicas` to a value. 

I'm wondering if we need to make sure that the [scalingConfig function](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7c82a73b304c6b85d19dfe87f4ef7ffdd5d9a7ef/pkg/cloud/services/eks/nodegroup.go#L97:L116) is aware that its externally managed and keep `DesiredSize` to what it currently is and just change min/max....instead of setting it back to 1.  

wdyt @Skarlso @ionutbalutoiu ?

ionutbalutoiu:
> > @richardcase Do you see any problems with a potential race where the desired capacity keeps getting overridden then set again then overridden...? I don't think I do, but I would love to have a second opinion on that.
> 
> I think there may be a problem with the value flip/flopping. And even more so if you are using GitOps and you set the `MachinePool.Spec.Replicas` to a value.

When using GitOps, you would ignore differences in `spec.replicas`.
It's in the [official docs](https://cluster-api-aws.sigs.k8s.io/topics/machinepools.html#autoscaling), and this is what I'm doing also.

> 
> I'm wondering if we need to make sure that the [scalingConfig function](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7c82a73b304c6b85d19dfe87f4ef7ffdd5d9a7ef/pkg/cloud/services/eks/nodegroup.go#L97:L116) is aware that its externally managed and keep `DesiredSize` to what it currently is and just change min/max....instead of setting it back to 1.
> 
> wdyt @Skarlso @ionutbalutoiu ?

If we have this PR, the `DesiredSize` is never adjusted since we have this check:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7c82a73b304c6b85d19dfe87f4ef7ffdd5d9a7ef/pkg/cloud/services/eks/nodegroup.go#L99-L101

And with this PR chages, we update `s.scope.MachinePool.Spec.Replicas` before `scalingConfig` is called.
That was my idea to make sure that `AWSManagedMachinePool` controller doesn't conflict with `cluster-autoscaler`.

However, if you would have `min` / `max` size changes, those would be applied, but the `DesiredSize` is always what's being observed on the node group, which is supposed to be scaled outside the `AWSManagedMachinePool` controller.

It's exactly the same fix adopted for the `AWSMachinePool` controller:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3683/files#diff-9c6c4cb74d0076521d841bb705940e29b5b7b6f81a7c7af31f8586616beec425R272

The only problem is that it wasn't ported to the `AWSManagedMachinePool` controller.

ionutbalutoiu:
> Its also got be thinking, do we need to consider the same thing as #4135 (not as part of this PR though)?

The `cluster-autoscaler` would never adjust the `DesiredSize` outside the interval `[min, max]` node group size.
If it did, the AWS API would error out.

I believe that it's safe to assume that having `external-autoscaler` annotation, the `DesiredSize` is always correct.

Because, what we do is to observe what is the current AWS managed node group size. AWS API would never allow the external autoscaler to set a wrong `DesiredSize`.

richardcase:
> The `cluster-autoscaler` would never adjust the `DesiredSize` outside the interval `[min, max]` node group size. If it did, the AWS API would error out.

So what happens when a user first creates a cluster and the nodegroup is initially created, and they set min to 2 and max to 4?

richardcase:
> When using GitOps, you would ignore differences in `spec.replicas`.
> It's in the [official docs](https://cluster-api-aws.sigs.k8s.io/topics/machinepools.html#autoscaling), and this is what I'm doing also.

Good point, thanks for pointing that out.

richardcase:
> And with this PR chages, we update `s.scope.MachinePool.Spec.Replicas` before `scalingConfig` is called.
> That was my idea to make sure that `AWSManagedMachinePool` controller doesn't conflict with `cluster-autoscaler`.

Ah yes it does, I misread it. Doh.  Thanks for pointing that out @ionutbalutoiu .

Lets run some e2e and get this merged.

richardcase:
Until the EKS e2e passes:

/hold

And from my side:

/approve

richardcase:
Both e2e have passed so:

/unhold

ionutbalutoiu:
> So what happens when a user first creates a cluster and the nodegroup is initially created, and they set min to 2 and max to 4? We can look at that later as it doesn't relate to this PR.

This is a very good point, and I opened the issue https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4166 to keep this tracked.

richardcase:
@ionutbalutoiu - i forgot to say thank you for this change :bowing_man: 

The next releases will be:
- v2.1.0 - we will cut a release branch from main for this (so this changer will be included in that)
- v1.5.3 - we could potentially backport this, but i don't think we need to

Skarlso:
Fantastic. Thanks. Sorry I was super busy today and didn't get to this. Thank you for resolving. Nice job all. 😊

PR #4147, opened 2023-03-15
Fix InstanceMetadataOptions defaults

muraee:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:
Fixes `InstanceMetadataOptions` defaults not being set for AWSMachines

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4037#discussion_r1129064749

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Fix InstanceMetadataOptions defaults not being set for AWSMachines
```

PR #4149, opened 2023-03-15
add a nil check on options

luthermonson:
/kind bug

after this code was added i got here with a nil options which caused a panic. adding a safety nil check


Ankitasw:
@luthermonson there is another PR #4147 open for this

Issue #4184, opened 2023-03-30
Implement MachinePool Machines clusterAPI proposal


cnmcavoy:
/kind feature

**Describe the solution you'd like**
Provide an implementation for MachinePool Machines based off of [this proposal](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220209-machinepool-machines.md).

AWSMachinePools & AWSManagedMachinePools should implement a status field for `infrastructureMachineKind` and `infrastructureMachineSelector` and manage these status fields in the controller reconcile loop. Each controller should spawn AWSMachine resources to represent EC2 instances managed by their external AWS autoscaling group resource. These AWSMachines need to have special handling in the AWSMachine reconciliation to skip any unnecessary steps for externally managed nodes.

**Anything else you would like to add:**
See the clusterAPI implementation of their part of the proposal here: https://github.com/kubernetes-sigs/cluster-api/pull/7938

com6056:
Do we think it is worth updating the docs in https://cluster-api-aws.sigs.k8s.io/topics/machinepools.html#autoscaling to reflect that in it's current state, you can't use the `clusterapi` `cluster-autoscaler` provider to scale `MachinePools`? It looks like this feature is required to be able to scale down properly.

PR #4194, opened 2023-04-03
Improve launch template diff detection for MachinePools

### LINKED ISSUE #4070: AWSMachinePool detects LaunchTemplate diffs multiple times and updates LaunchTemplate versions multiple times with no changes
/kind bug

The AWSMachinePool reconciler creates new LaunchTemplate versions without changes being made.

These are the MachinePool, AWSMachinePool, and KubeadmConfig I'm working with.
```
---
kind: MachinePool
metadata:
  annotations:
    cluster.x-k8s.io/replicas-managed-by: external-autoscaler
  name: worker-private-efficient
spec:
  clusterName: awscmhdev2
  replicas: 2
  template:
    metadata: {}
    spec:
      bootstrap:
        configRef:
          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
          kind: KubeadmConfig
          name: worker-private-efficient
      clusterName: awscmhdev2
      infrastructureRef:
        kind: AWSMachinePool
        name: worker-private-efficient
      version: v1.24.7
---
kind: AWSMachinePool
metadata:
  annotations:
    cluster-api-provider-aws: "true"
  name: worker-private-efficient
spec:
  additionalTags:
    k8s.io/cluster-autoscaler/awscmhdev2: owned
    k8s.io/cluster-autoscaler/enabled: "true"
  awsLaunchTemplate:
    additionalSecurityGroups:
    - filters:
      - name: tag:Name
        values:
        - capi-hostports
      - name: tag:network-zone
        values:
        - qa
      - name: tag:region
        values:
        - us-east-2
    ami:
      id: ami-0f6b6efcd422b9d85
    iamInstanceProfile: nodes.cluster-api-provider-aws.sigs.k8s.io
    instanceType: c5a.2xlarge
    rootVolume:
      deviceName: /dev/sda1
      encrypted: true
      iops: 16000
      size: 100
      throughput: 1000
      type: gp3
    sshKeyName: ""
  capacityRebalance: true
  defaultCoolDown: 5m0s
  maxSize: 5
  minSize: 0
  mixedInstancesPolicy:
    overrides:
    - instanceType: c5a.2xlarge
    - instanceType: m5a.2xlarge
  subnets:
  - filters:
    - name: availability-zone-id
      values:
      - use2-az1
    - name: tag:type
      values:
      - nodes
  - filters:
    - name: availability-zone-id
      values:
      - use2-az2
    - name: tag:type
      values:
      - nodes
  - filters:
    - name: availability-zone-id
      values:
      - use2-az3
    - name: tag:type
      values:
      - nodes
---
apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
kind: KubeadmConfig
metadata:
  name: worker-private-efficient
spec:
  files:
  - content: |
      vm.max_map_count=262144
    path: /etc/sysctl.d/90-vm-max-map-count.conf
  - content: |
      fs.inotify.max_user_instances=256
    path: /etc/sysctl.d/91-fs-inotify.conf
  format: cloud-config
  joinConfiguration:
    nodeRegistration:
      kubeletExtraArgs:
        cloud-provider: aws
        eviction-hard: memory.available<500Mi,nodefs.available<10%
        kube-reserved: cpu=500m,memory=2Gi,ephemeral-storage=1Gi
        node-labels: role.node.kubernetes.io/worker=true
        protect-kernel-defaults: "true"
        system-reserved: cpu=500m,memory=1Gi,ephemeral-storage=1Gi
      name: '{{ ds.meta_data.local_hostname }}'
  preKubeadmCommands:
  - sudo systemctl restart systemd-sysctl
```

This is the current state of the `runcmd` in the LaunchTemplate in AWS, version 1566.
```
runcmd:
  - "sudo systemctl restart systemd-sysctl"
  - kubeadm join --config /run/kubeadm/kubeadm-join-config.yaml  && echo success > /run/cluster-api/bootstrap-success.complete
```

I apply a change to add a command to the KubeadmConfig, such as this.
```
  preKubeadmCommands:
  - sudo systemctl restart systemd-sysctl
  - echo "hello world"
```

After waiting for 5-10 minutes I see the LaunchTemplate has 2 new versions present in AWS, 1567 and 1568.

In the capa-system logs, I see that there are multiple regular entries that detect a diff and update the ASGs, despite the specs above remaining stable. Every 3 detected diffs, a new LaunchTemplate is created, despite no changes having been made.
```
I0215 17:42:03.907312       1 logger.go:68] "Reconciling AWSMachinePool" controller="awsmachinepool" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSMachinePool" aWSMachinePool="capi-awscmhdev2/worker-private-efficient" namespace="capi-awscmhdev2" name="worker-private-efficient" reconcileID=48412683-7c0a-498f-963f-e6f09897acac machinePool="capi-awscmhdev2/worker-private-efficient" cluster="capi-awscmhdev2/awscmhdev2"
I0215 17:42:03.907928       1 logger.go:68] "checking for existing launch template"
I0215 17:42:04.223663       1 logger.go:68] "creating new version for launch template" existing=&{Name:worker-private-efficient IamInstanceProfile:nodes.cluster-api-provider-aws.sigs.k8s.io AMI:{ID:0xc008ac0e68 EKSOptimizedLookupType:<nil>} ImageLookupFormat: ImageLookupOrg: ImageLookupBaseOS: InstanceType:c5a.2xlarge RootVolume:<nil> SSHKeyName:<nil> VersionNumber:0xc007612640 AdditionalSecurityGroups:[{ID:0xc008ac1048 Filters:[]} {ID:0xc008ac1108 Filters:[]} {ID:0xc008ac11c8 Filters:[]}] SpotMarketOptions:<nil>} incoming=&{Name: IamInstanceProfile:nodes.cluster-api-provider-aws.sigs.k8s.io AMI:{ID:0xc008244ce0 EKSOptimizedLookupType:<nil>} ImageLookupFormat: ImageLookupOrg: ImageLookupBaseOS: InstanceType:c5a.2xlarge RootVolume:0xc0083f4500 SSHKeyName:0xc008244cf0 VersionNumber:<nil> AdditionalSecurityGroups:[{ID:<nil> Filters:[{Name:tag:Name Values:[capi-hostports]} {Name:tag:network-zone Values:[qa]} {Name:tag:region Values:[us-east-2]}]}] SpotMarketOptions:<nil>}
I0215 17:42:05.001698       1 logger.go:68] "got a mixed diff here" incoming=&{InstancesDistribution:<nil> Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]} existing=&{InstancesDistribution:0xc000ceb3e0 Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]}
I0215 17:42:05.001714       1 logger.go:68] "updating AutoScalingGroup"
I0215 17:42:07.601630       1 request.go:601] Waited for 1.181851585s due to client-side throttling, not priority and fairness, request: GET:https://internal-awscmhdev2-apiserver-754512249.us-east-2.elb.amazonaws.com:6443/apis/node.k8s.io/v1?timeout=10s
I0215 17:42:16.426360       1 logger.go:68] "Reconciling AWSMachinePool" controller="awsmachinepool" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSMachinePool" aWSMachinePool="capi-awscmhdev2/worker-private-efficient" namespace="capi-awscmhdev2" name="worker-private-efficient" reconcileID=6e9a5e38-8159-4f47-af70-9a2017af6266 machinePool="capi-awscmhdev2/worker-private-efficient" cluster="capi-awscmhdev2/awscmhdev2"
I0215 17:42:16.427022       1 logger.go:68] "checking for existing launch template"
I0215 17:42:16.603086       1 logger.go:68] "got a mixed diff here" incoming=&{InstancesDistribution:<nil> Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]} existing=&{InstancesDistribution:0xc00841bf80 Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]}
I0215 17:42:16.603112       1 logger.go:68] "updating AutoScalingGroup"
I0215 17:42:19.024945       1 request.go:601] Waited for 1.195479079s due to client-side throttling, not priority and fairness, request: GET:https://internal-awscmhdev2-apiserver-754512249.us-east-2.elb.amazonaws.com:6443/apis/prometheusservice.aws.crossplane.io/v1alpha1?timeout=10s
I0215 17:44:19.847853       1 logger.go:68] "Reconciling AWSMachinePool" controller="awsmachinepool" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSMachinePool" aWSMachinePool="capi-awscmhdev2/worker-private-efficient" namespace="capi-awscmhdev2" name="worker-private-efficient" reconcileID=e16f4bf4-b71d-4213-8133-36f002610601 machinePool="capi-awscmhdev2/worker-private-efficient" cluster="capi-awscmhdev2/awscmhdev2"
I0215 17:44:19.848650       1 logger.go:68] "checking for existing launch template"
I0215 17:44:20.009601       1 logger.go:68] "got a mixed diff here" incoming=&{InstancesDistribution:<nil> Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]} existing=&{InstancesDistribution:0xc00663d2c0 Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]}
I0215 17:44:20.009616       1 logger.go:68] "updating AutoScalingGroup"
I0215 17:44:22.389003       1 request.go:601] Waited for 1.181136775s due to client-side throttling, not priority and fairness, request: GET:https://internal-awscmhdev2-apiserver-754512249.us-east-2.elb.amazonaws.com:6443/apis/s3.aws.crossplane.io/v1alpha3?timeout=10s
I0215 17:52:25.983736       1 logger.go:68] "Reconciling AWSMachinePool" controller="awsmachinepool" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSMachinePool" aWSMachinePool="capi-awscmhdev2/worker-private-efficient" namespace="capi-awscmhdev2" name="worker-private-efficient" reconcileID=4d16218e-b7e9-4175-ac5e-bb119320a17e machinePool="capi-awscmhdev2/worker-private-efficient" cluster="capi-awscmhdev2/awscmhdev2"
I0215 17:52:25.984388       1 logger.go:68] "checking for existing launch template"
I0215 17:52:26.299754       1 logger.go:68] "creating new version for launch template" existing=&{Name:worker-private-efficient IamInstanceProfile:nodes.cluster-api-provider-aws.sigs.k8s.io AMI:{ID:0xc008c48628 EKSOptimizedLookupType:<nil>} ImageLookupFormat: ImageLookupOrg: ImageLookupBaseOS: InstanceType:c5a.2xlarge RootVolume:<nil> SSHKeyName:<nil> VersionNumber:0xc0017d12a0 AdditionalSecurityGroups:[{ID:0xc008c48808 Filters:[]} {ID:0xc008c488c8 Filters:[]} {ID:0xc008c48988 Filters:[]}] SpotMarketOptions:<nil>} incoming=&{Name: IamInstanceProfile:nodes.cluster-api-provider-aws.sigs.k8s.io AMI:{ID:0xc008791300 EKSOptimizedLookupType:<nil>} ImageLookupFormat: ImageLookupOrg: ImageLookupBaseOS: InstanceType:c5a.2xlarge RootVolume:0xc00b3d8d70 SSHKeyName:0xc008791310 VersionNumber:<nil> AdditionalSecurityGroups:[{ID:<nil> Filters:[{Name:tag:Name Values:[capi-hostports]} {Name:tag:network-zone Values:[qa]} {Name:tag:region Values:[us-east-2]}]}] SpotMarketOptions:<nil>}
I0215 17:52:27.046433       1 logger.go:68] "got a mixed diff here" incoming=&{InstancesDistribution:<nil> Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]} existing=&{InstancesDistribution:0xc006a203c0 Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]}
I0215 17:52:27.046450       1 logger.go:68] "updating AutoScalingGroup"
I0215 17:52:29.445737       1 request.go:601] Waited for 1.176573904s due to client-side throttling, not priority and fairness, request: GET:https://internal-awscmhdev2-apiserver-754512249.us-east-2.elb.amazonaws.com:6443/apis/bootstrap.cluster.x-k8s.io/v1alpha3?timeout=10s
I0215 17:52:38.274727       1 logger.go:68] "Reconciling AWSMachinePool" controller="awsmachinepool" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSMachinePool" aWSMachinePool="capi-awscmhdev2/worker-private-efficient" namespace="capi-awscmhdev2" name="worker-private-efficient" reconcileID=58867893-5c4a-46ca-8bda-140ff8c010b8 machinePool="capi-awscmhdev2/worker-private-efficient" cluster="capi-awscmhdev2/awscmhdev2"
I0215 17:52:38.275485       1 logger.go:68] "checking for existing launch template"
I0215 17:52:38.655905       1 logger.go:68] "got a mixed diff here" incoming=&{InstancesDistribution:<nil> Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]} existing=&{InstancesDistribution:0xc008f68990 Overrides:[{InstanceType:c5a.2xlarge} {InstanceType:m5a.2xlarge}]}
I0215 17:52:38.655921       1 logger.go:68] "updating AutoScalingGroup"
I0215 17:52:41.064482       1 request.go:601] Waited for 1.193402218s due to client-side throttling, not priority and fairness, request: GET:https://internal-awscmhdev2-apiserver-754512249.us-east-2.elb.amazonaws.com:6443/apis/cluster.x-k8s.io/v1beta1?timeout=10s
```

**Environment:**

```
$ kubectl version --short
Flag --short has been deprecated, and will be removed in the future. The --short output will become the default.
Client Version: v1.25.1
Kustomize Version: v4.5.7
Server Version: v1.24.7
```

cnmcavoy:
**What type of PR is this?**
/kind bug

Improve launch template diff detection for MachinePools:
  * Store tags related to the launch template sooner (unnecessary churn occurs when reconciliation errors part-way)
  * Handle missing ASG case better (presently 2 launch template versions are created for the initial asg)
  * Defaulting webhook should set necessary defaults for MachinePool launch templates (unbounded churn due to missing defaults)

Fixes #4070

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Improve launch template diff detection for MachinePools to reduce unnecessary launch template versions. Logging around launch template diffs has also been improved to highlight why new versions are created.
```

### DIFF
--- a/exp/controllers/awsmachinepool_controller.go
+++ b/exp/controllers/awsmachinepool_controller.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/davecgh/go-spew/spew"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/pkg/errors"
@@ -229,6 +228,13 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 	ec2Svc := r.getEC2Service(ec2Scope)
 	asgsvc := r.getASGService(clusterScope)
 
+	canUpdateLaunchTemplate := func() (bool, error) {
+		// If there is a change: before changing the template, check if there exist an ongoing instance refresh,
+		// because only 1 instance refresh can be "InProgress". If template is updated when refresh cannot be started,
+		// that change will not trigger a refresh. Do not start an instance refresh if only userdata changed.
+		return asgsvc.CanStartASGInstanceRefresh(machinePoolScope)
+	}
+
 	// Find existing ASG
 	asg, err := r.findASG(machinePoolScope, asgsvc)
 	if err != nil {
@@ -238,15 +244,6 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 	if asg == nil {
 		machinePoolScope.Debug("asg not created yet")
 		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, expinfrav1.ASGNotFoundReason, "Unable to find matching ASG")
-	}
-
-	canUpdateLaunchTemplate := func() (bool, error) {
-		// If there is a change: before changing the template, check if there exist an ongoing instance refresh,
-		// because only 1 instance refresh can be "InProgress". If template is updated when refresh cannot be started,
-		// that change will not trigger a refresh. Do not start an instance refresh if only userdata changed.
-		return asgsvc.CanStartASGInstanceRefresh(machinePoolScope)
-	}
-	if asg == nil {
 		canUpdateLaunchTemplate = func() (bool, error) {
 			return true, nil
 		}
@@ -427,9 +424,16 @@ func (r *AWSMachinePoolReconciler) updatePool(machinePoolScope *scope.MachinePoo
 		"subnets of machinePoolScope", subnetIDs,
 		"subnets of existing asg", existingASG.Subnets)
 	less := func(a, b string) bool { return a < b }
-	subnetChanges := cmp.Diff(subnetIDs, existingASG.Subnets, cmpopts.SortSlices(less)) != ""
+	subnetDiff := cmp.Diff(subnetIDs, existingASG.Subnets, cmpopts.SortSlices(less))
+	if subnetDiff != "" {
+		machinePoolScope.Debug("asg subnet diff detected", "diff", subnetDiff)
+	}
 
-	if asgNeedsUpdates(machinePoolScope, existingASG) || subnetChanges {
+	asgDiff := diffASG(machinePoolScope, existingASG)
+	if asgDiff != "" {
+		machinePoolScope.Debug("asg diff detected", "diff", subnetDiff)
+	}
+	if asgDiff != "" || subnetDiff != "" {
 		machinePoolScope.Info("updating AutoScalingGroup")
 
 		if err := asgSvc.UpdateASG(machinePoolScope); err != nil {
@@ -516,42 +520,23 @@ func (r *AWSMachinePoolReconciler) findASG(machinePoolScope *scope.MachinePoolSc
 	return asg, nil
 }
 
-// asgNeedsUpdates compares incoming AWSMachinePool and compares against existing ASG.
-func asgNeedsUpdates(machinePoolScope *scope.MachinePoolScope, existingASG *expinfrav1.AutoScalingGroup) bool {
+// diffASG compares incoming AWSMachinePool and compares against existing ASG.
+func diffASG(machinePoolScope *scope.MachinePoolScope, existingASG *expinfrav1.AutoScalingGroup) string {
+	detectedMachinePoolSpec := machinePoolScope.MachinePool.Spec.DeepCopy()
+
 	if !scope.ReplicasExternallyManaged(machinePoolScope.MachinePool) {
-		if machinePoolScope.MachinePool.Spec.Replicas != nil {
-			if existingASG.DesiredCapacity == nil {
-				machinePoolScope.Info("DesiredCapacity unset, diff detected", "incoming", *machinePoolScope.MachinePool.Spec.Replicas, "existing", existingASG.DesiredCapacity)
-				return true
-			}
-			if *machinePoolScope.MachinePool.Spec.Replicas != *existingASG.DesiredCapacity {
-				machinePoolScope.Info("DesiredCapacity diff detected", "incoming", *machinePoolScope.MachinePool.Spec.Replicas, "existing", *existingASG.DesiredCapacity)
-				return true
-			}
-		} else if existingASG.DesiredCapacity != nil {
-			machinePoolScope.Info("Replicas nil, DesiredCapacity diff detected", "incoming", machinePoolScope.MachinePool.Spec.Replicas, "existing", *existingASG.DesiredCapacity)
-			return true
-		}
+		detectedMachinePoolSpec.Replicas = existingASG.DesiredCapacity
+	}
+	if diff := cmp.Diff(machinePoolScope.MachinePool.Spec, *detectedMachinePoolSpec); diff != "" {
+		return diff
 	}
 
-	if machinePoolScope.AWSMachinePool.Spec.MaxSize != existingASG.MaxSize {
-		machinePoolScope.Info("MinSize diff detected", "incoming", machinePoolScope.AWSMachinePool.Spec.MaxSize, "existing", existingASG.MaxSize)
-		return true
-	}
-
-	if machinePoolScope.AWSMachinePool.Spec.MinSize != existingASG.MinSize {
-		machinePoolScope.Info("MinSize diff detected", "incoming", machinePoolScope.AWSMachinePool.Spec.MinSize, "existing", existingASG.MinSize)
-		return true
-	}
-
-	if machinePoolScope.AWSMachinePool.Spec.CapacityRebalance != existingASG.CapacityRebalance {
-		machinePoolScope.Info("CapacityRebalance diff detected", "incoming", machinePoolScope.AWSMachinePool.Spec.CapacityRebalance, "existing", existingASG.CapacityRebalance)
-		return true
-	}
-
+	detectedAWSMachinePoolSpec := machinePoolScope.AWSMachinePool.Spec.DeepCopy()
+	detectedAWSMachinePoolSpec.MaxSize = existingASG.MaxSize
+	detectedAWSMachinePoolSpec.MinSize = existingASG.MinSize
+	detectedAWSMachinePoolSpec.CapacityRebalance = existingASG.CapacityRebalance
 	{
 		mixedInstancesPolicy := machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy
-
 		// InstancesDistribution is optional, and the default values come from AWS, so
 		// they are not set by the AWSMachinePool defaulting webhook. If InstancesDistribution is
 		// not set, we use the AWS values for the purpose of comparison.
@@ -561,12 +546,11 @@ func asgNeedsUpdates(machinePoolScope *scope.MachinePoolScope, existingASG *expi
 		}
 
 		if !cmp.Equal(mixedInstancesPolicy, existingASG.MixedInstancesPolicy) {
-			machinePoolScope.Info("MixedInstancesPolicy diff detected", "incoming", spew.Sprintf("%#v", mixedInstancesPolicy), "existing", spew.Sprintf("%#v", existingASG.MixedInstancesPolicy))
-			return true
+			detectedAWSMachinePoolSpec.MixedInstancesPolicy = existingASG.MixedInstancesPolicy
 		}
 	}
 
-	return false
+	return cmp.Diff(machinePoolScope.AWSMachinePool.Spec, *detectedAWSMachinePoolSpec)
 }
 
 // getOwnerMachinePool returns the MachinePool object owning the current resource.
--- a/exp/controllers/awsmachinepool_controller.go
+++ b/exp/controllers/awsmachinepool_controller.go
@@ -19,7 +19,9 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
+	"github.com/davecgh/go-spew/spew"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/pkg/errors"
@@ -227,13 +229,58 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 	ec2Svc := r.getEC2Service(ec2Scope)
 	asgsvc := r.getASGService(clusterScope)
 
+	// Find existing ASG
+	asg, err := r.findASG(machinePoolScope, asgsvc)
+	if err != nil {
+		conditions.MarkUnknown(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, expinfrav1.ASGNotFoundReason, err.Error())
+		return ctrl.Result{}, err
+	}
+	if asg == nil {
+		machinePoolScope.Debug("asg not created yet")
+		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, expinfrav1.ASGNotFoundReason, "Unable to find matching ASG")
+	}
+
 	canUpdateLaunchTemplate := func() (bool, error) {
 		// If there is a change: before changing the template, check if there exist an ongoing instance refresh,
 		// because only 1 instance refresh can be "InProgress". If template is updated when refresh cannot be started,
 		// that change will not trigger a refresh. Do not start an instance refresh if only userdata changed.
 		return asgsvc.CanStartASGInstanceRefresh(machinePoolScope)
 	}
+	if asg == nil {
+		canUpdateLaunchTemplate = func() (bool, error) {
+			return true, nil
+		}
+	}
+
 	runPostLaunchTemplateUpdateOperation := func() error {
+		launchTemplateID := machinePoolScope.GetLaunchTemplateIDStatus()
+		asgName := machinePoolScope.Name()
+		resourceServiceToUpdate := []scope.ResourceServiceToUpdate{
+			{
+				ResourceID:      &launchTemplateID,
+				ResourceService: ec2Svc,
+			},
+			{
+				ResourceID:      &asgName,
+				ResourceService: asgsvc,
+			},
+		}
+		// Update the AWSMachinePool annotation immediately after creating a new template
+		err = ec2Svc.ReconcileTags(machinePoolScope, resourceServiceToUpdate)
+		if err != nil {
+			return errors.Wrap(err, "error updating tags")
+		}
+
+		asg, err := r.findASG(machinePoolScope, asgsvc)
+		if err != nil {
+			return fmt.Errorf("failed to find asg %q for instance refresh: %w", asgName, err)
+		}
+		if asg == nil {
+			machinePoolScope.Debug("asg not created yet, skipping instance refresh")
+			r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, expinfrav1.ASGNotFoundReason, "Unable to find matching ASG")
+			return nil
+		}
+
 		// skip instance refresh if explicitly disabled
 		if machinePoolScope.AWSMachinePool.Spec.RefreshPreferences != nil && machinePoolScope.AWSMachinePool.Spec.RefreshPreferences.Disable {
 			machinePoolScope.Debug("instance refresh disabled, skipping instance refresh")
@@ -260,20 +307,13 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 	// set the LaunchTemplateReady condition
 	conditions.MarkTrue(machinePoolScope.AWSMachinePool, expinfrav1.LaunchTemplateReadyCondition)
 
-	// Find existing ASG
-	asg, err := r.findASG(machinePoolScope, asgsvc)
-	if err != nil {
-		conditions.MarkUnknown(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, expinfrav1.ASGNotFoundReason, err.Error())
-		return ctrl.Result{}, err
-	}
-
 	if asg == nil {
 		// Create new ASG
 		if err := r.createPool(machinePoolScope, clusterScope); err != nil {
 			conditions.MarkFalse(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, expinfrav1.ASGProvisionFailedReason, clusterv1.ConditionSeverityError, err.Error())
 			return ctrl.Result{}, err
 		}
-		return ctrl.Result{}, nil
+		return ctrl.Result{RequeueAfter: time.Minute}, nil
 	}
 
 	if scope.ReplicasExternallyManaged(machinePoolScope.MachinePool) {
@@ -292,23 +332,6 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 	if err := r.updatePool(machinePoolScope, clusterScope, asg); err != nil {
 		machinePoolScope.Error(err, "error updating AWSMachinePool")
 		return ctrl.Result{}, err
-	}
-
-	launchTemplateID := machinePoolScope.GetLaunchTemplateIDStatus()
-	asgName := machinePoolScope.Name()
-	resourceServiceToUpdate := []scope.ResourceServiceToUpdate{
-		{
-			ResourceID:      &launchTemplateID,
-			ResourceService: ec2Svc,
-		},
-		{
-			ResourceID:      &asgName,
-			ResourceService: asgsvc,
-		},
-	}
-	err = ec2Svc.ReconcileTags(machinePoolScope, resourceServiceToUpdate)
-	if err != nil {
-		return ctrl.Result{}, errors.Wrap(err, "error updating tags")
 	}
 
 	// Make sure Spec.ProviderID is always set.
@@ -347,7 +370,7 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
 
 	if asg == nil {
 		machinePoolScope.Debug("Unable to locate ASG")
-		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, "NoASGFound", "Unable to find matching ASG")
+		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, expinfrav1.ASGNotFoundReason, "Unable to find matching ASG")
 	} else {
 		machinePoolScope.SetASGStatus(asg.Status)
 		switch asg.Status {
@@ -374,7 +397,7 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
 
 	if launchTemplate == nil {
 		machinePoolScope.Debug("Unable to locate launch template")
-		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, "NoASGFound", "Unable to find matching ASG")
+		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, expinfrav1.ASGNotFoundReason, "Unable to find matching ASG")
 		controllerutil.RemoveFinalizer(machinePoolScope.AWSMachinePool, expinfrav1.MachinePoolFinalizer)
 		return ctrl.Result{}, nil
 	}
@@ -497,29 +520,50 @@ func (r *AWSMachinePoolReconciler) findASG(machinePoolScope *scope.MachinePoolSc
 func asgNeedsUpdates(machinePoolScope *scope.MachinePoolScope, existingASG *expinfrav1.AutoScalingGroup) bool {
 	if !scope.ReplicasExternallyManaged(machinePoolScope.MachinePool) {
 		if machinePoolScope.MachinePool.Spec.Replicas != nil {
-			if existingASG.DesiredCapacity == nil || *machinePoolScope.MachinePool.Spec.Replicas != *existingASG.DesiredCapacity {
+			if existingASG.DesiredCapacity == nil {
+				machinePoolScope.Info("DesiredCapacity unset, diff detected", "incoming", *machinePoolScope.MachinePool.Spec.Replicas, "existing", existingASG.DesiredCapacity)
+				return true
+			}
+			if *machinePoolScope.MachinePool.Spec.Replicas != *existingASG.DesiredCapacity {
+				machinePoolScope.Info("DesiredCapacity diff detected", "incoming", *machinePoolScope.MachinePool.Spec.Replicas, "existing", *existingASG.DesiredCapacity)
 				return true
 			}
 		} else if existingASG.DesiredCapacity != nil {
+			machinePoolScope.Info("Replicas nil, DesiredCapacity diff detected", "incoming", machinePoolScope.MachinePool.Spec.Replicas, "existing", *existingASG.DesiredCapacity)
 			return true
 		}
 	}
 
 	if machinePoolScope.AWSMachinePool.Spec.MaxSize != existingASG.MaxSize {
+		machinePoolScope.Info("MinSize diff detected", "incoming", machinePoolScope.AWSMachinePool.Spec.MaxSize, "existing", existingASG.MaxSize)
 		return true
 	}
 
 	if machinePoolScope.AWSMachinePool.Spec.MinSize != existingASG.MinSize {
+		machinePoolScope.Info("MinSize diff detected", "incoming", machinePoolScope.AWSMachinePool.Spec.MinSize, "existing", existingASG.MinSize)
 		return true
 	}
 
 	if machinePoolScope.AWSMachinePool.Spec.CapacityRebalance != existingASG.CapacityRebalance {
+		machinePoolScope.Info("CapacityRebalance diff detected", "incoming", machinePoolScope.AWSMachinePool.Spec.CapacityRebalance, "existing", existingASG.CapacityRebalance)
 		return true
 	}
 
-	if !cmp.Equal(machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy, existingASG.MixedInstancesPolicy) {
-		machinePoolScope.Info("got a mixed diff here", "incoming", machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy, "existing", existingASG.MixedInstancesPolicy)
-		return true
+	{
+		mixedInstancesPolicy := machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy
+
+		// InstancesDistribution is optional, and the default values come from AWS, so
+		// they are not set by the AWSMachinePool defaulting webhook. If InstancesDistribution is
+		// not set, we use the AWS values for the purpose of comparison.
+		if mixedInstancesPolicy != nil && mixedInstancesPolicy.InstancesDistribution == nil {
+			mixedInstancesPolicy = machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy.DeepCopy()
+			mixedInstancesPolicy.InstancesDistribution = existingASG.MixedInstancesPolicy.InstancesDistribution
+		}
+
+		if !cmp.Equal(mixedInstancesPolicy, existingASG.MixedInstancesPolicy) {
+			machinePoolScope.Info("MixedInstancesPolicy diff detected", "incoming", spew.Sprintf("%#v", mixedInstancesPolicy), "existing", spew.Sprintf("%#v", existingASG.MixedInstancesPolicy))
+			return true
+		}
 	}
 
 	return false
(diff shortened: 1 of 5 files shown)


exp/api/v1beta2/awsmachinepool_webhook.go
```diff
@@ -162,4 +163,15 @@ func (r *AWSMachinePool) Default() {
 		log.Info("DefaultCoolDown is zero, setting 300 seconds as default")
 		r.Spec.DefaultCoolDown.Duration = 300 * time.Second
 	}
+	if r.Spec.MixedInstancesPolicy != nil {
+		if r.Spec.MixedInstancesPolicy.InstancesDistribution == nil {
+			// Set AWS Defaults if we have mixed instances but no distribution policy set to avoid unnecessary template churn
```
dlipovetsky, line 168: In the PR description, you mention "unbounded" churn. I think what you mean is that, if `InstancesDistribution` is undefined, then creating the launch configuration template fails.

I see that `InstancesDistribution` is not marked optional in the API types.

I think we only default optional fields.

dlipovetsky:
Thanks for working on this, @cnmcavoy!

A small nit: There are 3 separate improvements. I think it would have helped reviewers  if they were each at least a separate commit, but perhaps even a separate PR, because they are really distinct pieces of work.

exp/api/v1beta2/awsmachinepool_webhook.go
cnmcavoy: > In the PR description, you mention "unbounded" churn. I think what you mean is that, if InstancesDistribution is undefined, then creating the launch configuration template fails.

The behavior I observed was that when InstanceDistribution was nil, the Launch Template was created successfully, but a diff was always detected. So every reconciliation, a new version was created. The launch template was successfully "updated" each time, there was no error in the creation or update processes. I had a launch template on version 7000+ until I stopped the controller, over the course of a few hours.

cnmcavoy:
For additional context, the reason these all ended up in a single PR is that this was already a split-off from my work in https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4184. I am open to split this further up, if we feel it's necessary for a proper review, but the tradeoff is it slows down getting this fixed.

pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -126,34 +126,17 @@ func (s *Service) ASGIfExists(name *string) (*expinfrav1.AutoScalingGroup, error
 	case err != nil:
 		record.Eventf(s.scope.InfraCluster(), "FailedDescribeAutoScalingGroups", "failed to describe ASG %q: %v", *name, err)
 		return nil, errors.Wrapf(err, "failed to describe AutoScaling Group: %q", *name)
+	case len(out.AutoScalingGroups) == 0:
```
dlipovetsky, line 129: We call this code from 2 places:
1. https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/6ff9319973b929cb4a2ffceca37bbb748e048c44/exp/controllers/awsmachinepool_controller.go#L342. If we do not find an ASG, we record an Event here: https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/6ff9319973b929cb4a2ffceca37bbb748e048c44/exp/controllers/awsmachinepool_controller.go#L349.
2. https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/6ff9319973b929cb4a2ffceca37bbb748e048c44/exp/controllers/awsmachinepool_controller.go#L263. if we do not find an ASG, we do not record any Event. I suggest recording an Event here. WDYT?



(As an aside, using the `FailedDescribeAutoScalingGroups` event seems incorrect. We successfully described the autoscaling groups. We just did not find a group matching our name.)

exp/api/v1beta2/awsmachinepool_webhook.go
dlipovetsky: Thanks for clarifying!

I took a closer look at the issue description, and the code that compares the actual and desired MixedInstancesPolicy: https://github.com/kubernetes-sigs/cluster-api-provider-aws//blob/6ff9319973b929cb4a2ffceca37bbb748e048c44/exp/controllers/awsmachinepool_controller.go#L519.

I don't understand how you came up with the default values. Are they the AWS defaults?

cnmcavoy: > I don't understand how you came up with the default values. Are they the AWS defaults?

Yes, I believe so, I got these by examining the response AWS was supplying to CAPA.

> How will this change affect an AutoScalingGroup created by an earlier CAPA version that did not have these defaults?

My understanding is that it won't affect them, because existing resources won't refire the default webhook logic, it only fires when a new resource is created. We would have to move the defaults into the reconcile loop to set them for existing resources, which I am not sure is necessary.

dlipovetsky: These are the AWS Console defaults for

- 100% on-demand instances
image

- 100% spot instances
image

pkg/cloud/services/autoscaling/autoscalinggroup.go
cnmcavoy: Good catch, I'll take a closer look at this.

dlipovetsky: Thank you!

exp/api/v1beta2/awsmachinepool_webhook.go
cnmcavoy: > The webhook is called for both create and update. Please see the kubebuilder metadata at the top of the file, namely:

I see. It definitely took me a bit to fully grasp the kubebuilder metadata jargon.

> The default strategy for 100% on-demand is "lowest price," not the default you chose, "prioritized."

I see, checking out the codebase, CAPA only supports prioritized as a strategy for `OnDemandAllocationStrategy`:
```
var (
	// OnDemandAllocationStrategyPrioritized uses the order of instance type overrides
	// for the LaunchTemplate to define the launch priority of each instance type.
	OnDemandAllocationStrategyPrioritized = OnDemandAllocationStrategy("prioritized")
)
```

> The default strategy for 100% spot is "price capacity optimized," a strategy that is not even represented in our types!

:open_mouth: 

Do we need to expand this to include the AWS default strategies for both of these enums? The intent of my change wasn't to add new features... But that explains my odd choices in this PR. I'm already starting to forget why I made some of these decisions. :sweat_smile: 

> But these defaults will also apply to existing AWSMachinePool objects, and if there's a difference in the configuration to the one in the actual Launch Template, the Launch Template will be replaced. That might surprise users. WDYT?

To me this suggests we shouldn't introduce any new strategy types in this PR, since users can not select them today, so it would be surprising to me if a bugfix introduced a new strategy _and started defaulting to this new strategy._ I'd expect CAPA defaults to match whatever strategies are supported at the present. 

However, this is a very weak opinion, I can see the other side easily as well - AWS always supported these, the real surprise is that CAPA didn't, sticking to AWS defaults makes CAPA more self-documenting and easier to learn, etc.

Which outcome do we prefer? I don't have much stake since Indeed only uses the strategies CAPA has exposed presently.

pkg/cloud/services/autoscaling/autoscalinggroup.go
cnmcavoy: I switched the `FailedDescribeAutoScalingGroups` to a `corev1.EventTypeNormal` and added an events to the `reconcileNormal` loop. There are two places added, one if the asg does not exist in `reconcileNormal`, and a second event emitted if the instance refresh is enabled as well. Not sure if we want to distinguish them in some way.

exp/api/v1beta2/awsmachinepool_webhook.go
dlipovetsky: > To me this suggests we shouldn't introduce any new strategy types in this PR, since users can not select them today, so it would be surprising to me if a bugfix introduced a new strategy and started defaulting to this new strategy.

I agree with you 100%.

cnmcavoy: In that case, are we comfortable leaving these as the defaults? I think the risk for "updating" existing resources is small, because of the two if-guards for the defaulting here. If both conditions are true, and defaults are unset, the resource is in a broken state on the current release. So defaulting it would fix the resource.

pkg/cloud/services/autoscaling/autoscalinggroup.go
dlipovetsky: Thanks!

What do you think if, as part of this PR, we perform a small clean-up: For consistency, let's replace the string literal `"NoASGFound"` with `expinfrav1.ASGNotFoundReason` everywhere.

exp/api/v1beta2/awsmachinepool_webhook.go
cnmcavoy: > One option is to set defaults in the controller, not the webhook. That is, if MixedInstancesPolicy is undefined, set the values using the configuration returned by AWS, and update the AWSMachinePool resource.

This approach sounds fine to me, but it will require adding the missing AWS strategies so they can map correctly to CAPA values.

dlipovetsky: > This approach sounds fine to me, but it will require adding the missing AWS strategies so they can map correctly to CAPA values.

Thanks for pointing that out (again). I keep forgetting that.

Let's suppose we want to add these missing strategies.

Is that a breaking change? I _think_ it is, because an AWSMachinePool resource with a new strategy would be incompatible with a CRD that doesn't support the strategy.

https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api_changes.md#on-compatibility

dlipovetsky: > If both conditions are true, and defaults are unset, the resource is in a broken state on the current release. So defaulting it would fix the resource.

That's true? You mean, in the current release, AWSMachinePool with `mixedInstancesPolicy` set, but `mixedInstancesPolicy.InstancesDistribution` not set is broken?

I have to agree: If such an AWSMachinePool does not result in a valid Launch Template in the current release,, then we do not need to worry about defaults causing an existing Launch Template to be replaced.

dlipovetsky: What about this option:

If the `mixedInstancesPolicy` is not nil, but `mixedInstancesPolicy.InstancesDistribution` is nil, then we do nothing (same as prior to this PR) in the webhook.

Instead, we change the controller so that it skips the comparison on https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4194/files#diff-9c6c4cb74d0076521d841bb705940e29b5b7b6f81a7c7af31f8586616beec425R551.

cnmcavoy: > That's true? You mean, in the current release, AWSMachinePool with mixedInstancesPolicy set, but mixedInstancesPolicy.InstancesDistribution not set is broken?

Sorta. If you have instance refreshes enabled, it's definitely broken, you will be cycling ec2 nodes endlessly. If you have instance refreshes off, you have version churn, but I believe the machine pool will deploy the ec2 nodes correctly because the asg will be pinned to the first version, but that makes it immutable.

We discovered this because we are testing Machine Pools with mixed instances of m6a and m6i's as a secondary type.

> Instead, we change the controller so that it skips the comparison on https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4194/files#diff-9c6c4cb74d0076521d841bb705940e29b5b7b6f81a7c7af31f8586616beec425R551.

Okay, yeah I like that approach. I think that will work and be fully compatible with existing users.

pkg/cloud/services/autoscaling/autoscalinggroup.go
cnmcavoy: Should all be addressed now.

cnmcavoy:
anything left that I missed on this PR? @dlipovetsky

dlipovetsky:
I didn't realize you had addressed the default issue as well. I'll review again.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -496,29 +519,56 @@ func (r *AWSMachinePoolReconciler) findASG(machinePoolScope *scope.MachinePoolSc
...
+			}
+			if !cmp.Equal(mixedInstancesPolicy.Overrides, existingASG.MixedInstancesPolicy.Overrides) {
+				machinePoolScope.Info("MixedInstancesPolicy.Overrides diff detected", "incoming", spew.Sprintf("%#v", mixedInstancesPolicy.Overrides), "existing", spew.Sprintf("%#v", existingASG.MixedInstancesPolicy.Overrides))
+				return true
+			}
+		} else {
+			machinePoolScope.Info("MixedInstancesPolicy diff detected", "incoming", spew.Sprintf("%#v", mixedInstancesPolicy), "existing", spew.Sprintf("%#v", existingASG.MixedInstancesPolicy))
+			return true
+		}
 	}
```
dlipovetsky, line 572: I had trouble understanding this.

I think what we want is to compare desired and actual MixedInstancesPolicy, but ignore differences in InstancesDistribution, if the desired InstancesDistribution is nil.

I took a different approach here: https://github.com/dlipovetsky/cluster-api-provider-aws/commit/474881358c32153af3ac8cce710d5e449204c784

What do you think?

Skarlso:
@Skarlso take a look at this

dlipovetsky:
I still need to review the reconcile loop changes. I'll get to that today.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -497,29 +520,50 @@ func (r *AWSMachinePoolReconciler) findASG(machinePoolScope *scope.MachinePoolSc
...
+		// InstancesDistribution is optional, and the default values come from AWS, so
+		// they are not set by the AWSMachinePool defaulting webhook. If InstancesDistribution is
+		// not set, we use the AWS values for the purpose of comparison.
+		if mixedInstancesPolicy != nil && mixedInstancesPolicy.InstancesDistribution == nil {
+			mixedInstancesPolicy = machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy.DeepCopy()
+			mixedInstancesPolicy.InstancesDistribution = existingASG.MixedInstancesPolicy.InstancesDistribution
+		}
+
+		if !cmp.Equal(mixedInstancesPolicy, existingASG.MixedInstancesPolicy) {
+			machinePoolScope.Info("MixedInstancesPolicy diff detected", "incoming", spew.Sprintf("%#v", mixedInstancesPolicy), "existing", spew.Sprintf("%#v", existingASG.MixedInstancesPolicy))
```
Skarlso, line 564: We never use spew in the logs. I'm pretty sure this would look super ugly. :D Can you print individual values instead or somehow make a prettier output?

cnmcavoy: Not without expanding the footprint of the code there significantly. `MixedInstancesPolicy` is optional (could be nil on either object) and has further optional nil-able inner struct fields, and golang lacks a safe navigator operator :(

The output is the struct object. `%#v` shows the struct + field names, so it's human readable.

dlipovetsky: I think our goal is to help the user understand the diff, if one occurs. For representing the diff, I think https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff is very readable. Had you already considered it?

For example, this output

```console
2023/05/16 14:58:17 (-existing, +actual)   &v1beta2.MixedInstancesPolicy{
  	InstancesDistribution: &v1beta2.InstancesDistribution{
- 		OnDemandAllocationStrategy:          "",
+ 		OnDemandAllocationStrategy:          "prioritized",
- 		SpotAllocationStrategy:              "lowest-price",
+ 		SpotAllocationStrategy:              "",
  		OnDemandBaseCapacity:                nil,
  		OnDemandPercentageAboveBaseCapacity: nil,
  	},
  	Overrides: nil,
  }
````

is the result of the example

```go
package main

import (
	"log"

	"github.com/google/go-cmp/cmp"

	expinfrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/exp/api/v1beta2"
)

func main() {
	existing := &expinfrav1.MixedInstancesPolicy{
		InstancesDistribution: &expinfrav1.InstancesDistribution{
			SpotAllocationStrategy: expinfrav1.SpotAllocationStrategyLowestPrice,
		},
		Overrides: nil,
	}

	actual := &expinfrav1.MixedInstancesPolicy{
		InstancesDistribution: &expinfrav1.InstancesDistribution{
			OnDemandAllocationStrategy: expinfrav1.OnDemandAllocationStrategyPrioritized,
		},
		Overrides: nil,
	}
	diff := cmp.Diff(existing, actual)
	log.Println("(-existing, +actual)", diff)
}
```

However, the output is readable because it's multiline, and our logs are structured, so I think the newlines will be escaped and everything fit into a single line, which will be very hard to read.

Have you considered logging that the diff occurred, but writing the diff itself to an Event?

cnmcavoy: > I think our goal is to help the user understand the diff, if one occurs. For representing the diff, I think https://pkg.go.dev/github.com/google/go-cmp/cmp#Diff is very readable. Had you already considered it?

I'm not very familiar with that library, so no. But that functionality looks like it could produce the output we want. 

@dlipovetsky what if we use inspiration from your previous suggestion about how to refactor the  `mixedInstancesPolicy` to take a different approach, using `cmp.Diff` to detect the differences in the ASG, and then use that output for the event?


```
// detectASGDrift compares incoming AWSMachinePool and compares against existing ASG.
func detectASGDrift(machinePoolScope *scope.MachinePoolScope, existingASG *expinfrav1.AutoScalingGroup) string {
	detectedMachinePoolSpec := machinePoolScope.MachinePool.Spec.DeepCopy()

	if !scope.ReplicasExternallyManaged(machinePoolScope.MachinePool) {
		detectedMachinePoolSpec.Replicas = existingASG.DesiredCapacity
	}
	if diff := cmp.Diff(machinePoolScope.MachinePool.Spec, *detectedMachinePoolSpec); diff != "" {
		return diff
	}

	detectedAWSMachinePoolSpec := machinePoolScope.AWSMachinePool.Spec.DeepCopy()
	detectedAWSMachinePoolSpec.MaxSize = existingASG.MaxSize
	detectedAWSMachinePoolSpec.MinSize = existingASG.MinSize
	detectedAWSMachinePoolSpec.CapacityRebalance = existingASG.CapacityRebalance
	{
		mixedInstancesPolicy := machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy
		// InstancesDistribution is optional, and the default values come from AWS, so
		// they are not set by the AWSMachinePool defaulting webhook. If InstancesDistribution is
		// not set, we use the AWS values for the purpose of comparison.
		if mixedInstancesPolicy != nil && mixedInstancesPolicy.InstancesDistribution == nil {
			mixedInstancesPolicy = machinePoolScope.AWSMachinePool.Spec.MixedInstancesPolicy.DeepCopy()
			mixedInstancesPolicy.InstancesDistribution = existingASG.MixedInstancesPolicy.InstancesDistribution
		}

		if !cmp.Equal(mixedInstancesPolicy, existingASG.MixedInstancesPolicy) {
			detectedAWSMachinePoolSpec.MixedInstancesPolicy = existingASG.MixedInstancesPolicy
		}
	}

	return cmp.Diff(machinePoolScope.AWSMachinePool.Spec, *detectedAWSMachinePoolSpec)
}
```

And then in `updatePool` use that:

```
	asgDiff := detectASGDrift(machinePoolScope, existingASG)
	if asgDiff != "" {
		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, "ASGDiffDetected", asgDiff)
	}
	if asgDiff != "" || subnetChanges {
		machinePoolScope.Info("updating AutoScalingGroup")
...
```

If prefer that approach, I can update this PR to do that and run some tests.

cnmcavoy:
Any final concerns on this PR?

exp/controllers/awsmachinepool_controller.go
dlipovetsky: I like it!

exp/controllers/awsmachinepool_controller.go
```diff
@@ -404,9 +426,16 @@ func (r *AWSMachinePoolReconciler) updatePool(machinePoolScope *scope.MachinePoo
 		"subnets of machinePoolScope", subnetIDs,
 		"subnets of existing asg", existingASG.Subnets)
 	less := func(a, b string) bool { return a < b }
-	subnetChanges := cmp.Diff(subnetIDs, existingASG.Subnets, cmpopts.SortSlices(less)) != ""
+	subnetChanges := cmp.Diff(subnetIDs, existingASG.Subnets, cmpopts.SortSlices(less))
```
dlipovetsky, line 429: nit: Let's please use a consistent suffix in identifiers that represent "diffs."

I humbly suggest `subnetsDiff` (note the plural) and `asgDiff`.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -493,36 +522,37 @@ func (r *AWSMachinePoolReconciler) findASG(machinePoolScope *scope.MachinePoolSc
 	return asg, nil
 }
 
-// asgNeedsUpdates compares incoming AWSMachinePool and compares against existing ASG.
-func asgNeedsUpdates(machinePoolScope *scope.MachinePoolScope, existingASG *expinfrav1.AutoScalingGroup) bool {
+// detectASGDrift compares incoming AWSMachinePool and compares against existing ASG.
+func detectASGDrift(machinePoolScope *scope.MachinePoolScope, existingASG *expinfrav1.AutoScalingGroup) string {
```
dlipovetsky, line 526: nit: Let's please name the function after what it does, not the purpose for which it is used. For instance, `diffASG`.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -404,9 +426,16 @@ func (r *AWSMachinePoolReconciler) updatePool(machinePoolScope *scope.MachinePoo
...
-	subnetChanges := cmp.Diff(subnetIDs, existingASG.Subnets, cmpopts.SortSlices(less)) != ""
+	subnetChanges := cmp.Diff(subnetIDs, existingASG.Subnets, cmpopts.SortSlices(less))
+	if subnetChanges != "" {
+		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, "ASGDiffDetected", subnetChanges)
+	}
 
-	if asgNeedsUpdates(machinePoolScope, existingASG) || subnetChanges {
+	asgDiff := detectASGDrift(machinePoolScope, existingASG)
+	if asgDiff != "" {
+		r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeNormal, "ASGDiffDetected", asgDiff)
```
dlipovetsky, line 436: Having thought more about it, I'm not sure an Event is what we want. Events have a 1 hour TTL by default.

Is our goal to record this information _always_, or only if the user is asking for more verbose logs?

If it's the latter, then we could use `machinePoolScope.Debug`, as we do in other places.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -227,13 +228,58 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
...
+	}
+
 	canUpdateLaunchTemplate := func() (bool, error) {
 		// If there is a change: before changing the template, check if there exist an ongoing instance refresh,
 		// because only 1 instance refresh can be "InProgress". If template is updated when refresh cannot be started,
 		// that change will not trigger a refresh. Do not start an instance refresh if only userdata changed.
 		return asgsvc.CanStartASGInstanceRefresh(machinePoolScope)
 	}
+	if asg == nil {
+		canUpdateLaunchTemplate = func() (bool, error) {
```
Skarlso, line 249: You can pull this one up into the previous `asg == nil` check.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -227,13 +228,58 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
...
+			return true, nil
+		}
+	}
+
 	runPostLaunchTemplateUpdateOperation := func() error {
+		launchTemplateID := machinePoolScope.GetLaunchTemplateIDStatus()
```
Skarlso, line 255: Shouldn't all this happen after the skipping instance refresh check? If instance refresh is completely disabled, none of these things should happen shouldn't they? Like, updating the tags.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -260,20 +306,13 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
...
 	conditions.MarkTrue(machinePoolScope.AWSMachinePool, expinfrav1.LaunchTemplateReadyCondition)
 
-	// Find existing ASG
-	asg, err := r.findASG(machinePoolScope, asgsvc)
-	if err != nil {
-		conditions.MarkUnknown(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, expinfrav1.ASGNotFoundReason, err.Error())
-		return ctrl.Result{}, err
-	}
-
 	if asg == nil {
```
Skarlso, line 309: so I'm a bit confused here. We reconcile the launch template which does check for an ASG, but we only do the if asg doesn't exist create and requeue AFTER we reconciled the launch template which was looking for an ASG. I mean it's obvious that the ASG won't be there, so why not, if the ASG doesn't exists, create it, then requeue and proceed. I'm not sure I get the order of things in here. :D Can you educate me please? :)

exp/controllers/awsmachinepool_controller.go
Skarlso: Also, Kubernetes drops events if the queue is full. So events should be something that is okay to be missed.

exp/controllers/awsmachinepool_controller.go
Skarlso: Ah, never mind. Of course, the ASG is using the LaunchTemplate. So we assure that it's up to date. Right?

cnmcavoy: We must reconcile tags after updating the launch template, which is why this logic was moved into the beginning of `runPostLaunchTemplateUpdateOperation` - because we need to update our record of the tags on the launch template regardless of whether or not we are doing a refresh. The tags are a property of launch templates, not asgs.

I didn't write this logic, only relocated it from elsewhere in the codepath. Previously we only reconciled tags if the asg could be created, which is too late, because an error could have occurred.

exp/controllers/awsmachinepool_controller.go
cnmcavoy: Done.

exp/controllers/awsmachinepool_controller.go
cnmcavoy: sgtm

exp/controllers/awsmachinepool_controller.go
cnmcavoy: Okay, I don't have strong opinions here, so I swapped back to `machinePoolScope.Debug`

exp/controllers/awsmachinepool_controller.go
cnmcavoy: Done.

exp/controllers/awsmachinepool_controller.go
cnmcavoy: The ASG creation depends on the launch template creation, and the ASG instance refresh depends on the launch template being modified. I think these two dependencies are coupled in the implementation and causing confusion.

My understanding is that we can't move the ASG creation before `ec2Svc.ReconcileLaunchTemplate` is invoked, because we need that to succeed to ensure a launch template exists to use for the ASG.

However, that is also the function that detects if we modified the launch template and runs the post-update callback.

Maybe we should open a separate issue about clarifying these codepaths? I'm hesitant to add more to this PR or else it will never get merged...

Ankitasw:
/lgtm

There were few comments by @dlipovetsky, so I would defer to him for final approval 😅

dlipovetsky:
I haven't had time to look recently, but will do so this week.

cnmcavoy:
Split up into 3 PRs for ease of review.

PR #4250, opened 2023-05-08
Set httpPutResponseHopLimit to 2 when creating instances

### LINKED ISSUE #4247: Set httpPutResponseHopLimit to 2 in instanceMetadataOptions as default when creating instance
/kind feature

Regarding https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4037 support, I would propose to set the default HTTPPutResponseHopLimit to 2 in container environment. 

When customers is using instance profile role instead of using base64 aws credentials (very typical usage in production env), capa container needs 2 hops to retrieve aws credentials from metadata service. If default hop limit is 1,  capa fails to get credentials and fail at the first with:
```
E0430 03:18:00.552022       1 controller.go:329] "Reconciler error" err=<
	.spec.vpc.id is set but VPC resource is missing in AWS; failed to describe VPC resources. (might be in creation process): failed to query ec2 for VPCs: NoCredentialProviders: no valid providers in chain. Deprecated.
		For verbose messaging see aws.Config.CredentialsChainVerboseErrors
 > controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="tkg-system/tkg-mgmt-aws-pvbgc" namespace="tkg-system" name="tkg-mgmt-aws-pvbgc" reconcileID=8e5f1348-7f16-4441-9d18-a4b141eb973b
```

If we set HTTPPutResponseHopLimit to 2  as default, it will avoid capa failure and other applications that needs to access AWS. Otherwise we need customers to set the awsmachine template explicitly:
```
  template:
    spec:
      instanceMetadataOptions:
        httpPutResponseHopLimit: 2
```

They are very likely to forget or not aware of this knowledge and get a failed env.

Another benefit is customers don't need to change awsmachinetemplate very often due to the HopLimit issue in production env , which as we known, is immutable and it is a burden to update to a new machinetemplate.

I also see HTTPPutResponseHopLimit to 2 is recommended in container environment:
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instancedata-data-retrieval.html#imds-considerations

> To avoid problems with instance metadata retrieval, consider the following:
> The AWS SDKs use IMDSv2 calls by default. If the IMDSv2 call receives no response, the SDK retries the call and, if still unsuccessful, uses IMDSv1. This can result in a delay. In a container environment, if the hop limit is 1, the IMDSv2 response does not return because going to the container is considered an additional network hop. To avoid the process of falling back to IMDSv1 and the resultant delay, in a container environment we recommend that you set the hop limit to 2

https://aws.amazon.com/about-aws/whats-new/2020/08/amazon-eks-supports-ec2-instance-metadata-service-v2/
> Now, newly launched and any updated EKS managed node groups will be configured with a metadata token response hop limit set to 2. For self-managed nodes, CloudFormation templates and eksctl have been updated to launch nodes by default with a hop limit of 2.

**Environment:**

wyike:
**What type of PR is this?**
/kind feature

Fixes #4247

Set default httpPutResponseHopLimit as 2 when creating instances
Keep crd default value as 1, in case that customers want to set it as 1 for some machine deployment

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Set default httpPutResponseHopLimit to 2 in IMDSv2 enablement when creating instances
```

### DIFF
--- a/api/v1beta2/awsmachinetemplate_webhook_test.go
+++ b/api/v1beta2/awsmachinetemplate_webhook_test.go
@@ -127,7 +127,7 @@ func TestAWSMachineTemplateValidateUpdate(t *testing.T) {
 							InstanceType: "test",
 							InstanceMetadataOptions: &InstanceMetadataOptions{
 								HTTPEndpoint:            InstanceMetadataEndpointStateEnabled,
-								HTTPPutResponseHopLimit: 1,
+								HTTPPutResponseHopLimit: 2,
 								HTTPTokens:              HTTPTokensStateRequired,
 								InstanceMetadataTags:    InstanceMetadataEndpointStateDisabled,
 							},
--- a/api/v1beta2/types.go
+++ b/api/v1beta2/types.go
@@ -285,7 +285,7 @@ func (obj *InstanceMetadataOptions) SetDefaults() {
 		obj.HTTPEndpoint = InstanceMetadataEndpointStateEnabled
 	}
 	if obj.HTTPPutResponseHopLimit == 0 {
-		obj.HTTPPutResponseHopLimit = 1
+		obj.HTTPPutResponseHopLimit = 2 // Defaults to 2 in container environment
 	}
 	if obj.HTTPTokens == "" {
 		obj.HTTPTokens = HTTPTokensStateRequired // Defaults to IMDSv2
--- a/controllers/awsmachine_controller_unit_test.go
+++ b/controllers/awsmachine_controller_unit_test.go
@@ -2551,7 +2551,7 @@ func TestAWSMachineReconcilerReconcileDefaultsToLoadBalancerTypeClassic(t *testi
 						},
 						MetadataOptions: &ec2.InstanceMetadataOptionsResponse{
 							HttpEndpoint:            aws.String(string(infrav1.InstanceMetadataEndpointStateEnabled)),
-							HttpPutResponseHopLimit: aws.Int64(1),
+							HttpPutResponseHopLimit: aws.Int64(2),
 							HttpTokens:              aws.String(string(infrav1.HTTPTokensStateRequired)),
 							InstanceMetadataTags:    aws.String(string(infrav1.InstanceMetadataEndpointStateDisabled)),
 						},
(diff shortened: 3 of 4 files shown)


Ankitasw:
@pydctw please add release note anytime a new PR is raised.

test/e2e/data/infrastructure-aws/withoutclusterclass/kustomize_sources/addons/csi/data/aws-ebs-csi-external.yaml
```diff
@@ -4,8 +4,8 @@ metadata:
   name: aws-secret
   namespace: kube-system
 stringData:
-  key_id: ${AWS_ACCESS_KEY_ID}
```
Skarlso, line 7: I remember there was a discussion around this, but it's worth noting why these have been deleted?

Skarlso: Meh. I can't find it. I swear someone explained why that was removed. :D

Ankitasw: I guess the test failures are because of that (except for MD remediation which is already failing)
@wyike could you please revert this change, or maybe find a solution such that other tests passes.

wyike: comment here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4147#discussion_r1185647464

wyike: It should be removed. After it is removed, the csi add on will fallback to workload cluster control plane instance role to get credentials from metadata service.

Skarlso:
The remediation test is sadly expected to fail. Anything else might be of more interest to you.

wyike:
Hi @Skarlso all tests pass except for the remediation test.
Could you help review and approve, thanks

PR #4296, opened 2023-05-28
Revert creating instance with wait blocking

### LINKED ISSUE #4293: Self-hosted management cluster control plane upgrade takes more than 15min to finish one control plane
/kind bug

**Steps:**

1. Use CAPA v2.1.x

2. Create a self-hosted management cluster (capa is managing the management cluster it is located on)

3. Upgrade this self-host cluster control plane (kubernetes version)
Each control plane takes 15min+ to finish the new node creation.
Previously, or in normal simple testing env, we expect ~5min for one node.

4. As a result, the whole cluster upgrade time is very long, up to 1 hour to finish 3 control planes.

5. The upgrade can be successfully finally. Basically it's a performance (big) downgrade, no functionality is impacted.

**This issue only happen on:**
- capa v2.1.x code branch
- upgrade self-host cluster control plane

**This issue doesn't happen on:**
- capa v2.0.2
- If capa is in bootstrap kind cluster, it upgrades other workload clusters, no such issue.
- If upgrading the workers of the self-host management cluster, it is almost good.
- If upgrading the workload cluster of the self-host management cluster, it is also good.

**Environment:**

wyike:
**What type of PR is this?**
/kind bug

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4092 is a very good performance enhancement to do concurrent work in large scale deployment. But in the awsmachine reconcile (creation part)workflow, it lacks proper requeue mechanism to requeue pending or even other necessary status instance for next loop reconciliation.
As a result, a very basic self-host management cluster upgrade scenario is experiencing a great performance downgrade because of no too much object changes in such scenario, and no aws machine requeue with only having to rely on the fixed 10min resync period.

To be safer for capa v2.1.x, I just revert previous behavior on awsmachine creation that has been verified during releases. In future major/minor release, we can add the enhancement (creation part)of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4092 back. In the meanwhile redesign the requeue behavior for awsmachine reconcile(creation part), and do fully testing and validations.

Fixes #4293

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Revert creating an instance with previous waiting until it's running
```

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -17,12 +17,15 @@ limitations under the License.
 package ec2
 
 import (
+	"context"
 	"encoding/base64"
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/aws/request"
 	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/pkg/errors"
 	"k8s.io/utils/pointer"
@@ -31,6 +34,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/awserrors"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/converters"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/filter"
+	awslogs "sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/logs"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/userdata"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/record"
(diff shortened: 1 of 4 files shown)


Skarlso:
@wyike Can you add release notes please? :)

wyike:
> @wyike Can you add release notes please? :)

Hi @Skarlso I added.

Because of the downstream product bump timeline, it's risky to involve new changes (creation requeue support) for capa 2.1.x at this point that might cause new regression issues :( . So I just revert the previous working good code. If we want to improve, we can go through the requeue thing in main in future/next release.  Are you good with it. :)

Skarlso:
Sure. Let's contact the previous author of that pr as well? I'm on the phone but I think it was Marcus?

wyike:
> Sure. Let's contact the previous author of that pr as well? I'm on the phone but I think it was Marcus?
It's Cameron. 

Hi @cnmcavoy, tagging you for your awareness of this PR.  Could you please also take a look.  Thanks.

pkg/cloud/services/ec2/instances.go
```diff
@@ -606,6 +610,19 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 		return nil, errors.Errorf("no instance returned for reservation %v", out.GoString())
 	}
 
+	waitTimeout := 1 * time.Minute
```
YanzhaoLi, line 613: If the ec2 instance were not ready in 1 minute, what will happen?

Skarlso:
Thanks @wyike :)

cnmcavoy:
I'm confused at the justification for this PR, because CAPA doesn't make any guarantees about how fast machines will be reconciled to update their status (besides the mentioned 10 min sync period), so reverting this is purely loss of functionality that lowers the ceiling on the maximum cluster size. 

Setting that aside, an improvement without a revert seems to be straightfoward here. I originally considered setting the `requeueAfter` for one minute duration in #4092 but left it out as it wasn't necessary for my changes to work and therefore hard to justify arbitrary magic numbers. 

The AWSMachine reconcile never sets `requeueAfter` as part of the result, but it could set and track a variable for a one minute (or some other) requeue duration [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmachine_controller.go#L506-L598) and include it as part of the final return value with minimal code changes, and it seems like that would resolve this without also reverting the desired functionality. It's a holiday here currently, but someone can attempt this, otherwise I can look at a demonstrative PR later this week.

wyike:
> The AWSMachine reconcile never sets requeueAfter as part of the result, but it could set and track a variable for a one minute (or some other) requeue duration [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmachine_controller.go#L506-L598) and include it as part of the final return value with minimal code changes

Hi @cnmcavoy you're right, this way can resolve the issue. 
I am just worried any new regressions issues if we don't have enough time to fully validate it and insert requeue throughout this part of code also need to be careful. Because when we bump capa 2.1.x release, we have experienced a lot of regression issues that we cannot bump it successfully. And our release timeline is very near and we don't have another chances to try new code changes. So I just revert that creation part.

Could we do the requeue thing after this PR as a full enhancement?  I also mentioned this future work in the description. If you need, I can open another issue to track it.

wyike:
Close PR.

We'll do the fix rather than the revert.

PR #4304, opened 2023-06-01
Allow customization of ingress rules in control plane LB security group

fiunchinho:
**What type of PR is this?**

/kind feature

We would like to customize the ingress rules that are included in the security group created for the control plane Load Balancer. Currently, capa always adds an ingress rule to allow traffic towards the k8s api from everywhere (`0.0.0.0/0`). But we would like to have more control over this. We would like to only allow traffic coming from our VPNs.

With the changes in this PR, when no additional ingress rules are defined, we add the ingress rule to allow traffic from everywhere. That way we are backwards compatible. But when additional rules are defined, these will be used instead.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Allow customization of ingress rules in control plane Load Balancer security group
```

### DIFF
--- a/api/v1beta2/awscluster_types.go
+++ b/api/v1beta2/awscluster_types.go
@@ -208,10 +208,9 @@ type AWSLoadBalancerSpec struct {
 	// +optional
 	AdditionalSecurityGroups []string `json:"additionalSecurityGroups,omitempty"`
 
-	// AdditionalIngressRules sets the additional ingress rules for the control plane load balancer. If no source security group ids are specified, the
-	// default control plane security group will be used.
+	// IngressRules sets the ingress rules for the control plane load balancer.
 	// +optional
-	AdditionalIngressRules []IngressRule `json:"additionalIngressRules,omitempty"`
+	IngressRules []IngressRule `json:"ingressRules,omitempty"`
 
 	// LoadBalancerType sets the type for a load balancer. The default type is classic.
 	// +kubebuilder:default=classic
--- a/pkg/cloud/services/securitygroup/securitygroups.go
+++ b/pkg/cloud/services/securitygroup/securitygroups.go
@@ -520,18 +520,18 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
 			rules = append(rules, s.defaultSSHIngressRule(s.scope.SecurityGroups()[infrav1.SecurityGroupBastion].ID))
 		}
 		if s.scope.ControlPlaneLoadBalancer() != nil {
-			additionalIngressRules := s.scope.ControlPlaneLoadBalancer().AdditionalIngressRules
-			for i := range additionalIngressRules {
-				if additionalIngressRules[i].SourceSecurityGroupIDs == nil && additionalIngressRules[i].SourceSecurityGroupRoles == nil { // if the rule doesn't have a source security group, use the control plane security group
-					additionalIngressRules[i].SourceSecurityGroupIDs = []string{s.scope.SecurityGroups()[infrav1.SecurityGroupControlPlane].ID}
+			ingressRules := s.scope.ControlPlaneLoadBalancer().IngressRules
+			for i := range ingressRules {
+				if ingressRules[i].SourceSecurityGroupIDs == nil && ingressRules[i].SourceSecurityGroupRoles == nil { // if the rule doesn't have a source security group, use the control plane security group
+					ingressRules[i].SourceSecurityGroupIDs = []string{s.scope.SecurityGroups()[infrav1.SecurityGroupControlPlane].ID}
 					continue
 				}
 
-				for _, sourceSGRole := range additionalIngressRules[i].SourceSecurityGroupRoles {
-					additionalIngressRules[i].SourceSecurityGroupIDs = append(additionalIngressRules[i].SourceSecurityGroupIDs, s.scope.SecurityGroups()[sourceSGRole].ID)
+				for _, sourceSGRole := range ingressRules[i].SourceSecurityGroupRoles {
+					ingressRules[i].SourceSecurityGroupIDs = append(ingressRules[i].SourceSecurityGroupIDs, s.scope.SecurityGroups()[sourceSGRole].ID)
 				}
 			}
-			rules = append(rules, s.scope.ControlPlaneLoadBalancer().AdditionalIngressRules...)
+			rules = append(rules, s.scope.ControlPlaneLoadBalancer().IngressRules...)
 		}
 		return append(cniRules, rules...), nil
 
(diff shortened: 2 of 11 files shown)


fiunchinho:
Adding @richardcase and @alexander-demicev as we talked about this in Slack.

fiunchinho:
> Also, what about the IPv6 rules below?

I've changed the logic so that when no additional rules are defined and we want to use the default rules, we either use IPv6 or IPv4 depending on whether IPv6 is enabled or not.

fiunchinho:
I pushed a new change to rename the field from `additionalIngressRules` to simply `IngressRules`. The name `CustomIngressRules` seem redundant to me, because we have many fields where we use a default when it's not set, and we don't call it `CustomSomething`. So `IngressRules` made sense to me. What do you think?

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -577,23 +577,29 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
...
-				Protocol:       infrav1.SecurityGroupProtocolTCP,
-				FromPort:       int64(s.scope.APIServerPort()),
-				ToPort:         int64(s.scope.APIServerPort()),
-				IPv6CidrBlocks: []string{services.AnyIPv6CidrBlock},
-			})
+		var rules infrav1.IngressRules
+		if s.scope.ControlPlaneLoadBalancer() != nil && len(s.scope.ControlPlaneLoadBalancer().IngressRules) > 0 {
+			rules = s.scope.ControlPlaneLoadBalancer().IngressRules
+		} else {
+			if s.scope.VPC().IsIPv6Enabled() {
```
AverageMarcus, line 584: These could be merged to avoid extra nesting.

api/v1beta2/awscluster_types.go
```diff
@@ -208,10 +208,9 @@ type AWSLoadBalancerSpec struct {
 	// +optional
 	AdditionalSecurityGroups []string `json:"additionalSecurityGroups,omitempty"`
 
-	// AdditionalIngressRules sets the additional ingress rules for the control plane load balancer. If no source security group ids are specified, the
-	// default control plane security group will be used.
+	// IngressRules sets the ingress rules for the control plane load balancer.
```
AverageMarcus, line 211: Do we still want to indicate that if not provided the defaults will be used?

fiunchinho: > If no source security group ids are specified, the default control plane security group will be used.

I don't think so because that is not true anymore.

alexander-demicev (commented):
Looks good to me, there is also a small piece of documentation that needs to be updated https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure.html?highlight=ingress#control-plane-load-balancer

api/v1beta2/awscluster_types.go
AverageMarcus: Isn't this the default though? https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4304/files#diff-59c59dbf183fdde59cf0030bd23388381951860f0002a073a55f330c4733f8ddR583-R602

fiunchinho: I may be miss understanding the original code comment so bear with me for a second :smile: 

The code comment says

> If no source security group ids are specified, the default control plane security group will be used.

But in the current implementation, there is no defaulting for the source security group of the ingress rules. The ingress rules are used "as is", without defaulting the source of the traffic. That's why I believe the comment is not true anymore.

fiunchinho:
> Looks good to me, there is also a small piece of documentation that needs to be updated https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure.html?highlight=ingress#control-plane-load-balancer

Updated.

AverageMarcus:
/lgtm

Just needs approval from a maintainer :) 

/assign @richardcase @Ankitasw @Skarlso

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -799,3 +785,27 @@ func ingressRulesFromSDKType(v *ec2.IpPermission) (res infrav1.IngressRules) {
...
+		return infrav1.IngressRules{
+			{
+				Description:    "Kubernetes API IPv6",
+				Protocol:       infrav1.SecurityGroupProtocolTCP,
+				FromPort:       int64(s.scope.APIServerPort()),
+				ToPort:         int64(s.scope.APIServerPort()),
+				IPv6CidrBlocks: []string{services.AnyIPv6CidrBlock},
+			},
+		}
+	} else {
```
Skarlso, line 800: Don't need the else, you are already returning in the above section. :)

Skarlso: I mean, drop the else, and just return the `IngressRule`.

Skarlso:
Looks okay except for that small refactor. :) Then it should be fine. :)

pkg/cloud/services/securitygroup/securitygroups.go
fiunchinho: Done. Please take a look.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -577,23 +577,9 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
...
-		}
-		if s.scope.VPC().IsIPv6Enabled() {
-			rules = append(rules, infrav1.IngressRule{
-				Description:    "Kubernetes API IPv6",
-				Protocol:       infrav1.SecurityGroupProtocolTCP,
-				FromPort:       int64(s.scope.APIServerPort()),
-				ToPort:         int64(s.scope.APIServerPort()),
-				IPv6CidrBlocks: []string{services.AnyIPv6CidrBlock},
-			})
+		rules := s.getDefaultIngressRulesForControlPlaneLB()
```
Skarlso, line 580: Nice, looks lovely.

Skarlso:
```
E0606 11:34:44.467623       1 controller.go:329] "Reconciler error" err=<
	failed to create one or more IP addresses for NAT gateways: failed to allocate Elastic IP: AddressLimitExceeded: The maximum number of addresses has been reached.
		status code: 400, request id: 162c8357-0949-45bc-b298-c9b3b41b9a17
 > controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="functional-test-ssm-parameter-store-clusterclass-6jt77v/cluster-vetjuj-sc9q9" namespace="functional-test-ssm-parameter-store-clusterclass-6jt77v" name="cluster-vetjuj-sc9q9" reconcileID="5e8ecaea-9c47-492c-86a8-2281666b5bc8"
```

Pfff...

Now we have to find out if we actually created that many, or there is a problem.

fiunchinho:
> ```
> E0606 11:34:44.467623       1 controller.go:329] "Reconciler error" err=<
> 	failed to create one or more IP addresses for NAT gateways: failed to allocate Elastic IP: AddressLimitExceeded: The maximum number of addresses has been reached.
> 		status code: 400, request id: 162c8357-0949-45bc-b298-c9b3b41b9a17
>  > controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="functional-test-ssm-parameter-store-clusterclass-6jt77v/cluster-vetjuj-sc9q9" namespace="functional-test-ssm-parameter-store-clusterclass-6jt77v" name="cluster-vetjuj-sc9q9" reconcileID="5e8ecaea-9c47-492c-86a8-2281666b5bc8"
> ```
> 
> Pfff...
> 
> Now we have to find out if we actually created that many, or there is a problem on the account. I've seen sometimes that when there is an error it kept creating security groups when it eventually reached the limit withing a couple seconds haha.

What should I do now? I think the only thing I can do is to re-trigger and hope for the best :smile:

fiunchinho:
All green now :smile_cat:

PR #4321, opened 2023-06-07
chore: Bump CAPI to v1.4.3

Ankitasw:
**What type of PR is this?**
/kind support

This PR bumps CAPI to v1.4.3.

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Bump CAPI to v1.4.3
```


Ankitasw:
PR is up for review.

bootstrap/eks/controllers/eksconfig_controller.go
```diff
@@ -212,26 +212,26 @@ func (r *EKSConfigReconciler) joinWorker(ctx context.Context, cluster *clusterv1
...
 			clusterv1.ConditionSeverityInfo, "")
-		return ctrl.Result{}, nil
+		return nil
 	}
 
 	if !conditions.IsTrue(cluster, clusterv1.ControlPlaneInitializedCondition) {
 		log.Info("Control Plane has not yet been initialized")
 		conditions.MarkFalse(config, eksbootstrapv1.DataSecretAvailableCondition, eksbootstrapv1.WaitingForControlPlaneInitializationReason, clusterv1.ConditionSeverityInfo, "")
-		return ctrl.Result{}, nil
+		return nil
```
richardcase, line 221: Would we not want to requeue in this scenario? (i know we didn't originally).

exp/controllers/awsmachinepool_controller.go
```diff
@@ -200,28 +200,28 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
...
-			return ctrl.Result{}, err
+			return err
 		}
 	}
 
 	if !machinePoolScope.Cluster.Status.InfrastructureReady {
 		machinePoolScope.Info("Cluster infrastructure is not ready yet")
 		conditions.MarkFalse(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, infrav1.WaitingForClusterInfrastructureReason, clusterv1.ConditionSeverityInfo, "")
-		return ctrl.Result{}, nil
+		return nil
```
richardcase, line 217: We should be requeuing in this scenario.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -200,28 +200,28 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
...
 	}
 
 	// Make sure bootstrap data is available and populated
 	if machinePoolScope.MachinePool.Spec.Template.Spec.Bootstrap.DataSecretName == nil {
 		machinePoolScope.Info("Bootstrap data secret reference is not yet available")
 		conditions.MarkFalse(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, infrav1.WaitingForBootstrapDataReason, clusterv1.ConditionSeverityInfo, "")
-		return ctrl.Result{}, nil
+		return nil
```
richardcase, line 224: And we should requeue/requeueafter here

richardcase (commented):
@Ankitasw - there are a lot of changes in here that aren't specifically related to bumping CAPI to v1.4.3.

I'd say that we shouldn't change the return values of the reconcoleNormal/reconcileDelete functions to remove the ctrl.Result for 2 reasons:
- its the pattern used widely across CAPI
- There are some scenarios where we should be requeueing (and we're currently not).

Ankitasw:
@richardcase I think they dealt in CAPI the same way, isn't it?
You can see changes here: https://github.com/kubernetes-sigs/cluster-api/pull/8796/files#diff-47cf5969a7a97b70232c5d97b882424745bd1ed511b29aa56f55194688ab833c

I understand that not all changes are related to CAPI bump, shall I revert golangci bump changes and raise another PR?

bootstrap/eks/controllers/eksconfig_controller.go
Ankitasw: we can move this discussion to separate PR

Ankitasw:
PR is ready for review.

richardcase:
Just seen that you reverted some of the changes, sorry @Ankitasw i should've pinged you on slack to discuss.

As it stands now and with the e2e passing:

/approve
/lgtm

Ankitasw:
> Looking through the logic in the reconcileDelete/Normal, I think there are situations where we should be requeuing, and we aren't currently, so if we removed the result, then we would have to add them back in if we wanted to change these to requeue. But agree as it stands with no requeues then having the extra return value is pointless.

Addressed this in #4326

Issue #4342, opened 2023-06-15
Unnecessary wider permissions required for consistent API intent


enxebre:
/kind bug

**What steps did you take and what happened:**
I'm expressing intent in the API similar to
```
spec:
  additionalSecurityGroups:
  - id: sg-0e42acdc7c8b1f9a1
  additionalTags:
    expirationDate: 2023-06-14T14:51+00:00
    kubernetes.io/cluster/example-vmtl4: owned
  ami:
    id: ami-01735c4f217bf0ae4
  cloudInit:
    insecureSkipSecretsManager: true
    secureSecretsBackend: secrets-manager
  iamInstanceProfile: example-vmtl4-worker
  instanceType: m5.large
  rootVolume:
    size: 64
    type: gp3
  subnet:
    id: subnet-0d10940bdfb2aeb29
  uncompressedUserData: true
```

No InstanceMetadataOptions.

Now when updating to latest version of the controller I got a requirement for wider permissions even though I'm not exercising them.

```
 - lastTransitionTime: "2023-06-14T11:25:42Z"
    message: "failed to create AWSMachine instance: UnauthorizedOperation: You are
      not authorized to perform this operation.\n\tstatus code: 403, request id: 8b50824a-16ab-40cf-b737-47cb80ad34e4"
    reason: InstanceProvisionFailed
    severity: Error
    status: "False"
    type: InstanceReady
```

The field is optional and my API intent is the same than it was before, and so I consider the controller requiring any additional permissions to operate a disruptive breaking change

I suspect this is related to the instance metadata changes github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4037
and the fact we are setting defaults in the backend `infrav1.SetDefaults_AWSMachineSpec(&awsMachine.Spec)`  and so bypassing API intent.
I'd expect `SetDefaults_AWSMachineSpec` to honour nil API input as `getInstanceMetadataOptionsRequest` does. Then
seems to me that the in [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4037/files#r1129064749) should be fixed by accounting for nil derefs in `ensureInstanceMetadataOptions`

**What did you expect to happen:**
New version of controller do not require wider permissions for same API intent.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
[...]

PR #4347, opened 2023-06-19
Use default arch x86_64 for AMI lookup if `ec2:DescribeInstanceTypes` permission is missing

muraee:
**What type of PR is this?**

/kind bug

when upgrading to v2.1.2 we encountered the following error:
```
- lastTransitionTime: "2023-06-14T11:25:42Z"
    message: "failed to create AWSMachine instance: UnauthorizedOperation: You are
      not authorized to perform this operation.\n\tstatus code: 403, request id: 8b50824a-16ab-40cf-b737-47cb80ad34e4"
    reason: InstanceProvisionFailed
    severity: Error
    status: "False"
    type: InstanceReady
 ```
the message doesn't really provide any hints on the cause. After some investigation we found the issue to be a missing `ec2:DescribeInstanceTypes` permission which was introduced here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4054. However, new permissions should not be required by a new version of the controller.

This PR warns the user about permissions error (i.e. missing `ec2:DescribeInstanceTypes`) and returns the default arch `x86_64` instead of failing. Also provide a better error message for other types of errors.

Fixes #https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4342

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
use default arch x86_64 for AMI lookup if `ec2:DescribeInstanceTypes` permission is missing
```

### DIFF
--- a/pkg/cloud/services/ec2/ami.go
+++ b/pkg/cloud/services/ec2/ami.go
@@ -33,11 +33,15 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/cmd/clusterawsadm/api/bootstrap/v1beta1"
+	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/awserrors"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/record"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/util/system"
 )
 
 const (
+	// DefaultArchitectureTag is the default architecture used when the architcture can't be determined from instance type.
+	DefaultArchitectureTag = Amd64ArchitectureTag
+
 	// Amd64ArchitectureTag is the reference AWS uses for amd64 architecture images.
 	Amd64ArchitectureTag = "x86_64"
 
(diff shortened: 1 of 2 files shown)


enxebre:
I think I'd expect arch to be API user input eventually.

Test complain
```
level=info msg="[runner] linters took 1m13.399499211s with stages: goanalysis_metalinter: 1m13.280064857s"
pkg/cloud/services/ec2/ami.go:33: File is not `gci`-ed with --skip-generated -s standard,default,prefix(sigs.k8s.io/cluster-api) (gci)
	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/awserrors"
pkg/cloud/services/ec2/ami.go:36: File is not `gci`-ed with --skip-generated -s standard,default,prefix(sigs.k8s.io/cluster-api) (gci)
	"sigs.k8s.io/cluster-api-provider-aws/v2/cmd/clusterawsadm/api/bootstrap/v1beta1"
```
lgtm otherwise.

faiq:
This is a great change. Although I think it would be better to let them know to update their permissions in a future release and take this code out.

enxebre:
>This is a great change. Although I think it would be better to let them know to update their permissions in a future release and take this code out.

The change that requires widen up IAM was introduced a while ago and has already shipped. The change was driven to support a concrete use case related to eks https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4054 / https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3121

Now, because of that every user is forced to widen up iam, which is not always acceptable in restricted environments, specially if it's adding zero value for them.

I'd expect CAPA as a project strives to only force users to widen up permissions if they actually need to and add value for their use cases. This is particularly disrupting if the additional IAM is required even though API intent was untouched by a consumer.
I created https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4342 to discuss this cc @richardcase @Skarlso 
This PR is proposing a step in that direction.

pkg/cloud/services/ec2/ami.go
```diff
@@ -114,7 +118,13 @@ func (s *Service) pickArchitectureForInstanceType(instanceType string) (string,
 	}
 	describeInstanceTypeResult, err := s.EC2Client.DescribeInstanceTypes(descInstanceTypeInput)
 	if err != nil {
-		return "", err
+		// if call to DescribeInstanceTypes fails due to permissions error, log a warning and return the default architecture.
+		if awserrors.IsPermissionsError(err) {
+			record.Warnf(s.scope.InfraCluster(), "FailedDescribeInstanceTypes", "insufficient permissions to describe instance types for instance type %q: %v", instanceType, err)
```
richardcase, line 123: It would be nice in the warning to mention that we are falling back to the default architecture of x86_64.

richardcase (commented):
Thanks for this @muraee . One minor comment about expanding the warning message and the following lint error:

```
pkg/cloud/services/ec2/ami.go:36: File is not `gci`-ed with
```

Apart from that this looks good.

muraee:
thanks @richardcase, addressed your comment and fixed the lint error.

PR #4360, opened 2023-06-26
:bug: compare string value of protocol in health check instead of pointer

faiq:
**What type of PR is this?**
/kind bug 

when a user specified healthCheck to be HTTPS or HTTP they would see an error like this in the logs 

```
failed to configure health check for classic load balancer: &{<redacted object>}: ValidationError: HealthCheck HTTPS Target must specify a port followed by a path that begins with a slash. e.g. HTTP:80/ping/this/path 
```

This was because the comparison on line 1441 of this file was comparing the addresses of the ELBHealthCheck, rather than the object. That comparison always failed and so the value from line 1445 was always returned. 

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
fixes an error for HTTPS and HTTP health checks for classic ELBs 
```

### DIFF
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -1438,7 +1438,7 @@ func (s *Service) getHealthCheckTarget() string {
 	protocol := &infrav1.ELBProtocolSSL
 	if controlPlaneELB != nil && controlPlaneELB.HealthCheckProtocol != nil {
 		protocol = controlPlaneELB.HealthCheckProtocol
-		if protocol == &infrav1.ELBProtocolHTTP || protocol == &infrav1.ELBProtocolHTTPS {
+		if protocol.String() == infrav1.ELBProtocolHTTP.String() || protocol.String() == infrav1.ELBProtocolHTTPS.String() {
 			return fmt.Sprintf("%v:%d/readyz", protocol, infrav1.DefaultAPIServerPort)
 		}
 	}
--- a/pkg/cloud/services/elb/loadbalancer_test.go
+++ b/pkg/cloud/services/elb/loadbalancer_test.go
@@ -2317,6 +2317,10 @@ func TestChunkELBs(t *testing.T) {
 }
 
 func TestGetHealthCheckProtocol(t *testing.T) {
+	testHTTP := infrav1.ELBProtocol("HTTP")
+	testHTTPS := infrav1.ELBProtocol("HTTPS")
+	testTCP := infrav1.ELBProtocol("TCP")
+
 	tests := []struct {
 		testName                  string
 		lbSpec                    *infrav1.AWSLoadBalancerSpec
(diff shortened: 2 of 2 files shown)


pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -1438,7 +1438,7 @@ func (s *Service) getHealthCheckTarget() string {
 	protocol := &infrav1.ELBProtocolSSL
 	if controlPlaneELB != nil && controlPlaneELB.HealthCheckProtocol != nil {
 		protocol = controlPlaneELB.HealthCheckProtocol
-		if protocol == &infrav1.ELBProtocolHTTP || protocol == &infrav1.ELBProtocolHTTPS {
+		if *(protocol) == infrav1.ELBProtocolHTTP || *(protocol) == infrav1.ELBProtocolHTTPS {
```
Skarlso, line 1441: But the tests are passing that you added. :D So what is this solving? The protocol does have a `String` method.

Also the `*()` could just be `*` but honestly, I would rather not have this blown up:

```go
if protocol.String() == infrav1.ELBProtocolHTTP.String() || protocol.String() == infrav1.ELBProtocolHTTPS.String() {
```

Instead then? But again, why are the tests working then? :D

enxebre: Test is passing because the is forcing the same variable address. it is being set in 
```
			&infrav1.AWSLoadBalancerSpec{
				HealthCheckProtocol: infrav1.ELBProtocol("HTTP"),
			},
```
So then the addresses match. But is you set something like 

```
	http := infrav1.ELBProtocol("HTTP")
	&infrav1.AWSLoadBalancerSpec{
		HealthCheckProtocol: &http,
	},
```

then test would fail. I'd expect this to be comparing strings not variables addressees so either `*protocol == infrav1.ELBProtocolHTTP` or `protocol.String() == infrav1.ELBProtocolHTTP.String()` makes sense to me.

faiq: Let's use the `.String()` method instead, it looks a lot cleaner to me. Thank you both for the comments :)

faiq: Done! Please take a look at your convenience @Skarlso

enxebre: To prove the value, can we please add a testCase within `TestGetHealthCheckProtocol` where we don't set the API input as the built-in variable as in my example above?

faiq: This test should have captured the error https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0ca6090bfb6f5a61b4a52b2c9cb9eb7c23d86a32/pkg/cloud/services/elb/loadbalancer_test.go#L2332

so i'm very confused.

faiq: I've updated the tests

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -2330,21 +2334,21 @@ func TestGetHealthCheckProtocol(t *testing.T) {
 		{
 			"protocol http",
 			&infrav1.AWSLoadBalancerSpec{
-				HealthCheckProtocol: &infrav1.ELBProtocolHTTP,
```
enxebre, line 2333: can we please leave some test case using the built-in variables? so we cover both

enxebre:
just one more comment https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4360#discussion_r1246918395
then if yo can squash commits lgtm!

pkg/cloud/services/elb/loadbalancer_test.go
faiq: I don't think it makes sense to use the built-in variables. When a user will provide these values via a AWSCluster object, these will always be new variables

richardcase:
@faiq - this also lgtm, can you squash the commits and then we can get this in.

pkg/cloud/services/elb/loadbalancer_test.go
enxebre: that's not necessarily true. As a consumer of the API I can programatically choose to vendor capa and use the built-in variable to define my instance of the awsCluster. Just a nit anyway, lgtm.

PR #4379, opened 2023-07-05
Reconcile AWSMachinePool tags immediately after updating launch templates

cnmcavoy:
**What type of PR is this?**

/kind bug

Moves the tag reconciliation of AWSMachinePools to immediately after a launch template is created or updated. Tags need to be reconciled as early as possible after touching the launch template in AWS. If an error occurs in the reconciliation after touching a launch template, tags must be reconciled before ending reconciliation, or the next loop will falsely detect a difference between AWS and the AWSMachinePool spec.

Fixes #

Split off from #4194

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Reconcile AWSMachinePool tags immediately after updating launch templates
```

### DIFF
--- a/exp/controllers/awsmachinepool_controller.go
+++ b/exp/controllers/awsmachinepool_controller.go
@@ -19,6 +19,7 @@ package controllers
 import (
 	"context"
 	"fmt"
+	"time"
 
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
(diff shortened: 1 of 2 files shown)


exp/controllers/awsmachinepool_controller.go
```diff
@@ -233,7 +234,47 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 		// that change will not trigger a refresh. Do not start an instance refresh if only userdata changed.
 		return asgsvc.CanStartASGInstanceRefresh(machinePoolScope)
 	}
+
+	// Find existing ASG
+	asg, err := r.findASG(machinePoolScope, asgsvc)
+	if err != nil {
+		conditions.MarkUnknown(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, expinfrav1.ASGNotFoundReason, err.Error())
```
enxebre, line 241: This seems a bit confusing to me: We set the condition as unknown because of ASGNotFoundReason reason. Reasons represent contractual semantics and so I'd expect ASGNotFoundReason semantic to represent strictly a not found in aws but here it means something else - any error on the request.
Then yet in l246 we are using it for the not found semantic.

cnmcavoy: Yes, now that you point it out, that is confusing to me too. This is a bit of relocated code, but it's worth fixing. I updated this to `MarkFalse` and removed the event from the `asg == nil` branch, since that is normal on the first reconcile.

dlipovetsky:
> If an error occurs in the reconciliation after touching a launch template, _tags must be reconciled before ending reconciliation_

(emphasis mine) 

Does this PR reconcile tags before ending reconciliation of the launch template? I don't see how.

Ankitasw:
@cnmcavoy would you be able to continue with the PR ?

cnmcavoy:
There's still the possibility of errors handling reconciling AWSMachinePools when the error occurs before the tags are set, but I haven't been able to trigger this issue in a while, so I'll revisit if needed in the future.

PR #4406, opened 2023-07-18
Allow securing api LB, only allowing traffic from required sources

fiunchinho:
**What type of PR is this?**
/kind bug

On this PR https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4304 we introduced the possibility of passing custom ingress rules that would be applied to the api Load Balancer security group, so that users could specify more granular rules (for example, allowing traffic only from company VPNs) instead of allowing traffic from everywhere.
But the api Load Balancer needs to always allow traffic coming from the kubelet, or the cluster won't work.

On this PR, I'm changing the logic so that the custom ingress rules are appended to a set of basic rules that will always be added to make sure the kubelet can talk to the api Load Balancer. This set of basic rules will be different depending on whether the Load Balancer is a `internet-facing` or `internal` one.
- When using a `internet-facing` LB, we allow the public IPs of the Nat Gateways.
- When using an `internal` LB, we allow the workload cluster VPC CIDR.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fix control plane LB ingress rules so that kubelet can access the API
```


faiq:
This is a really good idea and the code looks reasonable to me. I think this change should be called out as a [caveat here](https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure.html?highlight=load#control-plane-load-balancer)

pkg/cloud/services/network/natgateways.go
```diff
@@ -106,12 +106,18 @@ func (s *Service) reconcileNatGateways() error {
 			}
 		}
 		ngws, err := s.createNatGateways(subnetIDs)
+		natGatewaysIPs := []string{}
```
Skarlso, line 109:
```suggestion
		var natGatewaysIPs []string
```

Skarlso: Otherwise, you'll get an empty list but a list nevertheless. It might make things awkward in certain places.

pkg/cloud/services/network/natgateways.go
```diff
@@ -106,12 +106,18 @@ func (s *Service) reconcileNatGateways() error {
...
 
 		for _, ng := range ngws {
 			subnet := s.scope.Subnets().FindByID(*ng.SubnetId)
 			subnet.NatGatewayID = ng.NatGatewayId
+			if len(ng.NatGatewayAddresses) > 0 && ng.NatGatewayAddresses[0].PublicIp != nil {
```
Skarlso, line 114: Hm, is it always the `0th` item?

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -786,7 +785,51 @@ func ingressRulesFromSDKType(v *ec2.IpPermission) (res infrav1.IngressRules) {
 	return res
 }
 
-func (s *Service) getDefaultIngressRulesForControlPlaneLB() infrav1.IngressRules {
+// getIngressRulesToAllowKubeletToAccessTheControlPlaneLB returns ingress rules required in the control plane LB.
+// The control plane LB will be accessed by in-cluster components like the kubelet, that means allowing the NatGateway IPs
+// when using an internet-facing LB, or the VPC CIDR when using an internal LB.
+func (s *Service) getIngressRulesToAllowKubeletToAccessTheControlPlaneLB() infrav1.IngressRules {
+	if s.scope.ControlPlaneLoadBalancer() != nil && s.scope.ControlPlaneLoadBalancer().Scheme != nil && *s.scope.ControlPlaneLoadBalancer().Scheme == infrav1.ELBSchemeInternal {
```
Skarlso, line 792:
```suggestion
	if s.scope.ControlPlaneLoadBalancer() != nil && s.scope.ControlPlaneLoadBalancer().Scheme != nil && *s.scope.ControlPlaneLoadBalancer().Scheme == infrav1.ELBSchemeInternal {
```

So this happens often enough now, honestly that it's just annoying. :D 

If you don't mind, would you be so kind to add this function:

```go

func (e ELBScheme) Equal(other *ELBScheme) bool {
	if other == nil {
		return false
	}

	return e == *other
}
```

in api/v1beta2/network_types.go. Then you can just call it with:

```go
if s.scope.ControlPlaneLoadBalancer() != nil && infrav1.ELBSchemeInternal.Equal(s.scope.ControlPlaneLoadBalancer().Scheme) {
```

What do you think?

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -786,7 +785,51 @@ func ingressRulesFromSDKType(v *ec2.IpPermission) (res infrav1.IngressRules) {
...
+		return infrav1.IngressRules{
+			{
+				Description: "Kubernetes API",
+				Protocol:    infrav1.SecurityGroupProtocolTCP,
+				FromPort:    int64(s.scope.APIServerPort()),
+				ToPort:      int64(s.scope.APIServerPort()),
+				CidrBlocks:  []string{s.scope.VPC().CidrBlock},
```
Skarlso, line 799: This needs to consider IPv6 cidr blocks.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -786,7 +785,51 @@ func ingressRulesFromSDKType(v *ec2.IpPermission) (res infrav1.IngressRules) {
...
+
+	natGatewaysIPs := s.scope.GetNatGatewaysIPs()
+	if len(natGatewaysIPs) > 0 {
+		return infrav1.IngressRules{
+			{
+				Description: "Kubernetes API",
+				Protocol:    infrav1.SecurityGroupProtocolTCP,
+				FromPort:    int64(s.scope.APIServerPort()),
+				ToPort:      int64(s.scope.APIServerPort()),
+				CidrBlocks:  natGatewaysIPs,
```
Skarlso, line 812: This too.

pkg/cloud/services/securitygroup/securitygroups_test.go
```diff
@@ -839,26 +951,12 @@ func TestControlPlaneLoadBalancerIngressRules(t *testing.T) {
 
 	for _, tc := range testCases {
 		t.Run(tc.name, func(t *testing.T) {
-			vpcSpec := infrav1.VPCSpec{}
-			if tc.useIPV6 {
```
Skarlso, line 843: Errr... you removed the settings for ipv6? 🤔

pkg/cloud/services/securitygroup/securitygroups.go
fiunchinho: Yeah, good idea! I'll do that.

pkg/cloud/services/securitygroup/securitygroups.go
fiunchinho: I don't think CAPA uses ipv6 for nat gateway IPs. Are you sure is possible that Nat Gateways use IPv6 here?

pkg/cloud/services/network/natgateways.go
fiunchinho: Done

pkg/cloud/services/securitygroup/securitygroups.go
fiunchinho: Done

pkg/cloud/services/securitygroup/securitygroups.go
Skarlso: Err, I have to check. You might be right that it's not.

pkg/cloud/services/securitygroup/securitygroups_test.go
Skarlso: Ahhha, interesting.

pkg/cloud/services/securitygroup/securitygroups.go
Skarlso: Looks like we only create them with EIPs. So.... _maybe_ this is fine. :D

pkg/cloud/services/securitygroup/securitygroups_test.go
Skarlso: Okay, I see now.

fiunchinho:
> [APPROVALNOTIFIER] This PR is **NOT APPROVED**
> 
> This pull-request has been approved by: **Once this PR has been reviewed and has the lgtm label**, please assign [enxebre](https://github.com/enxebre) for approval. For more information see [the Kubernetes Code Review Process](https://git.k8s.io/community/contributors/guide/owners.md#the-code-review-process).
> 
> The full list of commands accepted by this bot can be found [here](https://go.k8s.io/bot-commands?repo=kubernetes-sigs%2Fcluster-api-provider-aws).
> 
> Needs approval from an approver in each of these files:
> * **[OWNERS](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/OWNERS)**
> 
> Approvers can indicate their approval by writing `/approve` in a comment Approvers can cancel approval by writing `/approve cancel` in a comment

@enxebre could I get an approval here? :)

enxebre:
@fiunchinho I'm not approver here, not sure why the bot tagged me, but change lgtm
/lgtm

api/v1beta2/network_types.go
```diff
@@ -36,6 +36,9 @@ type NetworkStatus struct {
 
 	// APIServerELB is the Kubernetes api server load balancer.
 	APIServerELB LoadBalancer `json:"apiServerElb,omitempty"`
+
+	// NatGatewaysIPs contains the NAT Gateways public IPs.
+	NatGatewaysIPs []string `json:"natGatewaysIPs,omitempty"`
```
AndiDog, line 41:
```suggestion
	// NatGatewaysIPs contains the public IPs of the NAT Gateways
	NatGatewaysIPs []string `json:"natGatewaysIPs,omitempty"`
```

This variable is named plural-plural on purpose and we should make that clear. Using an apostrophe like "the NAT Gateways' public IPs" probably isn't clear enough, so my suggestion rewords it in another way.

It's different from what we have with

```
InstanceIDs
SecurityGroupIDs
SubnetIDs
SourceSecurityGroupIDs
PolicyARNs
```

because this collects the IP for each of the (typically) 3 NAT gateways, each gateway normally having one IP. So it shouldn't be named `NatGatewayIPs`.

Skarlso:
Uh, why did the APIDiff thing start to fail/

fiunchinho:
> Uh, why did the APIDiff thing start to fail?

I don't know :/ The job output is not very helpful.

Skarlso:
@fiunchinho What was your latest commit? It's squashed so I don't know. :) It didn't fail before.

fiunchinho:
I'd say [it's this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4406#discussion_r1302695179).

But what is the job complaining about? what's the change in the api?

Skarlso:
Ah, try running `make clean && make generate` please. See if anything changes.

fiunchinho:
> Ah, try running `make clean && make generate` please. See if anything changes.

Nothing changes.

Skarlso:
Usually that api-diff says that there is a backwards incompatible change between two API versions. I suggest reverting that commit, or try looking in the other packages for that same commit and change it also there.

fiunchinho:
> Usually that api-diff says that there is a backwards incompatible change between two API versions. I suggest reverting that commit, or try looking in the other packages for that same commit and change it also there.

Looking at the output of the job, I don't think there is a backwards incompatible change

```
sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope
  Incompatible changes:
  - NetworkScope.GetNatGatewaysIPs: added
  - NetworkScope.SetNatGatewaysIPs: added
  - SGScope.GetNatGatewaysIPs: added
  - SGScope.SetNatGatewaysIPs: added
  Compatible changes:
  - (*ClusterScope).GetNatGatewaysIPs: added
  - (*ClusterScope).SetNatGatewaysIPs: added
  - (*ManagedControlPlaneScope).GetNatGatewaysIPs: added
  - (*ManagedControlPlaneScope).SetNatGatewaysIPs: added
```

I think it should be ok to merge.

Skarlso:
These are the incompatible changes:
```
  Incompatible changes:
  - NetworkScope.GetNatGatewaysIPs: added
  - NetworkScope.SetNatGatewaysIPs: added
  - SGScope.GetNatGatewaysIPs: added
  - SGScope.SetNatGatewaysIPs: added
  - ```

fiunchinho:
> These are the incompatible changes:
> 
> ```
>   Incompatible changes:
>   - NetworkScope.GetNatGatewaysIPs: added
>   - NetworkScope.SetNatGatewaysIPs: added
>   - SGScope.GetNatGatewaysIPs: added
>   - SGScope.SetNatGatewaysIPs: added
>   - ```
> ```

But that's not the API, isn't it?. Isn't the job supposed to look for incompatible changes in the facing API, like changes inm `AWSClusterSpec` and so on? Also, how is adding functions an incompatible change?

Skarlso:
No. You modified the interface by adding two new functions. Which means that this change is now backwards incompatible since people requiring that interface will have a broken code.

I'm not sure how we usually deal with changes like these. @richardcase @Ankitasw ?

Ankitasw:
> No. You modified the interface by adding two new functions. Which means that this change is now backwards incompatible since people requiring that interface will have a broken code.
>
> I'm not sure how we usually deal with changes like these. @richardcase @Ankitasw ?

Generally we don't add API changes to the existing API version, we wait for the API version bump, as these APIs are exposed and people who are already using CAPA latest version, their clusters would break if they are using these exposed APIs. I would suggest we wait for v1beta3 API for this change, but I would also defer to @richardcase

richardcase:
There are 2 types of API changes going on here, the k8s CRD change and changes to exported Go packages.

The new `NatGatewaysIPs` field is optional, because of this, it's a compatible change as it doesn't require the user to do anything for normal behaviour of CAPA. If the field was required or if the change was changing the behaviour of an existing field, then it would be different. So, even though the apidiff is marking it as a compatible change, it's always worth checking the nuance of the field change.

With the other changes to the scopes that apidiff is complaining about. Historically, even though we have apidiff we haven't guaranteed not breaking the Go API for people that have imported the packages directly (we should probably revisit this decision and also the use of `internal` in the office hours soon). Also, as we are just adding functions and any consumers of the Go API can choose not to bump their imports, so i think we can override this.

/override pull-cluster-api-provider-aws-apidiff-main

Skarlso:
Sounds good. Thanks, Richard!
/lgtm
/approve

PR #4438, opened 2023-08-04
✨ Supporting externally managed Control Plane

### LINKED ISSUE #4437: Supporting externally managed Control Plane
/kind feature

[Kamaji](https://github.com/clastix/kamaji) is an Open Source project that is offering Kubernetes Control Plane as a Service. Thanks to the CAPI Control Plane provider contract and its [implementation](https://github.com/clastix/cluster-api-control-plane-provider-kamaji) it allows creating workload clusters just made of worker nodes and consuming the kube-apiserver as an external service.

At the current state, the Cluster API Provider for AWS is not supporting this contract since it gives for granted the Control Plane is made of Virtual Machines on AWS, thus provisioning of an AWS Load Balancer, or the reference to an already created one is mandatory and doesn't provide a chance to support an externally managed.

As briefly discussed with @richardcase, an annotation for the `AWSCluster` could skip the creation of the AWS Load Balancer: in a such situation, the AWS Cluster API provider will wait for the Control Plane provider to patch the `/AWSCluster/spec/controlPlaneEndpoint` object with the externally managed endpoint.

The proposed implementation sounds similar to what we achieved, as Kamaji developers, with other infrastructure providers we tested, such as vSphere, KubeVirt, Equinix, and OpenStack (although the latter supports a field to reference a fixed API server address and port).

I'm already working on a PR to address this use case and discussing the implementation.

**Environment:**

- Cluster-api-provider-aws version: v2.2.1
- Kubernetes version: (use `kubectl version`): N.R.
- OS (e.g. from `/etc/os-release`): N.R.

prometherion:
**What type of PR is this?**

/kind feature

Cluster API allows defining externally managed control plane for the given cluster, the current code-base doesn't consider this since it nevertheless performs the following actions:

- reconciliation of a Load Balancer, although already provisioned by the ControlPlane provider
- blocking the EC2 creation since it requires an ELB DNS name

Fixes #4437

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
A new enum value, `disabled`, is added for the `AWSCluster.spec.controlPlaneLoadBalancer.loadBalancerType` field, which skips the reconciliation of the load balancer for the given cluster, useful for clusters which are consuming an externally managed Control Plane.
```

### DIFF
--- a/api/v1beta2/awscluster_types.go
+++ b/api/v1beta2/awscluster_types.go
@@ -167,10 +167,11 @@ type Bastion struct {
 type LoadBalancerType string
 
 var (
-	LoadBalancerTypeClassic = LoadBalancerType("classic")
-	LoadBalancerTypeELB     = LoadBalancerType("elb")
-	LoadBalancerTypeALB     = LoadBalancerType("alb")
-	LoadBalancerTypeNLB     = LoadBalancerType("nlb")
+	LoadBalancerTypeClassic  = LoadBalancerType("classic")
+	LoadBalancerTypeELB      = LoadBalancerType("elb")
+	LoadBalancerTypeALB      = LoadBalancerType("alb")
+	LoadBalancerTypeNLB      = LoadBalancerType("nlb")
+	LoadBalancerTypeDisabled = LoadBalancerType("disabled")
 )
 
 // AWSLoadBalancerSpec defines the desired state of an AWS load balancer.
@@ -229,7 +230,7 @@ type AWSLoadBalancerSpec struct {
 
 	// LoadBalancerType sets the type for a load balancer. The default type is classic.
 	// +kubebuilder:default=classic
-	// +kubebuilder:validation:Enum:=classic;elb;alb;nlb
+	// +kubebuilder:validation:Enum:=classic;elb;alb;nlb;disabled
 	LoadBalancerType LoadBalancerType `json:"loadBalancerType,omitempty"`
 
 	// DisableHostsRewrite disabled the hair pinning issue solution that adds the NLB's address as 127.0.0.1 to the hosts
--- a/api/v1beta2/awscluster_webhook.go
+++ b/api/v1beta2/awscluster_webhook.go
@@ -298,5 +298,49 @@ func (r *AWSCluster) validateControlPlaneLBs() field.ErrorList {
 		}
 	}
 
+	if r.Spec.ControlPlaneLoadBalancer.LoadBalancerType == LoadBalancerTypeDisabled {
+		if r.Spec.ControlPlaneLoadBalancer.Name != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "name"), r.Spec.ControlPlaneLoadBalancer.Name, "cannot configure a name if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if r.Spec.ControlPlaneLoadBalancer.CrossZoneLoadBalancing {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "crossZoneLoadBalancing"), r.Spec.ControlPlaneLoadBalancer.CrossZoneLoadBalancing, "cross-zone load balancing cannot be set if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if len(r.Spec.ControlPlaneLoadBalancer.Subnets) > 0 {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "subnets"), r.Spec.ControlPlaneLoadBalancer.Subnets, "subnets cannot be set if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if r.Spec.ControlPlaneLoadBalancer.HealthCheckProtocol != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "healthCheckProtocol"), r.Spec.ControlPlaneLoadBalancer.HealthCheckProtocol, "healthcheck protocol cannot be set if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if len(r.Spec.ControlPlaneLoadBalancer.AdditionalSecurityGroups) > 0 {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "additionalSecurityGroups"), r.Spec.ControlPlaneLoadBalancer.AdditionalSecurityGroups, "additional Security Groups cannot be set if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if len(r.Spec.ControlPlaneLoadBalancer.AdditionalListeners) > 0 {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "additionalListeners"), r.Spec.ControlPlaneLoadBalancer.AdditionalListeners, "cannot set additional listeners if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if len(r.Spec.ControlPlaneLoadBalancer.IngressRules) > 0 {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "ingressRules"), r.Spec.ControlPlaneLoadBalancer.IngressRules, "ingress rules cannot be set if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if r.Spec.ControlPlaneLoadBalancer.PreserveClientIP {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "preserveClientIP"), r.Spec.ControlPlaneLoadBalancer.PreserveClientIP, "cannot preserve client IP if the LoadBalancer reconciliation is disabled"))
+		}
+
+		if r.Spec.ControlPlaneLoadBalancer.DisableHostsRewrite {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "disableHostsRewrite"), r.Spec.ControlPlaneLoadBalancer.DisableHostsRewrite, "cannot disable hosts rewrite if the LoadBalancer reconciliation is disabled"))
+		}
+	}
+
+	for _, rule := range r.Spec.ControlPlaneLoadBalancer.IngressRules {
+		if (rule.CidrBlocks != nil || rule.IPv6CidrBlocks != nil) && (rule.SourceSecurityGroupIDs != nil || rule.SourceSecurityGroupRoles != nil) {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "ingressRules"), r.Spec.ControlPlaneLoadBalancer.IngressRules, "CIDR blocks and security group IDs or security group roles cannot be used together"))
+		}
+	}
+
 	return allErrs
 }
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -266,6 +266,45 @@ func (r *AWSClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
 	return nil
 }
 
+func (r *AWSClusterReconciler) reconcileLoadBalancer(clusterScope *scope.ClusterScope, awsCluster *infrav1.AWSCluster) (*time.Duration, error) {
+	retryAfterDuration := 15 * time.Second
+	if clusterScope.AWSCluster.Spec.ControlPlaneLoadBalancer.LoadBalancerType == infrav1.LoadBalancerTypeDisabled {
+		clusterScope.Debug("load balancer reconciliation shifted to external provider, checking external endpoint")
+
+		return r.checkForExternalControlPlaneLoadBalancer(clusterScope, awsCluster), nil
+	}
+
+	elbService := r.getELBService(clusterScope)
+
+	if err := elbService.ReconcileLoadbalancers(); err != nil {
+		clusterScope.Error(err, "failed to reconcile load balancer")
+		conditions.MarkFalse(awsCluster, infrav1.LoadBalancerReadyCondition, infrav1.LoadBalancerFailedReason, infrautilconditions.ErrorConditionAfterInit(clusterScope.ClusterObj()), err.Error())
+		return nil, err
+	}
+
+	if awsCluster.Status.Network.APIServerELB.DNSName == "" {
+		conditions.MarkFalse(awsCluster, infrav1.LoadBalancerReadyCondition, infrav1.WaitForDNSNameReason, clusterv1.ConditionSeverityInfo, "")
+		clusterScope.Info("Waiting on API server ELB DNS name")
+		return &retryAfterDuration, nil
+	}
+
+	clusterScope.Debug("looking up IP address for DNS", "dns", awsCluster.Status.Network.APIServerELB.DNSName)
+	if _, err := net.LookupIP(awsCluster.Status.Network.APIServerELB.DNSName); err != nil {
+		clusterScope.Error(err, "failed to get IP address for dns name", "dns", awsCluster.Status.Network.APIServerELB.DNSName)
+		conditions.MarkFalse(awsCluster, infrav1.LoadBalancerReadyCondition, infrav1.WaitForDNSNameResolveReason, clusterv1.ConditionSeverityInfo, "")
+		clusterScope.Info("Waiting on API server ELB DNS name to resolve")
+		return &retryAfterDuration, nil
+	}
+	conditions.MarkTrue(awsCluster, infrav1.LoadBalancerReadyCondition)
+
+	awsCluster.Spec.ControlPlaneEndpoint = clusterv1.APIEndpoint{
+		Host: awsCluster.Status.Network.APIServerELB.DNSName,
+		Port: clusterScope.APIServerPort(),
+	}
+
+	return nil, nil
+}
+
 func (r *AWSClusterReconciler) reconcileNormal(clusterScope *scope.ClusterScope) (reconcile.Result, error) {
 	clusterScope.Info("Reconciling AWSCluster")
 
@@ -280,7 +319,6 @@ func (r *AWSClusterReconciler) reconcileNormal(clusterScope *scope.ClusterScope)
 	}
 
 	ec2Service := r.getEC2Service(clusterScope)
-	elbService := r.getELBService(clusterScope)
 	networkSvc := r.getNetworkService(*clusterScope)
 	sgService := r.getSecurityGroupService(*clusterScope)
 	s3Service := s3.NewService(clusterScope)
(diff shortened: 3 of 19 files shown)


pkg/cloud/scope/machine.go
```diff
@@ -370,6 +372,41 @@ func (m *MachineScope) IsEKSManaged() bool {
 	return m.InfraCluster.InfraCluster().GetObjectKind().GroupVersionKind().Kind == ekscontrolplanev1.AWSManagedControlPlaneKind
 }
 
+func (m *MachineScope) IsControlPlaneExternallyManaged() bool {
```
prometherion, line 375: Honestly, I'm not so much happy about this solution, since several steps are returning errors, and we should avoid that at the beginning for the scope.

I'd suggest introducing a new function to the interface `ClusterScoper`,  such as `ControlPlane`, which returns the Control Plane object in order to check its status if it is externally managed.

prometherion: Addressed with changes from be4b28aff.

controllers/awscluster_controller.go
```diff
@@ -252,6 +262,44 @@ func (r *AWSClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
...
+		return nil, err
+	}
+
+	if awsCluster.Status.Network.APIServerELB.DNSName == "" {
+		conditions.MarkFalse(awsCluster, infrav1.LoadBalancerReadyCondition, infrav1.WaitForDNSNameReason, clusterv1.ConditionSeverityInfo, "")
+		clusterScope.Info("Waiting on API server ELB DNS name")
+		return pointer.Duration(15 * time.Second), nil
+	}
+
+	clusterScope.Debug("looking up IP address for DNS", "dns", awsCluster.Status.Network.APIServerELB.DNSName)
```
vincepri, line 286:
```suggestion
	clusterScope.Debug("Looking up IP address for DNS", "dns", awsCluster.Status.Network.APIServerELB.DNSName)
```

controllers/awsmachine_controller.go
```diff
@@ -1141,3 +1149,19 @@ func (r *AWSMachineReconciler) ensureInstanceMetadataOptions(ec2svc services.EC2
 
 	return ec2svc.ModifyInstanceMetadataOptions(instance.ID, machine.Spec.InstanceMetadataOptions)
 }
+
+func (r *AWSMachineReconciler) getControlPlane(ctx context.Context, log *logger.Logger, cluster *clusterv1.Cluster) (*unstructured.Unstructured, error) {
+	controlPlane := unstructured.Unstructured{}
```
vincepri, line 1154:
```suggestion
	controlPlane := new(unstructured.Unstructured)
```
So we carry the pointer around?

controllers/awscluster_controller.go
prometherion: That message was inherited from the previous changes:  https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2891a57069b8365da6c601c2aa8a9ea9ad7cde7c/controllers/awscluster_controller.go#L316

We could address this renaming in a different PR to avoid too many files changes.

prometherion:
If the proposed approach is ok I can work on proper unit-tests for this.

controllers/awsmachine_controller.go
muraee: I think you can use `Get()` func from [sigs.k8s.io/cluster-api/controllers/external](https://github.com/kubernetes-sigs/cluster-api/blob/3a373971a953cdd1db97cdba89db5fba00452041/controllers/external/util.go#L34) pkg here

prometherion: Thanks, @muraee, implemented with bc6370bdb.

controllers/awscluster_controller.go
```diff
@@ -252,6 +262,44 @@ func (r *AWSClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
 	return reconcile.Result{}, nil
 }
 
+func (r *AWSClusterReconciler) reconcileLoadBalancer(clusterScope *scope.ClusterScope, awsCluster *infrav1.AWSCluster) (*time.Duration, error) {
+	if annotations.Has(awsCluster, DisableLoadBalancerReconciliationAnnotation) {
+		clusterScope.Info("load balancer reconciliation disabled")
```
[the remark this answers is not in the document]
Skarlso: Yep. Seconded the Debug level.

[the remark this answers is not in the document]
Skarlso: Actually... I would like to see in the log. But slightly different wording. Reconciliation is _not_ disabled. We still reconcile, we are just checking for external load balancer. So say something like this instead:

`load balancer reconciliation shifted to external provider, checking external endpoint`.

controllers/awsmachine_controller.go
```diff
@@ -197,10 +199,16 @@ func (r *AWSMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	infrav1.SetDefaults_AWSMachineSpec(&awsMachine.Spec)
 
+	cp, err := r.getControlPlane(ctx, log, cluster)
```
Skarlso, line 202: I'm sorry, I'm probably missing something, but how is this related change? 🤔

controllers/awscluster_controller.go
Ankitasw: Since its a minor nit change, feel free to do it in this PR itself

controllers/awscluster_controller.go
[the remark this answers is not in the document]
prometherion: Addressed with b438e14f2.

controllers/awsmachine_controller.go
prometherion: The Control Plane object is required since the Cluster API contract to determine if it's externally managed resides on its status (`/status/externalManagedControlPlane` boolean value).

Without it, we cannot understand if the scope is actually `IsControlPlaneExternallyManaged` or not.

Skarlso: Understood. Thanks for the explanation!

controllers/awscluster_controller.go
prometherion: Addressed with 58e62e7b0

controllers/awscluster_controller.go
```diff
@@ -58,6 +60,14 @@ import (
 	"sigs.k8s.io/cluster-api/util/predicates"
 )
 
+const (
+	// DisableLoadBalancerReconciliationAnnotation allows skipping the reconciliation
+	// of a Load Balancer for the given AWSCluster.
+	// This annotation is used then a Control Plane provider, such as an external one,
+	// is providing by itself the controlPlaneEndpoint resource.
+	DisableLoadBalancerReconciliationAnnotation = "sigs.k8s.io/cluster-api-provider-aws-disable-loadbalancer"
```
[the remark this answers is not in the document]
vincepri: A new field is a backward compatible change and does not require a new API revision

controllers/awscluster_controller.go
```diff
@@ -252,6 +263,44 @@ func (r *AWSClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
 	return nil
 }
 
+func (r *AWSClusterReconciler) reconcileLoadBalancer(clusterScope *scope.ClusterScope, awsCluster *infrav1.AWSCluster) (*time.Duration, error) {
+	if annotations.Has(awsCluster, DisableLoadBalancerReconciliationAnnotation) {
+		clusterScope.Debug("load balancer reconciliation shifted to external provider, checking external endpoint")
+
+		return r.checkForExternalControlPlaneLoadBalancer(clusterScope, awsCluster), nil
+	}
+
```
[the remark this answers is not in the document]
prometherion: We don't know if the externally managed control plane will create its own IP address, as e.g. Kamaji is going, or rather because the user is hard-coding a pre-defined IP address in the `/AWSCluster/spec/controlPlaneEndpoint` field.

In such edge-case scenarios, the knob through the annotation will skip the creation of the infrastructure components that would be unnecessary.

If you prefer, we could rework the PR and embed this logic in the retrieval of the Control Plane object: if it's marked as externally managed, we can skip the CP infrastructure components.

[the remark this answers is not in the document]
prometherion: Fixed by adding a field to the `AWSCluster` type, which made this outdated.

controllers/awscluster_controller.go
```diff
@@ -442,3 +460,27 @@ func (r *AWSClusterReconciler) requeueAWSClusterForUnpausedCluster(_ context.Con
...
 	}
 }
+
+func (r *AWSClusterReconciler) checkForExternalControlPlaneLoadBalancer(clusterScope *scope.ClusterScope, awsCluster *infrav1.AWSCluster) *time.Duration {
+	switch {
+	case len(awsCluster.Spec.ControlPlaneEndpoint.Host) == 0 && awsCluster.Spec.ControlPlaneEndpoint.Port == 0:
+		clusterScope.Info("AWSCluster control plane endpoint is still non-populated")
+		conditions.MarkFalse(awsCluster, infrav1.LoadBalancerReadyCondition, infrav1.WaitForExternalControlPlaneEndpointReason, clusterv1.ConditionSeverityInfo, "")
+
+		return pointer.Duration(15 * time.Second)
```
faiq, line 470: nit: let's make the duration as a constant

api/v1beta2/awscluster_types.go
```diff
@@ -221,7 +222,7 @@ type AWSLoadBalancerSpec struct {
 
 	// LoadBalancerType sets the type for a load balancer. The default type is classic.
 	// +kubebuilder:default=classic
-	// +kubebuilder:validation:Enum:=classic;elb;alb;nlb
+	// +kubebuilder:validation:Enum:=classic;elb;alb;nlb;disabled
```
faiq, line 225: One thing that I think would be helpful is having a webhook validation that looks to see if the other related LoadBalancer options are not set. 

For example, I shouldn't be able to set things like `Scheme, HealthCheckProtocol, AdditionalSecurityGroups` and so on if the loadbalancertype == disabled

dlipovetsky:
In today's CAPA office hours, we talked about adding a test that will verify the expected behavior, and prevent CAPA from breaking downstream implementations of a "managed control plane." During the discussion, an end-to-end test did not seem practical, since it requires running one (or more) of these implementations, but @prometherion will explore some other approaches.

api/v1beta2/awscluster_types.go
prometherion: Implemented.

prometherion:
I think this is ready, I tested it personally by combining it with Kamaji.

```yaml
kind: Cluster
metadata:
  labels:
    ccm: external
    csi: external
  name: capi-quickstart
  namespace: default
spec:
  clusterNetwork:
    pods:
      cidrBlocks:
      - 192.168.0.0/16
  controlPlaneRef:
    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
    kind: KamajiControlPlane
    name: capi-quickstart-control-plane
  infrastructureRef:
    kind: AWSCluster
    name: capi-quickstart
---
kind: AWSCluster
metadata:
  name: capi-quickstart
  namespace: default
spec:
  region: us-east-1
  sshKeyName: default
  controlPlaneLoadBalancer:
    loadBalancerType: disabled
---
apiVersion: controlplane.cluster.x-k8s.io/v1alpha1
kind: KamajiControlPlane
metadata:
  name: capi-quickstart-control-plane
  namespace: default
spec:
  dataStoreName: default
  addons:
    coreDNS: { }
    kubeProxy: { }
  kubelet:
    cgroupfs: systemd
    preferredAddressTypes:
      - ExternalIP
      - InternalIP
      - Hostname
  network:
    serviceAddress: 78.134.89.204
    serviceType: ClusterIP
  deployment:
  replicas: 2
  version: 1.26.0
---
kind: MachineDeployment
metadata:
  name: capi-quickstart-md-0
  namespace: default
spec:
  clusterName: capi-quickstart
  replicas: 2
  selector:
    matchLabels: null
  template:
    spec:
      bootstrap:
        configRef:
          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
          kind: KubeadmConfigTemplate
          name: capi-quickstart-md-0
      clusterName: capi-quickstart
      infrastructureRef:
        kind: AWSMachineTemplate
        name: capi-quickstart-md-0
      version: v1.26.0
---
kind: AWSMachineTemplate
metadata:
  name: capi-quickstart-md-0
  namespace: default
spec:
  template:
    spec:
      iamInstanceProfile: nodes.cluster-api-provider-aws.sigs.k8s.io
      instanceType: t3.large
---
apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
kind: KubeadmConfigTemplate
metadata:
  name: capi-quickstart-md-0
  namespace: default
spec:
  template:
    spec:
      joinConfiguration:
        nodeRegistration:
          kubeletExtraArgs:
            cloud-provider: external
          name: '{{ ds.meta_data.local_hostname }}'
```

> JFI, the Control Plane is running on my lab environment, outside of AWS

Once the provisioning is complete, the nodes are joining as expected:
```
NAME                          STATUS     ROLES    AGE    VERSION
ip-10-0-114-74.ec2.internal   NotReady   <none>   112s   v1.26.0
ip-10-0-96-155.ec2.internal   NotReady   <none>   113s   v1.26.0
```

The `clusterctl` describe output:
```
NAME                                                               READY  SEVERITY  REASON                       SINCE  MESSAGE                                                                
Cluster/capi-quickstart                                            True                                          3m20s                                                                          
├─ClusterInfrastructure - AWSCluster/capi-quickstart               True                                          4m6s                                                                           
├─ControlPlane - KamajiControlPlane/capi-quickstart-control-plane                                                                                                                               
└─Workers                                                                                                                                                                                       
  └─MachineDeployment/capi-quickstart-md-0                         False  Warning   WaitingForAvailableMachines  8m22s  Minimum availability requires 2 replicas, current 0 available           
    └─2 Machines...                                                True                                          2m32s  See capi-quickstart-md-0-lp5xc-f4bbx, capi-quickstart-md-0-lp5xc-q254x  
```

> please, notice I skipped the creation of the CCM along with the CNI plugin for the sake of speed testing

Could we trigger again the e2e testing, please?

richardcase:
Until #4733 merges

/hold

But this looks good to me, pending any changes required after 4733.

prometherion:
Thanks, @richardcase, rebased as requested.

The failed test seems unrelated to the changes we introduced.

richardcase:
Thanks @prometherion 

/unhold

faiq:
the tests are all passing, and the code looks good. i think this is ready to merge

api/v1beta2/awscluster_webhook.go
```diff
@@ -298,5 +298,49 @@ func (r *AWSCluster) validateControlPlaneLBs() field.ErrorList {
 		}
 	}
 
+	if r.Spec.ControlPlaneLoadBalancer.LoadBalancerType == LoadBalancerTypeDisabled {
+		if r.Spec.ControlPlaneLoadBalancer.Name != nil {
```
richardcase, line 302: I'm wondering if we need a `func Empty() bool` on the load balancer instead of checking each sub field. But lets discuss that separately.

controllers/awsmachine_controller.go
```diff
@@ -1178,3 +1186,22 @@ func (r *AWSMachineReconciler) ensureInstanceMetadataOptions(ec2svc services.EC2
 
 	return ec2svc.ModifyInstanceMetadataOptions(instance.ID, machine.Spec.InstanceMetadataOptions)
 }
+
+// +kubebuilder:rbac:groups=controlplane.cluster.x-k8s.io,resources=*,verbs=get;list;watch
```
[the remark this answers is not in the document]
prometherion: TBH, this wildcard makes obsolete the following ones:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/6193d172e0887d08319a63e07720ac9f7f08a841/controllers/rosacluster_controller.go#L55
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/6193d172e0887d08319a63e07720ac9f7f08a841/bootstrap/eks/controllers/eksconfig_controller.go#L62
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/6193d172e0887d08319a63e07720ac9f7f08a841/controllers/awsmanagedcluster_controller.go#L54
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/6193d172e0887d08319a63e07720ac9f7f08a841/controllers/rosacluster_controller.go#L55
...and many other lines.

Although we could optimize the role rules, I think these  are more intuitive for developers, since it's pretty straightforward trying to understand the objects controlled by the various controllers.

api/v1beta2/awscluster_webhook.go
prometherion: We still would need the check of each subfield, unfortunately. We could slender the webhook file, of course, but from a UX perspective, knowing which field is breaking the validation would be helpful.

api/v1beta2/awscluster_types.go
```diff
@@ -229,7 +230,7 @@ type AWSLoadBalancerSpec struct {
 
 	// LoadBalancerType sets the type for a load balancer. The default type is classic.
```
vincepri, line 231: Could we add a comment on what each value for this does and what the new value is useful for?

api/v1beta2/awscluster_webhook.go
```diff
@@ -298,5 +298,49 @@ func (r *AWSCluster) validateControlPlaneLBs() field.ErrorList {
 		}
 	}
 
+	if r.Spec.ControlPlaneLoadBalancer.LoadBalancerType == LoadBalancerTypeDisabled {
```
vincepri, line 301: Are we validating somewhere that the `type` cannot be changed once created? How are we dealing right now if a user creates the type = `disabled` and then wants `alb` and vice-versa?

api/v1beta2/awscluster_webhook.go
```diff
@@ -298,5 +298,49 @@ func (r *AWSCluster) validateControlPlaneLBs() field.ErrorList {
...
+		if r.Spec.ControlPlaneLoadBalancer.DisableHostsRewrite {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "disableHostsRewrite"), r.Spec.ControlPlaneLoadBalancer.DisableHostsRewrite, "cannot disable hosts rewrite if the LoadBalancer reconciliation is disabled"))
+		}
+	}
+
+	for _, rule := range r.Spec.ControlPlaneLoadBalancer.IngressRules {
+		if (rule.CidrBlocks != nil || rule.IPv6CidrBlocks != nil) && (rule.SourceSecurityGroupIDs != nil || rule.SourceSecurityGroupRoles != nil) {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "controlPlaneLoadBalancer", "ingressRules"), r.Spec.ControlPlaneLoadBalancer.IngressRules, "CIDR blocks and security group IDs or security group roles cannot be used together"))
+		}
+	}
```
vincepri, line 343: We should open an issue to track for the next API version to rework this, it seems to me that we're going around the current API structure to disable the control plane management

controllers/awscluster_controller.go
```diff
@@ -266,6 +266,45 @@ func (r *AWSClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
 	return nil
 }
 
+func (r *AWSClusterReconciler) reconcileLoadBalancer(clusterScope *scope.ClusterScope, awsCluster *infrav1.AWSCluster) (*time.Duration, error) {
+	retryAfterDuration := 15 * time.Second
```
vincepri, line 270: Could we rework this to just return an `error` and leave the `retryAfterDuration` outside?

controllers/awscluster_controller.go
```diff
@@ -447,3 +465,29 @@ func (r *AWSClusterReconciler) requeueAWSClusterForUnpausedCluster(_ context.Con
 		}
 	}
 }
+
+func (r *AWSClusterReconciler) checkForExternalControlPlaneLoadBalancer(clusterScope *scope.ClusterScope, awsCluster *infrav1.AWSCluster) *time.Duration {
+	requeueAfterPeriod := 15 * time.Second
+
```
vincepri, line 471: Same as above, the retry after should be outside in a caller

controllers/awsmachine_controller.go
```diff
@@ -200,10 +202,16 @@ func (r *AWSMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	infrav1.SetDefaults_AWSMachineSpec(&awsMachine.Spec)
 
+	cp, err := r.getControlPlane(ctx, log, cluster)
+	if err != nil {
+		return ctrl.Result{}, err
+	}
```
vincepri, line 208: It seems we're missing that the control plane is an optional reference, and this is going to panic when nil

controllers/awsmachine_controller.go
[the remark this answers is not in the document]
vincepri: Let's move these to a single place above

pkg/cloud/scope/machine.go
```diff
@@ -371,6 +377,10 @@ func (m *MachineScope) IsEKSManaged() bool {
...
```
vincepri, line 380: A bit confusing on why this is needed, the control plane reference can be anything, including a new CRD; why do we need to mark it as externally managed?

controllers/awsmachine_controller.go
prometherion: Potentially could be solved with #4817.

Issue #4026, opened 2023-01-31
SubnetSpec changes in `v1beta2` break existing use-cases


AverageMarcus:
/kind bug
/kind api-change

The changes introduced in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3748 (aimed at solving server side apply issues with Cluster Class) break existing use-cases around subnets created by CAPA.

The change makes the `ID` property on `SubnetSpec` now be a required field where previously this was optional. This change means that configuration to subnets used for CAPA clusters can now _only_ be done if infrastructure is created by the user resulting in CAPA only allowing for either the default subnet layout or a bring-your-own network where the user would be required to create everything (VPC, subnets, etc.) externally before the creation of the `AWSCluster` CR.

With `v1beta1` and earlier it was possible to specify some of the subnet configuration (such as the CIDR block and the AWS tags) but still have CAPA create the resources for you.

For example, we have a situation where we need more control over what resources end up in what subnets and rely on AWS tags and subnet filters to achieve this. 

```yaml
spec:
  network:
    subnets:
      - cidrBlock: 10.0.0.0/24
        availabilityZone: eu-west-1a
        tags:
          subnet-role: control-plane
      - cidrBlock: 10.0.1.0/24
        availabilityZone: eu-west-1b
        tags:
          subnet-role: control-plane
... etc ...
```

With this approach we can configure CAPA to create all the subnets we need, with the tags we can then filter on, without having to use an external process to create these resources. This is no longer possible with `v1beta2`.

In the PR it was acknowledged that this was a breaking change and that we'd need to come up with an alternative to solve this problem going forward but unfortunately was forgotten about so I'm opening this issue to have that discussion and try to come up with a solution to both the original issue the PR was fixing and the use cases currently in use that require configuring subnets in CAPA.

This also fits into the wider discussion around networking in CAPA and the improvements / changes desired.

**Related:**
* https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3748
* https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3883
* https://kubernetes.slack.com/archives/CD6U2V71N/p1675159432540519
* https://kubernetes.slack.com/archives/CD6U2V71N/p1675159036932519

**Environment:**

- Cluster-api-provider-aws version: v2.0.2 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

[...]

PR #4444, opened 2023-08-07
feat: support IPAM Manager for VPC creation

### LINKED ISSUE #4443: Support AWS IPAM for cluster VPC
/kind feature

AWS allows to use [IPAM Manager](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-ip-addressing.html#vpc-using-ipam]) when creating `VPC` to be able to provide `IPAM-allocated IPv4 CIDR block` to simplify IP address management.

This feature requires 2 parameters to be set:
image

`ID` (or `name`) of the `IPAM pool` and `Netmask`, which will be used for `VPC CIDR`.
It would be nice to support `IPAM Manager` when creating `VPC` for `AWSCluster`

MaxFedotov:
**What type of PR is this?**
/kind feature

Adds support for [IPAM Manager](https://docs.aws.amazon.com/vpc/latest/userguide/vpc-ip-addressing.html#vpc-using-ipam%5D) when creating `VPC` for `AWSCluster`

Fixes #4443

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Support AWS IPAM Manager for VPC CIDR Management
```


api/v1beta2/network_types.go
```diff
@@ -254,6 +266,9 @@ type VPCSpec struct {
 	// Defaults to 10.0.0.0/16.
 	CidrBlock string `json:"cidrBlock,omitempty"`
 
+	// IPAMPool defines the IPAM pool to be used for VPC.
+	IPAMPool *IPAMPool `json:"ipamPool,omitempty"`
```
Skarlso, line 270: Can we add ipv6 ipam as well?

pkg/cloud/services/network/vpc.go
```diff
@@ -211,10 +241,25 @@ func (s *Service) createVPC() (*infrav1.VPCSpec, error) {
...
 
-	if s.scope.VPC().CidrBlock == "" {
-		s.scope.VPC().CidrBlock = defaultVPCCidr
+	if s.scope.VPC().IPAMPool != nil {
+		ipamPoolID, err := s.getIPAMPoolID()
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to get IPAM Pool ID")
+		}
+
+		if s.scope.VPC().IPAMPool.NetmaskLength == 0 {
```
Skarlso, line 250: I'm wondering if this would cause any problems for IPv6 based IPAMs?

pkg/cloud/services/network/vpc.go
```diff
@@ -211,10 +241,25 @@ func (s *Service) createVPC() (*infrav1.VPCSpec, error) {
...
+			s.scope.VPC().IPAMPool.NetmaskLength = defaultIpamV4NetmaskLength
+		}
+
+		input.Ipv4IpamPoolId = ipamPoolID
```
Skarlso, line 254: Especially this part. It definitely has problems with ipv6-based IPAMs.

pkg/cloud/services/network/vpc.go
MaxFedotov: Hey @Skarlso!

If my understanding is correct, all settings in `spec.network.vpc` applies only to `ipv4`. For `ipv6` there is a separate configuration in `spec.network.vpc.ipv6`. 

And according to webhook, ipv6 cannot be used with managed clusters:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/29c442b8a72ddb5677047cf6fcb80358aad8c440/api/v1beta2/awscluster_webhook.go#L182-L193

As for EKS, there is no `ipv6` configuration in `AWSManagedClusterSpec`.

Skarlso: That's not completely true. The IPv6 part there only tells ipv6 specific settings. But this is IPAM. I wouldn't want to copy and replicate every setting there is to the ipv6 field. In the code, we check if ipv6 is enabled via that setting. But other than that some things are set separately by dedicated fields.

MaxFedotov: > I wouldn't want to copy and replicate every setting there is to the ipv6 field.
For IPv6 there should be a separate IPAM pool. So if a user wants to use IPAM for dual-stack vpc, he must create two different pools, and in this case IPAM parameter should be duplicated :(

Skarlso: Oh, that's interesting. I haven't considered that. 🤔 That messes things up a bit indeed. :D

Skarlso: I guess it's fine then.

richardcase:
This generally looks good to me and the e2e are passing.

However, i'm not up to speed with the impact on ipv6 so i will defer approval to those that know more:

/assign Skarlso

Skarlso:
@MaxFedotov I'm still a bit concerned that people will think this is the thing that they will need to use if they have IPAM for IPv6. Do you think we should rename this to IPAMv4?

MaxFedotov:
@Skarlso My assumption was that as `cidrBlock` field is for `ipv4`, all other fields (except `ipv6`) are also naturally related to `ipv4`.  If you think that `IPAM` name will be confusing for users - I'm entirely ok with renaming it to `IPAMv4` :)

Skarlso:
Yah I know, that right now, we don't segregate explicitly ipv4 and 6. Hm, don't know. @richardcase How do you feel? Do you think people would try setting up IPAM for v6 through network: vpc: ipam or network: vpc: ipv6: ipam: ? :)

api/v1beta2/network_types.go
richardcase, line 270: I think we should probably rename this so its specific to IPV4 and then also add a IPAMPool field in the [IPv6](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f43a36bfbe4284aa5f4e8c0b1feac857945adbfb/api/v1beta2/network_types.go#L234) struct.

richardcase: Or probably better the suggestion from @Skarlso 

```yaml
network: 
  vpc:
    ipam:
       ipv6Block: bla
       ipv4Block: blah
```

richardcase: We also need to think about the fact we want to completely evolve the networking side of CAPA as the base types have long assumed IPv4 and so what would the ideal structure be for IPv4 and IPv6 fields. 

We introduced the **IPv6** type (thanks @Skarlso), so do we think we want to also have an **IPv4** type as well going forward?

wdyt @Skarlso @MaxFedotov @AndiDog ?

Skarlso: I think the VPC/IPAM section would be more intuitive if it bundled both options. I can understand if you would like to do that in a separate PR though. But I wouldn't want to separate these. The IPv6 section is rather an indicator to enable ipv6 ATM. The IPAM as it is now conveys the feeling to set both.

Let's do the following. Let's add the ability to set ipv6 in a followup but add the fields now. But set a warning or an error if someone does set ipv6 IPAM to say, it's not yet supported. That way, the followup PR can then be added since the fields are already there.

WDYT?

MaxFedotov: > I think the VPC/IPAM section would be more intuitive if it bundled both options. I can understand if you would like to do that in a separate PR though. But I wouldn't want to separate these. The IPv6 section is rather an indicator to enable ipv6 ATM. The IPAM as it is now conveys the feeling to set both.

I agree that we need to have 2 separate configurations - one for `ipv4` and one for `ipv6`, because the user may use IPAM pools for both, and in order to do it he has to create 2 different pools and they will have different IDs.

But from my POV, as we already have some `ipv6` related configurations in separate `ipv6` struct (like `poolId` or `cidrBlock`), it will be quite properly from UX side to put IPAM configuration here as well. 

And also we have to keep in mind, that when using `ipv6`, the user is not able to specify both `IPv6 CIDR owned by me`(`poolId` in `ipv6` struct) and `ipam pool`. So keeping them in the same struct will make this grouping more logical.

Skarlso: That does make sense. Okay put it under the IPv6 tree then. I'm just cautious here that we don't start piling stuff in both sections of that makes sense.

AndiDog: Putting it in `type IPv6` makes sense, and we should document in the field/struct comments which fields are mutually exclusive (e.g. `IPv6.CidrBlock` and the new `IPv6.IPAMPool*`, I guess). Also, having some fields already added now, but unsupported for the moment, seems a good idea given that they have clear code comments.

MaxFedotov:
Hi @Skarlso and @richardcase. I've updated PR with IPv6 support. Can you please take a look?

Skarlso:
Will do today.

Skarlso:


One last time then
/lgtm

api/v1beta2/network_types.go
```diff
@@ -233,16 +233,36 @@ type NetworkSpec struct {
...
 	EgressOnlyInternetGatewayID *string `json:"egressOnlyInternetGatewayId,omitempty"`
+
+	// IPAMPool defines the IPAMv6 pool to be used for VPC.
+	// Mutually exclusive with CidrBlock.
+	// +optional
+	IPAMPool *IPAMPool `json:"ipamPool,omitempty"`
+}
+
+// IPAMPool defines the IPAM pool to be used for VPC.
+type IPAMPool struct {
```
AndiDog, line 257: Not quite sure about this one. In https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4026 and the subsequent effort to improve networking code, we figured that support for both managed (CAPA creates and deletes it) and unmanaged (CAPA finds it) resources should be very explicit in struct fields. In that issue, the problem was that `ID` could be either filled or empty to denote whether it's an existing or to-be-managed resource.

With the current comments on the fields `ID` and `Name`, as a user, I wouldn't know whether CAPA is going to create the IPAM pool for me. Also, when is which field needed? Can both be given, etc.?

Skarlso: That's a good point. We should add //+reqiure or optional and an easier to digest explanation.

MaxFedotov:
@Skarlso, @AndiDog is there anything else I should do to proceed with this PR?

Issue #3820, opened 2022-11-07
[E2E] Add tests for NLB feature


Skarlso:
/kind feature

**Describe the solution you'd like**
The PR for the NLB implementation is getting super big, so I'm extracting and writing an E2E test for it.
Related to this https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3088.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

Skarlso:
/assign

After the NLB PR is merged.

PR #4449, opened 2023-08-11
Set availability zones in status.networkStatus.apiServerElb when using BYO NLB

### LINKED ISSUE #4448: Multi-AZ control plane not created when using BYO NLB
/kind bug

I provision an NLB outside of CAPA for the control plane and set the name in the AWSCluster spec. It seems that the availability zones are not detected even when specifying subnets explicitly, leading to a control plane with all nodes in a single AZ.

The AWSCluster spec.subnets field is set with subnets spanning AZs, and each subnet has the availabilityZone field correctly set.

I notice in the status.networkStatus.apiServerElb field it has the subnet IDs set but not the availability zones. The status.failureDomains field looks like:
```yaml
failureDomains:
    us-west-2a: {}
    us-west-2b: {}
    us-west-2c: {}
```

Availability zones are detected for the specified subnets, allowing the failureDomains status to be set correctly

I believe the issue is in [fromSDKTypeToLB](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/29c442b8a72ddb5677047cf6fcb80358aad8c440/pkg/cloud/services/elb/loadbalancer.go#L338). Note that AvailabilityZones is not set in infrav1.LoadBalancer. 

It does not affect CAPA managed NLBs since the AvailabilityZones field is explicitly set in reconcileV2LB.

It is quite a simple fix that I will submit a PR for in a few minutes

**Environment:**

- Cluster-api-provider-aws version: v2.2.1
- Kubernetes version: (use `kubectl version`): 1.25
- OS (e.g. from `/etc/os-release`): ubuntu

ttreptow:
This ensures that failure domains are correctly set when using an externally managed (BYO) NLB

**What type of PR is this?**
/kind bug

When using a BYO NLB, control plane nodes are only created in a single AZ regardless of the subnet AZs specified. This is due to the availability zones not being set in the lb status field, which causes the failure domains to not be detected in the AWSCluster reconciliation.

Fixes #4448

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Availability zone status is correctly set when using BYO NLBs, fixing an issue where control plane nodes were only created in a single AZ
```

### DIFF
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -1467,8 +1467,10 @@ func fromSDKTypeToClassicELB(v *elb.LoadBalancerDescription, attrs *elb.LoadBala
 
 func fromSDKTypeToLB(v *elbv2.LoadBalancer, attrs []*elbv2.LoadBalancerAttribute, tags []*elbv2.Tag) *infrav1.LoadBalancer {
 	subnetIds := make([]*string, len(v.AvailabilityZones))
+	availabilityZones := make([]*string, len(v.AvailabilityZones))
 	for i, az := range v.AvailabilityZones {
 		subnetIds[i] = az.SubnetId
+		availabilityZones[i] = az.ZoneName
 	}
 	res := &infrav1.LoadBalancer{
 		ARN:       aws.StringValue(v.LoadBalancerArn),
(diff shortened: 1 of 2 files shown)


Ankitasw:
@Skarlso do we have NLB related E2E test?

Skarlso:
@Ankitasw Sadly we don't have those :/ #3820... :D I'm afraid I just never got around to it.

Ankitasw:
LGTM, but would need another pair of eyes @Skarlso 😄 

/approve

PR #4463, opened 2023-08-25
fix: don't generate nodegroup role name starting with hyphen

DavidSpek:
**What type of PR is this?**
/kind bug

Currently if `EnableIAM` is set to true and no role is specified for a node group the controller auto generates a role. However, the name formatting seems to be backwards from what it is intended to be, causing the role name to start with a `-`. What's surprising to me is that this doesn't seem to cause problems, other than that it looks weird. The name for the role that is currently generated looks like: `-nodegroup-iam-service-role_<cluster-name>-<nodegroup-name>`. This PR swaps the generated role name so it will be: `<cluster-name>-<nodegroup-name>_nodegroup-iam-service-role`. This is follows the same format as the name generated for Fargate.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Change generated nodegroup IAM role name from `-nodegroup-iam-service-role_<cluster-name>-<nodegroup-name>` to `<cluster-name>-<nodegroup-name>_nodegroup-iam-service-role`
```

### DIFF
--- a/pkg/cloud/services/eks/roles.go
+++ b/pkg/cloud/services/eks/roles.go
@@ -182,8 +182,8 @@ func (s *NodegroupService) reconcileNodegroupIAMRole() error {
 		} else {
 			s.scope.Info("no EKS nodegroup role specified, using role based on nodegroup name")
 			roleName, err = eks.GenerateEKSName(
+				"nodegroup-iam-service-role",
 				fmt.Sprintf("%s-%s", s.scope.KubernetesClusterName(), s.scope.NodegroupName()),
-				"-nodegroup-iam-service-role",
 				maxIAMRoleNameLength,
 			)
 			if err != nil {


DavidSpek:
@richardcase Have you had the chance to take a look at this PR?

Ankitasw:
/lgtm

cc @richardcase for approval

Skarlso:
This will definitely cause existing clusters to break right? Since it will look for a specific role format?

Ankitasw:
@Skarlso that's a good point, shall we hold it for next release?

Skarlso:
Yep, sounds good.

DavidSpek:
I'm not sure if it would break existing clusters, since this code is only used to create a new role if none were specified in the node group, which then get placed into the node group CR. Existing node group CRs would have the name of the role already in them.

Skarlso:
Okay, thinking about this some more, it should be fine. It should only happen for new clusters and old clusters should work. UNLESS, for whatever reason, the role name is wiped and it needs to look for it again.

DavidSpek:
Is there even logic to handle having the role name being wiped properly? The field should be immutable and if it gets removed I don't think there's a way to be able to recover what it should be (since it could also be a non-generated name).

Ankitasw:

DavidSpek:
@AndiDog Maybe you can have a look at this as well.

DavidSpek:
@Ankitasw @Skarlso What needs to be done before this PR can get merged?

Skarlso:
I'll take a look later today. Been busy with work.

Skarlso:
Okay, we did run tests, I thought this through, and I _think_ it shouldn't break anything. "Famous last words".

Here we go:
/lgtm
/approve

PR #4466, opened 2023-08-28
Make garbage collection configurable

Fedosin:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
This PR allows to configure garbage collection and define what clean-up tasks to execute during cluster deprovisioning.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #3541

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Make garbage collection configurable
```

PR #4485, opened 2023-09-04
Add validation for GC tasks annotation

Fedosin:
**What type of PR is this?**
/kind support

This PR adds validation for the GC config annotation. 

Fixes #

As a part of this work 2 constants `ExternalResourceGCAnnotation` and `ExternalResourceGCTasksAnnotation` were moved from `exp/api/v1beta2` to `api/v1beta2`.
**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
None
```

### DIFF
--- a/api/v1beta2/awscluster_webhook.go
+++ b/api/v1beta2/awscluster_webhook.go
@@ -18,6 +18,7 @@ package v1beta2
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/google/go-cmp/cmp"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
@@ -69,6 +70,8 @@ func (r *AWSCluster) ValidateDelete() error {
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type.
 func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 	var allErrs field.ErrorList
+
+	allErrs = append(allErrs, r.validateGCTasksAnnotation()...)
 
 	oldC, ok := old.(*AWSCluster)
 	if !ok {
@@ -173,6 +176,42 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 // Default satisfies the defaulting webhook interface.
 func (r *AWSCluster) Default() {
 	SetObjectDefaults_AWSCluster(r)
+}
+
+func (r *AWSCluster) validateGCTasksAnnotation() field.ErrorList {
+	var allErrs field.ErrorList
+
+	annotations := r.GetAnnotations()
+	if annotations == nil {
+		return nil
+	}
+
+	if gcTasksAnnotationValue := annotations[ExternalResourceGCTasksAnnotation]; gcTasksAnnotationValue != "" {
+		gcTasks := strings.Split(gcTasksAnnotationValue, ",")
+
+		supportedGCTasks := []GCTask{GCTaskLoadBalancer, GCTaskTargetGroup, GCTaskSecurityGroup}
+
+		for _, gcTask := range gcTasks {
+			found := false
+
+			for _, supportedGCTask := range supportedGCTasks {
+				if gcTask == string(supportedGCTask) {
+					found = true
+					break
+				}
+			}
+
+			if !found {
+				allErrs = append(allErrs,
+					field.Invalid(field.NewPath("metadata", "annotations"),
+						r.Annotations,
+						fmt.Sprintf("annotation %s contains unsupported GC task %s", ExternalResourceGCTasksAnnotation, gcTask)),
+				)
+			}
+		}
+	}
+
+	return allErrs
 }
 
 func (r *AWSCluster) validateSSHKeyName() field.ErrorList {
--- a/cmd/clusterawsadm/gc/gc.go
+++ b/cmd/clusterawsadm/gc/gc.go
@@ -29,7 +29,6 @@ import (
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta2"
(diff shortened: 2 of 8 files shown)


api/v1beta2/awscluster_webhook.go
```diff
@@ -163,6 +164,32 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 		)
 	}
 
+	if gcTasksAnotationValue := r.GetAnnotations()[ExternalResourceGCTasksAnnotation]; gcTasksAnotationValue != "" {
+		gcTasks := strings.Split(gcTasksAnotationValue, ",")
+
+		// TODO: Should we make these values constants?
```
Ankitasw, line 170: Yes, constants sounds like good idea 👍

Ankitasw (commented):
Also, I am not aware but does this require any changes in E2E GC tests as well? cc @richardcase

Ankitasw:
@richardcase since this is API change, shall we include this as part of v1beta3 changes?

Fedosin:
I think we can avoid introducing a breaking change if we move `ExternalResourceGCTasksAnnotation` constant only. A new version (2.3) hasn't been released yet, so we can modify that part safely.

api/v1beta2/awscluster_webhook.go
Fedosin: Let me add them then

cmd/clusterawsadm/gc/gc.go
```diff
@@ -132,7 +131,7 @@ func (c *CmdProcessor) setAnnotationAndPatch(ctx context.Context, annotationValu
 		return fmt.Errorf("creating patch helper: %w", err)
 	}
 
-	annotations.Set(infraObj, expinfrav1.ExternalResourceGCAnnotation, annotationValue)
+	annotations.Set(infraObj, infrav1.ExternalResourceGCAnnotation, annotationValue)
```
Ankitasw, line 134: These changes should be reverted as well.

Fedosin: I thought it may be easier to use these constants from api/v1beta2 directly.

exp/api/v1beta2/types.go
```diff
@@ -25,11 +25,11 @@ import (
 const (
 	// ExternalResourceGCAnnotation is the name of an annotation that indicates if
 	// external resources should be garbage collected for the cluster.
-	ExternalResourceGCAnnotation = "aws.cluster.x-k8s.io/external-resource-gc"
+	ExternalResourceGCAnnotation = infrav1.ExternalResourceGCAnnotation
```
Ankitasw, line 28: I think this would be duplication, either we keep these constants in exp/api/v1beta2, or in api/v1beta2

cmd/clusterawsadm/gc/gc.go
Ankitasw: It would be better to keep it in exp only, to avoid confusion

exp/api/v1beta2/types.go
Fedosin: Now these constants are in `api/v1beta2`, but I kept them as duplicates in `exp/api/v1beta2` to prevent a breaking change in the API. People who used the constants from `exp` can continue to do so, but now the correct location is `api/v1beta2`.

dlipovetsky:
This is a follow-up to #4466

exp/api/v1beta2/types.go
Ankitasw: I might need more opinions here @richardcase @Skarlso @dlipovetsky

Skarlso: It's OK not to duplicate it. If we remove it, make sure this gets into a minor release, though, because it might be a breaking change. Although, honestly, I doubt too many people call this from anywhere outside of CAPA.

Fedosin: alright! I'll remove it from `exp` completely since we don't consider it as a breaking change. for sure, it will make things simpler.

Skarlso: Noice.

api/v1beta2/awscluster_webhook.go
```diff
@@ -163,6 +164,31 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) error {
 		)
 	}
 
+	if gcTasksAnotationValue := r.GetAnnotations()[ExternalResourceGCTasksAnnotation]; gcTasksAnotationValue != "" {
```
Skarlso, line 167: Just being careful here. What if the annotation does exist but it _IS_ empty?

We should probably rather say something like `v, ok := r.GetAnnotations()[ExternalResourceGCTasksAnnotation]; ok` ?

pkg/cloud/services/gc/cleanup.go
```diff
@@ -71,21 +70,19 @@ func (s *Service) deleteResources(ctx context.Context) error {
...
+			infrav1.GCTaskLoadBalancer:  s.deleteLoadBalancers,
+			infrav1.GCTaskTargetGroup:   s.deleteTargetGroups,
+			infrav1.GCTaskSecurityGroup: s.deleteSecurityGroups,
 		}
 
 		cleanupFuncs = ResourceCleanupFuncs{}
 
 		tasks := strings.Split(val, ",")
 
-		// TODO: add some validation here.
```
Skarlso, line 85: Did the TODO get resolved?

Fedosin: Yes, by this PR

Skarlso: Oh, I thought that the validation is meant to signify some kind of other validation here. :)

api/v1beta2/awscluster_webhook.go
Fedosin: if the annotation exists and its value is empty, then there is nothing to validate and we don't enter this section

Fedosin:
I moved this validation code to a separate function.

PR #4508, opened 2023-09-18
Save Nat Gateways IPs to status on every reconciliation

fiunchinho:
**What type of PR is this?**

/kind bug

When we first create the nat gateways, they go to the `Pending` state. At that point, they don't have a PublicIP just yet. So if we only save the nat gateway public IP to the `status` field right after creating the nat gateway, we won't have the public IP available, meaning we won't be able to save it to the status.

Now the logic is changed so we always save the NAT Gateways IPs to the status field.

I've tested this version in our management clusters, adding our custom ingress rules and it all worked fine.

There was a missmatch in how the ingress rules were modeled in CAPA and the SDK, making the controller to keep reconciliating the same AWSCluster because the `SecurityGroups` saved in the field `.status.networkStatus.securityGroups` never matched the security groups returned by the AWS SDK. Now they match and no extra reconciliations are needed.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix how NAT gateways IPs are saved in the status field
```

### DIFF
--- a/pkg/cloud/services/network/natgateways.go
+++ b/pkg/cloud/services/network/natgateways.go
@@ -69,6 +69,7 @@ func (s *Service) reconcileNatGateways() error {
 		return err
 	}
 
+	natGatewaysIPs := []string{}
 	subnetIDs := []string{}
 
 	for _, sn := range s.scope.Subnets().FilterPublic() {
@@ -77,6 +78,9 @@ func (s *Service) reconcileNatGateways() error {
 		}
 
 		if ngw, ok := existing[sn.ID]; ok {
+			if len(ngw.NatGatewayAddresses) > 0 && ngw.NatGatewayAddresses[0].PublicIp != nil {
+				natGatewaysIPs = append(natGatewaysIPs, *ngw.NatGatewayAddresses[0].PublicIp)
+			}
 			// Make sure tags are up to date.
 			if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
 				buildParams := s.getNatGatewayTagParams(*ngw.NatGatewayId)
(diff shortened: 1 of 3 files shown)


AverageMarcus (commented):
/ok-to-test

pkg/cloud/services/network/natgateways.go
```diff
@@ -69,6 +69,7 @@ func (s *Service) reconcileNatGateways() error {
 		return err
 	}
 
+	var natGatewaysIPs []string
```
AverageMarcus, line 72: For consistency with the next line
```suggestion
	natGatewaysIPs := []string{}
```

Skarlso: Alternatively, the line could change to the var declaration. :)

Skarlso: Especially because it looks like it could be empty, so no need to pre-initialize it.

Ankitasw:
@fiunchinho did you test locally to check the status field for NAT gateway IP?

fiunchinho:
Yes, it's saving the IPs on the status field. But the controller is logging errors, because the ingress rules are being wrongly modified by a bug that it's fixed here https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4359

pkg/cloud/services/network/natgateways.go
fiunchinho: Why is one way preferred over the other? Isn't it a empty slice in both cases?

Ankitasw:
@fiunchinho we can ignore this test failure, it's not related to your change

/lgtm
/approve

PR #4512, opened 2023-09-21
fix: propagate AdditionalTags from AWSCluster to storage volumes

### LINKED ISSUE #4511: AWSCluster AdditionalTags are not propagated to AWSMachine storage volumes
## Bug Description

The AWSCluster resources has a spec.AdditionalTags k-v map that is a place end users (aka me!) can place kvs that I want to use as resource tags in AWS (mostly for cost tracking). 

The idea is that these tags are propagated to all resources descendent from the AWSCluster, such as ec2 instances and their root storage volumes.

Today, these tags make it to ec2 instances but not the root storage volumes. Only the awsmachine.spec.AdditionalTags make it to the root storage volumes. The layering of AdditionalTags is great for providing overrides (or additions) at various levels but the `AdditionalTags` from the `AWSCluster` should propagate to all resources. 

I believe this is a bug; but I could imagine someone arguing that it is a design choice - it would be a surprising one and one that creates more work for the end user (as the user would need to explicitly also set AdditionalTags on the AWSMachine manifest).

## Fix

This is an easy fix. See PR shortly. 

Here's the `ensureStorageTags` method which only pulls tags from the machine L1121.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b90c18cf1bac7308ace428245945a40e3099d758/controllers/awsmachine_controller.go#L1114-L1140

Just above its callsite, you can see that for ec2 instances the `ensureTags` method merges tags from the machine with those from the cluster through `machineScope.AdditionalTags()`.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b90c18cf1bac7308ace428245945a40e3099d758/controllers/awsmachine_controller.go#L585-L593

## Prior work

The `ensureStorageTags` method was introduced in this PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2463

philjb:
* see kubernetes-sigs/cluster-api-provider-aws#4511

**What type of PR is this?**
/kind bug

This PR gets AWSCluster AdditionalTags onto root volumes for the AWSMachines in the cluster. I believe this is the intended behavior: Set the additional tags on the cluster and have them show up on all resources for that cluster.

I believe it was a minor oversight that the current behavior isn't like this already.

Fixes #4511

None

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests: update existing unit tests
- [x] adds or updates e2e tests: none impacted

**Release note**:

```release-note
Fixed: Additional Tags specified on the AWSCluster will now propagate to root storage volumes.
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -589,7 +589,7 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		}
 
 		if instance != nil {
-			r.ensureStorageTags(ec2svc, instance, machineScope.AWSMachine)
+			r.ensureStorageTags(ec2svc, instance, machineScope.AWSMachine, machineScope.AdditionalTags())
 		}
 
 		if err := r.reconcileLBAttachment(machineScope, elbScope, instance); err != nil {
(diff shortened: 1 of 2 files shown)


philjb:
Please hold while I work through a few test failures i see locally.

philjb:
@Skarlso - back to you whenever you have time. I addressed the test failures.

cnmcavoy (approved):
lgtm, I don't see any other places that use `AWSMachine.Spec.AdditionalTags` that needs changing.

philjb:
> lgtm, I don't see any other places that use `AWSMachine.Spec.AdditionalTags` that needs changing.

I didn't either. It's just that reconcile method that touches tags for the instance and root volumes as far as i saw.

philjb:
thanks!

controllers/awsmachine_controller.go
```diff
@@ -1111,20 +1111,20 @@ func (r *AWSMachineReconciler) indexAWSMachineByInstanceID(o client.Object) []st
...
 		if subAnnotation, ok := annotations[volumeID].(map[string]interface{}); ok {
-			newAnnotation, err := r.ensureVolumeTags(ec2svc, aws.String(volumeID), subAnnotation, machine.Spec.AdditionalTags)
+			newAnnotation, err := r.ensureVolumeTags(ec2svc, aws.String(volumeID), subAnnotation, additionalTags)
 			if err != nil {
 				r.Log.Error(err, "Failed to fetch the changed volume tags in EC2 instance")
 			}
 			annotations[volumeID] = newAnnotation
 		} else {
-			newAnnotation, err := r.ensureVolumeTags(ec2svc, aws.String(volumeID), make(map[string]interface{}), machine.Spec.AdditionalTags)
+			newAnnotation, err := r.ensureVolumeTags(ec2svc, aws.String(volumeID), make(map[string]interface{}), additionalTags)
```
philjb, line 1127: this line change didn't make it into 2.2.4 release through https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4535/files

PR #4518, opened 2023-09-26
🌱 Tag S3 bucket as owned by cluster

AndiDog:
**What type of PR is this?**

/kind feature

I noticed that the S3 bucket isn't tagged. Some companies use AWS tags for grouping resources, calculating costs, or to clean up accidental remainders of clusters that CAPI/CAPA couldn't delete after a misconfiguration. With this change, the bucket is tagged as belonging to the cluster.

I'm not entirely sure if there's a use case where people provide a shared S3 bucket. In the S3 service code, there's such a hint in a `TODO`. For that case, we would need to introduce `AWSCluster.spec.s3Bucket.additionalTags` where users can set `sigs.k8s.io/cluster-api-provider-aws/cluster/...=shared`. And the implementation should be changed to only upsert tags instead of overwriting them, given that multiple `AWSCluster` objects could reconcile the same bucket. Let me know if that's a supported use case.

n/a

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Tag S3 bucket as owned by the cluster
```

### DIFF
--- a/pkg/cloud/services/s3/s3.go
+++ b/pkg/cloud/services/s3/s3.go
@@ -31,6 +31,7 @@ import (
 	"github.com/aws/aws-sdk-go/service/sts/stsiface"
 	"github.com/pkg/errors"
 
+	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	iam "sigs.k8s.io/cluster-api-provider-aws/v2/iam/api/v1beta1"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/util/system"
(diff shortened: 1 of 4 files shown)


pkg/cloud/services/s3/s3.go
```diff
@@ -230,6 +235,39 @@ func (s *Service) ensureBucketPolicy(bucketName string) error {
...
+		Tagging: &s3.Tagging{
+			TagSet: nil,
+		},
+	}
+
+	tags := infrav1.Build(infrav1.BuildParams{
+		ClusterName: s.scope.Name(),
+		Lifecycle:   infrav1.ResourceLifecycleOwned,
+		Name:        nil,
+		Role:        aws.String("node"),
```
vincepri, line 250: Is this a common role?

vincepri (commented):
One question, rest LGTM

vincepri:
/retitle 🌱 Tag S3 bucket as owned by cluster

pkg/cloud/services/s3/s3.go
AndiDog: Maybe I misunderstand the question. Do you mean to rather use `"common"` (`CommonRoleTagValue`) as we do for VPC, gateways and such? I used `"node"` because the bucket seems only used for bootstrapping purposes and therefore relates to nodes, similar to how ASGs are tagged, for example.

vincepri (approved):
/lgtm

PR #4301, opened 2023-05-31
fix: the quickstart guide needs update for external cloud provider

Skarlso:
**What type of PR is this?**
/kind bug
**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4299

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
The quick start guide is generated using this template and needs to use external provider.
```

PR #4519, opened 2023-09-26
flatcar: Set cloud-provider kubelet flag to "external"

johananl:
In-tree cloud provider support was removed in k8s v1.27.

See the following:

- https://github.com/kubernetes/kubernetes/pull/118899
- https://github.com/kubernetes/kubernetes/pull/115838
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4301

**What type of PR is this?**

/kind bug

Support for in-tree cloud providers was removed, which is currently causing errors when using `templates/cluster-template-flatcar.yaml`. Set `cloud-provider` to `external` to fix this.

None

* I'm not sure why in some places it's OK to use `cloud-provider: aws` and in others it's breaking things. See https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4301#issuecomment-1569863946.
* I'm not sure if I need to update the e2e tests.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/templates/cluster-template-flatcar.yaml
+++ b/templates/cluster-template-flatcar.yaml
@@ -45,19 +45,19 @@ spec:
       nodeRegistration:
         name: $${COREOS_EC2_HOSTNAME}
         kubeletExtraArgs:
-          cloud-provider: aws
+          cloud-provider: external
     clusterConfiguration:
       apiServer:
         extraArgs:
-          cloud-provider: aws
+          cloud-provider: external
       controllerManager:
         extraArgs:
-          cloud-provider: aws
+          cloud-provider: external
     joinConfiguration:
       nodeRegistration:
         name: $${COREOS_EC2_HOSTNAME}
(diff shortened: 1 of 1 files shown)


Ankitasw:
@johananl E2E tests doesn't cover flatcar tests, that's why this might not have been caught.
Thanks for fixing this.
The E2E tests which still uses `cloud-provider: aws` are the CSIMigration tests where we are supposed to test upgrades with intree provider to out of tree provider.
So I think if there is no other file for flatcar, this change should be ok.

johananl:
Thanks @Ankitasw!

>This is just the document change(templates) which was missed when the cloud provider migration was done.

IIUC this is more than a document change since the template files are used by `clusterctl generate` to produce workload cluster manifests (that's how I noticed the problem - my generated cluster wouldn't boot).

PR #4527, opened 2023-09-27
✨ Add AWSMachines to back the ec2 instances in AWSMachinePools 

### LINKED ISSUE #4184: Implement MachinePool Machines clusterAPI proposal

cnmcavoy:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: [ Implements the MachinePool Machines clusterAPI proposal ](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220209-machinepool-machines.md)

Fixes #4184

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Added AWSMachines to back the ec2 instances in AWSMachinePools and AWSManagedMachinePools.
```


cnmcavoy:
Still missing new tests to cover the functionality, have been testing locally with tilt

exp/controllers/awsmachinepool_controller.go
```diff
@@ -392,6 +425,163 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+func createAWSMachinesIfNotExists(ctx context.Context, awsMachineList *infrav1.AWSMachineList, mp *expclusterv1.MachinePool, existingASG *expinfrav1.AutoScalingGroup, client client.Client, ec2Svc services.EC2Interface) error {
+	log := ctrl.LoggerFrom(ctx).WithValues("machinepool", mp.Name, "namespace", mp.Namespace)
+	log.V(2).Info("create missing awsmachines")
+
+	providerIDs := make(map[string]infrav1.AWSMachine, len(awsMachineList.Items))
+	for i := range awsMachineList.Items {
+		awsMachine := awsMachineList.Items[i]
+		if *awsMachine.Spec.ProviderID != "" {
+			providerIDs[*awsMachine.Spec.ProviderID] = awsMachine
+		}
```
Skarlso, line 473: Use pointer to dereference the provider ID or this might crash with `nil`.

Also, you don't need the awsMachine extra value since you aren't using the built in for loop variable that could be overwritten. You are accessing the slice element directly.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -392,6 +425,163 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+		UID:        mp.UID,
+	}
+
+	for i := range existingASG.Instances {
+		instanceID := existingASG.Instances[i].ID
+		providerID := fmt.Sprintf("aws:///%s/%s", existingASG.Instances[i].AvailabilityZone, instanceID)
+
+		instanceLogger := log.WithValues("providerID", providerID, "instanceID", instanceID, "asg", existingASG.Name)
+		instanceLogger.V(5).Info("checking if machinepool awsmachine is up to date")
+		if _, exists := providerIDs[providerID]; !exists {
```
Skarlso, line 488: I would prefer an inverse if here saying `exists -> continue` instead of this huge if. :)

exp/controllers/awsmachinepool_controller.go
```diff
@@ -392,6 +425,163 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+		Kind:       mp.Kind,
+		Name:       mp.Name,
+		UID:        mp.UID,
+	}
+
+	for i := range existingASG.Instances {
+		instanceID := existingASG.Instances[i].ID
+		providerID := fmt.Sprintf("aws:///%s/%s", existingASG.Instances[i].AvailabilityZone, instanceID)
+
+		instanceLogger := log.WithValues("providerID", providerID, "instanceID", instanceID, "asg", existingASG.Name)
```
Skarlso, line 486: Hm, can you use the internal logger interface instead?

exp/controllers/awsmachinepool_controller.go
```diff
@@ -392,6 +425,163 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+			if err := client.Create(ctx, awsMachine); err != nil {
+				return fmt.Errorf("failed to create AWSMachine: %w", err)
+			}
+		}
+	}
+	return nil
+}
+
+func deleteOrphanedAWSMachines(ctx context.Context, awsMachineList *infrav1.AWSMachineList, existingASG *expinfrav1.AutoScalingGroup, client client.Client) error {
+	log := ctrl.LoggerFrom(ctx).WithValues("asg", existingASG.Name)
```
Skarlso, line 547: Use the internal logger instead.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -392,6 +425,163 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+				// XXX(cmcavoy): if we got here, something went wrong with the owner reference
+				if err := client.Delete(ctx, &awsMachine); err != nil {
+					return fmt.Errorf("failed to delete orphan awsMachine %s/%s: %w", awsMachine.Namespace, awsMachine.Name, err)
+				}
+				continue
+			}
+
+			if err := client.Delete(ctx, machine); err != nil {
+				return fmt.Errorf("failed to delete orphan machine %s/%s: %w", machine.Namespace, machine.Name, err)
+			}
```
Skarlso, line 579: So if I understand this correctly... 

If there is a machine for which there is no provider... check if there is an owner. If there is an owner, we delete the owner? And then we hope that cascading delete will also delete this machine?

pkg/cloud/scope/machine.go
```diff
@@ -117,6 +117,16 @@ func (m *MachineScope) IsControlPlane() bool {
 	return util.IsControlPlaneMachine(m.Machine)
 }
 
+// IsMachinePoolMachine returns true if the machine is created for a machinepool.
+func (m *MachineScope) IsMachinePoolMachine() bool {
+	for _, owner := range m.Machine.OwnerReferences {
+		if owner.Kind == "MachinePool" {
```
Skarlso, line 123: This should be in `api/constants.go`.

exp/controllers/awsmachinepool_controller.go
cnmcavoy: That is how I parsed the proposal language:

> When a MachinePool Machine is deleted manually, the system will delete the corresponding provider-specific resource. The opposite is also true: when a provider-specific resource is deleted, the system will delete the corresponding MachinePool Machine. This happens by virtue of the infrastructureRef <-> ownerRef relationship.

https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220209-machinepool-machines.md#proposal

@jont828 @mboersma @devigned wrote the proposal and perhaps can provide clarity ...?

What is the expected outcome when an AWSMachine and Machine pair, which is owned by a MachinePool, when it's ec2 instance is destroyed via an external action (e.g user terminates the instance in the aws console). Should the infrastructure provider (CAPA) detect this termination and delete the CAPI Machine resource?

Jont828: Yes, if an AWSMachine's associated provider instance is deleted, we expect the AWSMachinePool to delete the Machine and AWSMachine pair. This is so that we are not left with hanging resources in the event of an out-of-band instance deletion like you described.

Skarlso: Yes, but... if we delete its owner, is it okay that any OTHER machine this thing owns will also get deleted?

Jont828: I'm not quite sure I understand. Are you saying that if you delete the CAPI Machine, i.e. owner of the AWS Machine, you want to also delete another CAPI Machine associated with the same instance?

Skarlso: No, I mean is it a problem if the deleted owner owns more machines? Those would also be deleted then.

Jont828: When you delete the CAPI Machine, we expect it to delete the AWSMachine. It would be the same behavior we already have with MachineDeployments except there is no KubeadmConfig object to delete.

Ankitasw:
@cnmcavoy is there a final review pending for this PR, or any other items?

cnmcavoy:
@Ankitasw I think all the questions have been answered, it probably needs a new reviewer to give it a look over for a lgtm. It's ready to be merged if there are no more review items to address.

AndiDog:
I can have a look into this, @cnmcavoy – do you want to resolve conflicts first?

cnmcavoy:
> I can have a look into this, @cnmcavoy – do you want to resolve conflicts first?

Sure, I think I got them all sorted.

AndiDog (commented):
I reviewed most of the PR except the test and managed machine pool code. Sorry, I have quite a few questions and comments 😉.

Overall, this looks pretty good. Did you successfully test this already?

exp/controllers/awsmachinepool_controller.go
```diff
@@ -415,6 +448,164 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+			ObjectMeta: metav1.ObjectMeta{
+				Namespace:    mp.Namespace,
+				GenerateName: fmt.Sprintf("%s-", existingASG.Name),
+				Labels: map[string]string{
+					clusterv1.MachinePoolNameLabel: mp.Name,
+					clusterv1.ClusterNameLabel:     mp.Spec.ClusterName,
+				},
+				// Note: this AWSMachine will be owned by the MachinePool until the MachinePool controller
+				// creates its parent Machine which will adopt this resource and replace the owner reference.
+				// We set the MachinePool as a temporary owner to prevent this from becoming an orphan resource.
```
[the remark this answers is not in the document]
cnmcavoy: re: `BlockOwnerDeletion` I am not sure. I didn't set it bc this is supposed to be temporary ownership.

I'm not familiar with the CAPZ implementation. Do they have `AzureMachinePool` act as a long-lived owner, or does the ownership get passed to the `Machine` resource? I don't have a strong opinion here, I was trying to follow the specification as closely as possible.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -415,6 +448,164 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+	if err != nil {
+		return err
+	}
+	for i := range awsMachineList.Items {
+		awsMachine := awsMachineList.Items[i]
+		if !awsMachine.DeletionTimestamp.IsZero() {
+			// delete the owner Machine resource for the AWSMachine so that CAPI can clean up gracefully
+			machine, err := util.GetOwnerMachine(ctx, client, awsMachine.ObjectMeta)
+			if err != nil {
+				clusterScope.Warn("failed to get owner machine", "awsmachine", awsMachine, "machinepool", mp.Name, "namespace", mp.Namespace)
```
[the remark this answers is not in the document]
cnmcavoy: It could have no owner for legitimate reasons, e.g if the awsmachine was just provisioned by an ASG, CAPA reconciled and saw it, created an AWSMachine. Meanwhile, CAPI hasn't reconciled and hasn't yet created a machine for the new instance.

I _suppose_ we could distinguish between the two. As if it's the above scenario, the owner reference will be the AWSMachinePool. Then we could error and leave it in place if it was in some bizarre unknown state. Do you think it's worth the effort?

exp/controllers/awsmachinepool_controller.go
```diff
@@ -298,6 +311,23 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 		return nil
 	}
 
+	awsMachineList, err := getAWSMachines(ctx, machinePoolScope.MachinePool, r.Client)
```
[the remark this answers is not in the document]
cnmcavoy: We don't, you can only get lucky with the current implementation. If you reconcile right after the instance was provisioned in the cloud provider.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -415,6 +448,164 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
 }
 
+func reconcileDeleteAWSMachines(ctx context.Context, mp *expclusterv1.MachinePool, clusterScope cloud.ClusterScoper, client client.Client) error {
+	awsMachineList, err := getAWSMachines(ctx, mp, client)
+	if err != nil {
+		return err
+	}
+	for i := range awsMachineList.Items {
+		awsMachine := awsMachineList.Items[i]
+		if !awsMachine.DeletionTimestamp.IsZero() {
```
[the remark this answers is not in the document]
cnmcavoy: The `AWSMachine` resources are deleted via cascade from their owners, the `Machine`. When a `Machine` is deleted, it's dependents are also deleted.  The lifecycle is documented in the RFC: https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20220209-machinepool-machines.md#machine-controller

controllers/awsmachine_controller.go
```diff
@@ -488,6 +490,12 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		conditions.MarkUnknown(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, err.Error())
 		return ctrl.Result{}, err
 	}
+	if instance == nil && machineScope.IsMachinePoolMachine() {
+		err = errors.New("no instance found for machine pool")
+		machineScope.Error(err, "unable to find instance")
+		conditions.MarkUnknown(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceNotFoundReason, err.Error())
```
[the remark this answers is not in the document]
cnmcavoy: I do not recall being spammed with errors, but I last tested this cod in a cluster 6 months ago.

exp/controllers/awsmachinepool_controller.go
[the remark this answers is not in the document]
cnmcavoy: So I recalled why I chose the cluster scope. If we use the machinePoolScope here, the `AWSManagedMachinePool` controller can not re-use and call these methods. The `ManagedMachinePool` type is not a compatible interface.

What do you propose to do about that? Do we need to have duplicate code for all these methods in the `AWSManagedMachinePool` controller? I had hoped to avoid that.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -415,6 +448,164 @@ func (r *AWSMachinePoolReconciler) reconcileDelete(machinePoolScope *scope.Machi
...
+		instanceID := existingASG.Instances[i].ID
+		providerID := fmt.Sprintf("aws:///%s/%s", existingASG.Instances[i].AvailabilityZone, instanceID)
+
+		clusterScope.Trace("checking if machinepool awsmachine is up to date", "providerID", providerID, "instanceID", instanceID, "asg", existingASG.Name)
+		if _, exists := providerIDs[providerID]; exists {
+			continue
+		}
+
+		instance, err := ec2Svc.InstanceIfExists(&instanceID)
+		if err != nil {
```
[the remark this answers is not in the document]
cnmcavoy: I added a guard clause to try to handle this:
```
		if errors.Is(err, ec2.ErrInstanceNotFoundByID) {
			instanceLogger.V(5).Info("instance not found, it may have already been deleted.")
			continue
		}
```

AndiDog:
As an update: I'll try to test this major feature. PR looks mostly fine, I think. I'm only unsure about the owner reference in the non-happy case.

AndiDog:
I got a working test environment, and the `AWSMachine`/`Machine` objects get created correctly 🍀.

First blocking issue I found: deletion by downscaling the ASG/AWSMachinePool leads to this permanent error rather than deleting the `AWSMachine`:

```
  failureMessage: EC2 instance state "terminated" is unexpected
  failureReason: UpdateError
  instanceState: terminated
  ready: false
```

Other than that, the `Node` object went away and the cluster reacted according to the node deletion. @cnmcavoy what happened when you tested this? And which other cases do you think should be tested? I guess cluster-autoscaler scaling of ASGs may be a good external trigger that CAPA should have no problem with – I can test that.

cnmcavoy:
@AndiDog thanks for taking a look. I think that's slightly concerning bc there should be test cases covering that sort of condition.

I suspect this PR has been left too long and the code has rotted. Indeed doesn't use ASGs anymore for autoscaling, so I can't really justify the time investment it would take to cut a new PR and start this fresh.

PR #4529, opened 2023-09-28
flatcar: Run kubeadm after containerd

johananl:
**What type of PR is this?**

/kind bug

Ensure kubeadm runs after containerd to avoid a race condition between the two.

We intentionally add an `After=` directive and not a `Requires=` directive to avoid breaking things for distros which use Ignition and don't use containerd.

See https://github.com/kubernetes-sigs/image-builder/issues/939.

None

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Ensure kubeadm runs after containerd on Flatcar
```


Ankitasw:
/lgtm
/approve

Thanks @johananl

PR #4541, opened 2023-10-04
✨ Add AWSMachine fields to control vpc placement for the instance

### LINKED ISSUE #4540: Ability to spawn machines in an external vpc to the cluster vpc
/kind feature

As a platform administrator that uses CAPA, I'd like to be able to define and declare a cluster that lives in VPC "foo" and spawn machines in subnets under that VPC, as well as spawn machines in a VPC "bar", which has it's own separate subnets. This is useful in cases where the subnet space of the cluster is limited, or in scenarios where the VPCs have differing network configurations with regards to internet access.

We can assume VPC "foo" and VPC "bar" have peering, a transit gateway, or some other way to crosstalk.

This is currently not possible today, even if you specify an AWSMachine with a subnet in another VPC, using the existing `subnet` field, because the VPC ID is specified when filtering subnets.

**Environment:**

- Cluster-api-provider-aws version: 2.2.1
- Kubernetes version: (use `kubectl version`): 1.24.14
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04.6 LTS

cnmcavoy:
**What type of PR is this?**
/kind feature

Adds fields to the AWSMachine and AWSMachineTemplate resources to allow VPC selection, as well as overriding the security groups chosen for various infrastructure roles.

Fixes #4540

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
 Added AWSMachine and AWSMachineTemplate fields to control vpc placement for individual instances
```

### DIFF
--- a/api/v1beta2/awsmachine_types.go
+++ b/api/v1beta2/awsmachine_types.go
@@ -109,10 +109,20 @@ type AWSMachineSpec struct {
 	// +optional
 	AdditionalSecurityGroups []AWSResourceReference `json:"additionalSecurityGroups,omitempty"`
 
+	// VPC is a reference to the VPC to use when picking a subnet to use for this
+	// instance. Only valid if the subnet (id or filters) is also specified.
+	// +optional
+	VPC *AWSResourceReference `json:"vpc,omitempty"`
+
 	// Subnet is a reference to the subnet to use for this instance. If not specified,
 	// the cluster subnet will be used.
 	// +optional
 	Subnet *AWSResourceReference `json:"subnet,omitempty"`
+
+	// SecurityGroupOverrides is an optional set of security groups to use for the node.
+	// This is optional - if not provided security groups from the cluster will be used.
+	// +optional
+	SecurityGroupOverrides map[SecurityGroupRole]string `json:"securityGroupOverrides,omitempty"`
 
 	// SSHKeyName is the name of the ssh key to attach to the instance. Valid values are empty string (do not use SSH keys), a valid SSH key name, or omitted (use the default SSH key name)
 	// +optional
--- a/api/v1beta2/awsmachine_types.go
+++ b/api/v1beta2/awsmachine_types.go
@@ -109,11 +109,6 @@ type AWSMachineSpec struct {
 	// +optional
 	AdditionalSecurityGroups []AWSResourceReference `json:"additionalSecurityGroups,omitempty"`
 
-	// VPC is a reference to the VPC to use when picking a subnet to use for this
-	// instance. Only valid if the subnet (id or filters) is also specified.
-	// +optional
-	VPC *AWSResourceReference `json:"vpc,omitempty"`
-
 	// Subnet is a reference to the subnet to use for this instance. If not specified,
 	// the cluster subnet will be used.
 	// +optional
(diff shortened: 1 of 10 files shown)


cnmcavoy:
Some verification of this I did with one of our clusters:
Screenshot 2023-10-05 at 12-56-19 Instances EC2 us-east-2
Screenshot 2023-10-05 at 12-57-49 awscmhinfra3 Clusters Elastic Kubernetes Service us-east-2
Screenshot 2023-10-05 at 12-58-46 awscmhinfra3 Clusters Elastic Kubernetes Service us-east-2

api/v1beta2/awsmachine_types.go
```diff
@@ -109,11 +109,21 @@ type AWSMachineSpec struct {
 	// +optional
 	AdditionalSecurityGroups []AWSResourceReference `json:"additionalSecurityGroups,omitempty"`
 
+	// VPC is a reference to the VPC to use when picking a subnet to use for this
+	// instance. Only valid if the subnet (id or filters) is also specified.
+	// +optional
+	VPC *AWSResourceReference `json:"vpc,omitempty"`
```
vincepri, line 115: Reading the related use case, I was wondering if we should be able to infer the VPC id from the subnet identifier? 

What could confuse a bit users with the changes proposed here is that we have `AdditionalSecurityGroups` and now `SecurityGroupOverrides`, which might conflict with each other?

cnmcavoy: > Reading the related use case, I was wondering if we should be able to infer the VPC id from the subnet identifier?

I think this is possible, it would require rewriting the existing logic to always look up the VPC, but it does have the advantage of simplifying the fields for the AWSMachine. I'll take a look at doing this.

> What could confuse a bit users with the changes proposed here is that we have AdditionalSecurityGroups and now SecurityGroupOverrides, which might conflict with each other?

I agree there is some confusion because of the two fields for managing the same concept in AWS. The problem I tried to solve is that we need a different set of "default" SGs in the other VPC, because SGs are _bound_ to a particular VPC, and we can't re-use the cluster default SGs. So I provided a way for the user to supply their own. `AdditionalSecurityGroups` is insufficient because we need default SGs that are present in the secondary VPC. I don't care about `SecurityGroupOverrides` as a field, except as a way to provide SGs for the secondary VPC.

So the other option is we could eliminate the `SecurityGroupOverrides` field and instead have the CAPA duplicate the SGs from the cluster VPC into any secondary VPC, and use those for these nodes. But there would also be no good way to tell when we could tear down these SGs, without tracking them at the cluster level too. We'd probably have to add them to the AWSCluster or AWSManagedCluster's status field to keep track. Otherwise they might get left dangling after a cluster is torn down. Can you think of a better way to make this work?

cnmcavoy: @vincepri updated to look up the VPC as needed, if you want to give it another look.

Ankitasw (commented):
@cnmcavoy Would it be possible to add documentation for relevant changes in the CAPA book?

vincepri (approved):
/lgtm

cnmcavoy:
Rebased to fix tests.

Ankitasw:

/hold

@cnmcavoy please feel free to unhold once tests passes

Issue #4333, opened 2023-06-12
Ability to specify Availability Zones from the `VPCSpec` object


synthe102:
/kind feature

**Describe the solution you'd like**

Currently the AZ selection can be done using `availabilityZoneUsageLimit` and `availabilityZoneSelection`. However, there is no direct way to provide a list of availability zones to use.
I'd like a field that takes a list of availability zone names (or zone IDs) and create the VPC and subnets only in those availability zones.

**Environment:**

- Cluster-api-provider-aws version: 2.1.4
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

Skarlso:
Since availability zones are per region, defining a list of availability zones might not work in some of the regions.

Question... @synthe102 Would you expect the two settings of confinement ( AvailabilityZoneUsageLimit, AvailabilityZoneSelection ) to still work, or for them to be completely ignored if the list if defined ( thereby accepting the consequences that things might start to fail ).

Skarlso:
I'm going to assume for now that ignoring the other two if this one is set, is okay and the user accepts the consequences.

synthe102:
@Skarlso it makes sense for me to ignore those and make the user assume the consequences

yosefmih:
@Ankitasw came across this issue while investigating this error on the `AWSManagedControlPlane`. 

>  "Cannot create cluster 'x' because EKS does not support creating control plane instances in us-east-1c, the targeted availability zone. Retry cluster creation using control plane subnets that span at least two of these availability zones: us-east-1a, us-east-1b, us-east-1d, us-east-1e, us-east-1f. Note, post cluster creation, you can run worker nodes in separate subnets/availability zones from control plane subnets/availability zones passed during cluster creation"

According to AWS [docs](https://docs.aws.amazon.com/eks/latest/userguide/network-reqs.html), some zones can't be used for control planes. We were unlucky enough to have `us-east-1c` be mapped to one of these zones. What is the status on this issue? Happy to work on a fix as this is a blocking issue for us rn.

AndiDog:
PR needs a rebase to continue

/triage accepted

PR #4543, opened 2023-10-05
✨ feat: create vpc objects in explicitly provided availability zones

Skarlso:
**What type of PR is this?**

/kind feature

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4333#

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
 Ability to specify Availability Zones from the VPCSpec object.
```

### DIFF
--- a/api/v1beta2/network_types.go
+++ b/api/v1beta2/network_types.go
@@ -341,6 +341,11 @@ type VPCSpec struct {
 	// +kubebuilder:validation:Enum=Ordered;Random
 	AvailabilityZoneSelection *AZSelectionScheme `json:"availabilityZoneSelection,omitempty"`
 
+	// AvailabilityZones defines a list of Availability Zones in which to create network resources in.
+	// If defined, both AvailabilityZoneUsageLimit and AvailabilityZoneSelection are ignored.
+	// +optional
+	AvailabilityZones []string `json:"availabilityZones,omitempty"`
+
 	// EmptyRoutesDefaultVPCSecurityGroup specifies whether the default VPC security group ingress
 	// and egress rules should be removed.
 	//
--- a/pkg/cloud/services/network/subnets_test.go
+++ b/pkg/cloud/services/network/subnets_test.go
@@ -2316,6 +2316,146 @@ func TestReconcileSubnets(t *testing.T) {
 					After(zone2PrivateSubnet)
 			},
 		},
+		{
+			name: "Managed VPC, no existing subnets exist, one az is explicitly defined, expect one private and one public from default",
+			input: NewClusterScope().WithNetwork(&infrav1.NetworkSpec{
+				VPC: infrav1.VPCSpec{
+					ID: subnetsVPCID,
+					Tags: infrav1.Tags{
+						infrav1.ClusterTagKey("test-cluster"): "owned",
+					},
+					CidrBlock:         defaultVPCCidr,
+					AvailabilityZones: []string{"us-east-1b"},
+				},
+				Subnets: []infrav1.SubnetSpec{},
+			}),
+			expect: func(m *mocks.MockEC2APIMockRecorder) {
+				describeCall := m.DescribeSubnetsWithContext(context.TODO(), gomock.Eq(&ec2.DescribeSubnetsInput{
+					Filters: []*ec2.Filter{
+						{
+							Name:   aws.String("state"),
+							Values: []*string{aws.String("pending"), aws.String("available")},
+						},
+						{
+							Name:   aws.String("vpc-id"),
+							Values: []*string{aws.String(subnetsVPCID)},
+						},
+					},
+				})).Return(&ec2.DescribeSubnetsOutput{}, nil)
+
+				m.DescribeRouteTablesWithContext(context.TODO(), gomock.AssignableToTypeOf(&ec2.DescribeRouteTablesInput{})).
+					Return(&ec2.DescribeRouteTablesOutput{}, nil)
+
+				m.DescribeNatGatewaysPagesWithContext(context.TODO(),
+					gomock.Eq(&ec2.DescribeNatGatewaysInput{
+						Filter: []*ec2.Filter{
+							{
+								Name:   aws.String("vpc-id"),
+								Values: []*string{aws.String(subnetsVPCID)},
+							},
+							{
+								Name:   aws.String("state"),
+								Values: []*string{aws.String("pending"), aws.String("available")},
+							},
+						},
+					}),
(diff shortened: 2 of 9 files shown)


Skarlso:
Oh totally forgot all the conversion shenanigans.

api/v1beta2/network_types.go
```diff
@@ -323,6 +323,11 @@ type VPCSpec struct {
 	// +kubebuilder:default=Ordered
 	// +kubebuilder:validation:Enum=Ordered;Random
 	AvailabilityZoneSelection *AZSelectionScheme `json:"availabilityZoneSelection,omitempty"`
+
+	// AvailabilityZoneList defines a list of Availability Zones in which to create network resources in.
+	// If defined, both AvailabilityZoneUsageLimit and AvailabilityZoneSelection are ignored.
+	// +optional
+	AvailabilityZoneList []string `json:"availabilityZoneList,omitempty"`
```
vincepri, line 330:
```suggestion
	// AvailabilityZoneList defines a list of Availability Zones in which to create network resources in.
	// Cannot be defined alongside AvailabilityZoneUsageLimit and AvailabilityZoneSelection.
	// +optional
	AvailabilityZones []string `json:"availabilityZones,omitempty"`
```

We should probably refuse usage when both fields are set, that can cause confusion. We could check in a validation webhook, and given that this field is new, the change is backward compatible.

Skarlso: Sounds good. 👍

Skarlso: Actually, both fields have mandatory defaults. They are always defined. Meaning we cannot differentiate between deliberately set or the default value defined. Which means we cannot have a validation against them.

vincepri: We could move the defaults in the defaulting webhook instead of openapi schema? That way, the defaulting logic can be conditional.

Skarlso: I'm not a 100% sure how that works. It wouldn't break existing clusters, right? Like, the default in the kubebuilder results in a different yaml, while the defaulting webhook... doesn't? Or does it?

cnmcavoy: The defaulting webhook is translated into a mutating admission webhook by kubebuilder. New resources + resource updates will be seen by the defaulting webhook, mutated, and stored in the cluster after the mutation, but _existing_ resources will not be re-examined or updated. 

This can be a problem when the new field being introduced needs to be set to a non-nil default value for existing resources. However, that is not the case here...

So I think because the default behavior of the field is `nil` defaulting would work as intended, because existing resources are in the correct state, because none of them want to use this new functionality.

cnmcavoy (approved):

api/v1beta2/network_types.go
Skarlso: I'm not talking about defaulting for my new API I'm talking about defaulting for existing ones. I think that's a change of behaviour I would advice against.

vincepri: We already have defaulting and validation webhooks in place, are you referring to something else?

Skarlso: I'm talking about what you suggested. Where we switch the two others from using kubebuilder defaults to the validating webhook / defaulting webhook. That would be a change in behaviour I think?

vincepri: No it's not a change in behavior; the webhooks are considered part of the codebase, and so is the same the openapi/kubebuilder tags. This would move the validation from one place to the other; it's also brand new validation

Skarlso: Also... Okay. So... Kubebuilder Markers GENERATE default values so the generated output will be different. But mutating admission webhooks change values in flight so the output OBJECT's value will be different.

Unless I'm super mistaken these two scenarios ARE different.

vincepri: The yaml is going to be different, although the defaulting logic moves to the webhook, which historically we have not treated as a breaking change, but as a compatible one.

vincepri: That said, what I was suggesting is to validate this new field in the webhook, and reject it when the other two colliding fields are set.

Skarlso: Yes, but validating it would require removing the defaults from here. And that's what started this conversation. :)

vincepri: Yeah that's not a breaking change, @fabriziopandini @sbueringer from the upstream project can keep me honest though.

richardcase:
#4575  is probably needed for the eks e2e to pass.

pkg/cloud/services/network/subnets_test.go
```diff
@@ -2316,6 +2316,146 @@ func TestReconcileSubnets(t *testing.T) {
...
+					},
+				})).Return(&ec2.CreateSubnetOutput{
+					Subnet: &ec2.Subnet{
+						VpcId:               aws.String(subnetsVPCID),
+						SubnetId:            aws.String("subnet-2"),
+						CidrBlock:           aws.String("10.0.128.0/17"),
+						AvailabilityZone:    aws.String("us-east-1b"),
+						MapPublicIpOnLaunch: aws.Bool(false),
+					},
+				}, nil).After(zone1PublicSubnet)
```
AndiDog, line 2453: [minor]

Isn't it an implementation detail in which order private and public subnets are created i.e. we shouldn't check for `private.After(public)` here?

AndiDog (commented):
LGTM except one minor remark

pkg/cloud/services/network/subnets_test.go
Skarlso: I don't think it matters that much here? I copied this entire section from previous tests. :) So whatever they do this one does as well. :)

AndiDog: Yeah, not a blocker at all

synthe102:
Hey !
Thanks a lot for doing the implementation of this feature. Could it be possible to have it merged in the coming month?

Skarlso:
I can rebase and the rest is up to the capa team. :)

Skarlso:
Interesting, I ran `generate`

vincepri:
The only thing left in this PR is to adjust the webhook validation like mentioned above

Skarlso:
Yeah, not gonna do that. Someone else can pick this up.

synthe102:
I picked up where @Skarlso stopped and added the validating and mutating webhook changes @vincepri requested here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4950

PR #4575, opened 2023-10-12
fix: eks e2e test failing as subnet name was used

richardcase:
**What type of PR is this?**

/kind bug
/kind regression

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #4574 

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits

**Release note**:

```release-note
Fix issue with EKE e2e tests due to a issue when specifying which subnets to use.
```

PR #4589, opened 2023-10-25
Fix launch template reconciliation if bootstrap data secret cannot be read

AndiDog:
**What type of PR is this?**

/kind bug

When changing the machine pool's bootstrap config, e.g. `KubeadmConfig/{a => b}`, which is supported after CAPI's fix https://github.com/kubernetes-sigs/cluster-api/pull/8667, CAPA had a bug where it would continue even after the new bootstrap secret cannot be read, for instance because it doesn't exist yet as the bootstrap provider still needs a few seconds to make it ready. That could lead to a catastrophic misconfiguration where the launch template's user data is empty, typically leading to new instances not getting bootstrapped at all. Detailed problem description in [this comment](https://github.com/kubernetes-sigs/cluster-api/issues/8858#issuecomment-1737930898).

This provides the obvious fix. I couldn't find a great place to test this, since `ReconcileLaunchTemplate` is mocked away in tests.

I tested this successfully by changing the bootstrap config reference and observing how CAPI sets the new data secret name, CAPA intermittently failing to read the new secret until it's ready, and only then CAPA writing a new launch template version with the new bootstrap secret's data.

Part of https://github.com/kubernetes-sigs/cluster-api/issues/8858, but not a fix. Even with this patch, changing the bootstrap config object doesn't lead to an instance refresh (rolling out new nodes) because CAPA doesn't consider it a change. We need to discuss further how to solve the main issue.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fix launch template reconciliation if bootstrap data secret cannot be read
```

### DIFF
--- a/pkg/cloud/services/ec2/launchtemplate.go
+++ b/pkg/cloud/services/ec2/launchtemplate.go
@@ -59,6 +59,7 @@ func (s *Service) ReconcileLaunchTemplate(
 	bootstrapData, err := scope.GetRawBootstrapData()
 	if err != nil {
 		record.Eventf(scope.GetMachinePool(), corev1.EventTypeWarning, "FailedGetBootstrapData", err.Error())
+		return err
 	}
 	bootstrapDataHash := userdata.ComputeHash(bootstrapData)
 
--- a/pkg/cloud/scope/machinepool.go
+++ b/pkg/cloud/scope/machinepool.go
@@ -167,7 +167,7 @@ func (m *MachinePoolScope) getBootstrapData() ([]byte, string, error) {
 	key := types.NamespacedName{Namespace: m.Namespace(), Name: *m.MachinePool.Spec.Template.Spec.Bootstrap.DataSecretName}
 
 	if err := m.Client.Get(context.TODO(), key, secret); err != nil {
-		return nil, "", errors.Wrapf(err, "failed to retrieve bootstrap data secret for AWSMachine %s/%s", m.Namespace(), m.Name())
+		return nil, "", errors.Wrapf(err, "failed to retrieve bootstrap data secret %s for AWSMachine %s/%s", key.Name, m.Namespace(), m.Name())
 	}
 
 	value, ok := secret.Data["value"]


pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -59,6 +59,7 @@ func (s *Service) ReconcileLaunchTemplate(
 	bootstrapData, err := scope.GetRawBootstrapData()
 	if err != nil {
 		record.Eventf(scope.GetMachinePool(), corev1.EventTypeWarning, "FailedGetBootstrapData", err.Error())
+		return err
```
richardcase, line 62: Its probably worth logging as well here.

AndiDog: In the whole function, errors are only returned, not logged. The callers then log it like this:

```
if err := ec2Svc.ReconcileLaunchTemplate(machinePoolScope, canUpdateLaunchTemplate, runPostLaunchTemplateUpdateOperation); err != nil {
	r.Recorder.Eventf(machinePoolScope.AWSMachinePool, corev1.EventTypeWarning, "FailedLaunchTemplateReconcile", "Failed to reconcile launch template: %v", err)
	machinePoolScope.Error(err, "failed to reconcile launch template")
	return err
}
```

I prefer keeping it this way since we don't get two log lines about the same error.

richardcase: Sounds good to me.

cnmcavoy (approved):
/lgtm

AndiDog:
@richardcase The test `pull-cluster-api-provider-aws-e2e-eks` was waiting for an EKS cluster to be created and the result here in GitHub says "Pod got deleted unexpectedly". Do you have an idea for this issue? Can it be because of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4575 not merged into my PR?

AndiDog:
Trying the test again after having merged `main` (incl. possible test fix https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4575)

AndiDog:
@cnmcavoy @faiq I merged `main` and the E2E test worked this time. Switched to rebase since merge isn't allowed. Need another LGTM – no changes happened to my diff.

faiq (approved):

PR #4590, opened 2023-10-25
Fix the backport of #4512 into release-2.2 branch

### LINKED ISSUE #4511: AWSCluster AdditionalTags are not propagated to AWSMachine storage volumes

cnmcavoy:
**What type of PR is this?**
/kind regression

**What this PR does / why we need it**: Regression from main branch, backport of #4512 into release-2.2 failed and introduced errors and missed adding a unit test. See conversation on slack by @philjb : https://kubernetes.slack.com/archives/CD6U2V71N/p1697658532510839 

Fixes #4511 

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Fixed: Additional Tags specified on the AWSCluster will now propagate to root storage volumes.
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -1126,7 +1126,7 @@ func (r *AWSMachineReconciler) ensureStorageTags(ec2svc services.EC2Interface, i
 			}
 			annotations[volumeID] = newAnnotation
 		} else {
-			newAnnotation, err := r.ensureVolumeTags(ec2svc, aws.String(volumeID), make(map[string]interface{}), machine.Spec.AdditionalTags)
+			newAnnotation, err := r.ensureVolumeTags(ec2svc, aws.String(volumeID), make(map[string]interface{}), additionalTags)
 			if err != nil {
 				r.Log.Error(err, "Failed to fetch the changed volume tags in EC2 instance")
 			}
--- a/controllers/awsmachine_controller_unit_test.go
+++ b/controllers/awsmachine_controller_unit_test.go
@@ -481,7 +481,7 @@ func TestAWSMachineReconciler(t *testing.T) {
 					}
 				})
 
-				t.Run("should tag instances from machine and cluster tags", func(t *testing.T) {
+				t.Run("should tag instances and volumes with machine and cluster tags", func(t *testing.T) {
 					g := NewWithT(t)
 					awsMachine := getAWSMachine()
 					setup(t, g, awsMachine)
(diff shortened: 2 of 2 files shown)


philjb:
@richardcase @cnmcavoy  - how does this work now? Is there a new release of 2.2? aka 2.2.5?

Thanks for making this fix - I'd been meaning to do it. I just yesterday double checked the behavior of `main` and the 2.2.4 release. 2.2.4 with only the partial cherry pick is broken still with regards to #4512. The code in this PR should fix pushing tags to storage volumes on the 2.2 branch line. 

Same thing said differently, based on my (local) testing yesterday, #4512 should not be added to 2.2.4 release notes as the issue isn't fixed there. This pr should fix it in the next release of 2.2.

philjb (approved):
looks good to me. 

I tested (tilt based test setup) that someone with a cluster with CAPA 2.2.4 can upgrade to this patch (future 2.2.5) and it works properly. I was concerned that the prevAnnotation annotation might not push tag updates through after upgrades but it does. 

EDIT: for more context, `prevAnnotations` isn't even used when updating tags (which is good - prevents issues with an annotations state being wrong - but it doesn't mean more API tag requests - on every reconcile i believe); `prevAnnotations` is used when deciding when to update the annotation which says which tags were last applied.

richardcase:
@philjb - once this merges we can thinking about doing a v2.2.5 release.

richardcase:
We just need some to review and provide the lgtm

AndiDog (commented):
/lgtm

PR #4604, opened 2023-10-31
Filter out reconciling events on externally managed AWSManagedClusters

fiunchinho:
**What type of PR is this?**

/kind bug

Users may specify that some AWS resources should not be managed by CAPA controllers, because they may be managed by a different mechanism. We expose the `cluster.x-k8s.io/managed-by: "<name-of-system>"` annotation, so that users can specify which resources should not be reconciled by capa controllers. [We use it for our `AWSCluster` controller](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awscluster_controller.go#L385), but we don't use it for our `AWSManagedCluster` controller. I believe it should also use it.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Filter out reconciling events on externally managed AWSManagedClusters
```

### DIFF
--- a/controllers/awsmanagedcluster_controller.go
+++ b/controllers/awsmanagedcluster_controller.go
@@ -125,6 +125,7 @@ func (r *AWSManagedClusterReconciler) SetupWithManager(ctx context.Context, mgr 
 		WithOptions(options).
 		For(awsManagedCluster).
 		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(ctrl.LoggerFrom(ctx), r.WatchFilterValue)).
+		WithEventFilter(predicates.ResourceIsNotExternallyManaged(log.GetLogger())).
 		Build(r)
 
 	if err != nil {


dlipovetsky:
Just for my own understanding, is this the use case?

I have an EKS control plane that I manage externally, but I want CAPA to manage the EKS worker machines.

fiunchinho:
> Just for my own understanding, is this the use case?
> 
> I have an EKS control plane that I manage externally, but I want CAPA to manage the EKS worker machines.

Yes I think so. CAPA won't reconcile that CR, but if you add `MachineDeployments` or `MachinePools` that don't contain the annotation, those will be reconciled normally by CAPA.

dlipovetsky:
Since we already do this for unmanaged clusters, this seems reasonable.

/approve

fiunchinho:
Can you take a look @faiq @cnmcavoy ? Thanks!

cnmcavoy (approved):
/lgtm

Issue #4618, opened 2023-11-06
MachinePool remains in WaitingForReplicasReady because CAPA does not reconcile node references after instance refresh


AndiDog:
/kind bug

**What steps did you take and what happened:**

Related to https://github.com/kubernetes-sigs/cluster-api/issues/8858, https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4071

CAPA's `AWSMachinePool` reconciler unconditionally returns `return ctrl.Result{}, r.reconcileNormal(ctx, machinePoolScope, infraScope, infraScope)`, i.e. does not schedule reconciliation of the ASG's EC2 instances into `.Status.Instances` at regular intervals.

I made a change where CAPA triggers an instance refresh (e.g. change of AMI IDs), rolling out new EC2 instances. The parent `MachinePool` object remained in non-ready state with reason `WaitingForReplicasReady`, with CAPI continuously logging `NodeRefs != ReadyReplicas` messages. Only the next, random reconciliation of my `AWSMachinePool` object solves this by checking which instances exist in the ASG.

**What did you expect to happen:**

CAPA should regularly reconcile in order to check the ASG for a changed set of instances. Particularly if it's _expected_ because CAPA triggered an instance refresh.

**Environment:**

- Cluster-api-provider-aws version: v2.2.4 + preliminary fixes from https://github.com/kubernetes-sigs/cluster-api/issues/8858 which trigger instance refresh on change of the bootstrap config reference
- Kubernetes version: (use `kubectl version`): v1.24.14
- OS (e.g. from `/etc/os-release`): Flatcar Linux

cnmcavoy:
In your opinion, is this resolved by the MachinePools Machines implementation in CAPA? If not, what is missing that would need to be added? 

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4527

AndiDog:
@cnmcavoy I think your PR is separate from this issue. CAPI reconciles based on `AWSMachinePool.Spec.ProviderIDList` (unless you tell me that will change once infra providers create `<Infra>Machine` objects for machine pools?). That field is already correctly updated in `awsmachinepool_controller.go`, but CAPA does not regularly update it when the ASG (or an explicit instance refresh) creates/rolls instances.

cnmcavoy:
[...]

Issue #4071, opened 2023-02-15
Updating MachinePool, AWSMachinePool, and KubeadmConfig resources does not trigger an ASG instanceRefresh


wmgroot:
/kind bug

**What steps did you take and what happened:**
Updating the MachinePool, AWSMachinePool or KubeadmConfig resources does not trigger an instanceRefresh on the AWS ASG.

I expect that with the `awsmachinepool.spec.refreshPreferences.disable` left on it's default value or `false`, that changes to the MachinePool, AWSMachinePool, and KubeadmConfig would automatically trigger an instance refresh to rotate nodes in the pool to use the updated settings. Currently, I must manually start instance refreshes using the AWS UI or CLI in order for instances to be replaced when my specs change.

**What did you expect to happen:**
These are the MachinePool, AWSMachinePool, and KubeadmConfig I'm working with.
```
---
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachinePool
metadata:
  annotations:
    cluster.x-k8s.io/replicas-managed-by: external-autoscaler
  name: worker-private-efficient
spec:
  clusterName: awscmhdev2
  replicas: 2
  template:
    metadata: {}
    spec:
      bootstrap:
        configRef:
          apiVersion: bootstrap.cluster.x-k8s.io/v1beta1
          kind: KubeadmConfig
          name: worker-private-efficient
      clusterName: awscmhdev2
      infrastructureRef:
        apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
        kind: AWSMachinePool
        name: worker-private-efficient
      version: v1.24.7
---
apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
kind: AWSMachinePool
metadata:
  annotations:
    cluster-api-provider-aws: "true"
  name: worker-private-efficient
spec:
  additionalTags:
    k8s.io/cluster-autoscaler/awscmhdev2: owned
    k8s.io/cluster-autoscaler/enabled: "true"
  awsLaunchTemplate:
    additionalSecurityGroups:
    - filters:
      - name: tag:Name
        values:
        - capi-hostports
      - name: tag:network-zone
        values:
        - qa
      - name: tag:region
        values:
        - us-east-2
    ami:
      id: ami-0f6b6efcd422b9d85
    iamInstanceProfile: nodes.cluster-api-provider-aws.sigs.k8s.io
    instanceType: c5a.2xlarge
    rootVolume:
[...]

PR #4619, opened 2023-11-06
✨ Trigger machine pool instance refresh (node rollout) if bootstrap config reference changes

AndiDog:
**What type of PR is this?**

/kind feature

We discussed this and alternative solutions at length in https://github.com/kubernetes-sigs/cluster-api/issues/8858 and CAPI/CAPA office hours. Eventually, we turned to a solution that is not surprising for users: if the bootstrap config reference (e.g. `KubeadmConfig`) changes within a `MachinePool`/`AWSMachinePool` combo, i.e. its name changes, new nodes should be rolled out. With this, it is not necessary to parse the bootstrap data (= user data) and detect changes within those scripts (cloud-init, ignition, maybe other formats in the future). Instead, we only need to store the previously-used bootstrap data secret key (namespace+name) in the launch template, and compare on every reconciliation whether that reference changed. For example, if an operating user wants to roll out a change to `KubeadmConfig.spec.files`, i.e. add new files on the nodes, it's now sufficient to create a newly-named `KubeadmConfig`. For my specific usage, this means naming the object `KubeadmConfig/my-config-<hash of its spec>` and the Helm chart upgrade will take care to deploy this new object and delete the old one (which I tested and works perfectly fine).

Testing of this change showed that CAPI+CAPA correctly, and in the correct order without possible race conditions (🤞), render the bootstrap data secret, mark the bootstrap config ready, create a new launch template version and trigger instance refresh (which creates new EC2 instances). The only minor glitch I found is https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4618, but that can be solved separately.

Since reconciliation of launch templates wasn't covered in tests at all, I changed the mock interfaces a little so it could be tested, and added some test cases.

Towards https://github.com/kubernetes-sigs/cluster-api/issues/8858 and https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4071

For the feature to work, users require a CAPI version that includes https://github.com/kubernetes-sigs/cluster-api/pull/8667. Without that, CAPA won't see an updated `MachinePool.Spec.Template.Spec.Bootstrap.DataSecretName` to reconcile.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Trigger machine pool instance refresh (node rollout) if bootstrap config reference changes
```


pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -845,19 +873,24 @@ func (s *Service) buildLaunchTemplateTagSpecificationRequest(scope scope.LaunchT
 		Additional:  additionalTags,
 	})
 
-	if len(tags) > 0 {
-		// tag instances
+	// tag instances
+	{
```
fiunchinho, line 877: Do we use blocks like `{ }` on other parts of the code? I'm wondering if we want to use it, or if it improves things. If we don't use it anywhere else, I wouldn't use it.

cnmcavoy (commented):
Take a look at the linter errors failing the tests. Overall this looks solid.

pkg/cloud/services/ec2/launchtemplate.go
AndiDog: If there are zero tags per category (here: `ec2.ResourceTypeInstance` and `ec2.ResourceTypeVolume`), we just shouldn't add the tag specification. That's why the guard is there. And I think having a block, whether it's just `{ ... }` or `if ... { ... }` around each category, together with a comment above each block, separates the logic somewhat cleanly.

pkg/cloud/services/interfaces.go
```diff
@@ -65,22 +66,29 @@ type EC2Interface interface {
...
 	LaunchTemplateNeedsUpdate(scope scope.LaunchTemplateScope, incoming *expinfrav1.AWSLaunchTemplate, existing *expinfrav1.AWSLaunchTemplate) (bool, error)
 	DeleteBastion() error
 	ReconcileBastion() error
 }
 
+// ReconcileInterface encapsulates high-level reconciliation functions regarding EC2 reconciliation. It is separate
+// from EC2Interface so that we can test the behavior of our non-mock implementations. For example, by not mocking
+// the ReconcileLaunchTemplate function, but mocking EC2Interface, we can test which EC2 API operations would have
+// been called.
+type ReconcileInterface interface {
```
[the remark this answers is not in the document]
AndiDog: `EC2Interface` mocks the AWS EC2 API, so `ReconcileLaunchTemplate` (which I moved out) and `ReconcileTags` (out of scope for my PR) should not be in there. If I move the `ReconcileTags` function as well into the new interface, it should become clearer for someone working on the code, particularly if we state that it's for the functionality of the controller as opposed to raw AWS requests. Given that change, what about `MachinePoolReconcileInterface` since those functions are exclusively used in `awsmachinepool_controller.go` and `awsmanagedmachinepool_controller.go`?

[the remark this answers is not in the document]
cnmcavoy: I think that would clear up my confusion, sgtm.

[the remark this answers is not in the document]
AndiDog: Done. Also fixed lint errors.

pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -689,14 +706,29 @@ func (s *Service) SDKToLaunchTemplate(d *ec2.LaunchTemplateVersion) (*expinfrav1
...
+	var launchTemplateUserDataSecretKey *apimachinerytypes.NamespacedName
+	for _, tagSpecification := range v.TagSpecifications {
+		if tagSpecification.ResourceType != nil && *tagSpecification.ResourceType == ec2.ResourceTypeInstance {
+			for _, tag := range tagSpecification.Tags {
+				if tag.Key != nil && *tag.Key == infrav1.LaunchTemplateBootstrapDataSecret && tag.Value != nil && strings.Contains(*tag.Value, "/") {
+					parts := strings.SplitN(*tag.Value, "/", 2)
+					launchTemplateUserDataSecretKey = &apimachinerytypes.NamespacedName{
+						Namespace: parts[0],
+						Name:      parts[1],
+					}
```
cnmcavoy, line 725: Can we return here?

pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -689,14 +706,29 @@ func (s *Service) SDKToLaunchTemplate(d *ec2.LaunchTemplateVersion) (*expinfrav1
...
+				}
+			}
+		}
 	}
 
-	return i, userdata.ComputeHash(decodedUserData), nil
+	return i, userdata.ComputeHash(decodedUserData), launchTemplateUserDataSecretKey, nil
```
cnmcavoy, line 731: is it okay for `launchTemplateUserDataSecretKey` to be potentially nil here, if the above loop fails? Or should we return an error here?

pkg/cloud/services/ec2/launchtemplate.go
AndiDog: Done

pkg/cloud/services/ec2/launchtemplate.go
AndiDog: We must be backward-compatible and support cases where the tag is missing by mistake/bug/intervention. So `nil` is fine. I changed this line to explicit `nil` here, and added the `return` for the found case into the loop as you commented above.

pkg/cloud/services/ec2/launchtemplate.go
```diff
@@ -845,19 +873,24 @@ func (s *Service) buildLaunchTemplateTagSpecificationRequest(scope scope.LaunchT
...
 				Value: aws.String(value),
 			})
 		}
 		tagSpecifications = append(tagSpecifications, spec)
+	}
 
-		// tag EBS volumes
-		spec = &ec2.LaunchTemplateTagSpecificationRequest{ResourceType: aws.String(ec2.ResourceTypeVolume)}
+	// tag EBS volumes
+	if len(tags) > 0 {
```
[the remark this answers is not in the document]
AndiDog: Does the above answer solve your question? Here, `tags` could be empty and we want an `if`, while for instance tags, we now have at least one (the newly added tag used for the feature) and need no `if`.

cnmcavoy (approved):
/lgtm

AndiDog:
Rebased and only one (obvious to me) test line was conflicted. @cnmcavoy @fiunchinho could you take another look? @richardcase we'd need approval here as well. I'm not sure if this should go into v2.3.x?

richardcase:
> I'm not sure if this should go into v2.3.x?

@AndiDog - as its classed as a feature i'd say it needs to go into v2.4.0 instead. We generally only consider bugs for a patch release.

AndiDog:
@cnmcavoy @richardcase The PR needs LGTM (again after rebase) and approval.

I agree it's a feature, so no need to backport. We have integrated it already in our CAPA fork and it's working.

cnmcavoy (approved):
/lgtm

richardcase:
This looks good to me. Lets check the e2e passes before merging:

damdo:
@AndiDog see for your e2e failure, [see this](https://kubernetes.slack.com/archives/CD6U2V71N/p1705574166982269?thread_ts=1705565476.663379&cid=CD6U2V71N)

AndiDog:
Rebased because that fixed E2E tests in another PR.

fiunchinho (approved):
/lgtm

AndiDog:
Still one E2E failure related to LBs. The test function `createLBService` look strange to me. It expects some status field to be filled immediately or otherwise silently continues.

Ankitasw:
/approve

cc @fiunchinho @cnmcavoy @AndiDog for final review

richardcase (commented):
Looks good to me as well

/approve

For the final lgtm:

/assign @cnmcavoy @fiunchinho

Issue #3989, opened 2023-01-17
User-provided `Name` tag on subnets get overwritten


AverageMarcus:
/kind bug

**What steps did you take and what happened:**

Specify a subnet on an `AWSCluster` CR that has a `Name` tag defined. E.g. 

```yaml
apiVersion: infrastructure.cluster.x-k8s.io/v1beta1
kind: AWSCluster
metadata:
  name: example
spec:
  network:
    subnets:
    - availabilityZone: eu-north-1a
      cidrBlock: 10.0.0.0/24
      isPublic: false
      tags:
        Name: example-custom-subnet-name
```

This `Name` tag will be instead replaced with one built in code and instead become `Name: example-subnet-private-eu-north-1a`

**What did you expect to happen:**

If a `Name` tag is provided by the user it is what should be used when creating / updating the subnet. If not provided CAPA should default to the current generated format.

**Anything else you would like to add:**

Block of code building the Name value: https://github.com/giantswarm/cluster-api-provider-aws/blob/70373384dc117ea387d28a8548d49ecc58e7cca0/pkg/cloud/services/network/subnets.go#L426-L441

Block of code overwriting any user-provided Name tag: https://github.com/giantswarm/cluster-api-provider-aws/blob/70373384dc117ea387d28a8548d49ecc58e7cca0/api/v1beta1/tags.go#L198-L200

As far as I can see, that tag isn't used anywhere in the codebase for matching or looking up of subnets. It seem purely for display.

Multiple subnets can end up with the same Name value.

**Environment:**

- Cluster-api-provider-aws version: >= 1.5
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

PR #4630, opened 2023-11-13
WIP: Do not allow overriding the `Name` tag so that the AWSMachine controller can reliably find its EC2 instance

### LINKED ISSUE #4629: AWSMachine controller leaks EC2 instances when specifying a Name tag in `additionalTags`
/kind bug

After applying two `infrastructure.cluster.x-k8s.io/v1beta2` AWSMachines with the following additionalTags:

```yaml
spec:
  additionalTags:
    Name: sample-name-tag
```

124 EC2 instances were created in quick succession for these AWSMachines before `.spec.providerID` was populated. All of the EC2 instances (leaked and not-leaked) do have the tags `Name: sample-name-tag`

```
❯ k get awsmachine                                                  
NAME                                    CLUSTER                            STATE     READY   INSTANCEID                              MACHINE
sample-name-tag-251d8688-cjd7r   ${ID}   running   true    aws:///us-east-2a/i-019c51ee0ded3e51a   sample-name-tag-workers-55b788fd4dx5n4zb-kn2h6
sample-name-tag-251d8688-hwhb4   ${ID}   running   true    aws:///us-east-2a/i-0c9169fb1781d2a8e   sample-name-tag-workers-55b788fd4dx5n4zb-8brbs
```

One EC2 instance to be created per AWSMachine, with the tag `Name: sample-name-tag`

I think this occurs because:

1. We try to find an instance during a normal reconcile loop - https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/56c9a39dd834640ee4a027e679ad2a5757098dfd/controllers/awsmachine_controller.go#L483-L489 
2. We look for the instance with these tags, including the name of the AWSMachine https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/abd444c27083d5054c772c7bf5101266da10f76f/pkg/cloud/services/ec2/instances.go#L46-L53
3. If `instance == nil`, then a new EC2 instance is created https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/56c9a39dd834640ee4a027e679ad2a5757098dfd/controllers/awsmachine_controller.go#L517C29-L522 
4. However, this only happens when the instance cannot be found by tags (or an AWS not found/permission error). So the bug is with a change in v1beta2 to allow overwriting the `Name` tag: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta2/tags.go#L244-L264

vs v1beta1 https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/abd444c27083d5054c772c7bf5101266da10f76f/api/v1beta1/tags.go#L250-L269 where the AWSMachine's own `.metadata.name` had precedence

So I think we need to either not allow overwriting the `Name` tag or we need to be smarter about how we find the instance based off of tags. Even if we update the filtering logic to account for the additional Name tag, we don't want to find an instance that's already been claimed by another AWSMachine.

**Environment:**

```
Client Version: v1.28.3
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.6+f67aeb3
```

mjlshen:
/kind bug

This commit handles an edge-case where an error occurs after EC2 instance creation and before the output of a successful EC2 instance creation is returned. This causes the AWSMachine controller to attempt to find the EC2 instance by a set of filters instead of being able to explicitly find an EC2 instance by ID. Currently, the logic for finding an EC2 instance is not guaranteed to find the correct instance because one of the filters is by the `Name` tag, which can be overwritten by the user in the v1beta2 API.

In essence, this PR reverts #3991 because one of the assumptions in the backing issue, #3989:

> As far as I can see, that tag isn't used anywhere in the codebase for matching or looking up of subnets. It seem purely for display.

Was true for subnets, but is not true for EC2 instances.

Fixes #4629 

I'm not sure if this is the right way to solve this problem, two other possibilities I thought of are:
* Allowing users to override the `Name` tag for subnets, but not for EC2 instances
* Making `findInstances` more idempotent when overriding the `Name` tag https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/abd444c27083d5054c772c7bf5101266da10f76f/pkg/cloud/services/ec2/instances.go#L46-L53 this could be done by adding a new/unique CAPA-specific "name" tag so that `findInstances` can filter on that new tag instead of a `Name` tag that a user can overwrite, resulting in a unique EC2 instance match.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Disallow users from setting the `Name` tag on created AWS resources
```


enxebre:
Thanks @mjlshen! This is awesome but I don't think it handles all the edge cases where the providerID might still not be reliable. E.g. aws API eventual consistency [1], process crash right after the API call... So the tags fallback must be trustable. The name of the replicas in the scalable resource is an implementation detail and letting it being overridden is a bug. I think we should consider reverting the original change that enabled it to be overridden, or if there's still justification for the network use case which drove that change, then we need to differentiate so this is not allowed for instances. 

[1] https://docs.aws.amazon.com/AWSEC2/latest/APIReference/query-api-troubleshooting.html#eventual-consistency

mjlshen:
@AndiDog my PR is essentially reverting your PR for #3989 - do you have a use-case for setting the `Name` tag on subnets?

AndiDog:
Users should be able to set the `Name` tag since like all non AWS-internal tags, it has no specific meaning apart from display purposes (e.g. AWS Console). That goes for EC2 instances, subnets and probably most other AWS resources.

If CAPA cannot find the instance correctly because it is tagged with `{AWSCluster,AWSMachine}.spec.additionalTags["Name"]`, as expected, instead of `scope.Name()` (which currently returns `AWSMachine.Name`), then that's the bug and the find-by-name functionality needs to be fixed instead of forbidding overrides of the value. I think the tagging code is in `func (r *AWSMachineReconciler) ensureTags`, but I didn't read this complex function and assume it uses the additional tags map to set the `Name`, and does not make use of `scope.Name()`. If that's correct, then `filter.EC2.Name(scope.Name()),` is wrong and needs fixing plus a unit test to cover the case where the name tag is overridden.

mjlshen:
Ah the `Name` tag is getting set correctly on EC2 instances, but it's the same for every EC2 instance, so `findInstances` can no longer uniquely find an instance in certain circumstances.

I think this means we need to introduce a new tag to uniquely link an EC2 instance to an AWS machine CR

AndiDog:
True, that's definitely a problem since we only reconcile a single object and can't reason about others (such as "3 AWSMachine objects with same name tag").

Can you point me to the code which sets the EC2 instance name if no `AdditionalTags["name"]` is given?

mjlshen:
Sorry for the delayed response - yes!

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b003e69199b3b598a1c0f26a5726466d0a331fb7/pkg/cloud/services/ec2/instances.go#L123-L131 sets the EC2 instance name on creation

Where `infrav1.Build()` is the same Build function to assemble tag precedence: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/abd444c27083d5054c772c7bf5101266da10f76f/api/v1beta2/tags.go#L243-L264 (where an additionally specified `Name` tag now has precedence over the AWSMachine `.metadata.name` in v1beta2 now)

richardcase:
It would be good to get @AverageMarcus input on this as well.

AverageMarcus:
It's been quite a while since I've been deeply involved in CAPA code so I'm gonna keep this high level...

My general thought is that the users should be able to configure any of the tags they like, including `Name`, so that they can adhere to any possible tag policies they have in their organisation. If we're needing to use tags for any logic based decisions we should always be using CAPA-specific tag keys so instead of using the `Name` tag we would instead use something like `sigs.k8s.io/cluster-api-provider-aws/MachineName`.

Some background on the initial introduction of the configurable `Name` tag:
* PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/3991
* Issue: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3989
The driving force behind the initial change for us (Giant Swarm) is we wanted to allow for some complex network configurations where different machine pools could have different subnets and wanted to have control over how those subnets were named so they'd be identifiable. I forget what exactly the problem was with them all having the same name but if I remember correctly it was related to configuring transit gateway attachments and not being able to pick the correct subnet.

mjlshen:
Yeah, a new tag makes sense to me too, given that we want to be able to override a `Name` tag. I will update this PR to go in a new direction and introduce a new CAPA-specific tag based on AWSMachine name, thanks for all the feedback!

PR #4648, opened 2023-11-21
✨ Add separate eks kubeconfig secret keys for the cluster-autoscaler

### LINKED ISSUE #4607: Refactor the AWS & EKS control-plane controllers to split the kubeconfig secret into two for Cluster Autoscaler
/kind feature

Currently, the AWSControlPlane and AWSManagedControlPlane controllers create and manage the cluster kubeconfig secrets (`"%s-kubeconfig"` in the cluster's namespace). The contents of this secret is a single key-value, with the value being the cluster's kubeconfig in the shape of:
```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0t...
    server: https://XXXXX.us-east-2.eks.amazonaws.com
  name: cluster-name
contexts:
- context:
    cluster: cluster-3
    user: cluster-3-capi-admin
  name: cluster-3-capi-admin@cluster-3
current-context: cluster-3-capi-admin@cluster-3
kind: Config
preferences: {}
users:
- name: cluster-3-capi-admin
  user:
    token: k8s-aws-v1...
```

In the case of EKS clusters, the bearer token embedded in the `token` field expires quickly and the CAPA controller regularly updates the secret to keep it fresh. However, this shape of config is hard for other controllers, like the cluster autoscaler to consume, and leads to bugs like https://github.com/kubernetes/autoscaler/issues/4784 which can not be easily resolved by those controllers. If they reload the entire config, the user or other critical content may change, if they reload only the token within, then they are disregarding the contents and making unsafe assumptions.

I have discussed this problem with the folks over in SIG API-Machinery and it was suggested to split this secret into two key-values, in this shape:
```yaml
apiVersion: v1
clusters:
- cluster:
    certificate-authority-data: LS0t...
    server: https://XXXXX.us-east-2.eks.amazonaws.com
  name: cluster-name
contexts:
- context:
    cluster: cluster-3
    user: cluster-3-capi-admin
  name: cluster-3-capi-admin@cluster-3
current-context: cluster-3-capi-admin@cluster-3
kind: Config
preferences: {}
users:
- name: cluster-3-capi-admin
  user:
    tokenFile: ./bearer-token
```
And a second key-value, "`bearer-token`" with the contents:
```
k8s-aws-v1...
```
Additional context: https://github.com/kubernetes/autoscaler/pull/5951#discussion_r1378786610

This will allow controllers like the Cluster Autoscaler to mount the secret contents to a path, and it will populate two files. The existing api machinery then should be able to handle the reloading of the 2nd bearer-token file without complications.

Changing the secret value to depend on the 2nd value existing will break any users that consume only the one secret key-value explicitly, in a subPath volume mount. Normal volume mounts should work, as both files will exist relative to each other.

cnmcavoy:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**: Cluster Autoscaler can not mount and consume the [Cluster API Kubeconfig](https://cluster-api-aws.sigs.k8s.io/topics/eks/creating-a-cluster.html?highlight=kubeconfig#cluster-api-capi-kubeconfig) because the secret contents are refreshed every ten minutes, and no API Machinery exists to reload a kubeconfig safely.

Initially, I attempted solve this in the Cluster Autoscaler: https://github.com/kubernetes/autoscaler/issues/4784 - [However meeting with SIG APIMachinery on Nov 1 2023](https://youtu.be/JVs79gb7qJE), the SIG cautioned against this approach and advised splitting the token out from the kubeconfig, [as there is existing machinery to reload a token file auth](https://github.com/kubernetes/kubernetes/pull/67359). By switching to this approach, no change in the Cluster Autoscaler is needed, users only need to update their Cluster Autoscaler configuration to use the correct secret file from their secret volume mount.

Fixes #4607

**Checklist**:

- [X] squashed commits
- [X] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Add separate eks kubeconfig secret keys for the cluster-autoscaler to support refreshing the token automatically, see eks kubeconfig for more info.
```

### DIFF
--- a/docs/book/src/topics/eks/creating-a-cluster.md
+++ b/docs/book/src/topics/eks/creating-a-cluster.md
@@ -34,4 +34,12 @@ kubectl --namespace=default get secret managed-test-user-kubeconfig \
 
 This kubeconfig is used internally by CAPI and shouldn't be used outside of the management server. It is used by CAPI to perform operations, such as draining a node. The name of the secret that contains the kubeconfig will be `[cluster-name]-kubeconfig` where you need to replace **[cluster-name]** with the name of your cluster. Note that there is NO `-user` in the name.
 
-The kubeconfig is regenerated every `sync-period` as the token that is embedded in the kubeconfig is only valid for a short period of time. When EKS support is enabled the maximum sync period is 10 minutes. If you try to set `--sync-period` to greater than 10 minutes then an error will be raised.
+There are three keys in the CAPI kubeconfig for eks clusters:
+
+| keys        | purpose                                                                                                                                                                            |
+|-------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| value       | contains a complete kubeconfig with the cluster admin user and token embedded                                                                                                      |
+| relative    | contains a kubeconfig with the cluster admin user, referencing the token file in a relative path - assumes you are mounting all the secret keys in the same dir                    |
+| single-file | contains the same token embedded in the complete kubeconfig, it is separated into a single file so that existing APIMachinery can reload the token file when the secret is updated |
+
+The secret contents are regenerated every `sync-period` as the token that is embedded in the kubeconfig and token file is only valid for a short period of time. When EKS support is enabled the maximum sync period is 10 minutes. If you try to set `--sync-period` to greater than 10 minutes then an error will be raised.
--- a/pkg/cloud/services/eks/config.go
+++ b/pkg/cloud/services/eks/config.go
@@ -31,9 +31,12 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/client-go/tools/clientcmd/api"
+	"sigs.k8s.io/controller-runtime/pkg/client"
 
 	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/record"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+	"sigs.k8s.io/cluster-api/util"
 	"sigs.k8s.io/cluster-api/util/kubeconfig"
 	"sigs.k8s.io/cluster-api/util/secret"
 )
@@ -42,6 +45,9 @@ const (
 	tokenPrefix       = "k8s-aws-v1." //nolint:gosec
 	clusterNameHeader = "x-k8s-aws-id"
 	tokenAgeMins      = 15
+
+	relativeKubeconfigKey = "relative"
+	relativeTokenFileKey  = "token-file"
 )
 
 func (s *Service) reconcileKubeconfig(ctx context.Context, cluster *eks.Cluster) error {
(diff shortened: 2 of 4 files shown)


cnmcavoy:
This is ready to be reviewed, but I haven't had an opportunity to test the full setup with this change + cluster autoscaler end to end yet, so marked as a draft until I verify it.

docs/book/src/topics/eks/creating-a-cluster.md
```diff
@@ -34,4 +34,12 @@ kubectl --namespace=default get secret managed-test-user-kubeconfig \
...
+There are three keys in the CAPI kubeconfig for eks clusters:
+
+| keys                          | purpose                                                                                                                                                                            |
+|-------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| value                         | contains a complete kubeconfig with the cluster admin user and token embedded                                                                                                      |
+| cluster-autoscaler-value      | contains a kubeconfig with the cluster admin user, referencing the token file in a relative path - assumes you are mounting all the secret keys in the same dir                    |
+| cluster-autoscaler-token-file | contains the same token embedded in the complete kubeconfig, it is separated into a single file so that existing APIMachinery can reload the token file when the secret is updated |
```
mloiseleur, line 43: Since there are more use cases than just cluster-autoscaler, wdyt about using descriptive names like `embedded`, `relative` and `single-file` ?

cnmcavoy: These names are good, but changing the name of `value` means we have a different Secret shape in EKS clusters than self-managed clusters. I'll update the other two, but I am not sure changing the name of `value` is worth it, as it will be surprising to users.

richardcase:
It would be good to get cluster autoscaler added to our e2e tests. Lets create an issue to follow up on this.

From my side this look good:

/approve

When the e2e passes we can unhold and merge:

/hold

cnmcavoy:
> New changes are detected. LGTM label has been removed.

Only a rebase.

mloiseleur:
You will have to fix the test first.
/lgtm cancel

cnmcavoy:
@mloiseleur seems like the test failure is a flake? I pushed a commit to simply expose the error being suppressed in cloudformation and now it passed. My new commit should not have fixed any tests.

mloiseleur:
@cnmcavoy I updated to v2.6.1 this morning and I see in the logs an error about ownership in the log for all control plane.

```shell
E0807 08:02:32.550249       1 controller.go:329] "Reconciler error" err="failed to reconcile control plane for AWSManagedControlPlane flux-system/xxxxxxxxx-cp: failed reconciling kubeconfig: updating kubeconfig secret: EKS kubeconfig flux-system/s00-euw2-eks-hub-kubeconfig missing expected AWSManagedControlPlane ownership" controller="awsmanagedcontrolplane" controllerGroup="controlplane.cluster.x-k8s.io" controllerKind="AWSManagedControlPlane" AWSManagedControlPlane="flux-system/xxxxxxxxx-cp" namespace="flux-system" name="xxxxxxxxx-cp" reconcileID="c634fd01-5c99-4947-9ff7-3297fcaff97c"
```

Did you encounter this issue when you tested on your side ?

cnmcavoy:
>  @cnmcavoy I updated to v2.6.1 this morning and I see in the logs an error about ownership of the kubeconfig secret in the log for all control plane.
The secret was created with an owner reference in the previous releases as well. Previously the shape of the secret was assumed, the new behavior that you encountered is that the controller checks if it owns the secret before operating on it. Deleting and recreating was also going to be my suggestion, but the owner reference should already exist unless the secret was not created by the controller. My suspicion is flux or some other system created the secret and so CAPA didn't set the owner reference.

PR #4654, opened 2023-11-23
:bug: Update ASG - do not set desired value for machinepool which have externally managed replicas

calvix:
/kind bug

For machine pools that have externally managed replicas (for example by `cluster-autoscaler`) the cluster-api-provider-aws should not set the desired replicas otherwise it can cause a problem where the controller is trying to update an ASG with desired replicas that are out of bounds of min/max.

Example:
The machine pool is set to have `min:1`  instances and `max: 3` instances and it's currently running on 2 desired replicas. The user wants to update to new values `min: 5` and `max: 10`.  The update operation fails with an AWS API error 
```
E1122 16:24:14.026636       1 controller.go:324] "Reconciler error" err=<
	unable to update ASG: failed to update ASG "tsa05-ndp01": ValidationError: Desired capacity:2 must be between the specified min size:5 and max size:10
		status code: 400, request id: f142fa2f-6582-43c7-8296-0a88ac104adf
```
Updating the `MachinePool.spec.replicas` will not help as the field is updated based on the current amount of nodes/replicas due to externally managed replicas configuration so it will be set back to the current value of `2`

**Release note**:
```release-note
Treat the annotation `cluster.x-k8s.io/replicas-managed-by` on MachinePool CR as described in [CAPI documentation](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/architecture/controllers/machine-pool.md#externally-managed-autoscaler) and do not set desired replicas on ASG when the annotation is present.
```

### DIFF
--- a/pkg/cloud/services/autoscaling/autoscalinggroup.go
+++ b/pkg/cloud/services/autoscaling/autoscalinggroup.go
@@ -34,6 +34,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/converters"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/record"
+	"sigs.k8s.io/cluster-api/util/annotations"
 )
 
 // SDKToAutoScalingGroup converts an AWS EC2 SDK AutoScalingGroup to the CAPA AutoScalingGroup type.
@@ -46,7 +47,7 @@ func (s *Service) SDKToAutoScalingGroup(v *autoscaling.Group) (*expinfrav1.AutoS
 		MaxSize:           int32(aws.Int64Value(v.MaxSize)),
 		MinSize:           int32(aws.Int64Value(v.MinSize)),
 		CapacityRebalance: aws.BoolValue(v.CapacityRebalance),
-		//TODO: determine what additional values go here and what else should be in the struct
+		// TODO: determine what additional values go here and what else should be in the struct
 	}
 
 	if v.VPCZoneIdentifier != nil {
@@ -177,7 +178,7 @@ func (s *Service) CreateASG(machinePoolScope *scope.MachinePoolScope) (*expinfra
 	// Ignore the problem for externally managed clusters because MachinePool replicas will be updated to the right value automatically.
 	if mpReplicas >= machinePoolScope.AWSMachinePool.Spec.MinSize && mpReplicas <= machinePoolScope.AWSMachinePool.Spec.MaxSize {
 		input.DesiredCapacity = &mpReplicas
-	} else if !scope.ReplicasExternallyManaged(machinePoolScope.MachinePool) {
+	} else if !annotations.ReplicasManagedByExternalAutoscaler(machinePoolScope.MachinePool) {
 		return nil, fmt.Errorf("incorrect number of replicas %d in MachinePool %v", mpReplicas, machinePoolScope.MachinePool.Name)
 	}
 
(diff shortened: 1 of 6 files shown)


pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -284,35 +284,35 @@ func (s *Service) DeleteASG(name string) error {
 }
 
 // UpdateASG will update the ASG of a service.
-func (s *Service) UpdateASG(scope *scope.MachinePoolScope) error {
-	subnetIDs, err := s.SubnetIDs(scope)
+func (s *Service) UpdateASG(machinePoolScope *scope.MachinePoolScope) error {
```
calvix, line 287: the renaming of the variable from `scope` to `machinePoolScope` is necessary to be able to access the package `scope` and the function `scope.ReplicasExternallyManaged`

pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -284,35 +284,35 @@ func (s *Service) DeleteASG(name string) error {
...
+		MaxSize:              aws.Int64(int64(machinePoolScope.AWSMachinePool.Spec.MaxSize)),
+		MinSize:              aws.Int64(int64(machinePoolScope.AWSMachinePool.Spec.MinSize)),
 		VPCZoneIdentifier:    aws.String(strings.Join(subnetIDs, ",")),
-		CapacityRebalance:    aws.Bool(scope.AWSMachinePool.Spec.CapacityRebalance),
+		CapacityRebalance:    aws.Bool(machinePoolScope.AWSMachinePool.Spec.CapacityRebalance),
 	}
 
-	if scope.MachinePool.Spec.Replicas != nil {
-		input.DesiredCapacity = aws.Int64(int64(*scope.MachinePool.Spec.Replicas))
+	if machinePoolScope.MachinePool.Spec.Replicas != nil && !scope.ReplicasExternallyManaged(machinePoolScope.MachinePool) {
```
calvix, line 301: here is the actual change

fiunchinho: Could you add a test case for this change?

AndiDog:
Change looks good per se, however the existing implementation of the annotation check function seems strange:

```
func ReplicasExternallyManaged(mp *expclusterv1.MachinePool) bool {
	val, ok := mp.Annotations[ReplicasManagedByAnnotation]
	return ok && val == ExternalAutoscalerReplicasManagedByAnnotationValue
}
```

CAPA requires a specific value which is defined like this:

```
const (
	// ReplicasManagedByAnnotation is an annotation that indicates external (non-Cluster API) management of infra scaling.
	// The practical effect of this is that the capi "replica" count is derived from the number of observed infra machines,
	// instead of being a source of truth for eventual consistency.
	//
	// N.B. this is to be replaced by a direct reference to CAPI once https://github.com/kubernetes-sigs/cluster-api/pull/7107 is meged.
	ReplicasManagedByAnnotation = "cluster.x-k8s.io/replicas-managed-by"

	// ExternalAutoscalerReplicasManagedByAnnotationValue is used with the "cluster.x-k8s.io/replicas-managed-by" annotation
	// to indicate an external autoscaler enforces replica count.
	//
	// N.B. this is to be replaced by a direct reference to CAPI once https://github.com/kubernetes-sigs/cluster-api/pull/7107 is meged.
	ExternalAutoscalerReplicasManagedByAnnotationValue = "external-autoscaler"
)
```

But the [CAPI "contract" for this annotation says](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/architecture/controllers/machine-pool.md#externally-managed-autoscaler):

> Cluster API treats the annotation as a "boolean", meaning that the presence of the annotation is sufficient to indicate external replica count management, with one exception: if the value is `"false"`, then that indicates to Cluster API that replica enforcement is nominal, and managed by Cluster API.
>
> Providers may choose to implement the `cluster.x-k8s.io/replicas-managed-by` annotation with different values (e.g., `external-autoscaler`, or `karpenter`) that may inform different provider-specific behaviors, but those values will have no effect upon Cluster API.

So the check should be patched as follows (untested):

```
func ReplicasExternallyManaged(mp *expclusterv1.MachinePool) bool {
	val, ok := mp.Annotations[ReplicasManagedByAnnotation]
	return ok && val != "false"
}
```

And we should delete the const `ExternalAutoscalerReplicasManagedByAnnotationValue` since any non-false value is okay.

This is out of scope for the PR, but I quickly wrote a test + fixed the above, so that will be pushed to this PR branch soon by my colleague @calvix.

We should have someone from another company, and with expertise with that code, to review as well, please.

AndiDog:
@calvix Please squash into a single commit with a reasonable commit message (e.g. the PR title). Also, please edit the PR to include a release note (see [PR template](https://github.com/kubernetes-sigs/cluster-api-provider-aws/edit/main/.github/PULL_REQUEST_TEMPLATE.md) if you need to know the formatting).

This now contains my commits, so I shouldn't review.

pkg/cloud/services/autoscaling/autoscalinggroup.go
AndiDog: I added a test

PR #4656, opened 2023-11-23
:sparkles: Prevent users setting values that are invalid configuration on AWS

fiunchinho:
**What type of PR is this?**

/kind feature

Users can currently set both `SpotMarketOptions` and `MixedInstancesPolicy`, but this is not a valid configuration on AWS, resulting on this error
```
InvalidQueryParameter: Incompatible launch template: You cannot use a launch template that is set to request Spot Instances (InstanceMarketOptions) when you configure an Auto Scaling group with a mixed instances policy. Add a different launch template to the group and try again.
```

It's better to prevent this configuration early, so users get quick feedback on what's wrong with their clusters.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Prevent users setting SpotMarketOptions and MixedInstancesPolicy at the same time
```

### DIFF
--- a/exp/api/v1beta2/awsmachinepool_webhook.go
+++ b/exp/api/v1beta2/awsmachinepool_webhook.go
@@ -108,6 +108,13 @@ func (r *AWSMachinePool) validateAdditionalSecurityGroups() field.ErrorList {
 	}
 	return allErrs
 }
+func (r *AWSMachinePool) validateSpotInstances() field.ErrorList {
+	var allErrs field.ErrorList
+	if r.Spec.AWSLaunchTemplate.SpotMarketOptions != nil && r.Spec.MixedInstancesPolicy != nil {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.awsLaunchTemplate.AdditionalSecurityGroups"), "either spec.awsLaunchTemplate.SpotMarketOptions or spec.MixedInstancesPolicy should be used"))
+	}
+	return allErrs
+}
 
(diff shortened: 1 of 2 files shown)


exp/api/v1beta2/awsmachinepool_webhook.go
```diff
@@ -108,6 +108,13 @@ func (r *AWSMachinePool) validateAdditionalSecurityGroups() field.ErrorList {
 	}
 	return allErrs
 }
+func (r *AWSMachinePool) validateSpotInstances() field.ErrorList {
+	var allErrs field.ErrorList
+	if r.Spec.AWSLaunchTemplate.SpotMarketOptions != nil && r.Spec.MixedInstancesPolicy != nil {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.awsLaunchTemplate.AdditionalSecurityGroups"), "either spec.awsLaunchTemplate.SpotMarketOptions or spec.MixedInstancesPolicy should be used"))
```
AndiDog, line 114:
```suggestion
		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.awsLaunchTemplate.spotMarketOptions"), "either spec.awsLaunchTemplate.spotMarketOptions or spec.mixedInstancesPolicy should be used"))
```

AndiDog (commented):
/lgtm

faiq (approved):

PR #4662, opened 2023-11-28
:bug:  There is no need to check instance refresh if ASG is not found

### LINKED ISSUE #4655: AWSMachinePool reconciliation stuck if ASG could not be created
/kind bug

We created a new cluster having a `AWSMachinePool`. Upon adding a second `AWSMachinePool`, for some reason, the IAM instance profile wasn't ready/found yet and the ASG creation failed. That's out of scope for the issue. Just mentioning the error for completeness and visibility:

```
failed to create AWSMachinePool: failed to create autoscaling group: ValidationError: You must use a valid fully-formed launch template. Value (nodes-nodepool2-theclustername) for parameter iamInstanceProfile.name is invalid. Invalid IAM Instance Profile name
```

This means that the `AWSMachinePool` reconciler had successfully created the launch template (and set the IAM instance profile name on the launch template), but then failed on ASG creation.

The problem is that reconciliation is now stuck due to a bug: on next `Reconcile` attempt, `ReconcileLaunchTemplate` detects a diff, calls `canUpdateLaunchTemplate` and that in turn calls `CanStartASGInstanceRefresh`. In there, an AWS request `DescribeInstanceRefreshesInput{AutoScalingGroupName: aws.String(scope.Name())}` is made and fails because the ASG doesn’t exist yet. The error is:

```
E1123 13:38:44.326581       1 logger.go:83] "failed to reconcile launch template" err=<
	ValidationError: AutoScalingGroup name not found - AutoScalingGroup theclustername-nodepool2 not found
		status code: 400, request id: 17090c90-0275-40dd-9a1d-2e2994105934
```

We must handle, log and ignore that error. If the ASG does not exist, `CanStartASGInstanceRefresh` should return false. Or `ReconcileLaunchTemplate` should not even consider instance refresh through some other means.

Reconciliation continues and eventually succeeds.

**Environment:**

- Cluster-api-provider-aws version: roughly v2.2.4

fiunchinho:
**What type of PR is this?**

/kind bug

[On this PR](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4660) we tried to fix #4655 by ignoring the "not found" error when describing instance refreshes resources on AWS. But the AWS API returns a `400` error with code `ValidationError` when trying to describe the instance refreshes for a non existing AutoScalingGroup. Since `ValidationError` sounds super generic, instead of ignoring that error, on this PR I'm moving the call to find the AutoScalingGroup before defining the function that will check if we can do an instance refresh. That way, if the AutoScalingGroup is nil, we know we can skip checking. Also, we still want to update the `LaunchTemplate` when the AutoScalingGroup does not exist, because an error in the `LaunchTemplate` may be the root cause why the ASG is not created yet.

Fixes #4655

Unfortunately, there is no easy way to test this change, because `ReconcileLaunchTemplate`, which is the one receiving the anonymous function that we want to modify here is mocked.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Skip instance refresh attempt if ASG does not yet exist
```

### DIFF
--- a/exp/controllers/awsmachinepool_controller.go
+++ b/exp/controllers/awsmachinepool_controller.go
@@ -227,13 +227,30 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 	ec2Svc := r.getEC2Service(ec2Scope)
 	asgsvc := r.getASGService(clusterScope)
 
+	// Find existing ASG
+	asg, err := r.findASG(machinePoolScope, asgsvc)
+	if err != nil {
+		conditions.MarkUnknown(machinePoolScope.AWSMachinePool, expinfrav1.ASGReadyCondition, expinfrav1.ASGNotFoundReason, err.Error())
+		return err
+	}
+
 	canUpdateLaunchTemplate := func() (bool, error) {
 		// If there is a change: before changing the template, check if there exist an ongoing instance refresh,
 		// because only 1 instance refresh can be "InProgress". If template is updated when refresh cannot be started,
 		// that change will not trigger a refresh. Do not start an instance refresh if only userdata changed.
+		if asg == nil {
+			// If the ASG hasn't been created yet, there is no need to check if we can start the instance refresh.
+			// But we want to update the LaunchTemplate because an error in the LaunchTemplate may be blocking the ASG creation.
+			return true, nil
+		}
 		return asgsvc.CanStartASGInstanceRefresh(machinePoolScope)
 	}
 	runPostLaunchTemplateUpdateOperation := func() error {
+		// skip instance refresh if ASG is not created yet
+		if asg == nil {
+			machinePoolScope.Debug("ASG does not exist yet, skipping instance refresh")
+			return nil
+		}
 		// skip instance refresh if explicitly disabled
 		if machinePoolScope.AWSMachinePool.Spec.RefreshPreferences != nil && machinePoolScope.AWSMachinePool.Spec.RefreshPreferences.Disable {
 			machinePoolScope.Debug("instance refresh disabled, skipping instance refresh")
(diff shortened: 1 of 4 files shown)


exp/controllers/awsmachinepool_controller.go
```diff
@@ -227,13 +227,29 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
...
+			// If the ASG hasn't been created yet, there is no need to check if we can start the instance refresh.
+			// But we want to update the LaunchTemplate because an error in the LaunchTemplate may be blocking the ASG creation.
+			return true, nil
+		}
 		return asgsvc.CanStartASGInstanceRefresh(machinePoolScope)
 	}
 	runPostLaunchTemplateUpdateOperation := func() error {
+		// skip instance refresh if ASG is not created yet
+		if asg == nil {
+			return nil
```
AndiDog, line 251:
```suggestion
			machinePoolScope.Debug("ASG does not exist yet, skipping instance refresh")
			return nil
```

pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -323,9 +323,6 @@ func (s *Service) CanStartASGInstanceRefresh(scope *scope.MachinePoolScope) (boo
 	describeInput := &autoscaling.DescribeInstanceRefreshesInput{AutoScalingGroupName: aws.String(scope.Name())}
 	refreshes, err := s.ASGClient.DescribeInstanceRefreshesWithContext(context.TODO(), describeInput)
 	if err != nil {
-		if awserrors.IsNotFound(err) {
```
AndiDog, line 326: For the record, this is removed because AWS returns `ValidationError` instead of a clear "not found" code, right?

AndiDog (commented):
Mostly LGTM – one minor comment.

Can you please fill the release note on the PR description? The release note, commit and PR title could read something like "Skip instance refresh attempt if ASG does not yet exist".

pkg/cloud/services/autoscaling/autoscalinggroup.go
fiunchinho: Exactly. It never returns an error that would match `IsNotFound()` so it's useless to have the check here.

fiunchinho:
> Mostly LGTM – one minor comment.
> 
> Can you please fill the release note on the PR description? The release note, commit and PR title could read something like "Skip instance refresh attempt if ASG does not yet exist".

I left the release note intentionally because [I added it here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4660), but I guess it's better to have a more specific one.

PR #4667, opened 2023-11-29
:bug: cleanup: eliminate log spam when using S3 secrets

thefirstofthe300:
/kind cleanup

Currently, the S3 object store generates a large amount of log spam during reconciliation. Objects are stated to be deleting regardless of objects already being deleted.

Example:
```
I1129 02:23:28.008345       1 logger.go:67] "Deleting unneeded entry from AWS S3" secretPrefix=""
I1129 02:23:28.008394       1 logger.go:67] "Deleting object" controller="awsmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSMachine" AWSMachine="clusters-system/us-west-2-production-worker-us-west-2d-infra-vln7r-xzcsk" namespace="clusters-system" name="us-west-2-production-worker-us-west-2d-infra-vln7r-xzcsk" reconcileID="c10cc649-57ce-40f4-99bb-17d06590c2cf" machine="clusters-system/us-west-2-production-worker-us-west-2d-5lgbh-6cbb9999d5xlnxtcvd" cluster="clusters-system/us-west-2-production" bucket_name="cluster-api-provider-aws-gremlin-us-west-2-production" key="node/us-west-2-production-worker-us-west-2d-infra-vln7r-xzcsk"
I1129 02:24:05.691604       1 awsmanagedcluster_controller.go:114] "Successfully reconciled AWSManagedCluster" controller="awsmanagedcluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSManagedCluster" AWSManagedCluster="clusters-system/us-east-1-production-testing" namespace="clusters-system" name="us-east-1-production-testing" reconcileID="8cbaa551-85ef-4feb-9fac-533577193662" cluster="us-east-1-production-testing" controlPlane="us-east-1-production-testing-control-plane"
```

- [x] squashed commits
- [x] includes documentation (N/A)
- [x] adds unit tests (N/A

**Release note**:
```release-note
If you're using S3 buckets for bootstrap data, the following permissions are required on the controller to automatically detect whether an object is still needing to be deleted: `s3:GetObject`. The following permissions are optional: `s3:ListBucket`.
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -850,8 +850,6 @@ func (r *AWSMachineReconciler) deleteIgnitionBootstrapDataFromS3(machineScope *s
 		return nil
 	}
 
-	machineScope.Info("Deleting unneeded entry from AWS S3", "secretPrefix", machineScope.GetSecretPrefix())
-
 	if err := objectStoreSvc.Delete(machineScope); err != nil {
 		return errors.Wrap(err, "deleting bootstrap data object")
 	}
--- a/pkg/cloud/services/s3/s3.go
+++ b/pkg/cloud/services/s3/s3.go
@@ -173,25 +173,50 @@ func (s *Service) Delete(m *scope.MachineScope) error {
 	bucket := s.bucketName()
 	key := s.bootstrapDataKey(m)
 
-	s.scope.Info("Deleting object", "bucket_name", bucket, "key", key)
-
-	_, err := s.S3Client.DeleteObject(&s3.DeleteObjectInput{
+	_, err := s.S3Client.HeadObject(&s3.HeadObjectInput{
 		Bucket: aws.String(bucket),
 		Key:    aws.String(key),
 	})
-	if err == nil {
-		return nil
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "Forbidden":
+				// In the case that the IAM policy does not have sufficient
+				// permissions to get the object, we will attempt to delete it
+				// anyway for backwards compatibility reasons.
+				s.scope.Debug("Received 403 forbidden from S3 HeadObject call. If GetObject permission has been granted to the controller but not ListBucket, object is already deleted. Attempting deletion anyway in case GetObject permission hasn't been granted to the controller but DeleteObject has.", "bucket", bucket, "key", key)
+
+				_, err = s.S3Client.DeleteObject(&s3.DeleteObjectInput{
+					Bucket: aws.String(bucket),
+					Key:    aws.String(key),
+				})
+				if err != nil {
+					return errors.Wrap(err, "deleting S3 object")
+				}
+
+				s.scope.Debug("Delete object call succeeded despite missing GetObject permission", "bucket", bucket, "key", key)
+
+				return nil
+			case s3.ErrCodeNoSuchKey:
+				s.scope.Debug("Object already deleted", "bucket", bucket, "key", key)
+				return nil
+			case s3.ErrCodeNoSuchBucket:
+				s.scope.Debug("Bucket does not exist", "bucket", bucket)
+				return nil
+			default:
+				return errors.Wrap(aerr, "deleting S3 object")
+			}
+		}
 	}
 
-	aerr, ok := err.(awserr.Error)
-	if !ok {
+	s.scope.Info("Deleting S3 object", "bucket", bucket, "key", key)
+
+	_, err = s.S3Client.DeleteObject(&s3.DeleteObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+	})
+	if err != nil {
 		return errors.Wrap(err, "deleting S3 object")
-	}
-
-	switch aerr.Code() {
-	case s3.ErrCodeNoSuchBucket:
-	default:
-		return errors.Wrap(aerr, "deleting S3 object")
 	}
 
 	return nil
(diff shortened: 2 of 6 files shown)


pkg/cloud/services/s3/s3.go
```diff
@@ -173,27 +173,35 @@ func (s *Service) Delete(m *scope.MachineScope) error {
 	bucket := s.bucketName()
 	key := s.bootstrapDataKey(m)
 
-	s.scope.Info("Deleting object", "bucket_name", bucket, "key", key)
-
-	_, err := s.S3Client.DeleteObject(&s3.DeleteObjectInput{
+	_, err := s.S3Client.HeadObject(&s3.HeadObjectInput{
```
AndiDog, line 176: `HeadObject` must be added as allowed operation, or else the controller will fail on upgrade since users don't have this allowed in their IAM policy (see `cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go` – and maybe other places??).

This should also go into release notes (your PR has no release note at the moment).

pkg/cloud/services/s3/s3_test.go
```diff
@@ -627,6 +627,8 @@ func TestDeleteObject(t *testing.T) {
 			},
 		}
 
+		s3Mock.EXPECT().HeadObject(gomock.Any()).Times(1)
```
AndiDog, line 630:
```suggestion
		s3Mock.EXPECT().HeadObject(gomock.Any())
```

since that's the same as `.Times(1)` [repeated below]

controllers/awsmachine_controller.go
```diff
@@ -850,8 +850,6 @@ func (r *AWSMachineReconciler) deleteIgnitionBootstrapDataFromS3(machineScope *s
 		return nil
 	}
 
-	machineScope.Info("Deleting unneeded entry from AWS S3", "secretPrefix", machineScope.GetSecretPrefix())
```
vincepri, line 853: Move to debug?

pkg/cloud/services/s3/s3.go
```diff
@@ -173,27 +173,38 @@ func (s *Service) Delete(m *scope.MachineScope) error {
 	bucket := s.bucketName()
 	key := s.bootstrapDataKey(m)
 
-	s.scope.Info("Deleting object", "bucket_name", bucket, "key", key)
-
-	_, err := s.S3Client.DeleteObject(&s3.DeleteObjectInput{
```
vincepri, line 178: Another thought on Deletion, would be good to have the bucket overall with an object lifecycle so that every objects gets automatically deleted after 1h (or more?). This way we have some guarantees from S3 that the objects aren't going to be staying around.

pkg/cloud/services/s3/s3.go
thefirstofthe300: S3 is weird in that not all operations require a permission of the same name. In the case of HeadObject, it only requires GetObject permission: https://docs.aws.amazon.com/AmazonS3/latest/API/API_HeadObject.html

thefirstofthe300: I didn't realize the controller didn't already have this permission. Added.

pkg/cloud/services/s3/s3.go
thefirstofthe300: With this PR, I'm trying to eliminate the log spam around these actions. I haven't noticed any issues with objects sticking around since this code path is called every time a machine is reconciled after initial creation.

Not that I'm opposed to the addition of a lifecycle policy on the bucket, but I also don't want to expand the scope of the PR beyond a clean up to adding a new feature as well.

controllers/awsmachine_controller.go
thefirstofthe300: I don't think this line needs to be moved to debug as the statement is misleading. The implication that a deletion is actually occurring is false.

The information conveyed by this line is replaced by the changes made in `s3.go`.

pkg/cloud/services/s3/s3.go
```diff
@@ -173,27 +173,38 @@ func (s *Service) Delete(m *scope.MachineScope) error {
...
-	}
-
-	aerr, ok := err.(awserr.Error)
-	if !ok {
+	if err != nil {
+		if aerr, ok := err.(awserr.Error); ok {
+			switch aerr.Code() {
+			case "Forbidden":
+				s.scope.Debug("Received 403 forbidden. If ListBucket permission has not been granted to this controller, object may already be deleted", "bucket", bucket, "key", key)
+				return nil
```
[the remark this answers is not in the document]
thefirstofthe300: Great point. I added back the delete operation in this error block but added a bit more context to the debug log statement to explain why the delete operation was happening anyways.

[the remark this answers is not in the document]
AndiDog: Good idea. We should then also handle `ErrCodeNoSuchKey` etc. instead of failing, since the object may not even exist.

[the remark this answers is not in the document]
thefirstofthe300: Ummm...it is in the next case statement?

If you're referring to deleting a non-existent object, the S3 API will return a 204 No Content regardless of the object existing or not.

All of my research indicates there isn't a way to actually determine if S3 deleted the object specified or if S3 did nothing. The return code is apparently supposed to indicate that S3 did not experience an internal error and not indicate the existence of the object prior to the deletion.

Ankitasw:
/approve

cc @AndiDog for another review

pkg/cloud/services/s3/s3.go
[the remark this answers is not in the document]
AndiDog: Oh, that deletion behavior has been bugging me every time... You're right. All good then 👍.

PR #4676, opened 2023-12-05
:sparkles: Enable transit encryption to S3 bucket

fiunchinho:
**What type of PR is this?**

/kind feature

Fixes #

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Enable transit encryption on S3 buckets
```

### DIFF
--- a/pkg/cloud/services/s3/s3.go
+++ b/pkg/cloud/services/s3/s3.go
@@ -306,6 +306,20 @@ func (s *Service) bucketPolicy(bucketName string) (string, error) {
 			Action:   []string{"s3:GetObject"},
 			Resource: []string{fmt.Sprintf("arn:%s:s3:::%s/control-plane/*", partition, bucketName)},
 		},
+		{
+			Sid:    "ForceSSLOnlyAccess",
+			Effect: iam.EffectDeny,
+			Principal: map[iam.PrincipalType]iam.PrincipalID{
+				iam.PrincipalAWS: []string{"*"},
+			},
+			Action:   []string{"s3:*"},
+			Resource: []string{fmt.Sprintf("arn:%s:s3:::%s/*", partition, bucketName)},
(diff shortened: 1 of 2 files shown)


faiq:
Does this require any changes for IAM permissions on the roles

fiunchinho:
> Does this require any changes for IAM permissions on the roles

[Official docs](https://aws.amazon.com/blogs/storage/enforcing-encryption-in-transit-with-tls1-2-or-higher-with-amazon-s3/) don't mention anything about IAM so I'd say no, it doesn't require any changes on IAM permissions.

AndiDog (commented):
/lgtm

pkg/cloud/services/s3/s3.go
```diff
@@ -306,6 +306,20 @@ func (s *Service) bucketPolicy(bucketName string) (string, error) {
...
+			Sid:    "ForceSSLOnlyAccess",
+			Effect: iam.EffectDeny,
+			Principal: map[iam.PrincipalType]iam.PrincipalID{
+				iam.PrincipalAWS: []string{"*"},
+			},
+			Action:   []string{"s3:*"},
+			Resource: []string{fmt.Sprintf("arn:%s:s3:::%s/*", partition, bucketName)},
+			Condition: iam.Conditions{
+				"Bool": map[string]interface{}{
+					"aws:SecureTransport": false,
```
richardcase, line 319: Should this be set to **true**?

fiunchinho: The `Effect` of the policy is to `deny` requests when the condition `SecureTransport: false` is met. Hence only allowing requests with TSL. Does that make sense?

richardcase: Ah, yes. That makes sense....thanks @fiunchinho

PR #4678, opened 2023-12-06
🐛: Use external CCM in default template

### LINKED ISSUE #4677: Quickstart template does not install AWS cloud-controller-manager
/kind bug

Following the CAPI quickstart with Kubernetes v1.28, no AWS cloud-controller-manager is installed and the in-tree cloud controller manager is not used, apparently - trying to follow the quickstart results in only 4 nodes (one control plane, three workers) being created, but never being `Ready`.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/templates/cluster-template.yaml sets the cloud provider to `external`, but does not contain a `Deployment` or `DaemonSet` that installs the AWS cloud-controller-manager.

Following the quickstart results in a usable cluster.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/templates/cluster-template-external-cloud-provider.yaml contains the external cloud provider information, but is not the current default template. As the cloud controller managers are being removed from the Kubernetes code base into their own repositories, we should default to using the external code. 

**Environment:**

```
clusterawsadm version: &version.Info{Major:"2", Minor:"2", GitVersion:"v2.2.4", GitCommit:"56c9a39dd834640ee4a027e679ad2a5757098dfd", GitTreeState:"clean", BuildDate:"2023-10-05T09:43:40Z", GoVersion:"go1.21.0", AwsSdkVersion:"v1.44.213", Compiler:"gc", Platform:"darwin/arm64"}
```
```
❯ k version
Client Version: v1.28.3
Kustomize Version: v5.0.4-0.20230601165947-6ce0bf390ce3
Server Version: v1.27.3
```
- OS (e.g. from `/etc/os-release`): macOS 13.6.2

nrb:
**What type of PR is this?**

The default cluster template uses the `external` cloud provider, but does not provide the necessary information to install the cloud provider.

Copies the structure of the
cluster-template-external-cloud-provider.yaml into the default template, with the folloiwng changes.

* The following version updates were made to keep up-to-date:
    aws-cloud-controller-manager: v1.28.3
    CSI driver:                   v1.25.0
    CSI provisioner:              v3.6.2
    CSI attacher:                 v4.4.2
    CSI resizer:                  v1.9.2
    CSI livenessprobe:            v2.11.0
    CSI register:                 v2.9.2

* The CNI resource references were removed, since the Quickstart lets the user install their own CNI.

* Updated the contents of the aws-ccm-external.yaml file based on files in https://github.com/kubernetes/cloud-provider-aws/tree/master/examples/existing-cluster/base. 
   * Changed `--configure-cloud-routes=false` from the default
   * Changed node selector from `node-role.kubernetes.io/master` to `node-role.kubernetes.io/controle-plane`. I was not able to see the CCM pods get scheduled until this was changed.

Fixes #4677 

NOTE: Because this change requires the use of ClusterResourceSet, the Quickstart will need to be updated to include `export EXP_CLUSTER_RESOURCE_SET=true` prior to initializing the management cluster.

**Checklist**:

- [x] squashed commits

**Release note**:

```release-note
Updated the default clusterctl template to use external cloud-controller-manager
```

### DIFF
--- a/templates/cluster-template.yaml
+++ b/templates/cluster-template.yaml
@@ -1,46 +1,38 @@
----
 apiVersion: cluster.x-k8s.io/v1beta1
 kind: Cluster
 metadata:
-  name: "${CLUSTER_NAME}"
+  labels:
+    ccm: external
+    csi: external
+  name: ${CLUSTER_NAME}
 spec:
   clusterNetwork:
     pods:
-      cidrBlocks: ["192.168.0.0/16"]
+      cidrBlocks:
+      - 192.168.0.0/16
+  controlPlaneRef:
+    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+    kind: KubeadmControlPlane
+    name: ${CLUSTER_NAME}-control-plane
   infrastructureRef:
     apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
     kind: AWSCluster
-    name: "${CLUSTER_NAME}"
-  controlPlaneRef:
-    kind: KubeadmControlPlane
-    apiVersion: controlplane.cluster.x-k8s.io/v1beta1
-    name: "${CLUSTER_NAME}-control-plane"
+    name: ${CLUSTER_NAME}
 ---
 apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
 kind: AWSCluster
 metadata:
-  name: "${CLUSTER_NAME}"
+  name: ${CLUSTER_NAME}
 spec:
-  region: "${AWS_REGION}"
-  sshKeyName: "${AWS_SSH_KEY_NAME}"
+  region: ${AWS_REGION}
+  sshKeyName: ${AWS_SSH_KEY_NAME}
 ---
-kind: KubeadmControlPlane
 apiVersion: controlplane.cluster.x-k8s.io/v1beta1
+kind: KubeadmControlPlane
 metadata:
-  name: "${CLUSTER_NAME}-control-plane"
+  name: ${CLUSTER_NAME}-control-plane
 spec:
-  replicas: ${CONTROL_PLANE_MACHINE_COUNT}
-  machineTemplate:
-    infrastructureRef:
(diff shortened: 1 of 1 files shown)


nrb:
Given this requires an experimental feature, is this an acceptable solution?

We could perhaps instruct the user to manually install the CCM, though my initial attempt of installing the CCM via [Helm](https://github.com/kubernetes/cloud-provider-aws/blob/master/charts/aws-cloud-controller-manager/Readme.md) needed to have the `--config-cloud-routes` flag changed at the least; I'm not sure if the node selector change also needs to be made, too.

vincepri (approved):
/lgtm
/assign @richardcase

nrb:
Double checking this, I can't get it to trigger the admission webhook warnings about CRSes, so I'm going to remove the breaking change warnings.

richardcase:
> Given this requires an experimental feature, is this an acceptable solution?

I think its fine. We can update the quick start guide to say CRS needs to be enabled and the webhook should also reject the CRS if its not enabled.

templates/cluster-template.yaml
```diff
@@ -1,114 +1,972 @@
...
 kind: AWSCluster
 metadata:
-  name: "${CLUSTER_NAME}"
+  name: ${CLUSTER_NAME}
 spec:
-  region: "${AWS_REGION}"
-  sshKeyName: "${AWS_SSH_KEY_NAME}"
+  network:
+    vpc:
+      availabilityZoneUsageLimit: 1
```
richardcase, line 29: Was there a reason this was added?

nrb: Mistakenly copied. I can remove if you'd like.

nrb:
@richardcase Removed the extra `network` stanza.

richardcase:
Lets make sure we change the upstream quickstart as well.

/approve

richardcase:
Adding the lgtm in that @vincepri did a while back:

/lgtm

PR #4701, opened 2023-12-18
🐛 Fix deregistering of deleted CAPI Machines

### LINKED ISSUE #4697: Control plane node replacement causes kube-apiserver access issues due to delayed AWSMachine deletion
/kind bug

During a rolling update of a KCP, I noticed that the instance was deregistered at the load balancer (NLB) as soon as the AWSMachine was deleted.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/2a790d41db929d9ffaac8d7db7d64462ac27989e/controllers/awsmachine_controller.go#L330-L332
However, there was a timeframe of about 3 minutes between CAPI Machine deletion (because of the node drain) and CAPA AWSMachine deletion, which caused the kube-apiserver access to be faulty during this timeframe.

tl;dr of what's happening:

- 🆗 scale down of a control plane
- 🆗 CAPI deletes the control plane Machine
- 🆗 the etcd member of the node will be removed immediately
- ⚡ kube-apiserver requests are broken to this control plane
- ⚡all Pods will be drained from the control plane
- ⚡ as soon as all Pods are drained CAPI will delete the InfraMachine (i.e., AWSMachine)
- 🆗 the AWSMachine controller deregisters the instance from the LB

One additional issue I can see is: as the kube-apiserver request via the LB is partially broken *all* nodes will become `NotReady` and will remove their endpoints to all Service load balancers after a while. Because of this all load balancers are inaccessible during a control plane rollout. This should never happen IMO.

I expected the AWSMachine to be deleted immediately after the CAPI Machine was deleted, so that the kube-apiserver access would not be affected. I expected the deletion process to be seamless and not cause any disruptions to my cluster.
But maybe this is something we cannot handle in CAPA but more like in CAPI.

We have mitigated the issue by using a custom controller that watches on the `pre-drain.delete.hook.machine.cluster.x-k8s.io` hook for control plane machines and deletes the corresponding AWSMachine. AFAIK the deletion of the AWSMachine will not wait for all Pods to be drained. But this is okay in our clusters as there are only our controllers and exporters running on the control plane nodes.

**Environment:**

- Cluster-api-provider-aws version: 2.0.3 (w/o using CC)
- Kubernetes version: v1.26.10
- OS: Ubuntu

<sub>Tobias Giese <tobias.giese@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

tobiasgiese:
**What type of PR is this?**

/kind bug

We have an issue that there is a gap between the CAPI Machine deletion and the AWSMachine deletion. In this timeframe the kube-apiserver to the to be deleted control plane is no longer possible, but the deregistering of the LB member will be performed only for AWSMachine deletions. This PR fixes the gap and also deregisteres the instance if the CAPI Machine is deleted.

Fixes #4697

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Resolved issue where deleted control planes were still registered in the kube-apiserver load balancer, causing potential disruptions.
```

<sub>Tobias Giese <tobias.giese@mercedes-benz.com>, Mercedes-Benz Tech Innovation GmbH, [legal info/Impressum](https://github.com/mercedes-benz/foss/blob/master/PROVIDER_INFORMATION.md)</sub>

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -867,8 +867,8 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 	elbsvc := r.getELBService(elbScope)
 
 	// In order to prevent sending request to a "not-ready" control plane machines, it is required to remove the machine
-	// from the ELB as soon as the machine gets deleted or when the machine is in a not running state.
-	if !machineScope.AWSMachine.DeletionTimestamp.IsZero() || !machineScope.InstanceIsRunning() {
+	// from the ELB as soon as the machine or infra machine gets deleted or when the machine is in a not running state.
+	if machineScope.AWSMachineIsDeleted() || machineScope.MachineIsDeleted() || !machineScope.InstanceIsRunning() {
 		if elbScope.ControlPlaneLoadBalancer().LoadBalancerType == infrav1.LoadBalancerTypeClassic {
 			machineScope.Debug("deregistering from classic load balancer")
 			return r.deregisterInstanceFromClassicLB(machineScope, elbsvc, i)
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -360,9 +360,14 @@ func (m *MachineScope) InstanceIsInKnownState() bool {
 	return state != nil && infrav1.InstanceKnownStates.Has(string(*state))
 }
 
-// AWSMachineIsDeleted checks if the machine was deleted.
+// AWSMachineIsDeleted checks if the AWS machine was deleted.
 func (m *MachineScope) AWSMachineIsDeleted() bool {
 	return !m.AWSMachine.ObjectMeta.DeletionTimestamp.IsZero()
+}
+
+// MachineIsDeleted checks if the machine was deleted.
+func (m *MachineScope) MachineIsDeleted() bool {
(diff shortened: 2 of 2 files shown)


richardcase:
Until the e2e passes:

/hold

richardcase:
@tobiasgiese (or anyone else) - when the e2e run finished feel free to do an unhold.

AndiDog (commented):
/lgtm

but as usual, changes will haunt us later if there's no test. I'll have maintainers decide here, since I'm not an expert of this part of the code.

tobiasgiese:
> but as usual, changes will haunt us later if there's no test. I'll have maintainers decide here, since I'm not an expert of this part of the code.

Indeed this should be done. But I think it's fair to say that this will be addressed in a follow up PR/issue.
WDYT @richardcase

tobiasgiese:
e2e tests passed
/unhold

tobiasgiese:
@AndiDog issue https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3820 will cover this issue IMO

tobiasgiese:
/cherrypick release-2.2

I think we should backport it @richardcase

PR #4707, opened 2023-12-20
:sparkles: Remove ingress and egress rules from vpc default security group

### LINKED ISSUE #4675: Remove ingress/egress rules from VPC default security group
/kind feature

When creating clusters CAPA creates a VPC that contains a default security group that's created automatically by AWS. This default security group contains an ingress rule and an egress rules, allowing all egress traffic, and allowing all ingress coming from the default security group.

In highly secure environments having this security group with these wide policies is not desirable. CAPA could take care of removing the ingress and egress rules from the default security group on the VPC created when reconciling a cluster.

fiunchinho:
**What type of PR is this?**

/kind feature

We want to secure the VPC as much as possible.

Fixes #4675 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Remove ingress and egress rules from vpc default security group
```

### DIFF
--- a/api/v1beta1/awscluster_conversion.go
+++ b/api/v1beta1/awscluster_conversion.go
@@ -90,7 +90,7 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 		restoreIPAMPool(restored.Spec.NetworkSpec.VPC.IPv6.IPAMPool, dst.Spec.NetworkSpec.VPC.IPv6.IPAMPool)
 	}
 
-	dst.Spec.NetworkSpec.AdditionalControlPlaneIngressRules = restored.Spec.NetworkSpec.AdditionalControlPlaneIngressRules
+	dst.Spec.NetworkSpec.VPC.EmptyRoutesDefaultVPCSecurityGroup = restored.Spec.NetworkSpec.VPC.EmptyRoutesDefaultVPCSecurityGroup
 
 	// Restore SubnetSpec.ResourceID field, if any.
 	for _, subnet := range restored.Spec.NetworkSpec.Subnets {
--- a/api/v1beta2/network_types.go
+++ b/api/v1beta2/network_types.go
@@ -323,6 +323,18 @@ type VPCSpec struct {
 	// +kubebuilder:default=Ordered
 	// +kubebuilder:validation:Enum=Ordered;Random
 	AvailabilityZoneSelection *AZSelectionScheme `json:"availabilityZoneSelection,omitempty"`
+
+	// EmptyRoutesDefaultVPCSecurityGroup specifies whether the default VPC security group ingress
+	// and egress rules should be removed.
+	//
+	// By default, when creating a VPC, AWS creates a security group called `default` with ingress and egress
+	// rules that allow traffic from anywhere. The group could be used as a potential surface attack and
+	// it's generally suggested that the group rules are removed or modified appropriately.
+	//
+	// NOTE: This only applies when the VPC is managed by the Cluster API AWS controller.
+	//
+	// +optional
+	EmptyRoutesDefaultVPCSecurityGroup bool `json:"emptyRoutesDefaultVPCSecurityGroup,omitempty"`
 }
 
 // String returns a string representation of the VPC.
--- a/pkg/cloud/services/securitygroup/securitygroups.go
+++ b/pkg/cloud/services/securitygroup/securitygroups.go
@@ -64,6 +64,11 @@ func (s *Service) ReconcileSecurityGroups() error {
 
 	var err error
 
+	err = s.revokeIngressAndEgressRulesFromVPCDefaultSecurityGroup()
+	if err != nil {
+		return err
+	}
+
 	// Security group overrides are mapped by Role rather than their security group name
 	// They are copied into the main 'sgs' list by their group name later
 	var securityGroupOverrides map[infrav1.SecurityGroupRole]*ec2.SecurityGroup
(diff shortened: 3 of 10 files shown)


api/v1beta1/awscluster_conversion.go
```diff
@@ -90,7 +90,7 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 		restoreIPAMPool(restored.Spec.NetworkSpec.VPC.IPv6.IPAMPool, dst.Spec.NetworkSpec.VPC.IPv6.IPAMPool)
 	}
 
-	dst.Spec.NetworkSpec.AdditionalControlPlaneIngressRules = restored.Spec.NetworkSpec.AdditionalControlPlaneIngressRules
```
fiunchinho, line 93: This was already done on line 75

api/v1beta2/network_types.go
```diff
@@ -323,6 +323,11 @@ type VPCSpec struct {
 	// +kubebuilder:default=Ordered
 	// +kubebuilder:validation:Enum=Ordered;Random
 	AvailabilityZoneSelection *AZSelectionScheme `json:"availabilityZoneSelection,omitempty"`
+
+	// SecureDefaultVPCSecurityGroup specifies whether the default VPC security group ingress
+	// and egress rules should be removed.
+	// +optional
+	SecureDefaultVPCSecurityGroup bool `json:"secureDefaultVPCSecurityGroup,omitempty"`
```
fiunchinho, line 330: I'd remove this field on the next major release containing breaking changes. And change CAPA behavior as if this field was always true. Should we add a @todo somewhere?

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -363,6 +368,53 @@ func (s *Service) describeSecurityGroupsByName() (map[string]infrav1.SecurityGro
...
+		ingressRules := infrav1.IngressRules{
+			{
+				Protocol:               infrav1.SecurityGroupProtocolAll,
+				FromPort:               -1,
+				ToPort:                 -1,
+				SourceSecurityGroupIDs: []string{defaultSecurityGroupID},
+			},
+		}
+		err = s.revokeSecurityGroupIngressRules(defaultSecurityGroupID, ingressRules)
+		if awserrors.IsIgnorableSecurityGroupError(errors.Cause(err)) != nil {
```
fiunchinho, line 397: The ingress and egress rules are removed successfully, but on the next reconciliations, I keep getting
```I1221 17:25:06.920228       1 securitygroups.go:386] "Removing ingress and egress rules from default security group in VPC" controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="org-giantswarm/jose4" namespace="org-giantswarm" name="jose4" reconcileID="7cee43d6-5cb8-4509-ad48-47e3c27b1dff" cluster="org-giantswarm/jose4" defaultSecurityGroupID="sg-09b4f9143ee0aeda4" vpc-id="vpc-0480fd3a1f02bc76b"```

What would be the best way to unwrap the error so that it's ignored? I figured that we don't want to do anything if the ingress or egress rules were already removed.

File: api/v1beta2/network_types.go (hunk 2 above)
Comment (AndiDog) [reply api/v1beta2/network_types.go:330]: At least, we should add here in the description that the default will become true in the future, while currently it's false. Regarding a TODO, an issue could be opened for the next milestone.

File: pkg/cloud/services/securitygroup/securitygroups.go [hunk 4]
```diff
@@ -363,6 +368,53 @@ func (s *Service) describeSecurityGroupsByName() (map[string]infrav1.SecurityGro
 	return res, nil
 }
 
+// revokeIngressAndEgressRulesFromVPCDefaultSecurityGroup revokes ingress and egress rules from the VPC default security group.
+// The VPC default security group is created by AWS and cannot be deleted.
+// But we can revoke all ingress and egress rules from it to make it more secure. This security group is not used by CAPA.
+func (s *Service) revokeIngressAndEgressRulesFromVPCDefaultSecurityGroup() error {
+	if s.scope.VPC().SecureDefaultVPCSecurityGroup {
```
AndiDog, line 375: Please return early instead of indenting the whole function code. Or alternatively, you could move the condition to the caller.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -363,6 +368,53 @@ func (s *Service) describeSecurityGroupsByName() (map[string]infrav1.SecurityGro
...
+		securityGroups, err := s.EC2Client.DescribeSecurityGroupsWithContext(context.TODO(), &ec2.DescribeSecurityGroupsInput{
+			Filters: []*ec2.Filter{
+				filter.EC2.VPC(s.scope.VPC().ID),
+				filter.EC2.SecurityGroupName("default"),
+			},
+		})
+		if err != nil {
```
AndiDog, line 382: `len(securityGroups) != 1` is also an error

pkg/cloud/services/securitygroup/securitygroups.go
AndiDog: We can use the same logic as `func (s *Service) ReconcileSecurityGroups`: first read existing rules with `s.getSecurityGroupIngressRules` and then reconcile the differences.

pkg/cloud/services/securitygroup/securitygroups_test.go
```diff
@@ -90,6 +92,29 @@ func TestReconcileSecurityGroups(t *testing.T) {
 				},
 			},
 			expect: func(m *mocks.MockEC2APIMockRecorder) {
```
AndiDog, line 94: We should also have a test for reconciliation if the rules are already gone – in that case, no requests apart from `DescribeSecurityGroups` and listing its rules should be done.

api/v1beta2/network_types.go
vincepri: It might be nice if this is `true` for new clusters by default, but it stays `false` for older clusters. The defaulting can happen if in the validation webhook if the vpc id of a managed VPC is already filled.

Part of me was also considering of not adding this flag at all, which can be confusing?

fiunchinho: I'd prefer not adding the field/flag, because it makes things simpler and easier to understand. I don't think users should rely on this security group anyway. But someone in Slack suggested adding it to be backwards compatible, which also makes sense.

pkg/cloud/services/securitygroup/securitygroups.go
fiunchinho: The `default` security group can't be removed. When creating a new security group, group names must be unique, so there can't be two groups named `default`. I don't think `len(securityGroups)` can ever be `!= 1`.

pkg/cloud/services/securitygroup/securitygroups.go
fiunchinho: I don't think it makes sense in this case, because all we want is to delete the existing rules. We don't care what's inside, or "the differences". We want the group to have no rules at all.

pkg/cloud/services/securitygroup/securitygroups.go
fiunchinho: Done

api/v1beta2/network_types.go
vincepri, line 330:
```suggestion

	// EmptyRoutesDefaultVPCSecurityGroup specifies whether the default VPC security group ingress
	// and egress rules should be removed.
	//
	// By default, when creating a VPC, AWS creates a security group called `default` with ingress and egress
	// rules that allow traffic from anywhere. The group could be used as a potential surface attack and
	// it's generally suggested that the group rules are removed or modified appropriately.
	//
	// NOTE: This only applies when the VPC is managed by the Cluster API AWS controller.
	//
	// +optional
	EmptyRoutesDefaultVPCSecurityGroup bool `json:"emptyRoutesDefaultVPCSecurityGroup,omitempty"`
```

pkg/cloud/services/securitygroup/securitygroups_test.go
fiunchinho: Added a test case to cover that.

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -363,6 +368,55 @@ func (s *Service) describeSecurityGroupsByName() (map[string]infrav1.SecurityGro
...
+		{
+			Protocol:               infrav1.SecurityGroupProtocolAll,
+			FromPort:               -1,
+			ToPort:                 -1,
+			SourceSecurityGroupIDs: []string{defaultSecurityGroupID},
+		},
+	}
+	err = s.revokeSecurityGroupIngressRules(defaultSecurityGroupID, ingressRules)
+	if awserrors.IsIgnorableSecurityGroupError(errors.Cause(err)) != nil {
+		return errors.Wrapf(err, "failed to revoke ingress rules from vpc default security group in vpc %q", s.scope.VPC().ID)
```
AndiDog, line 401:
```suggestion
		return errors.Wrapf(err, "failed to revoke ingress rules from vpc default security group %q in VPC %q", defaultSecurityGroupID, s.scope.VPC().ID)
```

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -363,6 +368,55 @@ func (s *Service) describeSecurityGroupsByName() (map[string]infrav1.SecurityGro
...
+	ingressRules := infrav1.IngressRules{
+		{
+			Protocol:               infrav1.SecurityGroupProtocolAll,
+			FromPort:               -1,
+			ToPort:                 -1,
+			SourceSecurityGroupIDs: []string{defaultSecurityGroupID},
+		},
+	}
+	err = s.revokeSecurityGroupIngressRules(defaultSecurityGroupID, ingressRules)
+	if awserrors.IsIgnorableSecurityGroupError(errors.Cause(err)) != nil {
```
AndiDog, line 400: Are `GroupNotFound`/`PermissionNotFound` (which `IsIgnorableSecurityGroupError` handles) really errors that we want to ignore? The default group cannot be deleted, so such errors here should fail, I think. From the test, it seems that you found that AWS throws a permission error if the given rules do not exist. So we could instead use `ec2:DescribeSecurityGroupRules` (new permission which CAPA does not have yet) or handle the very specific error?

fiunchinho: After deleting the ingress/egress rules from the default security group, the next time we try to delete them we'll get a `PermissionNotFound` error. I think this one can be ignored, because that's the state we want. To avoid swallowing the `GroupNotFound` error, that should not happen in this scenario, we could handle `PermissionNotFound` more specifically, like you mentioned. I'll add a new `IsPermissionNotFoundError` function. How does that sound?

AndiDog (commented):
/lgtm

vincepri (approved):

AndiDog:
@vincepri did you mean to `/approve` instead of GitHub "approve"?

PR #4710, opened 2023-12-21
✨ Use go 1.21.5, fix kubectl version detection after `--short` parameter was removed

fiunchinho:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Use latest Go version.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Use go 1.21.5
```

PR #4720, opened 2024-01-03
🐛 Updating verify kubectl version script

### LINKED ISSUE #4719: New test-infra images have kubectl updated to v1.29, causing error in CI
/kind bug

[With the update of the test-infra images](https://github.com/kubernetes/test-infra/pull/31499), we're facing [errors](https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/4710/pull-cluster-api-provider-aws-e2e/1737890517250740224) in the CI.

The root cause is the presence of the `kubectl` binary updated to the 1.29 version, which has deprecated the `--short` options for the `kubectl version` command.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/bba04c43cd7a25b1384aa22a2ebcb65a8da97c4b/hack/ensure-kubectl.sh#L45

Having the local CI scripts with the removal of the deprecated `kubectl version` flag.

N.R.

**Environment:**

- Cluster-api-provider-aws version: N.R.
- Kubernetes version: (use `kubectl version`): N.R.
- OS (e.g. from `/etc/os-release`): N.R.

prometherion:
**What type of PR is this?**

/kind bug

Fixes #4719

This PR is a pre-requisite to fix the blocked CI of #4710. 

Essentially, the `--short` flag has been deprecated and removed from v1.29.
To keep backward compatibility, although the infra image has already `kubectl` baked in, the `--output=yaml` flag is used as a replacement.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/ensure-kubectl.sh
+++ b/hack/ensure-kubectl.sh
@@ -14,27 +14,38 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+# This has been copied from https://github.com/kubernetes-sigs/cluster-api/blob/v1.6.0/hack/ensure-kubectl.sh
+
 set -o errexit
 set -o nounset
 set -o pipefail
 
+if [[ "${TRACE-0}" == "1" ]]; then
+    set -o xtrace
+fi
+
+
+# shellcheck source=./hack/utils.sh
+source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
+
 GOPATH_BIN="$(go env GOPATH)/bin/"
-MINIMUM_KUBECTL_VERSION=v1.19.0
-GOARCH="$(go env GOARCH)"
-GOOS="$(go env GOOS)"
+MINIMUM_KUBECTL_VERSION=v1.28.0
+goarch="$(go env GOARCH)"
+goos="$(go env GOOS)"
 
 # Ensure the kubectl tool exists and is a viable version, or installs it
 verify_kubectl_version() {
 
   # If kubectl is not available on the path, get it
   if ! [ -x "$(command -v kubectl)" ]; then
-    if [ "$GOOS" == "linux" ] || [ "$GOOS" == "darwin" ]; then
+    if [ "$goos" == "linux" ] || [ "$goos" == "darwin" ]; then
       if ! [ -d "${GOPATH_BIN}" ]; then
         mkdir -p "${GOPATH_BIN}"
       fi
       echo 'kubectl not found, installing'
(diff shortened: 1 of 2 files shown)


richardcase:
Thanks for this @prometherion. I had just done the same, but you beat me to it :)

What do you think to bringing the script more in line with the [upstream CAPI version](https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/ensure-kubectl.sh)? Which will also require a small change to utils.sh (i.e. add the **verify_gopath_bin** function).

prometherion:
Thanks, @richardcase, I just aligned the files as you suggested.

richardcase:
I think after rebasing because of #4710 this will go green

prometherion:
Correct, thanks @richardcase!

hack/utils.sh
```diff
@@ -25,3 +25,27 @@ cd_root_path() {
     cd "$(get_root_path)" || exit
 }
 
+# get_capd_root_path returns the root path of CAPD source tree
+get_capd_root_path() {
```
richardcase, line 29: nit: i don't think we need the capd functions

prometherion: I thought we would just need to copy the whole file, I can amend.

hack/ensure-kubectl.sh
```diff
@@ -14,35 +14,46 @@
...
+
+# shellcheck source=./hack/utils.sh
+source "$(dirname "${BASH_SOURCE[0]}")/utils.sh"
+
 GOPATH_BIN="$(go env GOPATH)/bin/"
-MINIMUM_KUBECTL_VERSION=v1.19.0
-GOARCH="$(go env GOARCH)"
-GOOS="$(go env GOOS)"
+MINIMUM_KUBECTL_VERSION=v1.28.0
+goarch="$(go env GOARCH)"
```
Ankitasw, line 33: Please revert back to GOARCH and GOOS

prometherion: May I ask for the reason? We're doing a bare copy/paste of the source file, if there's a specific reason it would be great having an explicit statement for future changes.

Ankitasw: I didnt understand this change, and what affect would it make differently than the previous change

vincepri: Technically the change above might be more in line with bash common practices, where userspace variables are generally lowercase.

I'll leave the decision up to @Ankitasw though

Ankitasw: Sure makes sense to me if this adheres to the best practice 👍

PR #4723, opened 2024-01-05
🐛 Make VPC creation idempotent to avoid indefinite creation of new VPCs if storage of the ID fails

AndiDog:
**What type of PR is this?**

/kind bug

In case of a repeating storage error, for example

```
"Reconciler error" err="failed to reconcile network for AWSManagedControlPlane mynamespace/mycluster: failed to patch conditions: AWSManagedControlPlane.controlplane.cluster.x-k8s.io \"mycluster\" is invalid: spec.network.subnets[7]: Duplicate value: map[string]interface {}{\"id\":\"\"}" [...]
```

which was caused by something totally different, CAPA would create an indefinite number of VPCs (and potentially other resources depending how far reconciliation succeeds!), filling up the AWS account to its limits and requiring manual cleanup of the whole mess (the one case I had wasn't fun 😆).

This is terrifying but based on a very simple, horrible bug: VPC creation wasn't idempotent. This PR introduces the typical look-up-else-create logic.

Unfortunately, VPC creation tests were faulty because the `mockCtrl` object was shared across all test cases, so the expected mock calls of test case A could be "used" by a unrelated test case B, potentially making a test case pass that normally should fail. Or in short, we weren't testing what we thought we were. Also, some test cases had a description differing from the test, and made no sense overall, so I removed those. The tests now describe that VPC creation should happen only once. On top, since some tests were related to IPv6, I added a check to see if the AWS VPC IPv6 support matches our spec, or fail on mismatch. This made all tests green again. To ensure that we test the right thing in the future, test cases now check for specific errors.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Make VPC creation idempotent to avoid indefinite creation of new VPCs if storage of the ID fails
```

### DIFF
--- a/pkg/cloud/services/network/vpc.go
+++ b/pkg/cloud/services/network/vpc.go
@@ -103,23 +103,44 @@ func (s *Service) reconcileVPC() error {
 		return nil
 	}
 
-	// .spec.vpc.id is nil, Create a new managed vpc.
+	// .spec.vpc.id is nil. This means no managed VPC exists or we failed to save its ID before. Check if a managed VPC
+	// with the desired name exists, or if not, create a new managed VPC.
+
+	vpc, err := s.describeVPCByName()
+	if err == nil {
+		// An VPC already exists with the desired name
+
+		if !vpc.Tags.HasOwned(s.scope.Name()) {
+			return errors.Errorf(
+				"found VPC %q which cannot be managed by CAPA due to lack of tags (either tag the VPC manually with `%s=%s`, or provide the `vpc.id` field instead if you wish to bring your own VPC as shown in https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure)",
+				vpc.ID,
+				infrav1.ClusterTagKey(s.scope.Name()),
+				infrav1.ResourceLifecycleOwned)
+		}
+	} else {
+		if !awserrors.IsNotFound(err) {
+			return errors.Wrap(err, "failed to describe VPC resources by name")
+		}
+
+		// VPC with that name does not exist yet. Create it.
+		vpc, err = s.createVPC()
+		if err != nil {
+			return errors.Wrap(err, "failed to create new managed VPC")
+		}
+		s.scope.Info("Created VPC", "vpc-id", vpc.ID)
+	}
+
+	s.scope.VPC().CidrBlock = vpc.CidrBlock
+	s.scope.VPC().IPv6 = vpc.IPv6
+	s.scope.VPC().Tags = vpc.Tags
+	s.scope.VPC().ID = vpc.ID
+
 	if !conditions.Has(s.scope.InfraCluster(), infrav1.VpcReadyCondition) {
 		conditions.MarkFalse(s.scope.InfraCluster(), infrav1.VpcReadyCondition, infrav1.VpcCreationStartedReason, clusterv1.ConditionSeverityInfo, "")
 		if err := s.scope.PatchObject(); err != nil {
 			return errors.Wrap(err, "failed to patch conditions")
 		}
 	}
-	vpc, err := s.createVPC()
-	if err != nil {
-		return errors.Wrap(err, "failed to create new vpc")
-	}
-	s.scope.Info("Created VPC", "vpc-id", vpc.ID)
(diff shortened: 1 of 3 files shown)


pkg/cloud/services/network/vpc.go
```diff
@@ -103,24 +107,45 @@ func (s *Service) reconcileVPC() error {
...
+	// .spec.vpc.id is nil. This means no managed VPC exists or we failed to save its ID before. Check if a managed VPC
+	// with the desired name exists, or if not, create a new managed VPC.
+
+	vpc, err := s.describeVPCByName()
+	if err == nil {
+		// An VPC already exists with the desired name
+
+		if !vpc.Tags.HasOwned(s.scope.Name()) {
+			return errors.Errorf("an unmanaged VPC %s already exists, refusing to create another managed VPC", vpc.ID)
 		}
```
vincepri, line 119: The VPC today is created with tags within the same call, the error message is confusing to users given that if there are no tags, there isn't another managed VPC
```suggestion
		if !vpc.Tags.HasOwned(s.scope.Name()) {
			return errors.Errorf("found VPC named %s which cannot be managed by CAPA due to lack of tags, either tag the VPC manually, or provide the `vpc.id` field instead if you wish to bring your own VPC", vpc.ID)
		}
```

pkg/cloud/services/network/vpc.go
```diff
@@ -103,24 +107,45 @@ func (s *Service) reconcileVPC() error {
...
+
+		if s.scope.VPC().IsIPv6Enabled() != (vpc.IPv6 != nil) {
```
vincepri, line 121: split this in two lines for readability?

AndiDog: done in both locations

pkg/cloud/services/network/vpc.go
AndiDog: done

pkg/cloud/services/network/vpc.go
```diff
@@ -54,6 +54,11 @@ func (s *Service) reconcileVPC() error {
 			return errors.Wrap(err, ".spec.vpc.id is set but VPC resource is missing in AWS; failed to describe VPC resources. (might be in creation process)")
 		}
 
+		actualIPv6Enabled := (vpc.IPv6 != nil)
+		if s.scope.VPC().IsIPv6Enabled() != actualIPv6Enabled {
+			return errors.Errorf("IPv6 support of found unmanaged VPC %s differs from desired spec. Changing IP family is currently not supported.", vpc.ID)
+		}
```
vincepri, line 60: This seems a net-new check (repeated in the other function as well); should we remove this for now and add it in a different PR? Generally these checks would be best in a webhook at Update time

AndiDog: The webhook already checks that IPv6 can't be toggled later:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a1918aa166c219fbd64d32b30d01696db3fd79f7/controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go#L162-L165

IIRC, my change was to get a clear error message in unit tests, since they can simulate mismatching situations such as someone manually having turned on IPv6 via AWS Console. In such scenarios, the webhook wouldn't be called, but we still want an error instead of being silent about not reconciling the difference. I don't think this would be a user-facing breakage since normally, the reconciled state (here: IPv6 support on/off in the VPC) matches the spec.

Ankitasw: Agree with @vincepri here, it's better to do it in separate PR.

AndiDog:
E2E tests didn't work some weeks ago, it seems. Rebased onto `main` to see if that fixes the problem.

pkg/cloud/services/network/vpc.go
AndiDog: I removed the 2 occurrences of the check, and the new test case for that. I'll take note to re-add this in a separate PR, potentially targeting a later milestone.

vincepri (approved):
/lgtm
/assign @Ankitasw

PR #4686, opened 2023-12-12
✨  Rosa machinepools

enxebre:
**What type of PR is this?**
Add initial support for ROSAMachinePools

Includes https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4665

**What this PR does / why we need it**:
This PR introduces basic support to create/delete ROSAMachinePools
It also add several improvements to the rosa control plane to get cluster creation functional:
    - Add ROSAControlPlaneReadyCondition
    - Add helpers for ocmclient
    - Add WorkerRoleARN field to the API to satisfy latest ocm API requirements

Sample:
```
---
apiVersion: cluster.x-k8s.io/v1beta1
kind: MachinePool
metadata:
  name: replace-me
  namespace: replace-me
spec:
  clusterName: replace-me
  replicas: 2
  template:
    spec:
      clusterName: replace-me
      bootstrap:
        dataSecretName: ""
      infrastructureRef:
        apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
        kind: ROSAMachinePool
        name: replace-me
---
apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
kind: ROSAMachinePool
metadata:
  name: replace-me
  namespace: replace-me
spec:
  aws:
    instanceType: "m5.xlarge"
    subnet: "replace-me"
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4429

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Introduce basic support for ROSAMachinePools

```

PR #4725, opened 2024-01-05
✨ ROSA machinePools support

### LINKED ISSUE #4429: ROSA: rosaMachinePool for data plane
/kind feature

[This](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4344) introduces a common ground for collaboration on ROSA CAPA and basic ability to spin up a new cluster.

As a follow up we want to provide explore automated machine management for the data plane. For that we'll most likely want to have a new CAPA rosaMachinePool abstraction to consume the OCM machinePool API.

**Environment:**

muraee:
/kind feature

supersedes https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4686

Add initial support for ROSAMachinePools

What this PR does / why we need it:
This PR introduces basic support to create/delete ROSAMachinePoolsand a new cluster template to work with ROSAMachinePools.

It also add several improvements and fixes to the rosa control plane to get cluster creation functional:
- Add ROSAControlPlaneReadyCondition
- Introduce rosaClient helper to communicate with ocm API
- Add WorkerRoleARN field to the API to satisfy latest ocm API requirements
- Add RosaClusterName field to the API to satisfy ocm API requirements for cluster names
- Read credentials(token) from a referenced secret, e.g.:
 ```yaml
    kind: ROSAControlPlane
    metadata:
      name: "capi-rosa-quickstart-control-plane"
    spec:
      credentialsSecretRef:
        nmae: rosa-creds-secret
    ...
```

Fixes #4429, #4460

**Checklist**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Introduce basic support for ROSAMachinePools
```


muraee:
cc @vincepri this is the updated PR for machinePools

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -169,9 +170,47 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		}
 	}
 
+	// TODO: token should be read from a secret: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4460
+	token := os.Getenv("OCM_TOKEN")
```
vincepri, line 174: Can this be implemented before we merge this? Seems like a trivial fix to read the token from a plain secret

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -169,9 +170,47 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
+	rosaClient, err := rosa.NewRosaClient(token)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to create a rosa client: %w", err)
+	}
+
+	defer func() {
+		rosaClient.Close()
+	}()
```
vincepri, line 182:
```suggestion
	defer rosaClient.Close()
```

exp/controllers/rosamachinepool_controller.go
```diff
@@ -0,0 +1,310 @@
...
+	machinePoolScope.Info("Reconciling RosaMachinePool")
+
+	if controllerutil.AddFinalizer(machinePoolScope.RosaMachinePool, expinfrav1.RosaMachinePoolFinalizer) {
+		if err := machinePoolScope.PatchObject(); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	// TODO: token should be read from a secret: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4460
+	token := os.Getenv("OCM_TOKEN")
```
vincepri, line 164: same as above, seems we can also reuse this code?

pkg/rosa/client.go
```diff
@@ -0,0 +1,59 @@
...
+	return &rosaClient{
+		ocm: connection,
+	}
+}
+
+func NewRosaClient(token string) (*rosaClient, error) {
+	ocmAPIUrl := os.Getenv("OCM_API_URL")
+	if ocmAPIUrl == "" {
+		ocmAPIUrl = "https://api.openshift.com"
+	}
```
vincepri, line 25: Given that this is a new implementation in tree, can we have the RosaClient here directly read the secret, and eventually fallback to these environment variables?

muraee: sure

Ankitasw:
@muraee looks like unit tests are failing, could you fix it?

muraee:
@Ankitasw fixed! I was chasing down this error since yesterday. seems like one of the new packages we import is initilizing klog somewhere and then we end up with a panic: `flag x is redefined`

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -171,7 +173,7 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 
 	// Create the cluster:
 	clusterBuilder := cmv1.NewCluster().
-		Name(rosaScope.ControlPlane.Name).
+		Name(rosaScope.ControlPlane.Name[:15]).
```
stevekuznetsov: This requires a length check

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -426,13 +407,66 @@ func (r *ROSAControlPlaneReconciler) rosaClusterToROSAControlPlane(log *logger.L
...
-	clusterKey := rosaScope.ControlPlane.Name
+// OCMClient is a temporary helper to talk to OCM API.
+// TODO(alberto): vendor this from https://github.com/openshift/rosa/tree/master/pkg/ocm or build its own package here.
+type OCMClient struct {
+	ocm *sdk.Connection
+}
+
+func newOCMClient() (*OCMClient, error) {
+	// Create the connection, and remember to close it:
+	token := os.Getenv("OCM_TOKEN")
```
stevekuznetsov: I think we'd be better of reading this kind of thing at startup and passing content down the stack normally, as opposed to doing something like this in helpers.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -426,13 +407,66 @@ func (r *ROSAControlPlaneReconciler) rosaClusterToROSAControlPlane(log *logger.L
...
+	ocmClient := OCMClient{ocm: connection}
+
+	return &ocmClient, nil
+}
+
+func (client *OCMClient) Close() error {
+	return client.ocm.Close()
+}
+
+func (client *OCMClient) CreateCluster(clusterSpec *cmv1.Cluster) (*cmv1.Cluster, error) {
```
stevekuznetsov: Do we really need to build a further abstraction on top of the existing client SDK?

pkg/cloud/scope/rosamachinepool.go
```diff
@@ -0,0 +1,183 @@
...
+	return s.RosaMachinePool.Name
+}
+
+// NodePoolName returns the nodePool name of this machine pool.
+func (s *RosaMachinePoolScope) NodePoolName() string {
+	return s.RosaMachinePool.Spec.NodePoolName
+}
+
+// ClusterName returns the cluster name.
+func (s *RosaMachinePoolScope) RosaClusterName() string {
```
stevekuznetsov, line 113: Why do we need named methods to access public fields?

pkg/cloud/scope/rosamachinepool.go
```diff
@@ -0,0 +1,183 @@
...
+		err,
+	)
+	if err := s.PatchObject(); err != nil {
+		return errors.Wrap(err, "failed to mark rosa machinepool not ready")
+	}
+	return nil
+}
+
+// PatchObject persists the control plane configuration and status.
+func (s *RosaMachinePoolScope) PatchObject() error {
```
stevekuznetsov, line 163: IMO implicit calls without inputs are pretty hard to reason about.

pkg/cloud/scope/rosamachinepool.go
```diff
@@ -0,0 +1,183 @@
...
+	return s.patchHelper.Patch(
+		context.TODO(),
```
stevekuznetsov, line 165: We should not write new code that does this - thread the context please.

pkg/cloud/scope/rosamachinepool.go
```diff
@@ -0,0 +1,183 @@
...
+		MachinePool:     params.MachinePool,
+		controllerName:  params.ControllerName,
+	}, nil
+}
+
+// RosaMachinePoolScope defines the basic context for an actuator to operate upon.
+type RosaMachinePoolScope struct {
+	logger.Logger
+	client.Client
+	patchHelper                *patch.Helper
```
stevekuznetsov, line 91: What is `*patch.Helper`? Why not just use SSA?

pkg/rosa/nodepools.go
```diff
@@ -0,0 +1,67 @@
+package rosa
+
+import cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
+func (c *rosaClient) CreateNodePool(clusterID string, nodePool *cmv1.NodePool) (*cmv1.NodePool, error) {
```
stevekuznetsov, line 5: Why do we need these wrappers?

pkg/rosa/client.go
```diff
@@ -1,27 +1,56 @@
...
+		ocm:       connection,
+		rosaScope: rosaScope,
 	}
 }
 
-func NewRosaClient(token string) (*rosaClient, error) {
-	ocmAPIUrl := os.Getenv("OCM_API_URL")
-	if ocmAPIUrl == "" {
-		ocmAPIUrl = "https://api.openshift.com"
+func NewRosaClient(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope) (*rosaClient, error) {
```
stevekuznetsov, line 32: Why are we doing the lookup inside here instead of loading config at startup and using it as needed?

Do we need to handle re-loading this value if the Secret changes?

pkg/cloud/scope/rosamachinepool.go
muraee: what do you mean by new code?

pkg/cloud/scope/rosamachinepool.go
muraee: patch helper is a util provided by CAPI that takes care of doing some checks and patching status among other things. It uses SSA under the hood

pkg/cloud/scope/rosamachinepool.go
muraee: Its just a shorthand and commonly used pattern across the repo.

pkg/rosa/client.go
muraee: the secret is referenced through the API, so we can only read it in the reconciliation loop.
Also its content could be changed, so this ensures we always fetch the latest version.

stevekuznetsov: Are we re-loading this and re-creating the client + connections on every single loop? Wouldn't we prefer to load it once and keep track of changes? inotify watch should suffice

pkg/rosa/nodepools.go
muraee: Well, the ocm sdk is not really intuitive to use directly.
Also it helps keep the controller reconciliation code cleaner. We should actually move the building of the Cluster and NodePool ocm objects out of the controller as well.

pkg/cloud/scope/rosamachinepool.go
muraee: I think why no body cared to thread this, is that controller-runetime ensures that only 1 thread is handling an object reconciliation at a time.
But yeah I see your point, I will add a `context` to this func.

pkg/rosa/client.go
muraee: I think watching for secret content changes produces more overhead and code complexity, given that reading the secret every loop is not really a problem with caching enabled.

We would also need a different mechanism to detect if `.spec.CredentialsSecreRef` has changed.
It is possible that `.spec.CredentialsSecreRef` has been updated in the new reconciliation loop, so we need to keep reading the field and fetch the secret.

stevekuznetsov: That seems super wasteful from a resources standpoint - open connections are costly. What's our expected rate of change here?

pkg/rosa/nodepools.go
stevekuznetsov: Perhaps we send a PR to the SDK repo? I don't think the upstream cluster-api-provider-aws repo is a good place to wrap the SDK to make it more intuitive.

muraee: This is not just a wrapper, it provides reusable functions, like `GetCluster()` and `handleError()`. It serves to decuple the controller logic from the implementation details of the underlying sdk (ocm). As I said we should be moving all ocm specific logic here, like constructing the ocm cluster spec.
Obviously this will contain more helpful functionalities as we add more features.

stevekuznetsov: Generic error handling in a wrapper seems like a useful and novel addition - no argument there.

As for the rest of it, I would question the necessity. This is a leaky abstraction that requires an ever-increasing parameter list to handle the options in the SDK that the users need, or copy-paste. See your `GetNodePools` and `GetNodePool` methods. Why do we need both? Is it that onerous to pass an integer to the SDK?

As a consideration - why don't we wrap a `*kubernetes.Clientset` with some method like

```go
type wrapper struct {
    delegate *kubernetes.Clientset
}

func (w *wrapper) GetPod(namespace, name string) (*corev1.Pod, error) {
    return w.delegate.Core().V1().Pods(namespace).Get(context.TODO(), name, metav1.GetOptions{})
}
```

From my perspective, if `*rosaClient` is justified, the `*wrapper` above would be, too.

Let's get the PR in but reconsider if we truly need this in the final versions.

pkg/rosa/client.go
stevekuznetsov: If `rosaScope.Client.Get` is using a lister cache, this point is moot and we can iterate on it if we see a performance problem. Many Kubernetes controllers use custom caches (see the PKI caches in core, or every single controller using a lister), so I don't know how dogmatic the stance is on in-memory state between loops. :)

vincepri (approved):
/approve
/lgtm

PR #4732, opened 2024-01-11
:bug: S3 Bucket should be created in the same region, always add transport encryption policy

vincepri:
**What type of PR is this?**

/kind bug

Currently the S3 bucket is always created in us-east-1, regardless of where the cluster is located. In addition, this PR always ensures that we add a policy to ensure transport encryption is enabled, even when using presigned URLs.

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
When using Ignition, the S3 bucket was previously created in us-east-1 regardless of where the cluster was located, new S3 buckets will be created within the same region as the cluster.
```


AndiDog (commented):
/lgtm

PR #4733, opened 2024-01-11
✨ Add support for a secondary control plane load balancer

### LINKED ISSUE #4554: Add the ability to add a secondary load balancer to the control plane
/kind feature
/assign
/priority important-soon

In a lot of topologies, the control plane has a public and internal load balancer. Currently the AWSCluster CRD offers only ways to configure a single control plane.

Given the long term proposal for an AWSNetwork CRD to address the current shortcomings of AWSCluster, I'd like to propose to add a way to create and attach also a _secondary_ load balancer, configured by the user, and limited only to NLB or ALB (aka elbv2, not classic) load balancer types.

There are a few options we can consider, currently leaning towards the first one:

- [preferred] Add a `SecondaryControlPlaneLoadBalancer` field alongside the `ControlPlaneLoadBalancer` one and take into account of the new field and behavior in the current elb scopes and reconcilers.
  - This would be a backward compatible change.
  - The code changes should not be too invasive, and can be made in a next minor release.
  - Compatible with ClusterClass and SSA.
- Add a list of `AdditionalControlPlaneLoadBalancers` alongside the `ControlPlaneLoadBalancer`.
  - Requires a CRD breaking change to support SSA and ClusterClass if we want to reuse the struct in `ControlPlaneLoadBalancer`, given that the `name` is optional.
  - Unclear if more than two load balancers for the control plane is useful at all.
  - Handling a large number of load balancer and storing their information might be tricky.

cc @richardcase @JoelSpeed @nrb

nrb:
**What type of PR is this?**
/kind feature

Certain topologies of clusters may use an internal load balancer for the API server for a few reasons. A primary reason is to keep traffic within the cloud provider in order to reduce egress costs.

This PR adds support for a secondary control plane load balancer, internal to the Kubernetes cluster for such a use case.

Fixes #4554 

Previously, it was believed that security groups could not be added to a network load balancer, which meant that there were exceptions and exclusions around NLBs in the code. This is only true if the load balancer was created without any security groups - SGs cannot be added after creation. If the network load balancer is created with at least one security group, more can be added or removed.

See the [Network Load Balancer docs](https://docs.aws.amazon.com/elasticloadbalancing/latest/network/load-balancer-security-groups.html) for more details.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Adds support for a secondary, internal API server load balancer within clusters.
```


nrb:
I'm now able to create a 2nd load balancer within my AWS clusters.

Screenshot 2024-01-18 at 6 20 06 PM


The target group associated with it is also healthy.

Screenshot 2024-01-18 at 6 21 12 PM

I just have rebasing, writing automated tests, and writing docs left before I can remove the WIP.

nrb:
Confirming that nodes within the cluster can reach the load balancer:

```
✦ ❯ kubectl --kubeconfig $HOME/secondlb.kubeconfig attach curl-nrb -c curl-nrb -it
If you don't see a command prompt, try pressing enter.
[ root@curl-nrb:/ ]$ curl -k https://numbertwo-e4d5b20bb984138d.elb.us-east-2.amazonaws.com:6443
{
  "kind": "Status",
  "apiVersion": "v1",
  "metadata": {},
  "status": "Failure",
  "message": "forbidden: User \"system:anonymous\" cannot get path \"/\"",
  "reason": "Forbidden",
  "details": {},
  "code": 403
}[ root@curl-nrb:/ ]$
```

api/v1beta1/awscluster_conversion.go
```diff
@@ -44,6 +44,15 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 	}
 	restoreControlPlaneLoadBalancerStatus(&restored.Status.Network.APIServerELB, &dst.Status.Network.APIServerELB)
 
+	if restored.Spec.SecondaryControlPlaneLoadBalancer != nil {
+		if dst.Spec.SecondaryControlPlaneLoadBalancer == nil {
+			dst.Spec.SecondaryControlPlaneLoadBalancer = &infrav2.AWSLoadBalancerSpec{}
+		}
+		restoreControlPlaneLoadBalancer(restored.Spec.SecondaryControlPlaneLoadBalancer, dst.Spec.SecondaryControlPlaneLoadBalancer)
+	}
+	//TODO(nrb): Do we need this for secondary LBs? Probably.
```
vincepri, line 53: Yeah we do, because it didn't exist in v1beta1

api/v1beta2/awscluster_webhook.go
```diff
@@ -174,6 +133,55 @@ func (r *AWSCluster) ValidateUpdate(old runtime.Object) (admission.Warnings, err
...
 }
 
+func (r *AWSCluster) validateControlPlaneLoadBalancerUpdate(oldlb, newlb *AWSLoadBalancerSpec) field.ErrorList {
+	var allErrs field.ErrorList
+
+	// TODO(nrb): The validate-on-update default test was failing when newlb was nil
+	// Upon investigation, both values were nil, which is unlikely in a real world scenario
+	if oldlb == nil && newlb == nil {
+		return nil
+	}
```
vincepri, line 143:
```
newLoadBalancer := &AWSLoadBalancerSpec{}
	existingLoadBalancer := &AWSLoadBalancerSpec{}

	if r.Spec.ControlPlaneLoadBalancer != nil {
		newLoadBalancer = r.Spec.ControlPlaneLoadBalancer.DeepCopy()
	}

	if oldC.Spec.ControlPlaneLoadBalancer != nil {
		existingLoadBalancer = oldC.Spec.ControlPlaneLoadBalancer.DeepCopy()
	}
```

This was previously handled this way

vincepri: Which to be honest, maybe better to restructure, given that it was a bit confusing

api/v1beta2/awscluster_webhook.go
```diff
@@ -243,26 +251,44 @@ func (r *AWSCluster) validateNetwork() field.ErrorList {
...
 
-func (r *AWSCluster) validateControlPlaneLB() field.ErrorList {
+func (r *AWSCluster) validateControlPlaneLBs() field.ErrorList {
 	var allErrs field.ErrorList
 
-	if r.Spec.ControlPlaneLoadBalancer == nil {
-		return allErrs
+	// If the secondary is defined, check that the name is not empty and different from the primary.
+	// Also, ensure that the secondary load balancer is an NLB
+	if r.Spec.SecondaryControlPlaneLoadBalancer != nil {
```
vincepri, line 263: We might also want to check that the `Scheme` of the load balancer is different than the primary?

cc @JoelSpeed

controllers/awsmachine_controller.go
```diff
@@ -868,10 +869,28 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
...
+		return kerrors.NewAggregate(errs)
 	}
 
-	switch elbScope.ControlPlaneLoadBalancer().LoadBalancerType {
+	if err := r.registerInstanceToLBs(machineScope, elbsvc, i, elbScope.ControlPlaneLoadBalancer()); err != nil {
+		return errors.Wrapf(err, "could not register machine to primary load balancer")
+	}
+	if err := r.registerInstanceToLBs(machineScope, elbsvc, i, elbScope.SecondaryControlPlaneLoadBalancer()); err != nil {
+		return errors.Wrapf(err, "could not register machine to secondary load balancer")
+	}
```
vincepri, line 883: Convert to a for loop?

controllers/awsmachine_controller.go
```diff
@@ -868,10 +869,28 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
 			return r.deregisterInstanceFromClassicLB(machineScope, elbsvc, i)
 		}
 		machineScope.Debug("deregistering from v2 load balancer")
-		return r.deregisterInstanceFromV2LB(machineScope, elbsvc, i)
+		errs := []error{}
+		errs = append(errs, r.deregisterInstanceFromV2LB(machineScope, elbsvc, i, elbScope.ControlPlaneLoadBalancer()))
+		errs = append(errs, r.deregisterInstanceFromV2LB(machineScope, elbsvc, i, elbScope.SecondaryControlPlaneLoadBalancer()))
```
vincepri, line 874: Covert to a for loop?

vincepri: It might be good to add a helper in scope that returns all control plane load balancers?

controllers/awsmachine_controller.go
```diff
@@ -868,10 +869,28 @@ func (r *AWSMachineReconciler) reconcileLBAttachment(machineScope *scope.Machine
...
+		return errors.Wrapf(err, "could not register machine to secondary load balancer")
+	}
+
+	return nil
+}
+
+func (r *AWSMachineReconciler) registerInstanceToLBs(machineScope *scope.MachineScope, elbsvc services.ELBInterface, i *infrav1.Instance, lb *infrav1.AWSLoadBalancerSpec) error {
+	if lb == nil {
+		return nil
+	}
```
vincepri, line 891: Remove this and let's not call this function if it's nil?

docs/book/src/topics/secondary-load-balancer.md
```diff
@@ -0,0 +1,31 @@
...
+```yaml
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
+kind: AWSCluster
+metadata:
+  name: test-aws-cluster
+spec:
+  region: us-east-2
+  sshKeyName: nrb-default
+  secondaryControlPlaneLoadBalancer:
```
vincepri, line 27: Very nice!

docs/book/src/topics/secondary-load-balancer.md
```diff
@@ -0,0 +1,31 @@
...
+    name: internal-apiserver
+    loadBalancerType: nlb
```
vincepri, line 29: Can we make this the default for the secondary?

api/v1beta2/awscluster_webhook.go
nrb: I was wondering if we should reject if the secondary LB wasn't using an `Internal` Scheme, but I think if we do put bounds on it that just making sure they're different is probably sufficient.

pkg/cloud/scope/elb.go
```diff
@@ -49,4 +49,6 @@ type ELBScope interface {
```
vincepri, line 49: Should these methods be deprecated?

pkg/cloud/scope/elb.go
```diff
@@ -49,4 +49,6 @@ type ELBScope interface {
 
 	// ControlPlaneEndpoint returns AWSCluster control plane endpoint
 	ControlPlaneEndpoint() clusterv1.APIEndpoint
+
+	SecondaryControlPlaneLoadBalancer() *infrav1.AWSLoadBalancerSpec
```
vincepri, line 53: If we add a function in this interface that returns all load balancers, could we do away with the specific functions?

docs/book/src/topics/secondary-load-balancer.md
nrb: Yeah, that makes sense. Would that go in the webhook? Because right now the defaults are set via kubebuilder annotations on the types. Not sure it's worth making a new type for the secondary LB's spec.

vincepri: Yes defaulting webhook

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -140,21 +149,32 @@ func (s *Service) reconcileV2LB() error {
 	} else {
 		s.scope.Trace("Unmanaged control plane load balancer, skipping load balancer configuration", "api-server-elb", lb)
 	}
-	lb.DeepCopyInto(&s.scope.Network().APIServerELB)
+
+	if s.scope.SecondaryControlPlaneLoadBalancer() != nil && lb.Name == *s.scope.SecondaryControlPlaneLoadBalancer().Name {
```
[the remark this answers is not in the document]
vincepri: I guess we can keep using these, ideally in the future the status is a keyed slice, instead of a single field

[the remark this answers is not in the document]
vincepri: Is the field below status, or did you mean spec?

In either case though, the caller should be able to determine where it came from?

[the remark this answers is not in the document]
nrb: I had meant status, but I suppose if we have a function returning a slice of LBs while retaining two stored fields, we can inspect the `Schemes` to know which one goes where in the `Status`.

api/v1beta2/awscluster_webhook.go
JoelSpeed: I think having them different is an acceptable provision. I wouldn't want to restrict the primary to either public or private, as there are use cases for the primary to be either. But we definitely have the option to limit this API to 1 of each scheme.

If users need additional public load balancers, those are likely not for the control plane APIs and therefore should be served by the service controller rather than CAPI building out the infra IMO

nrb:
Looks like some sort of account lookup issue with Boskos; it fails [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/hack/boskos.py#L42)

vincepri (approved):
/lgtm
/assign @Ankitasw @richardcase

vincepri:
cc @JoelSpeed

richardcase (commented):
Great work @nrb . We should try to get this covered via e2e in the future.

docs/book/src/topics/secondary-load-balancer.md
```diff
@@ -0,0 +1,36 @@
+# Enabling a Secondary Control Plane Load Balancer
+
+## Overview
+
+It is possible to use a second control plane load balancer within a CAPA cluster.
+This secondary control plane load balancer is primarily meant to be used for internal cluster traffic, for use cases where traffic between nodes and pods should be kept internal to the VPC network.
+This adds a layer of privacy to traffic, as well as potentially saving on egress costs for traffic to the Kubernetes API server.
+
+A dual load balancer topology is not used as a default in order to maintain backward compatibility with existing CAPA clusters.
```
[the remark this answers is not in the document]
nrb: Yeah, I'll make an issue for the template.

One thing I'm looking at upstream is to support a [slice of ControlPlaneEndpoints](https://github.com/kubernetes-sigs/cluster-api/issues/5295), which I think would tend toward a default set up of two LBs, but I'm not positive. This PR is sort of a trial run, though I think CAPZ already supports internal/external LBs.

Issue #4391, opened 2023-07-11
Publish clusteraswadm as clusterctl plugin


richardcase:
/kind feature
/help
/milestone v2.2.0

**Describe the solution you'd like**
With a recent change in CAPI (https://github.com/kubernetes-sigs/cluster-api/pull/8957) it's possible to create plugins for clusterctl (like you can do with kubectl).

We should publish `clusterawsadm` so it can be used as an plubin like this:

`clusterctl aws bootstrap iam create-cloudformation-stack`

**Anything else you would like to add:**

This means we need to publish a binary called `clusterctl-aws`

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

richardcase:
We can publish it on our GitHub release, and it can be downloaded and placed in a users' PATH.

Ankitasw:
@richardcase we are already publishing clusterawsadm in our release binary, how do we publish it so that it can be used as plugin?

Issue #4610, opened 2023-11-03
clusterawsadm binary not working on Ubuntu 20.04


cgiraldo:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

running `clusterawsadm bootstrap iam create-cloudformation-stack` exits with error.

**What did you expect to happen:**
cloudformation-stack created with success

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

I think the problem is that the released binary was built against glibc>=2.32 but ubuntu 20.04 is using glibc2.31.
If I build the binary in my ubuntu 20.04 host, It works correctly.
If I use a previous released version (e.g. 2.2.1) the binary also works correctly.

```
clusterawsadm bootstrap iam create-cloudformation-stack
Attempting to create AWS CloudFormation stack cluster-api-provider-aws-sigs-k8s-io
clusterawsadm: dl-call-libc-early-init.c:37: _dl_call_libc_early_init: Assertion `sym != NULL' failed.
SIGABRT: abort
PC=0x2782c7c m=0 sigcode=18446744073709551610
signal arrived during cgo execution

goroutine 73 [syscall]:
runtime.cgocall(0x276a0a0, 0xc000330d88)
	/usr/local/go/src/runtime/cgocall.go:157 +0x4b fp=0xc000330d60 sp=0xc000330d28 pc=0x40a0eb
net._C2func_getaddrinfo(0xc0002367b0, 0x0, 0xc0007b2ba0, 0xc00075f178)
	_cgo_gotypes.go:100 +0x55 fp=0xc000330d88 sp=0xc000330d60 pc=0x587055
net._C_getaddrinfo.func1(0x413425?, 0x8?, 0x2a37b20?, 0x1?)
	/usr/local/go/src/net/cgo_unix_cgo.go:78 +0x7a fp=0xc000330df0 sp=0xc000330d88 pc=0x5879da
net._C_getaddrinfo(0xc000236600?, 0x27?, 0x0?, 0x0?)
	/usr/local/go/src/net/cgo_unix_cgo.go:78 +0x13 fp=0xc000330e20 sp=0xc000330df0 pc=0x587913
net.cgoLookupHostIP({0x3579119, 0x3}, {0xc000236600, 0x27})
	/usr/local/go/src/net/cgo_unix.go:166 +0x24f fp=0xc000330f60 sp=0xc000330e20 pc=0x5538cf
[...]

Issue #4309, opened 2023-06-02
Release v2.1.3 of clusterawsadm-linux-amd64 broken


muellerst-hg:
/kind bug

**What steps did you take and what happened:**
[A clear and concise description of what the bug is.]

2.1.2 of clusterawsadm-linux-amd64 was running fine on alpine 3.18.0 with musl
2.1.3 of clusterawsadm-linux-amd64 misses libc6 dependencies and requires `gcompat`package on alpine 3.18.0
2.1.3 version is marked with GitTreeState:"dirty"

Stes to reproduce on Alpine 3.18.0:
```
curl --location --output /usr/local/bin/clusterawsadm "https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/download/v2.1.3/clusterawsadm-linux-amd64"
chmod ugo+x /usr/local/bin/clusterawsadm
/usr/local/bin/clusterawsadm version
sh: /usr/local/bin/clusterawsadm: not found
ldd /usr/local/bin/clusterawsadm
        /lib64/ld-linux-x86-64.so.2 (0x7f4f06656000)
        libc.so.6 => /lib64/ld-linux-x86-64.so.2 (0x7f4f06656000)
Error relocating /usr/local/bin/clusterawsadm: __vfprintf_chk: symbol not found
Error relocating /usr/local/bin/clusterawsadm: __fprintf_chk: symbol not found

# Adding glibc compat package:
apk add gcompat

/usr/local/bin/clusterawsadm version
clusterawsadm version: &version.Info{Major:"2", Minor:"1", GitVersion:"v2.1.3-dirty", GitCommit:"7f7af3e0a3fc2bac59ca3683f420dea54ffcae2d", GitTreeState:"dirty", BuildDate:"2023-06-01T12:33:04Z", GoVersion:"go1.20.2", AwsSdkVersion:"v1.44.213", Compiler:"gc", Platform:"linux/amd64"}
```

**What did you expect to happen:**

`clusterawsadm` to work without `gcompat`, build from clean state

```
curl --location --output /usr/local/bin/xclusterawsadm "https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/download/v2.1.3/clusterawsadm-linux-amd64"
chmod ugo+x /usr/local/bin/clusterawsadm
/usr/local/bin/clusterawsadm version
[...]

Issue #4091, opened 2023-02-21
Feature Request: Automated Release Process


faiq:
/kind feature

This clones: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/2899

Describe the solution you'd like
Automate release process with Github Action that generates and uploads all the artifacts and release notes when a tag is created.

Anything else you would like to add:
Example: https://github.com/kubernetes-sigs/cluster-api-provider-azure/issues/1569 and https://github.com/kubernetes-sigs/cluster-api/pull/5188

Environment:

    Cluster-api-provider-aws version:
    Kubernetes version: (use kubectl version):
    OS (e.g. from /etc/os-release):

richardcase:
Also related to this:

- #4610 
- #4309

Ankitasw:
@faiq any progress here, since we want to include this in v2.4.0?

faiq:
Unfortunately, I haven't had an opportunity to work on it, nor am I too familiar with how the current release process works, so I don't feel confident enough to port it over. 

Maybe we can add an agenda item to the next CAPA office hours to briefly go over what needs to be done?

PR #4748, opened 2024-01-19
✨ use goreleaser for releasing binary

kranurag7:
**What type of PR is this?**
/kind feature

For automating releases. 

Fixes: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4391 https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4091

Related: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4610 https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4309

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Use goreleaser for releasing binary
```


.github/workflows/release.yaml
```diff
@@ -0,0 +1,31 @@
+name: release
+
+on:
+  push:
+    tags:
+    - 'release-**'
```
richardcase, line 6: Currently we sometimes tag main and then create a release branch. We should think about whether we stop this as part of the release.

.goreleaser.yaml
```diff
@@ -0,0 +1,33 @@
+builds:
+- id: "clusterctl-aws"
+  main: ./cmd/clusterawsadm
+  binary: bin/clusterctl-aws
```
richardcase, line 4: We also need to build a `clusterawsadm` binary.

cmd/clusterawsadm/cmd/version/version.go
```diff
@@ -63,7 +63,7 @@ func RunVersion(out io.Writer, cmd *cobra.Command) error {
 
 	switch of {
 	case "":
-		fmt.Fprintf(out, "clusterawsadm version: %#v\n", v.ClientVersion)
+		fmt.Fprintf(out, "clusterctl-aws version: %#v\n", v.ClientVersion)
```
richardcase, line 66: We will need to get the binary name to cover both binaries we build.

richardcase (commented):
Thanks for this @kranurag7 . I like the use of goreleaser here and creating the draft release.

.github/workflows/release.yaml
kranurag7: I switched to using `v*` for this workflow to trigger.

nrb:
This lgtm, but I am not familiar with the release process and therefore don't feel comfortable putting an actual review on this.

dlipovetsky:
From today's officeh ours: We'd like to have more maintainers/reviewers familiar with the current release process before changing this.

/milestone v2.5.0

vincepri:
this seems fine to me although I never used goreleaser, for comparison this is what upstream does https://github.com/kubernetes-sigs/cluster-api/blob/main/.github/workflows/release.yaml

Let's wait to cut v2.4.0 then merge this for subsequent releases?

nrb:
After going through the v2.4.1 release process, we ran into #4881. Using goreleaser bypasses this.

I tested this with the following process:

```shell
gh pr checkout 4748
goreleaser release --snapshot --clean
```

It created all the binaries faster than the existing Docker method did, along with tarballs including `clusterctl-aws` and `clusterawsadm` binaries for every platform.

I'd like for Damiano to take a look at this too, but I think this is worth setting up for at least creating the release and uploading binaries.

We can look at how the release notes and images get built later. I'm guessing we could roll the release notes process into this, but maybe not the image building/promotion.

/lgtm

/assign @damdo

.github/workflows/release.yaml
nrb: I like this solution - we can still do the release branch and build image while the binaries are building after the push.

kranurag7:
Hey there, 
Happy to work on this till completion. Please ask if there's any question on the PR. I'll be responsive :) 

I rebased the PR and I created a demo release on my fork.
https://github.com/kranurag7/cluster-api-provider-aws/releases/tag/v2.4.2 

Other things like removing binaries existing binaries related make targets, updating release docs etc. can be done in a subsequent PR or even this PR as well.

nrb:
> Happy to work on this till completion.

Thank you!

> Other things like removing binaries existing binaries related make targets

Let's not do that quite yet, IMO. I think once we get a release or two out with goreleaser we can do it, though.

> updating release docs

That would be great! We've been using [this doc](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/book/src/development/releasing.md).

.github/workflows/release.yaml
```diff
@@ -0,0 +1,31 @@
...
+        with:
+          go-version: '1.21'
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v5
+        with:
+          distribution: goreleaser
+          version: latest
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
nrb, line 31: @richardcase I think a maintainer would need to create a GitHub token for the goreleaser job to use. I do not have access to the repo secrets.

nrb: Ah, good call. Thanks!

nrb:
/lgtm

I think we should merge this and iterate on the docs.

What do you think @vincepri and @richardcase? I know it's not the same as the upstream CAPI process, but I do feel like using goreleaser is less likely to run into filesystem mounting issues between mac/Linux when run locally.

.goreleaser.yaml
```diff
@@ -0,0 +1,67 @@
...
+  goos:
+  - linux
+  - darwin
+  - windows 
+  goarch:
+  - amd64
+  - arm64
+  ignore:
+  - goos: windows
+    goarch: arm64
```
damdo, line 51: What does ignore do here?
It looks like we are releasing a windows for arm64 at the moment.
[clusterawsadm-windows-arm64.exe](https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/download/v2.4.1/clusterawsadm-windows-arm64.exe)

damdo: cc. @kranurag7

kranurag7: I think I've never seen windows running on arm. If something like that exists 🙂, I can update it. 
Also, I'm unsure if we also need to support freebsd here.

nrb: > I think I've never seen windows running on arm.

It's rare, but Microsoft made a couple Surface laptops that support ARM, as well as a RasPi image. See https://learn.microsoft.com/en-us/windows/arm/overview. Since we're already shipping it, probably worth continuing.

I don't think supporting BSD is a priority. We do not currently build for it.

damdo: +1 to what @nrb said.
I'd personally re-add windows arm64, and leave freeBSD out.
Thanks!

nrb: Looks like it's still ignoring windows-arm

kranurag7: We will be releasing two binaries `clusterawsadm` and `clusterctl-aws`
I removed it only from the first binary and missed it in the second one so I updated the same.  

The other thing I updated is following given goreleaser new release includes this. 
```diff
-  - -X 'sigs.k8s.io/cluster-api-provider-aws/v2/version.gitTreeState={{ if .IsGitDirty }}dirty{{ else }}clean{{ end }}'
+  - -X 'sigs.k8s.io/cluster-api-operator/version.gitTreeState={{.GitTreeState}}'
```

richardcase (changes requested):
Thanks for this @kranurag7 . As part of this i think we also need to update the makefile as we currently create a release:
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/Makefile#L506

And also we need to update the release process docs:
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/book/src/development/releasing.md

richardcase:
It would also be good to initially disable the change log:

```
changelog:
  disable: true
```

nrb:
https://prow.k8s.io/view/gs/kubernetes-jenkins/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/4748/pull-cluster-api-provider-aws-build-docker/1773297258993815552#1:build-log.txt%3A530

>  make: *** No rule to make target 'release-binaries'.  Stop. 

Could we rewrite this target to invoke goreleaser instead?

.goreleaser.yaml
```diff
@@ -0,0 +1,64 @@
...
+- id: clusterawsadm
+  builds:
+  - clusterawsadm
+  format_overrides:
+  - goos: windows
+    format: zip
+  name_template: "clusterawsadm_{{ .Tag }}_{{ .Os }}_{{ .Arch }}"
+
+changelog:
+  disable: true
```
nrb, line 64:
```suggestion
  disable: true

release:
  discussion_category_name: General

  extra_files:
    - glob: ./templates/*.yaml

  draft: true
```

This suggestion comes from examining the contents of https://github.com/kubernetes-sigs/cluster-api-provider-aws/releases/tag/v2.4.1. Mostly, it's including the template files which are fetched by clusterctl.

It also moves the choice to be a draft into the configuration file vs a command line argument.

Makefile
```diff
@@ -503,7 +503,7 @@ check-release-tag: ## Check if the release tag is set
 
 .PHONY: create-gh-release
 create-gh-release:$(GH) ## Create release on Github
-	$(GH) release create $(VERSION) -d -F $(RELEASE_DIR)/CHANGELOG.md -t $(VERSION) -R $(GH_REPO)
+	$(GH) release create $(VERSION) --draft --notes-file $(RELEASE_DIR)/CHANGELOG.md --title $(VERSION) --repo $(GH_REPO)
```
nrb, line 506:
```suggestion
	$(GH) release create $(VERSION) --notes-file $(RELEASE_DIR)/CHANGELOG.md --title $(VERSION) --repo $(GH_REPO)
```

Move the draft option into `.goreleaser.yaml`; see comment on that file.

nrb:
@kranurag7 I made a [commit](https://github.com/nrb/cluster-api-provider-aws/commit/b9367fdfd1a21d84280bfe23b790305ad80e666a) which I believe will fix the CI job, as well as align what goreleaser produces more closely to the current contents of a release.

This includes the suggestions I've made so far. You're welcome to grab it and incorporate it.

kranurag7:
Thank you, I think we should also decide how templates gets uploaded. As of now, we do it via `make` target. 

If we intend to do it via goreleaser then the corresponding make targets needs to be adjusted accordingly.

```make
.PHONY: release-templates
release-templates: $(RELEASE_DIR) ## Generate release templates
	cp templates/cluster-template*.yaml $(RELEASE_DIR)/

.PHONY: upload-gh-artifacts
upload-gh-artifacts: $(GH) ## Upload artifacts to Github release
	$(GH) release upload $(VERSION) -R $(GH_REPO) --clobber  $(RELEASE_DIR)/*
```

NOTE: This is not a blocker because we are using `--clobber` meaning that we are overwriting existing assets with the same name. But it would be good to have it only once.

nrb:
> If we intend to do it via goreleaser then the corresponding make targets needs to be adjusted accordingly.

Pushed a commit that removes these targets, as well as removing them from the documentation that we've used in the past.

@faiq @richardcase @damdo would be helpful if you double checked the docs changes I've made.

docs/book/src/development/releasing.md
```diff
@@ -53,30 +53,34 @@ Promote the container images from the staging registry to the production registr
     docker pull registry.k8s.io/cluster-api-aws/cluster-api-aws-controller:${VERSION}
     ```
 
-## Create release artifacts, and a GitHub draft release
 
-1. Again, make sure your repo is clean by git standards.
-1. Export the current branch `export BRANCH=release-1.5` (`export BRANCH=main`)and run `make release`.
```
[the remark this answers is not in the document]
nrb: Also need to retain the `make release-manifests` output to get our `metadata.yaml` file updated and uploaded

.goreleaser.yaml
```diff
@@ -0,0 +1,74 @@
...
+  - goos: windows
+    format: zip
+  name_template: "clusterctl-aws_{{ .Tag }}_{{ .Os }}_{{ .Arch }}"
+  format: binary
+- id: clusterawsadm
+  builds:
+  - clusterawsadm
+  format_overrides:
+  - goos: windows
+    format: zip
```
nrb, line 61: We can drop this format override since we're uploading binaries now, not archives.

.goreleaser.yaml
```diff
@@ -0,0 +1,74 @@
...
+  name_template: "clusterawsadm_{{ .Tag }}_{{ .Os }}_{{ .Arch }}"
+  format: binary
+
+changelog:
+  disable: true
+
+release:
+  discussion_category_name: General
+
+  extra_files:
```
nrb, line 71: Add `metadata.yaml` as an uploaded artifact

nrb:
/hold

Holding because there are some missing artifacts when releasing with this configuration.

Makefile
```diff
@@ -605,28 +600,8 @@ promote-images: $(KPROMO) $(YQ)
...
-		-e GOOS=$(GOOS) \
-		-e GOARCH=$(GOARCH) \
-		-e GOCACHE=/tmp/ \
-		--user $$(id -u):$$(id -g) \
-		-v "$$(pwd):/workspace$(DOCKER_VOL_OPTS)" \
-		-w /workspace \
-		$(GO_CONTAINER_IMAGE) \
-		go build -ldflags '$(LDFLAGS) -extldflags "-static"' \
-		-o $(RELEASE_DIR)/$(notdir $(RELEASE_BINARY))-$(GOOS)-$(GOARCH)$(EXT) $(RELEASE_BINARY)
+release-binaries: $(GORELEASER) ## Builds only the binaries, not a release. Deprecated, but kept until prow job definitions can be updated.
```
nrb, line 603: Un-deprecate and rename this so we know it's for CI/dry runs of the build.

.goreleaser.yaml
```diff
@@ -0,0 +1,64 @@
+builds:
```
nrb, line 1: Note for @richardcase since you asked about it before - the `changelog.disabled` flag actually disabled posting a changelog/release notes entirely; even customized ones passed via `--release-notes`. The `--release-notes` flag is what overrides the Goreleaser changelog generation.

See https://goreleaser.com/customization/changelog/

damdo (commented):
/lgtm

/assign @richardcase

.goreleaser.yaml
richardcase: Noted and understood, thanks @nrb . I mentioned it more in the context that currently the makefile would generate and attach the release notes. But if we are attaching the custom release notes directly with goreleaser then great :+1:

richardcase:
Great work @kranurag7 and @nrb , this is a real quality of life improvement :tada: 

/approve

nrb:
/cherry-pick release-2.4

The `.goreleaser.yaml` file needs to be present in the branch when its checked out during the build.

We worked around this for v2.4.2, but the cherry-pick will be useful for releases going forward.

PR #4764, opened 2024-01-31
✨ ROSA: Reconcile ROSAMachinePool version

muraee:
**What type of PR is this?**

/kind feature

Validates and reconcile `ROSAMachinePool.spec.version` to trigger an upgrade if version changes.

Fixes #

**Checklist**:

- [ ] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add & Reconcile ROSAMachinePools.spec.version
```


exp/controllers/rosamachinepool_controller.go
```diff
@@ -178,6 +180,15 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 	}
 	defer rosaClient.Close()
 
+	isValid, err := validateMachinePoolSpec(machinePoolScope)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to validate ROSAMachinePool.spec: %w", err)
+	}
+	if !isValid {
+		// dont' requeue because input is invalid and manual intervention is needed.
```
stevekuznetsov, line 188: We need to expose this in the status, or it's not actionable for users.

pkg/rosa/nodepools.go
```diff
@@ -70,3 +74,73 @@ func (c *RosaClient) DeleteNodePool(clusterID string, nodePoolID string) error {
...
+			return upgradePolicy, nil
+		}
+	}
+	return nil, nil
+}
+
+// ScheduleNodePoolUpgrade schedules a new nodePool upgrade to the specified version at the specified time.
+func (c *RosaClient) ScheduleNodePoolUpgrade(clusterID string, nodePool *cmv1.NodePool, version string, nextRun time.Time) (*cmv1.NodePoolUpgradePolicy, error) {
+	// earliestNextRun is set to at least 5 min from now by the OCM API.
+	// we set it to 6 min here to account for latencty.
```
stevekuznetsov, line 95: Can you speak to what this means, what latency we are accounting for and why it is necessary?

pkg/rosa/versions.go
```diff
@@ -2,13 +2,25 @@ package rosa
...
 import (
 	"fmt"
+	"strings"
 	"time"
 
+	"github.com/blang/semver"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
 )
 
+var MinSupportedVersion = semver.MustParse("4.14.0")
```
stevekuznetsov, line 12: This seems like something that OCM - not this repo - defines. Why do we have our own minimum?

pkg/rosa/versions.go
```diff
@@ -98,3 +110,47 @@ func (c *RosaClient) getControlPlaneUpgradePolicies(clusterID string) (controlPl
...
+		minVersion = MinSupportedVersion
+	}
+
+	return &minVersion, &maxVersion, nil
+}
+
+const versionPrefix = "openshift-v"
+
+// RawVersionID returns the rawID from the provided OCM version object.
+func RawVersionID(version *cmv1.Version) string {
```
stevekuznetsov, line 139: Please simply import [these](https://github.com/openshift/rosa/blob/8d35bae1dfe8a78c4f58c9d45f1467f1e853d449/pkg/ocm/versions.go#L238).

exp/api/v1beta2/rosamachinepool_types.go
```diff
@@ -33,6 +33,13 @@ type RosaMachinePoolSpec struct {
 	// +kubebuilder:validation:Pattern:=`^[a-z]([-a-z0-9]*[a-z0-9])?$`
 	NodePoolName string `json:"nodePoolName"`
 
+	// Version specifies the penshift version of the nodes associated with this machinepool.
+	// ROSAControlPlane version is used if not set.
+	//
+	// +optional
+	// +kubebuilder:validation:Pattern:=`^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`
```
stevekuznetsov, line 40: Do we have some provenance for this regex? Does it faithfully capture all possible semantic versions?

Also, please prefer to add code-gen comments above GoDoc so they do not get added to the user-facing documentation:

```go
	// +optional
	// +kubebuilder:validation:Pattern:=`^(0|[1-9]\d*)\.(0|[1-9]\d*)\.(0|[1-9]\d*)(?:-((?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*)(?:\.(?:0|[1-9]\d*|\d*[a-zA-Z-][0-9a-zA-Z-]*))*))?(?:\+([0-9a-zA-Z-]+(?:\.[0-9a-zA-Z-]+)*))?$`

	// Version specifies the penshift version of the nodes associated with this machinepool.
	// ROSAControlPlane version is used if not set.
	Version string `json:"version,omitempty"`
```

muraee: https://semver.org/#is-there-a-suggested-regular-expression-regex-to-check-a-semver-string

exp/controllers/rosamachinepool_controller.go
muraee: its exposed in the func
```go
		message := fmt.Sprintf("version %s is not supported, should be in the range: >= %s and <= %s", version, minSupportedVersion, maxSupportedVersion)
		machinePoolScope.RosaMachinePool.Status.FailureMessage = &message
		return false, nil
```
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4764/files#diff-ca8ae33b4375b0269fe24617620c4d304d1882cb8d562c7b544331d272c7aec1R352

exp/api/v1beta2/rosamachinepool_types.go
vincepri: See https://github.com/kubernetes-sigs/cluster-api/blob/main/util/version/version.go instead of inlined regex?

pkg/rosa/nodepools.go
muraee: from the time we send the request to the time ocm process it.
so if we now() = 12:00:00 and we set nextRun = 12:05:00, by the time the request is received by OCM, now() could be 12:00:30 and ocm will decide the earliest nextRun is actually at 12:05:30 not 12:05:00

pkg/rosa/versions.go
muraee: this is mainly for nodePools, OCM does define this but we can't retrieve this value.
`version.HostedControlPlaneEnabled()` will return false for `4.14.0` because only `4.14.5+` are supported for hosted control planes, but nodepools version can actually be `4.14.0`

exp/controllers/rosamachinepool_controller.go
stevekuznetsov: I see - it's surprising that a function named `validate...()` mutates its inputs. Can we make the control flow clearer?

exp/api/v1beta2/rosamachinepool_types.go
muraee: are you suggesting we add a validation webhook that uses this to validate the version? or is there a way to reference the regex variables in `+kubebuilder:validation:Pattern`?

pkg/rosa/versions.go
muraee: I thought we agreed to not import the rosa cli repo, but work on creating a separate shared library?

stevekuznetsov: No. We are importing this.

exp/api/v1beta2/rosamachinepool_types.go
vincepri: Yeah correct

muraee: added a webhook

vincepri (approved):
/approve
/assign @stevekuznetsov

PR #4765, opened 2024-01-31
:sparkles:  rosa: use the openshift/rosa libraries where we can

stevekuznetsov:
requires:

https://github.com/openshift/rosa/pull/1746
https://github.com/openshift/rosa/pull/1745
https://github.com/openshift/rosa/pull/1740

```release-note
NONE
```


stevekuznetsov:
Requires https://github.com/openshift/rosa/pull/1782

vincepri (approved):
/approve
/assign @muraee

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -200,13 +207,12 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		return ctrl.Result{}, nil
 	}
 
-	cluster, err := rosaClient.GetCluster()
```
muraee, line 203: why are we changing the flow here? I don't think we should rely on status fields to inform decisions, as those could be edited/removed manually.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -380,25 +349,32 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
+	creator, err := rosaaws.CreatorForCallerIdentity(rosaScope.Identity)
 	if err != nil {
-		return ctrl.Result{}, err
+		return ctrl.Result{}, fmt.Errorf("failed to transform caller identity to creator: %w", err)
 	}
 
+	var cluster *cmv1.Cluster
+	if clusterID := rosaScope.ControlPlane.Status.ID; clusterID != nil && *clusterID != "" {
```
muraee, line 364: same comment as above

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -560,18 +536,19 @@ func (r *ROSAControlPlaneReconciler) reconcileClusterAdminPassword(ctx context.C
...
 }
 
-func validateControlPlaneSpec(rosaClient *rosa.RosaClient, rosaScope *scope.ROSAControlPlaneScope) (bool, error) {
+func validateControlPlaneSpec(ocmClient *ocm.Client, rosaScope *scope.ROSAControlPlaneScope) (bool, error) {
 	// reset previous message.
 	rosaScope.ControlPlane.Status.FailureMessage = nil
 
-	version := rosaScope.ControlPlane.Spec.Version
-	isSupported, err := rosaClient.IsVersionSupported(version)
+	versionList, err := ocm.GetVersionMinorList(ocmClient)
```
muraee, line 543: `GetVersionMinorList` doesn't sound like the correct method to use here.
Also can we keep the current `IsVersionSupported()` function or at least port it to rosa repo? `ValidateVersion()` is doing a lot of redundant checks and requires an extra API call to list versions.

muraee: well, that was my original concern with importing and using rosa. Rosa handles both classic and HCP and therefore does a lot of extra steps that we don't need. + some of the their logic is related to the cli interactive workflow. I don't think we can just use anything they have as is.
the current validation is 100% simpler and more efficient.

stevekuznetsov: It's not our place to tell ROSA that their logic is wrong. If you can provide them with better logic, merge it into their repo, we can use it.

muraee: I never said its wrong, I am saying we have a different use-case here and this solution is not optimal for us.

stevekuznetsov: This is the logic that `openshift/rosa` uses to validate versions when users ask for HCP clusters on ROSA. We have the same requirements that they do. Their `ValidateVersion()` function explicitly handles `isSTS, isHostedCP bool`. Our use-case is a strict subset of their use-case. If we believe that the way they handle this subset is inefficient, any improvement to their logic would always be consumable by us.

stevekuznetsov: Copying their logic and changing it downstream means that we cannot consume any correctness changes or improvements they make upstream.

muraee: I am not arguing introducing this to openshift/rosa and then consuming here is. this what I suggested at the start.
I am saying we shouldn't consume inefficient code before the improvement lands upstream.

muraee: I will open a PR for this.

stevekuznetsov:
Now depends on https://github.com/openshift/rosa/pull/1792

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -563,20 +534,14 @@ func (r *ROSAControlPlaneReconciler) reconcileClusterAdminPassword(ctx context.C
...
 	version := rosaScope.ControlPlane.Spec.Version
-	isSupported, err := rosaClient.IsVersionSupported(version)
-	if err != nil {
-		return nil, fmt.Errorf("failed to verify if version is supported: %w", err)
-	}
-
-	if !isSupported {
-		message := fmt.Sprintf("version %s is not supported", version)
-		return &message, nil
+	if _, err := ocmClient.ValidateHypershiftVersion(version, "stable"); err != nil {
```
muraee, line 539: why are we ignoring the return value of `ValidateHypershiftVersion()`?
This function return a bool if the version is valid, an error is only returned if there was an error during the API call.

Also why are we changing the `validateControlPlaneSpec()` func return signature? we need to distinguish between user invalid input and real errors to inform whether to requeue or not, API errors for example could be intermittent and be resolved when we requeue.

exp/controllers/rosamachinepool_controller.go
```diff
@@ -307,13 +308,22 @@ func (r *ROSAMachinePoolReconciler) reconcileMachinePoolVersion(machinePoolScope
...
 	if err != nil {
 		return fmt.Errorf("failed to get existing scheduled upgrades: %w", err)
 	}
 
 	if scheduledUpgrade == nil {
-		scheduledUpgrade, err = rosaClient.ScheduleNodePoolUpgrade(clusterID, nodePool, version, time.Now())
+		rawVersion := ocm.GetRawVersionId(version)
+		policy, err := ocmClient.BuildNodeUpgradePolicy(rawVersion, nodePool.ID(), ocm.UpgradeScheduling{
+			AutomaticUpgrades: false,
+			NextRun:           time.Now(),
```
muraee, line 320: `time.now()` doesn't work as explained in the comment of the original function.

stevekuznetsov: Please explain what you mean.

exp/controllers/rosamachinepool_controller.go
```diff
@@ -307,13 +308,22 @@ func (r *ROSAMachinePoolReconciler) reconcileMachinePoolVersion(machinePoolScope
...
 	clusterID := *machinePoolScope.ControlPlane.Status.ID
-	scheduledUpgrade, err := rosaClient.CheckNodePoolExistingScheduledUpgrade(clusterID, nodePool)
+	_, scheduledUpgrade, err := ocmClient.GetHypershiftNodePoolUpgrade(clusterID, machinePoolScope.ControlPlane.Spec.RosaClusterName, nodePool.ID())
 	if err != nil {
 		return fmt.Errorf("failed to get existing scheduled upgrades: %w", err)
 	}
 
 	if scheduledUpgrade == nil {
-		scheduledUpgrade, err = rosaClient.ScheduleNodePoolUpgrade(clusterID, nodePool, version, time.Now())
+		rawVersion := ocm.GetRawVersionId(version)
```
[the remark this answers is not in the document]
stevekuznetsov: How do we know? Are we expecting a different format in the `spec` from what `rosa` CLI expects on the command-line?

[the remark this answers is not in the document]
muraee: because we are validating the version field is a sematic version, which is what rosa/ocm calls rawID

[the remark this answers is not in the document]
muraee: rosa/ocm version.ID() has a `openshift-v` prefix

PR #4767, opened 2024-02-01
:sparkles:  rosa: load identity dynamically

stevekuznetsov:
Instead of requiring users to provide their account IDs and ARNs in the spec of their hosted control planes, we can refer to AWS credentials like the other managed control plane controllers do in this provider.

```release-note
NONE
```


Makefile
```diff
@@ -76,7 +76,7 @@ DOCKER_BUILDKIT=1
 export ACK_GINKGO_DEPRECATIONS := 1.16.4
 
 # Set --output-base for conversion-gen if we are not within GOPATH
-ifneq ($(abspath $(REPO_ROOT)),$(shell go env GOPATH)/src/sigs.k8s.io/cluster-api-provider-aws)
+ifneq ($(abspath $(REPO_ROOT)),$(abspath $(shell go env GOPATH)/src/sigs.k8s.io/cluster-api-provider-aws))
```
vincepri, line 79: Why the change?

stevekuznetsov: When GOPATH has a trailing slash, the logic before my change would not detect correctly and all the generated bits would end up in a sub-directory of the repo:

```
	--output-file-base=zz_generated.conversion --output-base=/home/stevekuznetsov/code/kubernetes-sigs/cluster-api-provider-aws/src/sigs.k8s.io/cluster-api-provider-aws \
```

vincepri: Is your GOPATH manually set? If so, removing the trailing slash seems reasonable to me, we've never had anyone report any issues on this and it might break other uses, could we revert?

GOPATH from 1.8 is set by default to `$HOME/go`

stevekuznetsov: `abspath` is surjective. It does not seem possible for `$(abspath $(REPO_ROOT)` to equal some path `$foo` but NOT to equal `abspath $foo`.

pkg/cloud/scope/rosacontrolplane.go
```diff
@@ -53,11 +59,23 @@ func NewROSAControlPlaneScope(params ROSAControlPlaneScopeParams) (*ROSAControlP
...
+		patchHelper:    nil,
+		controllerName: params.ControllerName,
+	}
+
+	session, serviceLimiters, err := sessionForClusterWithRegion(params.Client, managedScope, *params.ControlPlane.Spec.Region, params.Endpoints, params.Logger)
+	if err != nil {
+		return nil, errors.Errorf("failed to create aws session: %v", err)
+	}
+
+	stsClient := NewSTSClient(managedScope, managedScope, managedScope, managedScope.ControlPlane)
```
muraee, line 75: we have to set
```go
managedScope.session = session
managedScope.serviceLimiters = serviceLimiters
```
before calling `NewSTSClient()`

stevekuznetsov: Good catch, thanks. We could also have every consumer make their own client and not in-line the identity but it seemed easier to do this for now

Makefile
vincepri, line 79: Revert this given we have a different PR open?

pkg/cloud/interfaces.go
```diff
@@ -84,3 +84,14 @@ type ClusterScoper interface {
 	// Close closes the current scope persisting the cluster configuration and status.
 	Close() error
 }
+
+type SessionMetadata interface {
```
vincepri, line 88: Add a comment?

Makefile
stevekuznetsov: The change was necessary to generate bits in this PR, keeping a diff locally is not ideal so hopefully we just merge the other one today?

vincepri: We'll have to wait for feedback from the other maintainers in the other PR, we can revert this bit for now if we want to merge this one in and iterate on the rest of the codebase

stevekuznetsov:
The `apidiff` failures are all expected - we added new fields to the structs here.

controlplane/rosa/api/v1beta2/defaults.go
```diff
@@ -0,0 +1,13 @@
+package v1beta2
+
+import "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+
+// SetDefaults_AWSClusterSpec is used by defaulter-gen.
+func SetDefaults_AWSClusterSpec(s *RosaControlPlaneSpec) { //nolint:golint,stylecheck
```
muraee, line 6:
```suggestion
func SetDefaults_ROSAControlPlaneSpec(s *RosaControlPlaneSpec) { //nolint:golint,stylecheck
```

stevekuznetsov: :facepalm:

PR #4780, opened 2024-02-06
🌱 Add missing Fields to RosaControlPlane

serngawy:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Adding the missing RosaControlPlan API fields;

 Autoscaling
 ComputeMachineType
 ServiceCIDR
 PodCIDR
 HostPrefix

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #
[issue](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4737)

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
None
```

Issue #4737, opened 2024-01-15
ROSA: Extend ROSAControlPlane API


muraee:
/kind feature

**Describe the solution you'd like**
`ROSAControlPlane` CRD should be extended to include the following fields:
- Autoscaling
- ComputeMachineType
- ServiceCIDR
- PodCIDR
- HostPrefix
- PrivateLink
- EtcdEncryption (bool) / EtcdEncryptionKMSKey
- Tags

ROSAControlePlaneController would need to be adjusted to reconcile these fields.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

stevekuznetsov:
- [ ] Autoscaling
- [ ] ComputeMachineType
- [ ] ServiceCIDR
- [ ] PodCIDR
- [ ] HostPrefix
- [x] PrivateLink https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4758
- [ ] EtcdEncryption (bool) / EtcdEncryptionKMSKey
- [ ] Tags

PR #4795, opened 2024-02-16
:sparkles:Add missing Fields to RosaControlPlane - tags, etcdEncryption

xiangjingli:
**What type of PR is this?**
/kind feature

Adding the missing RosaControlPlan API fields;
Tags,
EtcdEncryption,
EtcdEncryptionKMSArn

Also removing the accountID and creatorARN from ROSA cluster templates as it is not necessary to declare them explicitly in the rosaControlPlane spec along the recent change. 

Fixes #
https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4737

There is another PR for adding different fields to RosaControlplan delivered by @serngawy 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4780

I will rebase the PR after the 4780 PR is merged.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/controllers/awscluster_controller_test.go
+++ b/controllers/awscluster_controller_test.go
@@ -1437,24 +1437,6 @@ func mockedDeleteVPCCallsForNonExistentVPC(m *mocks.MockEC2APIMockRecorder) {
 }
 
 func mockedDeleteVPCCalls(m *mocks.MockEC2APIMockRecorder) {
-	m.DescribeVpcEndpointsPages(gomock.Eq(&ec2.DescribeVpcEndpointsInput{
-		Filters: []*ec2.Filter{
-			{
-				Name:   aws.String("vpc-id"),
-				Values: aws.StringSlice([]string{"vpc-exists"}),
-			},
-		}}),
-		gomock.Any()).Do(func(_, y interface{}) {
-		funct := y.(func(page *ec2.DescribeVpcEndpointsOutput, lastPage bool) bool)
-		funct(&ec2.DescribeVpcEndpointsOutput{VpcEndpoints: []*ec2.VpcEndpoint{{
-			VpcEndpointId: aws.String("vpce-12345"),
-		}}}, true)
-	}).Return(nil).AnyTimes()
-
-	m.DeleteVpcEndpoints(gomock.Eq(&ec2.DeleteVpcEndpointsInput{
-		VpcEndpointIds: aws.StringSlice([]string{"vpce-12345"}),
-	})).Return(&ec2.DeleteVpcEndpointsOutput{}, nil).AnyTimes()
-
 	m.DescribeSubnetsWithContext(context.TODO(), gomock.Eq(&ec2.DescribeSubnetsInput{
 		Filters: []*ec2.Filter{
 			{
--- a/pkg/cloud/services/network/vpc.go
+++ b/pkg/cloud/services/network/vpc.go
@@ -276,6 +276,15 @@ func (s *Service) deleteVPCEndpoints() error {
 		return nil
 	}
 
+	// Gather all services that might have been enabled.
+	services := sets.New[string]()
+	if s.scope.Bucket() != nil {
+		services.Insert(fmt.Sprintf("com.amazonaws.%s.s3", s.scope.Region()))
+	}
+	if services.Len() == 0 {
+		return nil
+	}
+
 	// Get all existing endpoints.
 	endpoints, err := s.describeVPCEndpoints()
 	if err != nil {


controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -276,6 +276,14 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		ExpirationTimestamp(time.Now().Add(1 * time.Hour)).
 		Hypershift(cmv1.NewHypershift().Enabled(true))
 
+	etcdEncryption := rosaScope.ControlPlane.Spec.EtcdEncryption
+	// If EtcdEncryptionKMSArn is set, it will override etcdEncryption flag to true.
```
serngawy, line 280: I believe better not to override the etcdEncryption flag, if the etcdEncryption set to false it is false. Otherwise the flag has no sense.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -350,6 +358,17 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
 		SubnetIDs(rosaScope.ControlPlane.Spec.Subnets...).
 		STS(stsBuilder)
+
+	// specify tags
+	if len(rosaScope.ControlPlane.Spec.Tags) > 0 {
+		awsBuilder = awsBuilder.Tags(rosaScope.ControlPlane.Spec.Tags)
+	}
+
+	// etcd encryption kms key arn
+	if rosaScope.ControlPlane.Spec.EtcdEncryptionKMSArn != "" {
```
serngawy, line 368: Add check if the etcdEncryption is true

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -576,7 +595,29 @@ func validateControlPlaneSpec(rosaClient *rosa.RosaClient, rosaScope *scope.ROSA
 		return false, nil
 	}
 
+	// validate Etcd Encryption KMS Arn
+	err = kmsArnRegexpValidator.ValidateKMSKeyARN(&rosaScope.ControlPlane.Spec.EtcdEncryptionKMSArn)
```
serngawy, line 599: move this to line 369 after the if check

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -81,6 +81,17 @@ type RosaControlPlaneSpec struct { //nolint: maligned
...
 	// If no identity is specified, the default identity for this controller will be used.
 	IdentityRef *infrav1.AWSIdentityReference `json:"identityRef,omitempty"`
+
+	// User-defined tags for AWS resources
+	// +optional
+	Tags map[string]string `json:"tags,omitempty"`
+
+	// Add etcd encryption. By default etcd data is encrypted at rest. This option configures etcd encryption on top of existing storage encryption.
+	// +optional
+	EtcdEncryption bool `json:"etcdEncryption,omitempty"`
```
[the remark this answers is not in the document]
xiangjingli: The EtcdEncryption is boolean field. In Go, if a boolean field is not explicitly initialized or assigned a value, its default value is false.

controlplane/rosa/controllers/rosacontrolplane_controller.go
xiangjingli: The EtcdEncryptionKMSArn  overriding the etcdEncryption flag is following ROSA CLI routine. Should we follow up it

https://github.com/openshift/rosa/blob/eaae3fad9fcf36b9954489de0160c652117f9742/cmd/create/cluster/cmd.go#L457-L462

controlplane/rosa/controllers/rosacontrolplane_controller.go
xiangjingli: Currently if EtcdEncryptionKMSArn is set, the etcdEncryption is overridded to be  true. If the overriding is removed, we will need to add the check

controlplane/rosa/controllers/rosacontrolplane_controller.go
xiangjingli: This follows the current implementation routine. There is separate function validateControlPlaneSpec for handling all rosa control plane spec validation. The validateControlPlaneSpec function is triggered before creating the cluster

controlplane/rosa/controllers/rosacontrolplane_controller.go
muraee: since EtcdEncryption can't be set true without providing a kms key ARN, I don't think we need this bool field at all.
Either the KMSKeyArn field is set -> we set encryption to true internally or its not set -> we set encryption to false

serngawy: The only case I can see is during testing you may don't want to delete the arn kms key and want to not apply kms encryption. Otherwise agree, we don't need the the flag at all.

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: No need to validate the arn kms, if it is not set.

xiangjingli: This is implemented inside of the function ValidateKMSKeyARN - if kmsKeyARN is not set, it return nil directly

https://github.com/openshift-online/ocm-common/blob/ff8c6c11d90947a20612d87b2b6435c72fe85db4/pkg/resource/validations/kms_arn_regex_validation.go#L13-L14

controlplane/rosa/controllers/rosacontrolplane_controller.go
xiangjingli: agree. I will remove the etcdEncryption flag from the spec

PR #4804, opened 2024-02-21
✨ ROSA: Reconcile ROSAMachinePool fields

muraee:
**What type of PR is this?**

/kind feature

reconcile changes to the ROSAMachinePool fields

Add 2 new fields
- `Taints` 
- `TuningConfigs`

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
reconcile ROSAMachinePool fields
```


exp/controllers/rosamachinepool_controller.go
```diff
@@ -250,22 +233,60 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
...
-		npBuilder.Version(cmv1.NewVersion().ID(ocm.CreateVersionID(rosaMachinePool.Spec.Version, ocm.DefaultChannelGroup)))
-	}
 
 	nodePoolSpec, err := npBuilder.Build()
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to build rosa nodepool: %w", err)
 	}
 
-	createdNodePool, err = ocmClient.CreateNodePool(*controlPlane.Status.ID, nodePoolSpec)
+	_, found, err := ocmClient.GetNodePool(*controlPlane.Status.ID, rosaMachinePool.Spec.NodePoolName)
```
stevekuznetsov, line 242: Why did this get logic move to the middle of the builder? Let's keep it at the top, it makes it more difficult to understand the building flow.

exp/controllers/rosamachinepool_controller.go
```diff
@@ -250,22 +233,60 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 	}
 
 	npBuilder.AWSNodePool(cmv1.NewAWSNodePool().InstanceType(rosaMachinePool.Spec.InstanceType))
-	if rosaMachinePool.Spec.Version != "" {
-		npBuilder.Version(cmv1.NewVersion().ID(ocm.CreateVersionID(rosaMachinePool.Spec.Version, ocm.DefaultChannelGroup)))
-	}
 
 	nodePoolSpec, err := npBuilder.Build()
```
stevekuznetsov, line 237: Let's do the optional version addition above this and only call `.Build()` once

muraee: the version field can't be part of the nodepool Update call. So its only set if we are creating not updating.

stevekuznetsov: I see, I missed the new `Update()` flow. As noted in Slack it seems like OCM would be expecting us to do a diff and send updates only as needed, let's make those changes.

If you can determine a more clear codepath here it would be good. Right now we have a builder, call a `.Build()` on it, then re-use the builder and re-build. The control flow of "create half a builder, get the current object, maybe update it, finish the builder, call create" is difficult to follow. It is not entirely clear that this is safe, expected, or normal. Other builder patterns will reclaim memory or otherwise change internally when `.Build()` is called so even aside from the control flow, it's surprising to see it.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -267,7 +268,7 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		billingAccount = rosaScope.ControlPlane.Spec.BillingAccount
 	}
 
-	spec := ocm.Spec{
+	ocmSpec := ocm.Spec{
```
serngawy, line 271: clusterSpec is the better name

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -296,44 +296,48 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
+			if err != nil {
+				// TODO: expose in status.
+				rosaScope.Error(err, "rosacontrolplane.spec.network.serviceCIDR invalid", networkSpec.ServiceCIDR)
+				return ctrl.Result{}, nil
+			}
+			ocmSpec.ServiceCIDR = *serviceCIDR
 		}
+
+		ocmSpec.HostPrefix = networkSpec.HostPrefix
+		ocmSpec.NetworkType = networkSpec.NetworkType
```
serngawy, line 331: outside networkSpec Do you need to set the type explicit to "OVNKubernete"  OR ocm set to "OVNKubernete" if its empty ?

controlplane/rosa/controllers/rosacontrolplane_controller.go
muraee: clusterSpec might be confused with CAPI cluster spec. the name should be specific to our implementation.

controlplane/rosa/controllers/rosacontrolplane_controller.go
muraee: ocm sets "OVNKubernete" by default if not set

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: better then ocmClusterSpec. ocmSpec is confusing as well ocm is not an object

stevekuznetsov (commented):
Looks great. Since you've already factored `nodePoolBuilder()` and `nodePoolToRosaMachinePoolSpec()` really well for unit testing, would you mind adding one for them?

muraee:
@stevekuznetsov added, we can add more test cases later.

stevekuznetsov:
Sounds good, the round-trip unit test is what I had in mind.

/lgtm
/approve

exp/api/v1beta2/rosamachinepool_types.go
```diff
@@ -45,37 +46,64 @@ type RosaMachinePoolSpec struct {
...
 	// ProviderIDList contain a ProviderID for each machine instance that's currently managed by this machine pool.
 	// +optional
 	ProviderIDList []string `json:"providerIDList,omitempty"`
 }
 
+type RosaTaint struct {
+	// +kubebuilder:validation:Required
+	//
+	// The taint key to be applied to a node.
+	Key string `json:"key"`
```
vincepri, line 93: We usually have the `+kubebuilder` tags after the comment, like:

```suggestion
	// The taint key to be applied to a node.
	//
	// +kubebuilder:validation:Required
	Key string `json:"key"`
```

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -296,44 +296,48 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
-		rosaScope.Error(err, "rosacontrolplane.spec.network.machineCIDR invalid", rosaScope.ControlPlane.Spec.Network.MachineCIDR)
-		return ctrl.Result{}, nil
-	}
+	if networkSpec := rosaScope.ControlPlane.Spec.Network; networkSpec != nil {
+		if networkSpec.MachineCIDR != "" {
+			_, machineCIDR, err := net.ParseCIDR(networkSpec.MachineCIDR)
+			if err != nil {
+				// TODO: expose in status
+				rosaScope.Error(err, "rosacontrolplane.spec.network.machineCIDR invalid", networkSpec.MachineCIDR)
+				return ctrl.Result{}, nil
```
vincepri, line 305: Could we move this to a validation webhook, and just return the error here (backoff loop) or set a condition?

stevekuznetsov: Returning an error when the data provided by the user is incorrect would only lead to infinite retry by the controller and not result in a good outcome. It seems universally correct to expose states like this in status so that the user knows that they need to update the object's `.spec` and so that the controller does not do extra reconciliation loops that cannot possibly result in a better outcome.

vincepri (approved):
/approve
/lgtm

PR #4808, opened 2024-02-22
🌱 Add basic unit test for EKS `AWSManagedControlPlaneReconciler`

AndiDog:
**What type of PR is this?**

/kind support

The EKS reconciliation was not covered in unit tests at all, making it hard to contribute changes with confidence. So like for `AWSCluster` reconciliation, I added a similar type of test that checks the basic creation of a cluster and the expected AWS API calls.

For now, the only test is for `AWSManagedControlPlane creation with managed VPC`, but more can be added more easily after this PR, since I'm making the controller testable.

This meant to fix a few small things along the way, such as version parsing which ignored errors or panicked until now.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```


Ankitasw:
This is really helpful @AndiDog, could you please add what all tests have you covered as part of description? It would be helpful in the review.

AndiDog:
@Ankitasw Added some more description, but really it's "just" about making the controller testable and adding a basic test `AWSManagedControlPlane creation with managed VPC`. I didn't want to blow up the PR size by adding much more. Adding more tests will be much easier after this.

richardcase:
Thanks @AndiDog. This is really helpful and as a base it will enable us to improve our test coverage on the EKS side. From my side:

/approve

Issue #4812, opened 2024-02-23
Generating dependency report fails on release


richardcase:
/kind bug

**What steps did you take and what happened:**

When trying to do recent releases of v2.3.x & v2.4.0-beta.0 we encountered the following problem when generating the dependency report for the release notes:

 ```
ERRO unable to run cmd: go list -mod=readonly -m all, workdir: /tmp/go-modiff3838785554, stdout: , stderr: go: updates to go.mod needed, disabled by -mod=readonly; to update it:
        go mod tidy
, error: exit status 1  file="modiff/modiff.go:276"
FATA generating dependency report: getting dependency changes: unable to run cmd: go list -mod=readonly -m all, workdir: /tmp/go-modiff3838785554, stdout: , stderr: go: updates to go.mod needed, disabled by -mod=readonly; to update it:
        go mod tidy
, error: exit status 1  file="release-notes/main.go:193
```

Digging into it (and looking at the temporary directory created) its requiring changes to go.mod including:

```
toolchain go1.21.5
```

However, adding this causes the staging image build to fail with:

```
/workspace/hack/tools/go.mod:5: unknown directive: toolchain
go: errors parsing go.mod:
``` 

**What did you expect to happen:**

The build of the staging image and release notes should succeed.

**Anything else you would like to add:**

Short term we can disable the dependency report generation. Then for the future we can look to do:

- add the `toolchain1 back into go mod
- Upgrade the gcb image 

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

richardcase:
There is an issue with **release-notes**. It turns out if the previous version (i.e specified in `--start-sha`) was for a version of go prior to 1.21 and the current version (i.e. specified in `--end-sha`) is now v1.21 then when it checks out the starting sha there is no `toolchain` directive and so when `go list -mod=readonly -m all` is run it tries to update go.mod to add `toolchain` and fails.

richardcase:
We will probably have to disable the dependency report generation short term using `--dependencies=false` and then re-enable it again in the next release.

I will also investigate if we can influence this behaviour via env vars.

PR #4813, opened 2024-02-23
🐛 fix: changes to release process

richardcase:
**What type of PR is this?**

/kind bug

Changes from the failed releases:

- Including the `toolchain` directive
- Bumping the GCB image so that it has a newer Go version (that understands the `toolchain` directive)
- Changing the clusterawsadm build to be inline with CAPI

Relates #4812 

This will be done for realease-2,3 manually instead of via automatic cherry pick

**Checklist**:

- x ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Various changes to the releases process.
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -21,7 +21,7 @@ include $(ROOT_DIR_RELATIVE)/common.mk
 
 # Go
 GO_VERSION ?=1.21.5
-GO_CONTAINER_IMAGE ?= public.ecr.aws/docker/library/golang:$(GO_VERSION)
+GO_CONTAINER_IMAGE ?= golang:$(GO_VERSION)
 
 # Directories.
 ARTIFACTS ?= $(REPO_ROOT)/_artifacts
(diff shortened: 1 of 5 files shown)


damdo (commented):
/lgtm
Thanks!

Makefile
```diff
@@ -607,18 +607,17 @@ release-binaries: ## Builds the binaries to publish with a release
...
+release-binary: $(RELEASE_DIR) versions.mk ## Release binary
 	docker run \
 		--rm \
 		-e CGO_ENABLED=0 \
 		-e GOOS=$(GOOS) \
 		-e GOARCH=$(GOARCH) \
-		--mount=source=gocache,target=/go/pkg/mod \
-		--mount=source=gocache,target=/root/.cache/go-build \
+		-e GOCACHE=/tmp/ \
+		--user $$(id -u):$$(id -g) \
```
faiq, line 617: Why do we need this?

richardcase: This is to be consistent with upstream CAPI (i.e. when it builds clusterctl) and it also us having to use the git config command.

PR #4831, opened 2024-02-28
✨ ROSA: Reconcile ROSAMachinePool.spec.ProviderIDList

muraee:
**What type of PR is this?**

/kind feature

`spec.ProviderIDList` is required to fulfill CAPI MachniePool contract so the CAPI controller can match then to Nodes and report readiness.

depends on https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4804

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Reconcile ROSAMachinePool.spec.ProviderIDList
```

### DIFF
--- a/exp/controllers/rosamachinepool_controller.go
+++ b/exp/controllers/rosamachinepool_controller.go
@@ -5,6 +5,8 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/aws/aws-sdk-go/aws"
+	"github.com/aws/aws-sdk-go/service/ec2"
 	"github.com/blang/semver"
 	"github.com/google/go-cmp/cmp"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
@@ -15,6 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/klog/v2"
+	"k8s.io/utils/ptr"
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
@@ -130,6 +133,7 @@ func (r *ROSAMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		MachinePool:     machinePool,
 		RosaMachinePool: rosaMachinePool,
 		Logger:          log,
+		Endpoints:       r.Endpoints,
 	})
 	if err != nil {
 		return ctrl.Result{}, errors.Wrap(err, "failed to create scope")
@@ -198,6 +202,17 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 	}
 
 	rosaMachinePool := machinePoolScope.RosaMachinePool
+	machinePool := machinePoolScope.MachinePool
+
+	if rosaMachinePool.Spec.Autoscaling != nil && !annotations.ReplicasManagedByExternalAutoscaler(machinePool) {
+		// make sure cluster.x-k8s.io/replicas-managed-by annotation is set on CAPI MachinePool when autoscaling is enabled.
+		annotations.AddAnnotations(machinePool, map[string]string{
+			clusterv1.ReplicasManagedByAnnotation: "rosa",
+		})
+		if err := machinePoolScope.PatchCAPIMachinePoolObject(ctx); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
 
 	nodePool, found, err := ocmClient.GetNodePool(machinePoolScope.ControlPlane.Status.ID, rosaMachinePool.Spec.NodePoolName)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


exp/controllers/rosamachinepool_controller.go
```diff
@@ -372,6 +379,47 @@ func validateMachinePoolSpec(machinePoolScope *scope.RosaMachinePoolScope) (*str
...
+		Filters: buildEC2FiltersFromTags(tags),
+	})
+	if err != nil {
+		return err
+	}
+
+	var providerIDList []string
+	for _, reservation := range response.Reservations {
+		for _, instance := range reservation.Instances {
+			providerID := fmt.Sprintf("aws:///%s/%s", *instance.Placement.AvailabilityZone, *instance.InstanceId)
```
stevekuznetsov, line 400: Is this importable from somewhere? How do we know that this will be the value that ends up in `node.spec.providerID`?

pkg/cloud/scope/rosamachinepool.go
```diff
@@ -70,6 +74,18 @@ func NewRosaMachinePoolScope(params RosaMachinePoolScopeParams) (*RosaMachinePoo
 		return nil, errors.Wrap(err, "failed to init MachinePool patch helper")
 	}
 
+	controlPlaneScope := &ROSAControlPlaneScope{
```
stevekuznetsov, line 77: Let's implement the required interface instead on the machine pool scope?

```
var _ cloud.SessionMetadata = &RosaMachinePoolScope{}
```

exp/controllers/rosamachinepool_controller.go
muraee: this is not importable, but ihttps://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/91941d950a27f94c078e4797c12a856f416b39a7/pkg/cloud/scope/machine.go#L152

stevekuznetsov: Could we refactor to make sure our implementation always remains in sync?

pkg/cloud/scope/rosamachinepool.go
muraee: the MachinePool doesn't doesn't have its own identify reference, so implementing the `IdentityRef()` interface func didn't seem right to me.
Happy to do it if you disagree.

exp/controllers/rosamachinepool_controller.go
muraee: @vincepri any thoughts/objections on refactoring?

pkg/cloud/scope/rosamachinepool.go
stevekuznetsov: It's probably a matter of semantics, but in a sense the `ROSAMachinePool` does have an identity reference - it must always use the reference on the `ROSAControlPlane` that it's associated with. Let's implement the interface here so we don't create and throw away a `ROSAControlPlaneScope{}` - we have all the information necessary to do the lookup in the machine pool scope params.

exp/controllers/rosamachinepool_controller.go
stevekuznetsov: To be clear, the "refactor" ask is to export a function that does this formatting ...

exp/controllers/rosamachinepool_controller.go
```diff
@@ -198,6 +202,17 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 	}
 
 	rosaMachinePool := machinePoolScope.RosaMachinePool
+	machinePool := machinePoolScope.MachinePool
+
+	if rosaMachinePool.Spec.Autoscaling != nil && !annotations.ReplicasManagedByExternalAutoscaler(machinePool) {
+		// make sure cluster.x-k8s.io/replicas-managed-by annotation is set on CAPI MachinePool when autoscaling is enabled.
+		annotations.AddAnnotations(machinePool, map[string]string{
+			clusterv1.ReplicasManagedByAnnotation: "true",
```
vincepri, line 210:
```suggestion
			clusterv1.ReplicasManagedByAnnotation: "rosa",
```

Or anything similar, we could also add a domain name in there, it might just be useful

pkg/cloud/scope/providerid.go
```diff
@@ -124,3 +125,7 @@ func (p *ProviderID) Validate() bool {
 func (p *ProviderID) IndexKey() string {
 	return p.String()
 }
+
+func ConstructAWSProviderID(instanceID, availabilityZone string) string {
+	return fmt.Sprintf("aws:///%s/%s", availabilityZone, instanceID)
+}
```
vincepri, line 131:
```suggestion
// ProviderIDPrefix is the prefix of AWS resource IDs to form the Kubernetes Provider ID.
// NOTE: this format matches the 2 slashes format used in cloud-provider and cluster-autoscaler.
const ProviderIDPrefix = "aws://"

// GenerateProviderID generates a valid AWS Node/Machine ProviderID field.
//
// By default, the last id provider is used as identifier (last part)
func GenerateProviderID(ids...string) string {
	return fmt.Sprintf("%s/%s", ProviderIDPrefix, strings.Join(ids, "/")))
}
```

Something like that? Might be useful to add some test cases to match

pkg/rosa/helpers.go
```diff
@@ -0,0 +1,21 @@
+package rosa
+
+import (
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+)
+
+func IsNodePoolReady(nodePool *cmv1.NodePool) bool {
```
vincepri, line 7: Add comment

vincepri (approved):
/approve
/lgtm

PR #4802, opened 2024-02-20
:sparkles:Add missing Fields to RosaControlPlane - tags, etcdEncryption

xiangjingli:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Adding the missing RosaControlPlan API fields;
Tags,
EtcdEncryption,
EtcdEncryptionKMSArn

Also removing the accountID and creatorARN from ROSA cluster templates as it is not necessary to declare them explicitly in the rosaControlPlane spec along the recent change. 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes # https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4737

**Special notes for your reviewer**:
There is another PR for adding different fields to RosaControlplan delivered by @serngawy 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4780

I will rebase the PR after the 4780 PR is merged.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
NONE
```

PR #4758, opened 2024-01-24
:sparkles: controlplane/rosa: allow configuring private link

stevekuznetsov:
controlplane/rosa: add a test for the cluster transform

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

controlpane/rosa: make cluster declaration idomatic

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

controlplane/rosa: allow configuring private link

Signed-off-by: Steve Kuznetsov <skuznets@redhat.com>

---

```release-note
NONE
```

PR #4832, opened 2024-02-29
:sparkles: rosa: add missing fields to spec

stevekuznetsov:
This supersedes #4802 and #4758


controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -99,6 +99,26 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// Autoscaling specifies auto scaling behaviour for the MachinePools.
 	// +optional
 	Autoscaling *expinfrav1.RosaMachinePoolAutoScaling `json:"autoscaling,omitempty"`
+
+	// +kubebuilder:validation:Optional
+
+	// AdditionalTags are user-defined tags to be added on the AWS resources associated with the control plane.
+	AdditionalTags infrav1.Tags `json:"additionalTags,omitempty"`
```
muraee, line 106: since you removed the webhook, we need to add the validation for this in runtime inside `validateControlPlaneSpec()` func.
 https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4e748e20799c96f20300aeedef461a1608220b2e/api/v1beta2/tags.go#L77

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -99,6 +99,26 @@ type RosaControlPlaneSpec struct { //nolint: maligned
...
+	// EtcdEncryptionKMSArn is the ARN of the KMS key used to encrypt etcd. The key itself needs to be
+	// created out-of-band by the user and tagged with `red-hat:true`.
+	EtcdEncryptionKMSArn string `json:"etcdEncryptionKMSArn,omitempty"`
+
+	// +kubebuilder:validation:Optional
+
+	// Private restricts master API endpoint and application routes to direct, private connectivity.
+	// Traffic to these endpoints will use AWS PrivateLink to have connectivity between VPCs, AWS services,
+	// and your on-premises networks without exposing your traffic to the public internet.
+	Private *bool `json:"private,omitempty"`
```
muraee, line 121: why is this a pointer? pointers imply a third option if not set, but we only have 2 options.
This should just be a `bool` with a default to false.

stevekuznetsov: That's what the OCM API / ROSA sdk wants, I don't see a compelling reason to change that. A `*bool` allows distinguishing "unset" from "not private."

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
stevekuznetsov: Good catch - not in the previous PR even with the webhook, but let me add.

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
vincepri: What does unset mean in this context? If it's the same as `false`, we usually don't use pointers

muraee: that is my question, what is unset? its either private=true or not=false, what happens when unset?

stevekuznetsov: We are a trivial passthrough to the OCM API. I don't see any reason why we should choose a different semantic than they do.

stevekuznetsov: By "trivial passthrough" I mean - we don't do any processing on this, we don't read the value, we don't do anything other than slap it into the OCM API call and call it a day.

vincepri: Can do the same with `&` before the field to OCM API

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -209,6 +205,44 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		rosaScope.ControlPlane.Status.FailureMessage = nil
 	}
 
+	if errs := rosaScope.ControlPlane.Spec.AdditionalTags.Validate(); errs != nil {
+		var msg []string
+		for _, err := range errs {
+			msg = append(msg, err.Error())
+		}
+		rosaScope.ControlPlane.Status.FailureMessage = ptr.To(strings.Join(msg, ", "))
```
vincepri, line 213: Hm, I see this is using `FailureMessage`, although Cluster API usually uses this as a terminal state; this is covered in https://cluster-api.sigs.k8s.io/developer/providers/cluster-infrastructure#infracluster-resources

These errors should be either logged, returned, or in a condition

/hold

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -209,6 +205,44 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
+		// dont' requeue because input is invalid and manual intervention is needed.
+		return ctrl.Result{}, nil
+	}
```
muraee, line 216: why outside of `validateControlPlaneSpec()` func? I think we should encapsulate all validations in one place.

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
stevekuznetsov: No, we lose the unset semantic.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -209,6 +205,44 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
+
+	if rosaScope.ControlPlane.Spec.Network.ServiceCIDR != "" {
+		_, cidr, err := net.ParseCIDR(rosaScope.ControlPlane.Spec.Network.ServiceCIDR)
+		if err == nil {
+			serviceCIDR = cidr
+		} else {
+			rosaScope.ControlPlane.Status.FailureMessage = ptr.To(fmt.Sprintf("rosacontrolplane.spec.network.serviceCIDR invalid: %v", err))
+			return ctrl.Result{}, nil
+		}
+	}
```
vincepri, line 244: Why are these validated in a controller, rather in a webhook?

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -99,6 +99,26 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// Autoscaling specifies auto scaling behaviour for the MachinePools.
 	// +optional
 	Autoscaling *expinfrav1.RosaMachinePoolAutoScaling `json:"autoscaling,omitempty"`
+
+	// +kubebuilder:validation:Optional
```
vincepri, line 103: Nit: These `// +kubebuilder:` comment tags should be underneath the field not above, we've used this pattern in most if not all of the APIs we build

controlplane/rosa/controllers/rosacontrolplane_controller.go
stevekuznetsov: It gets muddy with the CIDRs since having the validation for them inside of the func means we need to pass the parsed values back out. Are you OK with having the tag validation inside, but the CIDR outside?

stevekuznetsov:
/hold cancel

We are using a set of status conditions to expose errors now.

xiangjingli:
@stevekuznetsov  @muraee 

The PR has not implemented the `AdditionalTags`,`EtcdEncryptionKMSArn` and `Private` in the rosa controlplane controller

Are we going to to push more code to the PR or create a new PR? 

For the ref, here is how `AdditionalTags`,`EtcdEncryptionKMSArn` are implemented

image

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4802/files

muraee:
@xiangjingli https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4844 includes your changes.
/close

Issue #1691, opened 2020-04-15
Consider reflecting VPC DHCP dns options in machine status DNS addresses


enxebre:
/kind bug

**What steps did you take and what happened:**
Have a VPC with custom DNS DHCP options, e.g domainName `agl.com` https://docs.aws.amazon.com/vpc/latest/userguide/VPC_DHCP_Options.html

Machine addresses do not reflect this additional DNS addresses as the node would do via the controller manager cloud provider 
https://github.com/kubernetes/kubernetes/blob/6b13befdfb3e30862dc86fd1c7739f58901f0bae/staging/src/k8s.io/legacy-cloud-providers/aws/aws.go#L1400-L1479:
Node:
```
status:
  addresses:
  - address: 10.0.137.6
    type: InternalIP
  - address: ip-10-0-137-6.agl.com
    type: Hostname
  - address: ip-10-0-137-6.agl.com
    type: InternalDNS

```

**What did you expect to happen:**
We might want to consider honour this source, include the additional DNS addresses and be more consistent with cloud provider.

Accessing the metadata though would require to run the aws machine controller pod in the host network. Alternatively we could consider infer this information from the regular aws API and construct the ips known format, the likes of:

```
	vpc, err := r.awsClient.DescribeVpcs(&ec2.DescribeVpcsInput{
		VpcIds: []*string{newestInstance.VpcId},
	})
	if err != nil {
		return err
	}

	dhcp, err := r.awsClient.DescribeDHCPOptions(&ec2.DescribeDhcpOptionsInput{
		DhcpOptionsIds: []*string{vpc.Vpcs[0].DhcpOptionsId},
	})
	if err != nil {
		return err
	}

	klog.Infof("DHCP %+v", dhcp.DhcpOptions[0].DhcpConfigurations[0])
	for _, i := range dhcp.DhcpOptions[0].DhcpConfigurations {
		if *i.Key == "domain-name" {
			klog.Infof("Domain value %v", i.Values)
		}
	}
        // construct additional DNS addresses
```

**Anything else you would like to add:**
At some point having DHCP options awareness and reflecting it on the machine.status addresses might come in handy for scenarios where users wanting to adopt pre-existing vpcs.

Related to https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/70#issuecomment-614057170

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
[...]

PR #4841, opened 2024-03-01
:bug: fix: Fix instance PrivateDNSName when domain-name is set in dhcpOpts

jparrill:

When a VPC sets a DHCPOption's domain-name with a custom DNS name, it's being ignored. Now, with this PR, ensure that during instance creation, the controller checks the DHCPOptions and sets the proper PrivateDNSName.

Signed-off-by: Juan Manuel Parrilla Madrid <jparrill@redhat.com>

**What type of PR is this?**

/kind bug

- https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1691
Fixes #[OCPBUGS-29391](https://issues.redhat.com/browse/OCPBUGS-29391)

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
NONE
```

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -25,6 +25,7 @@ import (
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
+	"github.com/aws/aws-sdk-go/service/ec2/ec2iface"
 	"github.com/pkg/errors"
 	"k8s.io/utils/ptr"
 
@@ -886,6 +887,8 @@ func (s *Service) SDKToInstance(v *ec2.Instance) (*infrav1.Instance, error) {
 
 func (s *Service) getInstanceAddresses(instance *ec2.Instance) []clusterv1.MachineAddress {
 	addresses := []clusterv1.MachineAddress{}
+	// Check if the DHCP Option Set has domain name set
+	domainName := s.GetDHCPOptionSetDomainName(s.EC2Client, instance.VpcId)
 	for _, eni := range instance.NetworkInterfaces {
 		privateDNSAddress := clusterv1.MachineAddress{
 			Type:    clusterv1.MachineInternalDNS,
@@ -895,7 +898,17 @@ func (s *Service) getInstanceAddresses(instance *ec2.Instance) []clusterv1.Machi
 			Type:    clusterv1.MachineInternalIP,
 			Address: aws.StringValue(eni.PrivateIpAddress),
 		}
+
 		addresses = append(addresses, privateDNSAddress, privateIPAddress)
+
+		if domainName != nil {
+			// Add secondary private DNS Name with domain name set in DHCP Option Set
+			additionalPrivateDNSAddress := clusterv1.MachineAddress{
+				Type:    clusterv1.MachineInternalDNS,
+				Address: fmt.Sprintf("%s.%s", strings.Split(privateDNSAddress.Address, ".")[0], *domainName),
+			}
+			addresses = append(addresses, additionalPrivateDNSAddress)
+		}
 
 		// An elastic IP is attached if association is non nil pointer
 		if eni.Association != nil {
@@ -910,6 +923,7 @@ func (s *Service) getInstanceAddresses(instance *ec2.Instance) []clusterv1.Machi
 			addresses = append(addresses, publicDNSAddress, publicIPAddress)
 		}
 	}
+
 	return addresses
 }
 
@@ -1003,6 +1017,54 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
 	s.scope.Info("Updating instance metadata options", "instance id", instanceID, "options", input)
 	if _, err := s.EC2Client.ModifyInstanceMetadataOptionsWithContext(context.TODO(), input); err != nil {
 		return err
+	}
+
+	return nil
+}
+
+// GetDHCPOptionSetDomainName returns the domain DNS name for the VPC from the DHCP Options.
+func (s *Service) GetDHCPOptionSetDomainName(ec2client ec2iface.EC2API, vpcID *string) *string {
+	log := s.scope.GetLogger()
+
+	if vpcID == nil {
+		log.Info("vpcID is nil, skipping DHCP Option Set discovery")
+		return nil
+	}
+
+	vpcInput := &ec2.DescribeVpcsInput{
+		VpcIds: []*string{vpcID},
+	}
+
+	vpcResult, err := ec2client.DescribeVpcs(vpcInput)
+	if err != nil {
+		log.Info("failed to describe VPC, skipping DHCP Option Set discovery", "vpcID", *vpcID, "Error", err.Error())
+		return nil
+	}
+
+	dhcpInput := &ec2.DescribeDhcpOptionsInput{
+		DhcpOptionsIds: []*string{vpcResult.Vpcs[0].DhcpOptionsId},
+	}
+
+	dhcpResult, err := ec2client.DescribeDhcpOptions(dhcpInput)
+	if err != nil {
+		log.Error(err, "failed to describe DHCP Options Set", "DhcpOptionsSet", *dhcpResult)
+		return nil
+	}
+
+	for _, dhcpConfig := range dhcpResult.DhcpOptions[0].DhcpConfigurations {
+		if *dhcpConfig.Key == "domain-name" {
+			if len(dhcpConfig.Values) == 0 {
+				return nil
+			}
+			domainName := dhcpConfig.Values[0].Value
+			// default domainName is 'ec2.internal' in us-east-1 and 'region.compute.internal' in the other regions.
+			if (s.scope.Region() == "us-east-1" && *domainName == "ec2.internal") ||
+				(s.scope.Region() != "us-east-1" && *domainName == fmt.Sprintf("%s.compute.internal", s.scope.Region())) {
+				return nil
+			}
+
+			return domainName
+		}
 	}
 
 	return nil
--- a/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
+++ b/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
@@ -131,6 +131,7 @@ func (t Template) ControllersPolicy() *iamv1.PolicyDocument {
 				"ec2:DescribeSecurityGroups",
 				"ec2:DescribeSubnets",
 				"ec2:DescribeVpcs",
+				"ec2:DescribeDhcpOptions",
 				"ec2:DescribeVpcAttribute",
 				"ec2:DescribeVpcEndpoints",
 				"ec2:DescribeVolumes",
--- a/cmd/clusterawsadm/cloudformation/bootstrap/fixtures/customsuffix.yaml
+++ b/cmd/clusterawsadm/cloudformation/bootstrap/fixtures/customsuffix.yaml
@@ -190,6 +190,7 @@ Resources:
           - ec2:DescribeSecurityGroups
           - ec2:DescribeSubnets
           - ec2:DescribeVpcs
+          - ec2:DescribeDhcpOptions
           - ec2:DescribeVpcAttribute
           - ec2:DescribeVpcEndpoints
           - ec2:DescribeVolumes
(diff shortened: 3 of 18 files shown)


jparrill:
I need to change the approach, modifying the status.addresses to change the PrivateDnsName reported instead of change that value directly.

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1020,49 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+
+	for _, dhcpConfig := range dhcpResult.DhcpOptions[0].DhcpConfigurations {
+		if *dhcpConfig.Key == "domain-name" {
+			for _, value := range dhcpConfig.Values {
+				if value.Value != nil && *value.Value != fmt.Sprintf("%s.compute.internal", s.scope.Region()) {
+					return aws.String(*value.Value)
+				}
+			}
+		}
+	}
```
muraee, line 1061:
```suggestion
	for _, dhcpConfig := range dhcpResult.DhcpOptions[0].DhcpConfigurations {
		if *dhcpConfig.Key == "domain-name" {
			if len(dhcpConfig.Values) == 0 {
			   return nil
			}
			domainName := dhcpConfig.Values[0]
			// default domainName is 'ec2.internal' in us-east-1 and 'region.compute.internal' in the other regions.
			if strings.Contains(*domainName, "ec2.internal") || strings.Contains(*domainName, "compute.internal") {
			   return nil
			} 
			return domainName 
		}
	}
```
no need to loop over `dhcpConfig.Values`, it can only has 1 value.

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1020,49 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+		return nil
+	}
+
+	vpcInput := &ec2.DescribeVpcsInput{
+		VpcIds: []*string{vpcID},
+	}
+
+	vpcResult, err := ec2client.DescribeVpcs(vpcInput)
+	if err != nil {
+		return nil
```
muraee, line 1039: why not return the error, or log it?

pkg/cloud/services/ec2/instances.go
```diff
@@ -910,6 +912,16 @@ func (s *Service) getInstanceAddresses(instance *ec2.Instance) []clusterv1.Machi
...
 		}
 	}
+
+	// Add secondary private DNS Name only if DHCP Option Set has domain name set
+	if domainName := s.GetDHCPOptionSetDomainName(s.EC2Client, instance.VpcId); domainName != nil {
+		additionalPrivateDNSAddress := clusterv1.MachineAddress{
+			Type:    clusterv1.MachineInternalDNS,
+			Address: fmt.Sprintf("%s.%s", strings.Split(privateDNSAddress.Address, ".")[0], *domainName),
+		}
+		addresses = append(addresses, additionalPrivateDNSAddress)
```
muraee, line 922: I think this should be moved inside the first loop and use `eni.PrivateDnsName` instead of `privateDNSAddress.Address`

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1020,49 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+		return nil
+	}
+
+	dhcpInput := &ec2.DescribeDhcpOptionsInput{
+		DhcpOptionsIds: []*string{vpcResult.Vpcs[0].DhcpOptionsId},
+	}
+
+	log.Info("Fetching DHCP Options to get the domain-name", "dhcpOptionsID", *vpcResult.Vpcs[0].DhcpOptionsId)
+
+	dhcpResult, err := ec2client.DescribeDhcpOptions(dhcpInput)
```
muraee, line 1048: if this requires a new permission, you need to add it here 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/cbf531711d117b823c587622ba0e64b94d0fd314/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go#L78

pkg/cloud/services/ec2/instances.go
jparrill: Yeah, I think I can log it. Return the error could break the current workflow of provisioning (which does not returns any errors in `getInstaceAddresses`).

pkg/cloud/services/ec2/instances.go
muraee:
```
Key: "domain-name",
	          Values: [{
	              Value: "example.com"
	            }]
```
once you find the "domain-name key, the associated `Values` array field can only have 1 entry.

you still need to loop over `dhcpResult.DhcpOptions[0].DhcpConfigurations` thats right

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,55 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+	if err != nil {
+		return nil
```
muraee, line 1053: can we log the error here as well.

controllers/awsmachine_controller.go
```diff
@@ -600,7 +600,6 @@ func (r *AWSMachineReconciler) reconcileNormal(_ context.Context, machineScope *
 		}
 	}
 
-	// tasks that can only take place during operational instance states
```
muraee, line 603: why is this comment removed?

jparrill: my bad xD

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,56 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
 	return nil
 }
 
+// GetDHCPOptionSetDomainName returns the domain DNS name for the VPC from the DHCP Options.
+func (s *Service) GetDHCPOptionSetDomainName(ec2client ec2iface.EC2API, vpcID *string) *string {
+	fmt.Println("vpcID", vpcID)
+	log := s.scope.GetLogger()
+
+	if vpcID == nil {
+		log.Error(errors.New("vpcID is nil"), "vpcID is nil")
```
mjlshen, line 1031:
```suggestion
		log.Info("vpcID is nil, skipping DHCP Option Set discovery")
```
I think an info level log makes more sense as we're letting the reconciler proceed, what do you think?

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,56 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
 	return nil
 }
 
+// GetDHCPOptionSetDomainName returns the domain DNS name for the VPC from the DHCP Options.
+func (s *Service) GetDHCPOptionSetDomainName(ec2client ec2iface.EC2API, vpcID *string) *string {
+	fmt.Println("vpcID", vpcID)
```
mjlshen, line 1027:
```suggestion
```
Leftover print, if we want to keep it we could maybe log it at a debug level below the nil check

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,56 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+		return nil
+	}
+
+	vpcInput := &ec2.DescribeVpcsInput{
+		VpcIds: []*string{vpcID},
+	}
+
+	vpcResult, err := ec2client.DescribeVpcs(vpcInput)
+	if err != nil {
+		log.Error(err, "failed to describe VPC", "vpcID", *vpcID)
```
mjlshen, line 1041:
```suggestion
		log.Info(err, "failed to describe VPC, skipping DHCP Option Set discovery", "vpcID", *vpcID)
```

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,56 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+		return nil
+	}
+
+	dhcpInput := &ec2.DescribeDhcpOptionsInput{
+		DhcpOptionsIds: []*string{vpcResult.Vpcs[0].DhcpOptionsId},
+	}
+
+	log.Info("Fetching DHCP Options to get the domain-name", "dhcpOptionsID", *vpcResult.Vpcs[0].DhcpOptionsId)
```
mjlshen, line 1049:
```suggestion
```
I think this should reduce to a debug log level or perhaps not be logged at all

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,56 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+	for _, dhcpConfig := range dhcpResult.DhcpOptions[0].DhcpConfigurations {
+		if *dhcpConfig.Key == "domain-name" {
+			if len(dhcpConfig.Values) == 0 {
+				return nil
+			}
+			domainName := dhcpConfig.Values[0].Value
+			// default domainName is 'ec2.internal' in us-east-1 and 'region.compute.internal' in the other regions.
+			if strings.Contains(*domainName, "ec2.internal") || strings.Contains(*domainName, "compute.internal") {
+				return nil
+			}
```
mjlshen, line 1066: This check wouldn't be sufficient if someone supplied a `domain-name` of `us-east-1.compute.internal` right?

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,56 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+
+	dhcpInput := &ec2.DescribeDhcpOptionsInput{
+		DhcpOptionsIds: []*string{vpcResult.Vpcs[0].DhcpOptionsId},
+	}
+
+	log.Info("Fetching DHCP Options to get the domain-name", "dhcpOptionsID", *vpcResult.Vpcs[0].DhcpOptionsId)
+
+	dhcpResult, err := ec2client.DescribeDhcpOptions(dhcpInput)
+	if err != nil {
+		log.Error(err, "failed to describe DHCP Options Set", "DhcpOptionsSet", *dhcpResult)
```
mjlshen, line 1053:
```suggestion
		log.Info(err, "failed to describe DHCP Options Set, skipping DHCP Option Set discovery", "DhcpOptionsSet", vpcResult.Vpcs[0].DhcpOptionsId)
```
`dhcpResult` is effectively guaranteed to be nil if an error is returned here, also matching earlier log prints to add context around what is being skipped

mjlshen (changes requested):

pkg/cloud/services/ec2/instances.go
muraee: +1

pkg/cloud/services/ec2/instances.go
jparrill: Yep, you're right.

jparrill: Yep, you're right.

pkg/cloud/services/ec2/instances.go
```diff
@@ -1008,6 +1022,54 @@ func (s *Service) ModifyInstanceMetadataOptions(instanceID string, options *infr
...
+		if *dhcpConfig.Key == "domain-name" {
+			if len(dhcpConfig.Values) == 0 {
+				return nil
+			}
+			domainName := dhcpConfig.Values[0].Value
+			// default domainName is 'ec2.internal' in us-east-1 and 'region.compute.internal' in the other regions.
+			if (s.scope.Region() == "us-east-1" && strings.Contains(*domainName, "ec2.internal")) ||
+				(s.scope.Region() != "us-east-1" && strings.Contains(*domainName, "compute.internal")) {
+				return nil
+			}
```
mjlshen, line 1064: I think we should do exact matches here, like

```go
*domainName == "ec2.internal"
*domainName == fmt.Sprintf("%s.compute.internal", s.scope.Region())
```

To prevent other edge-cases like setting a DHCP option set domain-name to `example.compute.internal` and others in that vein.

mjlshen (approved):
/lgtm

jparrill:
@Ankitasw it was a good merge? I mean it was failing the cluster creation in E2E isn't it?

PR #4846, opened 2024-03-04
🐛 Allow overriding health check endpoint protocol

mtulio:
**What type of PR is this?**

/kind bug

This change intents to synchronize w/ CLB the health check attributes for
API LB, and use the correct health check path when using HTTP or HTTPS.

The following target group health check attributes for ELBv2
are setting based in the CLB configuration for all protocols:

- interval: current(30) want(10)
- timeout: current(10) want(5)
- health threshold count: current(5) want(5)
- unhealthy threshold count: current(2) want(3)

When using HTTP or HTTPS, the following attributes will be fixed:

- path: current("/") want("/readyz")

*'current' is the default values set by AWS, 'want' is the value set
in the CLB.

N/A

**Checklist**:

- [ ] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fixes in the health check probe configuration for API target group on ELBv2. The correct health check path now is set to `/readyz` when `AWSLoadBalancerSpec.HealthCheckProtocol` is `HTTP` or `HTTPS`.
```

### DIFF
--- a/api/v1beta2/network_types.go
+++ b/api/v1beta2/network_types.go
@@ -27,6 +27,16 @@ const (
 	DefaultAPIServerPort = 6443
 	// DefaultAPIServerPortString defines the API server port as a string for convenience.
 	DefaultAPIServerPortString = "6443"
+	// DefaultAPIServerHealthCheckPath the API server health check path.
+	DefaultAPIServerHealthCheckPath = "/readyz"
+	// DefaultAPIServerHealthCheckIntervalSec the API server health check interval in seconds.
+	DefaultAPIServerHealthCheckIntervalSec = 10
+	// DefaultAPIServerHealthCheckTimeoutSec the API server health check timeout in seconds.
+	DefaultAPIServerHealthCheckTimeoutSec = 5
+	// DefaultAPIServerHealthThresholdCount the API server health check threshold count.
+	DefaultAPIServerHealthThresholdCount = 5
+	// DefaultAPIServerUnhealthThresholdCount the API server unhealthy check threshold count.
+	DefaultAPIServerUnhealthThresholdCount = 3
 )
 
 // NetworkStatus encapsulates AWS networking resources.
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -172,6 +172,23 @@ func (s *Service) getAPIServerLBSpec(elbName string, lbSpec *infrav1.AWSLoadBala
 		scheme = *lbSpec.Scheme
 	}
 
+	// The default API health check is TCP, allowing customization to HTTP or HTTPS when HealthCheckProtocol is set.
+	apiHealthCheckProtocol := infrav1.ELBProtocolTCP
+	if lbSpec != nil && lbSpec.HealthCheckProtocol != nil {
+		s.scope.Trace("Found API health check protocol override in the Load Balancer spec, applying it to the API Target Group", "api-server-elb", lbSpec.HealthCheckProtocol)
+		apiHealthCheckProtocol = *lbSpec.HealthCheckProtocol
+	}
+	apiHealthCheck := &infrav1.TargetGroupHealthCheck{
+		Protocol:        aws.String(apiHealthCheckProtocol.String()),
+		Port:            aws.String(infrav1.DefaultAPIServerPortString),
+		Path:            nil,
+		IntervalSeconds: aws.Int64(infrav1.DefaultAPIServerHealthCheckIntervalSec),
+		TimeoutSeconds:  aws.Int64(infrav1.DefaultAPIServerHealthCheckTimeoutSec),
+		ThresholdCount:  aws.Int64(infrav1.DefaultAPIServerHealthThresholdCount),
+	}
+	if apiHealthCheckProtocol == infrav1.ELBProtocolHTTP || apiHealthCheckProtocol == infrav1.ELBProtocolHTTPS {
+		apiHealthCheck.Path = aws.String(infrav1.DefaultAPIServerHealthCheckPath)
+	}
 	res := &infrav1.LoadBalancer{
 		Name:          elbName,
 		Scheme:        scheme,
@@ -181,14 +198,11 @@ func (s *Service) getAPIServerLBSpec(elbName string, lbSpec *infrav1.AWSLoadBala
 				Protocol: infrav1.ELBProtocolTCP,
 				Port:     infrav1.DefaultAPIServerPort,
 				TargetGroup: infrav1.TargetGroupSpec{
-					Name:     fmt.Sprintf("apiserver-target-%d", time.Now().Unix()),
-					Port:     infrav1.DefaultAPIServerPort,
-					Protocol: infrav1.ELBProtocolTCP,
-					VpcID:    s.scope.VPC().ID,
-					HealthCheck: &infrav1.TargetGroupHealthCheck{
-						Protocol: aws.String(string(infrav1.ELBProtocolTCP)),
-						Port:     aws.String(infrav1.DefaultAPIServerPortString),
-					},
+					Name:        fmt.Sprintf("apiserver-target-%d", time.Now().Unix()),
+					Port:        infrav1.DefaultAPIServerPort,
+					Protocol:    infrav1.ELBProtocolTCP,
+					VpcID:       s.scope.VPC().ID,
+					HealthCheck: apiHealthCheck,
 				},
 			},
 		},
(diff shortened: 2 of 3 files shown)


api/v1beta2/network_types.go
```diff
@@ -94,12 +104,56 @@ var (
...
-	ThresholdCount  *int64  `json:"thresholdCount,omitempty"`
+
+	// The protocol to use to health check connect with the target.
+	// Defaults to TCP.
+	// +kubebuilder:default=TCP
+	// +optional
+	Protocol *string `json:"protocol,omitempty"`
+
+	// The destination for health checks on the targets when using the protocol HTTP or HTTPS.
+	// Defaults to /redyz.
```
faiq, line 115:
```suggestion
	// Defaults to /readyz.
```

faiq: can you use this default with TCP protocol.

mtulio:
Holding while more tests is done.

/hold

api/v1beta2/network_types.go
faiq: I think it would make sense to make the default HTTP wouldn't it?

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -1002,10 +1029,10 @@ func (s *Service) getAPIServerClassicELBSpec(elbName string) (*infrav1.LoadBalan
...
 		HealthCheck: &infrav1.ClassicELBHealthCheck{
 			Target:             s.getHealthCheckTarget(),
-			Interval:           10 * time.Second,
-			Timeout:            5 * time.Second,
-			HealthyThreshold:   5,
-			UnhealthyThreshold: 3,
+			Interval:           infrav1.DefaultAPIServerHealthCheckIntervalSec * time.Second,
+			Timeout:            infrav1.DefaultAPIServerHealthCheckTimeoutSec * time.Second,
+			HealthyThreshold:   infrav1.DefaultAPIServerHealthThresholdCount,
+			UnhealthyThreshold: infrav1.DefaultAPIServerUnhealthThresholdCount,
```
faiq, line 1035: nice change.

api/v1beta2/network_types.go
mtulio: @faiq yeah, I think it is the most correct path. But I am afraid of changing defaults in CAPA.
We are using HTTPS by default on OpenShift.

+ @vincepri thoughts?

mtulio:
Hey @faiq running successfully the `make test`. Is there any other suite I can run locally to validate this change before triggering the e2e? I've tried a few other `test-?` but failure to set the environment :)

mtulio:
This PR is ready to review.

api/v1beta2/network_types.go
```diff
@@ -94,12 +104,57 @@ var (
...
+	// Defaults to TCP.
+	// +kubebuilder:default=TCP
+	// +kubebuilder:validation:Enum=TCP;HTTP;HTTPS
+	// +optional
+	Protocol *string `json:"protocol,omitempty"`
+
+	// The destination for health checks on the targets when using the protocol HTTP or HTTPS,
+	// otherwise the path will be ignored.
+	// Defaults to "/readyz".
+	// +kubebuilder:default="/readyz"
```
vincepri, line 117: Why readyz vs healthz? This also poses the question if we should have the default to be kubernetes specific

api/v1beta2/network_types.go
vincepri: Having a default here makes it non-backward compatible. Why not leave these values empty for end users to setup instead?

api/v1beta2/network_types.go
faiq: healthz is deprecated https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

They should be kubernetes specific because this is a check against the kube-apiserver component

api/v1beta2/network_types.go
mtulio: @vincepri considering HTTP* is a best practice, specially for the graceful shutdown: should we need to plan in the future and emit some kind message to use this in new installs?

vincepri: we can try to default on create only, rather than default all the time?

mtulio: @vincepri Good question. I am new in CAPA project to provide a good answer. :)
I am suggesting that possibility for CAPA considering the AWS and KAS recommendation. But imo for "downstream" projects, all they need is to be allowed to customize the health check attributes (addressed in #4849 , for now blocked by this PR).

Anyway this PR is not changing the default, so we can move that thread to an issue if you think it is something to be considered.

api/v1beta2/network_types.go
```diff
@@ -27,6 +27,16 @@ const (
...
+	// DefaultAPIServerHealthCheckPath the API server health check path.
+	DefaultAPIServerHealthCheckPath = "/readyz"
+	// DefaultAPIServerHealthCheckIntervalSec the API server health check interval in seconds.
+	DefaultAPIServerHealthCheckIntervalSec = 10
+	// DefaultAPIServerHealthCheckTimeoutSec the API server health check timeout in seconds.
+	DefaultAPIServerHealthCheckTimeoutSec = 5
+	// DefaultAPIServerHealthThresholdCount the API server health check threshold count.
+	DefaultAPIServerHealthThresholdCount = 5
+	// DefaultAPIServerUnhealthThresholdCount the API server unhealthy check threshold count.
+	DefaultAPIServerUnhealthThresholdCount = 3
```
nrb, line 39: Thanks for moving these to consts!

pkg/cloud/services/elb/loadbalancer.go
```diff
@@ -172,6 +172,24 @@ func (s *Service) getAPIServerLBSpec(elbName string, lbSpec *infrav1.AWSLoadBala
 		scheme = *lbSpec.Scheme
 	}
 
+	// The default API health check is TCP, allowing customization to HTTP or HTTPS when HealthCheckProtocol is set.
+	apiHealthCheckProtocol := infrav1.ELBProtocolTCP.String()
```
nrb, line 176: You shouldn't need the `.String()` calls here and on L190, I don't think.

mtulio: Good point, thanks. Removed.

api/v1beta2/network_types.go
```diff
@@ -94,12 +104,57 @@ var (
...
+	// a failed health check.
+	// Defaults to 5.
+	// +kubebuilder:default=5
+	// +kubebuilder:validation:Minimum=5
+	// +optional
+	TimeoutSeconds *int64 `json:"timeoutSeconds,omitempty"`
+
+	// The number of consecutive health check successes required before considering
+	// a target healthy.
+	// Defaults to 5.
```
vincepri, line 145: Just to clarify, are these the _current_ defaults? What's the current behavior if left unset?

mtulio: Yes, the current default for CLB. Those values comes from the default static set to CLB[1], that I am proposing to moving it to constants[2] to be reused in ELBv2.
[1] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4846/files#diff-ee38bda5efe2606e7e7746a556c49a2d53d14eca60c029b8efebea657b8718f1L1005-L1008
[2] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4846/files#diff-1c519200705fb576b64e7fb271f7f683a2e6efadd54abd2a70b243e5069490e9R32-R39

mtulio: @vincepri for your later question, thanks for poping it up. I am revising this current fix. The goal here is to propose a fix in the API's TG health check setting it correctly, and not change or expose the API. So I removed the updates in `TargetGroupHealthCheck`, setting the defaults using the same method as CLB.

The PR #4849 is addressing the wider scope, proposing changes in the API allowing users to customize the health checks setting it accordingly when creating LB, addressing it as a feature instead of a bug fix.

[1] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4849/files#diff-ee38bda5efe2606e7e7746a556c49a2d53d14eca60c029b8efebea657b8718f1

Let me know what do you think, you all, @vincepri @nrb @faiq .

faiq: I left my comments on that PR.

For this one, I think it is fine. From my reading this PR is simply setting the same health check for ELBv2 as there currently is for ELBv1. Can you please confirm if I got that right?

mtulio: > this PR is simply setting the same health check for ELBv2 as there currently is for ELBv1
> Can you please confirm if I got that right?

@faiq yes.

mtulio:
> /test pull-cluster-api-provider-aws-e2e

@nrb @faiq  ✅

api/v1beta2/network_types.go
```diff
@@ -94,12 +104,13 @@ var (
...
-	IntervalSeconds *int64  `json:"intervalSeconds,omitempty"`
-	TimeoutSeconds  *int64  `json:"timeoutSeconds,omitempty"`
-	ThresholdCount  *int64  `json:"thresholdCount,omitempty"`
+	Protocol                *string `json:"protocol,omitempty"`
+	Path                    *string `json:"path,omitempty"`
+	Port                    *string `json:"port,omitempty"`
+	IntervalSeconds         *int64  `json:"intervalSeconds,omitempty"`
+	TimeoutSeconds          *int64  `json:"timeoutSeconds,omitempty"`
+	ThresholdCount          *int64  `json:"thresholdCount,omitempty"`
+	UnhealthyThresholdCount *int64  `json:"unhealthyThresholdCount,omitempty"`
```
faiq, line 113: wait is this a new API field?

faiq:
/hold


is there an API change?

api/v1beta2/network_types.go
mtulio: @faiq IIUC the missing field when setting the LB is part of the status[1], updated by controller. Do you think we can drop this status field here moving to the other PR #4849 which is exposing those attributes? Forgive me I am new in the project, trying to understand overall flow. :)

[1] 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4846/files#diff-aec4bb6b6279e42ec359b4c273734702a032460b7e5fc6ffae4766317a7247b4
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4846/files#diff-819e1fc8b69dba75e117c922715dbad8c4ed5bd8ef5608adbdc00f1eee18a546

faiq: Let’s keep the API objects as they are for right now and focus on parity between ELB v1 and 2

mtulio: @faiq thanks for the feedback. Commit removed. Using the unhealthy threshold to the constant when creating the TG.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/7a658ca873bcc1b5630c84f8b4d7de84947d6680/pkg/cloud/services/elb/loadbalancer.go#L338

mtulio:
> /retest

@faiq three unit were failing, just fixed it adding the Unhealthy threshold to the mock `&elbv2.CreateTargetGroupInput`

[1] 
"main create flow"
"created with ipv6 vpc"
"load balancer is not an NLB scope security groups will be added"

mtulio:
PTAL @nrb ? cc @vincepri

vincepri (approved):
/approv

nrb:
/retitle 🐛 Allow overriding health check endpoint protocol
/unhold

PR #4889, opened 2024-03-20
:bug: Add validation when PublicIP == true but user-supplied public subnet has MapPublicIPOnLauch == false

### LINKED ISSUE #4888: No validation done when only `PublicIP == true` but public subnet has `MapPublicIPOnLauch == false`
/kind bug

1. Create a shared VPC with private and public subnets, where all public subnets have `MapPublicIPOnLaunch == false`.
2. Create an AWSMachine with `PublicIP = true`  and `Subnet = nil`

CAPA throws an error letting the user know that no public IP will be assigned to the instance.

Instead what happens is the machine only has a private IP assigned and cannot talk to the internet because the internet gateway doesn't provide NAT.

https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure does not say that having `MapPublicIPOnLaunch == true` is a requirement for public subnets, even though it expects that. Note that the condition is checked if a subnet ID is set in the AWSMachine spec: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/ec2/instances.go#L337-L340

**Environment:**

- Cluster-api-provider-aws version: main @ d62768f507f85b25db1e5a74e0e820ed1f8e4414
- OS (e.g. from `/etc/os-release`): RHCOS

r4f4:
**What type of PR is this?**
/kind bug

In the scenario where the user brings their own VPC, if no subnet ID is set in the machine spec and `PublicIP` is `true`, CAPA will choose one from the available public subnets. However, if the subnet doesn't have `MapPublicIPOnLaunch == true`, the instance will not be assigned a public IP. As a result, the instance will have no internet access, contrary to the user's expectation.

Fixes #4888

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Added validation for MapPublicIPOnLaunch be set to true for user-supplied public subnets when the AWSMachine has PublicIP == true.
```

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -379,6 +379,27 @@ func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
 			record.Eventf(scope.AWSMachine, "FailedCreate", errMessage)
 			return "", awserrors.NewFailedDependency(errMessage)
 		}
+
+		criteria := &ec2.Filter{
+			Name:   aws.String("subnet-id"),
+			Values: aws.StringSlice([]string{subnets[0].GetResourceID()}),
+		}
+		ec2Subnets, err := s.getFilteredSubnets(criteria)
+		if err != nil {
+			return "", errors.Wrapf(err, "failed to filter subnets for criteria %q", criteria)
+		}
+		if len(ec2Subnets) == 0 {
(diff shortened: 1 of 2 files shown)


r4f4:
Instead of the validation, we can also do https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4892

Ankitasw:
@r4f4 Can you close this PR in favor of #4892 ?

PR #4892, opened 2024-03-20
🐛 ec2: instances: fix assigning public IP

### LINKED ISSUE #4888: No validation done when only `PublicIP == true` but public subnet has `MapPublicIPOnLauch == false`

r4f4:
**What type of PR is this?**
/kind bug

In the scenario where the user brings their own VPC, if no subnet ID is set in the machine spec and PublicIP is true, CAPA will choose one from the available public subnets. However, if the subnet doesn't have MapPublicIPOnLaunch == true, the instance will not be assigned a public IP. As a result, the instance will have no internet access, contrary to the user's expectation.

This change guarantees an instance will be assigned a public IP even if the subnet doesn't do it on instance launch. Instead, we set the option in the instance's network interface.

Fixes #4888

I still need to fix and add unit tests. I want to get feedback on the approach first.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Fix machines getting a public IP even when user-supplied subnets don't have MapPublicIpOnLaunch.
```


r4f4:
/hold
Until I fix the unit tests.

pkg/cloud/services/ec2/instances.go
```diff
@@ -334,7 +359,7 @@ func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
 					*subnet.SubnetId, *subnet.AvailabilityZone, *failureDomain)
 				continue
 			}
-			if scope.AWSMachine.Spec.PublicIP != nil && *scope.AWSMachine.Spec.PublicIP && !*subnet.MapPublicIpOnLaunch {
+			if wantsPublicIP && !isSubnetPublic(routeTables, *subnet.SubnetId) {
```
mtulio, line 362: @r4f4 can we use the `IsPublic` flag for the public subnet check?

```suggestion
			if wantsPublicIP && subnet.IsPublic {
```

r4f4: @mtulio not blindly. At this point all we have is an `ec2.Subnet` which doesn't have a simple `IsPublic` field. We might be able to do that if we separate the case subnetID-supplied from the subnetFilter-supplied.

mtulio: @r4f4 apologies, I should have linked the field :)

The [SubnetSpec.IsPublic](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/f67715ea3c17728fc531ec488fb0e186f31b2457/api/v1beta2/network_types.go#L413-L416) is filled when the subnet is created. Can you access in this stage the flag from the subnet ID from `subnet.SubnetId`? Something like:
```sh
s.scope.Subnets().FindByID(*subnet.SubnetId).IsPublic
```

r4f4: Aha! That makes sense. Nice suggestion, will incorporate!

r4f4: Done

damdo:
/cc. @vincepri

damdo:
@r4f4 something needs fixing in API generation and units

r4f4:
> @r4f4 something needs fixing in API generation and units. Could you PTAL? Thanks!

Sure, I also need to add unit tests.

r4f4:
@damdo I don't know how to fix the fuzzy test. Would you mind giving some pointers?

damdo:
@r4f4 The Fuzzy testing in this instance refers to the Conversion across API types conversion.
More specifically `--- FAIL: TestFuzzyConversion/for_AWSCluster` refers to the AWSCluster conversion.

Try and add 
```
dst.Status.Bastion.PublicIPOnLaunch = restored.Status.Bastion.PublicIPOnLaunch
```
after this: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/daaa1e12a8065329a5f8f2d10c207b7b739d9538/api/v1beta1/awscluster_conversion.go#L59

r4f4:
@damdo thanks! That did the trick.

damdo (commented):
Code and reasoning look reasonable.
E2Es passed.

/lgtm

/cc @nrb 
/assign @richardcase @dlipovetsky @vincepri

r4f4:
I'll rebase and squash the commits.

damdo (commented):
/lgtm

pkg/cloud/services/ec2/instances.go
```diff
@@ -538,13 +553,25 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 				DeviceIndex:        aws.Int64(int64(index)),
 			})
 		}
+		netInterfaces[0].AssociatePublicIpAddress = i.PublicIPOnLaunch
 
 		input.NetworkInterfaces = netInterfaces
 	} else {
-		input.SubnetId = aws.String(i.SubnetID)
+		if i.PublicIPOnLaunch != nil && *i.PublicIPOnLaunch {
```
vincepri, line 560: Use ptr.Deref?

api/v1beta2/types.go
```diff
@@ -237,6 +237,10 @@ type Instance struct {
 	// PrivateDNSName is the options for the instance hostname.
 	// +optional
 	PrivateDNSName *PrivateDNSName `json:"privateDnsName,omitempty"`
+
+	// PublicIPOnLaunch is the option to associate a public IP on instance launch
+	// +optional
+	PublicIPOnLaunch *bool `json:"publicIPOnLaunch,omitempty"`
```
vincepri, line 243: Would this field now allow to set a public ip on a Machine in a private subnet?

r4f4: I think so, although I haven't tested it.

pkg/cloud/services/ec2/instances.go
r4f4: Done.

vincepri (approved):
/lgtm

vincepri:
@r4f4 squash commits?

damdo (commented):
/lgtm

damdo:
@vincepri @richardcase @dlipovetsky @Ankitasw this one needs an `/approve`

richardcase:
Until the e2e pass:

/hold

richardcase:
But from my side:

/approve

When the e2e passes feel free to unhold.

damdo:
It timed out

r4f4:
Are we good to remove the hold?

PR #4909, opened 2024-04-04
:seedling: Redefine managing IAM resources: Create and Delete

### LINKED ISSUE #3715: Re-think how we manage prerequisites in clusterawsadm
/kind feature
/area clusterawsadm

Currently, we use `clusterawsadm` to create the AWS prerequisites (i.e. IAM stuff). This creates a Cloud Formation stack. 

Users are running into issues with the clusterawsadm/cloud formation approach because CloudFormation is region specific but IAM is global. Customers end up in the position where the stacks are created in multiple regions but some fail as the IAM resources have been created by another stack in another reason. Other users have run into other issues with the cloud formation stack for various reasons.

I personally think we should remove the use of Cloud Formation and look at another solution.

**Environment:**

Atharva-Shinde:
**What type of PR is this?**
/kind feature
/area clusterawsadm

Currently, CAPA manages prerequisites required by AWS through CloudFormation which has caused numerous issues to CAPA end-users. This PR works as a stepping stone in migrating away from the use of AWS CloudFormation and relying on service specific API calls to manage IAM resources and gradually make the process idempotent.

This PR introduces 2 new commands and to enable them set the `CLUSTERAWSADM_ALPHA` environment variable to `true`.

1. `clusterawsadm bootstrap iam create`: creates IAM resources(roles, instances profiles and policies) from the bootstrap configuration file (uses default bootstrap configuration if not provided)
2. `clusterawsadm bootstrap iam delete` deletes IAM resources(roles, instances profiles and policies) created using the bootstrap configuration file (uses default bootstrap configuration if not provided)

Fixes #3715

Screenshots:

Screenshot 2024-09-02 at 9 35 26 PM

Screenshot 2024-09-02 at 9 37 03 PM

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note

```


nrb:
/retitle :seedling: Redefine managing IAM resources: Create and Delete

Updating the title should cause the PR verify job to pass.

Atharva-Shinde:
Thanks @nrb I've addressed the CI failures :)

cmd/clusterawsadm/iamservice/service.go
```diff
@@ -0,0 +1,500 @@
...
+				return errors.Wrapf(err, "failed to attach instance profile \"%s\" to IAM role \"%s\"", res.InstanceProfileName, roleName)
+			}
+		}
+	}
+	return nil
+}
+
+func attachPoliciesToRole(rolename *string, awsManagedPolicies []string, client *iam.IAM) error {
+	if awsManagedPolicies == nil {
+		// klog.Warningf("no policies defined to attach to the IAM role \"%s\"", *rolename) // TODO
```
nrb, line 258: What's the next action here? Just enabling the warning, or something else?

nrb: Reading further, looks like there's a question about outputting the ARNs elsewhere - is this the same?

cmd/clusterawsadm/iamservice/service.go
```diff
@@ -0,0 +1,500 @@
...
+			PolicyArn: &policy,
+		})
+		if err != nil {
+			if aerr, ok := err.(awserr.Error); ok {
+				switch aerr.Code() {
+				case iam.ErrCodeEntityAlreadyExistsException:
+					klog.Warningf("IAM role \"%s\" is already attached to policy", *rolename) // TODO should we output the policy arn? how safe is it
+					continue
+				default:
+					return errors.Wrapf(err, "failed to attach IAM role \"%s\" to policy", *rolename) // TODO should we output the policy arn? how safe is it
```
nrb, line 275: Thinking about where this command should be running, I can think of a couple environments:

First, a user's machine, while they set up a management cluster. I think outputting an ARN here is fine. They have some sort credentials on their machine, which are the higher risk asset.


Next, CI, which might be spinning up resources, either for testing or as part of a larger provisioning system. In this case, the ARNs would probably get output into logs. Hopefully, the CI is using short-lived credentials, but again, those would be the more valuable asset.

It's true that policy ARNs could be observed and an attacker could attempt to edit the policy to expand permissions. However, I'm not sure the actual ARN is all that important, since what really matters are the permissions to actually overwrite the policy the ARN points to.

Does that make sense?

cmd/clusterawsadm/iamservice/service.go
```diff
@@ -0,0 +1,500 @@
...
+	if err != nil {
+		return nil, errors.Wrap(err, "internal server error")
+	}
+	return iam.New(s), nil
+}
+
+func prioritySet(t go_cfn.Template) (rmap map[string][]go_cfn.Resource, err error) {
+	rmap = map[string][]go_cfn.Resource{}
+	for _, resource := range t.Resources {
+		if resource.AWSCloudFormationType() == configservice.ResourceTypeAwsIamRole {
```
nrb, line 67: The description says that this PR is looking to move away from CloudFormation, but we're using CloudFormation files here.

Is the intent to use CloudFormation templates as the input still, but introspect them and make our own API calls vs submitting them directly to the CF service endpoints?

nrb (commented):
This seems reasonable to me overall. I've added some questions to help me understand where this might go in the future before I approve it.

cmd/clusterawsadm/iamservice/service.go
Atharva-Shinde: I think at that time I was wondering if we should issue a warning that the user hasn't defined any awsManagedPolicies in their CloudFormation config or just return back to parent call directly without any warnings.

cmd/clusterawsadm/iamservice/service.go
Atharva-Shinde: The intent as discussed with my mentors Richard and Ankita was to think of a solution that eventually gets rid of the CloudFormation to manage IAM resources. The end goal is to create an idempotent solution that enforces API calls to AWS to manage the IAM resources and doesn't rely on CloudFormation Stack. And the intent is to do this incrementally and not all at once so that we avoid big code change altogether and maintain a low to no user side changes.

So this PR is the first step in achieving the above, where we rely on the CF config(template) to get the input and make API calls to manage these IAM resources without relying on CloudFormation stack.

cmd/clusterawsadm/iamservice/service.go
Atharva-Shinde: So should we give the warning?

cmd/clusterawsadm/iamservice/service.go
Atharva-Shinde: Yeah that make sense :) I'll make the change to include the ARNs in the outputs above.

richardcase:
Sorry @Atharva-Shinde i've not had a huge amount of time recently. I will make sure i spent time reviewing.

richardcase:
/lifecycle hold

We really need this. To get some extra eyes on it:

/help

richardcase:
Myself & @Atharva-Shinde just had a good chat on this. Notes from the call:

Initial items to work on:

1. Move the new command so its marked as alpha

Option 1

this
`clusterawsadm bootstrap iam create`
becomes:
`clusterawsadm alpha bootstrap iam create`

Option 2

```shell
export CLUSTERAWSADM_ALPHA=true
clusterawsadm bootstrap iam create
```


2. Add tests to determine if it created/deleted the IAM resources as expected
3. Ensure updates can support adding a new permission to existing policy


For the future:

- Always add a tag to say "owned" by CAPA

For Rich

- Create a graduation acceptance test to compare resources created by both sets of commands

damdo:
@Atharva-Shinde this should probably be updated to AWS SDK v2 now that we have migrated to it across the board for CAP. Thanks

Atharva-Shinde:
I will prioritize this next week.

PR #4943, opened 2024-04-22
🌱 test: add failure message to Eventually().Should(bool) everywhere

damdo:
**What type of PR is this?**
/kind cleanup

In Gomega, when Eventually().Should(<bool>) fails, the failure returns a 
```
  [FAILED] Timed out after 1200.001s.
  Expected
      <bool>: false
  to be true
```
Which takes a bit of time to then debug further.

There is a more immediate way to add some context to this, by adding a failure message to the should like so: `.Should(x, "message")`. This PR does this for all the current occurrences of this construct.

**Release note**:

```release-note
test: add failure message to Eventually().Should(bool) everywhere
```

### DIFF
--- a/test/e2e/suites/managed/machine_deployment_helpers.go
+++ b/test/e2e/suites/managed/machine_deployment_helpers.go
@@ -58,7 +58,7 @@ func waitForMachineDeploymentDeleted(ctx context.Context, input waitForMachineDe
 		err := input.Getter.Get(ctx, key, mp)
 		notFound := apierrors.IsNotFound(err)
 		return notFound
-	}, intervals...).Should(BeTrue())
+	}, intervals...).Should(BeTrue(), fmt.Sprintf("Eventually failed waiting for MachineDeployment %q to be deleted", input.MachineDeployment.GetName()))
 }
 
 type waitForMachineDeletedInput struct {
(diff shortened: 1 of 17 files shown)


test/e2e/suites/managed/machine_deployment_helpers.go
```diff
@@ -58,7 +58,7 @@ func waitForMachineDeploymentDeleted(ctx context.Context, input waitForMachineDe
 		err := input.Getter.Get(ctx, key, mp)
 		notFound := apierrors.IsNotFound(err)
 		return notFound
-	}, intervals...).Should(BeTrue())
+	}, intervals...).Should(BeTrue(), fmt.Sprintf("Eventually failed waiting for MachineDeployment %s to be deleted", input.MachineDeployment.GetName()))
```
AndiDog, line 61: Can you please change _all occurrences_ into `%q` instead of `%s`? If a value is empty by mistake, which is very likely when writing tests, it's super-hard to see two spaces as compared to a quoted string that clearly says it's empty: `waiting for MachineDeployment  to be deleted` vs. `waiting for MachineDeployment "" to be deleted`.

AndiDog (commented):
Otherwise LGTM and great change!

test/e2e/suites/managed/machine_deployment_helpers.go
damdo: Good suggestion! Updated thanks Andreas!

AndiDog (commented):
/lgtm

dlipovetsky:
/approve

Thanks for improving the tests!

PR #4976, opened 2024-05-10
🐛: skip APIServerELB DNS name resolution if internal

### LINKED ISSUE #4975: APIServerLB DNS name resolution causes deploy failure on air-gapped systems
/kind bug

After CAPA creates the APIServer Load Balancer and it gets a DNS name, CAPA waits for the DNS name to resolve before continuing the installation [1][2]. I think this check was never reconsidered when support for a secondary LB was added.

Now consider the scenario where a cluster is deployed on an air-gapped system. Trying to resolve the LB DNS name will never succeed .

[1] https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/1641
[2] https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/1651

Only check for DNS name resolution when the APIServer LB is public.

Air-gapped installations are a requirement in C2S/SC2S secret regions.

**Environment:**

- Cluster-api-provider-aws version: master

r4f4:
**What type of PR is this?**

/kind bug

Skip APIServerELB DNS name resolution check when the LB is internal. The check will never work in air-gapped systems.

Fixes #4975

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Skip DNS name resolution check for internal APIServerELB.
```

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -288,12 +288,16 @@ func (r *AWSClusterReconciler) reconcileLoadBalancer(clusterScope *scope.Cluster
 		return &retryAfterDuration, nil
 	}
 
-	clusterScope.Debug("Looking up IP address for DNS", "dns", awsCluster.Status.Network.APIServerELB.DNSName)
-	if _, err := net.LookupIP(awsCluster.Status.Network.APIServerELB.DNSName); err != nil {
-		clusterScope.Error(err, "failed to get IP address for dns name", "dns", awsCluster.Status.Network.APIServerELB.DNSName)
-		conditions.MarkFalse(awsCluster, infrav1.LoadBalancerReadyCondition, infrav1.WaitForDNSNameResolveReason, clusterv1.ConditionSeverityInfo, "")
-		clusterScope.Info("Waiting on API server ELB DNS name to resolve")
-		return &retryAfterDuration, nil
+	if awsCluster.Status.Network.APIServerELB.Scheme != infrav1.ELBSchemeInternal {
+		clusterScope.Debug("Looking up IP address for DNS", "dns", awsCluster.Status.Network.APIServerELB.DNSName)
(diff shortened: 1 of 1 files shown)


damdo (commented):
/lgtm

r4f4:
/hold
While we double check if the name is resolvable in a SC2S region.

AndiDog:
There's a [chat thread](https://kubernetes.slack.com/archives/CD6U2V71N/p1696888521817979) about why the DNS resolving wait code was likely there. It might be to check LB availability. Did you see problems or delays after this change, such as CAPA requiring an additional wait before reconciling further (e.g. because the LB/DNS wasn't available yet)? That's the only concern I have. Otherwise LGTM.

r4f4:
> There's a [chat thread](https://kubernetes.slack.com/archives/CD6U2V71N/p1696888521817979) about why the DNS resolving wait code was likely there. It might be to check LB availability. Did you see problems or delays after this change, such as CAPA requiring an additional wait before reconciling further (e.g. because the LB/DNS wasn't available yet)? That's the only concern I have. Otherwise LGTM.

When we encountered the issue we were creating a cluster in C2S via an emulator (SHIFT). The issue did not happen in another emulator (Combine). We're trying to test on a real C2S/SC2S env to determine if the internal LB name is resolvable. If so, it's an emulator bug. If not, I'll test again and pay attention to the reconcile process.

AndiDog:
Please also test on a real, public AWS region if you can. It would be interesting to see how removing this wait behaves, in order to find out whether it could be removed overall, or if then we need some retries for the control plane setup or so.

nrb:
I'll wait to review until hearing back on whether or not this happens in the real AWS environment.

r4f4:
We could not reproduce in a real environment and SHIFT confirmed there was a bug in the emulator. There is still a possible problem with waiting for the DNS resolution but I'll open a separate issue for that.

/close

PR #4996, opened 2024-05-30
🌱  Bump setup-envtest to use new controller-tools binaries

damdo:
**What type of PR is this?**
/kind cleanup

This was brought up in the weekly CAPI community meeting.

This is a commit from CR main (22.05.2024).
Intentionally using a commit from main to use a setup-envtest version that uses binaries from controller-tools, not GCS. CR PR: https://github.com/kubernetes-sigs/controller-runtime/pull/2811

**Release note**:

```release-note
NONE
```

TODO:
- [x] Bump `hack/tools` too once this is addressed: https://github.com/kubernetes-sigs/controller-runtime/pull/2837#issuecomment-2139918231


damdo:
/hold
Until WIP

hack/tools/go.mod
```diff
@@ -21,7 +21,7 @@ require (
 	k8s.io/gengo v0.0.0-20230829151522-9cce18d56c01
 	k8s.io/klog/v2 v2.120.1
 	sigs.k8s.io/cluster-api/hack/tools v0.0.0-20221121093230-b1688621953c
-	sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20211110210527-619e6b92dab9
+	sigs.k8s.io/controller-runtime/tools/setup-envtest v0.0.0-20240531134648-6636df17d67b
```
nrb, line 24: I'm a little confused - Why doesn't this match the version in the `Makefile`? Is that on purpose?

damdo: It didn't matter as the important thing is for the one in `hack/tools/go.mod]` to be referencing the module with the compatible Go version.
The other one is a plain go install of the binary which doesn't interfere with the Go versioning.
Anyway I aligned both now.

damdo:
@nrb should be ready for labelling.

PR #4998, opened 2024-05-31
✨ ROSA: Add upgrade config to ROSAMachinePool

serngawy:
**What type of PR is this?**
/kind feature

Add update configuration to ROSAMachinePool  APIs.

Fixes #

**Checklist**:

- [ X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add UpdateConfig to ROSAMachinePool APIs.
```


exp/api/v1beta2/rosamachinepool_types.go
```diff
@@ -105,6 +105,11 @@ type RosaMachinePoolSpec struct {
 	//
 	// +optional
 	NodeDrainGracePeriod *metav1.Duration `json:"nodeDrainGracePeriod,omitempty"`
+
+	// UpgradeManagementConfig specifies upgrade management configuration which set MaxUnavailable & MaxSurge number of nodes during upgrade.
+	//
+	// +optional
+	UpgradeManagementConfig *RosaMachinePoolUpgradeConfig `json:"upgradeManagementConfig,omitempty"`
```
muraee, line 112:
```suggestion
	RollingUpdate *RosaRollingUpdateConfig  `json:"rollingUpdate,omitempty"`
```

exp/api/v1beta2/rosamachinepool_types.go
```diff
@@ -134,6 +139,43 @@ type RosaMachinePoolAutoScaling struct {
...
+	// Example: when MaxUnavailable is set to 30%, old nodes can be deleted down to 70% of
+	// desired nodes immediately when the rolling update starts. Once new nodes
+	// are ready, more old nodes be deleted, followed by provisioning new nodes,
+	// ensuring that the total number of nodes available at all times during the
+	// update is at least 70% of desired nodes.
+	//
+	// +kubebuilder:validation:Pattern="^((100|[0-9]{1,2})%|[0-9]+)$"
+	// +kubebuilder:default=0
+	// +optional
+	MaxUnavailable string `json:"maxUnavailable,omitempty"`
```
muraee, line 159: please use `intstr.IntOrString` type instead of `string`.
see: https://github.com/kubernetes-sigs/cluster-api/blob/6941e8c66f3de69adfb1f19893afc0f3cf4571e6/api/v1beta1/machinedeployment_types.go#L186

exp/api/v1beta2/rosamachinepool_types.go
```diff
@@ -134,6 +139,43 @@ type RosaMachinePoolAutoScaling struct {
...
+	// Example: when MaxSurge is set to 30%, new nodes can be provisioned immediately
+	// when the rolling update starts, such that the total number of old and new
+	// nodes do not exceed 130% of desired nodes. Once old nodes have been
+	// deleted, new nodes can be provisioned, ensuring that total number of nodes
+	// running at any time during the update is at most 130% of desired nodes.
+	//
+	// +kubebuilder:validation:Pattern="^((100|[0-9]{1,2})%|[0-9]+)$"
+	// +kubebuilder:default=1
+	// +optional
+	MaxSurge string `json:"maxSurge,omitempty"`
```
muraee, line 176: same comment as above.

muraee:
please also update `nodePoolToRosaMachinePoolSpec()` function
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b99443a6c79f0a03543b5ce38bab2fbf471c9373/exp/controllers/rosamachinepool_controller.go#L483

exp/api/v1beta2/rosamachinepool_types.go
serngawy: I'm not against  the naming but in OCM lib it is NodePoolManagementUpgrade, in hyperShift it is NodePoolManagement. Also it may confuse with the rollingUpgrade for other CRs. I'm more to keep the same name.

muraee: OCM naming is irrelevant. and its also `rollingUpdate` in Hypershift (https://github.com/openshift/hypershift/blob/afd17d6a127d4bcffe31a063af988973b495b0e1/api/hypershift/v1beta1/nodepool_types.go#L277). its just wrapped within other fields.

> Also it may confuse with the rollingUpgrade for other CRs

This is what its called in CAPI, so CAPI users will know immediately what the field does.

exp/api/v1beta2/rosamachinepool_types.go
```diff
@@ -134,6 +140,45 @@ type RosaMachinePoolAutoScaling struct {
...
+	//
+	// Example: when MaxUnavailable is set to 30%, old nodes can be deleted down to 70% of
+	// desired nodes immediately when the rolling update starts. Once new nodes
+	// are ready, more old nodes be deleted, followed by provisioning new nodes,
+	// ensuring that the total number of nodes available at all times during the
+	// update is at least 70% of desired nodes.
+	//
+	// +kubebuilder:validation:Pattern="^((100|[0-9]{1,2})%|[0-9]+)$"
+	// +kubebuilder:validation:XIntOrString
+	// +kubebuilder:default=0
```
muraee, line 159: I don't think we need to provide defaults. OCM might have different values or might change them anytime.

serngawy: If OCM will change that it is API changes and we need to change it as well. Those default values same as the Hypershift default values and we have to expose it to the user.

exp/api/v1beta2/rosamachinepool_types.go
serngawy: right, so its wrapped in another field in hypershift as the upgrade Type is included [here](https://github.com/openshift/hypershift/blob/afd17d6a127d4bcffe31a063af988973b495b0e1/api/hypershift/v1beta1/nodepool_types.go#L356) (ocm had the management name as all fields together) , However I didn't expose that to CAPA as its only support Replace type for now.

exp/api/v1beta2/rosamachinepool_types.go
muraee: > If OCM will change that it is API changes and we need to change it as well.

that's the point, if we don't set defaults in the client side, then we don't have to change them when the backend(OCM) change theirs. If the fields are not set explicitly by the user, we send empty values to OCM so they can set the appropriate defaults.
Otherwise, we will end up in the same situation as the `autoRepair`

exp/api/v1beta2/rosamachinepool_types.go
muraee: exactly, we are not exposing any of the other fields. We are only exposing `RollingUpdate` fields, so we need to name it accordingly:

again, the field is named `RollingUpdate` in hypershift (no that it matters): https://github.com/openshift/hypershift/blob/afd17d6a127d4bcffe31a063af988973b495b0e1/api/hypershift/v1beta1/nodepool_types.go#L277-L282

but most importantly is named in CAPI `RollingUpdate` https://github.com/kubernetes-sigs/cluster-api/blob/6941e8c66f3de69adfb1f19893afc0f3cf4571e6/api/v1beta1/machinedeployment_types.go#L164

Once we have support for more fields, update strategies, etc. Then we can think about wrapping this in another upper management field.

serngawy:
@damdo would you approve this PR for test

exp/api/v1beta2/rosamachinepool_types.go
```diff
@@ -134,6 +140,51 @@ type RosaMachinePoolAutoScaling struct {
...
+	// Example: when MaxUnavailable is set to 30%, old nodes can be deleted down to 70% of
+	// desired nodes immediately when the rolling update starts. Once new nodes
+	// are ready, more old nodes be deleted, followed by provisioning new nodes,
+	// ensuring that the total number of nodes available at all times during the
+	// update is at least 70% of desired nodes.
+	//
+	// +kubebuilder:validation:Pattern="^((100|[0-9]{1,2})%|[0-9]+)$"
+	// +kubebuilder:validation:XIntOrString
+	// +optional
+	MaxUnavailable intstr.IntOrString `json:"maxUnavailable,omitempty"`
```
muraee, line 168: should be a pointer
```suggestion
	MaxUnavailable *intstr.IntOrString `json:"maxUnavailable,omitempty"`
```

serngawy: no need for pointers, we don't want to check for nil values. empty value will be submitted anyway.

muraee: I ma pretty sure its 0 :)  have you actually tested this?

serngawy: yes Mulham, its empty string and 0 all the time, type what determine which value (that is why originally data type was string to avoid this conversation issues. OCM uses string only) unit test is updated. Another issue I found, OCM raise error if both maxUnavailable & maxSurge not set like you cannot set the maxSurge and rely on OCM to set default for maxUnavailable (like the unit-test). I will update the PR today to avoid this.

dlipovetsky:
This adds some API fields.

/kind area/api

Issue #4743, opened 2024-01-18
EKS VPC CNI cannot be disabled because AWS now installs via Helm


AndiDog:
/kind bug

**What steps did you take and what happened:**

Creating an EKS cluster with a custom CNI (such as Cilium) is currently problematic because CAPA does not correctly remove the automatically preinstalled VPC CNI. That can be seen by `aws-node` pods running on the cluster.

CAPA has code to delete the AWS VPC CNI resources if `AWSManagedControlPlane.spec.vpcCni.disable=true`, but only deletes if they're not managed by Helm. I presume that is on purpose so that users of CAPA could deploy the VPC CNI with Helm in their own way.

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/3618d1c1567afe72781059cd9a3498e7ae44b3b5/pkg/cloud/services/awsnode/cni.go#L269-L293

Unfortunately, it seems that AWS introduced a breaking change by switching their own automagic deployment method to Helm, including the relevant labels. This is what a newly-created EKS cluster looks like (VPC CNI _not_ disabled, cluster created by CAPA ~v2.3.0):

```
$ kubectl get ds -n kube-system aws-node -o yaml
apiVersion: apps/v1
kind: DaemonSet
metadata:
  annotations:
    deprecated.daemonset.template.generation: "1"
  creationTimestamp: "2024-01-18T15:40:42Z"
  generation: 1
  labels:
    app.kubernetes.io/instance: aws-vpc-cni
    app.kubernetes.io/managed-by: Helm
    app.kubernetes.io/name: aws-node
    app.kubernetes.io/version: v1.15.1
    helm.sh/chart: aws-vpc-cni-1.15.1
    k8s-app: aws-node
  name: aws-node
  namespace: kube-system
  # [...]
spec:
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      k8s-app: aws-node
  template:
    metadata:
      creationTimestamp: null
      labels:
        app.kubernetes.io/instance: aws-vpc-cni
        app.kubernetes.io/name: aws-node
        k8s-app: aws-node
```

The deletion code must be fixed. Sadly, AWS does not provide extra labels to denote that the deployment is AWS-managed. And this breaking change even applies to older Kubernetes versions like 1.24.

Related: [E2E test wanted to cover this feature (issue)](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3544)

**Environment:**

[...]

PR #5009, opened 2024-06-10
:bug: Remove Helm condition for AWS VPC CNI deletion

jas-nik:
**What type of PR is this?**
/kind bug

Recently, AWS changed the way VPC CNI is installed from manifest apply to managing it via Helm (ref: https://github.com/aws/amazon-vpc-cni-k8s/issues/2763#issuecomment-1908510232). This makes `AWSManagedControlPlane.Spec.VpcCni.Disable` flag obsolete due to the `app.kubernetes.io/managed-by: Helm` condition in deleteCNI flow.

This PR removes that condition to make the flag functional.

Fixes # [4743](https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4743)

For users replacing the default VPC CNI with their own helm-managed VPC CNI - They will need to set  `AWSManagedControlPlane.Spec.VpcCni.Disable` to `false` (default) OR set label on VPC CNI resource `"prevent-deletion": "true"`

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Flag to remove the default Helm-managed AWS VPC CNI EKS Addon
```

### DIFF
--- a/pkg/cloud/services/awsnode/cni.go
+++ b/pkg/cloud/services/awsnode/cni.go
@@ -31,7 +31,6 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"sigs.k8s.io/kustomize/api/konfig"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/awserrors"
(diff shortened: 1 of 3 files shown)


jas-nik:
@nrb  can I get a review on this please?

richardcase (commented):
Thanks @jas-nik . Generally this looks good to me. 

I think as part of this we should also have the option suggested by @AndiDog to have a means to ignore deletion if there is a specific label (https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/4743#issuecomment-1907565715)

jas-nik:
@richardcase  - Done. Added a label `prevent-deletion`, please review when you get a chance.

pkg/cloud/services/awsnode/cni.go
```diff
@@ -272,22 +271,23 @@ func (s *Service) deleteResource(ctx context.Context, remoteClient client.Client
...
-			} else {
-				s.scope.Debug(fmt.Sprintf("resource %s was deleted", key))
-			}
-		} else {
-			s.scope.Debug(fmt.Sprintf("resource %s is managed by helm, not deleted", key))
+		return nil
+	}
+	// Check if the "prevent-deletion" label is present and set to "true", skip if present
+	labels := obj.GetLabels()
+	if value, exists := labels["prevent-deletion"]; exists && value == "true" {
```
alexander-demicev, line 278: what do you think about checking the key presence instead of comparing it to `true` value? It's easy to make a mistake for the user when setting it because keys and values are just strings and we can't validate them. Adding "prevent-deletion" even with an empty value can be enough.

jas-nik: Thanks for the review. That sounds fine to me.

cnmcavoy:
/lgtm

We probably need to document the label in the future.

richardcase:
> We probably need to document the label in the future.

Thats a good point @cnmcavoy . I think we should probably include documentation with this PR. wdyt @jas-nik @cnmcavoy ?

jas-nik:
Updated `pod networking` doc

pkg/cloud/services/awsnode/cni.go
```diff
@@ -272,22 +271,23 @@ func (s *Service) deleteResource(ctx context.Context, remoteClient client.Client
...
+	// Check if the "prevent-deletion" label is present and set to "true", skip if present
+	labels := obj.GetLabels()
+	if _, exists := labels["prevent-deletion"]; exists {
+		s.scope.Debug(fmt.Sprintf("resource %s has 'prevent-deletion' label, skipping deletion", key))
+		return nil
+	}
+	// Delete the resource
+	if err := remoteClient.Delete(ctx, obj, &client.DeleteOptions{}); err != nil {
+		if !apierrors.IsNotFound(err) {
+			return fmt.Errorf("deleting %s: %w", key, err)
```
AndiDog, line 285:
```suggestion
			return fmt.Errorf("failed to delete %s: %w", key, err)
```

pkg/cloud/services/awsnode/cni.go
```diff
@@ -272,22 +271,23 @@ func (s *Service) deleteResource(ctx context.Context, remoteClient client.Client
...
-				s.scope.Debug(fmt.Sprintf(
-					"resource %s was not found, not deleted", key))
-			} else {
-				s.scope.Debug(fmt.Sprintf("resource %s was deleted", key))
-			}
-		} else {
-			s.scope.Debug(fmt.Sprintf("resource %s is managed by helm, not deleted", key))
+		return nil
+	}
+	// Check if the "prevent-deletion" label is present and set to "true", skip if present
```
AndiDog, line 276:
```suggestion
	// Don't delete if the `prevent-deletion` label exists. It could be there because CAPA added it (see below),
	// or because it was added externally, for example if a custom version of AWS CNI was already installed.
	// Either way, CAPA should not delete such a labelled CNI installation.
```

AndiDog (commented):
Minor wording comments, otherwise

/lgtm

richardcase:
@jas-nik  - thanks for the updates. Could you resolve the 2 open items and then i think we are good to go. Thank you so much for this.

richardcase:
There was previous lgtm (and only a comment change) so:

/lgtm
/approve

PR #5048, opened 2024-07-11
🐛 Fix subtests running in same environment and reconciliation bug which was not covered because of that

AndiDog:
**What type of PR is this?**

/kind bug

I was the author of those tests. They shared one envtest and namespace by mistake, leading to conflicts. After fixing this, luckily, only one new problem appeared which the conflicting tests previously couldn't catch: the reconciliation would return after setting `AWSMachinePool.status.launchTemplateVersion`, but can just continue (else we would have to call 3x `reconcileNormal` in the test to make it all happen). One test was checking for something it didn't even need (again: `AWSMachinePool.status.launchTemplateVersion`).

n/a

**Special notes for your reviewer**: n/a

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Continue reconciliation after filling an empty `AWSMachinePool.status.launchTemplateVersion` field
```

### DIFF
--- a/exp/controllers/awsmachinepool_controller_test.go
+++ b/exp/controllers/awsmachinepool_controller_test.go
@@ -472,15 +472,15 @@ func TestAWSMachinePoolReconciler(t *testing.T) {
 		})
 
 		t.Run("ReconcileLaunchTemplate not mocked", func(t *testing.T) {
-			g := NewWithT(t)
-			setup(t, g)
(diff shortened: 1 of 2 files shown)


fiunchinho (approved):
/lgtm

dlipovetsky:
Thanks!

/approve

PR #5057, opened 2024-07-15
✨ Ensure Volumes and Network Interfaces are also tagged when created as part of instance creation

fahlmant:
**What type of PR is this?**

/kind feature

ENIs and EBS Volumes that are created as part of instance creation are not being tagged with the tags that are applied to the instance. Since these resources are created as part of the AWS API call to create the instance itself, it follows that the same tags would be applied at creation time. AWS allows for restrictions against CreateTags as a standalone API call vs CreateTags as part of a call to the RunInstances API call. This change allows restrictions against standalone CreateTags calls against existing resources, while still allowing the creation of tags at on the resource.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
- EBS Volumes and ENIs are now tagged when created as part of EC2 Instance Creation. 
```

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -616,21 +616,25 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 	}
 
 	if len(i.Tags) > 0 {
-		spec := &ec2.TagSpecification{ResourceType: aws.String(ec2.ResourceTypeInstance)}
-		// We need to sort keys for tests to work
-		keys := make([]string, 0, len(i.Tags))
-		for k := range i.Tags {
-			keys = append(keys, k)
-		}
-		sort.Strings(keys)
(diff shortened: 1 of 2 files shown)


fiunchinho (changes requested):
Please, update the tests so that the new behavior is also tested.

fahlmant:
@fiunchinho tests added

fiunchinho (approved):
lgtm

fahlmant:
@richardcase The EKS test seems to be failing due to the CF stack not creating the expected IAM roles. Any guidance on what's going wrong?

richardcase:
@fahlmant - i see it passing now. Sometimes we have flakes. If it persistently fails then it would be an issue with a specific test node and we'd need to look into that.

richardcase:
Considering @fiunchinho previously did an lgtm:

/lgtm
/approve

PR #5063, opened 2024-07-19
Remove sorting from instance keys when creating input tag specifications

AlexVulaj:
**What type of PR is this?**

/kind cleanup

Currently when generating input tag specifications, we sort all of the keys from the given `*infrav1.Instance` before appending them to `input.TagSpecifications`. From the comments, this is done "for tests to work". This currently comes at the performance cost of iterating through all of the tags to grab the keys, and then sorting them before iterating the list again to add them to our input.

By utilizing a custom `gomock.Matcher`, we can transfer that performance hit from our production code to our tests.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
NONE
```


pkg/cloud/services/ec2/instances.go
```diff
@@ -619,17 +619,10 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
...
-			keys := make([]string, 0, len(i.Tags))
-			for k := range i.Tags {
-				keys = append(keys, k)
-			}
-			sort.Strings(keys)
-			for _, key := range keys {
+			for tagKey, tagValue := range i.Tags {
 				spec.Tags = append(spec.Tags, &ec2.Tag{
-					Key:   aws.String(key),
-					Value: aws.String(i.Tags[key]),
```
iamkirkbater, line 632: Isn't there a golintci check that would fail here because we're using `tagValue` instead of `i.Tags[tagKey]` with `i` being a pointer?

pkg/cloud/services/ec2/instances_test.go
```diff
@@ -45,6 +47,69 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
+type runInstancesInputMatcher struct {
+	runInstancesInput *ec2.RunInstancesInput
+}
+
+func (m runInstancesInputMatcher) Matches(arg interface{}) bool {
+	riiArg := arg.(*ec2.RunInstancesInput)
```
iamkirkbater, line 55: Do we need some kind of error check here in case the arg cannot be cast to the proper type?

iamkirkbater (commented):
It's been a while since I've been deep into coding world, but just saw two things that immediately stood out to me as potential problems down the road. Not sure if the one about the for loop matters in this case because I don't specifically remember the conditions in which it's "bad" but figured I'd bring it up.

pkg/cloud/services/ec2/instances.go
AlexVulaj: Issues with using loop variables inside of the loop itself mostly become issues with things like goroutines, where the variable could change before it is used by the nested functions. In this case, we're only doing assignments in serial so there's no concern.

iamkirkbater: Ahh - yeah makes sense. Thanks for the explanation!

pkg/cloud/services/ec2/instances_test.go
iamkirkbater:
```suggestion
	riiArg, ok := arg.(*ec2.RunInstancesInput)
	if !ok {
		return false
	}
```

Something like this?

AlexVulaj: Good catch, we do actually need this. I'll apply this patch now.

pkg/cloud/services/ec2/instances.go
r4f4: @AlexVulaj I think @iamkirkbater has a point here. What you said is true if you are using the _value_ of the loop variables. However `aws.String` is taking their address. Before Golang 1.22, loop variables are reused, so `spec.Tags` elements are pointing to the same value (the last one in `i.Tags`) at the end of the loop.

For example, the following go code:
```go
package main

import "fmt"

type Tag struct {
	key   *string
	value *string
}

func main() {
	tags := map[string]string{"key1": "value1", "key2": "value2"}
	saved := make([]*Tag, 0, 2)
	for key, val := range tags {
		saved = append(saved, &Tag{
			key:   &key,
			value: &val,
		})
	}
	fmt.Println(saved)
	for _, tag := range saved {
		fmt.Printf("%s: %s\n", *tag.key, *tag.value)
	}
}
```
outputs with Golang 1.21:
```
root@6b38e5f25656:/go/src# ./loop
[0xc0000140a0 0xc0000140b0]
key2: value2
key2: value2
```
with Golang 1.22:
```
$ ./loop
[0xc000096050 0xc000096080]
key2: value2
key1: value1
```

AlexVulaj: I'll note - I'm surprised the tests didn't catch this, but it's probably because I'm running go 1.22

AlexVulaj: I will say - I'm surprised that CI didn't catch this. Locally I'm running go 1.22 so it makes sense, but the green CI checks have me wondering if something's wrong with my matcher. I wasn't able to determine anything looking deeper locally.

AlexVulaj: Ah, well we all learned something today! I'll revert that last change real quick then.

r4f4 (commented):
/lgtm

pkg/cloud/services/ec2/instances.go
AlexVulaj: For any other interested parties, this is the `aws.String` function in its entirety:
```
func String(v string) *string {
    return &v
}
```

So indeed as @r4f4 pointed out above, it's returning an address to the value passed in to the function, which is a different memory address than what's in the original caller.

AlexVulaj:
I rebased my PR today - not sure I understand the error in the PR Verify check but if anyone can point out what I need to do it'd be appreciated, thanks!

damdo:
@AlexVulaj would you be able to rebase this one? TY

damdo:
@AlexVulaj would you be able to rebase this? Thanks!

PR #5039, opened 2024-06-27
🐛 ec2/byoip: fix EIP leak when creating machine

mtulio:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

The instance creation flow is creating by default EIP to instances even if the BYO IP flow is set. BYO IPv4 creates and associates the EIP to instance after it is created, preventing paying for additional EIP (amazon-provided) when creating the instance when the BYO IPv4 Pool is defined to be used by the machine.

Furthermore, the fix provides additional checks to prevent duplicated EIP in the BYO IP reconciliation loop. The extra checks include running the EIP association many times, while the EIP is already associated, and failures in the log when running the EIP association prematurely - when the instance isn't ready, Eg ec2 in pending state.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #5038

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation (N/A)
- [x] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
fix duplicated/leaked EIP when using BYO IPv4 on Machines.
```

PR #5109, opened 2024-08-28
feat: Add IRSA support for self-managed clusters (rebase)

### LINKED ISSUE #3560: Support for non-EKS IRSA

sl1pm4t:
**What type of PR is this?**

/kind feature

Adds IRSA functionality to an IaaS cluster. This will bring the IaaS clusters inline in functionality with Managed clusters which provide most of this functionality out of the box.

With this PR, the following new resources are created:
* S3 bucket ( if not already created for the Ignition feature )
* Two public world readable objects in the S3 bucket:
  * `<cluster_name>/.well-known/openid-configuration` - OpenID Connect discovery document
  * `<cluster_name>/openid/v1/jwks` - Service Account signing public key
* AWS IAM Identity Provider, configured to trust the Issuer found at the S3 URL where the OIDC discovery doc and keys are published.

This is a continuation of an old unmerged PR #4094 - with some fixes and some functionality removed to reduce the scope of the PR. 

The functionality removed includes:
* This PR no longer deploys the amazon-pod-identity-webhook addon to the workload cluster. I felt there are already many ways to manage cluster addons, including ClusterResourceSets or CAAPH, and that it was unnecessary to install the addon via the controller which then becomes an ongoing maintenance burden. Instead, the requirement for the addon has been added to the documentation.
* This PR no longer modifies the API Server `service-account-issuer` argument through kubeadm patches. This is easily covered in the documentation and only requires a single line of config to be added to the `AWSCluster` resource, but also during testing I experienced issues with this being applied inconsistently, resulting in different values across the control plane nodes.

Fixes #3560
Supersedes #4094

This PR adds a new ReconcileOIDCProvider to the AWSCluster reconciliation loop.

* Created a new IAM service, in the future the same logic for EKS could be combined as previously the OIDC code was buried in the EKS service. Details on the reconciler can be found in [comments](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4094/files#diff-a41f3e76f5a70a6f449fecf5766e35a00299daf53c97fa84838d557f8c4766d7R44-R53).
* Extend the S3 service to allow uploading data to arbitrary keys, that can also be marked as public (OIDC discovery docs need to be public for 3rd party systems to retrieve public keys etc.)
* Exposed a ManagementClient and RemoteClient for both cluster types and exported Client.
* Moved OIDCProvider status type to v1beta2 and migrated out of the EKS API to make one type both clusters can reference a single type.
* This PR adds a new Experimental feature flag to enable this functionality. This feature is dependent on the S3 bucket associated with Ignition node configuration, but it felt unintuitive to need to enable the Ignition feature flag to get OIDC support.

**Checklist**:

- [X] squashed commits
- [X] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add IRSA support for self-hosted clusters
```


sl1pm4t:
There is some investigation of the failing e2e test happening over [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5039#issuecomment-2327100978).

sl1pm4t:
> @sl1pm4t do you need help with rebase?

No I will be able to rebase. I had just forgotten this was needed. I'll try to do this in the next day or two.

sl1pm4t:
I rebased and re-ran `make generate` which updated `config/rbac/role.yaml` to what is seen in the latest commit.

richardcase:
Both e2e have passed 🎉 
- https://gitlab.aws.omniva.com/ise/ccppday1/cluster-api-control-plane-provider-hosted/-/tree/main/docs/diagrams/c4?ref_type=heads
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5109/pull-cluster-api-provider-aws-e2e-eks/1884733722587041792

controllers/awscluster_controller.go
```diff
@@ -256,6 +267,14 @@ func (r *AWSClusterReconciler) reconcileDelete(ctx context.Context, clusterScope
 		allErrs = append(allErrs, errors.Wrap(err, "error deleting network"))
 	}
 
+	if err := iamService.DeleteOIDCProvider(ctx); err != nil {
```
richardcase, line 270: Does this need to be dependent on the feature flag?

controlplane/eks/api/v1beta1/awsmanagedcontrolplane_types.go
```diff
@@ -224,6 +224,8 @@ type EncryptionConfig struct {
 type OIDCProviderStatus struct {
 	// ARN holds the ARN of the provider
 	ARN string `json:"arn,omitempty"`
+	// IssuerURL holds the OIDC Issuer URL of the cluster.
+	IssuerURL string `json:"issuerUrl,omitempty"`
```
richardcase, line 228: This doesn't need adding to v1beta1. Any new fields should just be on the latest apiversion.

docs/book/src/topics/irsa-and-oidc-provider.md
```diff
@@ -0,0 +1,107 @@
...
+  name: capa-quickstart
+spec:
+  ...
+  associateOIDCProvider: true
+```
+
+
+### 4 - Set Service Account Issuer URL in KubeadmControlPlane Configuration
+
+The KubeadmControlPlane configuration should be updated to set the apiServer `service-account-issuer` argument with the S3 buckets URL:
```
richardcase, line 52: It might also be nice to create a new cluster flavor (i.e. template) in the **templates** directory.

controllers/awscluster_controller.go
```diff
@@ -352,6 +372,12 @@ func (r *AWSClusterReconciler) reconcileNormal(clusterScope *scope.ClusterScope)
 	}
 	conditions.MarkTrue(awsCluster, infrav1.S3BucketReadyCondition)
 
+	if err := iamService.ReconcileOIDCProvider(context.TODO()); err != nil {
```
richardcase, line 375: Should we wrap this with a test of the feature being enabled to be explicit? I know the validating webook shouldn't allow setting the associate field when the feature flag is disabled. Undecided about this if i'm honest ;)

pkg/cloud/services/iam/iam.go
```diff
@@ -0,0 +1,119 @@
...
+// ReconcileOIDCProvider replicates functionality already built into managed clusters by auto-deploying an oidc provider
+// that trusts the clusters Service Account issuer.
+// For more details see: https://github.com/aws/amazon-eks-pod-identity-webhook/blob/master/SELF_HOSTED_SETUP.md
+// 1. create an oidc provider in aws which points to the s3 bucket
+// 2. pause until kubeconfig and cluster access is ready
+// 3. build openid discovery config and upload to S3 bucket
+// 4. copy Service Account public signing JWKs to the s3 bucket.
+func (s *Service) ReconcileOIDCProvider(ctx context.Context) error {
+	if !s.scope.AssociateOIDCProvider() {
+		return nil
```
richardcase, line 68: Could be good to have a log entry to say its disabled at a high log verbosity level for debugging purposes.

controllers/awsmachine_controller.go
```diff
@@ -673,6 +673,50 @@ func (r *AWSMachineReconciler) reconcileOperationalState(ec2svc services.EC2Inte
 		return err
 	}
 
+	// check if the remote kubeconfig works and annotate the cluster
+	if _, ok := machineScope.InfraCluster.InfraCluster().GetAnnotations()[scope.KubeconfigReadyAnnotation]; !ok && machineScope.IsControlPlane() {
```
richardcase, line 677: I'm not sure about this as the mechanism to decide if we carry on with the OIDC reconciliation.

richardcase: Added a suggestion in the awscluster_controller.go

controllers/awscluster_controller.go
```diff
@@ -352,6 +372,12 @@ func (r *AWSClusterReconciler) reconcileNormal(clusterScope *scope.ClusterScope)
...
+		conditions.MarkFalse(awsCluster, infrav1.OIDCProviderReadyCondition, infrav1.OIDCProviderReconciliationFailedReason, clusterv1.ConditionSeverityError, "%s", err.Error())
+		clusterScope.Error(err, "failed to reconcile OIDC provider")
+		return reconcile.Result{RequeueAfter: 15 * time.Second}, nil
```
richardcase, line 378: You could move `if err := iamService.ReconcileOIDCProvider(context.TODO()); err != nil {` to after `awsCluster.Status.Ready = true` and then within `iamService.ReconcileOIDCProvider` instead return a ctrl.Result so that you can say RequeueAfter if the downstream  cluster isn't available yet. This would allow  provisioning to continue for the control plane and then when the cluster is accessible `iamService.ReconcileOIDCProvider` would do the second part and deploy the config map and return an empty ctrl.Result so there is no requeue.

The benefit is this removes the need for the AWSMachines controller annotating AWSCluster and keeps everything it the control of the AWSCluster controller.

Hopefully that all makes sense. If it doesn't ping me on slack and we can chat or arrange a call.

feature/feature.go
```diff
@@ -67,6 +67,9 @@ const (
 	// BootstrapFormatIgnition will allow an user to enable alternate machine bootstrap format, viz. Ignition.
 	BootstrapFormatIgnition featuregate.Feature = "BootstrapFormatIgnition"
 
+	// OIDCProviderSupport will allow a user to enable OIDC provider support for kubeadm clusters.
+	OIDCProviderSupport featuregate.Feature = "OIDCProviderSupport"
```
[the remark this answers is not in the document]
sl1pm4t: Good point, I'll update this.

api/v1beta1/s3bucket.go
```diff
@@ -37,12 +35,6 @@ func (b *S3Bucket) Validate() []*field.Error {
 		errs = append(errs, field.Required(field.NewPath("spec", "s3Bucket", "name"), "can't be empty"))
 	}
 
-	// Feature gate is not enabled but ignition is enabled then send a forbidden error.
-	if !feature.Gates.Enabled(feature.BootstrapFormatIgnition) {
```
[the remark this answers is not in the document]
sl1pm4t: I'm not sure, this PR was based on work started by @luthermonson so I don't have the full history.
I see the check still exists in the v1beta2 validation, so I'll revert this and do some tests.

sl1pm4t:
I', picking this back up today. Will hopefully be able to push changes to address the review comments this week.

richardcase:
Thats will be great @sl1pm4t , thank you. I was chatting with someone on Friday and they are really looking forward to this feature.

feature/feature.go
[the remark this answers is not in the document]
sl1pm4t: This feature was renamed to `OIDCProviderUnmanagedClusters`

cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
```diff
@@ -185,6 +185,11 @@ func (t Template) ControllersPolicy() *iamv1.PolicyDocument {
 				"ec2:DeleteLaunchTemplateVersions",
 				"ec2:DescribeKeyPairs",
 				"ec2:ModifyInstanceMetadataOptions",
+				"iam:CreateOpenIDConnectProvider",
```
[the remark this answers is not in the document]
sl1pm4t: This has been fixed in the latest commit.

controllers/awscluster_controller.go
sl1pm4t: I added a feature flag gate here, and in the `ReconcileNormal`

docs/book/src/topics/irsa-and-oidc-provider.md
sl1pm4t: A new cluster flavor has been added.

controllers/awscluster_controller.go
sl1pm4t: Done

pkg/cloud/services/iam/iam.go
sl1pm4t: I added:
```
log.V(4).Info("OIDC provider association is disabled, skipping reconciliation")
```

controllers/awscluster_controller.go
sl1pm4t: I implemented your suggestion, with a slight modification. 
In the IAM reconciler I change the way we build the `jwks` document, so that it can be done entirely locally. We already have the Service Account signgin key in the `<cluster>-sa` secret, so I added logic to build the JWKS keyset in code instead of reaching across to the workload cluster to retrieve it. I lifted code from the `kubernetes` codebase to calculate key hashes etc, so the result is identical to what ends up installed in the cluster. 

This means IdentityProvider and S3 objects can be fully provisioned before the workload cluster is available, so I removed the ready check + pause, and therefore didn't end up returning a `ctrl.Result`

I'm tempted to also remove the creation of the `boilerplate-oidc-trust-policy` ConfigMap on the remote cluster because I'm not really sure what value it provides?!

sl1pm4t:
@richardcase I think this is ready for re-review. I have rebased and I think I have addressed your review comments.

lukasmrtvy:
friendly ping

sl1pm4t:
Just a heads up that, at my company, we have decided to manage OIDC federation (IRSA) resources in a separate operator instead of within CAPA. 
Therefore, I'm probably unlikely to continue shepherding this PR any longer.

FilipNikolovski:
Really interested to see this cross the finish line, is there anything I could do to help push it along?

PR #5113, opened 2024-09-04
✨ Add support to specify dedicated host ID and Affinity

muraee:
**What type of PR is this?**

/kind feature
/kind api-change

This PR adds support to set placement settings for the instance when tenancy=host

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add support to specify dedicated host ID and Affinity
```


pkg/cloud/services/ec2/instances.go
```diff
@@ -653,7 +654,12 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 
 	if i.Tenancy != "" {
 		input.Placement = &ec2.Placement{
-			Tenancy: &i.Tenancy,
+			Tenancy: ptr.To(string(i.Tenancy)),
+		}
+
+		if i.Tenancy == infrav1.TenancyHost && i.DedicatedHostPlacement != nil {
+			input.Placement.Affinity = i.DedicatedHostPlacement.Affinity
```
nrb, line 661: These fields are labeled as optional; should we double check that they aren't nil as well? Or perhaps add a webhook to validate them before we get to this stage.

muraee: the AWS sdk accepts nil as input, so I don't think we need to validate them.

api/v1beta2/awsmachine_types.go
```diff
@@ -188,7 +200,11 @@ type AWSMachineSpec struct {
 	// Tenancy indicates if instance should run on shared or single-tenant hardware.
 	// +optional
 	// +kubebuilder:validation:Enum:=default;dedicated;host
-	Tenancy string `json:"tenancy,omitempty"`
+	Tenancy Tenancy `json:"tenancy,omitempty"`
+
+	// DedicatedHostPlacement denotes the placement settings for the instance when tenancy=host
+	// +optional
+	DedicatedHostPlacement *DedicatedHostPlacement `json:"dedicatedHostPlacment,omitempty"`
```
enxebre, line 207: I'd call this hostPlacement given the enum has both terms separately "Dedicated" and "Host" DedicatedHostPlacement seems confusing to me.

api/v1beta2/awsmachine_types.go
```diff
@@ -188,7 +200,11 @@ type AWSMachineSpec struct {
 	// Tenancy indicates if instance should run on shared or single-tenant hardware.
 	// +optional
 	// +kubebuilder:validation:Enum:=default;dedicated;host
-	Tenancy string `json:"tenancy,omitempty"`
+	Tenancy Tenancy `json:"tenancy,omitempty"`
```
enxebre, line 203: By reading this API is not clear to me which input should I set for tenancy host with autoplacement, can we document that here?

api/v1beta2/awsmachine_types.go
muraee: renamed.

api/v1beta2/awsmachine_types.go
enxebre: thanks! "If host tenancy is used and auto-placement is disabled for the Dedicated Host, It is required to set '.spec.hostPlacment.hostID'" though the requirement is actually not enforced. Can we document what happens if not set in that case, e.g. it's required or no host would be picked. Can we also document all matrix combinations this API enables: If both hostID and autoplacement are set, the hostID would take precedence? If the hostID doesn't exist would it fallback to autoplacement?

api/v1beta2/awsmachine_types.go
```diff
@@ -199,6 +216,19 @@ type AWSMachineSpec struct {
 	CapacityReservationID *string `json:"capacityReservationId,omitempty"`
 }
 
+// HostPlacement denotes the placement settings for the instance when tenancy=host.
+type HostPlacement struct {
+	// HostID pins the instance to a sepcific Dedicated Host.
+	// +optional
```
enxebre, line 222: isn't a hostID required if I set HostPlacement?

muraee: why? you might just want to set the affinity.

api/v1beta2/awsmachine_types.go
muraee: we can't enforce it, because we don't know if auto-placement is enabled or not on the dedicated Host, since its created directly by the user out of band.

muraee: updated.

api/v1beta2/awsmachine_types.go
```diff
@@ -186,9 +198,18 @@ type AWSMachineSpec struct {
...
+	// - if '.spec.hostPlacment.hostID' is set, the instance will be launched into the specified Host, disregarding any auto-placement settings.
+	// see https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-hosts-understanding.html#dedicated-hosts-auto-placement
+	//
 	// +optional
 	// +kubebuilder:validation:Enum:=default;dedicated;host
-	Tenancy string `json:"tenancy,omitempty"`
+	Tenancy Tenancy `json:"tenancy,omitempty"`
+
+	// HostPlacement denotes the placement settings for the instance when tenancy=host.
+	// +optional
```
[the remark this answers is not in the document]
muraee: how about just `Placement` then, so we can group all options inside it in the next api bump?

richardcase:
@muraee @enxebre - are there any outstanding questions on this?

PR #5114, opened 2024-09-04
🐛 Fix subnet sorting with multiple vpcs when launching an instance

cnmcavoy:
**What type of PR is this?**
/kind bug

https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4541 I introduced a change to allow subnets from other subnets to be returned from the AWS describe subnets query. However, the sorting is subtly bugged. `func(i, j int) bool {` is supposed to return true when i is _less_ than j, or should be ordered first. We want any subnets with the cluster VPC to be "less" than to sort the strings.

The existing implementation (which I wrote :disappointed: ) does not do that. `strings.Compare(*filtered[i].VpcId, clusterVPC)` returns `0` (string match) or `1` if the clusterVPC > VPCID, or `-1` if VPCID < clusterVIPC. The second comparison `strings.Compare(*filtered[j].VpcId, clusterVPC)` is the same. However, it does not make sense to compare the result of these two string comparisons to sort the strings, the goal is to ensure subnets with the cluster VPC is ordered earlier.

Fixes # N/A

The first commit updates only the unit tests to demonstrate the bug exists in the main branch. The second commit fixes the bug and makes the tests pass.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Fix subnet sorting with multiple vpcs when launching an instance
```

### DIFF
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -370,10 +370,10 @@ func (s *Service) findSubnet(scope *scope.MachineScope) (string, error) {
 
 			filtered = append(filtered, subnet)
 		}
-		// prefer a subnet in the cluster VPC if multiple match
+		// keep AWS returned orderz stable, but prefer a subnet in the cluster VPC
 		clusterVPC := s.scope.VPC().ID
 		sort.SliceStable(filtered, func(i, j int) bool {
-			return strings.Compare(*filtered[i].VpcId, clusterVPC) > strings.Compare(*filtered[j].VpcId, clusterVPC)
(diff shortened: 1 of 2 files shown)


richardcase:
Until the e2e pass:

/hold 

From my side:

/approve

damdo (commented):
/lgtm

damdo:
@richardcase are we happy to unhold?

richardcase:
As the e2e are passing, then sure why not 👍

damdo:
SGTM

PR #5140, opened 2024-10-08
:sparkles: Support setting maxHealthyPercentage to configure ASG instance refresh

fiunchinho:
**What type of PR is this?**

/kind feature

It allows to configure the `MaxHealthyPercentage`  setting of the ASGs instance refresh, to better control how instances are added and terminated.

This is adding a new field to the AWSMachinePool CRD, and it's not a breaking change.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Support setting maxHealthyPercentage to configure ASG instance refresh
```


exp/api/v1beta2/awsmachinepool_types.go
```diff
@@ -171,7 +171,20 @@ type RefreshPreferences struct {
 	// The amount of capacity as a percentage in ASG that must remain healthy
 	// during an instance refresh. The default is 90.
 	// +optional
+	// +kubebuilder:validation:Minimum=0
```
richardcase, line 174: Adding new validations  to an existing field may break some users.

Just to confirm, if a user is specifying 101 as the value for this prior to this change would AWS be returning an error? (i assume so but want to check).

exp/api/v1beta2/awsmachinepool_webhook.go
```diff
@@ -141,6 +142,26 @@ func (r *AWSMachinePool) validateSpotInstances() field.ErrorList {
...
 }
 
+func (r *AWSMachinePool) validateRefreshPreferences() field.ErrorList {
+	var allErrs field.ErrorList
+
+	if r.Spec.RefreshPreferences == nil {
+		return allErrs
+	}
+
+	if r.Spec.RefreshPreferences.MaxHealthyPercentage != nil && r.Spec.RefreshPreferences.MinHealthyPercentage == nil {
```
richardcase, line 152: We could also consider using CEL instead of new validation webhook checks.

However, we don't have any CEL at present so its fine to be consistent.

exp/api/v1beta2/awsmachinepool_types.go
fiunchinho: I haven't tested it, but I'd assume so, [according to AWS docs](https://docs.aws.amazon.com/cli/latest/reference/autoscaling/start-instance-refresh.html). I could leave this out of this PR if you prefer.

richardcase: It may be safer...unless we explicitly test it.

fiunchinho: I just removed it, can you take a look again?

richardcase:
Until the e2e pass:

/hold

But from my side:

/lgtm
/approve

richardcase:
The e2e is failing because the AMIs are no longer available 😢 We have 2 options:

1. Wait for the new AMIs to be published and then rebase this PR. This may take a few days to 1 week
2. Merge without the e2e passing.

Personally, i'd  lean towards 2). But would be good to get others view on this? @nrb @dlipovetsky @AndiDog any thoughts?

fiunchinho:
I don't think this PR will break the e2e tests since it's adding a new field. Having said that, waiting 1 week for this to get merged is fine by me.

nrb:
We have two other approvals, and I agree that we should not gate this on fixing the AMIs.

/lgtm
/approve
/unhold

PR #5147, opened 2024-10-09
✨ Cidr blocks for default NodePort services Ingress Rule

### LINKED ISSUE #3314: Security group rule "Node Port Services" can be more restrictive
When creating a new AWSCluster, part of the infrastructure is creating Security Groups for machines to use.

Currently, the `Node Port Services` security group allows access from any IP. Should we consider making this bit configurable to a set of pre-defined CIDR blocks, or allow the VPC CIDR's by default?

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/0eee2776780168fa4dd3f2edffdb9f3d614973a3/pkg/cloud/services/securitygroup/securitygroups.go#L525-L531

/area security
/kind bug
/assign @sedefsavas

krasoffski:
**What type of PR is this?**
/kind feature

There is issue #3314 which describes problem when default Security Group rule "Node Port Services" it too wide.

This PR allows to specify CIDR Blocks for Security Group rule "Node Port Services" with keeping original behavior if no CIDR Blocks are provided by attribute `AWSCluster.spec.network.NodePortIngressRuleCidrBlocks`.

In our practice, we've deployed stackgres and add LB for Admin page, but security team found opened ports with this admin page, cause stackgress by default open NodePort too under the hood.
Thus allowing to specify networks from which node port services can be accessed negotiated this problem. 

Fixes #3314 (possible fix).

I did not add similar attribute for IPv6 CIDR blocks because there is validation that `IPv6 cannot be used with unmanaged clusters at this time.`.
New unit tests for original and new behavior are added.
Added unit tests for webhook validation of correctness provided CidrBlocks.

**Checklist**:
- [x] squashed commits
- [x] includes documentation
- [x] adds unit tests

**Release note**:
```release-note
New `AWSCluster.spec.network.NodePortIngressRuleCidrBlocks` parameter  which allows to specify IPv4 CIDR blocks for Security group rule "Node Port Services" instead of default `0.0.0.0/`. 
```


api/v1beta2/awscluster_webhook.go
```diff
@@ -267,6 +268,12 @@ func (r *AWSCluster) validateNetwork() field.ErrorList {
 		allErrs = append(allErrs, r.validateIngressRule(rule)...)
 	}
 
+	for _, cidrBlock := range r.Spec.NetworkSpec.NodePortIngressRuleCidrBlocks {
+		if _, _, err := net.ParseCIDR(cidrBlock); err != nil {
+			allErrs = append(allErrs, field.Invalid(field.NewPath("nodePortIngressRuleCidrBlocks"), r.Spec.NetworkSpec.NodePortIngressRuleCidrBlocks, "CIDR block is invalid"))
```
AndiDog, line 273:
```suggestion
	for cidrBlockIndex, cidrBlock := range r.Spec.NetworkSpec.NodePortIngressRuleCidrBlocks {
		if _, _, err := net.ParseCIDR(cidrBlock); err != nil {
			allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "network", fmt.Sprintf("nodePortIngressRuleCidrBlocks[%d]", cidrBlockIndex)), r.Spec.NetworkSpec.NodePortIngressRuleCidrBlocks, "CIDR block is invalid"))
```

(the surrounding code examples use incomplete paths, unfortunately)

api/v1beta2/network_types.go
```diff
@@ -351,6 +351,10 @@ type NetworkSpec struct {
 	// AdditionalControlPlaneIngressRules is an optional set of ingress rules to add to the control plane
 	// +optional
 	AdditionalControlPlaneIngressRules []IngressRule `json:"additionalControlPlaneIngressRules,omitempty"`
+
+	// NodePortIngressRuleCidrBlocks is an optional set of CIDR blocks to allow traffic to nodes NodePort services.
```
AndiDog, line 355:
```suggestion
	// NodePortIngressRuleCidrBlocks is an optional set of CIDR blocks to allow traffic to nodes' NodePort services.
	// If none are specified here, all IPs are allowed to connect.
```

pkg/cloud/scope/sg.go
```diff
@@ -59,4 +59,7 @@ type SGScope interface {
 	// ControlPlaneLoadBalancers returns both the ControlPlaneLoadBalancer and SecondaryControlPlaneLoadBalancer AWSLoadBalancerSpecs.
 	// The control plane load balancers should always be returned in the above order.
 	ControlPlaneLoadBalancers() []*infrav1.AWSLoadBalancerSpec
+
+	//
```
AndiDog, line 63:
```suggestion
	// NodePortIngressRuleCidrBlocks returns the CIDR blocks for the node NodePort ingress rules.
```

pkg/cloud/services/securitygroup/securitygroups.go
```diff
@@ -645,6 +644,10 @@ func (s *Service) getSecurityGroupIngressRules(role infrav1.SecurityGroupRole) (
 		return append(cniRules, rules...), nil
 
 	case infrav1.SecurityGroupNode:
+		cidrBlocks := []string{services.AnyIPv4CidrBlock}
```
AndiDog, line 647: For the record: if someone uses an IPv6-enabled VPC, the result will be surprising (see some lines below where we allow all incoming traffic from IPv6 sources). Okay to leave that part out for now, since there's not much contribution interest in IPv6, nor is it used widely anywhere.

Side note (personal opinion): we should not distinguish IPv4/IPv6 with separate field names. Instead, allow parsing `NodePortIngressRuleCidrBlocks []string` as either. If it contains an IPv6 CIDR, deal with it.

AndiDog (commented):
Mostly looks good. I have some easy comments.

pkg/cloud/services/securitygroup/securitygroups.go
krasoffski: I agree about second point. But it looks like for me that `[]string` is not suite for this task. It will require parsing strings during provisioning. But not sure that usage of like `[]CidrBlock` where CidrBlock is interface will be better.
Thank you!

AndiDog:
The apidiff failure is informational, and in this case acceptable. I'd say let's either get this merged, or we wait until the E2E tests work again (there's a problem right now about AMIs).

/lgtm

krasoffski:
@AndiDog 
It looks like I cannot merge right now as there is no such option in this PR (no merge button for me). 
Probably till all checks are passed.

krasoffski:
@AndiDog
Hi, again.
Is there any way how to merge this PR with apidiff fail?

richardcase:
The incompatible apidiffs are from the exported go api and not the k8s api, as we don't provide guarantees over the exported go packages:

/override pull-cluster-api-provider-aws-apidiff-main

richardcase:
Thanks for this @krasoffski 🙇‍♂️ 

/approve

PR #4700, opened 2023-12-18
✨ Ignition: add option to store User Data in plain text

damdo:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

For Machine bootstrapping CAPA supports one main method: CloudInit and one experimental method: Ignition.

For the former, CAPA used to store the User Data that the bootstrapper would use to fulfill its job in the EC2 Instance User Data in plain text.

This was not considered the most secure way to provide this configuration to a machine as a process/subject with the power to query the instance metadata could be able to read this config, which may contain certs, keys and other forms of secrets.

As such to minimize this issue, a new default method was added, where the controller would now leverage AWS SSM to encrypt the user data before storing it in the EC2 Instance User Data. The previous method of storing User Data in plain text was kept available as an alternative option to users, which would be able to access it by providing the `InsecureSkipSecretsManager` option in the AWSMachine spec.

Later on, when the new Ignition bootstrapping support was added, the creators, already aware of the security implications of storing User Data in plain text,
decided to opt for a more secure default. After briefly looking at using AWS SSM encryption also for Ignition, they acknowledged that it wasn't possible due to Ignition's lack of support for multi mime types. Hence they went for using S3 Buckets to store the config in objects, and have ignition fetch it at bootstrap time from there, and this did the trick.

Up until now this has been the only option to achieve Ignition bootstrapping.
However even if this is a better option from a security standpoint than storing User Data in plain text, and as such the preferred option for a good default,
it requires the ability to provision/access S3 Buckets, which is not always possible/desirable in specific scenarios, as well as not suitable for certain type of bespoke CAPI bootstrap providers.

As such we would like to introduce for Ignition, similarly to what is available for CloudInit, the ability to store User Data in the EC2 Instance in plain text.
Of course we know this is not the best option from a security standpoint, and as such we want to make it a fallback option that users can leverage if their use case does not allow the use of S3 buckets. We mentioned in the docs, and in the API that this option is discouraged and we trust CAPA users will use this with care.

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [x] adds unit tests

**Release note**:
```release-note
Ignition: add option to store User Data in plain text
```

PR #5148, opened 2024-10-10
⚠️ Support storing Ignition user data in S3 bucket for `AWSMachinePool`

AndiDog:

EC2 user data is insecure and very limited in size (16 KB). Use the S3 bucket storage also for machine pools. The difference to `AWSMachine` is that the S3 objects must stay around because a new instance could be started by the ASG at any time, and user data with a valid token must exist. Both a lifecycle policy for regular deletion, and explicit S3 object deletion requests, are used for cleanup, so that normally, only 1-3 launch template versions' user data are stored in S3. Each launch template version relates to one S3 object.

**What type of PR is this?**

I marked this as breaking change for now, since the default storage – in the case of `AWSMachinePool` + Ignition user data format – changes to `IgnitionStorageTypeOptionClusterObjectStore` if nothing is explicitly specified. For me, this makes sense because the `AWSMachine` controller already defaulted to that storage type. It's very likely that an administrator who uses Ignition user data has also configured an S3 bucket, so I see low chances of breaking someone's workflow. If not, the reconciler will log an error and not continue, and the administrator could set `AWSMachinePool.spec.ignition.storageType` to `UnencryptedUserData` in order to keep machine pool related user data stored in plaintext on each EC2 instance.

/kind feature

Didn't find an existing issue. My company ran into EC2 user data size limit problems.

This change has been successfully used in the Giant Swarm fork of CAPA since July, on production clusters. I have adapted the change to the latest upstream changes, in particular the introduction of `IgnitionStorageTypeOption` (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4700). Therefore, I retested this feature using Tilt on a plain CAPI@main + CAPA@main setup, with AWSCluster + Flatcar AMI + AWSMachinePool in my manifests.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
Support storing Ignition user data in S3 bucket for `AWSMachinePool`
```


AndiDog:
Superseded by another branch => https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5172

PR #4917, opened 2024-04-10
🐛 fix: update network subnets prio tag failures

mtulio:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

The subnet information from unmanaged subnets, like the Availability Zone name discovered when reconciling subnets, is not updated for additional subnets when the are failures to tag subnets for the preceding subnets.

For example, given subnetId1 and subnetId2, if the subnet tag failed when setting tags on subnetId1, the subnetId2 will not be discovered and attributes not be correctly set.

This change enforces subnet updates before trying to apply tags and does not skip the loop when checking existing subnets.

Additionally, a new option was added to the unit for reconcileSubnets(), allowing to ensure expected SubnetSpec attributes from Subnets{} post reconciliation.

**Which issue(s) this PR fixes**

References:
- #4874 
- #4901 
- #4882 

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation (N/A)
- [x] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [x] adds unit tests
- [ ] adds or updates e2e tests (N/A)

**Release note**:

```release-note
NONE
```

PR #5133, opened 2024-10-04
✨refactor: update to use new community ami account

richardcase:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Updates to the project to start using the new CNCF owned AWS account for publishing AMIs. The code has had the old VMWare account removed and the new account ID added. Additionally the docs have been updated including clarifying that the public AMIs should only be used for non-production usage.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Relates #5131
Relates #4984 
Fixes #5161 

**Special notes for your reviewer**:

A new GitHub Actions workflow (see #5138 ) can be invoked to publish new images. Also #5169 will be done as a follow-up.

**Checklist**:

- [X] squashed commits
- [X] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
action required
A new community owned AWS account is now being used for publishing AMIs. The old account owner by a company is no longer accessible and the images are being automatically deleted by AWS. Support for ubuntu 18.04 & 20.04 has been dropped and ubuntu 24.04 is now the default OS if no OS is supplied.
```

PR #5175, opened 2024-10-22
🐛: Tags defined in subnet spec should be applied

fiunchinho:
**What type of PR is this?**

/kind bug

Since 2.5.0 release, we can't tag subnets anymore by adding tags to the subnet spec in the `AWSCluster` CR. I believe the bug got introduced [on this PR](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4917/files#r1810976840), by overwriting the subnet tags defined in the spec by the tags from the AWS Subnet. This means that the existing tags on the AWS Subnet will always be applied, while the tags defined in the subnet spec will be ignored.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
Tags defined in subnet spec should be applied
```

### DIFF
--- a/pkg/cloud/services/network/subnets.go
+++ b/pkg/cloud/services/network/subnets.go
@@ -58,14 +58,6 @@ func (s *Service) reconcileSubnets() error {
 		existing infrav1.Subnets
 	)
 
-	// Describing the VPC Subnets tags the resources.
-	if s.scope.TagUnmanagedNetworkResources() {
-		// Describe subnets in the vpc.
-		if existing, err = s.describeVpcSubnets(); err != nil {
-			return err
-		}
-	}
-
 	unmanagedVPC := s.scope.VPC().IsUnmanaged(s.scope.Name())
 
 	if len(subnets) == 0 {
@@ -93,12 +85,9 @@ func (s *Service) reconcileSubnets() error {
 		}
 	}
 
-	// Describing the VPC Subnets tags the resources.
-	if !s.scope.TagUnmanagedNetworkResources() {
-		// Describe subnets in the vpc.
-		if existing, err = s.describeVpcSubnets(); err != nil {
-			return err
-		}
+	// Describe subnets in the vpc.
+	if existing, err = s.describeVpcSubnets(); err != nil {
+		return err
 	}
 
 	if s.scope.SecondaryCidrBlock() != nil {
@@ -139,11 +128,12 @@ func (s *Service) reconcileSubnets() error {
 				existingSubnet.ID = sub.ID
 			}
 
+			// Make sure tags are up-to-date.
+			subnetTags := sub.Tags
+
 			// Update subnet spec with the existing subnet details
 			existingSubnet.DeepCopyInto(sub)
 
-			// Make sure tags are up-to-date.
-			subnetTags := sub.Tags
 			if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
 				buildParams := s.getSubnetTagParams(unmanagedVPC, existingSubnet.GetResourceID(), existingSubnet.IsPublic, existingSubnet.AvailabilityZone, subnetTags, existingSubnet.IsEdge())
 				tagsBuilder := tags.New(&buildParams, tags.WithEC2(s.EC2Client))
--- a/pkg/cloud/services/network/subnets_test.go
+++ b/pkg/cloud/services/network/subnets_test.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"reflect"
+	"slices"
 	"testing"
 
 	"github.com/aws/aws-sdk-go/aws"
@@ -63,10 +64,7 @@ func TestReconcileSubnets(t *testing.T) {
 		{ID: "subnet-private-us-east-1-wl1-nyc-wlz-1", AvailabilityZone: "us-east-1-wl1-nyc-wlz-1", CidrBlock: "10.0.7.0/24", IsPublic: false},
 		{ID: "subnet-public-us-east-1-wl1-nyc-wlz-1", AvailabilityZone: "us-east-1-wl1-nyc-wlz-1", CidrBlock: "10.0.8.0/24", IsPublic: true},
 	}
-	// TODO(mtulio): replace by slices.Concat(...) on go 1.22+
-	stubSubnetsAllZones := stubSubnetsAvailabilityZone
-	stubSubnetsAllZones = append(stubSubnetsAllZones, stubSubnetsLocalZone...)
-	stubSubnetsAllZones = append(stubSubnetsAllZones, stubSubnetsWavelengthZone...)
+	stubSubnetsAllZones := slices.Concat(stubSubnetsAvailabilityZone, stubSubnetsLocalZone, stubSubnetsWavelengthZone)
 
 	// NetworkSpec with subnets in zone type availability-zone
 	stubNetworkSpecWithSubnets := &infrav1.NetworkSpec{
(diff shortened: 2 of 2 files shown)


pkg/cloud/services/network/subnets.go
```diff
@@ -93,12 +85,9 @@ func (s *Service) reconcileSubnets() error {
 		}
 	}
 
-	// Describing the VPC Subnets tags the resources.
-	if !s.scope.TagUnmanagedNetworkResources() {
-		// Describe subnets in the vpc.
-		if existing, err = s.describeVpcSubnets(); err != nil {
-			return err
-		}
+	// Describe subnets in the vpc.
```
fiunchinho, line 88: We had two blocks to describe the subnets in the VPC. One was executed when the feature flag for tagging unmanaged network resources was enabled, and the other block was executed when the feature flag was disabled. So it was always executed anyway.

pkg/cloud/services/network/subnets.go
```diff
@@ -139,11 +128,12 @@ func (s *Service) reconcileSubnets() error {
 				existingSubnet.ID = sub.ID
 			}
 
+			// Make sure tags are up-to-date.
+			subnetTags := sub.Tags
+
```
fiunchinho, line 133: This is the key change that fixes the issue. By storing the subnet tags from the spec to a local variable, we won't overwrite them when using `DeepCopyInto`.

damdo:
@fiunchinho FYI E2E tests are broken until #5133 merges

fiunchinho:
> @fiunchinho FYI E2E tests are broken until #5133 merges

Ok, thanks for the heads up.

BTW, since this is fixing a bug, shouldn't it be released as patch for v2.6 instead of v2.7?

damdo:
We should definitely release it in both.
So we'll merge it for 2.7 and then do a backport for the next 2.6.z.
/cc @richardcase

pkg/cloud/services/network/subnets_test.go
```diff
@@ -2625,6 +2575,183 @@ func TestReconcileSubnets(t *testing.T) {
 					}, nil).AnyTimes()
 			},
 		},
+		{
+			name: "Managed VPC, existing public and private subnets, 2 subnets in spec, custom tags in spec should be created",
```
mtulio, line 2579: is "Managed VPC" covers "existing subnets"?

mtulio: In BYO infra / [Configuring the AWSCluster Specification](https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure#configuring-the-awscluster-specification) it states:
> Specifying existing infrastructure for Cluster API to use takes place in the specification for the AWSCluster object. Specifically, you will need to add an entry with the VPC ID and the IDs of all applicable subnets into the network field.

So, should it be "unmanaged VPC" when you specify the VPC ID and subnet IDs?

pkg/cloud/services/network/subnets.go
mtulio: Hi @fiunchinho - is this behavior considering managed and unmanaged scenarios?

Checking the documentation about BYO/unmanaged and tags, it [states][1]:
> Note: All the tagging of resources should be the responsibility of the users and are not managed by CAPA controllers.

My understanding is this block will be reached at once when *unmanaged* (`existingSubnets!=nil`), and later interactions in the reconcile loop when *managed* (with subnets created h[ere ][2], preserving subnets tags defined on spec), considering the documentation that *unmanaged* should not create additional tags, than required from k8s, the tags from spec should be skipped. Is it make sense? (+ @richardcase @nrb  would love to hear your thoughts too)

[1]: https://cluster-api-aws.sigs.k8s.io/topics/bring-your-own-aws-infrastructure#tagging-aws-resources
[2]: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/c8eb4797b15e4925c07b05782b90add67f59d148/pkg/cloud/services/network/subnets.go#L197-L201

fiunchinho: hey @mtulio thanks for the comment. If I understand it correctly, with this approach, when reconciling an unmanaged vpc, the tags in the subnet on AWS are applied to the AWSCluster spec, and new tags in the AWSCluster spec are ignored/never applied into AWS. When reconciling a managed VPC, the tags in the AWSCluster spec are applied to the subnet on AWS.

[In the subnets.go file, line 626](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/subnets.go#L664), we only take into consideration the tags defined in the `AWSCluster` spec (called `manualTags` in the function) if the vpc is managed.

So I think this approach is not going against current documentation of BYOI/unmanaged infra.

Also, I think finding the subnet in AWS (`existingSubnets!=nil`) doesn't mean it's unmanaged. When CAPA creates the subnets, on following reconciliations it will find the same subnets on AWS, and they are managed by CAPA. A managed VPC is defined as a VPC with an `id` set, and the `sigs.k8s.io/cluster-api-provider-aws/cluster/$clusterName` set.

pkg/cloud/services/network/subnets_test.go
fiunchinho: [I tried explaining here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5175/files#r1815091234). Let me know your thoughts, please!

pkg/cloud/services/network/subnets.go
mtulio: > hey @mtulio thanks for the comment. If I understand it correctly, with this approach, when reconciling an unmanaged vpc, the tags in the subnet on AWS are applied to the AWSCluster spec, and new tags in the AWSCluster spec are ignored/never applied into AWS. [...]

Yep, that's the current behavior, isn't it aligned with the documentation for BYOI/unmanaged infra?

The [change](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4917/files#r1810976840) wasn't intentionally added, but revisiting the docs with your findings I am a little bit confused what would be the expected behavior for unmanaged on CAPA. @richardcase @nrb thoughts?

Just sharing some thoughts how we are using: we, on OpenShift, consider tags are managed by the user when they opt to provide their own network/subnets, and we don't touch on it when installing a cluster. So this change won't affect our automation as we are not setting the subnet tags for unmanaged VPC.

richardcase: For unmanaged, I'd say:

- Only apply required k8s tags to the subnets in AWS
- No other custom tags to be added to the subnets in AWS

As for syncing existing tags on the subnets back to AWSCluster, I'm not strongly opinionated on this.

Every time i look at the subnets code, it makes me 😢 IMHO, it's historically one of the most error-prone and horrible bits of code in CAPA.

AndiDog: `getSubnetTagParams` still does the right thing for unmanaged VPCs (= don't use additional tags unless `TagUnmanagedNetworkResources()`), so I think the behavior is fine even with this PR. If tags are directly defined on `AWSCluster.spec.network.subnets.tags`, it's a conscious decision by the admin and those tags should be added even for unmanaged ones – why else have a field for tags?

The field description could be more precise, though (out of scope for the PR):

```
$ kubectl explain awscluster.spec.network.subnets.tags
GROUP:      infrastructure.cluster.x-k8s.io
KIND:       AWSCluster
VERSION:    v1beta2

FIELD: tags <map[string]string>


DESCRIPTION:
    Tags is a collection of tags describing the resource.
```

AndiDog (commented):
/lgtm

(IMO, see comment thread about tagging behavior)

damdo:
/assign @nrb 
for approval

fiunchinho:
If there is anything preventing this PR from being approved just let me know.

richardcase:
@fiunchinho - would you be able to rebase this? That way we can pick up the latests e2e changes and run the eks e2e before merging.

I think once the e2e pass (after rebasing) we could look to get this merged.

@AndiDog - wdyt?

fiunchinho:
I just rebased, please take a look. Thank you.

PR #5181, opened 2024-10-24
🐛 Filter out AWS internal tags when reconciling

### LINKED ISSUE #5120: Feature Flag TagUnmanagedNetworkResources does not work if network resources have internal aws tags
/kind bug

We want to use already existing subnets. But we want to enable the `TagUnmanagedNetworkResources` feature flag, so that we can manage the subnet tags. But the subnets were created using CloudFormation, so they contain some tags of the form `aws::cloudformation::*`. When trying to add or delete tags of that form, the AWS API returns an error, because those are reserved/internal AWS tags.
```
InvalidParameterValue: Value ( aws:cloudformation:stack-name ) for parameter key is invalid. Tag keys starting with 'aws:' are reserved for internal use
```

When enabling the TagUnmanagedNetworkResources feature flag, VPC and subnets are reconciled without errors, and tags are applied normally.

Would it be possible to change CAPA code so that it ignores tags that start with `aws:`?

**Environment:**

- Cluster-api-provider-aws version: 2.5.2
- Kubernetes version: (use `kubectl version`): 1.28

fiunchinho:
**What type of PR is this?**

/kind bug

Tags that start with `aws:` are considered internal AWS tags and can't be add or removed.

Fixes #5120

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Filter out AWS internal tags when reconciling AWS infra
```

### DIFF
--- a/pkg/cloud/tags/tags.go
+++ b/pkg/cloud/tags/tags.go
@@ -21,6 +21,7 @@ import (
 	"context"
 	"fmt"
 	"sort"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/service/ec2"
@@ -30,6 +31,10 @@ import (
 	"github.com/pkg/errors"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+)
+
+const (
+	AwsInternalTagPrefix = "aws:" // AwsInternalTagPrefix is the prefix for AWS internal tags, which are reserved for internal AWS use.
 )
 
 var (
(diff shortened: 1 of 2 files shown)


pkg/cloud/tags/tags_test.go
```diff
@@ -34,14 +34,7 @@ import (
 )
 
 var (
-	bp = infrav1.BuildParams{
```
fiunchinho, line 37: I inlined the `BuildParams` object on each test case to make it easier to see what's being tested.

pkg/cloud/tags/tags.go
```diff
@@ -127,10 +131,12 @@ func WithEKS(eksclient eksiface.EKSAPI) BuilderOption {
 	return func(b *Builder) {
 		b.applyFunc = func(params *infrav1.BuildParams) error {
 			tags := infrav1.Build(*params)
-
 			eksTags := make(map[string]*string, len(tags))
 			for k, v := range tags {
-				eksTags[k] = aws.String(v)
+				// We want to filter out the tag keys that start with `aws:` as they are reserved for internal AWS use.
+				if !strings.HasPrefix(k, "aws:") {
```
s3rj1k, line 137: Maybe make `aws:` as const ? (Will help readability)

fiunchinho: Nice suggestion! I've introduced a constant, please take another look.

s3rj1k: Yep LGTM, thanks for taking care of this.

PR #5198, opened 2024-11-01
🐛 fix(securitygroups): Look up unmanaged NAT Gateway IPs, so provider doesn't add 0.0.0.0/0 SG rule

### LINKED ISSUE #5196: The provider adds a 0.0.0.0/0 SG rule to Control Plane LB in unmanaged mode.
/kind bug

* Create a cluster in an unmanaged VPC mode (i.e. the VPC is created by some other tool).
* Review the Control Plane load balancer security group rules and see the provider has added a rule allowing all IPv4 addresses (`0.0.0.0/0`). 

* I would expect the LB security group rules to only include the external IPs of the pre-existing NAT GW, and any specified in the AWSCluster `spec.controlPlaneLoadBalancer.ingressRules` field.

In my test, the provider was configured to create a public NLB for the apiserver, and I had specified additional allowed sources addresses via the AWSCluster `spec.controlPlaneLoadBalancer.ingressRules` field

I did some digging through the code, and CAPA adds the `0.0.0.0/0` rule if there are no NAT GW IPs available:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/securitygroup/securitygroups.go#L951-L952

 The [NAT GW IPs are set](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/natgateways.go#L105) on the `AWSCluster` status in `reconcileNatGateways()` but, this short circuits in unmanaged mode, and doesn't attempt to detect the IPs associated with the pre-existing NAT Gateways:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/network/natgateways.go#L41-L44

**Environment:**

- Cluster-api-provider-aws version: `v2.6.1`
- Kubernetes version: (use `kubectl version`): `v1.29.8`

sl1pm4t:
**What type of PR is this?**

/kind bug

This fixes an issue that caused the provider to add a security group rule allowing all addresses (`0.0.0.0/0`) access to the workload cluster API Server, if a cluster was created on an unmanaged VPC. The rule would be added even if the user had provided a list of source address in `spec.controlPlaneLoadBalancer.ingressRules`.

The rule was being added so the kubelets could reach the API, but `0.0.0.0/0` was a fall back source if the NAT Gateway IPs were not (yet) known. Prior to this PR, In the case of unmanaged VPC, the NAT Gateways IPs were never retrieved so `getIngressRulesToAllowKubeletToAccessTheControlPlaneLB()` in [securitygroups.go](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/securitygroup/securitygroups.go#L929) would always fall back to using `0.0.0.0/0`.

Fixes #5196

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Action Required: If deploying clusters to an existing VPC (not managed by the AWS provider), the provider will no longer automatically create a security group rule allowing traffic from all addresses (`0.0.0.0/0`). You may need to update `AWSCluster.spec.controlPlaneLoadBalancer.ingressRules` with the source address of your Management Cluster.
```

### DIFF
--- a/pkg/cloud/services/network/natgateways.go
+++ b/pkg/cloud/services/network/natgateways.go
@@ -41,6 +41,10 @@ import (
 func (s *Service) reconcileNatGateways() error {
 	if s.scope.VPC().IsUnmanaged(s.scope.Name()) {
 		s.scope.Trace("Skipping NAT gateway reconcile in unmanaged mode")
+		_, err := s.updateNatGatewayIPs(s.scope.TagUnmanagedNetworkResources())
+		if err != nil {
+			return err
+		}
 		return nil
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/cloud/services/network/natgateways.go
```diff
@@ -41,6 +41,10 @@ import (
 func (s *Service) reconcileNatGateways() error {
 	if s.scope.VPC().IsUnmanaged(s.scope.Name()) {
 		s.scope.Trace("Skipping NAT gateway reconcile in unmanaged mode")
+		_, err := s.updateNatGatewayIPs(false)
```
AndiDog, line 44:
```suggestion
		_, err := s.updateNatGatewayIPs(s.scope.TagUnmanagedNetworkResources())
```

AndiDog (commented):
Change looks mostly fine. Did you manually test this? I'm wondering why tests are failing.

sl1pm4t:
> Change looks mostly fine. Did you manually test this? I'm wondering why tests are failing.

Yes I have tested the change, and we are using regularly in our fork. 

I have not attempted to run the e2e test suite locally though, as I'm not familiar with them and not sure how to run against our own infrastructure.

sl1pm4t:
@AndiDog I will add, the tests seem to fail on something different each time, so I'm not sure it's related to my change.

joejulian:
This is a great change! Thank you. I needed this, too.

nrb:
EKS failure appears to be an account issue unrelated to our code.

Standard e2e tests were hitting #5252, but also `[It] [unmanaged] [functional] Workload cluster with EFS driver should pass dynamic provisioning test`. I don't think should be affected by NAT gateways, so that appears to be a flake.

Looking at the PR history, I also saw one occurrence of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5242.

Kicking off the e2es again to double check the flake theory, but I think this change looks alright.

/lgtm

nrb:
Dunno why this didn't work last time, but last run was last week.

nrb:
CI's fixed, but I'm not going to block this any longer.

/approve

Issue #5139, opened 2024-10-07
ELB health check fails with Kubernetes >=v1.30.x


dkoshkin:
/kind bug

**What steps did you take and what happened:**
Follow the [quickstart documentation ](https://cluster-api.sigs.k8s.io/user/quick-start.html) with Kubernetes v1.30.5 and a custom built AMI (the public AMIs are missing for that version and the default v1.31.0 version).
The ELB Health Check fails and the cluster is stuck after creating the first control-plane instance. The AWS console shows that 0 of 1 instanced are in service.

* The CAPA API defaults [create a Classic ELB](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1245ff7a81a552a2767d21ed1128678656c24042/api/v1beta2/defaults.go#L70) with an [SSL health check target](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/1245ff7a81a552a2767d21ed1128678656c24042/pkg/cloud/services/elb/loadbalancer.go#L1724). (HTTPS also doesn't work, but TCP does) 
* Starting in Go 1.22 the RSA ciphers were removed - https://github.com/golang/go/issues/63413
* Kubernetes >v1.30 switched to Go 1.22 in recent releases https://github.com/kubernetes/kubernetes/commit/ddb0b8da9dbcd3d507001c5cf8a01708816f3cd5

**What did you expect to happen:**
The defaults should result in a working cluster.

**Anything else you would like to add:**
1. Changing the health check to TCP in the AWS console did fix the check, but this update is not allowed by a webhook here and even after removing the webhook, the new value from AWSCluster never got updated.

2. Setting this on the apiserver and other control-plane components allowed the ELB health check to pass
```
tls-cipher-suites: ...,TLS_RSA_WITH_AES_128_GCM_SHA256,TLS_RSA_WITH_AES_256_GCM_SHA384,TLS_RSA_WITH_AES_128_CBC_SHA,TLS_RSA_WITH_AES_256_CBC_SHA,TLS_RSA_WITH_3DES_EDE_CBC_SHA
```
3. Using an NLB loadbalancer works
```
  controlPlaneLoadBalancer:
    loadBalancerType: nlb
```

Some discussion about this in the Kuberentes slack https://kubernetes.slack.com/archives/C3QUFP0QM/p1726622974749509

**Environment:**

- Cluster-api-provider-aws version: 2.6.1
[...]

PR #5209, opened 2024-11-06
🌱 Bump CAPI to v1.9.z

damdo:
**What type of PR is this?**
/kind support

CAPI v1.9.z is out now, so this PRs bumps the CAPI version used by CAPA with the new released one.

Bumps CAPI to v1.9.x based on https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/book/src/developer/providers/migrations/v1.8-to-v1.9.md

```release-note
action required
Bump CAPI to v1.9.z
!! ACTION REQUIRED BEFORE UPGRADING !!
If you are using the AWSManagedControlPlane to provision EKS clusters and you do not have a spec.Version specified in such resource (meaning you are relying on the default that AWS provides), you will need to either:
a) explicitly set such `spec.Version` field before upgrading CAPA
or b) disable the MachineSetPreflightChecks in your cluster either:
b1) by setting this core CAPI feature gate to `false`
b2) or by disabling it via the relevant annotation on all the machineSets belonging to said cluster (follow this guide on how to do this: https://cluster-api.sigs.k8s.io/tasks/experimental-features/machineset-preflight-checks).
This is necessary as core CAPI 1.9 introduces a feature gate change, setting MachineSetPreflightChecks=true, which in turn relies on the presence of spec.Version and status.Version on the AWSManagedControlPlane object.
We are planning a future refactor of these API fields in v1beta3 (https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3853).
Other places where you can find details on this are:
- https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5225
- https://github.com/kubernetes-sigs/cluster-api/issues/11117
- https://kubernetes.slack.com/archives/CD6U2V71N/p1739783013734149
```

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5335


damdo:
We are hitting https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5139 with this, as we are now using the 1.30+ k8s version in testing.

damdo:
/hold

These are still not passing, even after I rebased on top of the latest e2e fix #5338 

cc. @richardcase

faiq:
I'm seeing a lot of errors regarding missing AMIs 

```
E0211 12:37:39.777314       1 awsmachine_controller.go:528] "unable to create instance" err="failed to create AWSMachine instance: failed to find ami: found no AMIs with the name: \"capa-ami-ubuntu-24.04-?1.31.0-*\""
```

and also

```
I0211 12:38:00.504024       1 recorder.go:104] "Failed to find ami for OS=flatcar-stable, Architecture=x86_64 and Kubernetes-version=v1.31.0: found no AMIs with the name: \"capa-ami-flatcar-stable-?1.31.0-*\"" logger="events" type="Normal" object={"kind":"AWSCluster","namespace":"functional-test-ignition-qfhdxc","name":"functional-test-ignition-q4zmt4","uid":"ffd1ab2d-fe0f-4237-b72a-33c9b053466e","apiVersion":"infrastructure.cluster.x-k8s.io/v1beta2","resourceVersion":"3572"} reason="FailedDescribeImages"
```

damdo:
For extra context more info here: https://kubernetes.slack.com/archives/CD6U2V71N/p1739291982497849

damdo:
Unrelated issue

damdo:
ClusterClass also needed to switch to NLBs, added a new commit for that.

/test pull-cluster-api-provider-aws-e2e-eks pull-cluster-api-provider-aws-e2e

pkg/cloud/services/eks/cluster.go
```diff
@@ -141,6 +141,10 @@ func (s *Service) reconcileCluster(ctx context.Context) error {
 }
 
 func (s *Service) setStatus(cluster *eks.Cluster) error {
+	// Set the current Kubernetes control plane version in the status.
+	s.scope.ControlPlane.Status.Version = cluster.Version
```
richardcase, line 145: We may need to do some manipulation here. AWS will return something like `1.31` and not a full semver.

The Version field in the spec is in the format of vX.Y.Z or vX.Y

So if AWS returns `1.31` we should put `v1.31` or `v1.31.0 into the field

damdo: Hm. That's a bit tricky then.
What if the user had `v1.31.4` and then specifies a `v1.31.5`, to upgrade, for example?
How do we know if that was successfully rolled out if we can only read `1.31`?

damdo: I had a deeper look at this. From what I have been able to gather AWS EKS directly manages patch versions and so the user should only specify `major.minor` in the `Version` field of the SDK/Console, even though that's not explicitly specified in the API godoc (https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/eks#CreateClusterInput).

It seems though that if you provide a version with a patch (for example, 1.30.2), EKS will ignore the patch portion and effectively use the major.minor version (1.30). It does not reject the request; it silently truncates the patch version.

This behavior occurs because EKS manages Kubernetes patch updates automatically and only supports specifying the version in a major.minor format.

damdo: I also had a look at the [controlPlaneStablePreflightCheck](https://github.com/kubernetes-sigs/cluster-api/blob/3abcd216748c1dcae6828afb317099526e237ae9/internal/controllers/machineset/machineset_preflight.go#L94) where the strictest check on spec/status.Version is the [IsUpgrading](https://github.com/kubernetes-sigs/cluster-api/blob/3abcd216748c1dcae6828afb317099526e237ae9/internal/contract/controlplane.go#L210-L241) one.

In this check [`semver.ParseTolerant()`](https://github.com/blang/semver/blob/2ee87856327ba09384cabd113bc6b5d174e9ec0f/semver.go#L203-L224) is used for both spec.Version and status.Version, which effectively trims the `v` and automatically puts a trailing `.0` patch section on the version if not specified.

This is then followed up by a [`version.Compare(spec.Version, status.Version, ...)`](https://github.com/kubernetes-sigs/cluster-api/blob/3abcd216748c1dcae6828afb317099526e237ae9/util/version/version.go#L99-L104) which marks IsUpgrading=true when spec.Version is greater than status.Version.

damdo: So considering:
- `s.scope.ControlPlane.Status.Version = cluster.Version` sets `status.Version` to the `major.minor` current version returned by the AWS EKS API (e.g. `1.31`) 
- considering what said in (b), that the `v`s are ignored, and semver normalised
- considering that what is specified in `spec.Version` is then internally reduced by the EKS API to `major.minor`

I see two viable options:

option 1) change the `spec.Version` validation https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/8cc4bebb0737a89dcd05fcb4e64673ef8db13d6f/controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go#L72-L74 to restrict it to only `major.minor`. This should then always result in a comparable version between what the user provides (spec.Version) and what AWS EKS spits out (status.Version). This is likely a validation breaking change.

option 2) rather than directly setting `s.scope.ControlPlane.Status.Version = cluster.Version`, we cheat and look at what the user set in the spec.Version (`s.scope.ControlPlane.Spec.Version`), we then look at whether they set a patch version, we check it matches with the `major.minor` returned by AWS EKS API, and if it does we set the exact/or a greater `major.minor.patch` in status.Version to what the user specified. In case we opt for a greater one, we could use `major.minor.999` or something that highlights it is always greater.

WDYT?

richardcase: @damdo - great write up and thanks for looking into [semver.ParseTolerant()](https://github.com/blang/semver/blob/2ee87856327ba09384cabd113bc6b5d174e9ec0f/semver.go#L203-L224) and proposing options.

Option 1) - i think this is the better long term solution, but it will require an API version bump as we are changing the validation rules on the field. We'd need to have the conversion functions offered by the API version bump to be able to change the value of the version field, from something like `v1.31.2` to `v1.31`, for any existing AWSManagedControlPlanes.

Option 2 ) - is the easier short term solution.

damdo: Updated the PR with Option 2.

richardcase: Added option 1 to the issue tracking v1beta3 changes we want to make

damdo: Great! Thanks!
I've fixed a nil pointer exception that was happening in the mock tests.
So this is done too now.
PR ready for a review @richardcase

richardcase:
@damdo - would you be able to squash your commits?

damdo:
@richardcase done, I left the 3 most relevant atomic commits there. Squashed the rest.

nrb:
@damdo The top line of the description says this is blocked by an open issue - is that still true?

damdo:
@nrb not anymore as we moved to use NLBs in testing

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -283,6 +283,10 @@ type AWSManagedControlPlaneStatus struct {
 	// associated identity provider
 	// +optional
 	IdentityProviderStatus IdentityProviderStatus `json:"identityProviderStatus,omitempty"`
+	// version represents the minimum Kubernetes version for the control plane machines
```
richardcase, line 286: nit:
```suggestion
	// Version represents the minimum Kubernetes version for the control plane machines
```

richardcase (commented):
Looks great @damdo .

Just 1 nit, normally i would say to ignore it but as it will be in the CRD it might be worth correcting. wdyt?

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
damdo: @richardcase  Ah good catch! Updated thanks :)

PR #5214, opened 2024-11-12
:seedling:Add initial Rosa machine pool integration tests

PanSpagetka:
**What type of PR is this?**
Adding tests. I am not sure what kind should apply so I am not applying any.,

Adding basic integration tests for `ROSAMachinePoolReconciler`. One test case for creating new machine pool and for deleting.

To be able to mock OCM and STS calls I had to do small refactoring.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] adds unit tests

**Release note**:

```release-note
NONE
```


serngawy:
@nrb  would you add ok-to-test

main.go
```diff
@@ -238,6 +239,8 @@ func main() {
 			WatchFilterValue: watchFilterValue,
 			WaitInfraPeriod:  waitInfraPeriod,
 			Endpoints:        awsServiceEndpoints,
+			NewOCMClient:     rosa.NewOCMClient,
```
serngawy, line 242: This is not a good idea to create the clients here in the main . why do you need to do that ? for  integration test you may just mock the clients.

pkg/rosa/idps.go
```diff
@@ -14,7 +13,7 @@ const (
 
 // CreateAdminUserIfNotExist creates a new admin user withe username/password in the cluster if username doesn't already exist.
 // the user is granted admin privileges by being added to a special IDP called `cluster-admin` which will be created if it doesn't already exist.
-func CreateAdminUserIfNotExist(client *ocm.Client, clusterID, username, password string) error {
```
serngawy, line 17: We are using the github.com/openshift/rosa/pkg/ocm in order to avoid duplicate the effort of creating a lib that communicate with OCM. Please keep using it for this first integration test iteration

serngawy (changes requested):

serngawy (commented):
Please add the skeleton for RosaControlPlane_controller  integration test as well, it is  important as the RosaMachinePool integration test

main.go
PanSpagetka: Ok, I moved the initialization to `SetupWithManager` that seems to me as the only `ROSAMachinePoolReconciler` method where it fits. I am using same mechanism to create the mock implementation as the STS api.

pkg/rosa/idps.go
```diff
@@ -4,7 +4,6 @@ import (
 	"fmt"
 
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
-	"github.com/openshift/rosa/pkg/ocm"
```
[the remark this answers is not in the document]
PanSpagetka: We can't use  `github.com/openshift/rosa/pkg/ocm` directly here, because it doesn't expose interface that we could mock. So I had to create mockable interface at our side to be able to mock the ocm calls.

The guys that are developing `github.com/openshift/rosa/pkg/ocm`  are planning to create (and I hope also expose) OCM calls as interface to increase the testability of OCM code. So we could get rid of this in the future. But I am not sure how long will it take them to do it.

pkg/rosa/idps.go
PanSpagetka: Same as above

exp/controllers/rosamachinepool_controller.go
```diff
@@ -186,7 +194,7 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 		}
 	}
 
-	ocmClient, err := rosa.NewOCMClient(ctx, rosaControlPlaneScope)
+	ocmClient, err := r.NewOCMClient(ctx, rosaControlPlaneScope)
```
serngawy, line 197: same, check for nil

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +79,245 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "MachinePool",
+				APIVersion: expclusterv1.GroupVersion.String(),
+			},
+			Spec: expclusterv1.MachinePoolSpec{
+				ClusterName: ownerCluster.Name,
+				Template: clusterv1.MachineTemplateSpec{
+					Spec: clusterv1.MachineSpec{
+						ClusterName: ownerCluster.Name,
+					},
```
serngawy, line 189: We should add the infrastructureRef pointing to RosaMachinePool to have proper test,

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +79,245 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+						Kind:       "MachinePool",
+						APIVersion: clusterv1.GroupVersion.String(),
+					},
+				},
+			},
+			TypeMeta: metav1.TypeMeta{
+				Kind:       "ROSAMachinePool",
+				APIVersion: expinfrav1.GroupVersion.String(),
+			},
+			Spec: expinfrav1.RosaMachinePoolSpec{},
```
serngawy, line 211: we should at least set the version, otherwise it should fail

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +79,245 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+				return ocmMock, nil
+			},
+		}
+
+		req := ctrl.Request{}
+		req.NamespacedName = types.NamespacedName{Name: "rosa-machinepool", Namespace: ns.Name}
+
+		result, err := r.Reconcile(ctx, req)
+
+		g.Expect(err).ToNot(HaveOccurred())
```
serngawy, line 262: Please add more validations, check for the RosaMachinePool.status

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +79,245 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+		g.Expect(result).To(Equal(ctrl.Result{}))
+	})
+
+	t.Run("Reconcile delete", func(t *testing.T) {
+		setup(t)
+		defer teardown()
+
+		deleteTime := metav1.NewTime(time.Now().Add(5 * time.Second))
+		rosaMachinePool.ObjectMeta.Finalizers = []string{"finalizer-rosa"}
```
serngawy, line 271: Not sure what are you testing here, but finalizer should be added by the RosaMachinePool controller

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +79,245 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+				return ocmMock, nil
+			},
+		}
+
+		req := ctrl.Request{}
+		req.NamespacedName = types.NamespacedName{Name: "rosa-machinepool", Namespace: ns.Name}
+
+		result, err := r.Reconcile(ctx, req)
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(result).To(Equal(ctrl.Result{}))
```
serngawy, line 308: You need to validate if the RosaMachinePool deleted.

pkg/rosa/client.go
```diff
@@ -20,16 +22,142 @@ const (
 	ocmAPIURLKey = "ocmApiUrl"
 )
 
+type ocmclient struct {
```
serngawy, line 25: OCMClient not ocmclient and please move the OCMClient struct, interface and funcs to another file ex; ocmclient.go

pkg/rosa/idps.go
serngawy: okay, just github duplication

serngawy (changes requested):
add some comments and please add test for rosacontrolplane_controller.go as well.

exp/controllers/rosamachinepool_controller_test.go
PanSpagetka: I am testing that `DeleteNodePool`. It seems that `reconcileDelete` doesn't modify the `ROSAMachinePool` object directly.

exp/controllers/rosamachinepool_controller_test.go
PanSpagetka: Are you sure? I think it should be removed on delete https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosamachinepool_controller.go#L318

serngawy (changes requested):

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +81,459 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+	createObject(g, identity, ns.Name)
+	defer cleanupObject(g, secret)
+	defer cleanupObject(g, identity)
+
+	for i, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			// This is set by CAPI MachinePool reconcile
+			test.old.OwnerReferences = []metav1.OwnerReference{
+				{
+					Name:       ownerMachinePool(i).Name,
```
serngawy, line 356: Usually in k8s project we use the rand [lib](https://pkg.go.dev/k8s.io/apimachinery@v0.32.0/pkg/util/rand#String)  to generate a random string. If you can change it will be better otherwise add TODO note to change that.

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +81,459 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+			},
+			Spec: expinfrav1.RosaMachinePoolSpec{
+				NodePoolName: "test-nodepool-1",
+				Version:      "4.14.5",
+
+				Subnet:       "subnet-id",
+				InstanceType: "m5.large",
+			},
+		}
+		oc := ownerCluster(9)
```
serngawy, line 444: So here you are passing 9 to be similar to the rosaMachinePool naming, its better to use the rand lib generate random string and pass it to all the test CRs

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +81,459 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+
+		ocmMock := mocks.NewMockOCMClient(mockCtrl)
+		expect := func(m *mocks.MockOCMClientMockRecorder) {
+			m.GetNodePool(gomock.Any(), gomock.Any()).DoAndReturn(func(clusterId string, nodePoolID string) (*cmv1.NodePool, bool, error) {
+				nodePoolBuilder := nodePoolBuilder(mp.Spec, omp.Spec)
+				nodePool, err := nodePoolBuilder.ID("node-pool-1").Build()
+				g.Expect(err).NotTo(HaveOccurred())
+				return nodePool, true, nil
+			}).Times(1)
+			m.DeleteNodePool("rosa-control-plane-9", "node-pool-1").DoAndReturn(func(clusterId string, nodePoolID string) error {
```
serngawy, line 467: where is this name `"rosa-control-plane-9"` come from ? if it is the cp CR then better is `cp.Name`

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +81,459 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+		cp := rosaControlPlane(9)
+		objects := []client.Object{oc, omp, cp, mp}
+
+		for _, obj := range objects {
+			createObject(g, obj, ns.Name)
+		}
+
+		cpPh, err := patch.NewHelper(cp, testEnv)
+		cp.Status.Ready = true
+		cp.Status.ID = "rosa-control-plane-9"
```
serngawy, line 455: "rosa-control-plane-9" is repeated many places in the code please define var OR ref to it from RosaControlPlane CR.

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +81,459 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+			NewStsClient:   r.NewStsClient,
+		})
+		g.Expect(err2).ToNot(HaveOccurred())
+
+		err3 := r.reconcileDelete(ctx, machinePoolScope, rosaControlPlaneScope)
+		g.Expect(err3).ToNot(HaveOccurred())
+
+		machinePoolScope.Close()
+		time.Sleep(50 * time.Millisecond)
+		m := &expinfrav1.ROSAMachinePool{}
```
serngawy, line 515: better naming `rosaMachinePool := &expinfrav1.ROSAMachinePool{}`

exp/controllers/rosamachinepool_controller_test.go
```diff
@@ -60,3 +81,459 @@ func TestNodePoolToRosaMachinePoolSpec(t *testing.T) {
...
+		key := client.ObjectKey{Name: mp.Name, Namespace: ns.Name}
+		err4 := testEnv.Get(ctx, key, m)
+		g.Expect(err4).ToNot(HaveOccurred())
```
serngawy, line 518: I don't understand this, After deleting the RosaMachinePool we expect to have error [IsNotFound](https://pkg.go.dev/k8s.io/apimachinery/pkg/api/errors#IsNotFound)  Please check if the error is IsNotFound  error  using the API mentioned not any other error

serngawy (changes requested):
we almost there :)  , few comments and please run` make lint-fix` before pushing your changes to fix all the golang-ci job issue

exp/controllers/rosamachinepool_controller_test.go
PanSpagetka: In  https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosamachinepool_controller.go#L295 the only function that might delete the CR is `DeleteNodePool` that I am mocking. I can mock it in a way that it deletes the machinePool CR, but is that right? Does calling ocm actually delete the CR?

serngawy (approved):
Thanks Robin, looks good as initial test-cases, will add more test-cases later.

serngawy:
@AndiDog @damdo would you review and lgtm this PR

pkg/rosa/client.go
```diff
@@ -32,6 +32,23 @@ func NewOCMClient(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope) (
...
+// NewOCMClient2 creates a new OCM client.
+func NewOCMClient2(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope) (OCMClient, error) {
+	token, url, err := ocmCredentials(ctx, rosaScope)
+	if err != nil {
+		return nil, err
+	}
+	ocmClient, err := ocm.NewClient().Logger(logrus.New()).Config(&ocmcfg.Config{
+		AccessToken: token,
+		URL:         url,
+	}).Build()
```
nrb, line 44: I'm not exactly sure what the difference between this function and `NewOCMClient` is. I see that we wrap the built client in another struct, but the name and godoc string don't really communicate why that's necessary.

nrb: Ok, so it looks like this is creating a struct that fulfills the new interface, and every call site currently is being replaced with this new function attached to the reconciler.

Is there a reason to make a new function at all? Could we update the logic of the old one? It looks like this PR already phases out use of it.

I don't know if there are external users of the `rosa` package in this repo, but if the concern is about not breaking them, then I'd suggest at least calling this new function something like `NewWrappedOCMClient`.

PanSpagetka: @nrb Are the changes ok with you? Or do you have more question/suggestions?

nrb: Thanks! And for future reference, the API diff job is mostly so that reviewers are aware of changes; it's not required to merge if the breaking change is acceptable.

serngawy (approved):

nrb:

Max number of VPCs reached

nrb:
/retest

VPC capacity issues, not flakes in the suite.

damdo:
@PanSpagetka the test failure could be legit, PTAL:

https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5214/pull-cluster-api-provider-aws-test/1884498590013853696#

PanSpagetka:
@serngawy @nrb I resolved the test failure, could you please lgtm once more?

serngawy (approved):
/lgtm

PR #5222, opened 2024-11-20
✨ Support for BoostrapSelfManagedAddons flag for EKS cluster creation

jas-nik:
**What type of PR is this?**
/kind feature

Add flag to support BootstrapSelfManagedAddons to provision Bare EKS cluster without default addons (coreDNS, kube-proxy, aws-vpc-cni)

https://docs.aws.amazon.com/eks/latest/userguide/create-cluster.html

>By default, EKS installs multiple networking add-ons during cluster creation. This includes the Amazon VPC CNI, CoreDNS, and kube-proxy.

>If you’d like to disable the installation of these default networking add-ons, use the parameter below. This may be used for alternate CNIs, such as Cilium. Review the [EKS API reference](https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html) for more information.

Fixes #

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Add flag to support BootstrapSelfManagedAddons to provision Bare EKS cluster without default addons (coreDNS, kube-proxy, aws-vpc-cni)

```

### DIFF
--- a/pkg/cloud/scope/managedcontrolplane.go
+++ b/pkg/cloud/scope/managedcontrolplane.go
@@ -424,6 +424,11 @@ func (s *ManagedControlPlaneScope) DisableVPCCNI() bool {
 	return s.ControlPlane.Spec.VpcCni.Disable
 }
 
+// BootstrapSelfManagedAddons returns whether the AWS EKS networking addons should be disabled.
+func (s *ManagedControlPlaneScope) BootstrapSelfManagedAddons() bool {
+	return s.ControlPlane.Spec.BootstrapSelfManagedAddons
+}
+
 // VpcCni returns a list of environment variables to apply to the `aws-node` DaemonSet.
 func (s *ManagedControlPlaneScope) VpcCni() ekscontrolplanev1.VpcCni {
 	return s.ControlPlane.Spec.VpcCni
--- a/pkg/cloud/services/eks/cluster.go
+++ b/pkg/cloud/services/eks/cluster.go
@@ -473,6 +473,11 @@ func (s *Service) createCluster(eksClusterName string) (*eks.Cluster, error) {
 		Tags:                    tags,
 		KubernetesNetworkConfig: netConfig,
 	}
+	// Only set BootstrapSelfManagedAddons if it's explicitly set to false in the spec
+	// Default is true, so we don't need to set it in that case
+	if !s.scope.BootstrapSelfManagedAddons() {
+		input.BootstrapSelfManagedAddons = aws.Bool(false)
+	}
 
 	var out *eks.CreateClusterOutput
 	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
--- a/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
+++ b/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
@@ -2194,6 +2194,13 @@ spec:
                       will be the default.
                     type: string
                 type: object
+              bootstrapSelfManagedAddons:
+                default: true
+                description: |-
+                  BootstrapSelfManagedAddons is used to set configuration options for
+                  bare EKS cluster without EKS default networking addons
+                  If you set this value to false when creating a cluster, the default networking add-ons will not be installed
+                type: boolean
               controlPlaneEndpoint:
                 description: ControlPlaneEndpoint represents the endpoint used to
                   communicate with the control plane.
--- a/controlplane/eks/api/v1beta1/conversion.go
+++ b/controlplane/eks/api/v1beta1/conversion.go
@@ -43,7 +43,7 @@ func (r *AWSManagedControlPlane) ConvertTo(dstRaw conversion.Hub) error {
 	dst.Spec.Partition = restored.Spec.Partition
 	dst.Spec.RestrictPrivateSubnets = restored.Spec.RestrictPrivateSubnets
 	dst.Status.Version = restored.Status.Version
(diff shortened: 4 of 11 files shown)


jas-nik:
"failed to create new managed VPC: failed to create vpc: The maximum number of VPCs has been reached"

😞

nrb:
@jas-nik Sorry about the delay here, and thank you for this contribution.

I've been troubleshooting EKS CI issues, and this behavior is very welcome :)

I do ask that you add some tests for this case. We'll need at least a [cluster template](https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/test/e2e/data/eks) and to point to that template [in the e2e test config](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/data/e2e_eks_conf.yaml).

jas-nik:
@nrb  Apologies, this fell off my radar. I'll get them added

damdo:
Hey @jas-nik  CAPA doesn't do merge commits, would you be able to rebase instead?
Thanks!

nrb:

Hit VPC limit.

jas-nik:
Still the same issue.

jas-nik:
@damdo  @nrb  @richardcase would you be able to help with the VPC limit issue?

nrb:
@jas-nik Unfortunately we can't allocate more VPCs right now. The best we can do is retry the tests at off-peak times.

richardcase:
@jas-nik @nrb @damdo - are getting the "VPC limit" error when running just the unit tests or the e2e tests? If it's the unit tests, (i.e.  via `/test pull-cluster-api-provider-aws-test`), which i guess is the case from the discussion, then that should be within our control to change as it shouldn't be fitting AWS and this may be coming from out "resource counting" code.

Also, if it is the e2e then potentially we can increase the service limits.

richardcase:
Are we sure thats the real error? We have a test the checks for maximum number of VPCs:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/dcea94f191381f67e543464820291a97ea062107/controllers/awscluster_controller_test.go#L569

I will have a look at the logs in the morning to check.

jas-nik:
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5336/pull-cluster-api-provider-aws-test/1887504193065848832 - One of the latest PRs build passed even after VPC limit failure, so it might be a red herring after all. Thank you for chiming in. Still need to investigate what is the actual failure.

jas-nik:
@nrb @richardcase - Looks like the issue was due to the test cases expecting default value for `BootstrapSelfManagedAddons` flag. Adjusting expectedSpec for them solves the issue but not sure if that is the right approach here. Should I add a default value for the flag or remove the default  value all together since AWS already considers default as `True`.

pkg/cloud/scope/managedcontrolplane.go
```diff
@@ -424,6 +424,11 @@ func (s *ManagedControlPlaneScope) DisableVPCCNI() bool {
 	return s.ControlPlane.Spec.VpcCni.Disable
 }
 
+// BootstrapSelfManagedAddons returns whether the AWS EKS networking addons should be disabled.
+func (s *ManagedControlPlaneScope) BootstrapSelfManagedAddons() bool {
```
nrb, line 428: This doesn't appear to be used anywhere - is that on purpose?

nrb:
> Should I add a default value for the flag or remove the default value all together since AWS already considers default as True.

I'm fine with the code you've got as-is. Having it explicit in the CAPA side is nice for users to see at a glance.

Overall this looks good to me, just one question.

pkg/cloud/scope/managedcontrolplane.go
jas-nik:
```
if !s.scope.ControlPlane.Spec.BootstrapSelfManagedAddons
``` 
in /services/eks/cluster.go should have been 

```
if !s.scope.BootstrapSelfManagedAddons()
```

pkg/cloud/services/eks/cluster.go
```diff
@@ -473,6 +473,11 @@ func (s *Service) createCluster(eksClusterName string) (*eks.Cluster, error) {
 		Tags:                    tags,
 		KubernetesNetworkConfig: netConfig,
 	}
+	// Only set BootstrapSelfManagedAddons if it's explicitly set to false in the spec
+	// Default is true, so we don't need to set it in that case
+	if !s.scope.BootstrapSelfManagedAddons() {
```
nrb, line 478: Could this be simplified to `input.BootstrapSelfManagedAddons = aws.Bool(s.scope.BoostralSelfManagedAddons())`?

That's setting it no matter what, but it saves a conditional. The compiler might even be doing this behind the scenes, too.

jas-nik: With the default `BootstrapSelfManagedAddons` value (true), we do not need to set it for EKS control plane. We only need to adjust the `input` when BoostrapSelfManagedAddons is set to false. I assume this would be safer for CAPA users who are unaware of this flag and in case AWS does modify this flag in future.

nrb: Fair enough!

nrb:
/approve

I'll make a separate issue for the VPC limits; we seem to hit that a lot, so we likely need to up some limit or adjust the test code that Richard linked to.

nrb:
Issue for the maximum VPC flake: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5415

PR #5230, opened 2024-11-25
WIP: Proposal for dedicated hosts

faermanj:
/kind documentation

Adds the proposal for supporting [EC2 Dedicated Hosts](https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-hosts-overview.html)

**Checklist**:
- [X] squashed commits
- [X] includes documentation

**Release note**:
NONE

### DIFF
--- a/docs/proposal/20241125-ec2-dedicated-hosts.md
+++ b/docs/proposal/20241125-ec2-dedicated-hosts.md
@@ -1,0 +1,100 @@
+---
+title: Add Support for EC2 Dedicated Hosts
+authors:
+  - "@faermanj"
+reviewers:
+creation-date: 2024-11-25
+last-updated: 2024-11-25
+status: draft
+see-also: []
+replaces: []
+superseded-by: []
+---
+
+# Add Support for EC2 Dedicated Hosts
+
+## Table of Contents
+
+- [Add Support for EC2 Dedicated Hosts](#start)
+  - [Table of Contents](#table-of-contents)
+  - [Glossary](#glossary)
+  - [Summary](#summary)
+  - [Motivation](#motivation)
+    - [Goals](#goals)
+    - [Non-Goals/Future Work](#non-goalsfuture-work)
+  - [Proposal](#proposal)
+    - [User Stories](#user-stories)
+      - [Story 1](#story-1)
+    - [Requirements](#requirements)
+      - [Functional Requirements](#functional-requirements)
+      - [Non-Functional Requirements](#non-functional-requirements)
+    - [Implementation Details/Notes/Constraints](#implementation-detailsnotesconstraints)
+    - [Security Model](#security-model)
+    - [Risks and Mitigations](#risks-and-mitigations)
+  - [Alternatives](#alternatives)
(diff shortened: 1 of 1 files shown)


docs/proposal/20241125-ec2-dedicated-hosts.md
```diff
@@ -0,0 +1,100 @@
...
+- Dedicated Host allocation. For now, dedicated hosts must be pre-allocated and explicitly configured. In future work, we may want to auto-allocate / release dedicated hosts.
+- Custom replica allocation policy. For now, replicas are allocated in the first dedicated host that accepts the instance. In future work, we may let users configure other policies (round-robin, least-utilized, ...)
+
+## Proposal
+- Add list host id and affinity to the proper kind spec (`AWSMachineTemplate`?).
+- Define and document the initial policy for instance allocation
+- Add an E2E test case
+
+
+### Risks and Mitigations
```
mtulio, line 67: Is there any risk of control planes "losing" HW HA when allocating control planes in the same host, or will the proposal covers only the EC2 instances for worker/compute nodes?

docs/proposal/20241125-ec2-dedicated-hosts.md
```diff
@@ -0,0 +1,100 @@
...
+
+### Goals
+1. On cluster creation, let users specify the host id and affinity for instances.
+
+### Non-Goals/Future Work
+- Dedicated Host allocation. For now, dedicated hosts must be pre-allocated and explicitly configured. In future work, we may want to auto-allocate / release dedicated hosts.
+- Custom replica allocation policy. For now, replicas are allocated in the first dedicated host that accepts the instance. In future work, we may let users configure other policies (round-robin, least-utilized, ...)
+
+## Proposal
+- Add list host id and affinity to the proper kind spec (`AWSMachineTemplate`?).
```
damdo, line 62: Where is this available in the instance object exposed by the AWS client?

damdo (commented):
Hey @faermanj thanks for opening this proposal.
On a first pass I think it sounds reasonable to me. I left a question.

damdo:
/cc. @nrb

docs/proposal/20241125-ec2-dedicated-hosts.md
nrb: Would it be reasonable to require a minimum of 3 dedicated hosts?

nrb:
I haven't forgotten about this, but have been focusing on investigating CI failures the last couple weeks.

dlipovetsky:
From office hours:
- Let's not allow the user to specify a list of dedicated hosts, because that would put CAPA in the position of deciding which host to use for a specific AWSMachine.
- We could add a field to AWSMachineSpec, so that the user could specify a dedicated host in an AWSMachineTemplate. Then, the user could create a MachineDeployment that references the AWSMachineTemplate, and all VMs in the MD would be run on the dedicated host. The user also has the option of creating an AWSMachine "by hand."

damdo:
@faermanj I saw you closed this, was it intentional?

PR #5188, opened 2024-10-28
🐛chore: downgrade kustomize v5 version

richardcase:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

This downgrades customize v5 to version 5.3.0 from 5.5.0 as there is an issue using the higher vsrion as it requires go version 1.22.7 and the gcloud build image only has a lower version of go.

```
make -C hack/tools bin/kustomize
make[4]: Entering directory '/workspace/hack/tools'
mkdir -p bin
CGO_ENABLED=0 go build -tags=tools -o bin/kustomize sigs.k8s.io/kustomize/kustomize/v5
go: go.mod requires go >= 1.22.7 (running go 1.22.0; GOTOOLCHAIN=local)
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
NONE
```

PR #5240, opened 2024-12-06
:bug: Downgrade kustomize and prevent dependabot bumps

nrb:
**What type of PR is this?**

/kind regression

Kustomize previously broke builds
(https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5188)
and we mistakenly approved a version bump
(https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5232)
that caused the Go version to be >= 1.22.7. As of right now, the builder
image only has Go 1.22.0.

This change downgrades the relevant kustomize dependencies, as well as
prevents dependabot from incrementing kustomize automatically in the
future.

**Checklist**:

- [x] squashed commits
- [x] includes documentation

**Release note**:

```release-note
NONE
```


dlipovetsky:
Thank you!

/approve
/lgtm

damdo:
To avoid issues like this in the future I really think we should run pull-cluster-api-provider-aws-e2e by default on PRs(presubmits), and maybe have it non optional. It would catch all of these regressions.
cc. @nrb @dlipovetsky @richardcase @Ankitasw @AndiDog

damdo:
^ Folks what do you think of the above?

PR #5242, opened 2024-12-06
:bug: Attempt to clean up CF IAM users

nrb:
**What type of PR is this?**
/kind failing-test

Periodic tests seemed to get into a failure loop because an IAM user
with the same name already existed, which is not allowed. This then
failed the entire CloudFoundation stack. Depite the stack claiming to
have been rolled back, the next iteration would run into the same
problem.

This change includes IAM users in the list of resources we need to
specifically delete in the case of a CloudFoundation failure, just in
case they've leaked

The periodic tests at https://testgrid.k8s.io/sig-cluster-lifecycle-cluster-api-provider-aws#periodic-e2e-release-2-7 were failing roughly every other day between Nov 23, 2024 to Dec 6, 2024.
We'd seen failures prior to that, but testgrid's history doesn't appear to go that far back.

Nearly all the failures within the `capa-e2e.[SynchronizedBeforeSuite]` function contained this log entry:

```
STEP: Event details for AWSIAMUserBootstrapper : Resource: AWS::IAM::User, Status: CREATE_FAILED, Reason: Resource handler returned message: "Resource of type 'AWS::IAM::User' with identifier 'bootstrapper.cluster-api-provider-aws.sigs.k8s.io' already exists." (RequestToken: 9149fdc5-32aa-007f-086d-d60101e23ee9, HandlerErrorCode: AlreadyExists) @ 12/05/24 15:56:04.338
```

**Checklist**:

**Release note**:
-->
```release-note
NONE
```

### DIFF
--- a/test/e2e/shared/aws.go
+++ b/test/e2e/shared/aws.go
@@ -474,6 +474,7 @@ func deleteResourcesInCloudFormation(prov client.ConfigProvider, t *cfn_bootstra
 	iamSvc := iam.New(prov)
 	temp := *renderCustomCloudFormation(t)
 	var (
+		iamUsers         []*cfn_iam.User
 		iamRoles         []*cfn_iam.Role
 		instanceProfiles []*cfn_iam.InstanceProfile
 		policies         []*cfn_iam.ManagedPolicy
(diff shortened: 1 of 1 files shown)


nrb:
Probably needs to be rebased onto https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5240

nrb:
@richardcase Yeah, I asked about the janitor [on Slack](https://kubernetes.slack.com/archives/CCK68P2Q2/p1733513944450579). The IAM code only looks at roles and instance policies (https://github.com/kubernetes-sigs/boskos/tree/master/aws-janitor/resources).

I'm suspecting that what could be happening is that multiple periodics are using CF at the same time and stepping on each other. With your account logging PR, we can double check that in the future.

AndiDog:
I'm running in to the same issue regularly with PR E2E tests, thanks for fixing this!

/lgtm

nrb:
Removing WIP on this as I think it's ready, excepting problems with #5252

nrb:
Since this is tying up LF resources fairly consistently, I'm going to self-approve.

/approve

PR #5276, opened 2025-01-13
:book: Proposal for two new AWS self-managed cluster feature gates

mzazrivec:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** 
This proposal should be approved before https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5124 can be implemented.

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [X] includes documentation
- [x] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
NONE
```

PR #5284, opened 2025-01-15
:sparkles: Introduce AWSMachine and AWSCluster feature gates

mzazrivec:
**What type of PR is this?**

/kind feature

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Added AWSMachine and AWSCluster feature flags to control AWSMachineReconciler and AWSClusterReconciler.
```

### DIFF
--- a/docs/book/src/topics/reference/reference.md
+++ b/docs/book/src/topics/reference/reference.md
@@ -16,3 +16,5 @@
 | AlternativeGCStrategy         | EXP_ALTERNATIVE_GC_STRATEGY       | false   |
 | TagUnmanagedNetworkResources  | TAG_UNMANAGED_NETWORK_RESOURCES   | true    |
 | ROSA                          | EXP_ROSA                          | false   |
+| AWSMachine                    | CAPA_AWS_MACHINE                  | true    |
+| AWSCluster                    | CAPA_AWS_CLUSTER                  | true    |
--- a/feature/feature.go
+++ b/feature/feature.go
@@ -86,6 +86,16 @@ const (
 	// owner: @enxebre
 	// alpha: v2.2
 	ROSA featuregate.Feature = "ROSA"
+
+	// AWSMachine is used to control AWSMachine reconciler
+	// owner: @mzazrivec
+	// beta: v2.7
+	AWSMachine featuregate.Feature = "AWSMachine"
+
+	// AWSCluster is used to control AWSCluster reconciler
+	// owner: @mzazrivec
+	// beta: v2.7
+	AWSCluster featuregate.Feature = "AWSCluster"
 )
 
 func init() {
@@ -108,4 +118,6 @@ var defaultCAPAFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	AlternativeGCStrategy:         {Default: false, PreRelease: featuregate.Alpha},
 	TagUnmanagedNetworkResources:  {Default: true, PreRelease: featuregate.Alpha},
 	ROSA:                          {Default: false, PreRelease: featuregate.Alpha},
+	AWSMachine:                    {Default: true, PreRelease: featuregate.Beta},
+	AWSCluster:                    {Default: true, PreRelease: featuregate.Beta},
 }
--- a/config/manager/manager.yaml
+++ b/config/manager/manager.yaml
@@ -19,7 +19,7 @@ spec:
       containers:
       - args:
         - "--leader-elect"
-        - "--feature-gates=EKS=${CAPA_EKS:=true},EKSEnableIAM=${CAPA_EKS_IAM:=false},EKSAllowAddRoles=${CAPA_EKS_ADD_ROLES:=false},EKSFargate=${EXP_EKS_FARGATE:=false},MachinePool=${EXP_MACHINE_POOL:=false},EventBridgeInstanceState=${EVENT_BRIDGE_INSTANCE_STATE:=false},AutoControllerIdentityCreator=${AUTO_CONTROLLER_IDENTITY_CREATOR:=true},BootstrapFormatIgnition=${EXP_BOOTSTRAP_FORMAT_IGNITION:=false},ExternalResourceGC=${EXP_EXTERNAL_RESOURCE_GC:=false},AlternativeGCStrategy=${EXP_ALTERNATIVE_GC_STRATEGY:=false},TagUnmanagedNetworkResources=${TAG_UNMANAGED_NETWORK_RESOURCES:=true},ROSA=${EXP_ROSA:=false}"
+        - "--feature-gates=EKS=${CAPA_EKS:=true},EKSEnableIAM=${CAPA_EKS_IAM:=false},EKSAllowAddRoles=${CAPA_EKS_ADD_ROLES:=false},EKSFargate=${EXP_EKS_FARGATE:=false},MachinePool=${EXP_MACHINE_POOL:=false},EventBridgeInstanceState=${EVENT_BRIDGE_INSTANCE_STATE:=false},AutoControllerIdentityCreator=${AUTO_CONTROLLER_IDENTITY_CREATOR:=true},BootstrapFormatIgnition=${EXP_BOOTSTRAP_FORMAT_IGNITION:=false},ExternalResourceGC=${EXP_EXTERNAL_RESOURCE_GC:=false},AlternativeGCStrategy=${EXP_ALTERNATIVE_GC_STRATEGY:=false},TagUnmanagedNetworkResources=${TAG_UNMANAGED_NETWORK_RESOURCES:=true},ROSA=${EXP_ROSA:=false},AWSMachine=${CAPA_AWS_MACHINE:=true},AWSCluster=${CAPA_AWS_CLUSTER:=true}"
         - "--v=${CAPA_LOGLEVEL:=0}"
         - "--diagnostics-address=${CAPA_DIAGNOSTICS_ADDRESS:=:8443}"
         - "--insecure-diagnostics=${CAPA_INSECURE_DIAGNOSTICS:=false}"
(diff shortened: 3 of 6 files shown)


fiunchinho (commented):
Would it be possible to add tests?

feature/feature.go
```diff
@@ -108,4 +116,6 @@ var defaultCAPAFeatureGates = map[featuregate.Feature]featuregate.FeatureSpec{
 	AlternativeGCStrategy:         {Default: false, PreRelease: featuregate.Alpha},
 	TagUnmanagedNetworkResources:  {Default: true, PreRelease: featuregate.Alpha},
 	ROSA:                          {Default: false, PreRelease: featuregate.Alpha},
+	AWSMachine:                    {Default: true, PreRelease: featuregate.Alpha},
```
serngawy, line 119: I believe both feature gates should be beta not alpha

mzazrivec: Fixed

mzazrivec:
> Would it be possible to add tests?

While I understand the utility and usefulness of tests, creating tests for the feature gates would require changing `main.go`, maybe other files too (i.e. there are no tests for feature gates yet). Since this pull request implements the proposal from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5276 I really want to avoid any further changes here.

serngawy:
> > Would it be possible to add tests?
> 
> While I understand the utility and usefulness of tests, creating tests for the feature gates would require changing `main.go`, maybe other files too (i.e. there are no tests for feature gates yet). Since this pull request implements the proposal from #5276 I really want to avoid any further changes here.

You may update the e2e test to check for the feature-gate is enbaled/disabled similar to [AlternativeGCStrategy test](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/suites/gc_unmanaged/gc_unmanaged_test.go#L159)

feature/feature.go
```diff
@@ -86,6 +86,14 @@ const (
 	// owner: @enxebre
 	// alpha: v2.2
 	ROSA featuregate.Feature = "ROSA"
+
+	// owner: @mzazrivec
+	// alpha: v2.7
```
serngawy, line 91: beta

feature/feature.go
```diff
@@ -86,6 +86,14 @@ const (
...
+	AWSMachine featuregate.Feature = "AWSMachine"
+
+	// owner: @mzazrivec
+	// alpha: v2.7
```
serngawy, line 95: beta

serngawy (changes requested):
/ok-to-test

damdo (commented):
Thanks for working on this @mzazrivec @serngawy 
It looks reasonable to me.

/lgtm

feature/feature.go
damdo: Can you update this @mzazrivec ?

feature/feature.go
damdo: Can you update this @mzazrivec ?

damdo:
/assign @richardcase @nrb @dlipovetsky @AndiDog 

For approver review

feature/feature.go
mzazrivec: Done.

damdo (commented):
/lgtm

damdo:
Hey @mzazrivec did you have a look at the CAPA logs for these failed job runs: https://gcsweb.k8s.io/gcs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5284/pull-cluster-api-provider-aws-e2e/1896860470422278144/artifacts/clusters/

Anything suspicious there?

damdo:
Could be yeah, maybe it was missing some changes there @mzazrivec could you try rebase on top of the latest main? Thanks!

mzazrivec:
@damdo rebased, we'll see if it helps.

damdo:
Thanks I've already triggered the E2Es, let's see

damdo:
Looking at the changes I don't see how they could contribute to failing the e2e. They seem unrelated.


mzazrivec:
It seems that wich each e2e test run random tests time out and then the whole e2e job fails.

Anyway ... restarting.

mzazrivec:
The aws e2e test is green now.

damdo (commented):
/lgtm

Thanks

/assign @nrb @richardcase @dlipovetsky @Ankitasw @AndiDog

nrb (changes requested):
Thanks for this! I appreciate the work, but I think perhaps these should be proper arguments rather than feature flags.

The feature flags should be used for things that are being introduced for testing and opt-in. This PR makes the default invocation of the binary disable these reconcilers, and I don't think that's how most users expect things to work. See my comment inline for more info.

feature/feature.go
```diff
@@ -86,6 +86,16 @@ const (
 	// owner: @enxebre
 	// alpha: v2.2
 	ROSA featuregate.Feature = "ROSA"
+
+	// AWSMachine is used to control AWSMachine reconciler
+	// owner: @mzazrivec
+	// beta: v2.7
```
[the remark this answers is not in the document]
mzazrivec: Not sure if these new feature gates should be beta or something else. I changed that to beta because I was told to do so few comments above. I'm happy to change it to whatever makes sense though.

Regarding the question of feature gates vs separate cmd line: this pull request implements the proposal from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5276 as it has been approved & merged. We can certainly discuss whether cmd line is more appropriate (I don't feel strongly either way TBH), but it makes me wonder what that means for the proposal doc.

Regarding the question of the default behavior: both of the AWS reconcilers affected by these two new feature gates
would of course be enabled by default, i.e. upgrades would not be affected and you'd have to explicitly disable them on cmd line for those two reconcilers to be turned off.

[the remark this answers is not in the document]
damdo: Alpha Feature Gates are normally false by default, whereas Beta Feature Gates are true by default.
Since these controllers are already normally running setting them to Beta means they are true by default, and so they run by default, if nothing is specified.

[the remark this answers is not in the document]
damdo: Yes I also tend to agree with the approach @mzazrivec took, considering the existing featureGates for the other controllers.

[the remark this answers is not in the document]
richardcase: EKS has feature gates because historically it was experimental and as such we used feature gates to allow people to enable that feature.

We never removed the feature gates for EKS just in case we broke people. But really we should've removed them.

docs/book/src/topics/reference/reference.md
```diff
@@ -15,4 +15,6 @@
 | ExternalResourceGC            | EXP_EXTERNAL_RESOURCE_GC          | false   |
 | AlternativeGCStrategy         | EXP_ALTERNATIVE_GC_STRATEGY       | false   |
 | TagUnmanagedNetworkResources  | TAG_UNMANAGED_NETWORK_RESOURCES   | true    |
-| ROSA                          | EXP_ROSA                          | false   |
\ No newline at end of file
+| ROSA                          | EXP_ROSA                          | false   |
```
richardcase, line 18: Probably a naive question from my side. If you use ROSA do you disable AWSMachine and AWSCluster always?

feature/feature.go
```diff
@@ -86,6 +86,16 @@ const (
...
+	AWSMachine featuregate.Feature = "AWSMachine"
```
richardcase, line 93: With hindsight i think these should be a new single flag to disable both and not feature flags.

docs/book/src/topics/reference/reference.md
serngawy: It doesn't make conflict with ROSA, but in production env you don't run controllers that you don't use.

feature/feature.go
serngawy: In the proposal we discussed that theoretically end user can use awsCluster with machinePool without using awsMachine. That was the main reason for having 2 feature gates.

docs/book/src/topics/reference/reference.md
richardcase: So really it's either the Rosa controllers or the original capa controllers then.

feature/feature.go
richardcase: "awsCluster with machinePool without using awsMachine" that's not a goal of CAPA. With CAPA (excluding Rosa) the only choice is around enabling feature. This scenario is not something we want to support at this stage.

serngawy: that is okay, its just during discussion we ask is that possible and answer was could be :)

docs/book/src/topics/reference/reference.md
richardcase: Got it, thanks @serngawy .

nrb:
@serngawy @mzazrivec Apologies for changing the terms of the proposal on you - we very much appreciate the time and effort that went into it, and are not objecting to this implementation out of spite.

I've posted an alternative at #5395. This introduces an new argument that allows for selective disabling of controllers, along with a proposal for how to move forward with EKS and ROSA.

This new proposal does contradict @richardcase's [stance](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5284#discussion_r1992338068); however, I think it provides a compromise between the two aims that we're trying to meet.

We may also want to discuss separating binaries at some point - that would perhaps be the cleanest separation between ROSA and other controllers. However, I believe that will take more time to implement.

 @damdo @Ankitasw @AndiDog @dlipovetsky

mzazrivec:
Closing, as https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5395 got merged.

damdo:
Thanks all! Closing

PR #5286, opened 2025-01-16
✨ Support setting role path and permissions boundary on managed IAM roles

### LINKED ISSUE #5285: Support setting role path and permissions boundary on managed IAM roles
/kind feature

We use CAPA to create and manage the IAM roles for EKS and Fargate. We have the requirement that our roles must be created under a specific path, and with a permissions boundary attached. Setting a path and permissions boundary is a very common requirement in enterprise IAM environments. Unfortunately, this is not supported by CAPA.

The call that creates the IAM role is located [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/v2.7.1/pkg/cloud/services/eks/iam/iam.go#L187C4-L213):

```go
// CreateRole will create a role from the IAMService.
func (s *IAMService) CreateRole(
	roleName string,
	key string,
	trustRelationship *iamv1.PolicyDocument,
	additionalTags infrav1.Tags,
) (*iam.Role, error) {
	tags := RoleTags(key, additionalTags)

	trustRelationshipJSON, err := converters.IAMPolicyDocumentToJSON(*trustRelationship)
	if err != nil {
		return nil, errors.Wrap(err, "error converting trust relationship to json")
	}

	input := &iam.CreateRoleInput{
		RoleName:                 aws.String(roleName),
		Tags:                     tags,
		AssumeRolePolicyDocument: aws.String(trustRelationshipJSON),
	}

	out, err := s.IAMClient.CreateRole(input)
	if err != nil {
		return nil, errors.Wrap(err, "failed to call CreateRole")
	}

	return out.Role, nil
}
```

The path and permissions boundary are both string values that must be added to [`CreateRoleInput`](https://pkg.go.dev/github.com/aws/aws-sdk-go@v1.55.5/service/iam#CreateRoleInput).
Adding these fields to the Spec and passing them through the stack should be enough.

**Environment:**

- Cluster-api-provider-aws version: 2.7.1

robinkb:
**What type of PR is this?**

/kind feature

Allows setting the role path and permissions boundary on managed IAM roles for EKS control plane, EKS Fargate profile, and managed machine pools. This is typically needed in enterprise environments with more stringent IAM requirements.

Fixes #5285 

This is my first PR on this repository. I noticed a warning when generating code, and I wasn't sure if this needed to be fixed:

```
E0116 12:53:58.750053   55614 conversion.go:744] Warning: could not find nor generate a final Conversion function for sigs.k8s.io/cluster-api-provider-aws/v2/exp/api/v1beta2.FargateProfileSpec -> sigs.k8s.io/cluster-api-provider-aws/v2/exp/api/v1beta1.FargateProfileSpec
E0116 12:53:58.750274   55614 conversion.go:745]   the following fields need manual conversion:
E0116 12:53:58.750281   55614 conversion.go:747]       - RolePath
E0116 12:53:58.750286   55614 conversion.go:747]       - RolePermissionsBoundary
```

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] adds unit tests

**Release note**:

```release-note
Support setting role path and permissions boundary on managed IAM roles. **Action Required**: If you want to use permission boundaries, then you will need to update your IAM permissions by running **clusterawsadm bootstrap iam create-cloudformation-stack** again.
```


pavansokkenagaraj:
@robinkb PermissionsBoundary are required for CloudFormation template rendering for creating roles and permissions.

robinkb:
Hi @pavansokkenagaraj, thanks for the feedback.

I've amended my commit. However, I see more roles generated in the template:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/cmd/clusterawsadm/cloudformation/bootstrap/template.go#L183-L208

Do these need to be changed, too? I didn't use the bootstrap proces, so I'm not really familiar with how it works.

robinkb:
We are in the process of testing this on our own environment now. The most recent update includes some shenanigans around pointers. I'm not sure which way I am expected to handle them, as the codebase is inconsistent. The Spec for AWSManagedControlPlane uses pointers to strings for optional values, while the Spec for AWSFargateProfile does not, for example.

cmd/clusterawsadm/api/bootstrap/v1alpha1/types.go
```diff
@@ -83,6 +83,12 @@ type AWSIAMRoleSpec struct {
 	// ExtraStatements are additional IAM statements to be included inline for the role.
 	ExtraStatements []iamv1.StatementEntry `json:"extraStatements,omitempty"`
 
+	// Path sets the path to the role.
+	Path string `json:"path,omitempty"`
```
richardcase, line 87: The new fields only need to go into the latest API version (i.e. v1beta1). These can be removed.

But what this means is that the conversions need to be updated so that values for these new fields are restored when roundtripping.

A bit like we do here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/api/v1beta1/awscluster_conversion.go#L27

cmd/clusterawsadm/api/bootstrap/v1beta1/types.go
```diff
@@ -83,6 +83,12 @@ type AWSIAMRoleSpec struct {
 	// ExtraStatements are additional IAM statements to be included inline for the role.
 	ExtraStatements []iamv1.StatementEntry `json:"extraStatements,omitempty"`
 
+	// Path sets the path to the role.
+	Path string `json:"path,omitempty"`
```
richardcase, line 87: Could you add the "optional" kubebuilder comment?

cmd/clusterawsadm/api/bootstrap/v1beta1/types.go
```diff
@@ -83,6 +83,12 @@ type AWSIAMRoleSpec struct {
...
+
+	// PermissionsBoundary sets the ARN of the managed policy that is used to set the permissions boundary for the role.
+	PermissionsBoundary string `json:"permissionsBoundary,omitempty"`
```
richardcase, line 90: Same comment about optional.

exp/api/v1beta1/awsfargateprofile_types.go
```diff
@@ -60,6 +60,28 @@ type FargateProfileSpec struct {
 	// +optional
 	RoleName string `json:"roleName,omitempty"`
 
+	// RolePath sets the path to the role. For more information about paths, see IAM Identifiers
+	// (https://docs.aws.amazon.com/IAM/latest/UserGuide/Using_Identifiers.html)
+	// in the IAM User Guide.
+	//
+	// This parameter is optional. If it is not included, it defaults to a slash
+	// (/).
+	RolePath string `json:"rolePath,omitempty"`
```
richardcase, line 69: We don't need these in this API version (see previous comment).

pavansokkenagaraj:
@robinkb 
> Do these need to be changed, too? I didn't use the bootstrap proces, so I'm not really familiar with how it works.

Yes,  roles/permissions which are created in template function for eks need to be updated.

ref: - https://github.com/pavansokkenagaraj/cluster-api-provider-aws/blob/c518150b3a4aeabb38cab546d8a89988307cd053/cmd/clusterawsadm/cloudformation/bootstrap/template.go#L201-L238

The other thing I had to do was to set the permissionsBoundary if it is not nil/empty
if permissionsBoundary is empty, AWS IAM role creation returns an err expecting permissionsBoundary to be length of greater than 0(This needs some testing)

https://github.com/pavansokkenagaraj/cluster-api-provider-aws/blob/c518150b3a4aeabb38cab546d8a89988307cd053/pkg/cloud/services/eks/iam/iam.go#L190-L203

cmd/clusterawsadm/api/bootstrap/v1alpha1/types.go
robinkb: I managed to update the conversions for the other types, but I am confused as to why I need to implement conversion for this type in particular? The bootstrap API is only for a local configuration file, right? It's not stored in etcd?

There is no conversion in place for it today. If I do need to implement conversion for it, then sure, I will do that. But I will require help, because the test is spitting out vague errors that I can make neither heads nor tails out of.

robinkb:
@pavansokkenagaraj Thanks for the testing, I will get that fixed.

Regarding the failing `apidiff`: How do we handle that? I don't see how I could fix this issue without breaking the API of that particular package, unless we want to add a whole other function.

cmd/clusterawsadm/api/bootstrap/v1alpha1/types.go
richardcase: > I managed to update the conversions for the other types, but I am confused as to why I need to implement conversion for this type in particular? The bootstrap API is only for a local configuration file, right? It's not stored in etcd?

Normally for any Kubernetes API there is a test that does roundtrip conversion. But actually looking at this API we don't have those tests. With is great. So no need to worry about the conversion. I wonder if it is worth adding conversion functions at some point in the future. 

The new field(s) still only need to go into the latest API version.

richardcase:
@robinkb - we don't need to worry about the apidiff if its outside an `api` directory. The project doesn't provider guarantees for the exported API types outside of the APIs.

richardcase:
@robinkb  - would you be able to change the release note section to "action required". Normaly when we make changes to the IAM permissions we call it out the release notes to say something like `If you want to use permission boundaries, then you will need to update your IAM permissions by running **clusterawsadm bootstrap iam create-cloudformation-stack** again`

damdo:
@robinkb have you looked into the e2e-eks failure? It seems related to identities so worth checking if it is legit.

robinkb:
@damdo The tests did not run, failing in the `BeforeSuite`. The failure in the `AfterSuite` comes from trying to clean up a resource that was never created in the first place.

robinkb:
I will try to run the test again to see if anything changed.

/retest

damdo:
Let's get you reviews on this one
I'll try to have a look

also
/assign @nrb @richardcase @dlipovetsky @AndiDog @Ankitasw

richardcase:
The e2e tests are passing.

@robinkb - do you think it would be worth us following up on this in the future with an e2e that covers this?

aristosvo:
Thanks @robinkb and maintainers👌🚀

robinkb:
@richardcase Thank you! :tada: 

Because it's only passing through two strings fields, I don't think that adding an e2e test for this is really necessary. *Maybe* an integration test to verify that the parameters are correctly passed to the AWS API? But even that feels a bit superfluous. The `rolePath` parameter could be set in an existing e2e test, but that could break IAM policies.

Probably not worth the effort.

PR #5318, opened 2025-01-22
✨ Support storing Ignition user data in S3 bucket for `AWSMachinePool`

AndiDog:

Reopened from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5172 because of branch issues.

EC2 user data is insecure and very limited in size (16 KB). Use the S3 bucket storage also for machine pools. The difference to `AWSMachine` is that the S3 objects must stay around because a new instance could be started by the ASG at any time, and user data with a valid token must exist. Both a lifecycle policy for regular deletion, and explicit S3 object deletion requests, are used for cleanup, so that normally, only 1-3 launch template versions' user data are stored in S3. Each launch template version relates to one S3 object.

**What type of PR is this?**

/kind feature

Didn't find an existing issue. My company ran into EC2 user data size limit problems.

This change has been successfully used in the Giant Swarm fork of CAPA since July 2024, on production clusters. I have adapted the change to the latest upstream changes, in particular the introduction of `IgnitionStorageTypeOption` (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/4700) and made sure the previous default storage type is kept. Therefore, I retested this feature using Tilt on a plain CAPI@main + CAPA@main setup, with AWSCluster + Flatcar AMI + AWSMachinePool in my manifests.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] adds unit tests

**Release note**:

```release-note
action required
Support storing Ignition user data in S3 bucket for `AWSMachinePool`. If you plan to use this feature then you will need to update the permissions for CAPA using clusterawsadm. 
```

### DIFF
--- a/exp/controllers/awsmachinepool_controller.go
+++ b/exp/controllers/awsmachinepool_controller.go
@@ -50,6 +50,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services"
 	asg "sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/autoscaling"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/ec2"
+	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/s3"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/logger"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	expclusterv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
@@ -67,6 +68,7 @@ type AWSMachinePoolReconciler struct {
 	asgServiceFactory            func(cloud.ClusterScoper) services.ASGInterface
 	ec2ServiceFactory            func(scope.EC2Scope) services.EC2Interface
 	reconcileServiceFactory      func(scope.EC2Scope) services.MachinePoolReconcileInterface
+	objectStoreServiceFactory    func(scope.S3Scope) services.ObjectStoreInterface
 	TagUnmanagedNetworkResources bool
 }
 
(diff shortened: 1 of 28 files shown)


exp/controllers/awsmachinepool_controller.go
```diff
@@ -280,7 +296,7 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 		machinePoolScope.Info("starting instance refresh", "number of instances", machinePoolScope.MachinePool.Spec.Replicas)
 		return asgsvc.StartASGInstanceRefresh(machinePoolScope)
 	}
-	if err := reconSvc.ReconcileLaunchTemplate(machinePoolScope, ec2Svc, canUpdateLaunchTemplate, runPostLaunchTemplateUpdateOperation); err != nil {
+	if err := reconSvc.ReconcileLaunchTemplate(machinePoolScope, machinePoolScope, s3Scope, ec2Svc, objectStoreSvc, canUpdateLaunchTemplate, runPostLaunchTemplateUpdateOperation); err != nil {
```
fiunchinho, line 299: why do we need to pass the same `machinePoolScope` twice?

exp/controllers/awsmachinepool_controller.go
```diff
@@ -607,7 +623,7 @@ func machinePoolToInfrastructureMapFunc(gvk schema.GroupVersionKind) handler.Map
 	}
 }
 
-func (r *AWSMachinePoolReconciler) getInfraCluster(ctx context.Context, log *logger.Logger, cluster *clusterv1.Cluster, awsMachinePool *expinfrav1.AWSMachinePool) (scope.EC2Scope, error) {
+func (r *AWSMachinePoolReconciler) getInfraCluster(ctx context.Context, log *logger.Logger, cluster *clusterv1.Cluster, awsMachinePool *expinfrav1.AWSMachinePool) (scope.EC2Scope, scope.S3Scope, error) {
```
fiunchinho, line 626: It looks like there is no case where the first and second returned values are different. They are always both the same value, either `nil` or `clusterScope` or `managedControlplaneScope`. If both always have the same value, couldn't we just return one instead of two values?

exp/controllers/awsmachinepool_controller.go
AndiDog: It gets cast to two different interfaces (`IgnitionScope` and `LaunchTemplateScope`). The number of small interfaces is to blame here, but many of them are needed to easily mock functionality or share functionality between regular and EKS cluster code and types.

exp/controllers/awsmachinepool_controller.go
AndiDog: See above comment regarding interfaces

fiunchinho (approved):
/lgtm

richardcase:
Lets also check the EKS pass

AndiDog:
Trying a simple rebase

fiunchinho:
everything is green! 🙂

AndiDog:
Rebased onto AWS SDK Go v2 migration

AndiDog:
I've manually re-tested this after the major migration to AWS SDK for Go v2.

damdo:
@fiunchinho can you re-add your LGTM here?

AndiDog:
E2E test should hopefully now be fixed (problem in AWS SDK Go v2 change), trying again

Issue #5107, opened 2024-08-26
Cluster deletion stuck due to AWSCluster finalizer issue


a13x5:
/kind bug

**What steps did you take and what happened:**

We have CAPI objects deployed using Helm chart. Thus when `helm uninstall <release>` is executed all objects get deleted simultaneously.

Sometimes finalizer on `AWSCluster` object (and object itself) is removed before all `AWSMachine` resources are removed. This is causing `AWSMachines` to stuck forever, because awsmachine controller tries to patch `AWSCluster` somehow.

```
I0726 16:19:31.225463       1 awscluster_controller.go:208] "Reconciling AWSCluster delete" controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="defau
lt/aws-cl-1" namespace="default" name="aws-cl-1" reconcileID="da967d9f-4c3d-47a1-953a-cedf44e4d8d0" cluster="default/aws-cl-1"
I0726 16:19:33.955431       1 securitygroups.go:320] "Deleted security group" controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="default/aws-cl-1" n
amespace="default" name="aws-cl-1" reconcileID="da967d9f-4c3d-47a1-953a-cedf44e4d8d0" cluster="default/aws-cl-1" security-group-id="sg-068b633aae83d2e19" kind="cluster managed"
I0726 16:19:34.432437       1 securitygroups.go:320] "Deleted security group" controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="default/aws-cl-1" n
amespace="default" name="aws-cl-1" reconcileID="da967d9f-4c3d-47a1-953a-cedf44e4d8d0" cluster="default/aws-cl-1" security-group-id="sg-05fe37ab8f0a3ab15" kind="cluster managed"
I0726 16:19:36.516438       1 vpc.go:550] "Deleted VPC" controller="awscluster" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSCluster" AWSCluster="default/aws-cl-1" namespace="default" nam
e="aws-cl-1" reconcileID="da967d9f-4c3d-47a1-953a-cedf44e4d8d0" cluster="default/aws-cl-1" vpc-id="vpc-03b7241ad6eae9ab1"
[...]

PR #5365, opened 2025-02-27
✨ Wait for AWSCluster dependent object to be deleted

alexander-demicev:
**What type of PR is this?**
/kind feature

This PR adds a similar check https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/4e912b4e4d1f855abf9b5194acaf9f31b5763c57/controlplane/eks/controllers/awsmanagedcontrolplane_controller.go#L394 for `AWSCluster`. It ensures all machines are gone before proceeding with deletion.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5107

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -58,6 +58,10 @@ import (
 	"sigs.k8s.io/cluster-api/util/predicates"
 )
 
+const (
+	deleteRequeueAfter = 20 * time.Second
+)
+
 var defaultAWSSecurityGroupRoles = []infrav1.SecurityGroupRole{
 	infrav1.SecurityGroupAPIServerLB,
 	infrav1.SecurityGroupLB,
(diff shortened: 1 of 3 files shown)


richardcase:
Thanks for this @alexander-demicev . Looks good to me. Lets check that the e2e still pass.

controllers/awscluster_controller.go
```diff
@@ -191,21 +195,34 @@ func (r *AWSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
...
 	clusterScope.Info("Reconciling AWSCluster delete")
 
+	numDependencies, err := r.dependencyCount(ctx, clusterScope)
+	if err != nil {
+		clusterScope.Error(err, "error getting AWSCluster dependencies")
+		return reconcile.Result{}, err
+	}
+
+	if numDependencies > 0 {
+		clusterScope.Info("AWSCluster cluster still has dependencies - requeue needed", "dependencyCount", numDependencies)
```
damdo, line 220: To provide a bit more context we can say what sort of dependencies we are talking about here, as they could be confused with cloud resources.
```suggestion
		clusterScope.Info("AWSCluster still has dependent CAPI resources - requeuing", "dependencyCount", numDependencies)
```

damdo: Could we maybe list in the k/v of this log like, what CAPI resources are still depending on the AWSCluster, e.g. instead of the count report `nameSpace/name` of the resources, or would that be too much/too verbose? What do you think Alex?

damdo (commented):
Looks good, left a couple of non-blocking nits on the log lines.
Thanks for fixing this @alexander-demicev

controllers/awscluster_controller.go
richardcase: Having the list would be helpful but as it stands its consistent with the same check that is done on the EKS side. I'd say lets follow up on logging the list of resources.

damdo: Yeah happy to keep it with just the count in this one 👍

damdo (commented):
/lgtm

damdo:
Thanks @alexander-demicev 

/assign @richardcase @nrb

Issue #3166, opened 2022-02-07
Add AWSManagedControlPlaneTemplate type


sedefsavas:
Add a new CRD named `AWSManagedControlPlaneTemplate`, that will be used with `ClusterClass`.

This should have the same webhooks with `AWSManagedControlPlane`.
(Example `AWSClusterTemplate` PR: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/2585)

/kind feature

pydctw:
Adding clusterclass for EKS is blocked until we have a resolution on https://github.com/kubernetes-sigs/cluster-api/issues/6126

richardcase:
This will be possible after v2.0.0 and specifically after #3797 merges.

@pydctw would you still like to work on this or is this something i can pick up?

pydctw:
Hi @richardcase, do we plan to support ClusterClass for EKS after we re-introduce AWSManagedCluster? Asking because ClusterClass is not yet supported for MachinePool (https://github.com/kubernetes-sigs/cluster-api/issues/5991).

If we decide to support ClusterClass for EKS MachineDeployment flavor, I can work on it.

PR #5375, opened 2025-03-06
✨  Add ClusterClass support for EKS clusters

alexander-demicev:
**What type of PR is this?**
/kind feature

This PR adds ClusterClass support for EKS.
- It doesn't support managed machine pools for now, I will add them in a follow up
- I'm unsure which fields in AWSManagedControlPlane can be reused in AWSManagedControlPlaneTemplate so please correct me if I'm wrong.
- Moving some fields to a nested structure in `AWSManagedControlPlane` broke conversions and they have to be done manually.

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/3166

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Add ClusterClass support for EKS clusters
```


richardcase (commented):
This looks good @alexander-demicev .

Can you add an e2e test to cover creating a cluster with ClusterClass?

alexander-demicev:
@richardcase yes, can I do it in a follow-up PR?

richardcase:
> @richardcase yes, can I do it in a follow-up PR? I already have a draft locally

A follow-up is good, i think we'd want the e2e coverage before we release this though.....this feature will be really popular 😄

richardcase:
@alexander-demicev - theres some issue with the e2e on main.  One fix has gone on to main and we're just testing another and then it would be worth rebasing. Hopefully they will pass then.

richardcase:
Created a PR to your fork that should help: https://github.com/alexander-demicev/cluster-api-provider-aws/pull/1

api/v1beta2/awsmanagedclustertemplate_types.go
```diff
@@ -0,0 +1,66 @@
...
+type AWSManagedClusterTemplateResource struct {
+	Spec AWSManagedClusterTemplateResourceSpec `json:"spec"`
+}
+
+// AWSManagedClusterTemplateResourceSpec defines the desired state of an AWSManagedCluster.
+// This is equivalent to AWSManagedClusterSpec.
+type AWSManagedClusterTemplateResourceSpec struct {
+	// ControlPlaneEndpoint represents the endpoint used to communicate with the control plane.
+	// +optional
+	ControlPlaneEndpoint clusterv1.APIEndpoint `json:"controlPlaneEndpoint"`
```
richardcase, line 65: I would say that this shouldn't be in the ClusterClass template spec for `AWSManagedClusterTemplate`. The reasoning is that **ControlPlaneEndpoint** is set when the EKS cluster has been created and will just be set on the actual instance of `AWSManagedCluster`. Having it in the template means that user think they can set it which could cause issues.

But that then means  `AWSManagedClusterTemplate` would be empty.

alexander-demicev: Is it ok to keep the AWSManagedClusterTemplate empty?

config/crd/kustomization.yaml
```diff
@@ -36,6 +38,7 @@ patchesStrategicMerge:
 - patches/webhook_in_awsclusterroleidentities.yaml
 - patches/webhook_in_awsclustertemplates.yaml
 - patches/webhook_in_awsmanagedcontrolplanes.yaml
+- patches/webhook_in_awsmanagedcontrolplanetemplates.yaml
```
furkatgofurov7, line 41: do you need `patches/webhook_in_awsmanagedclustertemplates.yaml` as well?

furkatgofurov7 (commented):
Looks great, one tiny non-blocking question/comment

/lgtm

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go
```diff
@@ -208,69 +208,81 @@ func (r *AWSManagedControlPlane) validateEKSClusterNameSame(old *AWSManagedContr
...
 	if v.LessThan(minVersion) {
-		message := fmt.Sprintf("addons requires Kubernetes %s or greater", minAddonVersion)
-		allErrs = append(allErrs, field.Invalid(addonsPath, *r.Spec.Version, message))
+		message := fmt.Sprintf("addons require Kubernetes %s or greater", minAddonVersion)
+		allErrs = append(allErrs, field.Invalid(addonsPath, *eksVersion, message))
 	}
 
-	// validations for IPv6:
-	// - addons have to be defined in case IPv6 is enabled
-	// - minimum version requirement for VPC-CNI using IPv6 ipFamily is 1.10.2
```
Danil-Grigorev, line 266: Seems to me that this comment is worth to preserve.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go
```diff
@@ -321,157 +332,189 @@ func (r *AWSManagedControlPlane) validateIAMAuthConfig() field.ErrorList {
...
+	return validateNetwork("AWSManagedControlPlane", r.Spec.NetworkSpec, r.Spec.SecondaryCidrBlock, field.NewPath("spec"))
+}
+
+func validateNetwork(resourceName string, networkSpec infrav1.NetworkSpec, secondaryCidrBlock *string, path *field.Path) field.ErrorList {
 	var allErrs field.ErrorList
 
-	// If only `AWSManagedControlPlane.spec.secondaryCidrBlock` is set, no additional checks are done to remain
-	// backward-compatible. The `VPCSpec.SecondaryCidrBlocks` field was added later - if that list is not empty, we
-	// require `AWSManagedControlPlane.spec.secondaryCidrBlock` to be listed in there as well. This may allow merging
-	// the fields later on.
```
Danil-Grigorev, line 430: Same here - is this comment ok to remove?

Danil-Grigorev (commented):
I’m not a ClusterClass or AWS expert, but this change looks good to me.

/lgtm

controlplane/eks/api/v1beta1/conversion.go
```diff
@@ -21,25 +21,100 @@ import (
...
 	}
 
 	// Manually restore data.
 	restored := &ekscontrolplanev1.AWSManagedControlPlane{}
-	if ok, err := utilconversion.UnmarshalData(r, restored); err != nil || !ok {
+	if _, err := utilconversion.UnmarshalData(r, restored); err != nil {
+		return err
+	}
+
+	dst.Spec.IdentityRef = r.Spec.IdentityRef
```
richardcase, line 41: Normaly you only need to restore fields here that are new between two APIVersions (i.e. between v1beta1 & v1beta2). For fields that are the same they should be covered by the `autoConvert` and `Convert` functions.

Where you seeing issues?

richardcase: Or was it because of the `AWSManagedControlPlaneClassSpec` in the spec now?

richardcase (commented):
Looks good @alexander-demicev , just one query about the conversions (but i suspect there is nothing to change there).

controlplane/eks/api/v1beta1/conversion.go
alexander-demicev: Yes, I had to write it manually because of that

Danil-Grigorev: It seems to be a downside of the code generation, it should support inline nesting replacements. Personally I think the old diff `controlplane/eks/api/v1beta1/zz_generated.conversion.go` can be used as a copy/paste material, but as long as manual diff is the same (it seems to be), it looks ok.

richardcase:
From my side i think this is good to go.

/approve

Would be good to get one final look by another maintainer though if possible. @nrb @AndiDog do you have any cycles to have a look?

/hold

dlipovetsky:
I would like to know how this change affects an existing EKS cluster that does not use ClusterClass.

alexander-demicev:
@dlipovetsky it shouldn't affect existing clusters

templates/cluster-template-eks-clusterclass.yaml
```diff
@@ -0,0 +1,66 @@
+apiVersion: cluster.x-k8s.io/v1beta1
+kind: ClusterClass
```
jimmidyson, line 2: This isn't quite right, it doesn't even define a `Cluster` resource.

Please take a look at the [simple clusterclass template](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/templates/cluster-template-simple-clusterclass.yaml) and adapt accordingly. Note that you will need to use CC variables and patches for the template to be consumable as-is.

alexander-demicev: CAPA seems to be a bit of a special case here. Other providers don’t include a Cluster in their ClusterClass templates, and I think it’s better to leave that for users to define. I’m happy to add it if others feel it would be helpful, but it doesn’t seem to align with common practice.

jimmidyson: Hmm that's a very good point. However, I still think something is wrong with this `ClusterClass`. The `ClusterClass` should be agnostic of the `Cluster` that references it, yet the names of referenced templates in this `ClusterClass` include `${CLUSTER_NAME}` - shouldn't this be `CLUSTERCLASS_NAME` so that it is independent of the `Clusters` that reference it? e.g. `spec.controlPlane.ref.name`, `spec.controlPlane.infrastructure.ref.name`, `spec.workers.machineDeployments[0].template.bootstrap.ref.name`, spec.workers.machineDeployments[0].infrastructure.ref.name` all include `${CLUSTER_NAME}`. Otherwise you end up with a `ClusterClass` per `Cluster`, which seems unnecessary - the ClusterClass should provide variables and patches that make the `ClusterClass` reusable across Clusters.

alexander-demicev: I fixed these references, thanks for point it out

CharlieR-o-o-t:
Really looking forward to this change. Do you need any help (e.g. Go development assistance)?

api/v1beta2/awsmanagedclustertemplate_types.go
```diff
@@ -0,0 +1,60 @@
...
+	Items           []AWSManagedClusterTemplate `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&AWSManagedClusterTemplate{}, &AWSManagedClusterTemplateList{})
+}
+
+// AWSManagedClusterTemplateResource describes the data needed to create an AWSManagedCluster from a template.
+type AWSManagedClusterTemplateResource struct {
+	Spec AWSManagedClusterTemplateResourceSpec `json:"spec"`
```
jimmidyson: I think this should reference the `AWSManagedClusterSpec` to be templatable and patchable by topology mutation hooks.

```suggestion
	Spec AWSManagedClusterSpec `json:"spec"`
}
```

config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
```diff
@@ -2425,36 +2425,22 @@ spec:
...
-                  Supports substitutions for {{.BaseOS}} and {{.K8sVersion}} with the base
-                  OS and kubernetes version, respectively. The BaseOS will be the value in
-                  ImageLookupBaseOS or ubuntu (the default), and the kubernetes version as
-                  defined by the packages produced by kubernetes/release without v as a
-                  prefix: 1.13.0, 1.12.5-mybuild.1, or 1.17.3. For example, the default
-                  image format of capa-ami-{{.BaseOS}}-?{{.K8sVersion}}-* will end up
-                  searching for AMIs that match the pattern capa-ami-ubuntu-?1.18.0-* for a
-                  Machine that is targeting kubernetes v1.18.0 and the ubuntu base OS. See
-                  also: https://golang.org/pkg/text/template/
+                  ImageLookupFormat is the AMI naming format used to look up machine images when a machine does not specify an AMI.
```
jimmidyson: Not sure why the descriptions are being changed here - something to do with the refactoring in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5375/files#diff-7dd39fbde134e4f0211cac9cb830c36793209731cf02f3c728809c4c55a43f39R37?

controlplane/eks/api/v1beta2/awsmanagedcontrolplanetemplate_types.go
```diff
@@ -0,0 +1,56 @@
...
+	Items           []AWSManagedControlPlaneTemplate `json:"items"`
+}
+
+func init() {
+	SchemeBuilder.Register(&AWSManagedControlPlaneTemplate{}, &AWSManagedControlPlaneTemplateList{})
+}
+
+// AWSManagedControlPlaneTemplateResource describes the data needed to create an AWSManagedCluster from a template.
+type AWSManagedControlPlaneTemplateResource struct {
+	Spec AWSManagedControlPlaneTemplateResourceSpec `json:"spec"`
```
jimmidyson: Why can't this just use the original `AWSManagedControlPlaneClassSpec` without changing it?
```suggestion
	Spec AWSManagedControlPlaneClassSpec `json:"spec"`
```

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -34,181 +34,18 @@ const (
 
 // AWSManagedControlPlaneSpec defines the desired state of an Amazon EKS Cluster.
 type AWSManagedControlPlaneSpec struct { //nolint: maligned
+	AWSManagedControlPlaneClassSpec `json:",inline"`
```
jimmidyson: I don't think this is necessary to change, just reference the original `AWSManagedControlPlaneSpec` in the `AWSManagedControlPlaneTemplateResource`. This would prevent backwards incompatible Go API changes that can be seen in the changes being made in the tests.

controlplane/eks/api/v1beta2/types_class.go
```diff
@@ -0,0 +1,180 @@
...
+)
+
+// AWSManagedControlPlaneClassSpec defines the AWSManagedControlPlane properties that may be shared across several AWS managed control planes.
+type AWSManagedControlPlaneClassSpec struct {
+	// MachineTemplate contains information about how machines
+	// should be shaped when creating or updating a control plane.
+	// For the AWSManagedControlPlaneTemplate, this field is used
+	// only to fulfill the CAPI contract.
+	// +optional
+	MachineTemplate *AWSManagedControlPlaneTemplateMachineTemplate `json:"machineTemplate,omitempty"`
```
jimmidyson: I'm not sure why this is necessary? Can you link to the CAPI contract that nees to be fulfilled that requires this to be added?

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
jimmidyson: @alexander-demicev I've pushed https://github.com/alexander-demicev/cluster-api-provider-aws/pull/2 which includes reverting the existing API changes. I have tested locally and can create EKS clusters perfectly with this API via ClusterClass.

controlplane/eks/api/v1beta2/types_class.go
alexander-demicev: It's part of the control plane provider contract: https://cluster-api.sigs.k8s.io/developer/providers/contracts/control-plane#controlplane-machines.
CAPZ has the machine templates, I’m totally fine not having the machine template field here.

alexander-demicev:
@jimmidyson thanks a lot for the commit, I squashed it with mine

jimmidyson:
Thanks @alexander-demicev!

/lgtm

controlplane/eks/api/v1beta2/types_class.go
jimmidyson: It's not required here because the control plane nodes are not represented by `Machines`, as the note says:

> In case you are developing a control plane provider which uses a Cluster API Machine object to represent each control plane instance, following fields MUST be implemented in the ControlPlane spec.

jimmidyson:
@AndiDog @dlipovetsky @richardcase Can we remove the hold on this now please? LGTM and have tested it out succesfully. There's an e2e test too which is a bonus 😅

PR #5381, opened 2025-03-10
:book: Proposal for ROSA networking

mzazrivec:
**What type of PR is this?**

/kind documentation

This pull request contains a proposal for implementing provisioning of networking infrastructure for ROSA HCP clusters.

Fixes #

**Checklist**:

- [ ] squashed commits
- [x] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note

```

### DIFF
--- a/docs/proposal/20250224-rosa-networking.md
+++ b/docs/proposal/20250224-rosa-networking.md
@@ -1,0 +1,239 @@
+---
+title: Networking for ROSA HCP
+authors:
+  - "@mzazrivec"
+reviewers:
+  -
+creation-date: 2025-02-24
+last-updated: 2025-07-11
+status: provisional
+---
+
+# Networking for ROSA HCP
+
+## Table of Contents
+
+<!-- START doctoc generated TOC please keep comment here to allow auto update -->
+<!-- DON'T EDIT THIS SECTION, INSTEAD RE-RUN doctoc TO UPDATE -->
+
+- [Glossary](#glossary)
+- [Summary](#summary)
+- [Motivation](#motivation)
+  - [Goals](#goals)
+  - [Non-Goals/Future Work](#non-goalsfuture-work)
+- [Proposal](#proposal)
+  - [User Stories](#user-stories)
+    - [Functional Requirements](#functional-requirements)
+- [Alternatives](#alternatives)
+- [Upgrade Strategy](#upgrade-strategy)
+
+<!-- END doctoc generated TOC please keep comment here to allow auto update -->
+
+## Glossary
+
+Refer to the [Cluster API Book Glossary](https://cluster-api.sigs.k8s.io/reference/glossary.html).
+
+## Summary
+
+This proposal defines implementation of networking infrastructure in CAPA for ROSA Hosted Control Plane.
+
+## Motivation
+
+To be able to provision a new ROSA HCP kubernetes cluster using CAPA, one has to create and setup the underlying network infrastructure first: VPC, public and private subnets, internet gateway, routing tables for both subnets, elastic IP address.
+
+All of the above can be currently provisioned and configured via AWS CLI, AWS Management Console or Terraform. Motivation for this work is to be able to provision and configure the network infrastructure for ROSA HCP using CAPI.
+
+### Goals
+
+1. Implement a namespaced new custom resource `ROSANetwork` representing the networking stack for ROSA HCP.
+2. It will be possible to reference the new `ROSANetwork` resource from ROSA control plane resource
+3. Implement creation and deletion for the new `ROSANetwork` resource.
+4. Support the same networking scenarios as [ROSA CLI](https://github.com/openshift/rosa) using the same embeded AWS CloudFormation template that ROSA CLI uses.
+
+### Non-Goals/Future Work
+
+- Modify current networking code in AWS / EKS clusters.
+- Support custom CloudFormation template.
+
+## Proposal
+
+The goal of this proposal is to be able to provision the networking infrastructure required for a ROSA HCP cluster.
+
+[ROSA CLI](https://github.com/openshift/rosa) supports creation of the networking infrastructure for ROSA HCP and uses [AWS CloudFormation](https://aws.amazon.com/cloudformation/) template under the hood. The [CloudFormation template used by ROSA CLI](https://github.com/openshift/rosa/blob/master/cmd/create/network/templates/rosa-quickstart-default-vpc/cloudformation.yaml) allows to specify five parameters: CloudFormation stack name, AZ count or list of availability zones, region and CIDR block for the VPC. The created CloudFormation stack then contains a VPC, public and private subnets (each pair created in separate AZ), internet gateway attached to VPC, elastic IPs, NAT gateways, public and private routing tables and a security group.
+
+Adopting the CloudFormation template used by rosa-cli would mean that CAPA and the `ROSANetwork` custom resource would be relying on a mechanism that is know to work well and any changes or fixes implemented in ROSA CLI would be picked up automatically in CAPA.
+
+In practical terms, implementation of the proposal would mean:
+1. A new namespaced custom resource definition `ROSANetwork` in CAPA with five attributes: name, AZ count, list of availability zones, region and CIDR block for VPC. `availabilityZoneCount`, `availabilityZones`, `region` and `cidrBlock` will become the `spec` part of the new `ROSANetwork` type, name of the cloudformation stack will be the same as `metadata.name`.
+
+   `ROSANetwork` spec example:
+    ```
+    kind: ROSANetwork
+    metadata:
+      name: rosa-network-01
+      namespace: default
+    spec:
+      availabilityZoneCount: 3
+      region: us-west-2
+      cidrBlock: 10.0.0.0/16
+    ```
+
+   `ROSANetwork` spec example with specified availability zones:
+    ```
+    kind: ROSANetwork
+    metadata:
+      name: rosa-network-01
+      namespace: default
+    spec:
+      availabilityZones:
+      - us-west-2a
+      - us-west-2d
+      region: us-west-2
+      cidrBlock: 10.0.0.0/16
+    ```
+
+1. A new reconciler for the new custom resource, implementing creation and deletion. The reconciler will be using an existing [CloudFormation template from ROSA CLI](https://github.com/openshift/rosa/blob/master/cmd/create/network/templates/rosa-quickstart-default-vpc/cloudformation.yaml) and will use [AWS CloudFormation API](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudformation) to [create](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudformation#Client.CreateStack) and [delete](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudformation#Client.DeleteStack) the AWS CloudFormation stack.
+
+   Outputs and resources created in the cloudformation stack will be tracked under `status` of the `ROSANetwork` type. In particular, the `status` will contain the list of public and private subnets and availability zones, grouped together by the availability zones.
+
+   Example:
+   ```
+   kind: ROSANetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     subnets:
+     - availabilityZone: us-west-2a
+       privateSubnet: subnet-1d9f28ba992a83514
+       publicSubnet: subnet-0d9f28ba991b93514
+     - availabilityZone: us-west-2b
+       privateSubnet: subnet-2d7f58c09f1b43512
+       publicSubnet: subnet-2d7f18c09f1b43512
+     - availabilityZone: us-west-2c
+       privateSubnet: subnet-7d7e19c0af1f4d57f
+       publicSubnet: subnet-1d7e19c0af1c4c57f
+   ```
+
+   All resources created in the cloudformation stack will be tracked under `status.resources` array:
+   ```
+   kind: ROSANetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     resources:
+     - logicalId: AttachGateway
+       physicalId: IGW|vpc-0b3efe540b42d3561
+       reason: ""
+       resource: AWS::EC2::VPCGatewayAttachment
+       status: CREATE_COMPLETE
+     - logicalId: EC2VPCEndpoint
+       physicalId: vpce-0a361ac65e48031e5
+       reason: Resource creation Initiated
+       resource: AWS::EC2::VPCEndpoint
+       status: CREATE_IN_PROGRESS
+     - logicalId: EcrApiVPCEndpoint
+       physicalId: vpce-09f346abadcc09f61
+       reason: Resource creation Initiated
+       resource: AWS::EC2::VPCEndpoint
+       status: CREATE_IN_PROGRESS
+   ```
+   and will be reflecting the the values coming from [AWS CloudFormation API](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/cloudformation#Client.DescribeStackEvents) (`resource`, `logicalId`, `physicalId`, `reason` and `status`).
+
+   `status.conditions` of the `ROSANetwork` resource will be consistent with the CAPA conventions, example of a successful network stack creation:
+   ```
+   kind: ROSANetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     conditions:
+     - lastTransitionTime: "2025-07-11T13:51:40Z"
+       reason: Created
+       severity: Info
+       status: "True"
+       type: ROSANetworkReady
+   ```
+   Example of creation in progress:
+   ```
+   kind: ROSANetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     conditions:
+     - lastTransitionTime: "2025-07-11T13:51:40Z"
+       reason: Creating
+       severity: Info
+       status: "False"
+       type: ROSANetworkReady
+   ```
+   Example of failed network stack creation:
+   ```
+   kind: ROSANetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     conditions:
+     - lastTransitionTime: "2025-03-18T13:25:16Z"
+       status: "False"
+       type: ROSANetworkReady
+       severity: Error
+       reason: Failed
+   ```
+   Failed deletion:
+   ```
+   kind: ROSANetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     conditions:
+     - lastTransitionTime: "2025-03-18T13:25:16Z"
+       status: "False"
+       type: ROSANetworkReady
+       severity: Error
+       reason: DeletionFailed
+       message: ...
+   ```
+   
+1. Modifications in the ROSA control plane CRD & reconciler so that it would be possible to reference the `ROSANetwork` resource from control plane:
+   ```
+   kind: ROSAControlPlane
+     metadata:
+       name: hcp01-control-plane
+       namespace: default
+   spec:
+     rosaNetworkRef:
+       name: hcp01-rosa-network
+   ```
+   Should the ROSA control plane CR contain reference to ROSA network, the reconciler will read the AWS region, AZ and subnet ids parameters from the ROSA network CR. The ROSA control plane should also be validated through a webhook so that it does not contain both the reference to `ROSANetwork` and the subnet ids and / or availability zones.
+
+1. New tests.
+
+### User Stories
+
+1. As a CAPA user, I want to be able to provision the network infrastructure for ROSA HCP.
+2. As a CAPA user, I want to be able to use the provisioned network infrastructure in ROSA HCP control plane.
+3. As a CAPA user, I want to be able to delete the network infrastructure previously provisioned by CAPA.
+
+#### Functional Requirements
+
+1. Ability to create a new namespaced custom resource `ROSANetwork` with four attributes: name, AZ count, region and CIDR block for VPC.
+2. Reconciler implementing creation and deletion of the `ROSANetwork` resource.
+3. Ability to reference the new custom resource from ROSA HCP control plane.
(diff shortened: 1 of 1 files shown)


serngawy (changes requested):
Please add more details for the proposed CRD RosaNetwork. The proposal mention with `four attributes: name, AZ count, region and CIDR block for VPC` , what is the spec and status fields ?
- Is it namespace scope ? cluster scope ? 
- What is the status conditions ? ex; validate/phases/applied ?
- As you are using cloud formation, how do you propagate the errors from cloud formation template to the CR status ?
- ROSAControlPlane will refer to the RosaNetwork, What fields the RosaControlplane will read to figure out the subnet-id and zones
- How do you will handle deleting the RosaNetwork while there is a RosaControlPlane refer to it ?

mzazrivec:
> Please add more details for the proposed CRD RosaNetwork. The proposal mention with `four attributes: name, AZ count, region and CIDR block for VPC` , what is the spec and status fields ?
>

Yes, I added more details regarding the spec / status part of the new custom resource.

> * Is it namespace scope ? cluster scope ?

Namespaced. Added to the doc.

> * What is the status conditions ? ex; validate/phases/applied ?

This is the part I'm not 100% sure about, but I'd probably prefer being consistent with what CAPA already uses, e.g. something like:
```
status:
  conditions:
    - lastTransitionTime: "2025-03-19T16:29:24Z"
      status: "True"
      type: RosaNetworkReady
      message:
      reason:
```

> * As you are using cloud formation, how do you propagate the errors from cloud formation template to the CR status ?

Through the `status.resources` array. Added to the doc.

> * ROSAControlPlane will refer to the RosaNetwork, What fields the RosaControlplane will read to figure out the subnet-id and zones

Added to the doc.

> * How do you will handle deleting the RosaNetwork while there is a RosaControlPlane refer to it ?

On one hand it would make sense to prevent RosaNetwork deletion should it be referenced from an existing control plane, but not quite sure how to implement that without using some type of labeling or something else that would allow CAPA to see whether or not is the particular RosaNetwork being referenced or not. In short, I don't want to be implementing any checks in the initial implementation.

docs/proposal/2025-02-24-rosa-networking.md
```diff
@@ -0,0 +1,128 @@
...
+
+Adopting the CloudFormation template used by rosa-cli would mean that CAPA and the `RosaNetwork` custom resource would be relying on a mechanism that is know to work well and any changes or fixes implemented in ROSA CLI would be picked up automatically in CAPA.
+
+In practical terms, implementation of the proposal would mean:
+1. A new namespaced custom resource definition `RosaNetwork` in CAPA with four attributes: name, AZ count, region and CIRD block for VPC. These four attributes will become the `spec` part of the new `RosaNetwork` type.
+
+1. A new reconciler for the new custom resource, implementing creation and deletion. The reconciler will be using an existing [CloudFormation template from ROSA CLI](https://github.com/openshift/rosa/blob/master/cmd/create/network/templates/rosa-quickstart-default-vpc/cloudformation.yaml) and will use AWS CloudFormation API to create and delete the AWS CloudFormation stack.
+
+   Outputs and resources created in the cloudformation stack will be tracked under `status` of the `RosaNetwork` type. In particular, the `status` will have the following fields:
+   * publicSubnets: array of public subnet ids
```
serngawy, line 72: make it one list of ;
- availabilityZone
  privateSubnet
  publicSubnet

docs/proposal/2025-02-24-rosa-networking.md
```diff
@@ -0,0 +1,128 @@
...
+
+Adopting the CloudFormation template used by rosa-cli would mean that CAPA and the `RosaNetwork` custom resource would be relying on a mechanism that is know to work well and any changes or fixes implemented in ROSA CLI would be picked up automatically in CAPA.
+
+In practical terms, implementation of the proposal would mean:
+1. A new namespaced custom resource definition `RosaNetwork` in CAPA with four attributes: name, AZ count, region and CIRD block for VPC. These four attributes will become the `spec` part of the new `RosaNetwork` type.
```
serngawy, line 67: please add the go struct OR CR example for the rosaNetwork.spec

docs/proposal/2025-02-24-rosa-networking.md
```diff
@@ -0,0 +1,128 @@
...
+   Outputs and resources created in the cloudformation stack will be tracked under `status` of the `RosaNetwork` type. In particular, the `status` will have the following fields:
+   * publicSubnets: array of public subnet ids
+   * privateSubnets: array of private subnet ids
+   * availabilityZones: array of AZs
+
+   Resources created in the cloudformation stack will be tracked under `status.resources` array:
+   ```
+   status:
+     resources:
+       - resource: NATGateway1
```
serngawy, line 80: Is this will duplication for subnet-ids ? may we don't need the list above for (zone, pub-subnet, priv-subnet) ?

docs/proposal/2025-02-24-rosa-networking.md
```diff
@@ -0,0 +1,128 @@
...
+         id:
+         status: CREATE_IN_PROGRESS
+         reason: Eventual consistency check initiated
+       - resource: VPC
+         id: vpc-055edf3ebf27f18d6 
+         status: CREATE_COMPLETE
+         reason:
+   ```
+   and will be reflecting the resource definitions from the CloudFormation template (`resource`) and the values coming from AWS CloudFormation API (`id`, `status` and `reason).
```
serngawy, line 89: please add the go struct OR CR example for the rosaNetwork.status and include the suggested conditions

serngawy (changes requested):

docs/proposal/2025-02-24-rosa-networking.md
mzazrivec: I don't understand the above. Do you want all the AZs, private and public subnets to be rolled into one array?

docs/proposal/2025-02-24-rosa-networking.md
mzazrivec: Yes, duplicated, as a convenience function.

docs/proposal/2025-02-24-rosa-networking.md
mzazrivec: Added

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,212 @@
...
+1. A new namespaced custom resource definition `RosaNetwork` in CAPA with four attributes: name, AZ count, region and CIDR block for VPC. `availability_zone_count`, `region` and `cidr_block` will become the `spec` part of the new `RosaNetwork` type, name of the cloudformation stack will be the same as `metadata.name`.
+
+   `RosaNetwork` spec example:
+    ```
+    kind: RosaNetwork
+    metadata:
+      name: rosa-network-01
+      namespace: default
+    spec:
+      availability_zone_count: 3
```
serngawy, line 76: Is there a reason for not specifying the regions? which regions will be used if the count is 2 ?

serngawy: Checking the template [here](https://github.com/openshift/rosa/blob/master/cmd/create/network/templates/rosa-quickstart-default-vpc/cloudformation.yaml#L7) can you clarify why is the max count is 3 ? some region like us-west-2 has 4 AZ.  
Lets have the availabilityZone as list of strings similar to [rosaControlPlane](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/api/v1beta2/rosacontrolplane_types.go#L100) 
It should be okay if we will need to change the template a bit to specify the zones

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,212 @@
...
+      region: us-west-2
+      cidr_block: 10.0.0.0/16
```
serngawy, line 78: cidrBlock

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,212 @@
...
+   * availabilityZones: array of AZs
+
+   Example:
+   ```
+   kind: RosaNetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     publicSubnets:
```
serngawy, line 95: These lists cannot tell which public, private, zone belong to each other, better to have it as below 
```suggestion
     subnets:
     - availabilityZone: us-west-2a
       privateSubnet: subnet-1d9f28ba992a83514
       publicSubnet: subnet-0d9f28ba991b93514
     - availabilityZone: us-west-2b
       privateSubnet: subnet-1d9f28ba992a83513
       publicSubnet: subnet-0d9f28ba991b93513    
```

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,212 @@
...
+   ```
+   
+1. Modifications in the ROSA control plane CRD & reconciler so that it would be possible to reference the `RosaNetwork` resource from control plane:
+   ```
+   kind: ROSAControlPlane
+     metadata:
+       name: hcp01-control-plane
+       namespace: default
+   spec:
+     RosaNetworkRef:
```
serngawy, line 182: rosaNetworkRef
```suggestion
     rosaNetworkRef:
```

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,212 @@
...
+       name: hcp01-rosa-network
+   ```
+   Should the ROSA control plane CR contain reference to ROSA network, the reconciler will read the AWS region, AZ and subnet ids parameters from the ROSA network CR.
```
serngawy, line 185: Lets also mention that end user cannot set both RosaControlPlane->spec->[subnets](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/api/v1beta2/rosacontrolplane_types.go#L95) and rosaNetworkRef at same time.

serngawy (changes requested):

docs/proposal/20250224-rosa-networking.md
mzazrivec: > Is there a reason for not specifying the regions? which regions will be used if the count is 2 ?

It is possible to specify the region. It is one of the input parameters of the CF template.

I'm guessing you meant AZs here, not the regions. It's not possible to specify the AZs, because that's how the CF template is written. You just specify the number of AZs you want to be used and they are being taken in the alphabetical order.

mzazrivec: > Checking the template [here](https://github.com/openshift/rosa/blob/master/cmd/create/network/templates/rosa-quickstart-default-vpc/cloudformation.yaml#L7) can you clarify why is the max count is 3 ? some region like us-west-2 has 4 AZ.

Don't know why 3 is the limit. But that's how the current template is.

> Lets have the availabilityZone as list of strings similar to [rosaControlPlane](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/api/v1beta2/rosacontrolplane_types.go#L100) It should be okay if we will need to change the template a bit to specify the zones

That would be beyond the scope of this proposal document and the initial implementation. It's also not just a matter of changing the CF template. This change would mean that the AZs are the `spec` of the CR, not the `status` of the CR. Or maybe even both the `spec` and `status` depending on whether you want to specify the AZ count or the AZ list.

It could potentially be a future enhancement, though I'm not entirely convinced that it would all that useful. In the multi-zone cluster installation you're probably more interested in the number of zones you're going to be using rather than their names, which is how the template is written now.

docs/proposal/20250224-rosa-networking.md
mzazrivec: Fixed.

docs/proposal/20250224-rosa-networking.md
mzazrivec: Technically, the user can specify both, but CAPA would prefer one to the other, right?

serngawy: yes, we will need to validate that in the webook raising error if user set both.

docs/proposal/20250224-rosa-networking.md
hunterkepley: I'm not sure what the end goal for the network model here is, if it's to be used with clusters like I assume, then 3 is the correct number for multi-AZ. You can either have 1 AZ for single-AZ cluster, or 3 for multi-AZ. There is validation in the API which only allows one of 2 options (1 or 3) for every cluster created:

```
// ValidateAvailabilityZonesCount is responsible for verifying whether the number of availability zones adheres to specific rules:
//
// * The number of availability zones for a multi AZ cluster should be 3.
```

This validation is performed for every cluster in which the user has selected AZs: `if isAvailabilityZonesSet || selectAvailabilityZones {`

mzazrivec: > To let end user has the ability to set the AZs its not beyond the proposal scope.
> 
> * As endUser using us-west-2 region it has 4 AZs, Why I cannot create subnets in all AZs (forced to 3 AZs) ? Is there a technical reason for that ?

That particular problem / question is not really related to this proposal nor can I really address it here.

> * As endUser us-east-2 region, I want to specify which AZ that I can create subnet to it ? Is there technical reason that I cannot specify AZ ? (The RosaControlPlane has the option to specify the AZs and subnets)

Well, technical reason .... There's a technical limitation right now: the CF template in rosa-cli does not allow it / support it. Things may change on that front for sure and it may be an enhancement (meaning both in rosa-cli & CAPA) in the future, but right now it's not possible.

docs/proposal/20250224-rosa-networking.md
mzazrivec: Added to the doc.

docs/proposal/2025-02-24-rosa-networking.md
mzazrivec: Changed the `status` to be an array of the subnets grouped by AZs.

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,221 @@
...
+1. A new namespaced custom resource definition `RosaNetwork` in CAPA with five attributes: name, AZ count, list of availability zones, region and CIDR block for VPC. `availabilityZoneCount`, `availabilityZones`, `region` and `cidrBlock` will become the `spec` part of the new `RosaNetwork` type, name of the cloudformation stack will be the same as `metadata.name`.
+
+   `RosaNetwork` spec example:
+    ```
+    kind: RosaNetwork
+    metadata:
+      name: rosa-network-01
+      namespace: default
+    spec:
+      availabilityZoneCount: 3
```
serngawy, line 76: since we have the availabilityZones, no need for availabilityZoneCount,  what do you think ?

mzazrivec: I think that as long the CF template we're using supports both scenarios, we can support both as well.

serngawy:
Thanks @mzazrivec,  looks good. You may mention as well during ROSANetwork deletion a validation must be applied to check if there is a RosaControlplane refer to it.

serngawy (approved):

PanSpagetka (approved):

docs/proposal/20250224-rosa-networking.md
nrb: Are the `availabilityZones` and `availibilityZoneCount` fields intended to be mutually exclusive?

Based on the two examples provided, I'm assuming they are, however the proposal should state this explicitly.

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,221 @@
...
+
+### User Stories
+
+1. As a CAPA user, I want to be able to provision the network infrastructure for ROSA HCP.
+2. As a CAPA user, I want to be able to use the provisioned network infrastructure in ROSA HCP control plane.
+3. As a CAPA user, I want to be able to delete the network infrastructure previously provisioned by CAPA.
+
+#### Functional Requirements
+
+1. Ability to create a new namespaced custom resource `RosaNetwork` with four attributes: name, AZ count, region and CIDR block for VPC.
```
nrb, line 206: This should probably note the AZ list, too.

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,221 @@
...
+       privateSubnet: subnet-1d9f28ba992a83514
+     - availabilityZone: us-west-2b
+       publicSubnet: subnet-2d7f18c09f1b43512
+       privateSubnet: subnet-2d7f58c09f1b43512
+     - availabilityZone: us-west-2c
+       publicSubnet: subnet-1d7e19c0af1c4c57f
+       privateSubnet: subnet-7d7e19c0af1f4d57f
+   ```
+
+   All resources created in the cloudformation stack will be tracked under `status.resources` array:
```
nrb, line 118: What's the reason for setting `subnets` as a top-level field instead of listing them under `status.resources.subnets`?

docs/proposal/20250224-rosa-networking.md
```diff
@@ -0,0 +1,221 @@
...
+   ```
+   kind: RosaNetwork
+   metadata:
+     name: rosa-network-01
+     namespace: default
+   status:
+     conditions:
+     - lastTransitionTime: "2025-03-20T14:45:26Z"
+       status: "True"
+       type: RosaNetworkReady
```
nrb, line 151: How will the `true` status be calculated for this? That all resources are in `CREATE_COMPLETE` from the AWS API, or something else?

nrb: I'll also mention the changes that happened in upstream CAPI with regards to the definition of `Ready`[1]. In this context, I believe it should mean "The network resources can pass traffic for workloads".

I don't think anything in this proposal is counter to what I just said, but I wanted to point that out for any other reviewers.

We should also try to adhere to the recommendations made in that CAEP. Since CAPA's conditions have not yet been updated, conditions defined by this CRD may not match the past conventions, but IMO we should use the new conventions when adding new resources.

[1] https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20240916-improve-status-in-CAPI-resources.md#readiness-and-availability

nrb:
I agree with the general direction this proposal takes. I made some comments around clarity and explicitness, paying special attention to `status.conditions` since there has been a significant effort upstream to align CAPI's usage with that of the Kubernetes project itself.

The examples given in YAML are very helpful! It may also be useful to include Go definitions in order to better understand optional or mutually exclusive fields, since these can often be excluded from YAML output due to the `omitEmpty` tag.

docs/proposal/20250224-rosa-networking.md
mzazrivec: @nrb Those 2 parameters are mutually exclusive in the sense that explicit AZs take priority over the AZ count. But the CF template we'll be using is written so that if you specify both, the stack provisioning succeeds (with explicit AZs taking priority). I can mention this in this in the proposal doc as well.

docs/proposal/20250224-rosa-networking.md
mzazrivec: @nrb yep, added to the proposal doc.

docs/proposal/20250224-rosa-networking.md
mzazrivec: @nrb in the end the subnets would be listed under both `status.subnets` (grouped by AZ) and `status.resources`. The `status.resources` will be a list filled with the data propagated from AWS as the stack creation progresses. My thought here was to duplicate the results we're interested in (subnets & AZs) so that we don't have to parse it out of the `status.resources` (i.e. a convenience function).

docs/proposal/20250224-rosa-networking.md
mzazrivec: > How will the `true` status be calculated for this? That all resources are in `CREATE_COMPLETE` from the AWS API, or something else?

AWS API informs the user / consumer when the whole stack is `CREATE_COMPLETE`.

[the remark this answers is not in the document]
nrb: If you can add the reason in here, I think that's the last of my comments.

docs/proposal/20250224-rosa-networking.md
nrb: So if I could rephrase to confirm I understand - the subnets are promoted to a top level resource primarily for ease of readability?

mzazrivec: @nrb Yes, I guess you could say that. Although I meant it mostly for the benefit of readibility by other controllers: AZs and subnet ids is what the ROSA controller needs to provision the actual cluster.

Do you think this is not a good idea?

docs/proposal/20250224-rosa-networking.md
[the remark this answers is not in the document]
mzazrivec: @nrb yep, ack :smiley:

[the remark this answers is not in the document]
mzazrivec: Not 100% sure if I understand the linked document correctly. The document says: `Use of the Reason field is required (currently in Cluster API reasons is added only when condition are false)`

Does that mean we can have an empty string and still show the `Reason` field? Or do we have to have a non-empty reason string even for success?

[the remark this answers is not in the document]
mzazrivec: @nrb Added the `reason` to the example for a successful creation.

mzazrivec:
I updated the proposal document slightly (mainly filed names and such) to reflect the implementation in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5464

damdo:
@nrb are you happy with this?

damdo:
@mzazrivec what's the status of this one?

mzazrivec:
@damdo The status of this PR is the same as it has been for the past year -- I addressed comments & suggestions and now I'm waiting for someone for either merge it or close it (ideally with an explanation why). Nonetheless, the actual implementation for what this PR proposes has been merged long time ago.

richardcase:
As this is purely for ROSA i am going to:

/approve

And then leave it to others to lgtm and unhold

/hold

damdo (commented):
/lgtm

the main implementation merged for this proposal, so we can considered it done and keep this as documentation.

/unhold

PR #5383, opened 2025-03-11
✨ Set Paused condition on reconciled resources status upon reconciliation being paused

theobarberbany:
**What type of PR is this?**
/kind api-change

edit (@damdo):
To follow the recent updates in the CAPI Provider contract for InfraMachines: https://github.com/kubernetes-sigs/cluster-api/pull/11275 we should set the `Paused=true` condition on the resource's status to acknowledge we detected the addition of the `cluster.x-k8s.io/paused` annotation on a resource and paused its reconciliation to fulfill the user's request.

AWSManagedCluster is handled in a separate PR as there are API changes involved. See #5394

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Set Paused condition on reconciled resources status upon pausing reconciliation (CAPI provider contract change)
```

### DIFF
--- a/bootstrap/eks/controllers/eksconfig_controller.go
+++ b/bootstrap/eks/controllers/eksconfig_controller.go
@@ -40,12 +40,12 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/v2/bootstrap/eks/internal/userdata"
 	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/logger"
+	"sigs.k8s.io/cluster-api-provider-aws/v2/util/paused"
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 	bsutil "sigs.k8s.io/cluster-api/bootstrap/util"
 	expclusterv1 "sigs.k8s.io/cluster-api/exp/api/v1beta1"
 	"sigs.k8s.io/cluster-api/feature"
 	"sigs.k8s.io/cluster-api/util"
-	"sigs.k8s.io/cluster-api/util/annotations"
 	"sigs.k8s.io/cluster-api/util/conditions"
 	"sigs.k8s.io/cluster-api/util/patch"
 	"sigs.k8s.io/cluster-api/util/predicates"
@@ -113,9 +113,8 @@ func (r *EKSConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 	log = log.WithValues("cluster", klog.KObj(cluster))
 
-	if annotations.IsPaused(cluster, config) {
-		log.Info("Reconciliation is paused for this object")
-		return ctrl.Result{}, nil
+	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, config); err != nil || isPaused || conditionChanged {
+		return ctrl.Result{}, err
 	}
 
 	patchHelper, err := patch.NewHelper(config, r.Client)
@@ -294,7 +293,7 @@ func (r *EKSConfigReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
 	b := ctrl.NewControllerManagedBy(mgr).
 		For(&eksbootstrapv1.EKSConfig{}).
 		WithOptions(option).
-		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetScheme(), logger.FromContext(ctx).GetLogger(), r.WatchFilterValue)).
+		WithEventFilter(predicates.ResourceHasFilterLabel(mgr.GetScheme(), logger.FromContext(ctx).GetLogger(), r.WatchFilterValue)).
 		Watches(
 			&clusterv1.Machine{},
 			handler.EnqueueRequestsFromMapFunc(r.MachineToBootstrapMapFunc),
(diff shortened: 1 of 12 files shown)


controllers/awscluster_controller.go
```diff
@@ -167,9 +168,8 @@ func (r *AWSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	log = log.WithValues("cluster", klog.KObj(cluster))
 
-	if capiannotations.IsPaused(cluster, awsCluster) {
-		log.Info("AWSCluster or linked Cluster is marked as paused. Won't reconcile")
-		return reconcile.Result{}, nil
+	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, cluster); err != nil || isPaused || conditionChanged {
+		return ctrl.Result{}, err
 	}
```
damdo, line 173: Maybe we could retain the log line, it is useful in case users are wondering what's going on.
We can talk about the level of logging where this should be printed. What is core CAPI doing in this case?

damdo: Oh I see it is embedded in the EnsurePausedCondition, so maybe we can keep only that instead?

util/paused/paused.go
```diff
@@ -0,0 +1,114 @@
...
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// This is a copy of the paused helpers from cluster-api, without the
+// v1beta2conditions requirement. Once the v1beta2conditions migration is
+// complete, we will want to use the upstream package.
```
damdo, line 19: Thanks for adding this!

damdo:
`failed to create new managed VPC: failed to create vpc: The maximum number of VPCs has been reached`

How often do we clean up the VPCs? cc. @richardcase

controllers/awscluster_controller.go
```diff
@@ -167,9 +168,8 @@ func (r *AWSClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 
 	log = log.WithValues("cluster", klog.KObj(cluster))
 
-	if capiannotations.IsPaused(cluster, awsCluster) {
-		log.Info("AWSCluster or linked Cluster is marked as paused. Won't reconcile")
-		return reconcile.Result{}, nil
+	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, cluster); err != nil || isPaused || conditionChanged {
```
nrb, line 171: I might be misreading this, but doesn't this mean that when a machine is unpaused, `conditionChanged` is still set to `true`, and reconciliation will be skipped? Are we expecting it to just requeue in that case?

Apologies if this is answered upstream, just want to be clear on the expected behavior.

theobarberbany: Good catch, I think that's intentional behaviour: https://github.com/kubernetes-sigs/cluster-api/pull/11284/files#r1800591168

richardcase:
> `failed to create new managed VPC: failed to create vpc: The maximum number of VPCs has been reached`
> 
> How often do we clean up the VPCs? cc. @richardcase

aws-janitor should run after we finish using the account. This could also be that we are using more VPCs in general in the tests.....but this is e2e.

However, this is the unit tests which shouldn't be created any VPCs and so its probably coming from the resource tracking code we have in the tests.

damdo:
Got it, thanks @richardcase , it looks like it is passing now :)

bootstrap/eks/controllers/eksconfig_controller.go
```diff
@@ -294,7 +293,7 @@ func (r *EKSConfigReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Man
 	b := ctrl.NewControllerManagedBy(mgr).
 		For(&eksbootstrapv1.EKSConfig{}).
 		WithOptions(option).
-		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetScheme(), logger.FromContext(ctx).GetLogger(), r.WatchFilterValue)).
+		WithEventFilter(predicates.ResourceHasFilterLabel(mgr.GetScheme(), logger.FromContext(ctx).GetLogger(), r.WatchFilterValue)).
```
richardcase, line 296: Subtle change here is that Reconcile will now be called when the EKSConfig is paused, and we'll rely on the "EnsurePaused" condition.

richardcase: No action needed, just stating the change.

controllers/awscluster_controller.go
richardcase: If `conditionChanged` then this won't immediately requeue, and we'll have to wait for the next resync period (or one of the other events). 

Do we want it to immediately requeue?

util/paused/paused.go
richardcase, line 19: It would be worth creating an issue for this so that we don't forget it.

controllers/awscluster_controller.go
theobarberbany: I'm not 100% I'm following here, I thought we would effectively requeue. We patch the condition, which would cause a new event? Or am I missing something :)

util/paused/paused.go
theobarberbany: I've created #5399  :)

controllers/awscluster_controller.go
richardcase: Sorry i was the one missing something, doh. Yep as its being patched then it will cause it to be requeue.

Ignore me :)

theobarberbany: Ok fab! :)

theobarberbany:
@richardcase Have squashed  :)

nrb:
/lgtm

Thanks Theo!

PR #5386, opened 2025-03-12
:bug: ROSA: Apply CAPI machinepool changes to ROSAMachinePool

PanSpagetka:
**What type of PR is this?**

/kind bug

Apply CAPI machinepool changes to ROSAMachinePool when ROSAMachinePool is not managed by (external) autoscaler.

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Apply CAPI machinepool changes to ROSAMachinePool

```

### DIFF
--- a/templates/cluster-template-rosa-machinepool.yaml
+++ b/templates/cluster-template-rosa-machinepool.yaml
@@ -57,7 +57,7 @@ metadata:
   name: "${CLUSTER_NAME}-pool-0"
 spec:
   clusterName: "${CLUSTER_NAME}"
-  replicas: 1
+  replicas: 3
   template:
     spec:
       clusterName: "${CLUSTER_NAME}"
--- a/exp/controllers/rosamachinepool_controller.go
+++ b/exp/controllers/rosamachinepool_controller.go
@@ -366,6 +366,14 @@ func (r *ROSAMachinePoolReconciler) reconcileMachinePoolVersion(machinePoolScope
 	return nil
 }
 
+func (r *ROSAMachinePoolReconciler) shouldUpdateRosaReplicas(machinePoolScope *scope.RosaMachinePoolScope, nodePool *cmv1.NodePool) bool {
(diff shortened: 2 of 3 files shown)


templates/cluster-template-rosa-machinepool.yaml
```diff
@@ -57,7 +57,7 @@ metadata:
   name: "${CLUSTER_NAME}-pool-0"
 spec:
   clusterName: "${CLUSTER_NAME}"
-  replicas: 1
+  replicas: 3
```
serngawy, line 60: We should add logic in the func [shouldUpdateRosaReplicas](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5386/files#diff-ca8ae33b4375b0269fe24617620c4d304d1882cb8d562c7b544331d272c7aec1R374) if the user set the replica to 1 in the machinePool saying ROSAMachinePool must have at least 3 replica per zone and machinePool replica changes will be ignored

serngawy (changes requested):
one comment and can you add the reference for the min ROSAMachinePool num of replicas per zone in the PR description.

templates/cluster-template-rosa-machinepool.yaml
PanSpagetka: Setting 1 replica is ok, if there isnt the only machine pool. If there is the only one machine pool, number of replicas will not go lower than 2. If you try to set it to 1, reconcile with end with an error. I think that is OK behavior.

serngawy: hmm okay then, since we don't add default MP most likely we will not have this issue.  We may re-visit this when we work on adding default MP.

PanSpagetka:
@nrb Could you please review once more? I made some changes to fix the tests

PR #5395, opened 2025-03-14
:sparkles: Allow disabling controllers

nrb:
/kind feature

This PR introduces the ability to disable controllers within the controller manager binary. It is an alternative to #5284 using a dedicated argument instead of the existing feature gates.

It starts with AWSMachine and AWSCluster as `unmanaged`, since those are the two we're interested in at the moment.

Other controllers that can be optional, such as the EKS, ROSA, and MachinePool ones, are currently managed with feature flags. When they graduate, they should be controlled by the disabled controllers flag.

Fixes #

Locally, this prints the options that can be passed in. Other output is omitted in this example.

```shell
❯ ./bin/manager --help
 Usage of ./bin/manager:
 ...
      --disable-controllers strings            Sets of controllers that should be disabled for this instance of the controller manager in a comma-separated list. Options are: "unmanaged"
...
```

And when enabled, the following log message appears:

```shell
❯ go run main.go --disable-controllers=awsmachine,awscluster -v 4 --logtostderr --stderrthreshold 0
❮ ./bin/manager --disable-controllers=unmanaged -v 4 --logtostderr
I0317 11:49:25.286786   74733 logger.go:75] "feature gates: \n" logger="setup"
I0317 11:49:25.286901   74733 logger.go:75] "controller disabled" logger="setup" controller="AWSMachine" congrollerGroup="unmanaged"
I0317 11:49:25.286905   74733 logger.go:75] "controller disabled" logger="setup" controller="AWSCluster" controller-group="unmanaged"
....
```

Note also that this is an alternative to #5284, with apologies to @mzazrivec and @serngawy. The original proposal did follow our established process and was accepted through lazy consensus. However, the maintainers would like to avoid using feature flags for GA functionality, even if it is code that is rapidly removed. Such addition and removal is a form of interface stability, and doing so would cause unnecessary churn, while leaving it as a feature flag indefinitely communicates the wrong status.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Introduce the `--disable-controllers` argument to the controller manager binary.
```

### DIFF
--- a/controllers/disabled_controllers.go
+++ b/controllers/disabled_controllers.go
@@ -1,0 +1,65 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package controllers
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TODO: EKS and ROSA are excluded from this list for the time being because they are behind feature gates.
+// They should be added to this list when they graduate.
+const (
+	Unmanaged = "unmanaged"
+)
+
+// disabledControllers tracks which controllers are disabled.
+// A value of `false` (default) means a controller is _enabled_.
+var disabledControllers = map[string]bool{
+	Unmanaged: false,
+}
+
+var notValidErr = "%q is not a valid controller name"
+
+// IsDisabled checks if a controller is disabled.
+// If the name provided is not in the map, this will return 'false'.
+func IsDisabled(name string) bool {
+	return disabledControllers[name]
+}
+
+// GetValidNames returns a list of controller names that are valid to disable.
+func GetValidNames() []string {
+	ret := make([]string, 0, len(disabledControllers))
+	for name := range disabledControllers {
+		ret = append(ret, name)
+	}
+	return ret
+}
+
+// ValidateNamesAndDisable validates a list of controller names against the known set, and disables valid names.
+func ValidateNamesAndDisable(names []string) error {
+	// This list is not de-deduplicated, so in someone could specify valid names multiple times.
+	for _, n := range names {
+		// Make sure we're doing a case-insensitive comaparison
+		name := strings.ToLower(n)
+		if _, ok := disabledControllers[name]; !ok {
+			return fmt.Errorf(notValidErr, name)
+		}
+		disabledControllers[name] = true
+	}
+	return nil
+}
--- a/controllers/disabled_controllers_unit_test.go
+++ b/controllers/disabled_controllers_unit_test.go
@@ -1,0 +1,84 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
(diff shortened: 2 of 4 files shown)


damdo (commented):
The proposal seems reasonable to me and the implementation looks good.
One typo

/lgtm

/assign @richardcase @dlipovetsky @AndiDog @Ankitasw @serngawy @mzazrivec

controllers/disabled_controllers_unit_test.go
```diff
@@ -0,0 +1,91 @@
...
+
+	. "github.com/onsi/gomega"
+)
+
+func TestIsDisabled(t *testing.T) {
+	g := NewWithT(t)
+	defer resetDefaults()
+
+	// Valid names default to false
+	g.Expect(IsDisabled(AWSClusterName)).To(BeFalse(), "awscluster should defautl to false")
```
damdo, line 31:
```suggestion
	g.Expect(IsDisabled(AWSClusterName)).To(BeFalse(), "awscluster should default to false")
```

controllers/disabled_controllers.go
```diff
@@ -0,0 +1,67 @@
...
+package controllers
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TODO: EKS and ROSA are excluded from this list for the time being because they are behind feature gates.
+// They should be added to this list when they graduate.
+const (
```
serngawy, line 26: I believe better to have it as one constant ex; AWScontrollers.

serngawy (commented):
Thanks @nrb, having 1 comment other than that looks good to me.

controllers/disabled_controllers.go
```diff
@@ -0,0 +1,67 @@
...
+
+// IsDisabled checks if a controller is disabled.
+// If the name provided is not in the map, this will return 'false'.
+func IsDisabled(name string) bool {
+	return disabledControllers[name]
+}
+
+// GetValidNames returns a list of controller names that are valid to disable.
+func GetValidNames() []string {
+	ret := make([]string, 0)
```
PanSpagetka, line 48: Even though practical difference is negligible, I think `ret := make([]string, 0, len(disabledControllers))` would be better.

controllers/disabled_controllers.go
```diff
@@ -0,0 +1,67 @@
...
+*/
+
+package controllers
+
+import (
+	"fmt"
+	"strings"
+)
+
+// TODO: EKS and ROSA are excluded from this list for the time being because they are behind feature gates.
```
richardcase, line 24: I think it would be fine to now remove the EKS feature flag if we want to.

controllers/disabled_controllers.go
richardcase: I also wonder if we need such fine-grained control. Do we just need 3 options ultimately:
- unmanaged (i.e. non-eks)
- managed (i.e. eks)
- rosa

docs/proposal/20250314-optional-controllers.md
```diff
@@ -0,0 +1,73 @@
+# Problem
+
```
richardcase, line 2: Might be nice to explicitly state this supersedes the other proposal in the front matter.

nrb:
@richardcase @damdo @serngawy @PanSpagetka Comments should be addressed in this latest push. I've squashed everything into 1.

controllers/disabled_controllers.go
nrb: Ok - I'd prefer to split that out so we can give people a heads up.

Maybe deprecate in 2.8 and remove in 2.10?

nrb: (I should re-read our official stance on deprecation)

PR #5398, opened 2025-03-17
✨ WIP: Initial dedicated hosts implementation

faermanj:
/kind feature

Let users allocate machines in dedicated hosts.

Dedicated hosts must be pre-allocated.
https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/dedicated-hosts-overview.html

**Checklist**:
- [X] squashed commits
- [ ] includes documentation
- [X] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```


nrb:
Could use one of the simplest e2e templates for this.

Will also need to have the test create an EC2 instance to use, and then release it. It would need to happen regardless of any test failures; might have to update AWS janitor

api/v1beta2/awsmachine_types.go
```diff
@@ -223,6 +223,16 @@ type AWSMachineSpec struct {
...
 	MarketType MarketType `json:"marketType,omitempty"`
+
+	// HostID specifies the Dedicated Host on which the instance should be launched.
+	// +optional
+	HostID *string `json:"hostId,omitempty"`
+
+	// Affinity specifies the dedicated host affinity setting for the instance.
+	// When affinity is set to Host, an instance launched onto a specific host always restarts on the same host if stopped.
+	// +optional
+	// +kubebuilder:validation:Enum:=Defailt;Host
```
rvanderp3, line 234:
```suggestion
	// +kubebuilder:validation:Enum:=Default;Host
```

test/e2e/shared/aws.go
```diff
@@ -2366,3 +2366,39 @@ func GetMountTargetState(e2eCtx *E2EContext, mountTargetID string) (*string, err
...
+// ///////////////////////////////////////////////////////////////////////////////
+func getAvailabilityZone() string {
+	return "us-east-1a"
+}
+
+func getInstanceType() string {
+	return "t3.large"
+}
+
+func GetDedicatedHost(e2eCtx *E2EContext) (string, error) {
```
nrb, line 2379: Since this is actually allocating, we might want to rename this function. Seeing `Get` makes me think it's just retrieving one that already exists.

PR #5418, opened 2025-03-19
🐛 fix: efs & elb upgrade e2e tests

richardcase:
**What type of PR is this?**

/kind failing-test

The EFS e2e test was breaking for 2 reasons:

1. Running out if disk space on the control plane nodes. It only had 8Gb so this has been increased to 16gb 
2. The workload being deployed to test EFS was using centos with has been discontinued for a long time now. So changed to use Ubuntu

The classic ELB upgrade test was failing because on upgrade of CAPA, the CAPA controller manager was restarted. From the logs i could see that the AWSCluster was no being reconciled and thats because the "old" and "new" versions were the same basic ion the predicate logic in SetupWithManager. Changed SetupWithManager to be more consistent with other CAPI infra providers. Also it was failing because of the wait introduced by the new `paused.EnsurePausedCondition` functionality but this is fixed in #5425

Fixes #

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Fix the EFS & classic elb e2e tests.
```

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -21,7 +21,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/types"
@@ -32,9 +31,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
-	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
-	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 	"sigs.k8s.io/controller-runtime/pkg/source"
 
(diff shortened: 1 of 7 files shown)


test/e2e/suites/unmanaged/helpers_test.go
```diff
@@ -539,7 +539,7 @@ func createPodWithEFSMount(clusterClient crclient.Client) {
 			Containers: []corev1.Container{
 				{
 					Name:    "app",
-					Image:   "centos",
+					Image:   "ubnuntu",
```
damdo, line 542:
```suggestion
					Image:   "ubuntu",
```

Typo here.

test/e2e/shared/aws_helpers.go
```diff
@@ -196,7 +196,7 @@ func CheckClassicElbHealthCheck(input CheckClassicElbHealthCheckInput, intervals
 		}
 
 		if *lb.HealthCheck.Target != input.ExpectedTarget {
-			return fmt.Errorf("health check target does not match expected target")
+			return fmt.Errorf("health check target %s does not match expected target %s", *lb.HealthCheck.Target, input.ExpectedTarget)
```
nrb, line 199: Minor nit: `%q` might be easier to read, since it puts quotes around the interpolated string.

richardcase: Changed

nrb:
@richardcase The release-2.7 periodics are failing on the same efs issue, so we should backport at least the YAML changes in this PR. Do you think we should also backport the changes to the manager setup?

richardcase:
> @richardcase The release-2.7 periodics are failing on the same efs issue, so we should backport at least the YAML changes in this PR. Do you think we should also backport the changes to the manager setup?

@nrb - we'll probably have to do it all. This issue started as just the EFS problem and then expanded to the classic elb issue. I can just cherry pick the EFS changes first and then if we see the other issue we can then cherry pick that commit. Will create the PR for release-2.7 soon.

nrb:
> we'll probably have to do it all

Yeah, we can just do a cherry-pick then

main.go
```diff
@@ -128,6 +128,7 @@ var (
 // +kubebuilder:rbac:groups=authorization.k8s.io,resources=subjectaccessreviews,verbs=create
 
 func main() {
+	setupLog.Info("starting cluster-api-provider-aws", "version", version.Get().String())
```
damdo, line 131: Nice addition, thanks!

controllers/awscluster_controller.go
```diff
@@ -393,28 +390,6 @@ func (r *AWSClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
...
-					oldCluster.Status = infrav1.AWSClusterStatus{}
-					newCluster.Status = infrav1.AWSClusterStatus{}
-
-					oldCluster.ObjectMeta.ResourceVersion = ""
-					newCluster.ObjectMeta.ResourceVersion = ""
-
-					return !cmp.Equal(oldCluster, newCluster)
-				},
-			},
-		).
```
damdo, line 417: Do we reckon that removing this will have any unrelated side effects?
For example, downstream we do externally manage the AWSCluster, will this create many reconciliations for the CAPA infracluster controller then?

richardcase: I don't think there will be too may side affets. It will result in more Reconcile ruins potentially but:

- Our original code is at odds with how other infrastructure providers work
- Keeping this would cause issues with the Paused condition being added and would still cause that initial delay in reconciliation
- If we kept this our change for ELB wouldn't be called until there was a "Spec" change.

damdo: Ok thanks, I think this is fine to remove for now.
If we notice any issues we can open a follow-up thread to discuss countermeasures.
Thanks!

damdo (commented):
/lgtm

/assign @nrb

PR #5422, opened 2025-03-20
:seedling: Retry boskos account checkouts

nrb:
**What type of PR is this?**

/kind support

Right now, our custom boskos client does not retry checking out an account if there are no resources available.

Generally this is not a problem, but during busy periods our tests could fail simply because there were no available hosts at that exact moment.

The [go implementation](https://github.com/kubernetes-sigs/boskos) uses defaults of 3 tries with 10 seconds between, so that is replicated here.

**Checklist**:

- [x] squashed commits
- [x] includes emoji in title 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/boskos.py
+++ b/hack/boskos.py
@@ -23,9 +23,13 @@ import time
 
 BOSKOS_HOST = os.environ.get("BOSKOS_HOST", "boskos")
 BOSKOS_RESOURCE_NAME = os.environ.get('BOSKOS_RESOURCE_NAME')
+# Retry up to 3 times, with 10 seconds between tries. This is the same as defaults
+# on https://github.com/kubernetes-sigs/boskos/
+MAX_RETRIES = 3
+RETRY_WAIT = 10
 
 
-def checkout_account(resource_type, user, input_state="free"):
+def checkout_account(resource_type, user, input_state="free", tries=1):
     url = f'http://{BOSKOS_HOST}/acquire?type={resource_type}&state={input_state}&dest=busy&owner={user}'
 
     r = requests.post(url)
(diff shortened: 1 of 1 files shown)


nrb:
/hold
until #5418 merges, since CI will just fail until then anyway.

hack/boskos.py
```diff
@@ -37,7 +41,15 @@ def checkout_account(resource_type, user, input_state="free"):
         print(f"export BOSKOS_RESOURCE_NAME={result['name']}")
         print(f"export AWS_ACCESS_KEY_ID={result['userdata']['access-key-id']}")
         print(f"export AWS_SECRET_ACCESS_KEY={result['userdata']['secret-access-key']}")
-
+    # The http API has two possible meanings of 404s - the named resource type doesn't exist or there no available resources of the type.
+    # For our purposes, we don't need to differentiate.
+    elif r.status_code == 404:
+        print(f"could not find available host, retrying in {RETRY_WAIT}s")
```
richardcase, line 47: The 404 may also be because the resource is not found as well.

richardcase: I think based on what we saw earlier this week. But i will check for sure.

hack/boskos.py
```diff
@@ -37,7 +41,15 @@ def checkout_account(resource_type, user, input_state="free"):
...
+        if retries == MAX_RETRIES:
+            raise Exception(f"could not allocate host after {MAX_RETRIES} tries")
+        tries = tries + 1
```
damdo, line 50: I think there is a tries vs retries naming issue here

hack/boskos.py
richardcase: It does also return a 404 if the resource is not found: https://github.com/kubernetes-sigs/boskos/blob/master/handlers/handlers.go#L78

hack/boskos.py
nrb: Doh, thanks!

hack/boskos.py
nrb: I'll clarify this comment, `named resource type doesn't exist` was supposed to cover this case

nrb:
@damdo @richardcase I think your comments have been addressed. It would be great to have this in before branch cut for convenience, but if not we can backport it since it's not critical to users.

richardcase:
Thanks for this @nrb . This will stop some classes of test flakes.

damdo (commented):
/lgtm

/unhold

since the hold criteria was https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5418 merging, which it did.

PR #5429, opened 2025-03-21
:book: Document latest release obstacles

nrb:
**What type of PR is this?**

/kind documentation

While releasing v2.8.1, we ran into some issues with building the image due to mismatched Go versions in the Google Cloud Builder image and the hack/tool dependencies.
This documents what to do in that scenario.

Also adds the nightly image push jobs to prerequisites to check before starting the release, so that hopefully we can avoid skipping the .0 again.

**Checklist**:

- [x] includes documentation
- [x] includes emoji in title 

**Release note**:
```release-note
NONE
```


damdo (commented):
/lgtm

Thanks for updating them!

PR #5430, opened 2025-03-21
[release-2.7] :bug: fix: efs e2e test breaking

nrb:
**What type of PR is this?**

/kind failing-test

The EFS e2e test was breaking for 2 reasons:

1. Running out if disk space on the control plane nodes. It only had 8Gb so this has been increased to 16gb 2.The workload being deployed to test EFS was using centos with has been discontinued for a long time now. So changed to use Ubuntu

Note: This is a backport from main (#5418 specifically), but omits the ELB healthcheck test updates, as those are for v2.8+

**Release note**:

```release-note
NONE
```


nrb:
Failed on what #5431 fixes

nrb:
Self merging to (hopefully) get CI stable and get some periodic signal before Monday and a potential 2.7.2 release.

/approve

PR #5452, opened 2025-03-30
🌱 Fix envsubstr installation in Devbox

### LINKED ISSUE #5423: Devbox setup writes to /usr/bin
/kind bug

With the new Devbox changes it downloads and installs envsubst into /usr/bin. This is fine for when devbox is used within a devcontainer but not good for anyone that uses devbox on their machine directly.

```
Replaced existing envsubst with the new one at /usr/bin/envsubst
```

I'd expect envsubst to be either:

1. Be installed via Nix
2. Or installed into `bin` folder in the project directory

**Environment:**

//cc @faermanj

spandan541:
**What type of PR is this?**

For users using devbox directly on their machines, envsubstr is now installed in the isolated devbox environment. Thus not conflicting with the GNU system installation.

Fix the below clusterawsadm installation error when re-entering the devbox environment "ln: failed to create symbolic link '$INSTALL_DIR/clusterawsadm': File exists"

Fixes #5423

`-L` checks if symlink exists, `-e or -f` errors out because `clusterawsadm` does not exist initially when the devbox session is created.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/scripts/devbox-init-hook.sh
+++ b/scripts/devbox-init-hook.sh
@@ -2,29 +2,16 @@
 DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SRC_DIR="$(dirname $DIR)"
 
-# Set install path
-INSTALL_DIR="$HOME/.local/bin"
-mkdir -p "$INSTALL_DIR"
+# Set install path to devbox packages directory
+INSTALL_DIR="$DEVBOX_PACKAGES_DIR/bin"
 
-# Check if envsubst.old exists
-if [[ ! -f "/usr/bin/envsubst.old" ]]; then
-    echo "envsubst.old does not exist. Proceeding with installation..."
-
+if [[ ! -f "$INSTALL_DIR/envsubst" ]]; then
     ENVSUBST_VERSION="v1.4.3"
     URL="https://github.com/a8m/envsubst/releases/download/$ENVSUBST_VERSION/envsubst-$(uname -s)-$(uname -m)"
     echo "Downloading: $URL"
-
-    curl -sL $URL -o $INSTALL_DIR/envsubst
-    chmod +x "$INSTALL_DIR/envsubst"
-
-    # Replace existing envsubst if present
-    if [[ -f "/usr/bin/envsubst" ]]; then
-        sudo mv /usr/bin/envsubst /usr/bin/envsubst.old
-        sudo ln -s "$INSTALL_DIR/envsubst" /usr/bin/envsubst
-        echo "Replaced existing envsubst with the new one at /usr/bin/envsubst"
-    else
-        touch /usr/bin/envsubst.old
-    fi
+    echo "Installing github.com/a8m/envsubst into Devbox shell"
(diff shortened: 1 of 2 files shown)


scripts/devbox-init-hook.sh
```diff
@@ -2,29 +2,16 @@
...
 
-# Set install path
-INSTALL_DIR="$HOME/.local/bin"
-mkdir -p "$INSTALL_DIR"
-
-# Check if envsubst.old exists
-if [[ ! -f "/usr/bin/envsubst.old" ]]; then
-    echo "envsubst.old does not exist. Proceeding with installation..."
+# Set install path to devbox packages directory
+INSTALL_DIR="$DEVBOX_PACKAGES_DIR/bin"
```
richardcase, line 6: I would probably install envsubst and clusterawsadm into the `bin` folder of the project (like we do the tools installed from the makefile) instead of the devbox packages directory. Feels like that devbox package directory should be left to devbox.

scripts/devbox-init-hook.sh
```diff
@@ -2,29 +2,16 @@
...
-
-    # Replace existing envsubst if present
-    if [[ -f "/usr/bin/envsubst" ]]; then
-        sudo mv /usr/bin/envsubst /usr/bin/envsubst.old
-        sudo ln -s "$INSTALL_DIR/envsubst" /usr/bin/envsubst
-        echo "Replaced existing envsubst with the new one at /usr/bin/envsubst"
-    else
-        touch /usr/bin/envsubst.old
-    fi
+    echo "Installing github.com/a8m/envsubst into Devbox shell"
```
richardcase, line 12: We have a target for envsubst in the tools makefile that we could use instead: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/635e9934270e58096e3ea59d9b8b88c0e0a546f7/hack/tools/Makefile#L89

scripts/devbox-init-hook.sh
```diff
@@ -34,8 +21,7 @@ if [[ ! -f "/usr/bin/envsubst.old" ]]; then
 fi
 
 # Use build location by default
-if [[ ! -e "$INSTALL_DIR/clusterawsadm" ]]; then
+if [[ ! -L "$INSTALL_DIR/clusterawsadm" ]]; then
     echo "Linking [$SRC_DIR/bin/clusterawsadm] [$INSTALL_DIR/clusterawsadm]"
-    mkdir -p "$INSTALL_DIR"
-    ln -s "$SRC_DIR/bin/clusterawsadm" $INSTALL_DIR/clusterawsadm
+    sudo ln -s "$SRC_DIR/bin/clusterawsadm" $INSTALL_DIR/clusterawsadm
```
richardcase, line 26: Similar to the previous points i wonder if we should just to `make clusterawsadm` to get the binary built into the `bin` folder?

richardcase (commented):
Thanks for this @spandan541 . Let me know if you have any questions about my comments.

PR #5471, opened 2025-04-18
✨ Support EKS upgrade policy

### LINKED ISSUE #5183: Support EKS upgrade policy
/kind feature

Currently when creating an EKS cluster using CAPA there is no way to set the upgrade policy. This means the default of "extended" is used that means thate lcuster will automatically enter extended support (with the extra cost that entails).

There is no way to specify "standard" as the upgrade policy with CAPA. 

**Environment:**

phuhung273:
**What type of PR is this?**
/kind feature

Support EKS upgrade policy

Fixes #5183

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Support EKS upgrade policy
```


phuhung273:
Hi @richardcase, can you please help me review this PR if you have time

richardcase (commented):
Thanks for this @phuhung273 .

Feel free to ping me on the issue or in slack if you want a hand with anything.

pkg/cloud/services/eks/cluster.go
```diff
@@ -725,6 +739,63 @@ func (s *Service) reconcileClusterVersion(cluster *eks.Cluster) error {
...
+
+	input := &eks.UpdateClusterConfigInput{
+		Name: aws.String(s.scope.KubernetesClusterName()),
+		UpgradePolicy: &eks.UpgradePolicyRequest{
+			SupportType: s.scope.ControlPlane.Spec.UpgradePolicy,
+		},
+	}
+
+	if err := wait.WaitForWithRetryable(wait.NewBackoff(), func() (bool, error) {
+		if _, err := s.EKSClient.UpdateClusterConfig(input); err != nil {
```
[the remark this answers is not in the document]
phuhung273: Sure, updated to use current UpdateClusterConfig call

phuhung273:
Hi @richardcase, if you have time, please help me review the latest update

phuhung273:
Hi @richardcase, can you help me review this

phuhung273:
Hi @richardcase, can you help review the change addressing your comments.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -209,6 +209,15 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
...
 	KubeProxy KubeProxy `json:"kubeProxy,omitempty"`
+
+	// The support policy to use for the cluster.
+	// Extended support indicates that the cluster will not be automatically upgraded
+	// when it leaves the standard support period, and will enter extended support.
+	// Clusters in extended support have higher costs.
+	// The default value is extended. Use standard to disable extended support.
+	// +kubebuilder:validation:Enum=extended;standard
+	// +optional
+	UpgradePolicy *UpgradePolicy `json:"upgradePolicy,omitempty"`
```
damdo: Does it have to be a pointer? Considering it is marked as optional If we omit it it'll be empty right?
Then in that case we can default it to extended.

richardcase: Yes agree. As this is a type alias for a string, we can do without the pointer.

pkg/cloud/services/eks/cluster.go
```diff
@@ -867,3 +908,10 @@ func clusterUpdatingStateRetryable(ctx context.Context, input *eks.DescribeClust
 	}
 	return true, nil
 }
+
+func convertUpgradePolicy(input ekscontrolplanev1.UpgradePolicy) ekstypes.SupportType {
```
richardcase: We sometimes but these in the "convert" package when converting  to/from SDK types.

richardcase (commented):
Looks good @phuhung273 .

It would also be worth adding a line to the docs so state that by default we create a cluster with "extended" support, perhaps somehwre here: https://cluster-api-aws.sigs.k8s.io/topics/eks/creating-a-cluster

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
phuhung273: In case current cluster is STANDARD, eg: by workaround since CAPA doesn't support yet. Defaulting to EXTENDED will suddenly switch all clusters to EXTENDED, increasing the cost.

There is a testcase for this situation: if the field is nil, don't do anything https://github.com/phuhung273/cluster-api-provider-aws/blob/00efef15c7cb5263a6409d2b998f8583d5e0d1bd/pkg/cloud/services/eks/cluster_test.go#L703-L710

Do you think it is a valid case ?

damdo: What is the default in the SDK for EKS if nothing is specified as a policy at the moment? If the default is `"standard"` then we should keep standard as default here IMO. And that wouldn't cause the issue you are talking about

phuhung273: By default:
- CreateCluster is EXTENDED
- UpdateCluster doesn't touch that param if not specified.

The case im mentioning is when user have other stuff managing clusters on top of CAPA.

pkg/cloud/services/eks/cluster.go
phuhung273: Sure, can see a similar converter package. I moved this function there.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
damdo: Ah ok so the default in GCP EKS platform is extended.
We could then have values for the Enum as: `Extended`, `Standard` and `PlatformDefault`. (let's use CamelCase enum values)
 
Where`PlatformDefault`  indicates that the Upgrade Policy is the default provided by the cloud platform.
And it our case it would mean we leave it empty when we specify it as the EKS input, and the GCP Platform internally picks Extended for us.

richardcase: I'm not sure about having the "PlatformDefault" option. If the AWS default is "extended" then we can set our default to be extended.

damdo: On a second thought, we can use Omitted here (instead of PlatformDefault), so `""` to signal we don't have a preference.
That way we don't pass it explicity (we omit it) to the EKS create input, and EKS will pick Extended for us.
I think it is nice to leave the default as is to avoid undoing/faffing with existing EKS cluster configurations.

richardcase: > In case current cluster is STANDARD, eg: by workaround since CAPA doesn't support yet.

What's the workaround? Setting the value in the AWS console?

We don't support scenarios where a user changes something manually in the console and guarantee that won't ben changed.

So i think we need to put a "action required" release note that states that by default clusters will be created using extended support and if you have manually changed it in the console to standard you will need to update your cluster definition to add "standard"

damdo: @richardcase I think we were concurrently writing these comments :) 
Just to check, are you happy with
`Extended`, `Standard` and omitted ("") (default were we don't pass it to the platform and it picks the value for us/keeps it untouched)? LMK TY

phuhung273: > Setting the value in the AWS console?

Yes this is one case

> put a "action required" release note that states that by default clusters will be created using extended support and if you have manually changed it in the console to standard you will need to update your cluster definition to add "standard"

I like this most. How do you think @damdo ?

damdo: I think that's potentially more disruptive than what I am suggesting, which will be a no-op for the current users, and will leave a way for the user to rely on AWS to choose what's best for them (which to me is the least surprising change). Let's see what Richard says.

richardcase: I'm in favour of anything that adheres to the "principle of least surprise".

Having an unset value (i.e. "" in this case) to mean    "Whatever the aws default" is fine, as long as they don't change that default (which they shouldn't). It also then puts the onus on the user to go look up what that default value is. CAPA is also supposed to be opinionated and it would be good for the API to communicate the expectations.

I think for new clusters we should explicitly state what the default is in our API.

damdo: Yes I agree. In that case we can add a comment that explains how CAPA will behave when the user omits the value or sets it to `""`.
i.e. It will let the AWS Platform choose whatever default value is there for the upgrade policy (at the time of writing Extended, but this may change in the future).

phuhung273: Seems like we have an agreement on: Extended, Standard and omitted (""). Thank you both.

> I think for new clusters we should explicitly state what the default is in our API.

Regarding to this point, we should explicitly set EXTENDED in `createCluster` if user ommit rite ?

damdo: > Regarding to this point, we should explicitly set EXTENDED in createCluster if user ommit rite ?

We should omit setting it in createCluster if the user omits it

phuhung273: Thanks for the confirmation, let me update the PR

phuhung273: I've updated to: Extended, Standard or omitted (""). PTAL @damdo @richardcase, thanks.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -208,6 +208,15 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
 
 	// KubeProxy defines managed attributes of the kube-proxy daemonset
 	KubeProxy KubeProxy `json:"kubeProxy,omitempty"`
+
+	// The support policy to use for the cluster.
+	// Extended support indicates that the cluster will not be automatically upgraded
+	// when it leaves the standard support period, and will enter extended support.
+	// Clusters in extended support have higher costs.
+	// If omitted, new clusters use AWS default extended support and existing clusters stay unchanged.
```
damdo:
```suggestion
	// If omitted, new clusters will use the AWS default upgrade policy (which at the time of writing is "extended") and existing clusters will have their upgrade policy unchanged.
```

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -208,6 +208,15 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
 
 	// KubeProxy defines managed attributes of the kube-proxy daemonset
 	KubeProxy KubeProxy `json:"kubeProxy,omitempty"`
+
+	// The support policy to use for the cluster.
```
damdo:
```suggestion
	// The cluster upgrade policy to use for the cluster.
	// (Official AWS docs for this policy: https://docs.aws.amazon.com/eks/latest/userguide/view-upgrade-policy.html)
```

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -208,6 +208,15 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
...
+	// Extended support indicates that the cluster will not be automatically upgraded
+	// when it leaves the standard support period, and will enter extended support.
+	// Clusters in extended support have higher costs.
```
damdo:
```suggestion
	// `extended` upgrade policy indicates that the cluster will enter into extended support once the Kubernetes version reaches end of standard support. You will incur extended support charges with this setting. You can upgrade your cluster to a standard supported Kubernetes version to stop incurring extended support charges.
	// `standard` upgrade policy indicates that the cluster is eligible for automatic upgrade at the end of standard support. You will not incur extended support charges with this setting but you EKS cluster will automatically upgrade to the next supported Kubernetes version in standard support.
```

docs/book/src/topics/eks/creating-a-cluster.md
```diff
@@ -14,6 +14,9 @@ clusterctl generate cluster capi-eks-quickstart --flavor eks-managedmachinepool
 
 NOTE: When creating an EKS cluster only the **MAJOR.MINOR** of the `-kubernetes-version` is taken into consideration.
 
+By default, EKS cluster uses:
+- EXTENDED support. See more info about [cluster upgrade policy](https://docs.aws.amazon.com/eks/latest/userguide/view-upgrade-policy.html)
```
damdo:
```suggestion
By default CAPA relies on the default EKS cluster upgrade policy, which at the moment of writing is EXTENDED support.
See more info about [cluster upgrade policy](https://docs.aws.amazon.com/eks/latest/userguide/view-upgrade-policy.html)
```

controlplane/eks/api/v1beta2/types.go
```diff
@@ -220,6 +220,24 @@ type AddonIssue struct {
 	ResourceIDs []string `json:"resourceIds,omitempty"`
 }
 
+// UpgradePolicy defines the support policy to use for the cluster.
+type UpgradePolicy string
+
+var (
+	// UpgradePolicyExtended indicates that the cluster will not be automatically upgraded
+	// when it leaves the standard support period, and will enter extended support.
+	// Clusters in extended support have higher costs.
```
damdo:
```suggestion
	// UpgradePolicyExtended indicates that the cluster will enter into extended support once the Kubernetes version reaches end of standard support.
	// You will incur extended support charges with this setting.
	// You can upgrade your cluster to a standard supported Kubernetes version to stop incurring extended support charges.
```

controlplane/eks/api/v1beta2/types.go
```diff
@@ -220,6 +220,24 @@ type AddonIssue struct {
...
+	UpgradePolicyExtended = UpgradePolicy("extended")
+
+	// UpgradePolicyStandard indicates that the cluster will be automatically upgraded
+	// when it leaves the standard support period.
```
damdo:
```suggestion
	// UpgradePolicyStandard indicates that the cluster is eligible for automatic upgrade at the end of standard support.
	// You will not incur extended support charges with this setting but you EKS cluster will automatically upgrade to the next supported Kubernetes version in standard support.
```

damdo (commented):
Let's update also these

phuhung273:
Thanks so much for your review @damdo. Let me know if we should start e2e.

pkg/cloud/services/eks/cluster.go
```diff
@@ -526,6 +537,12 @@ func (s *Service) reconcileClusterConfig(ctx context.Context, cluster *ekstypes.
 		input.ResourcesVpcConfig = updateVpcConfig
 	}
 
+	updateUpgradePolicy := s.reconcileUpgradePolicy(cluster.UpgradePolicy)
+	if updateUpgradePolicy != nil {
+		needsUpdate = true
+		input.UpgradePolicy = updateUpgradePolicy
+	}
+
```
damdo: We could make this inline to reduce variable scope.
```suggestion
	if updateUpgradePolicy := s.reconcileUpgradePolicy(cluster.UpgradePolicy); updateUpgradePolicy != nil {
		needsUpdate = true
		input.UpgradePolicy = updateUpgradePolicy
	}

```

pkg/cloud/services/eks/cluster.go
```diff
@@ -719,6 +736,31 @@ func (s *Service) reconcileClusterVersion(ctx context.Context, cluster *ekstypes
...
+	if strings.ToLower(string(clusterUpgradePolicy)) == s.scope.ControlPlane.Spec.UpgradePolicy.String() {
+		s.Debug("upgrade policy unchanged, no action")
+		return nil
+	}
+
+	return &ekstypes.UpgradePolicyRequest{
+		SupportType: converters.SupportTypeToSDK(s.scope.ControlPlane.Spec.UpgradePolicy),
+	}
```
damdo: Personally I think we can drop all these logs, not sure they add a lot of value.
In place of the logs we could add comments though on why we do that.

phuhung273: Sure, log was added because I want to inform when should we not update the cluster (the omitted case). But we're on the same page now. All good.

damdo:
Looks reasonable to me, let's run some tests.

/test ?

damdo:
/assign @richardcase @nrb @AndiDog @faiq 

Are you happy with the latest changes?

damdo:
/hold 

For testing

config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanetemplates.yaml
```diff
@@ -1022,6 +1022,17 @@ spec:
                         - iam-authenticator
                         - aws-cli
                         type: string
+                      upgradePolicy:
+                        description: |-
+                          The cluster upgrade policy to use for the cluster.
+                          (Official AWS docs for this policy: https://docs.aws.amazon.com/eks/latest/userguide/view-upgrade-policy.html)
+                          `extended` upgrade policy indicates that the cluster will enter into extended support once the Kubernetes version reaches end of standard support. You will incur extended support charges with this setting. You can upgrade your cluster to a standard supported Kubernetes version to stop incurring extended support charges.
+                          `standard` upgrade policy indicates that the cluster is eligible for automatic upgrade at the end of standard support. You will not incur extended support charges with this setting but you EKS cluster will automatically upgrade to the next supported Kubernetes version in standard support.
```
faiq: If clusters are going to automatically get updated there's going to be discrepancy between the AWSMangedControlPlane.Version. At the very least, we should update the status of the object. 🤔  I'm curious how this works right now with longer lasting clusters

config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
```diff
@@ -3196,6 +3196,17 @@ spec:
                 - iam-authenticator
                 - aws-cli
                 type: string
+              upgradePolicy:
+                description: |-
+                  The cluster upgrade policy to use for the cluster.
+                  (Official AWS docs for this policy: https://docs.aws.amazon.com/eks/latest/userguide/view-upgrade-policy.html)
+                  `extended` upgrade policy indicates that the cluster will enter into extended support once the Kubernetes version reaches end of standard support. You will incur extended support charges with this setting. You can upgrade your cluster to a standard supported Kubernetes version to stop incurring extended support charges.
+                  `standard` upgrade policy indicates that the cluster is eligible for automatic upgrade at the end of standard support. You will not incur extended support charges with this setting but you EKS cluster will automatically upgrade to the next supported Kubernetes version in standard support.
```
faiq: nit minor typo

```suggestion
                  `standard` upgrade policy indicates that the cluster is eligible for automatic upgrade at the end of standard support. You will not incur extended support charges with this setting but your EKS cluster will automatically upgrade to the next supported Kubernetes version in standard support.
```

faiq:
This looks mostly good to me although I have some questions about how automatic updates with "Standard" upgrade policy are going to be reflected -- they might already be handled, but I'm not sure.

config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanetemplates.yaml
phuhung273: Current logic in main branch shows that cluster will be untouched if its version is higher than the object. I think it is ok.
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/60649f3ec6af96f294b1fea8b899977646ed090e/pkg/cloud/services/eks/cluster.go#L684

Agree with updating object status. I'm thinking about `Ready: false` and failure message `EKS cluster was automatically upgraded because version 1.25 is out of standard support`. WDYT ?

faiq: :100:  setting some sort of failure would be good. in this case what would be the remediation simply to bump the version to the correct version, maybe we can add that as a part of the failure message?

phuhung273: Sure would be great to include the instruction, let me add it.

phuhung273: Added with this commit https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5471/commits/854b3d2e9d22a3edcad567c16b4e8ccf131fe292.

damdo:
Good points @faiq 
Also would it be reasonable to add an e2e test in the eks e2e suite that deals with the change 🤔 
Maybe going from extended to standard and check whether the .status.version was propagated correctly or something similar? cc. @richardcase as you may have thoughts

phuhung273:
> add an e2e test in the eks e2e suite that deals with the change 🤔 Maybe going from extended to standard and check whether the .status.version was propagated correctly

I love this idea. If we all agree let me give it a try.

richardcase:
> Maybe going from extended to standard and check whether the .status.version was propagated correctly or something similar?

Sounds good to me 👍

pkg/cloud/services/eks/cluster.go
```diff
@@ -207,6 +211,18 @@ func (s *Service) setStatus(cluster *ekstypes.Cluster) error {
 			conditions.MarkFalse(s.scope.ControlPlane, ekscontrolplanev1.EKSControlPlaneUpdatingCondition, "updated", clusterv1.ConditionSeverityInfo, "")
 			record.Eventf(s.scope.ControlPlane, "SuccessfulUpdateEKSControlPlane", "Updated EKS control plane %s", s.scope.KubernetesClusterName())
 		}
+		if s.scope.ControlPlane.Spec.UpgradePolicy == ekscontrolplanev1.UpgradePolicyStandard &&
+			(specSemver.Major < clusterSemver.Major || specSemver.Minor < clusterSemver.Minor) {
```
faiq: this will make the code slightly easier

```suggestion
			specSemver.LT(clusterSemver) {
```

faiq: Reference here https://pkg.go.dev/github.com/blang/semver#Version.LT

pkg/cloud/services/eks/cluster.go
```diff
@@ -207,6 +211,18 @@ func (s *Service) setStatus(cluster *ekstypes.Cluster) error {
...
+			s.scope.ControlPlane.Status.Ready = false
+			failureMsg := fmt.Sprintf(
+				"EKS control plane %s was automatically upgraded to version %s because %s is out of standard support. "+
+					"This can be fixed by bumping to the actual version of the cluster",
```
faiq:
```suggestion
					"This can be fixed by changing to the version of the AWSManagedControlPlane to the one reported in the status",
```

pkg/cloud/services/eks/cluster.go
faiq: Or maybe there's a reason you're ignoring the patch version here. If so please let me know! 

also wouldn't this return true in a case where specSemver == 2.1.0 and clusterSemver == 1.2.0

phuhung273: > wouldn't this return true in a case where specSemver == 2.1.0 and clusterSemver == 1.2.0

Really appreciate for this, my fault...

> Reference here https://pkg.go.dev/github.com/blang/semver#Version.LT

I don't think we should use this. Since user doesn't control patch version for managed cluster, using `less than` might show lots of false alerts. Let me just correct the logic here.

phuhung273:
Hi @damdo @faiq @richardcase, just found out that we cannot update cluster from EXTENDED to STANDARD if current version is out of standard support
```console
$ aws eks update-cluster-config --name cluster-4mxn5e_cluster-u37ohv-control-plane --upgrade-policy supportType=STANDARD

An error occurred (InvalidParameterException) when calling the UpdateClusterConfig operation: The Kubernetes version "1.30" is not in standard support. Please update the cluster to a version in standard support to change the supportType
```

This means we cannot e2e test cases like `going from extended to standard and check whether the .status.version was propagated correctly`.

phuhung273:
Thanks @faiq for understanding. @damdo let me know if you agree with e2e limitation `An error occurred (InvalidParameterException) when calling the UpdateClusterConfig operation: The Kubernetes version "1.30" is not in standard support. Please update the cluster to a version in standard support to change the supportType`

richardcase:
> This means we cannot e2e test cases like `going from extended to standard and check whether the .status.version was propagated correctly`.

@phuhung273  - but we should be able to do a test that is with a version that is in standard support and still switch the cluster from extended to standard upgrade policy.

phuhung273:
> but we should be able to do a test that is with a version that is in standard support and still switch the cluster from extended to standard upgrade policy

Absolutely let me add one.

phuhung273:
The new e2e works, let me know what you think @richardcase @damdo @faiq

test/e2e/suites/managed/eks_upgrade_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+		cluster := framework.GetClusterByName(ctx, framework.GetClusterByNameInput{
+			Getter:    e2eCtx.Environment.BootstrapClusterProxy.GetClient(),
+			Namespace: namespace.Name,
+			Name:      clusterName,
+		})
+		Expect(cluster).NotTo(BeNil(), "couldn't find cluster")
+
+		WaitForEKSClusterUpgradePolicy(ctx, e2eCtx.BootstrapUserAWSSession, eksClusterName, upgradePolicy)
+
+		upgradePolicy = ekscontrolplanev1.UpgradePolicyExtended
```
damdo:
```suggestion
		changedUpgradePolicy = ekscontrolplanev1.UpgradePolicyExtended
```

test/e2e/suites/managed/eks_upgrade_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+		ginkgo.By(fmt.Sprintf("should update upgrade policy to %s", upgradePolicy))
```
damdo:
```suggestion
		ginkgo.By(fmt.Sprintf("Changing the UpgradePolicy from %s to %s", upgradePolicy, changedUpgradePolicy))
```

test/e2e/suites/managed/eks_upgrade_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+		shared.SetEnvVar(shared.UpgradePolicy, upgradePolicy.String(), false)
```
damdo:
```suggestion
		shared.SetEnvVar(shared.UpgradePolicy, changedUpgradePolicy.String(), false)
```

test/e2e/suites/managed/eks_upgrade_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+		ManagedClusterSpec(ctx, getManagedClusterSpec)
+		WaitForEKSClusterUpgradePolicy(ctx, e2eCtx.BootstrapUserAWSSession, eksClusterName, upgradePolicy)
```
damdo:
```suggestion
		WaitForEKSClusterUpgradePolicy(ctx, e2eCtx.BootstrapUserAWSSession, eksClusterName, changedUpgradePolicy)
```

test/e2e/suites/managed/eks_upgrade_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+			ClusterProxy:         e2eCtx.Environment.BootstrapClusterProxy,
+			Cluster:              cluster,
+			ClusterctlConfigPath: e2eCtx.Environment.ClusterctlConfigPath,
+			ArtifactFolder:       e2eCtx.Settings.ArtifactFolder,
+		}, e2eCtx.E2EConfig.GetIntervals("", "wait-delete-cluster")...)
+	})
+})
+
+func WaitForEKSClusterUpgradePolicy(ctx context.Context, sess *aws.Config, eksClusterName string, upgradePolicy ekscontrolplanev1.UpgradePolicy) {
+	ginkgo.By(fmt.Sprintf("EKS control plane upgrade policy should be %s", upgradePolicy))
```
damdo:
```suggestion
	ginkgo.By(fmt.Sprintf("Checking EKS control plane upgrade policy matches %s", upgradePolicy))
```

test/e2e/suites/managed/eks_upgrade_policy_test.go
```diff
@@ -0,0 +1,119 @@
...
+	Eventually(func() (bool, error) {
+		cluster, err := getEKSCluster(ctx, eksClusterName, sess)
+		if err != nil {
+			return false, err
+		}
+		return converters.SupportTypeToSDK(upgradePolicy) == cluster.UpgradePolicy.SupportType, nil
```
damdo: Can we differentiate between failing to get the EKS cluster vs failing to match the expected upgrade policy change? TY

phuhung273: Yeah we can. In that case it would fall into the `if err != nil {}`

damdo: Ok so maybe something like this (untested)? WDYT?
```go
func WaitForEKSClusterUpgradePolicy(ctx context.Context, sess *aws.Config, eksClusterName string, upgradePolicy ekscontrolplanev1.UpgradePolicy) {
	By(fmt.Sprintf("Waiting for EKS control plane upgrade policy to be %s", upgradePolicy))

	Eventually(func() (bool, error) {
		cluster, err := getEKSCluster(ctx, eksClusterName, sess)
		if err != nil {
			var notFoundErr *types.ResourceNotFoundException
			if errors.As(err, &notFoundErr) {
				// Unrecoverable error stop trying and fail early.
				return false, StopTrying(fmt.Errorf("unrecoverable error: cluster %q not found: %w", eksClusterName, err))
			}
			return false, err // For transient errors, retry
		}

		expectedPolicy := converters.SupportTypeToSDK(upgradePolicy)
		actualPolicy := cluster.UpgradePolicy.SupportType

		if actualPolicy != expectedPolicy {
			// The upgrade policy change hasn't been reflected in EKS yet, error and try again.
			return false, fmt.Errorf("upgrade policy mismatch: expected %s, but found %s", expectedPolicy, actualPolicy)
		}

		// Success in finding the change has been reflected in EKS.
		return true, nil

	}, 5*time.Minute, 10*time.Second).Should(BeTrue(), "timed out waiting for the correct EKS upgrade policy")
}
```

phuhung273: Yeah looking good. Lemme test it.

phuhung273:
Thank you @damdo. But please wait im testing this https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5471#discussion_r2374715973

phuhung273:
Took 2hrs just to realize a nil check bug in helper function 😅 But here it is the negative test
Screenshot 2025-09-24 212151

test/e2e/suites/managed/eks_upgrade_policy_test.go
phuhung273: Done with a bit change. Let me know what you think

phuhung273:
Look like test infra issue

damdo:
@phuhung273 context for that error here: https://kubernetes.slack.com/archives/CD6U2V71N/p1758795545213209

phuhung273:
> Context for that error here: https://kubernetes.slack.com/archives/CD6U2V71N/p1758795545213209

This one is spot instance test so if it is flaky i'm not surprised 😅 So i think it make sense to retry.

But this run https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5471/pull-cluster-api-provider-aws-e2e/1971153449110212608 it failed 3 other cases, while the spot one you're referring passed.

damdo:
> > Context for that error here: https://kubernetes.slack.com/archives/CD6U2V71N/p1758795545213209
> 
> This one is spot instance test so if it is flaky i'm not surprised 😅 So i think it make sense to retry.
> 
> But this run https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5471/pull-cluster-api-provider-aws-e2e/1971153449110212608 it failed 3 other cases, while the spot one you're referring passed.

The underlying issue is the same one. If we fix that, we fix all of them. Let's keep chatting on that thread

phuhung273:
Just rebase to see if thing gets better.

damdo:
@phuhung273 it looks like in the latest EKS test run the e2e failed on the STANDARD to EXTENDED upgrade policy test: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5471/pull-cluster-api-provider-aws-e2e-eks/1971544462022676480

So it might be a bit flakey, would you be able to investigate what happened there? Thanks!

phuhung273:
I can see most success e2e-eks took almost 4hr, failed one exceeded a bit, so seems like 4h is the limit. Since we add a new test, maybe increasing [this 5h limit](https://github.com/kubernetes/test-infra/blob/7161286445b3d6f9d4efa9c9e26375f85b1236f3/config/jobs/kubernetes-sigs/cluster-api-provider-aws/cluster-api-provider-aws-presubmits-main.yaml#L405) may help.

Is https://github.com/kubernetes/test-infra/tree/master/config/jobs/kubernetes-sigs/cluster-api-provider-aws the correct repo @damdo ?

damdo:
Ah it could be that that's the reason yes @phuhung273 
@richardcase WDYT of bumping the EKS e2e tests timeout?
I don't think it is worth splitting that suite up in this PR, I think we could extend the timeout and then evaluate if we want to shard these somehow.

damdo:
@phuhung273 opened https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5690

phuhung273:
Timeout bump commit is not included currently. If e2e-eks timeout again, i will rebase.

phuhung273:
Even when all tests passed, SynchronizeAfter failed 😅

phuhung273:
It finally passed 😅

phuhung273:
Thank you also for the help @damdo

PR #5496, opened 2025-05-15
:bug: Launchtemplate needs to be updated if spot options are changed

fiunchinho:
**What type of PR is this?**

/kind bug

When the spot instances options are changed, we want to update the launch template so that the changes take effect.

I've changed the unit tests for the `LaunchTemplateNeedsUpdate` function, because some cases were passing by pure chance. Basically these security groups are always added to the incoming `AWSLaunchTemplate`
```
{ID: aws.String("sg-111")},
{ID: aws.String("sg-222")},
```

So if the existing `AWSLaunchTemplate` that we use in the tests don't have these security groups, the function will always return true. Tests like the one testing changes to the `IamInstanceProfile` field were passing because of the different security groups, rather than the change on the `IamInstanceProfile` field.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
LaunchTemplate needs to be updated if spot options are changed
```

### DIFF
--- a/pkg/cloud/services/ec2/launchtemplate.go
+++ b/pkg/cloud/services/ec2/launchtemplate.go
@@ -680,6 +680,24 @@ func (s *Service) deleteLaunchTemplateVersion(id string, version *int64) error {
 	return nil
 }
 
+// SDKToSpotMarketOptions converts EC2 instance market options to SpotMarketOptions.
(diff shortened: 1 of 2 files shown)


alexander-demicev (approved):
/lgtm

fiunchinho:
@alexander-demicev After testing the changes in our management cluster, I needed to do more changes to make this work. PTAL, thanks.

PR #5451, opened 2025-03-28
📖 ROSA HCP IAM Roles proposal

serngawy:
**What type of PR is this?**
/kind documentation

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ *] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
NONE
```

PR #5499, opened 2025-05-21
✨ Rosa Config implementaiton

PanSpagetka:
Based on proposal https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5451
Adding RosaRoleConfig API with implementation. that should create Account/Operator roles and OIDC config/provider necessary to create ROSA cluster.

We need to move RosaMachinePoolAutoScaling definition to controlplane, because otherwise there would be circular dependency.

**What type of PR is this?**
/kind feature

Fixes #

**Checklist**:

- [] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Add Rosa Role Config API and implementation
```


Dockerfile
```diff
@@ -28,12 +28,17 @@ WORKDIR /workspace
 # Copy the Go Modules manifests
 COPY go.mod go.mod
 COPY go.sum go.sum
+COPY ./rosa /workspace/rosa
```
serngawy: why are we adding this ? what is the rosa file

Dockerfile
```diff
@@ -28,12 +28,17 @@ WORKDIR /workspace
...
+
+
 # Cache deps before building and copying source so that we don't need to re-download as much
 # and so that source changes don't invalidate our downloaded layer
 RUN  --mount=type=cache,target=/root/.local/share/golang \
      --mount=type=cache,target=/go/pkg/mod \
      go mod download
 
+# RUN go mod download
```
serngawy: no need for this line

PROJECT
serngawy, line 1: can you just add the RosaRoleConfig item without changing the order of other items

config/default/tmp.yaml
serngawy, line 1: why do we need this file ?

config/crd/kustomization.yaml
```diff
@@ -38,6 +39,7 @@ patchesStrategicMerge:
 - patches/webhook_in_awsmanagedcontrolplanes.yaml
 - patches/webhook_in_eksconfigs.yaml
 - patches/webhook_in_eksconfigtemplates.yaml
+#- patches/webhook_in_rosaroleconfigs.yaml
```
serngawy: I believe you need to uncomment this line

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,579 @@
...
+		return ctrl.Result{}, r.reconcileDelete(scope, ocmClient)
+	}
+
+	if controllerutil.AddFinalizer(scope.RosaRoleConfig, expinfrav1.RosaRoleConfigFinalizer) {
+		if err := scope.PatchObject(); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if scope.RosaRoleConfig.Status.OIDCID == "" {
```
serngawy: I believe you should check/set the RosaRoleConfig condition first , then get  the oidc using OCM client if it is not exist then create it. 
Same applied for account-roles and operator-roles

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,579 @@
...
+
+	oidcID := scope.RosaRoleConfig.Status.OIDCID
+	if oidcID != "" {
+		err := r.deleteOIDCProvider(ocmClient, awsClient, oidcID)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = r.deleteOperatorRoles(ocmClient, awsClient, scope.RosaRoleConfig.Spec.AccountRoleConfig.Prefix)
```
serngawy: Do we need to delete the operator roles before the oidc-provider ?

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,579 @@
...
+		}
+	}
+	return nil
+}
+
+func (r *ROSARoleConfigReconciler) deleteOIDCConfig(ocmClient *ocm.Client, oidcConfigID string) error {
+	return ocmClient.DeleteOidcConfig(oidcConfigID)
+}
+
+type reporter struct {
```
serngawy: please move this to another file , better to be under pkg/.../rosa

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: Nope, a changed it so it matches reverse creation order.

serngawy:
@PanSpagetka would you fix the PR title so PR verify job can pass

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +183,24 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
 	return nil
 }
 
+func (r *ROSAControlPlane) validateRosaRoleConfig() *field.Error {
```
serngawy, line 205: Not sure if the logic here is valid, hasDirectRoleFields will be true even if one condition is true and others are false. Ex; r.Spec.OIDCID can be true but all others fields can be false that case hasDirectRoleFields will be true.

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +183,24 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
...
+	hasRosaRoleConfigRef := r.Spec.RosaRoleConfigRef != nil
+	hasDirectRoleFields := r.Spec.OIDCID != "" || r.Spec.InstallerRoleARN != "" || r.Spec.SupportRoleARN != "" || r.Spec.WorkerRoleARN != "" ||
+		r.Spec.RolesRef.IngressARN != "" || r.Spec.RolesRef.ImageRegistryARN != "" || r.Spec.RolesRef.StorageARN != "" ||
+		r.Spec.RolesRef.NetworkARN != "" || r.Spec.RolesRef.KubeCloudControllerARN != "" || r.Spec.RolesRef.NodePoolManagementARN != "" ||
+		r.Spec.RolesRef.ControlPlaneOperatorARN != "" || r.Spec.RolesRef.KMSProviderARN != ""
+
+	if hasRosaRoleConfigRef && hasDirectRoleFields {
```
serngawy: Lets make the logic as below;
if RosaRoleConfigRef is set, we use the roleConfigRef to get all roles (ignore all other role fields even if set) just log warning.
if RosaRoleConfigRef not set and all other Role fields are set, we use the roles fields
if RosaRoleConfigRef not set and some Roles fields are missing we raise error
if RosaRoleConfigRef not set and all Roles fields are missing we raise error

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -226,6 +228,48 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
+				rosacontrolplanev1.ROSARoleConfigReadyCondition,
+				rosacontrolplanev1.ROSARoleConfigNotReadyReason,
+				clusterv1.ConditionSeverityWarning,
+				"RosaRoleConfig %s/%s is not ready", rosaScope.ControlPlane.Namespace, rosaScope.ControlPlane.Spec.RosaRoleConfigRef.Name)
+			return ctrl.Result{}, fmt.Errorf("RosaRoleConfig %s/%s is not ready", rosaScope.ControlPlane.Namespace, rosaScope.ControlPlane.Spec.RosaRoleConfigRef.Name)
+		}
+
+		conditions.MarkTrue(rosaScope.ControlPlane, rosacontrolplanev1.ROSARoleConfigReadyCondition)
+
+		// Update spec fields from RosaRoleConfig
```
serngawy: This logic is not correct, If the user is already setting those fields value, we are updating those values. I believe we should define internal roleConfig in rosacontrolPlane scope and then check which roles to use based on availability

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,183 @@
...
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ROSARoleConfigSpec defines the desired state of ROSARoleConfig
+type ROSARoleConfigSpec struct {
+	AccountRoleConfig  AccountRoleConfig             `json:"accountRoleConfig"`
+	OperatorRoleConfig OperatorRoleConfig            `json:"operatorRoleConfig"`
+	OIDCConfig         OIDCConfig                    `json:"oidcConfig"`
+	IdentityRef        *infrav1.AWSIdentityReference `json:"identityRef,omitempty"`
+	Region             string                        `json:"region,omitempty"`
```
serngawy: Missing description and kbuilder tags.

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,183 @@
...
+	//
+	// At most one provider can be configured.
+	//
+	// +listType=map
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=1
+	ExternalAuthProviders []rosacontrolplanev1.ExternalAuthProvider `json:"externalAuthProviders,omitempty"`
+	IdentityRef           *infrav1.AWSIdentityReference             `json:"identityRef,omitempty"`
+	Region                string                                    `json:"region,omitempty"`
+	Prefix                string                                    `json:"prefix"`
```
serngawy: Missing description and kbuilder tags.

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,183 @@
...
+	// +optional
+	SharedVPCConfig SharedVPCConfig `json:"sharedVPCConfig,omitempty"`
+}
+
+// OperatorRoleConfig defines cluster-specific operator IAM roles based on your cluster configuration.
+type OperatorRoleConfig struct {
+	//  User-defined prefix for generated AWS operator policies.
+	// +kubebuilder:validation:MaxLength:=4
+	// +kubebuilder:validation:Required
+	Prefix string `json:"prefix"`
```
serngawy, line 85: Prefix must be immutable cause changing the prefix will create new role

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,183 @@
...
+	Spec   ROSARoleConfigSpec   `json:"spec,omitempty"`
+	Status ROSARoleConfigStatus `json:"status,omitempty"`
+}
+
+// AccountRoleConfig defines account-wide IAM roles before creating your ROSA cluster.
+type AccountRoleConfig struct {
+	// User-defined prefix for all generated AWS resources
+	// +kubebuilder:validation:MaxLength:=4
+	// +kubebuilder:validation:Required
+	Prefix string `json:"prefix"`
```
serngawy, line 58: Prefix must be immutable cause changing the prefix will create new roles

serngawy (changes requested):

exp/api/v1beta2/rosaroleconfig_webhook.go
```diff
@@ -0,0 +1,48 @@
...
+	"context"
+
+	runtime "k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/webhook"
+	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+)
+
+// SetupWebhookWithManager will setup the webhooks for the ROSARoleConfig.
+func (r *ROSARoleConfig) SetupWebhookWithManager(mgr ctrl.Manager) error {
```
mzazrivec, line 13: This routine is never called from `main.go`, which is why I'm getting:
```
$ kubectl apply -f rosa-roleconfig-01.yaml
Error from server (InternalError): error when creating "rosa-roleconfig-01.yaml": Internal error occurred: failed calling webhook "default.rosaroleconfig.infrastructure.cluster.x-k8s.io": failed to call webhook: the server could not find the requested resource
```

With this change in place the above error goes away:
```
$ git diff
diff --git a/main.go b/main.go
index c65ff7356..6348fd0de 100644
--- a/main.go
+++ b/main.go
@@ -285,6 +285,11 @@ func main() {
                        setupLog.Error(err, "unable to create webhook", "webhook", "ROSAMachinePool")
                        os.Exit(1)
                }
+
+               if err := (&expinfrav1.ROSARoleConfig{}).SetupWebhookWithManager(mgr); err != nil {
+                       setupLog.Error(err, "unable to create webhook", "webhook", "ROSARoleConfig")
+                       os.Exit(1)
+               }
        }
 
        if err = (&expcontrollers.ROSARoleConfigReconciler{
```

mzazrivec: Although I don't quite understand the point of this webhook, since it's currently not doing anything.

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,183 @@
...
+type ROSARoleConfigSpec struct {
+	AccountRoleConfig  AccountRoleConfig             `json:"accountRoleConfig"`
+	OperatorRoleConfig OperatorRoleConfig            `json:"operatorRoleConfig"`
+	OIDCConfig         OIDCConfig                    `json:"oidcConfig"`
+	IdentityRef        *infrav1.AWSIdentityReference `json:"identityRef,omitempty"`
+	Region             string                        `json:"region,omitempty"`
+	// CredentialsSecretRef references a secret with necessary credentials to connect to the OCM API.
+	// The secret should contain the following data keys:
+	// - ocmToken: eyJhbGciOiJIUzI1NiIsI....
+	// - ocmApiUrl: Optional, defaults to 'https://api.openshift.com'
```
serngawy: Please remove those 2 lines. Currently the secret is having other info to authenticate

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,604 @@
...
+		if err := scope.PatchObject(); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if scope.RosaRoleConfig.Status.OIDCID == "" {
+		err = r.createOIDCConfig(roleConfig, scope, ocmClient)
+		if err != nil {
+			conditions.MarkFalse(scope.RosaRoleConfig, expinfrav1.RosaRoleConfigReadyCondition, expinfrav1.RosaRoleConfigReconciliationFailedReason, clusterv1.ConditionSeverityError, "Failed to create OIDC Config: %v", err)
+			return ctrl.Result{RequeueAfter: time.Second * 60}, fmt.Errorf("failed to OICD Config: %w", err)
```
mzazrivec: The above line (and all the other returns with non empty result and non-nil error) will produce the following log entry:
```
I0722 11:52:38.495821      15 controller.go:345] "Warning: Reconciler returned both a non-zero result and a non-nil error. The result will always be ignored if the error is non-nil and the non-nil error causes requeuing with exponential backoff. For more details, see: https://pkg.go.dev/sigs.k8s.io/controller-runtime/pkg/reconcile#Reconciler" controller="rosaroleconfig" 
```

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
PanSpagetka: This is intended, but we need both `&&` and `||` of all values to make behavior you described in comment below.

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +183,29 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
...
+		r.Spec.RolesRef.IngressARN != "" || r.Spec.RolesRef.ImageRegistryARN != "" || r.Spec.RolesRef.StorageARN != "" ||
+		r.Spec.RolesRef.NetworkARN != "" || r.Spec.RolesRef.KubeCloudControllerARN != "" || r.Spec.RolesRef.NodePoolManagementARN != "" ||
+		r.Spec.RolesRef.ControlPlaneOperatorARN != "" || r.Spec.RolesRef.KMSProviderARN != ""
+
+	hasAllDirectRoleFields := r.Spec.OIDCID != "" && r.Spec.InstallerRoleARN != "" && r.Spec.SupportRoleARN != "" && r.Spec.WorkerRoleARN != "" &&
+		r.Spec.RolesRef.IngressARN != "" && r.Spec.RolesRef.ImageRegistryARN != "" && r.Spec.RolesRef.StorageARN != "" &&
+		r.Spec.RolesRef.NetworkARN != "" && r.Spec.RolesRef.KubeCloudControllerARN != "" && r.Spec.RolesRef.NodePoolManagementARN != "" &&
+		r.Spec.RolesRef.ControlPlaneOperatorARN != "" && r.Spec.RolesRef.KMSProviderARN != ""
+
+	if hasRosaRoleConfigRef && hasAnyDirectRoleFields {
```
serngawy: Not sure why you making this check; if the rosaRoleConfig is defined we use it ignore the other roles fields check [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5499#discussion_r2218505200)

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +183,29 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
 	return nil
 }
 
+func (r *ROSAControlPlane) validateRosaRoleConfig() *field.Error {
+	hasRosaRoleConfigRef := r.Spec.RosaRoleConfigRef != nil
+	hasAnyDirectRoleFields := r.Spec.OIDCID != "" || r.Spec.InstallerRoleARN != "" || r.Spec.SupportRoleARN != "" || r.Spec.WorkerRoleARN != "" ||
```
serngawy, line 206: no need to check with || having 1 field missing should raise error if the rosaRoleConfig not set

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +183,29 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
 	return nil
 }
 
+func (r *ROSAControlPlane) validateRosaRoleConfig() *field.Error {
+	hasRosaRoleConfigRef := r.Spec.RosaRoleConfigRef != nil
```
serngawy:
```suggestion
	if r.Spec.RosaRoleConfigRef != nil {
	  return nil
	}
```

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +183,29 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
...
+	hasAllDirectRoleFields := r.Spec.OIDCID != "" && r.Spec.InstallerRoleARN != "" && r.Spec.SupportRoleARN != "" && r.Spec.WorkerRoleARN != "" &&
+		r.Spec.RolesRef.IngressARN != "" && r.Spec.RolesRef.ImageRegistryARN != "" && r.Spec.RolesRef.StorageARN != "" &&
+		r.Spec.RolesRef.NetworkARN != "" && r.Spec.RolesRef.KubeCloudControllerARN != "" && r.Spec.RolesRef.NodePoolManagementARN != "" &&
+		r.Spec.RolesRef.ControlPlaneOperatorARN != "" && r.Spec.RolesRef.KMSProviderARN != ""
+
+	if hasRosaRoleConfigRef && hasAnyDirectRoleFields {
+		return field.Invalid(field.NewPath("spec.rosaRoleConfigRef"), r.Spec.RosaRoleConfigRef, "rosaRoleConfigRef and direct role fields (oidcID, installerRoleARN, supportRoleARN, workerRoleARN, rolesRef) are mutually exclusive")
+	}
+
+	if !hasRosaRoleConfigRef && !hasAllDirectRoleFields {
```
serngawy:
```suggestion
	if !hasAllDirectRoleFields {
	// raise error here specifying which fields are missing 
	// OR do check for every field once it is missing return field.invalid 
	}
```

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,201 @@
...
+	ExternalAuthProviders []rosacontrolplanev1.ExternalAuthProvider `json:"externalAuthProviders,omitempty"`
+	// IdentityRef is the AWS identity reference for the OIDC config.
+	// +immutable
+	IdentityRef *infrav1.AWSIdentityReference `json:"identityRef,omitempty"`
+	// Region is the AWS region for the OIDC config.
+	// +immutable
+	Region string `json:"region,omitempty"`
+	// Prefix is the prefix for the OIDC config.
+	// +immutable
+	Prefix string `json:"prefix"`
```
serngawy: lets make all the prefix fields have limit length 4

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,201 @@
...
+	// +listMapKey=name
+	// +kubebuilder:validation:MaxItems=1
+	// +immutable
+	ExternalAuthProviders []rosacontrolplanev1.ExternalAuthProvider `json:"externalAuthProviders,omitempty"`
+	// IdentityRef is the AWS identity reference for the OIDC config.
+	// +immutable
+	IdentityRef *infrav1.AWSIdentityReference `json:"identityRef,omitempty"`
+	// Region is the AWS region for the OIDC config.
+	// +immutable
+	Region string `json:"region,omitempty"`
```
serngawy: The region is defined above for all accountRole, operatorRole and oidcConfig

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,603 @@
...
+	}
+
+	if controllerutil.AddFinalizer(scope.RosaRoleConfig, expinfrav1.RosaRoleConfigFinalizer) {
+		if err := scope.PatchObject(); err != nil {
+			return ctrl.Result{}, err
+		}
+	}
+
+	if scope.RosaRoleConfig.Status.OIDCID == "" {
+		err = r.createOIDCConfig(roleConfig, scope, ocmClient)
```
serngawy: the logic need to change as follow;
First you get the oidcConfig using ocm client;
1- if the oidcConfig exist we set the oidcConfig status info and condition
2- if the oidcConfig not exist, we create the oidcConfig then set the oidcConfig stats info and condition
Same applied to accountRoles, operatorRoles and oidcProvider

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -226,11 +228,58 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
...
+		}
+
+		// Check if RosaRoleConfig is ready
+		if !conditions.IsTrue(rosaRoleConfig, expinfrav1.RosaRoleConfigReadyCondition) {
+			conditions.MarkFalse(rosaScope.ControlPlane,
+				rosacontrolplanev1.ROSARoleConfigReadyCondition,
+				rosacontrolplanev1.ROSARoleConfigNotReadyReason,
+				clusterv1.ConditionSeverityWarning,
+				"RosaRoleConfig %s/%s is not ready", rosaScope.ControlPlane.Namespace, rosaScope.ControlPlane.Spec.RosaRoleConfigRef.Name)
+			return ctrl.Result{}, fmt.Errorf("RosaRoleConfig %s/%s is not ready", rosaScope.ControlPlane.Namespace, rosaScope.ControlPlane.Spec.RosaRoleConfigRef.Name)
```
mzazrivec: I think it would be more suitable to log an info message here and just requeue without returning the error.

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,169 @@
...
+	// +immutable
+	Version string `json:"version"`
+	// SharedVPCConfig is used to set up shared VPC.
+	// +optional
+	// +immutable
+	SharedVPCConfig SharedVPCConfig `json:"sharedVPCConfig,omitempty"`
+}
+
+// OperatorRoleConfig defines cluster-specific operator IAM roles based on your cluster configuration.
+type OperatorRoleConfig struct {
```
serngawy, line 80: Please follow the [proposal](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/proposal/20250324-rosa-iam-roles.md?plain=1#L79) definition for the OperatorRoleConfig. OperatorRoleConfig must have the option to assign the [oidc-Id](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/proposal/20250324-rosa-iam-roles.md?plain=1#L91) AND it should be mutual exclusive with the oidcConfig->[createManagedOIDC](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/proposal/20250324-rosa-iam-roles.md?plain=1#L103)

serngawy (changes requested):
the [OidcConfig](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/docs/proposal/20250324-rosa-iam-roles.md?plain=1#L101) is missing in the ROSARoleConfig API .

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -907,7 +958,7 @@ func validateControlPlaneSpec(ocmClient rosa.OCMClient, rosaScope *scope.ROSACon
 	return "", nil
 }
 
-func buildOCMClusterSpec(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpec, creator *rosaaws.Creator) (ocm.Spec, error) {
+func buildOCMClusterSpec(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpec, roleConfig *expinfrav1.ROSARoleConfig, creator *rosaaws.Creator) (ocm.Spec, error) {
```
serngawy, line 969: ExternalAuth need to be defined as well in the ROSARoleConfig and assigned to the cluster spec similar to [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/controllers/rosacontrolplane_controller.go#L946)

PanSpagetka: We have discussed this on meeting and we dont need to include `ExternalAuth` in RosaRoleConfig.

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,173 @@
...
+	Prefix string `json:"prefix"`
+	// The ARN of the policy that is used to set the permissions boundary for the operator roles.
+	// +optional
+	// +immutable
+	PermissionsBoundaryARN string `json:"permissionsBoundaryARN,omitempty"`
+	// SharedVPCConfig is used to set up shared VPC.
+	// +optional
+	// +immutable
+	SharedVPCConfig SharedVPCConfig `json:"sharedVPCConfig,omitempty"`
+	// OIDCID is the ID of the OIDC config that will be used to create the operator roles.
```
serngawy, line 94:
```suggestion
	// OIDCID is the ID of the OIDC config that will be used to create the operator roles. A managed OIDC-provider will be created if the OIDCID not specified
```

exp/api/v1beta2/rosaroleconfig_types.go
serngawy: as we discussed no need for region

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,658 @@
...
+func (r *ROSARoleConfigReconciler) reconcileDelete(scope *scope.RosaRoleConfigScope, ocmClient *ocm.Client) error {
+	log := rosalogging.NewLogger()
+	awsClient, err := aws.NewClient().Logger(log).Build()
+	if err != nil {
+		conditions.MarkFalse(scope.RosaRoleConfig, expinfrav1.RosaRoleConfigReadyCondition, expinfrav1.RosaRoleConfigDeletionFailedReason, clusterv1.ConditionSeverityError, "Failed to create AWS client: %v", err)
+		return err
+	}
+
+	oidcID := scope.RosaRoleConfig.Status.OIDCID
+	err = r.deleteOIDCProvider(ocmClient, awsClient, oidcID)
```
serngawy: We don't delete the oidc-provider if the user set it under the spec.operatorRole.OIDCConfig

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -124,13 +123,23 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// +kubebuilder:default=WaitForAcknowledge
 	VersionGate VersionGateAckType `json:"versionGate"`
 
+	// RosaRoleConfigRef is a reference to a RosaRoleConfig resource that contains account and operator roles and OIDC configuration.
+	// If specified, the roles and OIDC configuration will be taken from the referenced RosaRoleConfig instead of the direct fields.
+	//
+	// +optional
+	RosaRoleConfigRef *corev1.LocalObjectReference `json:"rosaRoleConfigRef,omitempty"`
```
serngawy, line 130: The OIDC is immutable , so validation should be applied (could be in the webhook [update](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go#L88) ) that RosaRoleConfigRef cannot be change if the cluster is provisioned

exp/controllers/rosaroleconfig_controller.go
serngawy: I guess the order here is not correct, deleting the operatrRoles first then oidc-provider then odic-config ?

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,662 @@
...
+	operatorRoles, err := runtime.AWSClient.ListOperatorRoles(version, "", config.Prefix)
+
+	if err != nil {
+		return err
+	}
+
+	if len(operatorRoles) > 0 {
+		for _, roles := range operatorRoles {
+			for _, role := range roles {
+				if strings.Contains(role.RoleName, fmt.Sprintf("%s-openshift-ingress-operator-cloud-credentials", config.Prefix)) {
```
serngawy: why strings.contain not exact string equal by name ?

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,662 @@
...
+				}
+				if strings.Contains(role.RoleName, fmt.Sprintf("%s-kube-system-control-plane-operator", config.Prefix)) {
+					scope.RosaRoleConfig.Status.OperatorRolesRef.ControlPlaneOperatorARN = role.RoleARN
+				}
+				if strings.Contains(role.RoleName, fmt.Sprintf("%s-kube-system-kms-provider", config.Prefix)) {
+					scope.RosaRoleConfig.Status.OperatorRolesRef.KMSProviderARN = role.RoleARN
+				}
+			}
+		}
+	} else {
```
serngawy: The logic here is not correct , what if one Role is missing not all of them ? we create 8 operator Roles, we should check every single one

PanSpagetka: No, if there is one role missing we don't create 8 roles because you can't have 2 roles with same name. And in rosa cli we handle roles creation as a batch operation, so we are consistent with that.

serngawy: that is not correct , rosa lib will re-create the missing operator roles if there are any. You can reproduce this scenario with rosa-cli

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,662 @@
...
+		return err
+	}
+
+	if len(operatorRoles) > 0 {
+		for _, roles := range operatorRoles {
+			for _, role := range roles {
+				if strings.Contains(role.RoleName, fmt.Sprintf("%s-openshift-ingress-operator-cloud-credentials", config.Prefix)) {
+					scope.RosaRoleConfig.Status.OperatorRolesRef.IngressARN = role.RoleARN
+				}
+				if strings.Contains(role.RoleName, fmt.Sprintf("%s-openshift-image-registry-installer-cloud-credentials", config.Prefix)) {
```
serngawy:
```suggestion
				 else if strings.Contains(role.RoleName, fmt.Sprintf("%s-openshift-image-registry-installer-cloud-credentials", config.Prefix)) {
```
better to set else if with the following if(s) check

serngawy: better to have the RoleNames in list/map so easier for you to check the roles count and existence

serngawy (changes requested):

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: No having roles names in list doesn't help, every role has its own field so every condition needs to be spelled out.

serngawy (approved):
Thanks @PanSpagetka, sounds good in general we can add more test in future PRs

serngawy:
Please add release note , mention adding Rosa RoleConfig API

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -714,8 +737,8 @@ type AWSRolesRef struct {
 	//	]
 	// }
 	// +immutable
-	ControlPlaneOperatorARN string `json:"controlPlaneOperatorARN"`
-	KMSProviderARN          string `json:"kmsProviderARN"`
+	ControlPlaneOperatorARN string `json:"controlPlaneOperatorARN,omitempty"`
```
nrb: Why are these fields being updated with `omitempty`?

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
nrb: True, but you don't know _which_ field is missing without checking all of them.

I can see why this information is being moved into a RoleConfig, because checking all these fields individually is tricky.

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +186,58 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
 	return nil
 }
 
+func (r *ROSAControlPlane) validateRosaRoleConfig() *field.Error {
+	hasAnyDirectRoleFields := r.Spec.OIDCID != "" || r.Spec.InstallerRoleARN != "" || r.Spec.SupportRoleARN != "" || r.Spec.WorkerRoleARN != "" ||
+		r.Spec.RolesRef.IngressARN != "" || r.Spec.RolesRef.ImageRegistryARN != "" || r.Spec.RolesRef.StorageARN != "" ||
```
nrb, line 207: For the sake of clarity, perhaps we could use a `r.Spec.RolesRef == nil` check to demonstrate that there is no `RolesRef` at all, without checking every field within it.

We can look at the specific `RolesRef` fields themselves further down in the function, since up here we're just looking for a high level conflict.

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
```diff
@@ -179,6 +186,58 @@ func (r *ROSAControlPlane) validateExternalAuthProviders() *field.Error {
...
+	if r.Spec.InstallerRoleARN == "" {
+		return field.Invalid(field.NewPath("spec.installerRoleARN"), r.Spec.InstallerRoleARN, "must be specified")
+	}
+	if r.Spec.SupportRoleARN == "" {
+		return field.Invalid(field.NewPath("spec.supportRoleARN"), r.Spec.SupportRoleARN, "must be specified")
+	}
+	if r.Spec.WorkerRoleARN == "" {
+		return field.Invalid(field.NewPath("spec.workerRoleARN"), r.Spec.WorkerRoleARN, "must be specified")
+	}
+	if r.Spec.RolesRef.IngressARN == "" {
```
nrb, line 230: Probably also want to make sure that r.Spec.RolesRef isn't nil at some point, possibly up above where my previous comment is.

Another suggestion would be to say if you reach this point in the code, you know all the previous fields aren't `""`, because the function would have returned.

Knowing that, we could say that at this point if r.Spec.RolesRef is not nil, then the user has specified all the direct fields and a RolesRef, triggering the mutual exclusion.

In terms of clarity, I think it would also be worth having a couple helper functions for checking the direct fields and the `RolesRef` fields; it keeps the conditionals grouped into smaller chunks.

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -0,0 +1,173 @@
...
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	rosacontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/rosa/api/v1beta2"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
```
nrb:
```suggestion
```

exp/api/v1beta2/rosaroleconfig_webhook.go
```diff
@@ -0,0 +1,48 @@
...
+// +kubebuilder:webhook:verbs=create;update,path=/validate-infrastructure-cluster-x-k8s-io-v1beta2-rosaroleconfig,mutating=false,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=rosaroleconfigs,versions=v1beta2,name=validation.rosaroleconfig.infrastructure.cluster.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+// +kubebuilder:webhook:verbs=create;update,path=/mutate-infrastructure-cluster-x-k8s-io-v1beta2-rosaroleconfig,mutating=true,failurePolicy=fail,matchPolicy=Equivalent,groups=infrastructure.cluster.x-k8s.io,resources=rosaroleconfigs,versions=v1beta2,name=default.rosaroleconfig.infrastructure.cluster.x-k8s.io,sideEffects=None,admissionReviewVersions=v1;v1beta1
+
+type rosaRoleConfigWebhook struct{}
+
+var _ webhook.CustomDefaulter = &rosaRoleConfigWebhook{}
+var _ webhook.CustomValidator = &rosaRoleConfigWebhook{}
+
+// ValidateCreate implements admission.Validator.
+func (r *rosaRoleConfigWebhook) ValidateCreate(ctx context.Context, obj runtime.Object) (warnings admission.Warnings, err error) {
```
nrb, line 31: I would think you want to check at least that your required fields are all set.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,663 @@
+/*
+Copyright The Kubernetes Authors.
```
nrb: Make sure to add the year, like https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/a0968d3c19ea7beb8908f14c7275290726d4fc08/exp/api/v1beta1/awsmanagedmachinepool_types.go#L2

This should appear in every file, it looks like some are currently in tree without a license header.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,663 @@
...
+
+func (r *ROSARoleConfigReconciler) Reconcile(ctx context.Context, req ctrl.Request) (res ctrl.Result, reterr error) {
+	log := logger.FromContext(ctx)
+
+	roleConfig := &expinfrav1.ROSARoleConfig{}
+	if err := r.Get(ctx, req.NamespacedName, roleConfig); err != nil {
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{Requeue: true}, nil
```
nrb, line 97: The error is never reported here, so the user won't know why we're re-queueing, which could result in an endless loop.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,663 @@
...
+		return ctrl.Result{}, fmt.Errorf("failed to create OCM client: %w", err)
+	}
+
+	if !roleConfig.ObjectMeta.DeletionTimestamp.IsZero() {
+		conditions.MarkFalse(scope.RosaRoleConfig, expinfrav1.RosaRoleConfigReadyCondition, expinfrav1.RosaRoleConfigDeletionStarted, clusterv1.ConditionSeverityInfo, "Deletion of RosaRolesConfig started")
+		return ctrl.Result{}, r.reconcileDelete(scope, ocmClient)
+	}
+
+	if controllerutil.AddFinalizer(scope.RosaRoleConfig, expinfrav1.RosaRoleConfigFinalizer) {
+		if err := scope.PatchObject(); err != nil {
```
nrb: This results in at least 2 calls to `PatchObject` - here and within the deferred clean up function.

Can we make all of our field changes and then do 1 patch call at the end of a happy path reconcile?

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,663 @@
...
+	isSharedVpc := config.SharedVPCConfig.VPCEndpointRoleARN != "" && config.SharedVPCConfig.RouteRoleARN != ""
+
+	operatorRoles, err := runtime.AWSClient.ListOperatorRoles(version, "", config.Prefix)
+
+	if err != nil {
+		return err
+	}
+
+	for _, roles := range operatorRoles {
+		for _, role := range roles {
```
nrb, line 242: I would re-arrange this a bit.

Outside the loop, possibly in the API package:

```go
const (
  IngressOperatorARNSuffix = "-openshift-ingress-operator-cloud-credentials"
  ImageRegistryARNSuffix = "-openshift-image-registry-installer-cloud-credentials"
....
)
```

Then, here,

```go
roleSuffix := strings.TrimPrefix(role.RoleName, config.Prefix)

if roleSuffix == role.RoleName {
  // return error, the prefix wasn't found
}

switch roleSuffix {
case IngressOperatorRoleSuffix:
  scope.RosaRoleConfig.Status.OperatorRolesRef.IngressARN = role.RoleARN
  break
case ImageRegistryRoleSuffix:
  scope.RosaRoleConfig.Status.OperatorRolesRef.ImageRegistryARN = role.RoleARN
  break
....
default:
   // error because the suffix didn't match known cases.
  } 
```

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,663 @@
...
+	if createRoles {
+		runtime.Reporter = (&rosa.Reporter{})
+		runtime.Logger = rosalogging.NewLogger()
+		runtime.AWSClient, err = aws.NewClient().Logger(runtime.Logger).Build()
+		if err != nil {
+			return fmt.Errorf("failed to create aws client: %w", err)
+		}
+
+		managedPolicies := true
+		isSharedVpc := config.SharedVPCConfig.VPCEndpointRoleARN != "" && config.SharedVPCConfig.RouteRoleARN != ""
```
nrb: This is used a few places in the code, might be worth making the check a helper function on config.SharedVPCConfig.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,663 @@
...
+	if err != nil {
+		return err
+	}
+
+	clusters, err := ocmClient.GetAllClusters(creator)
+	if err != nil {
+		return err
+	}
+
+	var err2, err3 error
```
nrb: Please use descriptive names instead of numbers for the variable names.

Although it looks like we only ever return 1 error, so you could re-use `err`, rather than have separate variables and extra nil checks.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,663 @@
...
+		} `json:"Statement"`
+	}
+
+	err = json.Unmarshal([]byte(decodedString), &policyDoc)
+	if err != nil {
+		return "", err
+	}
+
+	// Extract from the 'Federated' ARN
+	if len(policyDoc.Statement) > 0 {
```
nrb, line 550: Should we check for `len == 1`? Is it ok that any other statements are ignored?

pkg/cloud/scope/rosaroleconfig.go
```diff
@@ -0,0 +1,170 @@
...
+
+	RosaRoleConfigScope := &RosaRoleConfigScope{
+		Logger:         *params.Logger,
+		Client:         params.Client,
+		controllerName: params.ControllerName,
+		patchHelper:    nil,
+		RosaRoleConfig: params.RosaRoleConfig,
+	}
+
+	sessionv2, serviceLimitersv2, err := sessionForClusterWithRegionV2(params.Client, RosaRoleConfigScope, "", params.Endpoints, params.Logger)
```
nrb: The variables can drop `v2` since there's nothing in this file using the v1 SDK.

pkg/cloud/scope/rosaroleconfig.go
```diff
@@ -0,0 +1,170 @@
...
+ limitations under the License.
+*/
+
+package scope
+
+import (
+	"context"
+
+	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
+	iamv2 "github.com/aws/aws-sdk-go-v2/service/iam"
```
nrb: Could import these directly without an alias now that there's no reference to the older SDK.

pkg/cloud/scope/rosaroleconfig.go
```diff
@@ -0,0 +1,170 @@
...
+
+// GetClient Returns RosaRoleConfigScope client.
+func (s *RosaRoleConfigScope) GetClient() client.Client {
+	return s.Client
+}
+
+// PatchObject persists the RosaRoleConfig configuration and status.
+func (s *RosaRoleConfigScope) PatchObject() error {
+	return s.patchHelper.Patch(
+		context.TODO(),
```
nrb:
```suggestion
		context.Background(),
```

pkg/cloud/scope/rosaroleconfig.go
```diff
@@ -0,0 +1,170 @@
...
+		s.RosaRoleConfig)
+}
+
+// Close closes the current scope persisting the RosaRoleConfig configuration and status.
+func (s *RosaRoleConfigScope) Close() error {
```
nrb: Why have this wrapper function?

pkg/rosa/helpers.go
```diff
@@ -38,3 +42,34 @@ func handleErr(res *ocmerrors.Error, err error) error {
...
 }
+
+// Reporter is a helper struct used by rosa CLI runtime to report errors.
+// We need it to limit number of reports by rosa CLI runtime.
+type Reporter struct {
+	scope scope.RosaRoleConfigScope
+}
+
+// Debugf prints a debug message with the given format and arguments.
+func (r *Reporter) Debugf(format string, args ...interface{}) {
```
nrb, line 69: These functions just swallow the input? Are you trying to hide the output?

go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/cluster-api-provider-aws/v2
 
-go 1.23.0
+go 1.23.1
```
nrb: Leave this at 1.23.0.

nrb (changes requested):
Some questions and style suggestions.

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
PanSpagetka: Because its used at two place: `ROSAControlPlane.Spec.RolesRef` and  `ROSARoleConfigSpec.Status.OperatorRolesRef`. For `ROSARoleConfigSpec` usecase @serngawy wanted to scenario when some(not all) operator roles exist. 

E.g. `ControlPlaneOperatorARN` is missing, other roles are present. In the first reconcile we get 7 other roles and set them in status and also we initiate creation if the missing `ControlPlaneOperatorARN` role. But without `omitempty` this patch ended with error as `ControlPlaneOperatorARN` was missing. In second round of reconcile we pick up all roles and we should be fine.

controlplane/rosa/api/v1beta2/rosacontrolplane_webhook.go
PanSpagetka: The `hasAnyDirectRoleFields` is here to check that both `RosaRoleConfigRef` and direct role references in Control plane are present. As both are defining the same thing, only one can be set and it is up to the user to choose which one.

Block of many ifs below is to compensate adding `// +optional` to all references in Control plane. I can extract it into another function but I would rather kept the logic. I think it is better to provider user with error message that contains specific missing field.

`r.Spec.RolesRef.` is not a pointer, so it cant be nil. it contains only operator roles and is part of original direct roles settings. In contrast `r.Spec.RosaRoleConfigRef` contains reference to new `RosaRoleConfig` CR and can be nil(nil value is checked)

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: Similar pattern is present at multiple places around the codebase, so I assumed that hiding the error is intentional. I can log the error before requeuing.

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: So I will just remove this Patch, right?

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: I am changing this to use `errors.Join`

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: I am not sure. It seems plausible that either an user or we as RH could add another policy. We just want to extract OIDC here and first policy should contain it.

This is for usecase when user sets operator role prefix that already exist, than we want to use OIDC that is bound to those roles.

pkg/cloud/scope/rosaroleconfig.go
PanSpagetka: I guess we can delete it

pkg/rosa/helpers.go
PanSpagetka: Yep, its to hide output logs from rosa cli.

go.mod
PanSpagetka: This diff is probably outdated. it seems that in main we already have version 1.23.1 https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/go.mod

PanSpagetka:
@nrb Thank you for review! I created new temporary commit with changes, so its easier to review. I will squash it when we agree that code is good enough.

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
serngawy: @PanSpagetka Not sure if I understand your comment , The case we were discussing is We have the 8 Roles ARN are created for first time so all of them exist, then one or more roles get deleted by aws admin using aws console for example (not using our CR) . when reconcile happen some of those 8 roles will not retrieved by awsClient. So we will re-create the missing roles and set the ARN. Why we you need to set it empty ? there is no chance we will have this case.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,660 @@
...
+		return nil
+	}
+	// Try to get OIDC UUID from some operator role policy document.
+	roleName := fmt.Sprintf("%s-openshift-ingress-operator-cloud-credentials", roleConfig.Spec.OperatorRoleConfig.Prefix)
+	roleDetails, err := scope.IAMClient().GetRole(context.TODO(), &iamv2.GetRoleInput{
+		RoleName: &roleName,
+	})
+	if err != nil {
+		return r.createOIDCConfig(scope, ocmClient)
+	}
```
serngawy: @PanSpagetka the logic here not make sense , if there is an error we don't re-create the oidc config we return the error.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,660 @@
...
+	oidcID, err := r.GetOIDCIDFromOperatorRole(scope, roleDetails)
+	if err != nil {
+		return r.createOIDCConfig(scope, ocmClient)
+	}
```
serngawy: same here.

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: Its because `scope.IAMClient().GetRole` returns error when role is not found. And if the role is not found, we cant derive the OIDC id from it.

exp/controllers/rosaroleconfig_controller.go
PanSpagetka: Similarly `GetOIDCIDFromOperatorRole` returns error when we cant get OIDC id from the role. So we create new one

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,660 @@
...
+	if !r.operatorRolesReady(&scope.RosaRoleConfig.Status.OperatorRolesRef) {
+		err = operatorroles.CreateOperatorRoles(runtime, ocm.Production, config.PermissionsBoundaryARN, interactive.ModeAuto, policies, version, config.SharedVPCConfig.IsSharedVPC(), config.Prefix, hostedCp, installerRoleArn, forcePolicyCreation,
+			oidcConfigID, config.SharedVPCConfig.RouteRoleARN, ocm.DefaultChannelGroup, config.SharedVPCConfig.VPCEndpointRoleARN)
+		return err
+	}
+
+	return nil
+}
+
+func (r *ROSARoleConfigReconciler) reconcileOIDCConfig(roleConfig *expinfrav1.ROSARoleConfig, scope *scope.RosaRoleConfigScope, ocmClient *ocm.Client) error {
```
serngawy:
```
func (r *ROSARoleConfigReconciler) reconcileOIDCConfig(roleConfig *expinfrav1.ROSARoleConfig, scope *scope.RosaRoleConfigScope, ocmClient *ocm.Client) error {
	if scope.RosaRoleConfig.Status.OIDCID != "" {
       // retrieve using ocm client to be sure it is exist 
		if  disappeared {
		return error
		}
		return nil
	}
	if roleConfig.Spec.OperatorRoleConfig.OIDCID != "" {
		scope.RosaRoleConfig.Status.OIDCID = roleConfig.Spec.OperatorRoleConfig.OIDCID
		return nil
	}
	
	return r.createOIDCConfig(scope, ocmClient)
}
```

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,660 @@
...
+	oidcID, err := r.GetOIDCIDFromOperatorRole(scope, roleDetails)
+	if err != nil {
+		return r.createOIDCConfig(scope, ocmClient)
+	}
+	scope.RosaRoleConfig.Status.OIDCID = oidcID
+	return nil
+}
+
+func (r *ROSARoleConfigReconciler) createOIDCProvider(scope *scope.RosaRoleConfigScope, ocmClient *ocm.Client) error {
+	if scope.RosaRoleConfig.Status.OIDCProviderARN != "" {
```
serngawy:
```suggestion
	if scope.RosaRoleConfig.Status.OIDCProviderARN != "" {
	// we don't just return we have to retrieve the oidc-provider using ocm client 
	}
```

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,662 @@
...
+}
+
+func (r *ROSARoleConfigReconciler) createOperatorRoles(ctx context.Context, roleConfig *expinfrav1.ROSARoleConfig, scope *scope.RosaRoleConfigScope, ocmClient *ocm.Client) error {
+	installerRoleArn := scope.RosaRoleConfig.Status.AccountRolesRef.InstallerRoleARN
+	if installerRoleArn == "" {
+		return fmt.Errorf("installer role is empty")
+	}
+	oidcConfigID := scope.RosaRoleConfig.Status.OIDCID
+	if oidcConfigID == "" {
+		return fmt.Errorf("OIDCID is empty")
```
serngawy:
```suggestion
		return fmt.Errorf("OIDCID is required to create operator roles.")
```

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -0,0 +1,662 @@
...
+		oidcConfig, err := ocmClient.GetOidcConfig(scope.RosaRoleConfig.Status.OIDCID)
+		if err != nil || oidcConfig == nil {
+			return fmt.Errorf("failed to get OIDC config: %w", err)
+		}
+		return nil
+	}
+	if roleConfig.Spec.OperatorRoleConfig.OIDCID != "" {
+		scope.RosaRoleConfig.Status.OIDCID = roleConfig.Spec.OperatorRoleConfig.OIDCID
+		return nil
+	}
```
serngawy:
```suggestion
    oidcID := ""
   if scope.RosaRoleConfig.Status.OIDCID != "" {
	    oidcID = scope.RosaRoleConfig.Status.OIDCID
} else if roleConfig.Spec.OperatorRoleConfig.OIDCID != "" {
	oidcID = roleConfig.Spec.OperatorRoleConfig.OIDCID
}
if oidcID != "" {
  	oidcConfig, err := ocmClient.GetOidcConfig(oidcID)
if err != nil || oidcConfig == nil {
	    		return fmt.Errorf("failed to get OIDC config: %w", err)
    }
    scope.RosaRoleConfig.Status.OIDCID = oidcID
	    return nil
	     }
```

PanSpagetka:
@nrb @serngawy could you take another look? I think we should be good now

serngawy:
/close
for the favor of this one https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5667

PR #5520, opened 2025-06-04
✨  allow preserving addons on delete

alam0rt:
**What type of PR is this?**

feature

We are looking at migrating away from the managed addons (specifically kube-proxy) and wanting to use our own self-managed.

The issue is that both `kube-system/daemonset/kube-proxy`s can't exist at the same time for obvious reason.

Removing the addon from `.spec.addons` will of course, delete the addon.

Deploying _over_ the addon works, but again, removing the addon will remove the  now self-managed addon.

The only way (without juggling multiple daemonsets, using taints/tolerations to run both at the same time, etc) for a clean migration is to (please let me know if a good way comes to mind):

* Deploy over the managed addon
* Remove the addon while preserving the resources on the cluster

We _could_ potentially stop CAPA reconciliation and remove the addon manually, but kind of want to avoid that.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Allow preserving EKS addons.
```

### DIFF
--- a/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanetemplates.yaml
+++ b/config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanetemplates.yaml
@@ -83,6 +83,11 @@ spec:
                               description: Name is the name of the addon
                               minLength: 2
                               type: string
+                            preserveOnDelete:
+                              description: |-
+                                PreserveOnDelete indicates that the addon resources should be
+                                preserved in the cluster on delete.
+                              type: boolean
                             serviceAccountRoleARN:
                               description: ServiceAccountRoleArn is the ARN of an
                                 IAM role to bind to the addons service account


alam0rt:
Speaking with my team, would it be better if the `preserve` field was in `sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta1.Addon`?

image

alam0rt:
> Speaking with my team, would it be better if the `preserve` field was in `sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta1.Addon`?
> image

I went and did it. Seems cleaner.

richardcase:
@alam0rt - thanks for this. Just so i understand this correct. If you want to delete and use your own version of the addon, is this the process:
- Define the kube proxy addon with `preserveOnDelete=true` in AWSManagedControlPlane
- Delete the addon from AWSManagedControlPlane

richardcase:
Also, guessing this is only for the use case where `bootstrapSelfManagedAddons` wasn't specified at cluster creation time?

alam0rt:
> @alam0rt - thanks for this. Just so i understand this correct. If you want to delete and use your own version of the addon, is this the process:
> 
> - Define the kube proxy addon with `preserveOnDelete=true` in AWSManagedControlPlane
> 
> - Delete the addon from AWSManagedControlPlane

That's correct!

PR #5518, opened 2025-06-03
[WIP] Migrate gc to aws sdk v2

Danil-Grigorev:
**What type of PR is this?**

/kind support

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note

```

PR #5497, opened 2025-05-16
✨Migrate autoscaling to AWS SDK V2

alexander-demicev:
**What type of PR is this?**

/kind cleanup
/kind deprecation

**What this PR does / why we need it**:
This PR migrates autoscaling to AWS SDK V2. I had to add some changes to IAM in the test package, which I will clean up in a separate PR together with the IAM migration.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5401

**Special notes for your reviewer**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Migrate autoscaling to AWS SDK V2
```

Issue #5409, opened 2025-03-17
Migrate network code to AWS SDK v2


nrb:
Update calls in the following packages and files. Note that this is not necessarily an exhaustive list.

./pkg/cloud/scope/network.go
./pkg/cloud/services/network
./pkg/cloud/services/network/network.go
./pkg/cloud/services/mock_services/network_interface_mock.go

damdo:
See examples on how to migrate here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5389

damdo:
Hey @sowjanyadreddy thank you for stepping up and helping us with the AWS SDK v2 migration efforts for network code.
I was wondering if you had a branch or draft PR we can start to have a look at for an early review. We are looking at moving ahead as quickly as possible with the migrations as they are a blocker for the next CAPA release.

Thanks!

sowjanyadreddy:
Hey  damdo,
     Yes I have 8 files out of the 12 migrated in the network . Will send
in a draft MR once I am done with the last 4.

*Thanks,*
*Sowjanya*

damdo:
Awesome! Thanks @sowjanyadreddy :)

sowjanyadreddy:
In my code review with @richardcase  earlier today we realized that https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5521 addresses the same changes. So resolving this issue.

richardcase:
@sowjanyadreddy - i update the description for #5521 to say that it closes this issue. When the PR merges it will close this.

damdo:
It looks like when https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5521 the GH automation didn't close this issue for some reason. Closing it manually.

PR #5389, opened 2025-03-13
✨ Migrate S3, Sessions and Endpoints packages to AWS SDK v2

punkwalker:
**What type of PR is this?**
/kind feature
/kind cleanup

**What this PR does / why we need it**:
It is initial aws-sdk-go-v2 migration for few components. This PR only migrates S3 Client to V2.
I have created this PR as a duplicate of #5272 as @mjlshen does not have cycles to work on it. So I will be working on this migration going forward. I wanted to make sure we get quicker review on this migration PRs.

**Which issue(s) this PR fixes**:
Fixes: #5400
Fixes: #5414

**Special notes for your reviewer**:
This migration will require multiple such PRs. It would be great if we can get some thoughts on the approach of this migration. It might be missing some test cases for V2 clients but we will add it as we go along.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Migrate the S3 service to aws-sdk-go-v2
```

PR #5521, opened 2025-06-04
:sparkles:  WIP Migrate ec2 to AWS SDK V2

### LINKED ISSUE #5403: Migrate ec2 code to AWS SDK v2
Update calls in the following packages and files. Note that this is not necessarily an exhaustive list.

```
./test/mocks/aws_ec2api_mock.go
./pkg/cloud/scope/ec2.go
./pkg/cloud/filter/ec2.go
./pkg/cloud/services/mock_services/ec2_interface_mock.go
./pkg/cloud/services/gc/ec2.go
./pkg/cloud/services/instancestate/ec2events.go
./pkg/cloud/services/ec2
```

yiannistri:
**What type of PR is this?**

/kind feature
/kind cleanup

- Migrates ec2 packages to AWS SDK v2.

Fixes #5403 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Migrate EC2 packages to AWS SDK v2
```


damdo:
Hey @yiannistri would you be able to rebase? Thanks!

Makefile
```diff
@@ -606,7 +606,7 @@ promote-images: $(KPROMO) $(YQ)
 
 .PHONY: release-binaries
 release-binaries: $(GORELEASER) ## Builds only the binaries, not a release.
-	$(GORELEASER) build --config $(GORELEASER_CONFIG) --snapshot --clean
+	GOMAXPROCS=2 $(GORELEASER) build --config $(GORELEASER_CONFIG) --snapshot --clean
```
nrb: Why this change?

cmd/clusterawsadm/ami/copy.go
```diff
@@ -175,52 +169,57 @@ func copyWithSnapshot(input copyWithSnapshotInput) (string, string, error) {
...
 	copyInput := &ec2.CopySnapshotInput{
-		Description:       input.image.Description,
-		DestinationRegion: aws.String(input.destinationRegion),
-		DryRun:            aws.Bool(input.dryRun),
-		Encrypted:         aws.Bool(input.encrypted),
-		SourceRegion:      aws.String(input.sourceRegion),
-		KmsKeyId:          kmsKeyIDPtr,
-		SourceSnapshotId:  input.image.BlockDeviceMappings[0].Ebs.SnapshotId,
+		Description:      input.image.Description,
+		DryRun:           aws.Bool(input.dryRun),
```
nrb, line 173: ~Looks like `DestinationRegion` was removed; is it gone from the new version of the API?~

EDIT: Oh, I see, it's put on as part of a presigned request. 👍

pkg/cloud/services/common/common.go
```diff
@@ -0,0 +1,95 @@
...
+package common
+
+import (
+	"context"
+
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+)
+
+// EC2API defines the EC2 API interface.
+type EC2API interface {
```
nrb, line 27: This should probably still live in the `cloud/services/ec2/service.go` file, since there's no other interface being defined, plus it would mean we don't have to update the imports in other files.

Other changes for SDK v2 also keep their updated interfaces in their respective files, like https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5526, https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5497, and https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5389.

pkg/cloud/services/ec2/instances.go
```diff
@@ -906,26 +909,18 @@ func (s *Service) SDKToInstance(v *ec2.Instance) (*infrav1.Instance, error) {
...
-			metadataOptions.HTTPPutResponseHopLimit = *v.MetadataOptions.HttpPutResponseHopLimit
-		}
-		if v.MetadataOptions.HttpTokens != nil {
-			metadataOptions.HTTPTokens = infrav1.HTTPTokensState(*v.MetadataOptions.HttpTokens)
-		}
-		if v.MetadataOptions.InstanceMetadataTags != nil {
-			metadataOptions.InstanceMetadataTags = infrav1.InstanceMetadataState(*v.MetadataOptions.InstanceMetadataTags)
-		}
+		metadataOptions.HTTPEndpoint = infrav1.InstanceMetadataState(string(v.MetadataOptions.HttpEndpoint))
+		metadataOptions.HTTPPutResponseHopLimit = int64(*v.MetadataOptions.HttpPutResponseHopLimit)
```
nrb: This one would still need a nil check since it's de-referencing a pointer.

Makefile
yiannistri: I noticed this make target failing in CI (although looking back in CI history I can't find an example of that) and it looked like `goreleaser` would hang when building the binaries. I tried changing different things in `.goreleaser.yaml` to no effect until I came across this [issue](https://github.com/aws/aws-sdk-go-v2/issues/2871). This got me thinking that referencing the ec2 part of the aws-sdk-go-v2 lib from the code may cause failures when run on smaller machines. Setting the GOMAXPROCS env var to 2 (as described [here](https://goreleaser.com/customization/builds/go/#gomaxprocs-for-go-build-commands)) allowed the build to complete and artifacts were generated.

yiannistri: I can remove this if you prefer and assess the effect afterwards as it's a rather odd bug/fix.

pkg/cloud/services/common/common.go
yiannistri: I seem to recall that I had to move this here (or rather to a "common" location) to break a cyclic dependency. Let me try your suggestion to see what happens and report here.

Makefile
Danil-Grigorev: @nrb this resolved stuck builds for goreleaser in #5518 too. I was able to reproduce it locally, happens only in concurrent build.

damdo:
@yiannistri would you be able to rebase? Thanks!

pkg/cloud/scope/clients.go
```diff
@@ -73,19 +73,21 @@ func NewASGClient(scopeUser cloud.ScopeUsage, session cloud.Session, logger logg
...
+
+	ec2opts := []func(*ec2.Options){
+		func(o *ec2.Options) {
+			o.Logger = logger.GetAWSLogger()
+			o.ClientLogMode = awslogs.GetAWSLogLevelV2(logger.GetLogger())
+		},
+		ec2.WithAPIOptions(
+			awsmetricsv2.WithMiddlewares(scopeUser.ControllerName(), target),
+			awsmetricsv2.WithCAPAUserAgentMiddleware(),
+		),
```
punkwalker: @yiannistri Can you create EndpointResolver like this - https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/endpointsv2/endpoints.go#L109-L123

punkwalker: Also, the serviceLimiter implementation is also required.

pkg/utils/utils.go
```diff
@@ -36,6 +36,14 @@ func ToInt64Pointer(to *int32) *int64 {
...
 }
 
+// ToInt64Value converts an int32 pointer to an int64 value.
+func ToInt64Value(to *int32) int64 {
+	if to == nil {
+		panic("Cannot convert nil pointer to int64")
+	}
+	return int64(*to)
+}
+
```
punkwalker, line 46: Can we not use aws.Int64 here?

punkwalker (changes requested):
@yiannistri 
Thank you for working on this.
I am late to party but these changes are important for maintaining functionality between SDK v1 and v2

punkwalker:
@yiannistri 
update: Please use the ServiceLimiterMiddleware from PR #5574. You can find more details in it.

pkg/utils/utils.go
yiannistri: I don't think we can because the `aws` pointer conversion methods convert to a value of the same bit size as the pointer, i.e. int64 to *int64 and vice versa and in this case we convert from *int32 to int64.

punkwalker: Apologigies, I was suggesting [aws.ToInt64](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2@v1.36.5/aws#ToInt64). 
We could use the same logic with it like `aws.ToInt64(int64(*int32))`. This way we don't need to this additional code/method and we can avoid panicking.

alexander-demicev: isn't this code `aws.ToInt64(int64(*int32))` going to panic too if *int32 is nil?

damdo:
@yiannistri hey would you be able to rebate now that the GC or merged? Thanks!

pkg/utils/utils.go
punkwalker: @alexander-demicev Right, I felt that the usage of this was for IP range of SecurityGroupRule which may not be nil. Anyhow, looks like this method is not used anymore.

pkg/cloud/scope/clients.go
```diff
@@ -78,19 +67,26 @@ func NewASGClient(scopeUser cloud.ScopeUsage, session cloud.Session, logger logg
...
+	ec2opts := []func(*ec2.Options){
+		func(o *ec2.Options) {
+			o.Logger = logger.GetAWSLogger()
+			o.ClientLogMode = awslogs.GetAWSLogLevelV2(logger.GetLogger())
+			o.EndpointResolverV2 = ec2EndpointResolver
+		},
+		ec2.WithAPIOptions(
+			awsmetricsv2.WithMiddlewares(scopeUser.ControllerName(), target),
+			awsmetricsv2.WithCAPAUserAgentMiddleware(),
+		),
```
punkwalker: Can you add the ServiceLimiterMiddleware from PR #5574 . You can find more details in the PR.

pkg/cloud/services/autoscaling/autoscalinggroup.go
```diff
@@ -503,24 +504,24 @@ func mapToTags(input map[string]string, resourceID *string) []autoscalingtypes.T
...
-					Values: aws.StringSlice(eachFilter.Values),
+					Values: eachFilter.Values,
 				})
 			}
 		}
 	}
 
 	if len(inputFilters) > 0 {
-		out, err := s.EC2Client.DescribeSubnetsWithContext(context.TODO(), &ec2.DescribeSubnetsInput{
+		out, err := s.EC2Client.DescribeSubnets(context.TODO(), &ec2.DescribeSubnetsInput{
```
punkwalker, line 523: nit: Can we bring the context from parent calls here instead of using `context.TODO()`? Similarly, should we do it for all such EC2Client. i.e, if context is available in parent method, use it instead of TODO? 

P.S. I know the `context.TODO()` has been around but thought of improving it while we are at it :)

punkwalker (changes requested):
@yiannistri This is the biggest of V2 migration. Thank you for working on this. I have requested few changes, otherwise this LGTM.

damdo:
Hey @yiannistri it looks like the PR never passed the full E2Es and the EKS E2Es
Screenshot 2025-07-23 at 12 17 07

Whereas the periodics against main for those tests have good pass rates:
- https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/periodic-cluster-api-provider-aws-eks-e2e
- https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/logs/periodic-cluster-api-provider-aws-e2e

I suspect there might be something either causing issues, or exacerbating flakes and making them permafailures.

Would you be able to take a look at the e2e test, maybe you could even try running it from your machine if that helps debugging. Thanks!

sowjanyadreddy:
Looks like this issue also addressed changes needed for network code in https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5409 . cc : @richardcase

richardcase:
> Looks like this issue also addressed changes needed for network code in #5409 . cc : @richardcase

@yiannistri - i added that issue to the list of fixes in the description.

damdo:
/assign @richardcase @nrb @AndiDog @dlipovetsky 

For approval

damdo:
All e2es are finally passing 🎉 
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5521/pull-cluster-api-provider-aws-e2e/1949419816922124288
- https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5521/pull-cluster-api-provider-aws-e2e-eks/1949342928438562816

nrb:
/approve

Thank you for persevering on this change! It's hugely helpful.

PR #5173, opened 2024-10-22
✨ Cancel instance refresh on any relevant change to ASG instead of blocking until previous one is finished (which may have led to failing nodes due to outdated join token)

AndiDog:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

Changing any relevant `spec.*` for an `AWSMachinePool` triggers rolling of nodes via ASG instance refresh. If another change happens shortly afterwards, it has to wait until the first rollout is done, and will then trigger another instance refresh. But it is neither necessary nor desired to roll all worker nodes twice in such a case, and it's much slower. Instead, cancel the one pending instance refresh, wait until another one can be started, and apply the latest change as soon as possible with the second instance refresh.

This change has been running fine in Giant Swarm's CAPA fork for three months at the time of opening this PR.

**Special notes for your reviewer**:

This PR stacks on top of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5148, so _let's please review and merge that other PR first_. After that's done, _this_ PR can be retargeted to `main`. I didn't want to separate these independent changes because otherwise I have to deal with merge conflicts.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [x] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Cancel instance refresh on any relevant change to ASG instead of blocking until previous one is finished
```

PR #5543, opened 2025-06-11
✨ Cancel instance refresh on any relevant change to ASG instead of blocking until previous one is finished (which may have led to failing nodes due to outdated join token)

AndiDog:
(reopened from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5173 because I didn't have enough permissions to force-push)

**What type of PR is this?**

Changing any relevant `spec.*` for an `AWSMachinePool` triggers rolling of nodes via ASG instance refresh. If another change happens shortly afterwards, it has to wait until the first rollout is done, and will then trigger another instance refresh. But it is neither necessary nor desired to roll all worker nodes twice in such a case, and it's much slower. Instead, cancel the first pending instance refresh, wait until another one can be started, and apply the latest change as soon as possible with the second instance refresh.

This change has been running fine in Giant Swarm's CAPA fork for almost a year at the time of opening this PR.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Cancel instance refresh on any relevant change to ASG instead of blocking until previous one is finished
```


AndiDog:
@richardcase I think you have some machine pool experience and could maybe review this?

AndiDog:
Rebased onto AWS SDK Go v2 changes

AndiDog:
Rate limit errors in the test, trying again

fiunchinho (approved):
/lgtm

richardcase:
@AndiDog - do you think there is any benefit in adding an e2e that covers this scenario?

AndiDog:
> @AndiDog - do you think there is any benefit in adding an e2e that covers this scenario?

@richardcase I'd say the feature is not that critical to require an E2E test. The behavior is well-covered in `exp/controllers/awsmachinepool_controller_test.go` (and I'm saying that as a huge fan of those quickly-running, mock-based unit tests because they're much easier to write).

AndiDog:
@richardcase Do you see the E2E test as required? Or other concerns for the merge?

Btw this feature has been working for much over a year in our fork, so I'm very confident.

Ankitasw:
/approve

If E2E is needed, we can add it in follow up PR

PR #5553, opened 2025-06-17
Add support for EKSConfig LaunchTemplate bootstrapping for AL2023 using nodeadm

AmitSahastra:
# Add support for EKSConfig LaunchTemplate bootstrapping for AL2023 using nodeadm

## Description
This PR adds support for using launch template EKSConfig bootstrapping for Amazon Linux 2023 nodes. The EKSConfig controller now able to create bootstrap datasecrete with nodeConfig that will enable using AL2023 images with LaunchTemplates.

## Changes
- Modified EKSConfig controller to fetch AMI ID from AWSManagedMachinePool's launch template
- Added proper condition handling for control plane readiness
- Ensured proper reconciliation when dependencies (AWSManagedControlPlane and AWSManagedMachinePool) become ready

## Testing
- Verified that EKSConfig controller correctly retrieves AMI ID from launch template
- Confirmed that node labels are properly set with the AMI ID
- Tested reconciliation behavior when dependencies are not ready
- Validated that controller proceeds with data secret creation once dependencies are ready

## Impact
This changes will allow user to use launch templates with AL2023 images. Also to specify custom AMIs through launch templates while maintaining compatibility with CAPA's auto AMI lookup mechanism.

## Related Issues
Fixes #[issue_number]

**What type of PR is this?**

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note

```


bootstrap/eks/controllers/eksconfig_controller.go
```diff
@@ -271,22 +309,99 @@ func (r *EKSConfigReconciler) joinWorker(ctx context.Context, cluster *clusterv1
...
+			if err != nil {
+				log.Error(err, "Failed to create AWS session for EKS API")
+				conditions.MarkFalse(config, eksbootstrapv1.DataSecretAvailableCondition,
+					eksbootstrapv1.DataSecretGenerationFailedReason,
+					clusterv1.ConditionSeverityWarning,
+					"Failed to create AWS session: %v", err)
+				return ctrl.Result{}, err
+			}
+			eksClient := eks.New(sess)
+			describeInput := &eks.DescribeClusterInput{Name: aws.String(controlPlane.Spec.EKSClusterName)}
```
k0da: I just tested this code in my setup.

This doesn't work if cluster is deployed into different account than controller.

Perhaps we can get CA from kubeconfig secret which is already generated?

AmitSahastra: Good catch will update code.

k0da: I tested this way it works for me https://gist.github.com/k0da/aae082096faf8bfc11699d6ee33e6e96

k0da: The other thing I'm struggling with is my AWSManagedControlPlane have empty `spec.controlPlane.Host`. Maybe we should get endpoint from kubeconfig as well. But it should be a bug lying somewhere else. JFYI

AmitSahastra: I have updated pr with few changes and its working as of now.

AmitSahastra: > I tested this way it works for me https://gist.github.com/k0da/aae082096faf8bfc11699d6ee33e6e96

its using EKSConfig objectMetada here `extractClusterOwner(&config.ObjectMeta)` it shouldn't have ownerObject ref for `owner.Kind == "Cluster" `.

The EKSConfig object doesn't have a direct owner reference to the Cluster - it typically has an owner reference to the Machine or MachinePool object, which in turn is owned by the Cluster.

k0da: @AmitSahastra 
This is what I have:
```yaml
- apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
  kind: EKSConfig
  metadata:
    annotations:
      cluster.x-k8s.io/cloned-from-groupkind: EKSConfigTemplate.bootstrap.cluster.x-k8s.io
      cluster.x-k8s.io/cloned-from-name: eks-config
    creationTimestamp: "2025-07-10T19:51:43Z"
    generation: 1
    labels:
      cluster.x-k8s.io/cluster-name: test-pool-management
      topology.cluster.x-k8s.io/owned: ""
      topology.cluster.x-k8s.io/pool-name: eks-asg
    name: test-pool-management-eks-asg-r5rjn
    namespace: aws
    ownerReferences:
    - apiVersion: cluster.x-k8s.io/v1beta1
      kind: Cluster
      name: test-pool-management
      uid: 7b13697b-c927-4192-90f3-3d868694973c
    - apiVersion: cluster.x-k8s.io/v1beta1
      blockOwnerDeletion: true
      controller: true
      kind: MachinePool
      name: test-pool-management-eks-asg-b627d
      uid: f9561baa-ad2a-4aa4-9170-71842917cf52
```

AmitSahastra: In this solution, since we already have access to the Cluster object in joinWorker, I assume we can use it directly.
Is there a specific use case for relying on the Cluster’s ownerReference instead?

k0da: We just need to find kubeconfig with cluster.Name-kubeconfig

k0da:
LGTM. Running this in my dev environment. So far no further issues.

bootstrap/eks/internal/userdata/node.go
```diff
@@ -138,3 +221,88 @@ func NewNode(input *NodeInput) ([]byte, error) {
...
+	}
+
+	if input.MaxPods == nil {
+		if input.UseMaxPods != nil && *input.UseMaxPods {
+			input.MaxPods = ptr.To[int32](58)
+		} else {
+			input.MaxPods = ptr.To[int32](110)
+		}
+	}
+	if input.DNSClusterIP == nil {
```
k0da: DNSCluster IP is an override. So it is quite optional. It is not mandatory to specify.

rudimk:
@AmitSahastra bit curious about the use of `EKSConfig` here. Right now I've got about a hundred plus clusters on CAPA that are using `AWSManagedMachinePools` but I'm not using an `EKSConfig` with them. I guess my question is - will I have to plug in an `EKSConfig` for every cluster, in order to make the switch to AL2023? Also kind of curious as to how that might impact existing managed nodegroups running off AL2. 

Really appreciate your work on this, looking forward to having it out soon!

AmitSahastra:
> @AmitSahastra bit curious about the use of `EKSConfig` here. Right now I've got about a hundred plus clusters on CAPA that are using `AWSManagedMachinePools` but I'm not using an `EKSConfig` with them. I guess my question is - will I have to plug in an `EKSConfig` for every cluster, in order to make the switch to AL2023? Also kind of curious as to how that might impact existing managed nodegroups running off AL2.
> 
> Really appreciate your work on this, looking forward to having it out soon!

Had you enable launch template in AWSMMP it should create EKS config.

rudimk:
Aye we are using launch templates in our managed machine pools using the AWSLaunchTemplate parameter. Pretty certain no corresponding EKSConfig objects are being created - because I don't see any on our management cluster - hence my question around whether one needs to create EKSConfig objects for each cluster.

AmitSahastra:
Yes, for AL2023 support via CAPA, I added logic that uses nodeType: al2023 inside EKSConfig to generate the correct nodeadm-style userdata. So if you’re planning to switch to AL2023 and rely on CAPA for bootstrap, then yes, you’ll need to start creating EKSConfig objects with that field set.

As for existing AL2-based nodegroups, there’s no impact — they can stay as-is with your current launch templates and don’t require any changes unless you explicitly migrate them to AL2023.

I hope that answers your question.

rudimk:
That's baller - thanks a ton @AmitSahastra, appreciate you!

mloiseleur:
I have built and tested this PR on a sandbox environment for our use case. 
~~Maybe it works for `AWSManagedMachinePool`, but it does not support `AWSManagedControlPlane` with `MachineDeployment` and `EKSConfigTemplate`.~~

~~It still creates instances with previous format and so, fails to connect.~~

<details>

<summary>Example YAML</summary>

```yaml
---
kind: AWSMachineTemplate
metadata:
  name: test-3723977939
  namespace: flux-system
spec:
  template:
    spec:
      ami:
        eksLookupType: AmazonLinux2023
      instanceMetadataOptions:
        httpTokens: required
        httpPutResponseHopLimit: 2
      iamInstanceProfile: test-capa-nodes
      instanceType: t4g.medium
---
apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
kind: EKSConfigTemplate
metadata:
  name: test-al2023
  namespace: flux-system
spec:
  template:
    spec:
      nodeType: al2023
---
kind: MachineDeployment
metadata:
  name: t00-use1-eks-test-us-east-1a-md0
  namespace: flux-system
spec:
  clusterName: test
  template:
    spec:
      bootstrap:
        configRef:
          apiVersion: bootstrap.cluster.x-k8s.io/v1beta2
          kind: EKSConfigTemplate
          name: test-al2023
      clusterName: test
      infrastructureRef:
        kind: AWSMachineTemplate
        name: test-3723977939
      version: v1.32.3
```

</details>

<details>

<summary>CAPA Logs</summary>

```console
I0804 13:55:29.740725       1 eksconfig_controller.go:265] "Generating userdata" 
I0804 13:55:29.740740       1 eksconfig_controller.go:314] "Processing AL2023 node type" 
I0804 13:55:29.741301       1 eksconfig_controller.go:366] "Generating AL2023 userdata" 
I0804 13:55:29.748003       1 eksconfig_controller.go:443] "created bootstrap data secret for EKSConfig" 
I0804 13:55:29.766772       1 eksconfig_controller.go:197] "joinWorker called" [repeated multiple times, without success]
```

The instance is launched successfully, but with previous metadata format and so, it's not able to connect:

```console
I0804 13:55:30.090681       1 awsmachine_controller.go:732] "Creating EC2 instance"
I0804 13:55:30.306418       1 instances.go:135] "Obtained a list of supported architectures for instance type"  [...] supported architectures=["arm64"]
I0804 13:55:30.307461       1 instances.go:135] "Chosen architecture" [...]  instance type="t4g.medium" supported architectures=["arm64"] architecture="arm64"
I0804 13:55:30.353337       1 ami.go:374] "found AMI" [...] id="ami-0cbd69473baf7c079" version="1.32"
I0804 13:55:32.550263       1 awsmachine_controller.go:578] "EC2 instance state changed" state="pending" instance-id="i-xxx
I0804 13:56:03.485209       1 awsmachine_controller.go:578] "EC2 instance state changed" state="running" instance-id="i-xxx"
```

</details>

~~After further investigation, I can be more precise. The secret is created as expected, with the expected content.~~
~~The vm is deployed with userdata created by `pkg/cloud/services/secretsmanager/secret.go` (template is in `pkg/cloud/services/secretsmanager/secret_fetch_script.go`). So far, my guess is that it's supposed to get the secret from SecretManager, and that's where it's failing.~~

**EDIT**: It works when disabling Secrets Manager on user metadata in `AWSMachineTemplate`:

```diff
@@ -11,6 +11,8 @@
       {{- else }}
         eksLookupType: AmazonLinux2023
       {{- end }}
+      cloudInit:
+        insecureSkipSecretsManager: true
       instanceMetadataOptions:
         httpTokens: required
         httpPutResponseHopLimit: 2
```

mloiseleur:
@AmitSahastra I've done a review with a [commit](https://github.com/kubernetes-sigs/cluster-api-provider-aws/commit/546432d3cce8755cf3071612afa6e169793523ec) on my fork https://github.com/mloiseleur/cluster-api-provider-aws/tree/al2023-launch-template 

What I've done:

1. Fix maxPods (it was setting to 58 when setting the boolean in the CRD to true)
2. Make clusterDNS optional (following [this comment](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5553#discussion_r2216358747), confirmed with my test that it works as expected)
3. Move node-labels from gotemplate to go func, for readability
4. Add usage documentation in the book
5. Tested everything on a sandbox env

May I invite you to add it to your PR ? You can apply this [file](https://github.com/user-attachments/files/21614428/0001-review.patch.zip) or cherrypick from my branch.

AmitSahastra:
Thanks @mloiseleur for the detailed review and improvements! I’ve cherry-picked your changes on maxPods, clusterDNS, and the node-labels refactor, and added them to the PR. Also appreciated the test validation and doc additions. 🙌

bootstrap/eks/internal/userdata/node_test.go
```diff
@@ -386,3 +387,189 @@ users:
...
+				NodeGroupName:     "test-nodegroup",
+				UseMaxPods:        ptr.To[bool](true),
+				DNSClusterIP:      ptr.To[string]("10.100.0.10"),
+				AMIImageID:        "ami-123456",
+				ClusterCIDR:       "192.168.0.0/16",
+			},
+			expectErr: false,
+			verifyOutput: func(output string) bool {
+				return strings.Contains(output, "cidr: 192.168.0.0/16") &&
+					strings.Contains(output, "maxPods: 58") &&
```
mloiseleur:
```suggestion
					strings.Contains(output, "maxPods: 110") &&
```
It should be 110 when useMaxPods is set. 
That should fix the CI.

bootstrap/eks/api/v1beta2/eksconfig_types.go
```diff
@@ -24,6 +24,9 @@ import (
 
 // EKSConfigSpec defines the desired state of Amazon EKS Bootstrap Configuration.
 type EKSConfigSpec struct {
+	// NodeType specifies the type of node (e.g., "al2023")
+	// +optional
+	NodeType string `json:"nodeType,omitempty"`
```
jimmidyson: Is there any way to derive this from the AMI being used rather than asking the user to specify in the API?

jimmidyson: This is only required when using AL2023 so perhaps let's add an enum for this and only accept `al2023`? Something like this:

```suggestion
// +kubebuilder:validation:Enum=al2023
type NodeType string

const (
	NodeTypeAL2023 = "al2023
)

// EKSConfigSpec defines the desired state of Amazon EKS Bootstrap Configuration.
type EKSConfigSpec struct {
	// NodeType specifies the type of node (e.g., "al2023")
	// +optional
	NodeType NodeType `json:"nodeType,omitempty"`
```

AmitSahastra: Thanks for suggetion, I was using constans for NodeTypeAL2023 already, 
There was systax error in suggested commits so included suggetions manually, that is enum into code and crds as well.

bootstrap/eks/controllers/eksconfig_controller.go
```diff
@@ -251,7 +281,9 @@ func (r *EKSConfigReconciler) joinWorker(ctx context.Context, cluster *clusterv1
 		DiskSetup:                config.Spec.DiskSetup,
 		Mounts:                   config.Spec.Mounts,
 		Files:                    files,
+		ClusterCIDR:              controlPlane.Spec.NetworkSpec.VPC.CidrBlock,
```
jimmidyson: This should be the service CIDR of the cluster as per [documentation](https://github.com/awslabs/amazon-eks-ami/blob/v20250813/nodeadm/api/v1alpha1/nodeconfig_types.go#L52-L53):

```
// CIDR is your cluster's service CIDR block. This value is used to infer your cluster's DNS address.
CIDR string `json:"cidr,omitempty"`
```

Pushed a PR targeting your PR branch - https://github.com/AmitSahastra/cluster-api-provider-aws/pull/2.

rudimk:
Hey folks - curious as to when we're looking at seeing this merged / released?

bootstrap/eks/controllers/eksconfig_controller.go
jimmidyson: @AmitSahastra Could you take a look at this please?

damdo:
/assign @damdo @nrb @richardcase @AndiDog @dlipovetsky @Ankitasw 

What do you think of this?

bootstrap/eks/api/v1beta2/eksconfig_types.go
```diff
@@ -22,8 +22,20 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
 )
 
+// NodeType specifies the type of nodeq
+// +kubebuilder:validation:Enum=al2023
+type NodeType string
```
damdo, line 27: See my comments below regarding the naming of this field

bootstrap/eks/api/v1beta2/eksconfig_types.go
```diff
@@ -22,8 +22,20 @@ import (
...
+
+const (
+	// NodeTypeAL2023 represents the AL2023 node type.
+	NodeTypeAL2023 NodeType = "al2023"
+)
```
damdo, line 32: Same here a NodeTypeAL2

damdo: We could use the ones you defined for the AMIFamilyType also here
```
	// AMIFamilyAL2 is the Amazon Linux 2 AMI family.
	AMIFamilyAL2 = "AmazonLinux2"
	// AMIFamilyAL2023 is the Amazon Linux 2023 AMI family.
	AMIFamilyAL2023 = "AmazonLinux2023"
```

bootstrap/eks/api/v1beta2/eksconfig_types.go
```diff
@@ -22,8 +22,20 @@ import (
...
+
 // EKSConfigSpec defines the desired state of Amazon EKS Bootstrap Configuration.
 type EKSConfigSpec struct {
+	// NodeType specifies the type of node (e.g., "al2023")
+	// +optional
+	NodeType NodeType `json:"nodeType,omitempty"`
```
damdo, line 38: I think `NodeType` is a bit weak as a field name because Type is quite generic.
How about AMIFamilyType like the field we have in the NodeInput struct? So then we can directly pass that down to it?

damdo: We would also need to specify what the default is (I guess `AmazonLinux2`) and an enum validation for this.
```
	// +kubebuilder:validation:Enum="";AmazonLinux2;AmazonLinux2023
	// +optional
```

bootstrap/eks/internal/userdata/node.go
```diff
@@ -19,15 +19,25 @@ package userdata
...
 )
 
 const (
 	defaultBootstrapCommand = "/etc/eks/bootstrap.sh"
+	boundary                = "//"
+
+	// AMIFamilyAL2 is the Amazon Linux 2 AMI family.
+	AMIFamilyAL2 = "AmazonLinux2"
+	// AMIFamilyAL2023 is the Amazon Linux 2023 AMI family.
+	AMIFamilyAL2023 = "AmazonLinux2023"
```
damdo, line 40: Let's move them in the outer API and leverage them here too

bootstrap/eks/internal/userdata/node.go
```diff
@@ -66,6 +120,27 @@ type NodeInput struct {
...
+	AMIImageID        string
+	APIServerEndpoint string
+	Boundary          string
+	CACert            string
+	CapacityType      *v1beta2.ManagedMachinePoolCapacityType
+	ClusterCIDR       string // CIDR range for the cluster
+	ClusterDNS        string
+	MaxPods           *int32
+	NodeGroupName     string
+	NodeLabels        string // Not exposed in CRD, computed from user input
```
damdo: Hm, is there a way to make this more specific to AmazonLinux2023 if that's only for it?
How about a discriminated union?
Have we thought about what's going to happen if there is going to be, for example an AmazonLinux2027 or similar, which may have different fields from this one?

docs/book/src/topics/eks/enabling.md
```diff
@@ -56,3 +56,40 @@ clusterctl init --infrastructure aws
 ```
 
 NOTE: you will need to enable the creation of the default Fargate IAM role. The easiest way is using `clusterawsadm` and using the `fargate` configuration option, for instructions see the [prerequisites](../using-clusterawsadm-to-fulfill-prerequisites.md).
+
+### Amazon Linux 2023
```
damdo, line 60: I'd call it something more generic depending on what we agree on for the API, like AMIFamilyType, and then we can have various subsections for each type.

bootstrap/eks/controllers/eksconfig_controller_reconciler_test.go
```diff
@@ -199,6 +203,70 @@ func TestEKSConfigReconciler(t *testing.T) {
 			gomega.Expect(string(secret.Data["value"])).To(Not(Equal(string(expectedUserData))))
 		}).Should(Succeed())
 	})
+
+	t.Run("Should return requeue when control plane is not initialized", func(t *testing.T) {
```
damdo, line 207: We can also add webhook tests for the values we are going to be validating for the AMIFamilyType

bootstrap/eks/controllers/eksconfig_controller_reconciler_test.go
```diff
@@ -199,6 +203,70 @@ func TestEKSConfigReconciler(t *testing.T) {
...
+		g := NewWithT(t)
+		amcp := newAMCP("test-cluster")
+		cluster := newCluster(amcp.Name)
+		machine := newMachine(cluster, "test-machine")
+		config := newEKSConfig(machine)
+
+		// Set node type to AL2023 to trigger requeue
+		config.Spec.NodeType = "al2023"
```
damdo, line 215: Will need changing based on the API we agree on

bootstrap/eks/internal/userdata/node.go
```diff
@@ -138,3 +224,106 @@ func NewNode(input *NodeInput) ([]byte, error) {
...
+		return "ON_DEMAND"
+	default:
+		return strings.ToUpper(string(*ni.CapacityType))
+	}
+}
+
+// validateAL2023Input validates the input for AL2023 user data generation.
+func validateAL2023Input(input *NodeInput) error {
+	if input.APIServerEndpoint == "" {
+		return fmt.Errorf("API server endpoint is required for AL2023")
```
damdo, line 297: Everywhere in this PR it would be better to use the full extended AmazonLinux2023, rather than AL2023, same goes for AL2 vs AmazonLinux2.

damdo (commented):
Thanks for the PR, I think we polish a bit the API first, then we can adjust the logic according to it

bootstrap/eks/api/v1beta2/eksconfig_types.go
damdo: It looks like we allow all of these?
https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5555/files#diff-c82f4c3e71cee496ae15d0bb197a279d5911a9b5b86282fa3b6f4900673078e2R459-R469

rudimk:
Hey @damdo just want to flag that AWS is setting a deadline on this that's about 2.5 months away. Merging this in, cutting a new release, deploying that release in-house and testing it before we roll this out to clusters on our end is bound to take some time. Don't want to sound pushy but if we could figure out what kind of timeline we're looking at here, that'll help us determine what the best course of action is. 

Thanks so much!

bootstrap/eks/api/v1beta2/eksconfig_types.go
jimmidyson: The enum is added on the type definition (https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5553/files#diff-64e2d3d6a1323c5ae37a84cfc459f926fe7e329e83a4f695df28de89f19a3172R26). Only `al2023` is allowed at this point, no other values are required to change behaviour, default (unset) is current behaviour.

bootstrap/eks/controllers/eksconfig_controller.go
AmitSahastra: Merged, Thanks!

bootstrap/eks/api/v1beta2/eksconfig_types.go
faiq: This API is a little confusing to me. We're essentially specifying an AMI family here. We have a field like this already in AWSMachine 

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/cfdcb142e2df59e120457c28fac499c8f7cbff97/api/v1beta2/types.go#L59


I think we shouldn't repeat this information here.

faiq: Another thing that's worth noting here is that Amazon Linux 2 will be EOL coming up very soon - june 2026 source: https://aws.amazon.com/amazon-linux-2/faqs/

It seems like EKS has largely moved away from this AL2 support to either AL2023 or bottlerocket. 

Bottlerocket seems to have its own set backs see https://github.com/kubernetes-sigs/cluster-api/issues/7840

so realistically the _only_ option is going to be AL2023. 


EDIT: 
> [!IMPORTANT] Amazon EKS will no longer publish EKS-optimized Amazon Linux 2 (AL2) AMIs after November 26th, 2025. Additionally, Kubernetes version 1.32 is the last version for which Amazon EKS will release AL2 AMIs. From version 1.33 onwards, Amazon EKS will continue to release AL2023 and Bottlerocket based AMIs. To learn more, see [Guide to EKS AL2 & AL2-Accelerated AMIs transition features](https://docs.aws.amazon.com/eks/latest/userguide/eks-ami-deprecation-faqs.html).

source: https://awslabs.github.io/amazon-eks-ami/

it looks like they won't be building this AMI

bootstrap/eks/controllers/eksconfig_controller.go
```diff
@@ -271,22 +308,85 @@ func (r *EKSConfigReconciler) joinWorker(ctx context.Context, cluster *clusterv1
...
+					eksbootstrapv1.DataSecretGenerationFailedReason,
+					clusterv1.ConditionSeverityWarning,
+					"Failed to extract CA from kubeconfig secret: %v", err)
+				return ctrl.Result{}, err
+			}
+			nodeInput.CACert = ca
+		}
+
+		// Get AMI ID from AWSManagedMachinePool's launch template if specified
+		if configOwner.GetKind() == "AWSManagedMachinePool" {
```
faiq, line 344: Is there any reason that users who don' use managed machine pools can't set AMI IDs like users of `AWSMachineTemplate`?

faiq:
Generally speaking I think we can simplify this to just support the nodeadm script because the other stuff isn't going to be applicable. Please take a look at some of the work I did here that attempts to simplify some of this stuff https://github.com/faiq/cluster-api-provider-aws/commit/cf6a2805ee4ea723129a79f126e24c9d1bcb699b

jimmidyson:
@faiq 

> Generally speaking I think we can simplify this to just support the nodeadm script because the other stuff isn't going to be applicable.

AFAIK nodeadm will only apply to AL2023 so if using other distros they'll still use cloud-init. I think so anyway 😅

faiq:
@jimmidyson 

> AFAIK nodeadm will only apply to AL2023 so if using other distros they'll still use cloud-init. I think so anyway 😅

I think they'll still be using nodeadm :thinking: so from my POV there's 2 things happening here

1. The old bootstrap mechanism (bootstrap.sh) is being deprecated in favor of the new `nodeadm` tool. 
2. Because EKS mostly works with the images they provide what I think is happening here is that we're pairing the Operating Systems with the bootstrap methods -- ie: AL2 + bootstrap.sh and AL2023 + nodeadm. I think the bootstrap methods are independent of the OS and `nodeadm` is what they're using going forward. 

i could be wrong, but that's how im interpreting the situation :man_shrugging:

bootstrap/eks/internal/userdata/node.go
```diff
@@ -41,9 +51,53 @@ runcmd:
...
+set -o errexit
+set -o pipefail
+set -o nounset
+{{- if or .PreBootstrapCommands .PostBootstrapCommands }}
+
+{{- range .PreBootstrapCommands}}
+{{.}}
+{{- end}}
+{{- range .PostBootstrapCommands}}
+{{.}}
```
faiq, line 69: [Cloud-init handles the `text/x-shellscript` content type](https://docs.cloud-init.io/en/23.1/explanation/format.html#user-data-script) and all of that is going to [run before Nodeadm runs](https://github.com/awslabs/amazon-eks-ami/blob/main/templates/al2023/runtime/rootfs/etc/systemd/system/nodeadm-run.service#L6) to join the cluster. I think we should remove the `PostBootstrapCommands` because of this.

Running the nodeadm command and then running the postBootstrapCommands will likely be very cumbersome (disabling the service) and subject to a lot of downstream changes which will be difficult to maintain

faiq:
Hey @AmitSahastra after looking and thinking about this PR a little bit more I think the right approach here would be to create a new API type that handles bootstrapping for `Nodeadm`. 

The `EKSConfig` type was written as an API adapter layer above the old `bootstrap.sh` bootstrapping method. This current PR is essentially adding a new implementation for nodeadm, but there are a lot of fields that `nodeadm` supports (see [API reference](https://awslabs.github.io/amazon-eks-ami/nodeadm/doc/api/)) that `bootstrap.sh` doesn't support and [vice versa ](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5553#discussion_r2323459300)

I think the cleanest way to support this would be to create a new type so when Amazon Linux and `bootstrap.sh` get no longer supported by EKS we can cleanly remove that code. In the meantime, we can support both using the two different bootstrap methods.

faiq:
I wrote this document that outlines the need for a new bootstrap API here https://docs.google.com/document/d/1n2v0Q4D5VAzydu7aJno3gRjMlSHueJ8BlGIOrKLDaiQ/edit?tab=t.0#heading=h.ly39fzd1bgc0

pavansokkenagaraj:
> I wrote this document that outlines the need for a new bootstrap API here https://docs.google.com/document/d/1n2v0Q4D5VAzydu7aJno3gRjMlSHueJ8BlGIOrKLDaiQ/edit?tab=t.0#heading=h.ly39fzd1bgc0

@damdo @faiq @jimmidyson 
Native support for Bottlerocket in Managed Node Groups with EKS has been around since 2022(AFAIK).
If we are designing a new API, Bottlerocket configuration should be considered.

faiq:
@pavansokkenagaraj that's a good point however it seems that bottlerocket support in cluster API is set back due to https://github.com/kubernetes-sigs/cluster-api/issues/7840 as well as this https://github.com/kubernetes-sigs/cluster-api/issues/5294 

Given that it's not going to use a cloud-init based configuration, we should consider that out of the scope of this new API

dsanders1234:
The deadline for upgrading from Amazon EKS Amazon Linux 2 (AL2) to Amazon Linux 2023 (AL2023) is November 26, 2025. I'm assuming there is no workaround for people using managed nodegroups and AL2.

dlipovetsky:
Please see the [proposal](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5678) PR for a separate NodeadmConfig.

hsmade:
Am I wrong to assume here that both bottlerocket and AL2023 are currently not supported by CAPA, so spinning up a 1.34 cluster is not possible?

richardcase:
We need a way forward on this and/or #5700 . A topic for the next office hours?

jimmidyson:
IIUC this PR should be closed in favour of #5700 as that implements the agreed KEP from #5678.

damdo:
Closing in favor of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5700

PR #5560, opened 2025-06-19
Check control plane version skew before creating new launch template …

fiunchinho:
…version

**What type of PR is this?**

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note

```


exp/controllers/awsmachinepool_controller.go
```diff
@@ -442,6 +454,35 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
 	return ctrl.Result{}, nil
 }
 
+// isMachinePoolAllowedToUpgradeDueToControlPlaneVersionSkew checks if the control plane is being upgraded, in which case we shouldn't update the launch template.
+func (r *AWSMachinePoolReconciler) isMachinePoolAllowedToUpgradeDueToControlPlaneVersionSkew(clusterScope cloud.ClusterScoper, machinePoolScope *scope.MachinePoolScope) (bool, error) {
+	if machinePoolScope.Cluster.Spec.ControlPlaneRef == nil {
+		return true, nil
```
richardcase: I don't think we'd want to proceed in this case with the upgrade.

fiunchinho: I changed the logic to return an error in this case, PTAL.

exp/controllers/awsmachinepool_controller.go
```diff
@@ -442,6 +454,35 @@ func (r *AWSMachinePoolReconciler) reconcileNormal(ctx context.Context, machineP
...
+	}
+
+	controlPlane, err := clusterScope.UnstructuredControlPlane()
+	if err != nil {
+		return false, errors.Wrap(err, "failed to get ControlPlane")
+	}
+
+	cpVersion, found, err := unstructured.NestedString(controlPlane.Object, "status", "version")
+	if !found || err != nil {
+		return false, errors.Wrapf(err, "failed to get version of ControlPlane %s", machinePoolScope.Cluster.Spec.ControlPlaneRef.Name)
```
nrb, line 470: With the `||` here, is it possible that we'll get a case where `!found = true && err == nil`? That may cause an issue when generating the wrapped error here.

exp/controllers/awsmachinepool_controller_test.go
```diff
@@ -879,6 +891,65 @@ func TestAWSMachinePoolReconciler(t *testing.T) {
...
+					nil)
+				ec2Svc.EXPECT().DiscoverLaunchTemplateAMI(gomock.Any(), gomock.Any()).Return(ptr.To[string]("ami-existing"), nil)
+				// Mock changes to the launch template, ASG would be updated, but k8s version skew will prevent it
+				ec2Svc.EXPECT().LaunchTemplateNeedsUpdate(gomock.Any(), gomock.Any(), gomock.Any()).Return(true, nil)
+
+				// Control plane kubernetes version skew is checked before, so it won't be called
+				asgSvc.EXPECT().CanStartASGInstanceRefresh(gomock.Any()).Times(0)
+				// Won't be called due to version skew
+				ec2Svc.EXPECT().CreateLaunchTemplateVersion(gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any(), gomock.Any()).Times(0)
+				// Won't be called due to version skew
```
nrb, line 928: Appreciate these comments, thank you!

AndiDog:
We found this doesn't work if the `MachinePool` object still hasn't been updated. The patched logic then thinks `MachinePool.spec.template.spec.version` to be the desired version, but it's outdated and this might again lead to unwanted worker rollouts with too new Kubernetes version.

This patch improves the situation if the objects are deployed around the same time (before reconciliation runs), so I think we could keep it as a first step in the right direction. For a full skew check, we'd need to synchronize the `MachinePool` and `AWSMachinePool` objects, for example using a common field or label.

@fiunchinho One more question on behavior: The patch avoids the instance refresh. But what if the ASG scales up or replaces instances – would they have the new Kubernetes version? Also, we currently get `return errors.New("Cannot update the launch template, prerequisite not met")` – probably fine, but maybe we should fill the condition with a good reason text (returned by `canUpdateLaunchTemplate()`?!).

fiunchinho:
> @fiunchinho One more question on behavior: The patch avoids the instance refresh. But what if the ASG scales up or replaces instances – would they have the new Kubernetes version? Also, we currently get `return errors.New("Cannot update the launch template, prerequisite not met")` – probably fine, but maybe we should fill the condition with a good reason text (returned by `canUpdateLaunchTemplate()`?!).

It prevents creating a new launch template version, so new nodes (even when scaling up) won't get the new k8s version.

PR #5569, opened 2025-06-27
✨Create multiple control plane loadbalancers concurrently

### LINKED ISSUE #5568: Parallel creation of multiple loadbalancers
/kind feature

LoadBalancers take many minutes to become available after creation. We currently wait synchronously for a loadbalancer to become available immediately after creation: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/63b8bcf1b3860347d6f860a3dd5bd582d1899a95/pkg/cloud/services/elb/loadbalancer.go#L128-L133

This is inefficient when creating more than one ControlPlaneLoadBalancer, for example when creating both the internal and external LBs of an OpenShift cluster. Ideally we would create all loadbalancers first, then wait for them to become ready.

**Environment:**

- Cluster-api-provider-aws version: v2.8.1-129-g63b8bcf1b (main)

mdbooth:
**What type of PR is this?**

/kind feature

LoadBalancers take many minutes to become available after creation. We currently wait synchronously for a loadbalancer to become available immediately after creation. This increases total cluster installation time when creating multiple control plane loadbalancers as we don't create a loadbalancer until the previous one is fully initialised. It would be more time-efficient to create them all first, then wait for them to become available.

This PR splits loadbalancer creation into 2 phases:
* Get or create
* Reconcile

We ensure that we do the waiting in the reconcile phase after all getOrCreates have executes.

Fixes #5568

This PR doesn't introduce any parallel execution in CAPA. It simply changes the order of operations so that loadbalancer initialisations can proceed concurrently in AWS.

Apart from a minor refactor to enable re-use, the existing `ensure two load balancers are reconciled` test remains unchanged and continues to pass.

We add a new test, `ensure two load balancers are created concurrently`. This is also a general creation test, which did not previously exist. Concurrency is asserted by asserting that both `WaitUntilLoadBalancerAvailableWithContext` calls happen after both `CreateLoadBalancer` calls.

We don't add a test for classic load balancers, as support for these already seems to be problematic. Note that 2 classic loadbalancers cannot work because `getOrCreateClassicLoadBalancer` (formerly `reconcileClassicLoadBalancer`) does not take an lbSpec argument, therefore it hardcodes the primary load balancer. You could presumably mix a classic primary and v2 secondary LB, but I did not want to add coverage for an esoteric test case.

I believe I also spotted a latent bug in classic load balancer reconciliation where it will always reconcile a HealthCheck even if it is already set. I added a comment about this but did not attempt to fix it because it is unrelated.

**Checklist**:

- [X] squashed commits
- [X] includes documentation
- [X] includes emoji in title 
- [X] adds unit tests

**Release note**:

```release-note
Control plane load balancers are created concurrently, reducing cluster
installation time when specifying a secondary control plane load balancer.
```

### DIFF
--- a/pkg/cloud/services/elb/loadbalancer.go
+++ b/pkg/cloud/services/elb/loadbalancer.go
@@ -67,11 +67,21 @@ const additionalTargetGroupPrefix = "additional-listener-"
 // cantAttachSGToNLBRegions is a set of regions that do not support Security Groups in NLBs.
 var cantAttachSGToNLBRegions = sets.New("us-iso-east-1", "us-iso-west-1", "us-isob-east-1")
 
+type lbReconciler func() error
+
 // ReconcileLoadbalancers reconciles the load balancers for the given cluster.
 func (s *Service) ReconcileLoadbalancers() error {
 	s.scope.Debug("Reconciling load balancers")
 
 	var errs []error
+	var lbReconcilers []lbReconciler
+
+	// The following splits load balancer reconciliation into 2 phases:
+	// 1. Get or create the load balancer
+	// 2. Reconcile the load balancer
+	// We ensure that we only wait for the load balancer to become available in
+	// the reconcile phase. This is useful when creating multiple load
+	// balancers, as they can take several minutes to become available.
 
 	for _, lbSpec := range s.scope.ControlPlaneLoadBalancers() {
 		if lbSpec == nil {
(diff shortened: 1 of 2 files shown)


nrb:
> We don't add a test for classic load balancers, as support for these already seems to be problematic.

Correct. We're phasing out support of classic load balancers as much as possible.

>  You could presumably mix a classic primary and v2 secondary LB, but I did not want to add coverage for an esoteric test case.

This was definitely possible a few releases ago, but since Go dropped older ciphers that classic ELBs supported, we've since defaulted to NLBs.

I added the support for a second LB, and supporting classic there was never the intention.

So I think this test coverage is sufficient, and the change is a pretty easy speed improvement.

/lgtm
/approve

mdbooth:
This is still running after 33 hours! Wonder if we can kick it somehow.

/test pull-cluster-api-provider-aws-build-docker

PR #5580, opened 2025-07-02
🐛  Fix addition logic of getAttemptContextMiddleware

punkwalker:
**What type of PR is this?**
/kind bug

Fixes addition logic of getAttemptContextMiddleware relative to default "Retry" middleware of AWS SDK
Fixes #
NA

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [X] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Fix addition logic of getAttemptContextMiddleware
```

### DIFF
--- a/pkg/cloud/metricsv2/metrics.go
+++ b/pkg/cloud/metricsv2/metrics.go
@@ -24,6 +24,7 @@ import (
 	"time"
 
 	awsmiddleware "github.com/aws/aws-sdk-go-v2/aws/middleware"
+	"github.com/aws/aws-sdk-go-v2/aws/retry"
 	"github.com/aws/smithy-go/middleware"
 	smithyhttp "github.com/aws/smithy-go/transport/http"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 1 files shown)


pkg/cloud/metricsv2/metrics.go
```diff
@@ -100,8 +101,11 @@ func WithMiddlewares(controller string, target runtime.Object) func(stack *middl
 		if err := stack.Finalize.Add(getRequestMetricContextMiddleware(), middleware.Before); err != nil {
 			return fmt.Errorf("failed to add request metric context middleware: %w", err)
 		}
-		if err := stack.Finalize.Insert(getAttemptContextMiddleware(), "Retry", middleware.After); err != nil {
-			return fmt.Errorf("failed to add attempt context middleware: %w", err)
+		// Check if the Default Retry Middleware exists.
+		if _, ok := stack.Finalize.Get((*retry.Attempt)(nil).ID()); ok {
+			if err := stack.Finalize.Insert(getAttemptContextMiddleware(), (*retry.Attempt)(nil).ID(), middleware.After); err != nil {
```
nrb, line 106: Should also check if the AttemptContextMiddleware's in the stack already, otherwise we get:

`"ensuring bucket exists: creating S3 bucket: failed to add attempt context middleware: already exists, capa/AttemptMetricContextMiddleware"`

punkwalker: sure, we can do that but I feel we don't need it as only add this middleware once. Did you make the changes to your local copy of CAPA package?

punkwalker: Confirmed on Slack, this is not needed

damdo:
cc. @JoelSpeed @benluddy

nrb:
We should add an e2e that exercises the code path when ignition is being fetched from the s3 bucket, since that's where this happens.

@damdo @richardcase thoughts on merging this and following up with another PR?

damdo:
This was tested downstream and verified that it works as intended.
That said, yes, it would be nice to have a test that exercises it so we don't regress again.
+1 for merging the fix and then having someone follow up on that test.

damdo (commented):
/lgtm

/assign @richardcase

PR #5591, opened 2025-07-15
:bug: Set conditions to true when reconciliation is successful

fiunchinho:
**What type of PR is this?**

/kind bug

We only set these conditions to false when there is a problem in the reconciliation. But if the reconciliation eventually succeeds, we never set them back to true.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Set conditions to true when reconciliation is successful
```


fiunchinho:
> Looks like a reasonable fix @fiunchinho Do you have visual evidence these have moved back to True?
> 
> /lgtm

Yes, I tested this change in our management cluster.

damdo:
Cool thanks @fiunchinho 

/assign @nrb

nrb (commented):
/approve

AndiDog (commented):
/lgtm

PR #5598, opened 2025-07-23
[WIP]Add support for AMD SEV-SNP instances

fangge1212:
**What type of PR is this?**
/kind feature

This PR adds support for AMD SEV-SNP instances.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```Add support for AMD SEV-SNP instances

```


api/v1beta2/awsmachine_types.go
```diff
@@ -116,6 +132,10 @@ type AWSMachineSpec struct {
 	// +kubebuilder:validation:MinLength:=2
 	InstanceType string `json:"instanceType"`
 
+	// CpuOptions is the set of cpu options for the instance
+	// +optional
+	CpuOptions *CpuOptions `json:"cpuOptions,omitempty"`
```
bgartzi: I'm not personally sure about this one. It mimics the way it is structured on the AWS end, that's okay. But what if TDX support arrives at AWS at some point? I think it could be easier to extend an enum for confidential cpu/vm types (e.g. `Disabled`/`SNP`/`TDX`) than one boolean flag for each supported confidential VM type. This is also the approach that others cluster-api-providers went for IIUC.

In that case, I would personally propose a dedicated enum field such as ConfidentialComputing in the API, that would latter end up configuring the appropriate EC2 CpuOptions. However, let's wait for feedback for people that know about this API and codebase.

api/v1beta2/awsmachine_webhook.go
```diff
@@ -417,6 +420,17 @@ func (r *AWSMachine) validateNonRootVolumes() field.ErrorList {
 	return allErrs
 }
 
+func (r *AWSMachine) validateInstanceTypeForConfidentialCompute() field.ErrorList {
+	var allErrs field.ErrorList
+	if r.Spec.CpuOptions != nil {
+		if r.Spec.CpuOptions.AmdSevSnp != nil && *r.Spec.CpuOptions.AmdSevSnp && !slices.Contains(instanceTypesSupportingAmdSevsnp, r.Spec.InstanceType) {
```
bgartzi: Is there a way of also verifying that an AMI with `uefi` or `uefi-preferred` boot modes is being configured for the instance? I couldn't find much following `AMIReference`.

pkg/cloud/services/ec2/instances.go
```diff
@@ -561,6 +563,18 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
...
+		input.CpuOptions = &ec2.CpuOptionsRequest{}
+
+		if i.CpuOptions.AmdSevSnp != nil {
+			val := ec2.AmdSevSnpSpecificationDisabled
+			if *i.CpuOptions.AmdSevSnp {
+				val = ec2.AmdSevSnpSpecificationEnabled
+			}
+			input.CpuOptions.AmdSevSnp = aws.String(val)
+		}
+	}
```
bgartzi: Just a style nit: could we instead just focus on the condition that we are interested in? such as
```suggestion
	if i.CpuOptions != nil && i.CpuOptions.AmdSevSnp != nil && *i.CpuOptions.AmdSevSnp {
		input.CpuOptions = &ec2.CpuOptionsRequest{
			AmdSevSnp: aws.String(ec2.AmdSevSnpSpecificationEnabled),
		}
	}
```

api/v1beta2/awsmachine_webhook.go
fangge1212: One method is to use the describe-images API to retrieve the AMI BootMode:
```
# aws ec2 describe-images --image-ids ami-0fe07c1aadd2e4ac9 \
  --query 'Images[*].BootMode'
[
    "uefi-preferred"
]

```
But AFAIK, it is not possible to do this in webhook

api/v1beta2/awsmachine_types.go
fangge1212: Wait for more feedback before changing it

pkg/cloud/services/ec2/instances.go
fangge1212: 1. One question: If user sets CpuOptions.AmdSevSnp to false in AWSMachineSpec, do we need to reflect this in runInstancesInput? Or just leave it to default?
2. I expect to add other confidential computing technology in a same "if i.CpuOptions != nil" condition, so I have such logic.

fangge1212: If we decide to use a dedicated enum field such as ConfidentialComputing in the API, I will use switch-case logic here.
If we decide to use CpuOptions to mimic the way on the AWS end, I will change the type of CpuOptions.AmdSevSnp to string to make it consistent with the AWS end, so no type convesion is needed here.

api/v1beta2/awsmachine_webhook.go
bgartzi: Okay, thanks!

Yes, we would need an ec2 client, which does not seem to be available within this context.

pkg/cloud/services/ec2/instances.go
bgartzi: > One question: If user sets CpuOptions.AmdSevSnp to false in AWSMachineSpec, do we need to reflect this in runInstancesInput? Or just leave it to default?

We expect `CpuOptions.AmdSevSnp = false` to be the default, so nothing should change if the user sets it explicitly. Right now the logic does not set any SEV-SNP field in the runInstancesInput struct, so I wouldn't be concerned about reflecting that configuration.

> I expect to add other confidential computing technology in a same "if i.CpuOptions != nil" condition, so I have such logic.
> If we decide to use a dedicated enum field such as ConfidentialComputing in the API, I will use switch-case logic here.

That makes sense,

> I will change the type of CpuOptions.AmdSevSnp to string to make it consistent with the AWS end, so no type convesion is needed here.

Okay, let's see what maintainers think about those API enum opetion naming conventions.

api/v1beta2/awsmachine_types.go
nrb: It looks like CAPZ uses a pair of data fields (https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/virtualmachines/spec_test.go#L114) and CAPG uses an enum (https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/b0fd7d672dfb152eedb49be3dabcd7c9c6cb31fe/api/v1beta1/gcpmachine_types.go#L130).

Since these appear to be CPU extensions, is there ever a case where having both SNP and TDX is a valid configuration?

api/v1beta2/awsmachine_types.go
```diff
@@ -73,6 +73,33 @@ const (
...
+type CpuOptions struct {
+	// AmdSevSnp specifies AMD SEV-SNP for the instance.
+	// +kubebuilder:validation:Enum=enabled;disabled
+	// +optional
+	AmdSevSnp AmdSevSnpSpecification `json:"amdSevSnp,omitempty"`
+}
+
+// Confidentail computing support depends on the instance type.
+// Only certain instance types in M6a, R6a and C6a series support AMD SEV-SNP. Reference: https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/sev-snp.html
+var (
```
nrb, line 97: While having this list allows us to avoid an API call to validate whether the instance type is supported or not, it's also going to be a burden to maintain and keep up-to-date.

I wonder if we could describe the instance and check for support via the returned API object, raising an error if it's not present, something like [how CAPZ does it](https://github.com/kubernetes-sigs/cluster-api-provider-azure/blob/main/azure/services/virtualmachines/spec.go#L213), though obviously their data types are different.

fangge1212: CAPZ uses cache to store the sku data, and then get data from the cache. It seems a lot of code changes if we want to do the same. 
Or we don't do the check and rely on the AWS end to fail when creating the instance?

api/v1beta2/awsmachine_types.go
fangge1212: Currently TDX is unsupported for AWS instance. I think both of them can appear in the spec configuration, but only one of them can be enabled.

bgartzi: In case TDX is supported in AWS at some point, having both SNP and TDX in the same instance should not be a valid configuration. SNP is an AMD technology, while TDX is Intel's.

PR #5602, opened 2025-07-30
:sparkles: Create only one nat gateway per AZ

fiunchinho:
**What type of PR is this?**

/kind feature

We noticed that CAPA creates too many NAT gateways. One per AZ would be enough, even if there are multiple subnets in an AZ. This creates unnecessary costs. This applies only to clusters with multiple subnets per AZ, such as clusters with multiple VPC CIDRs (in the same region, of course).

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Create only one nat gateway per AZ
```


AndiDog (commented):
/lgtm

Ankitasw (approved):
/approve

PR #5625, opened 2025-08-21
:bug: Use Go 1.23.0 instead of a higher patch release

nrb:
**What type of PR is this?**

/kind bug

go 1.23.8 is not yet supported in our build environment, and moreover a
Go version should (generally) be kept to major.minor.0 in a go.mod file.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/hack/tools/go.mod
+++ b/hack/tools/go.mod
@@ -1,6 +1,6 @@
 module sigs.k8s.io/cluster-api-provider-aws/hack/tools
 
-go 1.23.8
+go 1.23.0
 
 require (
 	github.com/a8m/envsubst v1.4.2


nrb:
This won't actually work - https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5479 updated OPA for a CVE, and we can't back-rev it.

I think the better thing to do will be to use the Kustomize image for generating our manifests and completely forego using a local copy, since this is a recurrent problem.

Issue #5599, opened 2025-07-23
Support for tenancy=host with 1 instance - 1 host relation


enxebre:
/kind feature

**Describe the solution you'd like**
As a Service Consumer I want to be able to leverage dedicated hosts for BYOL scenarios.
As a Service Consumer I want to express intent for 1 - 1 relation between instance and host with matching capacity so I don’t need to carry the burden of reservations out of band.
As capi/capa API consumer I want capa to expose a knob so I can express intent for this use case. It would discover the most efficient reservation out of my instance type, create the reservation and pin the instance to it at run time

**Anything else you would like to add:**

Similar to https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5548 but covering a different use case. Both complementary.

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

PR #5631, opened 2025-08-26
:sparkles: Allocate dedicated host when a dedicated host doesn't exist

rvanderp3:
**What type of PR is this?**

/kind feature

Builds on #5548 to enable CAPA to allocate a dedicated host. This removes the requirement that a dedicated host exist prior to instance provisioning.  A core assumption of this PR is that there is a 1:1 mapping between an instance and the dedicated host on which it runs. Nothing in this implementation precludes scheduling multiple instances to a given dedicated host, but that is not the goal of this initial implementation.

**Checklist**:

- [x] squashed commits
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:
```
Adds support for provisioning a new instance on a newly allocated dedicated host. The dedicate host is associated only with the new instance.  When the instance is deprovisioned, the dedicated host will be released as well.
```


punkwalker:
Hi @rvanderp3,
Thank you for picking this up. 
I know the PR is still in draft but I wanted to add some suggestion at the start. Hope you don't mind.

api/v1beta2/awsmachine_webhook.go
```diff
@@ -469,14 +469,161 @@ func (r *AWSMachine) validateAdditionalSecurityGroups() field.ErrorList {
...
+	validSizes := map[string]bool{
+		"nano": true, "micro": true, "small": true, "medium": true, "large": true,
+		"xlarge": true, "2xlarge": true, "3xlarge": true, "4xlarge": true, "6xlarge": true,
+		"8xlarge": true, "9xlarge": true, "10xlarge": true, "12xlarge": true, "16xlarge": true,
+		"18xlarge": true, "24xlarge": true, "32xlarge": true, "48xlarge": true, "56xlarge": true,
+		"112xlarge": true, "224xlarge": true, "metal": true,
+	}
+
+	return validSizes[size]
+}
```
punkwalker: Can we use [DescribeInstanceTypes](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#Client.DescribeInstanceTypes) so that we have SDK validation about instance type? That way, we don't need static values in CAPA code.

api/v1beta2/awsmachine_webhook.go
```diff
@@ -469,14 +469,161 @@ func (r *AWSMachine) validateAdditionalSecurityGroups() field.ErrorList {
...
+
+	// Basic validation for region format
+	for _, char := range region {
+		if (char < 'a' || char > 'z') && (char < 'A' || char > 'Z') && (char < '0' || char > '9') && char != '-' {
+			return false
+		}
+	}
+
+	return true
+}
```
punkwalker: We can do similar for this one as well using [DescribeAvailabilityZones](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#Client.DescribeAvailabilityZones)

rvanderp3:
> Hi @rvanderp3, Thank you for picking this up. I know the PR is still in draft but I wanted to add some suggestion at the start. Hope you don't mind.

not at all, I appreciate the early feedback

api/v1beta2/awsmachine_webhook.go
rvanderp3: good point, no need to reproduce that logic.

rvanderp3: it looks like the ec2 client isn't typically used in the context of a webhook.  perhaps we forego the webhook logic and let this fail during reconciliation.  what do you think?

punkwalker: We can fallthrough webhook with basic [kubebuilder validation](https://book.kubebuilder.io/reference/markers/crd-validation) like, pattern and length validation. Then in reconcilation we can have strong validation.

pkg/cloud/services/ec2/instances.go
```diff
@@ -1276,6 +1289,79 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
...
+		if err != nil {
+			return "", errors.Wrap(err, "failed to get subnet for host allocation")
+		}
+		if subnet.AvailabilityZone != nil {
+			availabilityZone = subnet.AvailabilityZone
+		}
+	}
+
+	// Fallback to other AZ sources if needed
+	if availabilityZone == nil {
```
enxebre, line 1339: should this also error if the az inferred from the subnet doesn't match the one expressed in the reservation? Otherwise, if they don't match we fail later silently

pkg/cloud/services/ec2/instances.go
```diff
@@ -1276,6 +1289,79 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
...
+		return "", errors.New("dynamic host allocation spec is nil")
+	}
+
+	// Check if we already have an allocated host
+	if scope.AWSMachine.Status.AllocatedHostID != nil {
+		hostID := *scope.AWSMachine.Status.AllocatedHostID
+		s.scope.Debug("Using existing allocated host", "hostID", hostID)
+
+		// Validate the host is still compatible and available
+		err := s.ValidateHostCompatibility(ctx, hostID, scope.AWSMachine.Spec.InstanceType)
```
enxebre: Just to clarify, while this is good defensive check, this won't ever happen if we address the comments to fail early on creation via cel/webhook.
For day two operation this won't be a possible scenario either since awsMachine instances would be managed by scalable resources i.e. machineDeployment. A change to the awsTemplate would result in recreation of a new machine a deletion of this one releasing the allocation.
Agree?

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,260 @@
...
+	}
+
+	host := output.Hosts[0]
+	hostInfo := s.convertToHostInfo(host)
+
+	return hostInfo, nil
+}
+
+// ListDedicatedHosts lists dedicated hosts based on filters.
+func (s *Service) ListDedicatedHosts(ctx context.Context, filters map[string]string) ([]*infrav1.DedicatedHostInfo, error) {
```
enxebre: where is this used?

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,260 @@
...
+		input.TagSpecifications = tagSpecs
+	}
+
+	output, err := s.EC2Client.AllocateHosts(ctx, input)
+	if err != nil {
+		record.Warnf(s.scope.InfraCluster(), "FailedAllocateDedicatedHost", "Failed to allocate dedicated host: %v", err)
+		return "", errors.Wrap(err, "failed to allocate dedicated host")
+	}
+
+	if len(output.HostIds) == 0 {
```
enxebre: does aws give guarantees to always return an id immediately in the output response to the creation?
Otherwise that scenario would need to be handled with an async mechanism to fetch on going allocations

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,260 @@
...
+		return "", errors.New("no dedicated host ID returned from allocation")
+	}
+
+	hostID := output.HostIds[0]
```
enxebre, line 93: why [0] what happens if quantity > 1

api/v1beta2/awsmachine_types.go
```diff
@@ -432,6 +471,11 @@ type AWSMachineStatus struct {
 	// Conditions defines current service state of the AWSMachine.
 	// +optional
 	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
+
+	// AllocatedHostID tracks the dynamically allocated dedicated host ID.
+	// This field is populated when DynamicHostAllocation is used.
+	// +optional
+	AllocatedHostID *string `json:"allocatedHostID,omitempty"`
```
enxebre: If we allow N Quantity, how does that with having a single id in the status?

api/v1beta2/awsmachine_types.go
```diff
@@ -255,6 +260,40 @@ type AWSMachineSpec struct {
...
+	// +kubebuilder:validation:Maximum=10
+	// +kubebuilder:default=1
+	// +optional
+	Quantity *int32 `json:"quantity,omitempty"`
+
+	// AutoRelease determines whether to automatically release the dedicated host
+	// when the machine is deleted.
+	// +kubebuilder:default=true
+	// +optional
+	AutoRelease *bool `json:"autoRelease,omitempty"`
```
enxebre: can we make this a structured type with an enum instead of a bool? e.g. release: auto/none

api/v1beta2/awsmachine_types.go
```diff
@@ -255,6 +260,40 @@ type AWSMachineSpec struct {
...
+	AvailabilityZone *string `json:"availabilityZone,omitempty"`
+
+	// InstanceType specifies the specific instance type for the dedicated host.
+	// If not specified, derives from InstanceFamily.
+	// +optional
+	InstanceType *string `json:"instanceType,omitempty"`
+
+	// Quantity specifies the number of dedicated hosts to allocate.
+	// +kubebuilder:validation:Minimum=1
+	// +kubebuilder:validation:Maximum=10
```
enxebre: I'm curious where is this maximum coming from?

api/v1beta2/awsmachine_types.go
```diff
@@ -255,6 +260,40 @@ type AWSMachineSpec struct {
 	CapacityReservationPreference CapacityReservationPreference `json:"capacityReservationPreference,omitempty"`
 }
 
+// DynamicHostAllocationSpec defines the configuration for dynamic dedicated host allocation.
+type DynamicHostAllocationSpec struct {
+	// InstanceFamily specifies the EC2 instance family (e.g., "m5", "c5", "r5").
+	// +kubebuilder:validation:Required
+	InstanceFamily string `json:"instanceFamily"`
```
enxebre: can we have regex cel/webhook validation to prevent invalid input for InstanceFamily / InstanceType.

Can we also compute the upper level spec.InstanceType in that validation, so any invalid input fails earlier

api/v1beta2/awsmachine_types.go
```diff
@@ -246,6 +246,11 @@ type AWSMachineSpec struct {
 	// +kubebuilder:validation:Enum:=default;host
 	HostAffinity *string `json:"hostAffinity,omitempty"`
 
+	// DynamicHostAllocation enables automatic allocation of dedicated hosts.
```
enxebre: can we please enforce this mutually exclusivity with CEL or a webhook. I assume current maintainers of this repo will prefer the webhook.

enxebre:
covers https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5599

enxebre:
@rvanderp3 thanks! this is looking great. I dropped some early feedback.

api/v1beta2/awsmachine_types.go
rvanderp3: yep, good catch, was a bit torn on how to handle multiple dedicated hosts.  will change this to a slice and we can limit it to a max of 1 for now or until we have a chance to discuss.

pkg/cloud/services/ec2/instances.go
rvanderp3: yep, that makes sense

pkg/cloud/services/ec2/dedicatedhosts.go
rvanderp3: we are going to have a 1:1 mapping of machine to instance for now. quantity is being removed for now

pkg/cloud/services/ec2/instances.go
rvanderp3: the az will now be inferred directly from the machine subnet. there shouldnt be an opportunity for a mismatch now

api/v1beta2/awsmachine_types.go
rvanderp3: for sure, will add that

pkg/cloud/services/ec2/dedicatedhosts.go
rvanderp3: we'll be expecting a single host ID, which should be in the output if the call to allocate succeeds. if it doesn't, it will be retried in a later reconciliation attempt.

api/v1beta2/awsmachine_types.go
rvanderp3: this was removed altogether

api/v1beta2/awsmachine_types.go
rvanderp3: the host family is now inferred from the machine spec

pkg/cloud/services/ec2/dedicatedhosts.go
rvanderp3: it was a vestige from initial work on this, it is not used and has been removed.

api/v1beta2/awsmachine_types.go
```diff
@@ -41,6 +41,19 @@ const (
...
+// It determines when the dedicated host associated with an AWSMachine should be released, such as
+// automatically upon machine deletion or only when manually triggered.
+type DynamicHostReleaseStrategy string
+
+const (
+	// DedicatedHostReleaseStrategyOnMachineDeletion means the dedicated host will be released when the machine is deleted.
+	DedicatedHostReleaseStrategyOnMachineDeletion = DynamicHostReleaseStrategy("on-machine-deletion")
+
+	// DedicatedHostReleaseStrategyManual means the dedicated host will be released manually by the user.
+	DedicatedHostReleaseStrategyManual = DynamicHostReleaseStrategy("manual")
```
mtulio: Let me know if those doc string suggestions makes sense.
```suggestion
	// DedicatedHostReleaseStrategyOnMachineDeletion specifies the strategy to release the Dedicated Host instance when the machine is deleted.
	DedicatedHostReleaseStrategyOnMachineDeletion = DynamicHostReleaseStrategy("on-machine-deletion")

	// DedicatedHostReleaseStrategyManual specifies the strategy to release the Dedicated Host instance manually by the user.
	DedicatedHostReleaseStrategyManual = DynamicHostReleaseStrategy("manual")
```

rvanderp3:
/hold

investigating test failures

rvanderp3:
after conferring with my team, it sounds like the apidiff error may be expected due to changes to the API.  dropping the hold and ready for review.

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/record"
+)
+
+// AllocateDedicatedHost allocates a single dedicated host based on the specification.
+// This function always allocates exactly one dedicated host per call.
+func (s *Service) AllocateDedicatedHost(ctx context.Context, spec *infrav1.DynamicHostAllocationSpec, instanceType, availabilityZone string) (string, error) {
+	s.scope.Debug("Allocating single dedicated host", "instanceType", instanceType, "availabilityZone", availabilityZone)
+	instanceFamily := extractInstanceFamily(instanceType)
+	input := &ec2.AllocateHostsInput{
+		InstanceFamily:   aws.String(instanceFamily),
```
enxebre: AllocateHostsInput supports instance type as input. Any reason to not pass `InstanceType: instanceType` directly here? cc @punkwalker

api/v1beta2/awsmachine_types.go
```diff
@@ -235,17 +248,25 @@ type AWSMachineSpec struct {
...
+	// When HostAffinity is set to host, an instance started onto a specific host always restarts on the same host if stopped.
+	// When HostAffinity is set to default, and you stop and restart the instance, it can be restarted on any available host.
 	// When HostAffinity is defined, HostID is required.
 	// +optional
 	// +kubebuilder:validation:Enum:=default;host
+	// +kubebuilder:default=default
 	HostAffinity *string `json:"hostAffinity,omitempty"`
 
+	// DynamicHostAllocation enables automatic allocation of a single dedicated host.
+	// This field is mutually exclusive with HostID and always allocates exactly one host.
```
enxebre, line 263: nit: maybe include some more context for consumers here e.g:
"depending on the instance type used, the host allocation for that single instance might be less or more cost effective. This is usually a good fit for metal instance types..."
@punkwalker is it fair to say that the .metal size instance types would always use all the host allocation sockets/cpus for that same type? as in the example here https://aws.amazon.com/blogs/compute/understanding-the-lifecycle-of-amazon-ec2-dedicated-hosts

it would be also nice to document better the .Tenancy field, e.g.
"when .tenancy=host is set, you can either let aws find a suitable host for you from preexisting allocations that have auto-placement enabled, choose an specific .hostID, or use dynamicHostAllocation to create a dedicated host allocation for the instance.

enxebre:
thanks @rvanderp3, lgtm overall other than https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5631#discussion_r2329669957

pkg/cloud/services/ec2/dedicatedhosts.go
rvanderp3: Good point, this may be a misunderstanding on my part.  I thought the dedicated host was associated with a family and instances are deployed on it.  Perhaps if you pass an instance type, it derives the family from that.  I'll need to investigate.

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+
+	input := &ec2.ReleaseHostsInput{
+		HostIds: []string{hostID},
+	}
+
+	_, err := s.EC2Client.ReleaseHosts(ctx, input)
+	if err != nil {
+		record.Warnf(s.scope.InfraCluster(), "FailedReleaseDedicatedHost", "Failed to release dedicated host %s: %v", hostID, err)
+		return errors.Wrap(err, "failed to release dedicated host")
+	}
```
mtulio: I see the [ReleaseHosts API](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#Client.ReleaseHosts) supports the [retryer](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/aws#Retryer), do you think it can help to enhance API calls by recovering from expected failures with a backoff? I am not familiar with releasing DH to understand how long it could take, but do you think if adjusting with observations it could eventually save time until next service reconciliation (and perhaps cloud costs with DH allocation)?

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+		return nil, errors.Errorf("dedicated host %s not found", hostID)
+	}
+
+	host := output.Hosts[0]
+	hostInfo := s.convertToHostInfo(host)
+
+	return hostInfo, nil
+}
+
+// convertToHostInfo converts an AWS Host to our DedicatedHostInfo struct.
```
mtulio:
```suggestion
// convertToHostInfo converts an AWS Host to the DedicatedHostInfo struct.
```

pkg/cloud/services/ec2/instances.go
```diff
@@ -1276,6 +1289,120 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
...
+// allocateHostIfNeeded allocates a dedicated host if dynamic allocation is configured.
+func (s *Service) allocateHostIfNeeded(ctx context.Context, scope *scope.MachineScope) (string, error) {
+	spec := scope.AWSMachine.Spec.DynamicHostAllocation
+	if spec == nil {
+		return "", errors.New("dynamic host allocation spec is nil")
+	}
+
+	// Check if a host is already allocated for this machine
+	if scope.AWSMachine.Status.AllocatedHostID != nil {
+		existingHostID := *scope.AWSMachine.Status.AllocatedHostID
```
mtulio: nit
```suggestion
		existingHostID := aws.StringValue(scope.AWSMachine.Status.AllocatedHostID)
```

pkg/cloud/services/ec2/instances.go
```diff
@@ -1276,6 +1289,120 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
 	return request
 }
 
+// allocateHostIfNeeded allocates a dedicated host if dynamic allocation is configured.
+func (s *Service) allocateHostIfNeeded(ctx context.Context, scope *scope.MachineScope) (string, error) {
```
mtulio: What about to call something more specifically to it's purpose, such as `allocateHostWithDynamicAllocation`?

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+	for _, tag := range host.Tags {
+		if tag.Key != nil && tag.Value != nil {
+			hostInfo.Tags[*tag.Key] = *tag.Value
+		}
+	}
+
+	return hostInfo
+}
+
+// ValidateHostCompatibility validates that an instance type is compatible with a host.
```
mtulio: nit improving doc strings
```suggestion
// ValidateHostCompatibility validates that an instance type is compatible with a host.
// A host is incompatible when the instance family does not match with the Dedicated Host.
```

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+	instanceFamily := extractInstanceFamily(instanceType)
+
+	if hostInfo.InstanceFamily != instanceFamily {
+		return errors.Errorf("instance type %s (family %s) is not compatible with host %s (family %s)",
+			instanceType, instanceFamily, hostID, hostInfo.InstanceFamily)
+	}
+
+	// Check if host has available capacity
+	if hostInfo.AvailableCapacity <= 0 {
+		return errors.Errorf("host %s has no available capacity", hostID)
```
mtulio: What happens if the host already have **other** instances allocated (other than reconciled)? Will it purge others EC2 in the caller when `ReleaseDedicatedHost`?

pkg/cloud/services/ec2/instances.go
```diff
@@ -1276,6 +1289,120 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
...
+		// Validate that the existing host is still available and compatible
+		instanceType := scope.AWSMachine.Spec.InstanceType
+		if err := s.ValidateHostCompatibility(ctx, existingHostID, instanceType); err != nil {
+			s.scope.Warn("Existing allocated host is no longer compatible, will release and allocate new host",
+				"hostID", existingHostID, "instanceType", instanceType, "error", err)
+
+			// Release the incompatible host before allocating a new one
+			if releaseErr := s.ReleaseDedicatedHost(ctx, existingHostID); releaseErr != nil {
+				s.scope.Error(releaseErr, "Failed to release incompatible dedicated host", "hostID", existingHostID)
+				// Continue with new allocation even if release fails
```
mtulio: is there any risk to leak the DH in this case?

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+
+// ReleaseDedicatedHost releases a dedicated host.
+func (s *Service) ReleaseDedicatedHost(ctx context.Context, hostID string) error {
+	s.scope.Debug("Releasing dedicated host", "hostID", hostID)
+
+	input := &ec2.ReleaseHostsInput{
+		HostIds: []string{hostID},
+	}
+
+	_, err := s.EC2Client.ReleaseHosts(ctx, input)
```
mtulio: the [ReleaseHostsOutput](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#ReleaseHostsOutput) could be captured to provide more information in case the DH isn't released

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+	if err != nil {
+		record.Warnf(s.scope.InfraCluster(), "FailedReleaseDedicatedHost", "Failed to release dedicated host %s: %v", hostID, err)
+		return errors.Wrap(err, "failed to release dedicated host")
+	}
+
+	s.scope.Info("Successfully released dedicated host", "hostID", hostID)
```
mtulio: it would be nice to have here the IDs from output of release DH, instead of expected input already shown earlier in this function so users could have more information what steps controller takes while investigating issues.

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+func (s *Service) ReleaseDedicatedHost(ctx context.Context, hostID string) error {
+	s.scope.Debug("Releasing dedicated host", "hostID", hostID)
+
+	input := &ec2.ReleaseHostsInput{
+		HostIds: []string{hostID},
+	}
+
+	_, err := s.EC2Client.ReleaseHosts(ctx, input)
+	if err != nil {
+		record.Warnf(s.scope.InfraCluster(), "FailedReleaseDedicatedHost", "Failed to release dedicated host %s: %v", hostID, err)
```
mtulio, line 124: it would be nice to expand the failure reason extracting the messages from ReleaseHostsOutput.Unsuccessful[]

pkg/cloud/services/ec2/instances.go
```diff
@@ -1276,6 +1289,120 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
...
+		if subnet != nil && subnet.AvailabilityZone != nil {
+			availabilityZone = subnet.AvailabilityZone
+		}
+	}
+
+	instanceType := scope.AWSMachine.Spec.InstanceType
+	instanceFamily := extractInstanceFamily(instanceType)
+
+	if availabilityZone == nil {
+		return "", errors.New("availability zone could not be determined, please specify a subnet or subnet filters")
```
mtulio: Is it the Subnet ID to be specified as alternatives of filters?

```suggestion
		return "", errors.New("availability zone could not be determined, please specify a subnet ID or subnet filters")
```

pkg/cloud/services/ec2/dedicatedhosts.go
```diff
@@ -0,0 +1,199 @@
...
+	return nil
+}
+
+// extractInstanceFamily extracts the instance family from an instance type.
+func extractInstanceFamily(instanceType string) string {
+	// Instance types follow the pattern: family + generation + size (e.g., "m5.large")
+	parts := strings.Split(instanceType, ".")
+	if len(parts) < 2 {
+		return instanceType
+	}
```
mtulio: I wonder if returning `instanceType` would return different value of the signature (family). I see the calls to this function and, IIUC, may provide unexpected behavior, such as on [`AllocateHostsInput`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#AllocateHostsInput) which allows `InstanceFamily` or `InstanceType`.

pkg/cloud/services/ec2/dedicatedhosts.go
mtulio: > InstanceType: instanceType

if we are going to support one machine per DH on DHA, maybe we need to set only [the `AllocateHostsInput.InstanceType`](https://pkg.go.dev/github.com/aws/aws-sdk-go-v2/service/ec2#AllocateHostsInput) ?

rvanderp3:
/hold

holding while working through feedback

pkg/cloud/services/ec2/dedicatedhosts.go
punkwalker: for 1:1 mapping, we should set the InstanceType and omit InstanceFamily as both can not be used at the same time.

pkg/cloud/services/ec2/dedicatedhosts.go
rvanderp3: Yep, it resolves the instance of the dedicated host for you if you pass the instance type

pkg/cloud/services/ec2/dedicatedhosts.go
rvanderp3: by the time we release the dedicated host, all instances on the host should be gone.  Generally, it's a quick process.  regardless, I'm adding the retryer.

pkg/cloud/services/ec2/dedicatedhosts.go
rvanderp3: no, it will not purge others.  it will only purge instances assoiciated with the CAPA machine being deleted.

api/v1beta2/awsmachine_types.go
punkwalker: > is it fair to say that the .metal size instance types would always use all the host allocation sockets/cpus for that same type? 

It seems that this is mostly true for all instance families but there can be some exceptions as well. However, as we are going with 1:1 mapping, I don't think so we need to consider this as of now.

> "when .tenancy=host is set, you can either let aws find a suitable host for you from preexisting allocations that have auto-placement enabled, choose an specific .hostID, or use dynamicHostAllocation to create a dedicated host allocation for the instance.

Similarly for this, if we do 1:1 mapping then we should not use auto-placement as it will be redundant when we place only single instance on the host.

api/v1beta2/awsmachine_types.go
```diff
@@ -255,6 +276,21 @@ type AWSMachineSpec struct {
...
+	// when the machine is deleted.
+	// +kubebuilder:default=on-machine-deletion
+	// +kubebuilder:validation:Enum:=on-machine-deletion;manual
+	// +optional
+	Release DynamicHostReleaseStrategy `json:"release,omitempty"`
+
+	// Tags to apply to the allocated dedicated host.
+	// +optional
+	Tags map[string]string `json:"tags,omitempty"`
+}
```
punkwalker, line 284: Can we use the same tags that are present on Machine under `spec.additionalTags` and default tags as well?

api/v1beta2/awsmachine_types.go
```diff
@@ -255,6 +276,21 @@ type AWSMachineSpec struct {
...
 
+// DynamicHostAllocationSpec defines the configuration for dynamic dedicated host allocation.
+// This specification always allocates exactly one dedicated host per machine.
+type DynamicHostAllocationSpec struct {
+	// AutoRelease determines whether to automatically release the dedicated host
+	// when the machine is deleted.
+	// +kubebuilder:default=on-machine-deletion
+	// +kubebuilder:validation:Enum:=on-machine-deletion;manual
+	// +optional
+	Release DynamicHostReleaseStrategy `json:"release,omitempty"`
```
punkwalker: Do we really need two strategies? How would manual release work if the host is dynamically alllocated?

pkg/cloud/services/ec2/instances.go
rvanderp3: removed it from the logging just in case

api/v1beta2/awsmachine_types.go
```diff
@@ -432,6 +468,23 @@ type AWSMachineStatus struct {
 	// Conditions defines current service state of the AWSMachine.
 	// +optional
 	Conditions clusterv1.Conditions `json:"conditions,omitempty"`
+
+	// AllocatedHostID tracks the dynamically allocated dedicated host ID.
+	// This field is populated when DynamicHostAllocation is used.
+	// +optional
+	AllocatedHostID *string `json:"allocatedHostID,omitempty"`
```
punkwalker: IMO, this status field can be `DedicatedHostID`. We set it to hostID of either user specified hostID or dynamically allocated hostID. The lifecycle ownership of the host can be determined by the tag. If the tag is present for dynamic allocation then ownership is on CAPA otherwise user owns it.

pkg/cloud/services/ec2/dedicatedhosts.go
punkwalker: Based on the `ReleaseHostsOutput`, looks like the call may succeed and the host release will still be unsuccessful. So it's better to rely on output for release failures and API errors can be inferred from err returned by release call.

pkg/cloud/services/ec2/dedicatedhosts.go
punkwalker: Let's use either CAPA retryer or AWS SDK retryer whenever we need retries instead of implementing our own like this one `shouldRetryHostRelease`

api/v1beta2/awsmachine_types.go
rvanderp3: fair point, we can always add that later if we need to

api/v1beta2/awsmachine_types.go
rvanderp3: we could for sure.  will make that change

rvanderp3:
/unhold

aws-test appears to be failing due to VPC quota exhaustion

rvanderp3:
/hold 

i need to look a bit closer, it does appear one of the new unit tests is failing

api/v1beta2/awsmachine_types.go
rvanderp3: @punkwalker Ive made the change to include tags specified in spec.additionalTags.  I looked at including the default tags as well but it seemed odd to apply some of them to the dedicated host.  For example, the role tag that would be attached to the machine is node and that was be propagated to the dedicate host.  I assumed default tags meant those that would be applied to the machine but I may have misunderstood.  What did you mean by "default tags"?

controllers/awsmachine_controller.go
```diff
@@ -360,6 +361,84 @@ func (r *AWSMachineReconciler) reconcileDelete(ctx context.Context, machineScope
...
 		machineScope.Info("EC2 instance terminated successfully", "instance-id", instance.ID)
+
+		// Handle dedicated host cleanup AFTER instance is confirmed terminated
+		if machineScope.AWSMachine.Status.DedicatedHost != nil &&
+			machineScope.AWSMachine.Status.DedicatedHost.ID != nil &&
+			machineScope.AWSMachine.Spec.DynamicHostAllocation != nil {
+			hostID := *machineScope.AWSMachine.Status.DedicatedHost.ID
+
+			// Check if we should retry host release
+			shouldRetry, retryAfter := shouldRetryHostRelease(machineScope)
```
mtulio: Is that retry control fields still required even now it is using the native retry from SDK on `ReleaseDedicatedHost`?

pkg/cloud/services/ec2/instances.go
```diff
@@ -1279,6 +1292,95 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
...
+	if scope.AWSMachine.Status.DedicatedHost.ID != nil {
+		existingHostID := aws.ToString(scope.AWSMachine.Status.DedicatedHost.ID)
+		s.scope.Info("Found existing allocated host for machine", "hostID", existingHostID, "machine", scope.Name())
+		return existingHostID, nil
+	}
+
+	// Determine the availability zone for the host
+	var availabilityZone *string
+
+	// Get AZ from the machine's subnet
```
mtulio, line 1316: is it possible to use the service helper `findSubnet` in this context?

rvanderp3:
/hold

placing hold to work through feedback from @mtulio . thanks Marco!

controllers/awsmachine_controller.go
rvanderp3: no, it should not be required.  removed.

controllers/awsmachine_controller.go
```diff
@@ -360,6 +360,35 @@ func (r *AWSMachineReconciler) reconcileDelete(ctx context.Context, machineScope
...
+			machineScope.AWSMachine.Status.DedicatedHost.ID != nil &&
+			machineScope.AWSMachine.Spec.DynamicHostAllocation != nil {
+			hostID := *machineScope.AWSMachine.Status.DedicatedHost.ID
+
+			// Attempt to release the dedicated host
+			machineScope.Info("Releasing dynamically allocated dedicated host", "hostID", hostID)
+			if err := ec2Service.ReleaseDedicatedHost(ctx, hostID); err != nil {
+				// Host release failed, set up retry logic
+				machineScope.Error(err, "failed to release dedicated host", "hostID", hostID)
+				r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeWarning, "FailedReleaseHost", "Failed to release dedicated host %s: %v", hostID, err)
```
mtulio, line 375: Do we need to set the failure conditional here (maybe the `DedicatedHostReleaseCondition`?), as well returning when failed to release?

pkg/cloud/services/ec2/instances.go
```diff
@@ -1279,6 +1295,80 @@ func getInstanceMetadataOptionsRequest(metadataOptions *infrav1.InstanceMetadata
...
+	if scope.AWSMachine.Spec.Subnet != nil {
+		subnetID, err := s.findSubnet(scope)
+		if err != nil {
+			return "", errors.Wrap(err, "failed to find subnet for host allocation")
+		}
+
+		// Get the full subnet object to extract availability zone
+		subnet, err := s.getSubnet(&subnetID)
```
mtulio: nit: I wonder if we can prevent new functions (`getSubnet`) with similar SDK calls, instead use existing helper functions provided by controller. In this case `getFilteredSubnets` looks performing the same goal with `EC2Client.DescribeSubnets`.

controllers/awsmachine_controller.go
mtulio: Is the const `DedicatedHostReleaseFailedReason` created to use here?

controllers/awsmachine_controller.go
```diff
@@ -360,6 +360,35 @@ func (r *AWSMachineReconciler) reconcileDelete(ctx context.Context, machineScope
...
+			} else {
+				// Host release succeeded
+				machineScope.Info("Successfully released dedicated host", "hostID", hostID)
+				r.Recorder.Eventf(machineScope.AWSMachine, corev1.EventTypeNormal, "SuccessfulReleaseHost", "Released dedicated host %s", hostID)
```
mtulio: Same doubt about the const `DedicatedHostReleaseSucceededReason`

api/v1beta2/conditions_consts.go
```diff
@@ -191,4 +195,13 @@ const (
...
 	S3BucketFailedReason = "S3BucketCreationFailed"
+
+	// DedicatedHostReleaseSucceededReason used when the dedicated host is successfully released.
+	DedicatedHostReleaseSucceededReason = "DedicatedHostReleaseSucceeded"
+
+	// DedicatedHostReleaseFailedReason used when the dedicated host release fails.
+	DedicatedHostReleaseFailedReason = "DedicatedHostReleaseFailed"
+
+	// DedicatedHostReleaseRetryingReason used when the dedicated host release is being retried.
+	DedicatedHostReleaseRetryingReason = "DedicatedHostReleaseRetrying"
```
mtulio: I did some comments of where those reason consts are used.

mtulio (commented):
Hey @rvanderp3 , thanks for addressing the feedback. Couple of nits, and one question of failed condition.

controllers/awsmachine_controller.go
rvanderp3: yes, it is, im thinking we'll keep this condition

controllers/awsmachine_controller.go
rvanderp3: yep, we dont need this anymore

api/v1beta2/conditions_consts.go
rvanderp3: same, we dont need this anymore

pkg/cloud/services/ec2/instances.go
rvanderp3: yep, agreed.  fixed.

mtulio (commented):
Thanks for addressing the comments, @rvanderp3 ! This looks good to me! 👍🏽 

/lgtm

mtulio:
> Thanks for addressing the comments, @rvanderp3 ! This looks good to me! 👍🏽
> 
> /lgtm

@rvanderp3 nit: you may need to squash the commits :)

rvanderp3:
> > Thanks for addressing the comments, @rvanderp3 ! This looks good to me! 👍🏽
> > /lgtm
> 
> @rvanderp3 nit: you may need to squash the commits :)

good point.  done!

api/v1beta2/awsmachine_types.go
```diff
@@ -438,6 +461,21 @@ type AWSMachineStatus struct {
...
+
+	// DedicatedHost tracks the dynamically allocated dedicated host.
+	// This field is populated when DynamicHostAllocation is used.
+	// +optional
+	DedicatedHost *DedicatedHostStatus `json:"dedicatedHost,omitempty"`
+}
+
+// DedicatedHostStatus defines the observed state of a dynamically allocated dedicated host
+// associated with an AWSMachine. This struct is used to track the ID of the dedicated host
+// and any failure messages encountered during host release operations.
```
nrb: This part should be omitted if we're not including a field for the failure messages right now.

pkg/cloud/services/ec2/instances.go
```diff
@@ -258,9 +258,25 @@ func (s *Service) CreateInstance(ctx context.Context, scope *scope.MachineScope,
 
 	input.MarketType = scope.AWSMachine.Spec.MarketType
 
-	input.HostID = scope.AWSMachine.Spec.HostID
+	// Handle dynamic host allocation if specified
+	if scope.AWSMachine.Spec.DynamicHostAllocation != nil {
+		hostID, err := s.ensureDedicatedHostAllocation(ctx, scope)
+		if err != nil {
```
nrb, line 264: I would expect any error messages that `DedicatedHostStatus` would hold would be collected around here.

api/v1beta2/awsmachine_types.go
rvanderp3: thanks, i meant to remove that, actually.  that is a vestige from earlier in the effort.

nrb:
The apidiff failure is fine, just letting us know that a Go-level API was changed. Since it's additive here, I see no issue.

pkg/cloud/services/ec2/instances.go
rvanderp3: i believe this was addressed with the dropping of the comment, right?

nrb: Yep! Just put it here in case you were going to add fields back in.

nrb:
/lgtm
/approve

Doing both since the last edit was just updating a comment.

rvanderp3:
/release-note-edit
```release-note
Adds support for provisioning a new instance on a newly allocated dedicated host. The dedicate host is associated only with the new instance. When the instance is deprovisioned, the dedicated host will be released as well.
```

PR #5578, opened 2025-07-02
✨ feat: Support setting EKS authentication mode

joshfrench:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Support setting EKS [AuthenticationMode](https://docs.aws.amazon.com/eks/latest/userguide/setting-up-access-entries.html).

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Partially addresses #4854

**Special notes for your reviewer**:
This PR just finishes the work started in #5108 by rebasing, removing the API changes to v1beta1, and fixing the fuzzy tests. There were two seemingly unrelated changes in the original PR that I've removed:
- Add CGFLAGS to build command (I'm dropping this as unrelated)
- One documentation fix that I'll resubmit in its own PR

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Added support for setting EKS AuthenticationMode (required for migrating to EKS Access Entries)
```

Issue #5278, opened 2025-01-13
Add EKS AutoMode support


richardcase:
/kind feature
/priority important-soon
/triage accepted
/help

**Describe the solution you'd like**

Add support for [EKS Auto mode](https://docs.aws.amazon.com/eks/latest/userguide/automode.html) to CAPA.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

seanschneeweiss:
This will require the upgrade to AWS SDK v2 in #2225. We created a fork to include EKS Auto Mode support in the v1 SDK: https://github.com/c445/aws-sdk-go/commit/309a05d2c76eb43157f57cd8b4346cd94c6b99b5
The fork of SDK v1 was just a quick starter for us. I guess for CAPA upstream the SDK v2 and the CRD change is the way forward.

IMHO it fits into `AWSManagedControlPlane`. On the long run, a mixed mode of EKS Auto Mode and non EKS Auto Mode might be supported by EKS.
> Node Pools in EKS Auto Mode differ from Amazon EKS Managed Node Groups but can coexist in the same cluster. [src](https://docs.aws.amazon.com/eks/latest/userguide/automode-get-started-cli.html)

I wished this could be aligned with ACK controller for EKS, but there is no progress yet: https://github.com/aws-controllers-k8s/community/issues/2222

The AWS SDK exposes Auto Mode settings as follows:

**`ComputeConfig *ComputeConfigRequest`**
```go
type ComputeConfigRequest struct {
	// Request to enable or disable the compute capability on your EKS Auto Mode
	// cluster. If the compute capability is enabled, EKS Auto Mode will create
	// and delete EC2 Managed Instances in your Amazon Web Services account.
	Enabled *bool `locationName:"enabled" type:"boolean"`
	// Configuration for node pools that defines the compute resources for your
	// EKS Auto Mode cluster. For more information, see EKS Auto Mode Node Pools
	// in the EKS User Guide.
	NodePools []*string `locationName:"nodePools" type:"list"`
	// The ARN of the IAM Role EKS will assign to EC2 Managed Instances in your
	// EKS Auto Mode cluster. This value cannot be changed after the compute capability
	// of EKS Auto Mode is enabled. For more information, see the IAM Reference
[...]

PR #5642, opened 2025-09-02
✨ Add ability to control "EKS Auto Mode" for EKS clusters

CharlieR-o-o-t:
What type of PR is this?
/kind feature
/kind api-change

**What this PR does / why we need it**
This PR introduces support for controlling EKS Auto Mode for EKS clusters managed with AWSManagedControlPlane CRD. 
Previously, the EKSAutoMode was always enabled by default, for someone it's not needed. 
With this change, users can explicitly set the spec.eksAutoMode field in the AWSManagedControlPlane resource to true or false based on their needs.

Includes extra fix:
AWSManagedControlPlane.spec.bootstrapSelfManagedAddons was always defaulted to true by webhook.

**Checklist**:
- [X] squashed commits
- [X] includes documentation
- [X] includes emoji in title 
- [X] adds unit tests

**Release note**:
```release-note
Ability to control "EKS Auto Mode" with AWSManagedControlPlane CRD. New field "EKSAutoMode" has been added to spec.
```

### DIFF
--- a/.github/workflows/auto-publish-ami.yml
+++ b/.github/workflows/auto-publish-ami.yml
@@ -32,10 +32,10 @@ jobs:
       failures: ${{ steps.matrix.outputs.failures }}
     steps:
       - name: Checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Set up Go
-        uses: actions/setup-go@d35c59abb061a4a6fb18e82ac0862c26744d6ab5 # v5.5.0
+        uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
         with:
           go-version-file: go.mod
 
--- a/.github/workflows/build-ami-v2.yml
+++ b/.github/workflows/build-ami-v2.yml
@@ -56,7 +56,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           repository: ${{ inputs.image_builder_repo }}
           ref: ${{ inputs.image_builder_version }}
--- a/.github/workflows/build-ami-varsfile.yml
+++ b/.github/workflows/build-ami-varsfile.yml
@@ -33,7 +33,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           repository: ${{ inputs.image_builder_repo }}
           ref: ${{ inputs.image_builder_version }}
--- a/.github/workflows/build-ami.yml
+++ b/.github/workflows/build-ami.yml
@@ -52,7 +52,7 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - name: checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
         with:
           repository: ${{ inputs.image_builder_repo }}
           ref: ${{ inputs.image_builder_version }}
--- a/.github/workflows/codeql-analysis.yml
+++ b/.github/workflows/codeql-analysis.yml
@@ -21,18 +21,18 @@ jobs:
         language: [ 'go' ]
     steps:
     - name: Checkout repository
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
-      uses: github/codeql-action/init@7211b7c8077ea37d8641b6271f6a365a22a5fbfa # v4.36.0
+      uses: github/codeql-action/init@8aad20d150bbac5944a9f9d289da16a4b0d87c1e # v4.36.2
       with:
         languages: ${{ matrix.language }}
 
     # Autobuild attempts to build any compiled languages  (C/C++, C#, or Java).
     # If this step fails, then you should remove it and run the build manually (see below)
     - name: Autobuild
-      uses: github/codeql-action/autobuild@7211b7c8077ea37d8641b6271f6a365a22a5fbfa # v4.36.0
+      uses: github/codeql-action/autobuild@8aad20d150bbac5944a9f9d289da16a4b0d87c1e # v4.36.2
 
     - name: Perform CodeQL Analysis
-      uses: github/codeql-action/analyze@7211b7c8077ea37d8641b6271f6a365a22a5fbfa # v4.36.0
+      uses: github/codeql-action/analyze@8aad20d150bbac5944a9f9d289da16a4b0d87c1e # v4.36.2
--- a/.github/workflows/dependabot.yml
+++ b/.github/workflows/dependabot.yml
@@ -28,7 +28,7 @@ jobs:
         go-version: '1.24'
       id: go
     - name: Check out code into the Go module directory
-      uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
     - uses: actions/cache@27d5ce7f107fe9357f9df03efb73ab90386fccae # v5.0.5
       name: Restore go cache
       with:
--- a/.github/workflows/detect-k8s-releases.yml
+++ b/.github/workflows/detect-k8s-releases.yml
@@ -21,7 +21,7 @@ jobs:
       id-token: write # required for OIDC authentication to AWS
     steps:
       - name: Checkout code
-        uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
 
       - name: Set up Go
         uses: actions/setup-go@4a3601121dd01d1626a1e23e37211e3254c1c06c # v6.4.0
--- a/.github/workflows/md-link-checker.yml
+++ b/.github/workflows/md-link-checker.yml
@@ -8,7 +8,7 @@ jobs:
   check:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # v6.0.3
       - uses: artyom/mdlinks@3d49483ce078798ed3aee5abc1c65476040654dd # v0.4.2
         with:
           dir: 'docs/book'
--- a/.github/workflows/pr-golangci-lint.yaml
+++ b/.github/workflows/pr-golangci-lint.yaml
@@ -17,7 +17,7 @@ jobs:
         working-directory:
           - ""
     steps:
-      - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v4.2.2
+      - uses: actions/checkout@df4cb1c069e1874edd31b4311f1884172cec0e10 # tag=v4.2.2
       - name: Calculate go version
         id: vars
         run: echo "go_version=$(make go-version)" >> $GITHUB_OUTPUT
--- a/.github/workflows/pr-verify.yml
+++ b/.github/workflows/pr-verify.yml
@@ -9,7 +9,7 @@ jobs:
     runs-on: ubuntu-latest
     name: verify PR contents
     steps:
-    - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # tag=v4.2.2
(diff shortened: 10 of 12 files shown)


controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -200,7 +200,15 @@ type AWSManagedControlPlaneSpec struct { //nolint: maligned
...
-	BootstrapSelfManagedAddons bool `json:"bootstrapSelfManagedAddons,omitempty"`
+	BootstrapSelfManagedAddons *bool `json:"bootstrapSelfManagedAddons,omitempty"`
+
+	// EKSAutoMode indicates the EKS Auto Mode state for control-plane.
+	// If you set this value to false, the following params will be disabled for EKS:
+	// AWS::EKS::Cluster KubernetesNetworkConfig ElasticLoadBalancing Enabled -> false.
+	// AWS::EKS::Cluster StorageConfig blockStorage Enabled -> false.
+	// AWS::EKS::Cluster ComputeConfig Enabled -> false.
+	// +kubebuilder:default=true
+	EKSAutoMode *bool `json:"eksAutoMode"`
```
punkwalker: Can we name the field `autoMode` instead of `eksAutoMode`?
Also, IMO, this should not default to true and field should be optional as upstream EKS service does not default to AutoMode.

pkg/cloud/services/eks/cluster.go
```diff
@@ -297,9 +297,23 @@ func makeEksEncryptionConfigs(encryptionConfig *ekscontrolplanev1.EncryptionConf
...
-func makeKubernetesNetworkConfig(serviceCidrs *clusterv1.NetworkRanges) (*ekstypes.KubernetesNetworkConfigRequest, error) {
+func makeKubernetesNetworkConfig(serviceCidrs *clusterv1.NetworkRanges, isIPv6Enabled, isEKSAutoModeEnabled bool) (*ekstypes.KubernetesNetworkConfigRequest, error) {
+	netConfig := new(ekstypes.KubernetesNetworkConfigRequest)
+
+	// Should be explicitly defined.
+	if isEKSAutoModeEnabled {
+		netConfig.ElasticLoadBalancing = &ekstypes.ElasticLoadBalancing{Enabled: aws.Bool(true)}
+	} else {
+		netConfig.ElasticLoadBalancing = &ekstypes.ElasticLoadBalancing{Enabled: aws.Bool(false)}
+	}
```
punkwalker: I think we don't need this else as the field should be unset if autoMode is disabled.

pkg/cloud/services/eks/cluster.go
```diff
@@ -471,6 +478,14 @@ func (s *Service) createCluster(ctx context.Context, eksClusterName string) (*ek
 		Tags:                       tags,
 		KubernetesNetworkConfig:    netConfig,
 		BootstrapSelfManagedAddons: bootstrapAddon,
+		ComputeConfig:              &ekstypes.ComputeConfigRequest{Enabled: aws.Bool(false)},
+		StorageConfig:              &ekstypes.StorageConfigRequest{BlockStorage: &ekstypes.BlockStorage{Enabled: aws.Bool(false)}},
```
punkwalker: Let's keep the `ComputeConfig` and `StorageConfig` unset when autoMode disabled.

pkg/cloud/services/eks/cluster.go
```diff
@@ -471,6 +478,14 @@ func (s *Service) createCluster(ctx context.Context, eksClusterName string) (*ek
...
+	}
+
+	// Enable EKS Auto Mode if needed.
+	if s.scope.EKSAutoMode() {
+		input.ComputeConfig = &ekstypes.ComputeConfigRequest{Enabled: aws.Bool(true)}
+		input.StorageConfig = &ekstypes.StorageConfigRequest{BlockStorage: &ekstypes.BlockStorage{Enabled: aws.Bool(true)}}
```
punkwalker: The `computeConfig` has `nodepools` and if this is provided, it should also have `nodeRoleArn`. This is crucial if user wants to use default nodepools (general-purpose and system). When this is not specified, the user will need to create own nodepools.

Ref: https://docs.aws.amazon.com/eks/latest/APIReference/API_CreateCluster.html#AmazonEKS-CreateCluster-request-computeConfig

punkwalker (changes requested):
Thank you @CharlieR-o-o-t for working on this, I added some initial thoughts.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
CharlieR-o-o-t: yes, my bad. For some reason though it's default feature.

pkg/cloud/services/eks/cluster.go
CharlieR-o-o-t: good point, will handle nodepools/nodeRoleArn + validation

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go
```diff
@@ -572,6 +573,13 @@ func (*awsManagedControlPlaneWebhook) Default(_ context.Context, obj runtime.Obj
...
 	// Set default value for BootstrapSelfManagedAddons
-	r.Spec.BootstrapSelfManagedAddons = true
+	if r.Spec.BootstrapSelfManagedAddons == nil {
+		r.Spec.BootstrapSelfManagedAddons = aws.Bool(true)
+	}
+
+	// Set default value for AutoMode
+	if r.Spec.AutoMode == nil {
+		r.Spec.AutoMode = aws.Bool(false)
+	}
```
faiq: shouldn't the default be true considering this was always on by default in previous versions? a change like this could be breaking for users

CharlieR-o-o-t:
Thanks for review, will add new version soon.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go
CharlieR-o-o-t: agree, have changed that.

CharlieR-o-o-t:
@punkwalker , hello,

I’m unsure how we should handle the breaking change reported here: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5642/pull-cluster-api-provider-aws-apidiff-main/1965129880395845632

`bootstrapSelfManagedAddons` should be pointer, it's possible mistype in type definition, and it's not possible to set to false because of this:
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/60649f3ec6af96f294b1fea8b899977646ed090e/controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go#L575

Also I have added new field to struct `AccessConfig`, because `AutoMode` is not capable with CONFIG_MAP authenticationMode in EKS. + validation has been added with unit tests.

Thank you for your review, I appreciate it.

config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
```diff
@@ -2245,6 +2245,25 @@ spec:
...
+                properties:
+                  authenticationMode:
+                    description: |-
+                      AuthenticationMode mode controls how Kubernetes API authentication is performed:
+                        - CONFIG_MAP — uses only the aws-auth ConfigMap (legacy mode).
+                        - API — uses only EKS Access Entries (aws-auth is ignored).
+                        - API_AND_CONFIG_MAP — enables both Access Entries and aws-auth.
+                    type: string
+                  bootstrapAdminPermissions:
+                    description: |-
```
faiq, line 2354: Is it possible to rebase your changes off https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5578/files -- we're implementing the same changes there. 

Rebasing your changes off that would make the PR much smaller and less work down the line -- considering that has a lot of reviews already

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -221,6 +230,42 @@ type KubeProxy struct {
...
+	// +kubebuilder:default=false
+	Enabled bool `json:"enabled,omitempty"`
+	// Compute capability configuration for EKS Auto Mode.
+	// +optional
+	Compute Compute `json:"compute,omitempty"`
+}
+
+// Compute allows to run compute capability with EKS AutoMode.
+type Compute struct {
+	NodePools []string `json:"nodePools,omitempty"`
```
faiq: needs a comment describing the field.

config/crd/bases/controlplane.cluster.x-k8s.io_awsmanagedcontrolplanes.yaml
CharlieR-o-o-t: sure, will do

punkwalker:
> @punkwalker , hello,
> 
> I’m unsure how we should handle the breaking change reported here: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5642/pull-cluster-api-provider-aws-apidiff-main/1965129880395845632
> 
> `bootstrapSelfManagedAddons` should be pointer, it's possible mistype in type definition, and it's not possible to set to false because of this:
> 
> https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/60649f3ec6af96f294b1fea8b899977646ed090e/controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go#L575
> 
> Also I have added new field to struct `AccessConfig`, because `AutoMode` is not capable with CONFIG_MAP authenticationMode in EKS. + validation has been added with unit tests.
> 
> Thank you for your review, I appreciate it.

@CharlieR-o-o-t 
Can you confirm why are we changing `BootstrapSelfManagedAddons` from`bool` to `*bool`? The reason api-diff will fail because this change breaks API compatibility with previous API versions, so it would flag it. If you feel this change is necessary, Let's check with @richardcase about how to handle it!

pkg/cloud/services/eks/cluster.go
punkwalker: +1, Let's keep this PR focused on AutoMode.

richardcase:
> @CharlieR-o-o-t Can you confirm why are we changing `BootstrapSelfManagedAddons` from`bool` to `*bool`? The reason api-diff will fail because this change breaks API compatibility with previous API versions, so it would flag it. If you feel this change is necessary, Let's check with @richardcase about how to handle it!

Yeah we should probably change this back to `bool`. Was there a specific reason it was changed?

We will probably change all option `bool` fields to `*bool` in the future, like CAPI did (as part of v1beta2): https://github.com/kubernetes-sigs/cluster-api/pull/12436

CharlieR-o-o-t:
thank you, will change back to "bool". 
Will continue to work on this PR after merge of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5578

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook_test.go
```diff
@@ -877,6 +906,37 @@ func TestWebhookUpdate(t *testing.T) {
...
+						NodeRoleArn: aws.String("fooarn"),
+						NodePools:   []string{"pool1", "pool2"},
+					},
+				},
+			},
+			newClusterSpec: AWSManagedControlPlaneSpec{
+				EKSClusterName: "default_cluster1",
+				NetworkSpec: infrav1.NetworkSpec{
+					VPC: infrav1.VPCSpec{
+						IPv6: &infrav1.IPv6{},
```
Schmidt3000: This test should be testing one thing. It appears to also be testing whether or not you can switch to ipv6 after having it undeclared.

Schmidt3000 (changes requested):

musaprg:
xref: https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5278

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go
```diff
@@ -52,6 +53,9 @@ const (
 	cidrSizeMin    = 16
 	vpcCniAddon    = "vpc-cni"
 	kubeProxyAddon = "kube-proxy"
+
+	autoModeComputeNodePoolSystem  = "system"
+	autoModeComputeNodePoolGeneral = "general-purpose"
```
musaprg: We can implement this validation on the API side by using CRD validation schema instead of admission webhook.

https://book.kubebuilder.io/reference/markers/crd-validation

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -225,6 +228,29 @@ type KubeProxy struct {
...
+	Enabled bool `json:"enabled,omitempty"`
+	// Compute capability configuration for EKS Auto Mode.
+	// +optional
+	Compute Compute `json:"compute,omitempty"`
+}
+
+// Compute allows to run compute capability with EKS AutoMode.
+type Compute struct {
+	// NodePools that defines the compute resources for your EKS Auto Mode cluster.
+	NodePools []string `json:"nodePools,omitempty"`
```
musaprg: Why don't we use CRD validation marker here instead of checking values inside the admission webhook?

```
// Built-in node pools of EKS Auto Mode. 
// +kubebuilder:validation:Enum:=system;general-purpose
type AutoModeBuiltInNodePool string
```

```suggestion
	NodePools []AutoModeBuiltInNodePool `json:"nodePools,omitempty"`
```

https://book.kubebuilder.io/reference/markers/crd-validation

musaprg: One possible benefit of this is that we can use static validation on the API schema side. This would help kube-apiserver to reject invalid requests regarding this field before sending the request to the admissin webhook server.

musaprg (commented):
Hi. Thank you for working on this. I've left one minor suggestion.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
musaprg: I'm not sure but we could use `+kubebuilder:validation:items:Enum` instead and rewrite the above code like below.

```suggestion
        // +kubebuilder:validation:items:Enum:=system;general-purpose
	NodePools []string `json:"nodePools,omitempty"`
```

CharlieR-o-o-t: done

CharlieR-o-o-t:
Have rebased and made necessary changes. Ready for review.

PR #5643, opened 2025-09-03
🌱 enable requiredfields linter

### LINKED ISSUE #5492: [KAL] Enable the requiredfields linter
Enable the requiredfields linter.

v47:
**What type of PR is this?**

/kind cleanup

* Enables Kube API linter’s requiredfields check to ensure required API fields aren’t omitted or pointer types.
* Removes omitempty tags from required OIDC identity provider fields and MachinePool block device DeviceName fields to align with new linter rules and guarantee proper serialization.

Fixes #5492

**Checklist**:
- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:
```release-note
Enabled requiredfields Kube API linter check and removed `omitempty` tags from required EKS and experimental MachinePool API fields.
```


JoelSpeed:
Please update KAL to the latest version. After discussion with the Kube API reviewers, and updates to the Kube API conventions, the required fields rule has been rewritten and will now recommend `omitempty` on fields, so this PR is going in the opposite direction to the current best advice

richardcase:
Also the PR needs a retitle.

v47:
> Also the PR needs a retitle.

@richardcase  sorry,  do you have any hints about the title?

controlplane/eks/api/v1beta1/types.go
```diff
@@ -223,6 +223,7 @@ type OIDCIdentityProviderConfig struct {
 	// This is also known as audience. The ID for the client application that makes
 	// authentication requests to the OpenID identity provider.
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed: Have you verified that this ratchets?

controlplane/eks/api/v1beta1/types.go
```diff
@@ -250,6 +252,8 @@ type OIDCIdentityProviderConfig struct {
 	// and must be publicly accessible over the internet.
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:Pattern=`^https://.+`
```
JoelSpeed: Might be better to use a CEL validation if possible, `isURL` likely more comprehensive

JoelSpeed (commented):
In a lot of cases, changing the serialization can be breaking, though clients do need to update to the new library version for this to take effect.

Where we already have a non-zero min length, adding `omitempty`/`omitzero` should be safe, since no one should have persisted an empty value anyway

Where there aren't min lengths already, this could be problematic. Have you checked what the behaviour is when a zero value is persisted and you upgrade into this schema/client?

Existing values should be fine because of ratcheting validation, but, if the client now starts omitting the value, it will remove the value on write and this will lead to a required validation coming up, which would then be breaking

controlplane/eks/api/v1beta1/types.go
v47: Did you have something like this in mind?
```
// +kubebuilder:validation:XValidation:rule="isURL(self) && url(self).getScheme() == 'https'"
```

JoelSpeed: Exactly, and then you can add `,message="..."` with a custom error message too

exp/api/v1beta2/awsmachinepool_types.go
```diff
@@ -67,7 +67,7 @@ type AWSMachinePoolSpec struct {
 
 	// AWSLaunchTemplate specifies the launch template and version to use when an instance is launched.
 	// +kubebuilder:validation:Required
-	AWSLaunchTemplate AWSLaunchTemplate `json:"awsLaunchTemplate"`
+	AWSLaunchTemplate AWSLaunchTemplate `json:"awsLaunchTemplate,omitzero"`
```
damdo, line 70: I think doing this causes an issue in the tests because it omits the zero value for this which is required by the validation checks:

```
AWSMachinePool.infrastructure.cluster.x-k8s.io "test" is invalid: spec.awsLaunchTemplate: Required value
```

See: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5643/pull-cluster-api-provider-aws-test/1970412246974599168#

v47: yup, saw it, but now for me it's a question,  how required AWSLaunchTemplate should be treated?

controlplane/eks/api/v1beta1/types.go
v47: @JoelSpeed, the validation for https was addressed.

exp/api/v1beta2/awsmachinepool_types.go
v47: @damdo @JoelSpeed sorry, do you have an opinion about the changes related to AWSMachinePool?

damdo: How did you fix it? Did you run generation again?

v47: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5643/files#diff-870af3af95309601f80ec2a719be848510d92c57065a88a5f465ac646259db55R101
i added AWSLaunchTemplate in the tests.

damdo: Ok thanks, let's see what @JoelSpeed thinks of the overall changes

JoelSpeed: The linter is basically saying that you should require a non-zero AWSMachineTemplate. Does the AWSMachineTemplate have any required fields presently? Or can I today commit `{}` to the API and that be valid?

damdo:
@JoelSpeed anything pending to be addressed or are you happy with it?

JoelSpeed:
Some of my review comments, e.g. https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5643/changes#r2321528667, have not had a response yet

It's important to make sure that any changes we make here won't break currently stored invalid data, so ensuring ratcheting is important

damdo:
Thanks.
@v47 are you able to address @JoelSpeed 's comments? TY

exp/api/v1beta2/awsmachinepool_types.go
v47: At least one property must be non-empty. All its fields are optional (omitempty), so technically it's possible to commit `{}` to the API—it would fail the MinProperties validation.

JoelSpeed: Great, so, omitzero changing here makes sense in that case. We won't break anyone by doing so. The error message will change from requiring 1 property to saying the field is required, but that's more correct so this is a good change

v47:
@JoelSpeed i made changes to address the concerns. Can you please take a look again?

controlplane/eks/api/v1beta1/types.go
```diff
@@ -227,7 +227,7 @@ type OIDCIdentityProviderConfig struct {
 	// This is also known as audience. The ID for the client application that makes
 	// authentication requests to the OpenID identity provider.
 	// +kubebuilder:validation:Required
-	ClientID string `json:"clientId,omitempty"`
+	ClientID *string `json:"clientId,omitempty"`
```
JoelSpeed, line 230: What happens today if this value is empty? Is there any webhook validation that rejects this value being empty?

controlplane/eks/api/v1beta1/types.go
```diff
@@ -243,7 +243,7 @@ type OIDCIdentityProviderConfig struct {
 	//
 	// IdentityProviderConfigName is a required field
 	// +kubebuilder:validation:Required
-	IdentityProviderConfigName string `json:"identityProviderConfigName,omitempty"`
+	IdentityProviderConfigName *string `json:"identityProviderConfigName,omitempty"`
```
JoelSpeed, line 246: As above, what happens when this is the empty string?

controlplane/eks/api/v1beta1/types.go
```diff
@@ -254,7 +254,8 @@ type OIDCIdentityProviderConfig struct {
 	// and must be publicly accessible over the internet.
 	//
 	// +kubebuilder:validation:Required
-	IssuerURL string `json:"issuerUrl,omitempty"`
+	// +kubebuilder:validation:XValidation:rule="isURL(self) && url(self).getScheme() == 'https'"
```
JoelSpeed, line 257: Where did this validation come from? Is this enforced at a webhook level?

controlplane/eks/api/v1beta1/types.go
```diff
@@ -254,7 +254,8 @@ type OIDCIdentityProviderConfig struct {
...
+	IssuerURL *string `json:"issuerUrl,omitempty"`
```
JoelSpeed, line 258: What happens today when this field is the empty string?

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -285,7 +285,7 @@ type AccessConfig struct {
 type AccessEntry struct {
 	// PrincipalARN is the Amazon Resource Name (ARN) of the IAM principal
 	// +kubebuilder:validation:Required
-	PrincipalARN string `json:"principalARN"`
+	PrincipalARN *string `json:"principalARN,omitempty"`
```
JoelSpeed, line 288: What happens when this is the zero value, is that actually valid?

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_types.go
```diff
@@ -313,11 +313,11 @@ type AccessEntry struct {
 type AccessPolicyReference struct {
 	// PolicyARN is the Amazon Resource Name (ARN) of the access policy
 	// +kubebuilder:validation:Required
-	PolicyARN string `json:"policyARN"`
+	PolicyARN *string `json:"policyARN,omitempty"`
```
JoelSpeed, line 316: What happens when this is the zero value, is that actually valid?

controlplane/rosa/api/v1beta2/external_auth_types.go
```diff
@@ -46,9 +46,10 @@ type TokenIssuer struct {
 	// Must use the https:// scheme.
 	//
 	// +kubebuilder:validation:Pattern=`^https:\/\/[^\s]`
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 49: The minimum length is actually 8 right? `https://` must be in the value?

controlplane/rosa/api/v1beta2/external_auth_types.go
```diff
@@ -79,23 +80,24 @@ type OIDCClientConfig struct {
...
-	ComponentName string `json:"componentName"`
+	ComponentName string `json:"componentName,omitempty"`
 
 	// ComponentNamespace is the namespace of the component that is supposed to consume this
 	// client configuration
 	//
 	// +kubebuilder:validation:MinLength=1
 	// +kubebuilder:validation:MaxLength=63
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:Pattern=^[a-z0-9]([-a-z0-9]*[a-z0-9])?$
```
JoelSpeed, line 91: Where did this pattern come from? Already validated somewhere?

controlplane/rosa/api/v1beta2/external_auth_types.go
```diff
@@ -130,8 +132,9 @@ type PrefixedClaimMapping struct {
 	// Claim is a JWT token claim to be used in the mapping
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 135: Is this already validated?

controlplane/rosa/api/v1beta2/external_auth_types.go
```diff
@@ -151,8 +154,9 @@ type UsernameClaimMapping struct {
...
```
JoelSpeed, line 157: Is this already validated?

controlplane/rosa/api/v1beta2/external_auth_types.go
```diff
@@ -229,21 +233,22 @@ type TokenRequiredClaim struct {
...
-	RequiredValue string `json:"requiredValue"`
+	RequiredValue string `json:"requiredValue,omitempty"`
 }
 
 // LocalObjectReference references an object in the same namespace.
 type LocalObjectReference struct {
 	// Name is the metadata.name of the referenced object.
 	//
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 251: Is this already validated?

.golangci-kal.yml
```diff
@@ -25,17 +25,17 @@ linters:
...
               - "*" # We will manually enable new linters after understanding the impact. Disable all by default.
           lintersConfig:
-            conditions:
-              isFirstField: Warn # Require conditions to be the first field in the status struct.
-              usePatchStrategy: Forbid # Conditions should not use the patch strategy on CRDs.
-              useProtobuf: Forbid # We don't use protobuf, so protobuf tags are not required.
+            # conditions:
+            #   isFirstField: Warn # Require conditions to be the first field in the status struct.
+            #   usePatchStrategy: Forbid # Conditions should not use the patch strategy on CRDs.
+            #   useProtobuf: Forbid # We don't use protobuf, so protobuf tags are not required.
```
JoelSpeed, line 38: Why are you commenting out existing configuration?

exp/api/v1beta2/types.go
```diff
@@ -339,13 +341,15 @@ type Taint struct {
 	// Effect specifies the effect for the taint
 	// +kubebuilder:validation:Required
 	// +kubebuilder:validation:Enum=no-schedule;no-execute;prefer-no-schedule
-	Effect TaintEffect `json:"effect"`
+	Effect TaintEffect `json:"effect,omitempty"`
 	// Key is the key of the taint
 	// +kubebuilder:validation:Required
-	Key string `json:"key"`
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 347: Was this already being validated?

exp/api/v1beta2/types.go
```diff
@@ -339,13 +341,15 @@ type Taint struct {
...
+	Key string `json:"key,omitempty"`
 	// Value is the value of the taint
 	// +kubebuilder:validation:Required
-	Value string `json:"value"`
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 351: Was this already being validated?

exp/api/v1beta2/types.go
```diff
@@ -60,6 +61,7 @@ type BlockDeviceMapping struct {
 }
 
 // AWSLaunchTemplate defines the desired state of AWSLaunchTemplate.
+// +kubebuilder:validation:MinProperties=1
```
JoelSpeed, line 64: Was this already being validated?

exp/api/v1beta2/types.go
```diff
@@ -52,6 +52,7 @@ type EBS struct {
 type BlockDeviceMapping struct {
 	// The device name exposed to the EC2 instance (for example, /dev/sdh or xvdh).
 	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
```
JoelSpeed, line 55: Was this already being validated?

JoelSpeed (commented):
I've stopped commenting part way through, but for each change, we need to understand what the existing validations are. So for everywhere you've made a field a pointer, can you confirm what happens today when the value is the empty string, is it genuinely accepted by the API server. Everywhere where you have added a new minimum length, can you link to existing validation that enforces this?

v47:
> I've stopped commenting part way through, but for each change, we need to understand what the existing validations are. So for everywhere you've made a field a pointer, can you confirm what happens today when the value is the empty string, is it genuinely accepted by the API server. Everywhere where you have added a new minimum length, can you link to existing validation that enforces this?

Thanks for looking at it. I saw the issue with the schema change in CI, and it kept me thinking. Adding a pointer triggers it. 
Pointers allow the propper behavior on validation. But they definitely change the way it will be consumed.
@JoelSpeed i am a bit concerned that right now I don't see a way to have the requirefields linter on and no schema changes.
What do you feel is the right course?

JoelSpeed:
There are changes we can make, and changes we cannot, but they all depend on whether the field is validated in the webhook or not, so we need to tally the validations with the fields.

For each field we are making a change in, we need to review, what is the webhook validation. Does it require a non-empty value? 

If it requires a non empty value, we can retrofit kubebuilder markers to effect that change.

If it does not actually check for a non empty value, we may want to instead punt this change to the next API version. For now I'd suggest adding exceptions in the golangci-lint config file for this API version, and then when we introduce the next API version, we can make those changes.

We could also argue that there is ratcheting validation which we could rely on, but this requires a case by case analysis.

What we absolutely cannot do is change fields to pointers within the same API version.

I would be interested, and this is what I was trying to get to, in how you came to the decision on each field as to whether you want to add a minimum length or change it to a pointer?

PR #5675, opened 2025-09-22
🌱 test: Add dedicated unit tests for getInstanceCPUOptionsRequest

fangge1212:
**What type of PR is this?**
/kind cleanup

The unit tests for `getInstanceCPUOptionsRequest` are more lightweight and faster than testing various CPU option configurations through the larger `CreateInstance` function.

This commit refactors the existing tests by moving the specific CPU option test cases from `TestCreateInstance` into a new, more focused `TestGetInstanceCPUOptionsRequest` function. A single test case remains in `TestCreateInstance` to ensure the integration between the functions is correct.

Fixes #

**Checklist**:

- [X] includes emoji in title 

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/services/ec2/instances_test.go
+++ b/pkg/cloud/services/ec2/instances_test.go
@@ -5905,256 +5905,6 @@ func TestCreateInstance(t *testing.T) {
 				}
 			},
 		},
-		{
-			name: "with AMD SEV-SNP disabled",
-			machine: &clusterv1.Machine{
-				ObjectMeta: metav1.ObjectMeta{
-					Labels: map[string]string{"set": "node"},
-				},
-				Spec: clusterv1.MachineSpec{
-					Bootstrap: clusterv1.Bootstrap{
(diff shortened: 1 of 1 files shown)


pkg/cloud/services/ec2/instances_test.go
```diff
@@ -6766,3 +6516,51 @@ func TestGetCapacityReservationSpecification(t *testing.T) {
 		})
 	}
 }
+
+func TestGetInstanceCPUOptionsRequest (t *testing.T) {
+	testCases := []struct {
+		name            string
+		instance        *infrav1.Instance
```
AndiDog: Let's use only the `infrav1.CPUOptions` type here, since the rest of `infrav1.Instance` isn't even used in this test.

fangge1212: You are right. Updated. Thanks.

AndiDog:
Looks like an unrelated ROSA test failure. Maybe it's flaky, so let's try again.

PR #5677, opened 2025-09-22
✨ Skip AWSCluster reconciliation when owner Cluster is being deleted

fiunchinho:
**What type of PR is this?**

/kind feature

When a `Cluster` CR is deleted, CAPA takes care of deleting the dependant CRs in an specific order, i.e. first the worker nodes, then the control plane, and finally the infra cluster CR `AWSCluster`. Removing everything can take a little while, depending on the size of the cluster. While things are getting removed, the `awscluster_controllerl` keeps reconciling the `AWSCluster`, trying to keep the AWS resources such as the VPC or subnets up to date. But these resources will be removed shortly. 
With the changes on this PR, we are skipping the reconciliation of the `AWSCluster` when the owner `Cluster` has been set for deletion. That way we skip reconciliation loops done on resources that are about to be removed, saving CAPA resources and AWS API calls.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Skip AWSCluster reconciliation when owner Cluster is being deleted
```

### DIFF
--- a/controllers/awscluster_controller.go
+++ b/controllers/awscluster_controller.go
@@ -312,6 +312,11 @@ func (r *AWSClusterReconciler) reconcileLoadBalancer(ctx context.Context, cluste
 }
 
 func (r *AWSClusterReconciler) reconcileNormal(ctx context.Context, clusterScope *scope.ClusterScope) (reconcile.Result, error) {
+	if !clusterScope.Cluster.DeletionTimestamp.IsZero() {
+		clusterScope.Info("Cluster owning the AWSCluster is being deleted, skipping reconciliation")
+		return reconcile.Result{}, nil
+	}
+
 	clusterScope.Info("Reconciling AWSCluster")
 
 	awsCluster := clusterScope.AWSCluster
--- a/controllers/awscluster_controller_unit_test.go
+++ b/controllers/awscluster_controller_unit_test.go
@@ -73,6 +73,12 @@ func TestAWSClusterReconcilerReconcile(t *testing.T) {
 			expectError:  false,
 		},
 		{
+			name:         "Should not Reconcile if owner cluster is deleted",
+			awsCluster:   &infrav1.AWSCluster{ObjectMeta: metav1.ObjectMeta{GenerateName: "aws-test-"}},
+			ownerCluster: &clusterv1.Cluster{ObjectMeta: metav1.ObjectMeta{GenerateName: "capi-test-", DeletionTimestamp: &metav1.Time{Time: time.Now()}}},
+			expectError:  false,
+		},
+		{
 			name:        "Should Reconcile successfully if no AWSCluster found",
 			expectError: false,
 		},


damdo:
Is this the recommended way of dealing with the reconciliation of the ProviderCluster when the owning Cluster is being deleted?

What are other providers doing in this occasion?

I had a quick look at CAPV and CAPG but they don't seem to be doing that:
- https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/controllers/vspherecluster_reconciler.go
- https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/controllers/gcpcluster_controller.go

damdo:
^ cc. @chrischdi

chrischdi:
I agree with @damdo on this.

From core CAPI perspective this is not defined clearly I think (did not find something clearly about this in the InfraCluster contract [xref](https://main.cluster-api.sigs.k8s.io/developer/providers/contracts/infra-cluster#deleted-resource)). But it also does not mention the deletion should be the way of the issue description. So I'd conclude it should not be like that.

From my point of view: the infracluster provider should take care of keeping the environment usable, so deletion can succeed successfully.

So for the sake of making deletion reliable, IMHO CAPA should continue to reconcile the AWSCluster. Otherwise this could lead to even take longer for getting deleted and maybe produce more costs.
E.g. if CAPI is not able anymore to drain machines in a normal way, or CSI is not able to detach volumes and CAPI's waiting for successful detachment, etc.

Do we have numbers on how much it could save in terms of CAPA resources and AWS API calls?
Maybe the reconciliation could be optimized in other ways instead to be more efficient, and not only for AWSCluster's that have their Cluster marked for deletion.

fiunchinho:
> Is this the recommended way of dealing with the reconciliation of the ProviderCluster when the owning Cluster is being deleted?
> 
> What are other providers doing in this occasion?
> 
> I had a quick look at CAPV and CAPG but they don't seem to be doing that:
> 
> * https://github.com/kubernetes-sigs/cluster-api-provider-vsphere/blob/main/controllers/vspherecluster_reconciler.go
> * https://github.com/kubernetes-sigs/cluster-api-provider-gcp/blob/main/controllers/gcpcluster_controller.go

I don't think it's defined anywhere, it seems more like an oversight. But I don't see any downside of doing it this way, only the advantage of being more efficient.

> I agree with @damdo on this.
> 
> From core CAPI perspective this is not defined clearly I think (did not find something clearly about this in the InfraCluster contract [xref](https://main.cluster-api.sigs.k8s.io/developer/providers/contracts/infra-cluster#deleted-resource)). But it also does not mention the deletion should be the way of the issue description. So I'd conclude it should not be like that.

> From my point of view: the infracluster provider should take care of keeping the environment usable, so deletion can succeed successfully.
> 
> So for the sake of making deletion reliable, IMHO CAPA should continue to reconcile the AWSCluster. Otherwise this could lead to even take longer for getting deleted and maybe produce more costs. E.g. if CAPI is not able anymore to drain machines in a normal way, or CSI is not able to detach volumes and CAPI's waiting for successful detachment, etc.
> 
> Do we have numbers on how much it could save in terms of CAPA resources and AWS API calls? Maybe the reconciliation could be optimized in other ways instead to be more efficient, and not only for AWSCluster's that have their Cluster marked for deletion.

I'll close this if everyone is against it. Please, would you take care of adding this to the contract so that developers know what's mandatory from infra providers?

PR #5603, opened 2025-07-30
✨ IPv6 support for self-managed clusters

tthvo:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

As of today, CAPA supports IPv6 on EKS, but not self-managed clusters. Thus, these changes bring IPv6 support for self-managed clusters, both single-stack IPv6 and dualstack.

**Which issue(s) this PR fixes**:

Fixes #2420
Fixes #3381 

**Special notes for your reviewer**:

- EC2 instance type:
  - The networking capabilities of the instance type must support IPv4  
  - For launching instances in IPv6-only subnets, only nitro-based instance type is supported. 
- CNI: The CNI plugin needs to support IPV6. I include sample manifests in `test/e2e/data/cni`. Calico does not support IPv6 with "IP-in-IP" so we need to use VXLAN.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Enable IPv6 support for self-managed kubernetes clusters
```

PR #5684, opened 2025-09-24
🐛 fix: use cluster tag key to list managed egress-only internet gateway

### LINKED ISSUE #5683: [Bug] [EKS] CAPA deletes all EgressOnlyInternetGateways during IPv6 cluster deprovision
/kind bug

When deprovisioning an IPv6 for EKS cluster, CAPA is supposed to delete the eigw attached to the cluster-owned VPC. However, we observed that CAPA deleted **all** eigw in the region regardless of whether the eigw is attached to cluster-owned VPC or not.

This is very catastrophic as it can break other workloads in other VPCs.

CAPA should only clean up eigw that it manages.

The API for `DescribeEgressOnlyInternetGateways` does not support `attachment.vpc-id` filter. See [API Reference](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeEgressOnlyInternetGateways.html )) and below code reference from AWS SDK [reference](https://github.com/aws/aws-sdk-go-v2/blob/main/service/ec2/api_op_DescribeEgressOnlyInternetGateways.go):

```go
type DescribeEgressOnlyInternetGatewaysInput struct {
	// The filters.
	//
	//    * tag:<key> - The key/value combination of a tag assigned to the resource.
	//    Use the tag key in the filter name and the tag value as the filter value.
	//    For example, to find all resources that have a tag with the key Owner
	//    and the value TeamA, specify tag:Owner for the filter name and TeamA for
	//    the filter value.
	//
	//    * tag-key - The key of a tag assigned to the resource. Use this filter
	//    to find all resources assigned a tag with a specific key, regardless of
	//    the tag value.
	Filters []*Filter `locationName:"Filter" locationNameList:"Filter" type:"list"`
```

This means the AWS API will return **all** eigws. That being said, during provisioning steps, if there are more than 1 eigw available, CAPA might select the wrong eigw, instead of creating one for the cluster-owned VPC.

**Environment:**

- Cluster-api-provider-aws version: Latest `main`

tthvo:
**What type of PR is this?**
/kind bug

The API for `DescribeEgressOnlyInternetGateways` does not support `attachment.vpc-id` filter. Thus, the call will return all available eigw. Consequences:
- CAPA incorrectly selects an unintended eigw for use. Leading to route creation failure since the eigw belongs to a different VPC.
- CAPA incorrectly destroys all eigw of all VPCs. This is very catastrophic as it can break other workloads.

This changes the filter to use cluster tag instead. Additional safeguard is also included to check if the eigw is truly attached the VPC.

Fixes #5683 5683

This is a cherry-pick from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5603/commits/c4dff2a36bba4a24b47e1bda3b2217455cbcd7d0 (self-managed IPv6 cluster is in progress).

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Ensure CAPA only manages cluster-owned egress only internet gateway
```


pkg/cloud/services/network/egress_only_gateways.go
```diff
@@ -136,21 +136,35 @@ func (s *Service) createEgressOnlyInternetGateway() (*types.EgressOnlyInternetGa
...
+	eigws := make([]types.EgressOnlyInternetGateway, 0)
+	for _, eigw := range out.EgressOnlyInternetGateways {
+		for _, attachment := range eigw.Attachments {
+			if aws.ToString(attachment.VpcId) == s.scope.VPC().ID {
+				eigws = append(eigws, eigw)
+			}
+		}
+	}
+
+	if len(eigws) == 0 {
```
mtulio, line 163: It may not be possible, but do we need to check if API returned more than one object too - logging some unexpected behavior?

mtulio (commented):
one nit, overall looks good to me

/lgtm

pkg/cloud/services/network/egress_only_gateways.go
tthvo: Right, this should never happen as a vpc can only never have 1 eigw attached.

But we can never be too careful 😅 so I added a check and also updated the unit test to reflect the case. Thanks!

PR #5690, opened 2025-10-01
🌱 Bump ginkgo timeout

damdo:
**What type of PR is this?**
/kind support

Some e2e jobs have grown and need more than 5h now.
So let's bump it until we get around to decide what to do with them.
The overall job timeout in prow is 5h, let's use all of the available time.

```release-note
NONE
```


damdo:
cc. @phuhung273 @joshfrench

phuhung273:
Love this, thanks @damdo . Now i know we also have this internal timeout instead of test-infra

PR #5711, opened 2025-10-15
✨ feat: Implement autoscaling from zero by auto-populating AWSMachineTemplate capacity/nodeInfo

LiangquanLi930:
✨ Implement autoscaling from zero by auto-populating AWSMachineTemplate capacity and NodeInfo

**What type of PR is this?**
/kind feature

This PR implements the [Cluster API autoscaling from zero proposal](https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20210310-opt-in-autoscaling-from-zero.md) for CAPA by adding a controller that automatically populates `AWSMachineTemplate.Status.Capacity` with instance type information.

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Add autoscaling from zero support with auto-population of AWSMachineTemplate capacity/nodeInfo
```


controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,193 @@
...
+		if apierrors.IsNotFound(err) {
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Skip if capacity is already set
+	if len(awsMachineTemplate.Status.Capacity) > 0 {
+		return ctrl.Result{}, nil
+	}
```
chrischdi: What if the capacity is not correct? Or is the .spec.template.spec.instanceType immutable?

controllers/awsmachinetemplate_controller_unit_test.go
```diff
@@ -0,0 +1,295 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
+	clusterv1 "sigs.k8s.io/cluster-api/api/v1beta1"
+)
+
+func TestAWSMachineTemplateReconciler(t *testing.T) {
```
chrischdi, line 37: Could there also be a unit test which expects actual correct values instead of just checking "capacity" is set?

elmiko:
`nodeInfo` would be nice to have, architecture should be relatively easy to determine but i'm not sure about operating system.

that said, `nodeInfo` is an extension to the original opt-in scale from zero, if arch and o/s are not present the autoscaler will still be able to scale based on resources alone.

controllers/awsmachinetemplate_controller.go
LiangquanLi930: IIUC, spec.template.spec.instanceType is immutable: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/427f450d6acdca574cfb32755edcc47b662b71d1/api/v1beta2/awsmachinetemplate_webhook.go#L230-L240 
The webhook enforces that spec.template.spec cannot be changed after creation, except when `topology.ShouldSkipImmutabilityChecks()` returns true (which is only for ClusterClass topology patches).

Therefore, I think the current early-return logic is correct.

chrischdi: Sounds good

controllers/awsmachinetemplate_controller_unit_test.go
elmiko: +1

elmiko (commented):
in general this is looking good to me, i tend to agree with Christian's questions.

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,259 @@
...
+	// Memory
+	if info.MemoryInfo != nil && info.MemoryInfo.SizeInMiB != nil {
+		memoryBytes := *info.MemoryInfo.SizeInMiB * 1024 * 1024
+		resourceList[corev1.ResourceMemory] = *resource.NewQuantity(memoryBytes, resource.BinarySI)
+	}
+
+	// Extract node info from AMI if available
+	nodeInfo := &infrav1.NodeInfo{}
+	amiID := template.Spec.Template.Spec.AMI.ID
+	if amiID != nil && *amiID != "" {
```
LiangquanLi930: @elmiko @chrischdi Hi
I have a question: when an AWSMachineTemplate does not specify an AMI.ID, how should this be handled? Should we keep it as it is, or should we retrieve the AMI.ID in some way?

LiangquanLi930: Or in other words, do we have any other ways to obtain the architecture and OS?

elmiko: good question, i am not familiar enough with the aws sdk to know if we could query in another way. @damdo and @nrb have some experience with CAPA, perhaps they might have a suggestion.

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,259 @@
...
+	var arch infrav1.Architecture
+	switch image.Architecture {
+	case ec2types.ArchitectureValuesX8664:
+		arch = infrav1.ArchitectureAmd64
+	case ec2types.ArchitectureValuesArm64:
+		arch = infrav1.ArchitectureArm64
+	}
+
+	// Determine OS - check Platform field first (specifically for Windows identification)
+	var os string
```
nrb: Maybe we could initialize this with `linux` and only change to `windows` if the `image.Platform` or the `image.PlatformDetails` say it's windows? A fairly minor change, but it would eliminate the check at L242.

controllers/awsmachinetemplate_controller.go
nrb: Here's what we do when there's no AMI ID defined for a Machine: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/ec2/instances.go#L144

I would say we could probably re-use https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/ec2/ami.go#L174, and if that doesn't get what we need, it's an error and the user needs to give us either the AMI ID or some of the other lookup data.

api/v1beta2/awsmachinetemplate_types.go
```diff
@@ -23,13 +23,51 @@ import (
...
+
+// NodeInfo contains information about the node's architecture and operating system.
+type NodeInfo struct {
+	// Architecture is the CPU architecture of the node.
+	// Its underlying type is a string and its value can be any of amd64, arm64.
+	// +optional
+	Architecture Architecture `json:"architecture,omitempty"`
+	// OperatingSystem is a string representing the operating system of the node.
+	// This may be a string like 'linux' or 'windows'.
+	// +optional
```
chrischdi, line 58:
```suggestion
	// +kubebuilder:validation:Enum=linux;windows
	// +optional
```

api/v1beta2/awsmachinetemplate_types.go
```diff
@@ -23,13 +23,51 @@ import (
...
+	OperatingSystemLinux = "linux"
+	// OperatingSystemWindows represents the Windows operating system.
+	OperatingSystemWindows = "windows"
+)
+
+// NodeInfo contains information about the node's architecture and operating system.
+type NodeInfo struct {
+	// Architecture is the CPU architecture of the node.
+	// Its underlying type is a string and its value can be any of amd64, arm64.
+	// +optional
```
chrischdi, line 53:
```suggestion
	// +kubebuilder:validation:Enum=amd64;arm64
	// +optional
```

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,257 @@
...
+		if strings.Contains(platformDetails, infrav1.OperatingSystemWindows) {
+			os = infrav1.OperatingSystemWindows
+		}
+	}
+
+	return arch, os, nil
+}
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *AWSMachineTemplateReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
```
chrischdi: Let's put this function to the top for consistency.

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,257 @@
...
+	log := logger.FromContext(ctx)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&infrav1.AWSMachineTemplate{}).
+		WithOptions(options).
+		WithEventFilter(predicates.ResourceHasFilterLabel(mgr.GetScheme(), log.GetLogger(), r.WatchFilterValue)).
```
chrischdi:
```suggestion
		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetScheme(), log.GetLogger(), r.WatchFilterValue)).
```

controllers/awsmachinetemplate_controller.go
chrischdi: 👍 

That then requires to lookup the kubernetes version first, which should be set at the parent object (you should be able to get it via the owner reference).

Also this may lead to requiring updates so we should remove not skip in line 72 if ami is not set.

LiangquanLi930:
@elmiko @nrb @chrischdi Thanks for your review! When you have time, could you help to review again? Thanks!

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+	// CPU
+	if info.VCpuInfo != nil && info.VCpuInfo.DefaultVCpus != nil {
+		resourceList[corev1.ResourceCPU] = *resource.NewQuantity(int64(*info.VCpuInfo.DefaultVCpus), resource.DecimalSI)
+	}
+
+	// Memory
+	if info.MemoryInfo != nil && info.MemoryInfo.SizeInMiB != nil {
+		memoryBytes := *info.MemoryInfo.SizeInMiB * 1024 * 1024
+		resourceList[corev1.ResourceMemory] = *resource.NewQuantity(memoryBytes, resource.BinarySI)
+	}
```
LiangquanLi930, line 240: I'm not sure whether my handling of this part is correct, and I’m also uncertain about which unit I should be using.

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,376 @@
...
+			return ctrl.Result{}, nil
+		}
+		return ctrl.Result{}, err
+	}
+
+	// Skip if capacity and nodeInfo are already set
+	if len(awsMachineTemplate.Status.Capacity) > 0 && awsMachineTemplate.Status.NodeInfo != nil {
+		return ctrl.Result{}, nil
+	}
+
```
chrischdi:
```suggestion
```

Either drop the skip, or do not skip if `template.Spec.Template.Spec.AMI.ID` is empty.

LiangquanLi930: updated

config/rbac/role.yaml
```diff
@@ -101,6 +101,7 @@ rules:
   - controlplane.cluster.x-k8s.io
   resources:
   - '*'
+  - kubeadmcontrolplanes
```
chrischdi: `kubeadmcontrolplanes` should already be part of `"*"`.

api/v1beta2/awsmachinetemplate_types.go
```diff
@@ -40,6 +78,7 @@ type AWSMachineTemplateSpec struct {
 // +kubebuilder:object:root=true
 // +kubebuilder:resource:path=awsmachinetemplates,scope=Namespaced,categories=cluster-api,shortName=awsmt
 // +kubebuilder:storageversion
+// +kubebuilder:subresource:status
```
chrischdi, line 89: Can we just make it be a subresource now or is this a breaking change?

(Question to maintainers).

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,371 @@
...
+// getInstanceTypeCapacity queries AWS EC2 API for instance type capacity information.
+// Returns the resource list (CPU, Memory).
+func (r *AWSMachineTemplateReconciler) getInstanceTypeCapacity(ctx context.Context, ec2Client *ec2.Client, instanceType string) (corev1.ResourceList, error) {
+	// Query instance type information
+	input := &ec2.DescribeInstanceTypesInput{
+		InstanceTypes: []ec2types.InstanceType{ec2types.InstanceType(instanceType)},
+	}
+
+	result, err := ec2Client.DescribeInstanceTypes(ctx, input)
```
chrischdi, line 219: I wonder how expensive it will be to always reach out here to AWS or if we need some caching, or does the client have some caching included?

Same question for other API calls to AWS.

controllers/awsmachinetemplate_controller.go
elmiko: the machine-api annotations used a calculation of MiB for its value. for the cluster-api implementation we should simply be using a `Quantity` for this value, no need to convert to `MiB`.

you can see [an example](https://github.com/kubernetes/autoscaler/tree/master/cluster-autoscaler/cloudprovider/clusterapi#scale-from-zero-support) in the cluster-api autoscaler docs.

i /think/ the logic you have here is correct though, but i don't think we need to do the conversion. unless we need it for the `Quantity`.

elmiko (commented):
i think this is looking good, although i'm not sure we need to do the conversion for the memory value.

api/v1beta2/awsmachinetemplate_types.go
LiangquanLi930: @nrb Could you help to answer it? Thanks!

controllers/awsmachinetemplate_controller.go
LiangquanLi930: updated

controllers/awsmachinetemplate_controller.go
LiangquanLi930: @chrischdi Thanks for your review!
IMHO, if we add caching, we should only cache capacity, not nodeInfo. NodeInfo depends on the AMI configuration, and IIUC, each AWSMachineTemplate typically has its own unique AMI (either via AMI.ID or ImageLookup* parameters). Given that AMIs are unlikely to be reused across templates, I think we should keep the current implementation unchanged.
WDYT?

chrischdi: > Given that AMIs are unlikely to be reused across templates, I think we should keep the current implementation unchanged.

I could think of a user having multiple clutsers in different namespaces using the same AMI.

I'm not an expert for the aws sdk.

LiangquanLi930: @nrb Could you help to check it? Thanks!

api/v1beta2/awsmachinetemplate_types.go
```diff
@@ -23,13 +23,51 @@ import (
...
+	// Architecture is the CPU architecture of the node.
+	// Its underlying type is a string and its value can be any of amd64, arm64.
+	// +kubebuilder:validation:Enum=amd64;arm64
+	// +optional
+	Architecture Architecture `json:"architecture,omitempty"`
+	// OperatingSystem is a string representing the operating system of the node.
+	// This may be a string like 'linux' or 'windows'.
+	// +kubebuilder:validation:Enum=linux;windows
+	// +optional
+	OperatingSystem string `json:"operatingSystem,omitempty"`
```
richardcase: If its limited to the 2 values in the enum then we should create the type:

```go
type OperatingSystem string
```

api/v1beta2/awsmachinetemplate_types.go
richardcase: I think this should be ok as its only CAPA updating the status.

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,370 @@
...
+	cluster, err := util.GetOwnerCluster(ctx, r.Client, template.ObjectMeta)
+	if err != nil {
+		return "", err
+	}
+	if cluster == nil {
+		return "", errors.New("no owner cluster found")
+	}
+
+	// Get region from AWSCluster (standard EC2-based cluster)
+	if cluster.Spec.InfrastructureRef != nil && cluster.Spec.InfrastructureRef.Kind == "AWSCluster" {
```
richardcase: `AWSCluster` is for an non-EKS cluster. We also need to consider EKS which uses a combination of `AWSManagedCluster` and `AWSManagedControlPlane`. As an example of the template that you would use: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/templates/cluster-template-eks.yaml

controllers/awsmachinetemplate_controller.go
richardcase: We don't cache responses currently in CAPA. 

There have been periods of time when we've tried to reduce the number of API calls.

main.go
```diff
@@ -407,6 +407,15 @@ func setupReconcilersAndWebhooks(ctx context.Context, mgr ctrl.Manager,
 		}
 	}
 
+	setupLog.Debug("enabling AWSMachineTemplate controller")
```
richardcase: This should be in the same scope as the `AWSMachine` controller so that it's included in the same block.....as we have flags to disable certain controllers.

api/v1beta2/awsmachinetemplate_types.go
LiangquanLi930: Thanks! Resolve conversation

controllers/awsmachinetemplate_controller.go
LiangquanLi930: updated

main.go
LiangquanLi930: updated.

controllers/awsmachinetemplate_controller.go
LiangquanLi930: Thanks for your review. I completely agree with the idea of reducing API calls. I’ve reviewed the current code again and made some adjustments.

LiangquanLi930:
@richardcase @nrb @chrischdi Thanks for your review again! When you have time, could you help to review again? Thanks!

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,393 @@
...
+			case ec2types.ArchitectureTypeX8664:
+				architecture = ec2service.Amd64ArchitectureTag
+				nodeInfo.Architecture = infrav1.ArchitectureAmd64
+			case ec2types.ArchitectureTypeArm64:
+				architecture = ec2service.Arm64ArchitectureTag
+				nodeInfo.Architecture = infrav1.ArchitectureArm64
+			}
+		} else {
+			return nil, errors.Errorf("instance type must support exactly one architecture, got %d", len(instanceTypeInfo.ProcessorInfo.SupportedArchitectures))
+		}
```
chrischdi: Could it be that this part is unnecessary work, because the gathered information would be overwritten below anyway via the default AMI?

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,393 @@
...
+				ec2Client,
+				template.Spec.Template.Spec.ImageLookupOrg,
+				template.Spec.Template.Spec.ImageLookupBaseOS,
+				kubernetesVersion,
+				architecture,
+				template.Spec.Template.Spec.ImageLookupFormat,
+			)
+			if err == nil && image != nil {
+				// Successfully found AMI, extract accurate nodeInfo from it
+				arch, os, _ := r.getNodeInfoFromAMI(ctx, ec2Client, *image.ImageId)
```
chrischdi: If we keep this code we should change `getNodeInfoFromAMI` to directly take the image as argument so we don't do the additional API call of getting the same information again.

chrischdi: Could be e.g.:

```
func (r *AWSMachineTemplateReconciler) getNodeInfoFromAMI(image ec2types.Image) *infrav1.NodeInfo {
```

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,393 @@
...
+	if info.MemoryInfo != nil && info.MemoryInfo.SizeInMiB != nil {
+		resourceList[corev1.ResourceMemory] = resource.MustParse(fmt.Sprintf("%dMi", *info.MemoryInfo.SizeInMiB))
+	}
+
+	return resourceList, nil
+}
+
+// getNodeInfo queries node information (architecture and OS) for the AWSMachineTemplate.
+// It uses AMI ID if specified, otherwise attempts AMI lookup or falls back to instance type info.
+func (r *AWSMachineTemplateReconciler) getNodeInfo(ctx context.Context, ec2Client *ec2.Client, template *infrav1.AWSMachineTemplate, instanceType string) (*infrav1.NodeInfo, error) {
```
chrischdi, line 250: This function is still a bit confusing to me. Can we have proper comments what we try to do?

And maybe use early returns instead of if elsing till the end.

E.g.:

```go
// Resolve via ami
if amiID := ptr.Deref(template.Spec.Template.Spec.AMI.ID, ""); amiID != "" {
  // do work
  return nodeInfo, nil
}

// Resolve via InstanceType
...
// do work
...
  return nodeInfo, nil
}

// Resolve via default AMI
...
return nodeinfo,nil

```

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,393 @@
...
+	for _, kcp := range kcpList.Items {
+		if kcp.Spec.MachineTemplate.InfrastructureRef.Kind == "AWSMachineTemplate" &&
+			kcp.Spec.MachineTemplate.InfrastructureRef.Name == template.Name &&
+			kcp.Spec.Version != "" {
+			return kcp.Spec.Version, nil
+		}
+	}
+
+	return "", errors.New("no MachineDeployment or KubeadmControlPlane found referencing this AWSMachineTemplate with a version")
+}
```
chrischdi, line 399: We could improve this by setting another label selector based on the owning cluster.

```suggestion
func (r *AWSMachineTemplateReconciler) getKubernetesVersion(ctx context.Context, template *infrav1.AWSMachineTemplate) (string, error) {
	listOpts, err := getParentListOptions(template.ObjectMeta)
	if err != nil {
		return "", errors.Wrap(err, "failed to get parent list options")
	}

	// Try to find version from MachineDeployment first
	machineDeploymentList := &clusterv1.MachineDeploymentList{}
	if err := r.List(ctx, machineDeploymentList, listOpts...); err != nil {
		return "", errors.Wrap(err, "failed to list MachineDeployments")
	}

	// Find MachineDeployments that reference this AWSMachineTemplate
	for _, md := range machineDeploymentList.Items {
		if version := ptr.Deref(md.Spec.Template.Spec.Version, ""); md.Spec.Template.Spec.InfrastructureRef.Kind == "AWSMachineTemplate" &&
			md.Spec.Template.Spec.InfrastructureRef.Name == template.Name && version != "" {
			return version, nil
		}
	}

	// If not found in MachineDeployment, try KubeadmControlPlane
	kcpList := &controlplanev1.KubeadmControlPlaneList{}
	if err := r.List(ctx, kcpList, listOpts...); err != nil {
		return "", errors.Wrap(err, "failed to list KubeadmControlPlanes")
	}

	// Find KubeadmControlPlanes that reference this AWSMachineTemplate
	for _, kcp := range kcpList.Items {
		if kcp.Spec.MachineTemplate.InfrastructureRef.Kind == "AWSMachineTemplate" &&
			kcp.Spec.MachineTemplate.InfrastructureRef.Name == template.Name &&
			kcp.Spec.Version != "" {
			return kcp.Spec.Version, nil
		}
	}

	return "", errors.New("no MachineDeployment or KubeadmControlPlane found referencing this AWSMachineTemplate with a version")
}

func getParentListOptions(obj metav1.ObjectMeta) ([]client.ListOption, error) {
	listOpts := []client.ListOption{
		client.InNamespace(obj.Namespace),
	}

	for _, ref := range obj.GetOwnerReferences() {
		if ref.Kind != "Cluster" {
			continue
		}
		gv, err := schema.ParseGroupVersion(ref.APIVersion)
		if err != nil {
			return nil, errors.WithStack(err)
		}
		if gv.Group == clusterv1.GroupVersion.Group {
			listOpts = append(listOpts, client.MatchingLabels{
				clusterv1.ClusterNameLabel: ref.Name,
			})
			break
		}
	}
	return listOpts, nil
}
```

controllers/awsmachinetemplate_controller.go
LiangquanLi930: If ec2service.DefaultAMILookup fails, this part of the code is necessary.

LiangquanLi930: Or we can accept that nodeinfo is nil when ec2service.DefaultAMILookup fails.

LiangquanLi930: @chrischdi  WAYT?

controllers/awsmachinetemplate_controller.go
LiangquanLi930: Thanks!

controllers/awsmachinetemplate_controller.go
chrischdi: Ah ok, so below is a fallback 👍

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,410 @@
...
+
+// SetupWithManager sets up the controller with the Manager.
+func (r *AWSMachineTemplateReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
+	log := logger.FromContext(ctx)
+
+	return ctrl.NewControllerManagedBy(mgr).
+		For(&infrav1.AWSMachineTemplate{}).
+		WithOptions(options).
+		WithEventFilter(predicates.ResourceNotPausedAndHasFilterLabel(mgr.GetScheme(), log.GetLogger(), r.WatchFilterValue)).
```
damdo, line 68: @chrischdi is this `predicates.ResourceNotPausedAndHasFilterLabel` the new equivalent to `paused.EnsurePausedCondition` later on in the reconcile loop ? I guess it is similar but behaves differently as the `paused.EnsurePausedCondition` check allows for doing extra steps before deciding to stop reconciling the resource right?

chrischdi: Yes, these two (the predicate and the EnsurePausedCondition function) are supposed to work together.

controllers/awsmachinetemplate_controller.go
```diff
@@ -0,0 +1,410 @@
...
+	return resourceList, nil
+}
+
+// getNodeInfo queries node information (architecture and OS) for the AWSMachineTemplate.
+// It attempts to resolve nodeInfo using three strategies in order of priority:
+//  1. Directly from explicitly specified AMI ID
+//  2. From default AMI lookup (requires Kubernetes version from owner MachineDeployment/KubeadmControlPlane)
+//  3. From instance type architecture (OS cannot be determined, only architecture)
+func (r *AWSMachineTemplateReconciler) getNodeInfo(ctx context.Context, ec2Client *ec2.Client, template *infrav1.AWSMachineTemplate, instanceType string) (*infrav1.NodeInfo, error) {
+	if amiID := ptr.Deref(template.Spec.Template.Spec.AMI.ID, ""); amiID != "" {
```
chrischdi, line 252:
```suggestion
	// Strategy 1: Extract nodeInfo from the AMI if an ID is set.
	if amiID := ptr.Deref(template.Spec.Template.Spec.AMI.ID, ""); amiID != "" {
```

controllers/awsmachinetemplate_controller.go
damdo: So are we saying we need @LiangquanLi930 to also add `paused.EnsurePausedCondition` ?

LiangquanLi930:
squash commits

controllers/awsmachinetemplate_controller.go
LiangquanLi930: Updated: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5711/commits/03683f1998ae12b8631525d5c0346f991310dd69

LiangquanLi930:
rebase https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5720

damdo:
Hey @LiangquanLi930 can you squash the rebase commit and stack your ones on top? Thanks

LiangquanLi930:
> Hey @LiangquanLi930 can you squash the rebase commit and stack your ones on top? Thanks

@damdo Hi, I’m keeping the commits separate for now to make rebasing and reviewing easier.  Once you think everything looks good, I’ll squash the commits.

richardcase:
I know it approved but also had a look myself and i would 've also

/approve

chrischdi:
@LiangquanLi930 could you please take a look at the e2e test?

It seems to be very flaky!

See: [k8s-triage](https://storage.googleapis.com/k8s-triage/index.html?text=NodeInfo&job=.*cluster-api-provider-aws.*)

Example: https://prow.k8s.io/view/gs/kubernetes-ci-logs/logs/periodic-cluster-api-provider-aws-e2e-eks-canary/1992633554630086656

PR #5738, opened 2025-11-13
🐛 fix: ensure correct APIVersion in AWSClusterControllerIdentity conversion

dntosas:
/kind bug

The conversion webhook for AWSClusterControllerIdentity was failing with:
- expected infrastructure.cluster.x-k8s.io/v1beta2, received infrastructure.cluster.x-k8s.io/v1beta1
- Error from server: conversion webhook returned invalid object: invalid groupVersion

**Root cause:**
The auto-generated conversion function (Convert_v1beta1_AWSClusterControllerIdentity_To_v1beta2_AWSClusterControllerIdentity) copies TypeMeta directly from the source object, which includes the APIVersion field. This results in the converted v1beta2 object retaining the v1beta1 APIVersion, causing the conversion webhook to reject it as invalid.

**Fix:**
Explicitly set the APIVersion to the target version (v1beta2) after calling the auto-generated conversion function in both ConvertTo methods for AWSClusterControllerIdentity and AWSClusterControllerIdentityList. This ensures the converted object has the correct APIVersion that matches the target API version expected by Kubernetes.

This fix ensures that when Kubernetes requests conversion from v1beta1 to v1beta2, the webhook returns objects with the correct APIVersion, allowing the conversion to succeed.

**What type of PR is this?**

Fixes #

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Fixes a regression at AWSClusterControllerIdentity v1beta1 to v1beta2 conversion
```

### DIFF
--- a/test/e2e/suites/unmanaged/unmanaged_conversion_test.go
+++ b/test/e2e/suites/unmanaged/unmanaged_conversion_test.go
@@ -1,0 +1,208 @@
+//go:build e2e
+// +build e2e
+
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


damdo:
/close

in favour of #5739

PR #5447, opened 2025-03-26
🌱 Bump CAPI to v1.10.2

tamalsaha:
**What type of PR is this?**
I fixed the validator and defaulter interface broken by controller-runtime.
util/defaulting package is also internal to cluster-api now. I copied that. Not sure what is the right answer on that one.

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Bump CAPI to v1.10.2
```

PR #5744, opened 2025-11-14
[release-2.9] 🐛 fix: CAPA v1beta1 to v1beta2 generated conversion webhook

damdo:
Manual cherry-pick of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5739

---

/kind regression

**This PR addresses a conversion webhook failure introduced in CAPA v2.9.x**.

Users upgrading from CAPA v1.x (who have v1beta1 resources like AWSClusterControllerIdentity still stored in etcd) are unable to access these resources after upgrading to v2.9.x.

Any attempt to get these objects fails with an invalid groupVersion error, as the webhook incorrectly returns a v1beta1 object when v1beta2 is expected.

The root cause was identified in our old and custom copy of the upstream conversion-gen tooling.
A change (introduced in PR #5447 ) caused the generator to incorrectly add `out.TypeMeta = in.TypeMeta` to the generated conversion functions (`zz_generated.conversion.go`).

This line improperly copies the `apiVersion` from the `v1beta1` input object to the `v1beta2` output object, causing the webhook to return an invalid payload.

This pull request resolves the issue by removing the old custom copy of the upstream conversion-gen, in favour of the up-to-date/latest mainline conversion-gen.
It follow that up with a regeneration of the conversion code which removes the problematic TypeMeta assignment.

More details in the Slack thread: https://kubernetes.slack.com/archives/CD6U2V71N/p1762863174053729

```release-note
fix: CAPA v1beta1 to v1beta2 resources generated conversion webhook
```

### DIFF
--- a/api/v1beta1/zz_generated.conversion.go
+++ b/api/v1beta1/zz_generated.conversion.go
@@ -144,11 +144,6 @@ func RegisterConversions(s *runtime.Scheme) error {
 	}); err != nil {
 		return err
 	}
-	if err := s.AddGeneratedConversionFunc((*v1beta2.AWSClusterSpec)(nil), (*AWSClusterSpec)(nil), func(a, b interface{}, scope conversion.Scope) error {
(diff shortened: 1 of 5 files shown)


chrischdi (commented):
/lgtm

PR #5748, opened 2025-11-17
:bug: Default AWSMachine in admission controller instead of during reconciliation as it's not allowed to update the `awsmachine.spec` field after the initial creation

fiunchinho:
**What type of PR is this?**

/kind bug

While reconciling newly created `AWSMachines`, I noticed that this error usually happens

```
admission webhook \"validation.awsmachine.infrastructure.cluster.x-k8s.io\" denied the request: AWSMachine.infrastructure.cluster.x-k8s.io \"mycluster-somemachine\" is invalid: spec: Forbidden: cannot be modified" controller="awsmachine" controllerGroup="infrastructure.cluster.x-k8s.io" controllerKind="AWSMachine" AWSMachine="default/mycluster-somemachine" namespace="default" name="mycluster-mymachine" reconcileID="92584f4e-6f44-4f12-8007-352eab8a1429"
```

The problem is that we are setting the defaults on the existing `AWSMachine` while reconciling it. I think it'd be better to use the defaulting admission controller instead, and avoid default during the reconciliation.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Default AWSMachine in admission controller instead of during reconciliation as it's not allowed to update the `awsmachine.spec` field after the initial creation
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -213,8 +213,6 @@ func (r *AWSMachineReconciler) Reconcile(ctx context.Context, req ctrl.Request) 
 		return ctrl.Result{}, nil
 	}
 
-	infrav1.SetDefaults_AWSMachineSpec(&awsMachine.Spec)
-
 	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, awsMachine); err != nil || isPaused || conditionChanged {
 		return ctrl.Result{}, err
 	}
@@ -758,12 +756,6 @@ func (r *AWSMachineReconciler) reconcileOperationalState(ec2svc services.EC2Inte
 		return err
 	}
 	v1beta1conditions.MarkTrue(machineScope.AWSMachine, infrav1.SecurityGroupsReadyCondition)
-
-	err = r.ensureInstanceMetadataOptions(ec2svc, instance, machineScope.AWSMachine)
-	if err != nil {
-		machineScope.Error(err, "failed to ensure instance metadata options")
-		return err
-	}
 
 	return nil
 }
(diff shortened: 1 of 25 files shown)


controllers/awsmachine_controller.go
```diff
@@ -1320,6 +1318,12 @@ func (r *AWSMachineReconciler) ensureStorageTags(ec2svc services.EC2Interface, i
 }
 
 func (r *AWSMachineReconciler) ensureInstanceMetadataOptions(ec2svc services.EC2Interface, instance *infrav1.Instance, machine *infrav1.AWSMachine) error {
+	// If InstanceMetadataOptions is not set in the spec, don't modify the instance
+	// The webhook will set defaults on new resources, but we shouldn't force defaults on existing instances
```
AndiDog:
```suggestion
	// If InstanceMetadataOptions is not set in the spec, don't modify the AWSMachine.
	// The webhook will set defaults on new AWSMachines, but we cannot write defaults to existing objects
	// because validation doesn't allow changes to the spec.
```

controllers/awsmachine_controller_test.go
```diff
@@ -659,6 +665,7 @@ func mockedCreateInstanceCalls(m *mocks.MockEC2APIMockRecorder) {
 		Attribute:          ec2types.NetworkInterfaceAttributeGroupSet,
 	})).Return(&ec2.DescribeNetworkInterfaceAttributeOutput{Groups: []ec2types.GroupIdentifier{{GroupId: aws.String("3")}}}, nil).MaxTimes(1)
 	m.ModifyNetworkInterfaceAttribute(context.TODO(), gomock.Any()).AnyTimes()
+	m.ModifyInstanceMetadataOptions(context.TODO(), gomock.Any()).MaxTimes(1)
```
AndiDog: That's IMO not expected while creating an instance. The metadata should be set once in [RunInstances](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_RunInstances.html), and not modified afterwards.

fiunchinho: I think you're right. I gave it another shot, I completely removed `ModifyInstanceMetadataOptions` now, and rely only on the admission controller to change the metadata options.

AndiDog:
I think we should retitle the PR and release note to show what the change fixes.

For me, this looks fine but I'm not sure if it was intended that the full `.spec` field is immutable (as checked in `func (*awsMachineWebhook) ValidateUpdate`). I'll let some more experienced folks review and approve.

/lgtm

AndiDog:
@fiunchinho Let's rebase onto latest `main` since there have been some E2E flake fixes.

fiunchinho:
Test failed with
```
E1201 16:37:07.307961   48645 awscluster_controller.go:335] "failed to reconcile network" err="failed to create new managed VPC: failed to create vpc: The maximum number of VPCs has been reached"
```

PR #5751, opened 2025-11-18
🌱 Enable nofloats linter

### LINKED ISSUE #5490: [KAL] Enable the nofloats linter
Enable the nofloats linter.

moko-poi:
**What type of PR is this?**

/kind cleanup

This PR enables the `nofloats` linter as part of the KAL (Kube-API-Linter) configuration to enforce Kubernetes API best practices by preventing the use of float types in API definitions.

Float types are discouraged in Kubernetes APIs due to:
- Precision issues across different platforms
- Unreliable round-tripping in serialization/deserialization
- Difficulties in comparison operations

The internal rate limiter implementation (`pkg/internal/rate/rate.go`) legitimately uses float64 for token bucket calculations and has been exempted from this linter rule using `//nolint:kubeapilinter` comments, as it is not part of the Kubernetes API surface.

Fixes #5490

The `nofloats` linter only flags float usage in type definitions and struct fields, not in type conversions or temporary calculations. This is why `exp/api/v1beta2/validation.go` (which uses `float64()` conversions) doesn't trigger errors.

The existing `path-except: "api//*"` rule in `.golangci-kal.yml` doesn't effectively limit KAL linting to API folders only - KAL runs on all Go code similar to how `optionalorrequired` linter was previously enabled. Therefore, internal packages that legitimately use floats need explicit `//nolint` comments.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [ ] adds unit tests

**Release note**:
```release-note
NONE

### DIFF
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -23,7 +23,7 @@ linters:
               #- "jsontags" # Ensure every field has a json tag.
               #- "maxlength" # Ensure all strings and arrays have maximum lengths/maximum items.
               #- "nobools" # Bools do not evolve over time, should use enums instead.
-              #- "nofloats" # Ensure floats are not used.
+              - "nofloats" # Ensure floats are not used.
               #- "optionalorrequired" # Every field should be marked as `+optional` or `+required`.
               # - "requiredfields" # Required fields should not be pointers, and should not have `omitempty`.
               - "statussubresource" # All root objects that have a `status` field should have a status subresource.
--- a/pkg/internal/rate/rate.go
+++ b/pkg/internal/rate/rate.go
@@ -30,6 +30,8 @@ import (
 // Limit defines the maximum frequency of some events.
 // Limit is represented as number of events per second.
 // A zero Limit allows no events.
+//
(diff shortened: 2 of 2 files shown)


moko-poi:
The `nofloats` linter detected 4 violations in `pkg/internal/rate/rate.go` related to float usage in the token bucket rate limiter implementation. 
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/internal/rate/rate.go

Rather than refactoring the rate limiting logic (which would be out of scope for this PR), I've added `//nolint:kubeapilinter` comments to exempt this internal code.

This is justified because:
- The rate limiter is internal infrastructure, not part of the Kubernetes API surface
- The `nofloats` linter's purpose is to prevent floats in API definitions
- Modifying the rate limiter implementation would introduce unnecessary risk

The `nofloats` linter only flags type/field definitions, not conversions, which is why `exp/api/v1beta2/validation.go` doesn't trigger errors despite using `float64()`.

PR #5768, opened 2025-11-24
🐛 Fix invalid cloud-config when write_files is nil

joshfrench:
**What type of PR is this?**
/kind bug

When `EKSConfigTemplate.spec.template.spec.files` is empty, the resulting cloud config is invalid:
```
#cloud-config
write_files:
runcmd:
  - /etc/eks/bootstrap.sh cluster-name
```
```
cloud-init status --long
status: error
extended_status: error - done
boot_status_code: enabled-by-generator
last_update: Thu, 01 Jan 1970 00:25:50 +0000
detail: Failed due to systemd unit failure
errors:
        - ('write_files', TypeError("'NoneType' object is not iterable"))
```

This may only be an issue on newer versions of cloud-init (or maybe the underlying system Python?) We've been using CAPA for years and only saw this problem after updating our AMIs to Ubuntu 24.04:
```
cloud-init --version
/usr/bin/cloud-init 25.2-0ubuntu1~24.04.1
```

**Checklist**:
- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:
```release-note
Fix invalid cloud-config when EKS config has no files
```

### DIFF
--- a/bootstrap/eks/internal/userdata/node.go
+++ b/bootstrap/eks/internal/userdata/node.go
@@ -30,7 +30,9 @@ const (
 	defaultBootstrapCommand = "/etc/eks/bootstrap.sh"
 
 	nodeUserData = `#cloud-config
+{{- if .Files }}
 {{template "files" .Files}}
+{{- end }}
 runcmd:
 {{- template "commands" .PreBootstrapCommands }}
   - {{ .BootstrapCommand }} {{.ClusterName}} {{- template "args" . }}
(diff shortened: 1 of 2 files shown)


AndiDog:
/ok-to-test

Can you please retitle? The PR title sounds like an issue, but it should start with a verb, just like the proposed release note. At best, use present tense: "🐛 Fix ..."

PR #5808, opened 2026-01-05
✨ Eks pod identities

joshfrench:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Adds `PodIdentityAssociations` to the `AWSManagedControlPlane` spec. Pod identities allow mapping service accounts to IAM identities without OIDC or IRSA.

**Which issue(s) this PR fixes**:
Fixes #4747

**Checklist**:
- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:
```release-note
Adds EKS pod identity associations to AWSManagedControlPlane.
```

PR #5713, opened 2025-10-17
✨ feat: eks pod identity support for controllers

richardcase:
**What type of PR is this?**

/kind feature

**What this PR does / why we need it**:

This adds support for using EKS pod identity for the CAPA controller when the management cluster is an EKS cluster.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

This is in addition to #4906 .

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Add support for using EKS pod identity for the controller credentials.
```

PR #5809, opened 2026-01-06
✨ feat: Eks pod identity for controller role

joshfrench:
**What type of PR is this?**
/kind feature

Addresses feedback in the original PR #5713, which is a soft blocker to #5808.

Related to #4906

This mainly addresses the feedback in #5713. Most of the comments were stylistic/grammatical, but there was a question as to whether the `clusterawsadm controller zero-credentials --namespace=capa-system` step is required both for new clusters and when migrating existing clusters.

`clusterctl init` will always create the k8s secret, even if the AWS credentials were empty from the start. Therefore  even if `zero-credentials` is redundant, it should be safe to run for both new and existing clusters.

There's also one functional change: this only adds the `sts:TagSession` action to the trust policy when the principals include `pods.eks.amazonaws.com`. Pod identity associations require this, but it's superfluous otherwise.

**Checklist**:
- [ ] squashed commits
- [x] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:
```release-note
Add support for using EKS pod identity for the controller credentials.
```

### DIFF
--- a/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
+++ b/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go
@@ -51,8 +51,8 @@ func (t Template) controllersPolicyRoleAttachments() []string {
 	return attachments
 }
 
-func (t Template) controllersTrustPolicy() *iamv1.PolicyDocument {
-	policyDocument := ec2AssumeRolePolicy()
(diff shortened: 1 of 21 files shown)


joshfrench:
/remove-lifecycle stale

@richardcase Can we get this merged?

joshfrench:
boskos failure

/retest

PR #5210, opened 2024-11-07
✨ feat(addons): add updateAddonTagsWithLatestVersion to set addon tags to latest version

thiha-min-thant:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

- Feature enhancement: Adds support for specifying the latest version of EKS addons in AWSManagedControlPlane.
- Improves user experience: Automatically resolves the versioning of addons by fetching the latest supported versions.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #5184

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [ ] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
feat: Support installing the latest supported version of EKS addons by specifying version: latest in AWSManagedControlPlane.
```

PR #5811, opened 2026-01-06
✨ feat: Allow EKS addons to use latest tag

### LINKED ISSUE #5184: Support installing "latest" EKS addon
/kind feature

Currently when you specify EKS addons to be installed in `AWSManagedControlPlane` you need to specify the specific version of the addon. It would also be good to support installing the latest supported version of the addon.

So currently you need to do:

```yaml
  spec:
    addons:
    - conflictResolution: overwrite
      name: vpc-cni
      version: v1.18.1-eksbuild.3
```

But it would also be nice to support the following:

```yaml
  spec:
    addons:
    - conflictResolution: overwrite
      name: vpc-cni
      version: latest
```

If `latest` is specified then we need to query for the latest supported version for the cluster/k8s version (using `DescribeAddonVersions`) and use that.

**Environment:**

joshfrench:
**What type of PR is this?**
/kind feature

- Feature enhancement: Adds support for specifying the latest version of EKS addons in AWSManagedControlPlane.
- Improves user experience: Automatically resolves the versioning of addons by fetching the latest supported versions.

Fixes #5184

This PR is a continuation of #5210:
- Rebased to the latest `main`
- Uses the `semver` package to compare tags instead of simple lexicographic sort
- Adds basic tests

**Checklist**:
- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:
```release-note
Support installing the latest supported version of EKS addons by specifying `version: latest` in AWSManagedControlPlane.
```

### DIFF
--- a/pkg/cloud/services/eks/addons.go
+++ b/pkg/cloud/services/eks/addons.go
@@ -24,6 +24,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	"github.com/aws/aws-sdk-go-v2/service/eks"
 	ekstypes "github.com/aws/aws-sdk-go-v2/service/eks/types"
+	"golang.org/x/mod/semver"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta2"
@@ -272,22 +273,23 @@ func (s *Service) updateAddonTagsWithLatestVersion(ctx context.Context, addons [
 
 	latestVersions := make(map[string]string)
 	for _, addon := range output.Addons {
-		if len(addon.AddonVersions) > 0 {
-			latestVersions[*addon.AddonName] = *addon.AddonVersions[0].AddonVersion
-			for _, version := range addon.AddonVersions[1:] {
-				if *version.AddonVersion > latestVersions[*addon.AddonName] {
-					latestVersions[*addon.AddonName] = *version.AddonVersion
-				}
+		for _, version := range addon.AddonVersions {
+			current := latestVersions[*addon.AddonName]
+			if current == "" || semver.Compare(*version.AddonVersion, current) > 0 {
+				latestVersions[*addon.AddonName] = *version.AddonVersion
 			}
 		}
 	}
 
 	for _, addon := range addons {
-		if addon.Version != nil && *addon.Version == "latest" {
-			if latestVersion, ok := latestVersions[*addon.Name]; ok {
-				addon.Version = &latestVersion
-			}
+		if addon.Version == nil || *addon.Version != "latest" {
+			continue
 		}
+		latestVersion, ok := latestVersions[*addon.Name]
+		if !ok {
+			continue
+		}
+		addon.Version = &latestVersion
 	}
 
 	return addons, nil
--- a/pkg/cloud/services/eks/addons.go
+++ b/pkg/cloud/services/eks/addons.go
@@ -54,6 +54,11 @@ func (s *Service) reconcileAddons(ctx context.Context) error {
 	// Get the addons from the spec we want for the cluster
 	desiredAddons := s.translateAPIToAddon(s.scope.Addons())
 
+	// Update the tags with the latest version if available
+	if desiredAddons, err = s.updateAddonTagsWithLatestVersion(ctx, desiredAddons); err != nil {
+		return fmt.Errorf("updating addon tags with latest version: %w", err)
+	}
+
 	// If there are no addons desired or installed then do nothing
 	if len(installed) == 0 && len(desiredAddons) == 0 {
 		s.scope.Info("no addons installed and no addons to install, no action needed")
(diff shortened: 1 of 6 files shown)


pkg/cloud/services/eks/addons.go
```diff
@@ -248,3 +254,43 @@ func convertConfiguration(configuration string) *string {
...
 }
+
+func (s *Service) updateAddonTagsWithLatestVersion(ctx context.Context, addons []*eksaddons.EKSAddon) ([]*eksaddons.EKSAddon, error) {
+	s.Debug("Updating addon tags with the latest available version")
+
+	input := &eks.DescribeAddonVersionsInput{
+		KubernetesVersion: s.scope.ControlPlane.Spec.Version,
+	}
+
+	output, err := s.EKSClient.DescribeAddonVersions(ctx, input)
```
AndiDog, line 265: We must not call AWS APIs unconditionally, given that the feature seems gated by the `latest` value

pkg/cloud/services/eks/addons.go
```diff
@@ -248,3 +254,43 @@ func convertConfiguration(configuration string) *string {
 	}
 	return &configuration
 }
+
+func (s *Service) updateAddonTagsWithLatestVersion(ctx context.Context, addons []*eksaddons.EKSAddon) ([]*eksaddons.EKSAddon, error) {
```
AndiDog, line 258: Is the desired behavior to update the addons spec _once_ with the newest available version, write back the explicit version numbers, and then not reconcile again? Or should it continuously upgrade to the latest?

joshfrench: @AndiDog I picked up this issue because it will replace some functionality we built internally. For us, `latest` means "I do not care about the version of this addon" and it's fine to take a new version if one is available when we next reconcile a cluster.

I see the case for the alternate behavior you describe, but we'd need a way for the user to indicate, at a later point in time, that they want to move from the previously discovered version to the current value of `latest`.

Possible designs:
1. Add a field to the Addon spec, like `AutoUpdate (bool)` or `UpdatePolicy ("always"|"never")`.
1. Use Addon status or cluster annotation to record the resolved versions. Updating would require manually clearing the status field/annotation.
1. Change the implementation to a mutating webhook that does version discovery and patches the Addon spec directly. This would always happen on creation, and could happen conditionally on update (probably driven by a new field in the Addon spec, as in option 1.)

2 just feels bad, 3 would be annoying from a gitops perspective as my source of truth wouldn't match what was actually in k8s.

I'm leaning toward option 1 and have a mild preference for `UpdatePolicy ("always"|"never")` with a default of `never`. For consistency I'd expect the new field to be respected regardless of whether the version is a fixed tag or `latest`. But I suppose `always` and a fixed tag could be mutually exclusive instead.

Thoughts or preferences on which way to go with this?

joshfrench: `ifAvailable` might be better than `always` 🤔

joshfrench: This should also support `default`, as the latest version may not be marked as the default yet. The update policy should apply the same logic in either case.

AndiDog: To understand the current behavior: if I set `latest` (or `default`) in `spec.addons`, the code retrieves and reconciles the respective AWS version in each reconciliation, but doesn't write it back to `spec.addons`, right? If yes, that's the approach I'd like to have instead of extra fields. CAPA should not overwrite `latest` with `v1.2.3`, for example, since that sort of behavior is an absolute mess in some existing reconciliation code. So IMO, neither 1 nor 2 nor 3.

joshfrench: @AndiDog Agreed, CAPA should not mutate the spec if a new addon version is discovered during reconciliation. I do think we need an additional field to control when/if version discovery is attempted though. It offers the same kind of control as a container's `imagePullPolicy` when using a moving target like the `latest` image tag.

Users might want either of these behaviors, which seem equally reasonable to me:
1. Get the latest or default addon version at time of installation and then freeze to that version.
2. Take updates during future reconciliations as they become available.

I'm proposing these changes to the spec:
```diff
diff --git i/controlplane/eks/api/v1beta2/types.go w/controlplane/eks/api/v1beta2/types.go
index ab395d225..2bd74aac0 100644
--- i/controlplane/eks/api/v1beta2/types.go
+++ w/controlplane/eks/api/v1beta2/types.go
@@ -183,8 +183,21 @@ type Addon struct {
 	// +kubebuilder:validation:MinLength:=2
 	// +kubebuilder:validation:Required
 	Name string `json:"name"`
-	// Version is the version of the addon to use
+	// Version is the version of the addon to use.
+	// Use `default` to use the current default version for the
+	// cluster's Kubernetes version.
+	// Use `latest` to use the latest version for the cluster's
+	// Kubernetes version (which may be newer than the default.)
 	Version string `json:"version"`
+	// UpdatePolicy determines the behavior on subsequent reconciles when
+	// the addon version is `default` or `latest`.
+	// `ifAvailable` will update the addon to the newest default or latest
+	// version, if it has changed since the last reconcile.
+	// `never` will not update the addon version after initial
+	// installation, even if the default or latest version changes.
+	// +kubebuilder:validation:Enum=ifAvailable;never
+	// +kubebuilder:default=never
+	UpdatePolicy string `json:"updatePolicy,omitempty"`
 	// Configuration of the EKS addon
 	// +optional
 	Configuration string `json:"configuration,omitempty"`
```

I think this also implies a new validation to ensure `UpdatePolicy` cannot be `ifAvailable` unless `Version` is `default` or `latest`.

AndiDog: I think particularly the `default` value makes sense to use for everyone, to avoid hardcoding specific versions in templates. The `latest` value might be for a small number of users. For both, we'd want something like `updatePolicy` indeed. A good idea, in my opinion. We need to define what `updatePolicy=never` does on Kubernetes version upgrade – pretty sure users don't want to be stuck with old addon versions?!

Maybe let's hear what other maintainers think of this feature – @dlipovetsky @richardcase maybe? If we agree, then this spec change could be implemented. The PR anyway needs a rebase.

joshfrench: 1. If `latest` is of limited value, I'm fine dropping it from this proposal. We can always add it later.

2. I hadn't considered `updatePolicy=never` in the context of a k8s version upgrade. I'll have to double check this but if memory serves, when updating the EKS control plane would leave an addon at an unsupported version AWS doesn't automatically pull the addon forward. The addon remains where it is and just gets marked as "Degraded."
Given that, I'm not sure this controller should attempt to reconcile addons either in that scenario. There may be a range of overlapping versions that are compatible with both the old and new k8s versions; should it pick the lowest common version? Highest?
Maybe we just check addon health post-upgrade and emit a warning if any of the addons are in a degraded state after completion. The user can decide if they want to leave it as-in, switch to `updatePolicy=ifAvailable`, pin to a known compatible version, etc.

joshfrench:
this is just a pet feature, i'll get back to it eventually.

/remove-lifecycle stale

PR #5815, opened 2026-01-13
✨ Expose new field to add custom configuration to the S3 bucket policy

fiunchinho:
**What type of PR is this?**

/kind feature

CAPA currently exposes fields to configure which IAM instance profiles / roles will be allowed to connect to the S3 bucket.  These fields allow users to pass the name of instance profiles for the control plane and the machine pools. CAPA will allow those instance profiles into the right S3 bucket path, depending on whether they are for the control plane or for the machine pools. 

With this PR, we are exposing a new field so users can add more instance profiles to the S3 bucket policy. Our use case is that we are creating node pools that are not managed by CAPA (with karpenter). The nodes for these node pools need access to the cluster S3 bucket to get the instance bootstrap data.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Expose new field to add custom configuration to the S3 bucket policy
```


api/v1beta2/awscluster_types.go
```diff
@@ -283,6 +283,27 @@ type AWSClusterStatus struct {
 	Conditions     clusterv1beta1.Conditions     `json:"conditions,omitempty"`
 }
 
+// AdditionalIAMInstanceProfile defines an additional IAM instance profile
+// with a custom S3 prefix for accessing bootstrap data from S3 Bucket.
+// This enables support for custom node pools (e.g., Karpenter) that need
+// access to bootstrap data with custom S3 prefixes.
+type AdditionalIAMInstanceProfile struct {
+	// Name is the name of the IAM instance profile that will be granted access.
+	// This should match the IAM role name (not the ARN).
```
AndiDog:
```suggestion
	// This must match the IAM role name (not the ARN).
```

and let's use `+kubebuilder:validation:Pattern` to avoid users being able to fill an ARN (any regex that excludes `:` should do that; I guess role names follow some pattern).

api/v1beta2/awscluster_types.go
```diff
@@ -283,6 +283,27 @@ type AWSClusterStatus struct {
...
+	// +kubebuilder:validation:MinLength=1
+	// +kubebuilder:validation:MaxLength=128
+	Name string `json:"name"`
+
+	// Prefix is the S3 object prefix (path) that this profile can access.
+	// Must end with /* to allow wildcard access (e.g., "karpenter-nodes/*", "path/to/data/*").
+	// Can also be "*" for wildcard-only access to the entire bucket.
```
AndiDog: Do we really need to restrict this? Any `additionalXyz` fields are usually for the purpose of full customization.

api/v1beta2/awscluster_webhook_test.go
```diff
@@ -323,6 +323,151 @@ func TestAWSClusterValidateCreate(t *testing.T) {
...
+						NodesIAMInstanceProfiles:       []string{"bar"},
+						AdditionalIAMInstanceProfiles: []AdditionalIAMInstanceProfile{
+							{Name: "duplicate", Prefix: "path1/*"},
+							{Name: "duplicate", Prefix: "path2/*"},
+						},
+					},
+				},
+			},
+			wantErr: true,
+		},
```
AndiDog: A test case "rejects ARN instead of IAM instance profile name" would be great (see comment above)

api/v1beta2/s3bucket.go
```diff
@@ -62,6 +63,44 @@ func (b *S3Bucket) Validate() []*field.Error {
...
+		}
+
+		// Must not start with /
+		if profile.Prefix != "" && strings.HasPrefix(profile.Prefix, "/") {
+			errs = append(errs, field.Invalid(profilePath.Child("prefix"), profile.Prefix,
+				"must not start with /"))
+		}
+	}
+
+	// Check for duplicate names
```
AndiDog: Is this really a problem we need to check for? Would duplicates fail at AWS?

If it's just about this:

```
Sid:    fmt.Sprintf("additional-%s", additionalProfile.Name),
```

Then we can use the index instead of the name and avoid this validating code.

AndiDog: I'm addressing this as suggested above

api/v1beta2/awscluster_types.go
AndiDog: I'm addressing this as suggested above

api/v1beta2/awscluster_types.go
AndiDog: I'm addressing this as suggested above

nrb (approved):
/lgtm

dlipovetsky:
Forgive me if I've misunderstood something here, but I think S3 bucket access can be given to a specific IAM _role_, and that role can be acquired by making a request from an instance that has an  IAM _instance profile_ with that role.

Is karpenter (or whatever needs access) using the role directly, or indirectly via the instance profile? Does the field `AdditionalIAMRoles` make more sense than `AdditionalIAMInstanceProfiles`? I'd like to understand this better before approving.

The code changes lgtm.

AndiDog:
> Forgive me if I've misunderstood something here, but I think S3 bucket access can be given to a specific IAM _role_, and that role can be acquired by making a request from an instance that has an IAM _instance profile_ with that role.

@dlipovetsky You're right. I've renamed "instance profile" to "role" everywhere. Also, I've added impl+test for allowing multiple S3 object prefixes for the same role name instead of rejecting it (see `+listMapKey`). Kept it as separate commits for easier review. If you're fine with it and `/approve`, I can squash and merge this myself with `/lgtm` since I'm not the original author (I promise to not make unreviewed changes 😆).

PR #5817, opened 2026-01-14
boskos.py: add extra logging

damdo:
**What type of PR is this?**

/kind support

Add extra logging to boskos.py to improve failures visibility

**Release note**:
```release-note
NONE
```

### DIFF
--- a/hack/boskos.py
+++ b/hack/boskos.py
@@ -16,10 +16,18 @@
 
 import argparse
 import json
+import logging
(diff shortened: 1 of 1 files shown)


nrb:
This was for debugging some CI issues that were actually with the boskos checkouts and this extra logging didn't add much.

We're also planning to move to a Go utility and remove our Python script.

/close

PR #5818, opened 2026-01-14
🐛  Validate region when reconciling cluster

fiunchinho:
**What type of PR is this?**

/kind bug

When reconciling clusters using `AWSClusterRoleIdentity` with an **invalid or empty region name**, the wrong region credentials could be cached and reused for **ALL clusters** using the same identity, causing subsequent clusters to fail or use the wrong region.

The codebase had three critical issues:

1. **No validation of region names** - Any string was accepted as a region
2. **Global mutable caches** - Provider and session objects were cached globally using sync.Map
3. **Cache keys included region** - But invalid regions weren't rejected, so they poisoned the cache
4. **Subsequent clusters got poisoned sessions** - Next cluster with the same identity would get the wrong cached session

We implemented two layers of validation:

- Reject invalid regions at admission time before they reach the controller.
- Add defensive checks in session creation to prevent cache poisoning if an invalid region somehow gets through.

Fixes #

The high number of files changed in the PR are due to making the `spec.region` field to be required. It was optional before. So tests that did not specify any region or that used an invalid region name needed to be updated.

When a cluster with an empty region was created before, here's the sequence of events:

  1. No Webhook Validation: The resource was accepted without any validation
  2. Controller Reconciliation: At pkg/cloud/scope/cluster.go:79, the code would call:
  session, serviceLimiters, err := sessionForClusterWithRegion(
      params.Client,
      clusterScope,
      params.AWSCluster.Spec.Region,  // This would be ""
      params.Logger
  )
  3. Session Creation: The empty region "" was passed to AWS SDK:
  config.LoadDefaultConfig(context.Background(), config.WithRegion(""))
  4. AWS SDK Behavior:
    - When config.WithRegion("") is used, the SDK ignores that parameter
    - The SDK then falls back to checking:
        - AWS_REGION environment variable
      - AWS_DEFAULT_REGION environment variable
      - AWS config files
      - EC2 instance metadata (if running on EC2)
  5. Controller Environment: Looking at config/manager/manager.yaml, no AWS region environment variables are set in the controller pod

  Outcome: the AWS SDK would return 

```
  operation error <SERVICE>: <OPERATION>, failed to resolve service endpoint,
  an AWS region is required, but was not found
```

Taking all of this into consideration, I believe it makes sense to make the field required, as it was actually required in the first place.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
Validate region when reconciling cluster
```

### DIFF
--- a/api/v1beta2/awscluster_webhook_test.go
+++ b/api/v1beta2/awscluster_webhook_test.go
@@ -37,7 +37,12 @@ import (
 )
 
 func TestAWSClusterDefault(t *testing.T) {
-	cluster := &AWSCluster{ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"}}
+	cluster := &AWSCluster{
+		ObjectMeta: metav1.ObjectMeta{Name: "foo", Namespace: "default"},
+		Spec: AWSClusterSpec{
+			Region: "us-west-2",
+		},
+	}
 	t.Run("for AWSCluster", defaultValidateTest(context.Background(), cluster, &awsClusterWebhook{}, true))
 	cluster.Default()
 	g := NewWithT(t)
@@ -61,6 +66,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						LoadBalancerType: LoadBalancerTypeDisabled,
 						Name:             ptr.To("name"),
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -73,6 +79,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						CrossZoneLoadBalancing: true,
 						LoadBalancerType:       LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -85,6 +92,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						Subnets:          []string{"foo", "bar"},
 						LoadBalancerType: LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -97,6 +105,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						HealthCheckProtocol: &ELBProtocolTCP,
 						LoadBalancerType:    LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -109,6 +118,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						AdditionalSecurityGroups: []string{"foo", "bar"},
 						LoadBalancerType:         LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -126,6 +136,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						},
 						LoadBalancerType: LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -145,6 +156,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						},
 						LoadBalancerType: LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -157,6 +169,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						DisableHostsRewrite: true,
 						LoadBalancerType:    LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -169,6 +182,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						PreserveClientIP: true,
 						LoadBalancerType: LoadBalancerTypeDisabled,
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -179,6 +193,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
 					ControlPlaneLoadBalancer: &AWSLoadBalancerSpec{Scheme: &unsupportedIncorrectScheme},
+					Region:                   "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -193,6 +208,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 						strings.Repeat("CAPI", 33): "value-3",
 						"key-4":                    strings.Repeat("CAPI", 65),
 					},
+					Region: "us-west-2",
 				},
 			},
 			wantErr: true,
@@ -201,6 +217,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "accepts bucket name with acceptable characters",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region: "us-west-2",
 					S3Bucket: &S3Bucket{
 						Name:                           "abcdefghijklmnoprstuwxyz-0123456789",
 						ControlPlaneIAMInstanceProfile: "control-plane.cluster-api-provider-aws.sigs.k8s.io",
@@ -213,6 +230,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "rejects empty bucket name",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region:   "us-west-2",
 					S3Bucket: &S3Bucket{},
 				},
 			},
@@ -222,6 +240,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "rejects bucket name shorter than 3 characters",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region: "us-west-2",
 					S3Bucket: &S3Bucket{
 						Name: "fo",
 					},
@@ -233,6 +252,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "rejects bucket name longer than 63 characters",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region: "us-west-2",
 					S3Bucket: &S3Bucket{
 						Name: strings.Repeat("a", 64),
 					},
@@ -244,6 +264,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "rejects bucket name starting with not letter or number",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region: "us-west-2",
 					S3Bucket: &S3Bucket{
 						Name: "-foo",
 					},
@@ -255,6 +276,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "rejects bucket name ending with not letter or number",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region: "us-west-2",
 					S3Bucket: &S3Bucket{
 						Name: "foo-",
 					},
@@ -266,6 +288,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "rejects bucket name formatted as IP address",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region: "us-west-2",
 					S3Bucket: &S3Bucket{
 						Name: "8.8.8.8",
 					},
@@ -277,6 +300,7 @@ func TestAWSClusterValidateCreate(t *testing.T) {
 			name: "requires bucket control plane IAM instance profile to be not empty",
 			cluster: &AWSCluster{
 				Spec: AWSClusterSpec{
+					Region: "us-west-2",
 					S3Bucket: &S3Bucket{
 						Name:                           "foo",
 						ControlPlaneIAMInstanceProfile: "",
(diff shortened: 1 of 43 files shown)


api/v1beta2/awscluster_webhook_test.go
```diff
@@ -1490,6 +1629,197 @@ func TestAWSClusterDefaultAllowedCIDRBlocks(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.TODO()
+			cluster := tt.cluster.DeepCopy()
+			cluster.ObjectMeta = metav1.ObjectMeta{
+				GenerateName: "cluster-",
+				Namespace:    "default",
+			}
+			if cluster.Spec.Region == "" && !tt.wantErr {
+				cluster.Spec.Region = "us-west-2"
+			}
```
AndiDog, line 1740: Not needed for the above test cases if I see it correctly?

api/v1beta2/awscluster_webhook_test.go
```diff
@@ -1490,6 +1629,197 @@ func TestAWSClusterDefaultAllowedCIDRBlocks(t *testing.T) {
...
+		t.Run(tt.name, func(t *testing.T) {
+			ctx := context.TODO()
+			cluster := tt.oldCluster.DeepCopy()
+			cluster.ObjectMeta = metav1.ObjectMeta{
+				GenerateName: "cluster-",
+				Namespace:    "default",
+			}
+			if cluster.Spec.Region == "" {
+				cluster.Spec.Region = "us-west-2"
+			}
```
AndiDog, line 1809: Doesn't apply to above test cases and is surprising. Better remove?

api/v1beta2/awscluster_webhook_test.go
```diff
@@ -1490,6 +1629,197 @@ func TestAWSClusterDefaultAllowedCIDRBlocks(t *testing.T) {
...
+
+			if err := testEnv.Create(ctx, cluster); err != nil {
+				t.Errorf("failed to create cluster: %v", err)
+			}
+
+			cluster.Spec.Region = tt.newCluster.Spec.Region
```
AndiDog, line 1815: If all we do is test the `region` field, the test cases don't need to define the full `AWSCluster` object but could have `oldRegion, newRegion string` fields.

controlplane/eks/api/v1beta2/awsmanagedcontrolplane_webhook.go
```diff
@@ -566,6 +576,20 @@ func validatePrivateDNSHostnameTypeOnLaunch(networkSpec infrav1.NetworkSpec, pat
 	return allErrs
 }
 
+func (r *AWSManagedControlPlane) validateRegion() *field.Error {
+	if r.Spec.Region == "" {
+		return field.Required(field.NewPath("spec", "region"), "region is required")
+	}
+
+	// Validate region against all AWS partitions
```
AndiDog, line 584: I'm not really sure what this comment means. I think the call to `IsValidRegion` needs no explanation, even if this is somehow related to partitions?!

api/v1beta2/awscluster_webhook_test.go
```diff
@@ -1490,6 +1629,197 @@ func TestAWSClusterDefaultAllowedCIDRBlocks(t *testing.T) {
...
+		{
+			name: "invalid region - random string",
+			cluster: &AWSCluster{
+				Spec: AWSClusterSpec{
+					Region: "notaregion",
+				},
+			},
+			wantErr: true,
+		},
+	}
```
AndiDog, line 1728: Let's add the live example that brought us here, something like `eu-central-1a` (extraneous AZ specifier nto allowed)

pkg/cloud/services/ec2/bastion_test.go
```diff
@@ -224,7 +224,7 @@ func TestServiceDeleteBastion(t *testing.T) {
 				awsCluster := &infrav1.AWSCluster{
 					ObjectMeta: metav1.ObjectMeta{Name: "test"},
 					Spec: infrav1.AWSClusterSpec{
-						NetworkSpec: infrav1.NetworkSpec{
+						Region: "us-west-2", NetworkSpec: infrav1.NetworkSpec{
```
AndiDog, line 227:
```suggestion
						Region: "us-west-2",
						NetworkSpec: infrav1.NetworkSpec{
```

pkg/cloud/services/ec2/bastion_test.go
```diff
@@ -446,7 +446,7 @@ func TestServiceReconcileBastion(t *testing.T) {
...
```
AndiDog, line 449:
```suggestion
						Region: "us-west-2",
						NetworkSpec: infrav1.NetworkSpec{
```

pkg/cloud/services/elb/loadbalancer_test.go
```diff
@@ -89,7 +92,7 @@ func TestELBName(t *testing.T) {
 					Namespace: metav1.NamespaceDefault,
 				},
 				Spec: infrav1.AWSClusterSpec{
-					ControlPlaneLoadBalancer: &infrav1.AWSLoadBalancerSpec{
+					Region: "us-west-2", ControlPlaneLoadBalancer: &infrav1.AWSLoadBalancerSpec{
```
AndiDog, line 95:
```suggestion
					Region: "us-west-2",
					ControlPlaneLoadBalancer: &infrav1.AWSLoadBalancerSpec{
```

[repeated]

pkg/cloud/services/network/subnets_test.go
```diff
@@ -4383,7 +4383,7 @@ func TestDiscoverSubnets(t *testing.T) {
 				AWSCluster: &infrav1.AWSCluster{
 					ObjectMeta: metav1.ObjectMeta{Name: "test"},
 					Spec: infrav1.AWSClusterSpec{
-						NetworkSpec: *tc.input,
+						Region: "us-west-2", NetworkSpec: *tc.input,
```
AndiDog, line 4386:
```suggestion
						Region: "us-west-2",
						NetworkSpec: *tc.input,
```

[repeated]

pkg/cloud/services/s3/s3_test.go
```diff
@@ -137,7 +137,7 @@ func TestReconcileBucket(t *testing.T) {
 			},
 			AWSCluster: &infrav1.AWSCluster{
 				Spec: infrav1.AWSClusterSpec{
-					S3Bucket: &infrav1.S3Bucket{},
+					Region: "us-west-2", S3Bucket: &infrav1.S3Bucket{},
```
AndiDog, line 140:
```suggestion
					Region: "us-west-2",
					S3Bucket: &infrav1.S3Bucket{},
```

pkg/cloud/services/securitygroup/securitygroups_test.go
```diff
@@ -1136,7 +1136,7 @@ func TestReconcileSecurityGroups(t *testing.T) {
 			cluster := &infrav1.AWSCluster{
 				ObjectMeta: metav1.ObjectMeta{Name: "test"},
 				Spec: infrav1.AWSClusterSpec{
-					NetworkSpec: *tc.input,
+					Region: "us-west-2", NetworkSpec: *tc.input,
```
AndiDog, line 1139:
```suggestion
					Region: "us-west-2",
					NetworkSpec: *tc.input,
```

[repeated]

pkg/cloud/scope/rosaroleconfig.go
```diff
@@ -70,7 +70,9 @@ func NewRosaRoleConfigScope(params RosaRoleConfigScopeParams) (*RosaRoleConfigSc
 		RosaRoleConfig: params.RosaRoleConfig,
 	}
 
-	session, serviceLimiters, err := sessionForClusterWithRegion(params.Client, RosaRoleConfigScope, "", params.Logger)
+	// IAM is a global service, but AWS SDK still requires a region.
+	// Use us-east-1 as the default region for IAM operations.
+	session, serviceLimiters, err := sessionForClusterWithRegion(params.Client, RosaRoleConfigScope, "us-east-1", params.Logger)
```
AndiDog, line 75: This change might be surprising given the PR title. If it's really needed, please add it to the title and release notes.

AndiDog (commented):
I'm wondering why all `AWSClusterSpec` fields are "optional". This change makes the field required, which it actually was semantically, I think: for both managed and unmanaged VPC, the region is required to create or find it. I can't think of a case where the field can be empty.

LGTM with some minor code formatting requests.

damdo:
⚠️  Hey @fiunchinho I worry this might effectively be a borderline breaking change!

This PR introduces webhook validation for the Region field on AWSCluster and AWSManagedControlPlane resources. While this improves the user experience by catching invalid configurations at admission time rather than at runtime, it constitutes a breaking change for existing clusters.

Looking at the API definition for AWSCluster (for example) here:

https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/690d73b87de1fff68a14c9c03fd67c5258166962/api/v1beta2/awscluster_types.go#L39-L40

- No +optional marker exists (suggesting it was intended to be required)
- No +kubebuilder:validation:Required marker exists either
- The generated CRD has no `required:` constraint for Region at the spec level
- Hence OpenAPI validation does NOT enforce non-empty Region

So "API wise" the Region was, and is, optional.

Also before this change this was optional webhook wise too.

Which means that the empty Region field would reach runtime.

So this is what might have happened previously with an empty Region (with Claude's help, although not verified):

> - Session creation would succeed — AWS SDK's config.LoadDefaultConfig accepts empty region and falls back to environment variables (AWS_REGION, AWS_DEFAULT_REGION) or ~/.aws/config
> - If a default region existed (env var, config file), the cluster would silently use that region potentially unintentionally
> - If no default region was configured anywhere, API calls would fail later with confusing errors like MissingRegion: could not find region configuration
> 

This has a technically breaking impact on existing clusters with empty Region set.

> - Existing AWSCluster or AWSManagedControlPlane resources without Region set will fail webhook validation on any update operation
> - Users who relied on the implicit fallback to AWS_REGION environment variable will need to explicitly set Region in their cluster specs
> - New resources must specify a valid Region


IMO this validation change should be introduced as part of an API version bump (e.g., v1beta3) rather than in the existing v1beta2 API

/cc @nrb @richardcase @JoelSpeed

damdo:
/hold

For discussion on Breaking change

JoelSpeed:
You cannot make this field required without breaking existing users. Any resource that exists today that does not set the field would be unwritable unless you implement ratcheting in the webhook so that for new objects the field is required, but for existing objects that do not meet the validation, the existing value is allowed to be persist.

You could make it required in a future version, but to do that, you'd need a way to populate the field when it's empty on up-conversion, is there somewhere else we can infer the region from when it is otherwise unset?

AndiDog:
The field seems semantically required: any cluster needs a VPC if I'm not mistaken, and the VPC must either be created or looked up, and both these operations require the AWS API region. Therefore I don't think it _can_ be a breaking change for existing, working clusters. What about we check what happens in CAPA code if the field _isn't_ filled (to confirm this reasoning)?

JoelSpeed:
> The field seems semantically required: any cluster needs a VPC if I'm not mistaken, and the VPC must either be created or looked up, and both these operations require the AWS API region. Therefore I don't think it can be a breaking change for existing, working clusters. What about we check what happens in CAPA code if the field isn't filled (to confirm this reasoning)?

Has anyone tested creating a cluster and leaving this field blank? Does it fallback to a default (us-east-1) meaning that the cluster boots and is actually fine? Or does it default to something that could vary per some other configuration a user has provided?

If there's a fixed default that's applied by the AWS SDK, then we can likely up-convert and populate that value, but if it varies at all, we would have no idea how to correctly populate that field (this is supposing that an empty value still results in a cluster coming up correctly)

richardcase:
Thanks for this @fiunchinho . I agree with @damdo and @JoelSpeed changing this from optional to required will potentially break some of our users.

I'd favour including this in our next API version (v1beta3). We have a tracking issue for this #3853

nrb:
Given that the AWS credential chain falls back to environment variables and config files, it seems possible that this has been masked because, as mentioned in the initial issue, the AWS SDK is silently succeeding despite the missing value. A region is always required, but it is sometimes inferred from the environment rather than the Kube API.

@fiunchinho I assume this is something you could reproduce reliably?

I'm wondering if we can show that the environment is being used, maybe we can leverage that to do ratcheting validation? Something like 

```go
if spec.Region == "" && (env["AWS_REGION"] != "" || env["AWS_DEFAULT_REGION"] != "") {
  spec.Region = env["AWS_REGION"] // would need to check which one's actually used, of course.
}
```

To me, detecting the environment and persisting it in the Kube API is good for forward-compatible changes, but I also recognize that could be a change in behavior for anyone relying on the environment to be the authoritative source for the region.

pkg/cloud/scope/rosaroleconfig.go
serngawy: I'm sure about that, Does it raise error if we didn't set the region ?

AndiDog:
Upon creation of `AWSCluster` without setting the `region` field, we get this error:

```
    Message:               failed to create vpc endpoint for service "com.amazonaws..s3": operation error EC2: CreateVpcEndpoint, https response error StatusCode: 400, RequestID: 3e29b182-634e-4563-94ac-252108848349, api error InvalidServiceName: The Vpc Endpoint Service 'com.amazonaws..s3' does not exist
    Reason:                VpcEndpointsReconciliationFailed
```

And the field doesn't get filled automatically. So at current versions of CAPA, it's basically required already.

damdo:
@fiunchinho could you rebase when you have a moment? 



> Upon creation of `AWSCluster` without setting the `region` field, we get this error:
> 
> ```
>     Message:               failed to create vpc endpoint for service "com.amazonaws..s3": operation error EC2: CreateVpcEndpoint, https response error StatusCode: 400, RequestID: 3e29b182-634e-4563-94ac-252108848349, api error InvalidServiceName: The Vpc Endpoint Service 'com.amazonaws..s3' does not exist
>     Reason:                VpcEndpointsReconciliationFailed
> ```
> 
> And the field doesn't get filled automatically. So at current versions of CAPA, it's basically required already.

I feel like in light of this, we can add the validation, as it would otherwise error anyway. Thanks!

PR #5825, opened 2026-01-15
:bug: Fix: Changed dedicated host validation logic to require tenancy=host

vr4manta:
**What type of PR is this?**

/kind bug

This PR enhances the dedicated host logic to verify that the tenancy is `host` if any of the dedicated host fields (hostAffinity, hostID, dynamic hosts) are set.

Fixes #

In testing this feature in OCP, we came across configuration combinations that were working with AWS with CAPA but not as expected with CLI.  If we test this directly with AWS cli, and setting only `Affinity=default`, we see it is not allowed:

```
prompt > AWS_PROFILE=saml aws ec2 run-instances --region us-east-2 \
    --image-id ami-0bc8dda494f111572 --count 1 \
    --instance-type m6i.xlarge --key-name huliu-testdns \
    --security-group-ids <value> \
    --subnet-id <value> \
    --placement Affinity=default

An error occurred (InvalidParameterCombination) when calling the RunInstances operation: The parameter affinity cannot be used without specifying a tenancy of 'host'
```
This confirms that Affinity requires Tenancy=host to be set as well. We should align with AWS's behavior, shouldn't we?

**Checklist**:

- [X ] squashed commits
- [ ] includes documentation
- [X ] includes emoji in title 
- [X ] adds unit tests

**Release note**:

```release-note
Enhanced dedicated host logic to require tenancy=host when attempting to use dedicated hosts for instances.
```

### DIFF
--- a/api/v1beta2/awsmachine_webhook.go
+++ b/api/v1beta2/awsmachine_webhook.go
@@ -38,6 +38,11 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/v2/feature"
 )
 
+const (
+	hostTenancy  = "host"
+	hostAffinity = "host"
+)
+
 // log is for logging in this package.
 var log = ctrl.Log.WithName("awsmachine-resource")
 
@@ -484,6 +489,19 @@ func (r *AWSMachine) validateHostAllocation() field.ErrorList {
 	// If both hostID and dynamicHostAllocation are specified, return an error
 	if hasHostID && hasDynamicHostAllocation {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.hostID"), "hostID and dynamicHostAllocation are mutually exclusive"), field.Forbidden(field.NewPath("spec.dynamicHostAllocation"), "hostID and dynamicHostAllocation are mutually exclusive"))
+	}
+
+	// HostID, HostAffinity, and DynamicHostAllocation can only be set when Tenancy is "host"
+	if hasHostID && r.Spec.Tenancy != hostTenancy {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.hostID"), "hostID can only be set when tenancy is 'host'"))
+	}
+
+	if r.Spec.HostAffinity != nil && *r.Spec.HostAffinity == hostAffinity && r.Spec.Tenancy != hostTenancy {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.hostAffinity"), "hostAffinity can only be set to 'host' when tenancy is 'host'"))
+	}
+
+	if hasDynamicHostAllocation && r.Spec.Tenancy != hostTenancy {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.dynamicHostAllocation"), "dynamicHostAllocation can only be set when tenancy is 'host'"))
 	}
 
 	return allErrs
--- a/api/v1beta2/awsmachinetemplate_webhook_test.go
+++ b/api/v1beta2/awsmachinetemplate_webhook_test.go
@@ -88,6 +88,7 @@ func TestAWSMachineTemplateValidateCreate(t *testing.T) {
 					Template: AWSMachineTemplateResource{
 						Spec: AWSMachineSpec{
 							InstanceType: "test",
+							Tenancy:      "host",
 							HostID:       aws.String("h-1234567890abcdef0"),
 							DynamicHostAllocation: &DynamicHostAllocationSpec{
 								Tags: map[string]string{
@@ -108,6 +109,7 @@ func TestAWSMachineTemplateValidateCreate(t *testing.T) {
 					Template: AWSMachineTemplateResource{
 						Spec: AWSMachineSpec{
 							InstanceType: "test",
+							Tenancy:      "host",
 							HostAffinity: ptr.To("host"),
 						},
 					},
@@ -123,6 +125,7 @@ func TestAWSMachineTemplateValidateCreate(t *testing.T) {
 					Template: AWSMachineTemplateResource{
 						Spec: AWSMachineSpec{
 							InstanceType: "test",
+							Tenancy:      "host",
 							HostAffinity: ptr.To("host"),
 							HostID:       ptr.To("h-09dcf61cb388b0149"),
 						},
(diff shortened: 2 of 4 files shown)


api/v1beta2/awsmachinetemplate_webhook_test.go
```diff
@@ -164,6 +168,57 @@ func TestAWSMachineTemplateValidateCreate(t *testing.T) {
 			},
 			wantError: false,
 		},
+		{
```
nrb, line 171: You'd mentioned something about 8 character host IDs also being valid - do we want to include a test case for that here too, or is that not relevant within CAPA since we're mostly passing values through?

vr4manta: In one of the earlier PRs, Joel Speed had mentioned we could add formatting validation.  I think if we want to address that, we can do that via a different PR since this one is fixing an issue.  My opinion is that the older 8 character hosts should really no longer exist after the transition from 2018.  If someone is holding onto an older one that hasn't been migrated to new id format, what are the odds that it may not meet the requirements for running k8s clusters.  For the unit tests, I think we can ignore adding tests for the format until we actually start validating for them.

nrb: Sounds good, we can omit it until/unless we see it cause an actual problem.

api/v1beta2/awsmachine_webhook.go
```diff
@@ -486,6 +491,19 @@ func (r *AWSMachine) validateHostAllocation() field.ErrorList {
 		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.hostID"), "hostID and dynamicHostAllocation are mutually exclusive"), field.Forbidden(field.NewPath("spec.dynamicHostAllocation"), "hostID and dynamicHostAllocation are mutually exclusive"))
 	}
 
+	// HostID, HostAffinity, and DynamicHostAllocation can only be set when Tenancy is "host"
```
huali9, line 494: I'm curious what's the MAPI equivalent field for DynamicHostAllocation?

api/v1beta2/awsmachine_webhook.go
```diff
@@ -486,6 +491,19 @@ func (r *AWSMachine) validateHostAllocation() field.ErrorList {
...
+	}
+
+	if r.Spec.HostAffinity != nil && *r.Spec.HostAffinity == hostAffinity && r.Spec.Tenancy != hostTenancy {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.hostAffinity"), "hostAffinity can only be set to 'host' when tenancy is 'host'"))
+	}
+
+	if hasDynamicHostAllocation && r.Spec.Tenancy != hostTenancy {
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec.dynamicHostAllocation"), "dynamicHostAllocation can only be set when tenancy is 'host'"))
+	}
+
```
huali9, line 506: Can we also add validation for this configuration?
```
  hostAffinity: host
  tenancy: host
```
AWS doesn't allow this configuration combination. When attempting to create an instance with Affinity=host,Tenancy=host:
```
(my_virtualenv) liuhuali@Lius-MacBook-Pro huali-test % AWS_PROFILE=saml aws ec2 run-instances --region us-east-2 \
    --image-id ami-0bc8dda494f111572 --count 1 \
    --instance-type m6i.xlarge --key-name huliu-testdns \
    --security-group-ids sg-00efcc7e740bd7772 sg-016a9dc342ab25808 \
    --subnet-id subnet-04bc48559ab2a0bf8 \
    --placement Affinity=host,Tenancy=host

An error occurred (InvalidHostId) when calling the RunInstances operation: The Dedicated host ID null does not exist.
```

MAPI already has validation in place to prevent this:
```
liuhuali@Lius-MacBook-Pro huali-test % oc create -f ms1.yaml 
Error from server (Forbidden): error when creating "ms1.yaml": admission webhook "validation.machineset.machine.openshift.io" denied the request: spec.placement.host.dedicatedHost: Required value: dedicatedHost is required when hostAffinity is DedicatedHost, and optional otherwise
```
Without adding validation, this causes conversion failures:
```
    message: 'failed to convert CAPI machine set to MAPI machine set: spec.dedicatedHost.id:
      Required value: id is required and must start with ''h-'' followed by 8 or 17
      lowercase hexadecimal characters (0-9 and a-f)'
    reason: FailedToConvertCAPIMachineSetToMAPI
```

api/v1beta2/awsmachine_webhook.go
huali9: I tried `DynamicHostAllocation` in CAPI but get the error below. Does CAPI actually support `DynamicHostAllocation`? @vr4manta 
```
  - lastTransitionTime: "2026-01-19T07:34:23Z"
    message: 'failed to create AWSMachine instance: failed to allocate dedicated host:
      failed to allocate dedicated host: failed to allocate dedicated host: ......:
      operation error EC2: AllocateHosts, https response error StatusCode: 403, RequestID:
      79653b3c-6246-4c34-a4ab-d288c588dfc8, api error UnauthorizedOperation: You are
      not authorized to perform this operation. User: arn:aws:iam::301721915996:user/huliu-aws0119b-jpmjm-openshift-cluster-api-aws-hsrjr
      is not authorized to perform: ec2:AllocateHosts on resource: arn:aws:ec2:us-east-2:301721915996:dedicated-host/*
      because no identity-based policy allows the ec2:AllocateHosts action. Encoded
      authorization failure message: ......'
    reason: InstanceProvisionFailed
    severity: Error
    status: "False"
    type: InstanceReady

```

vr4manta: In MAPI / openshift API, this is still being worked on.

api/v1beta2/awsmachine_webhook.go
nrb: I think we need one more case in the validation, where both affinity and tenancy are `host` and `hostID` is empty - per the example at https://docs.aws.amazon.com/license-manager/latest/userguide/conversion-tenancy.html

```bash
aws ec2 modify-instance-placement --instance-id <instance_id> \
  --tenancy host --affinity host --host-id <host_id>
```

In other words, setting the tenancy and affinity requires you also set the host ID, but the validation as it is here doesn't enforce that at the Kube API level.

vr4manta: I can work on adding that logic next.  Let me sync all these up and see how the integration tests are looking.  Thanks!

api/v1beta2/awsmachine_webhook.go
huali9: Good to know, thank you!

PR #5828, opened 2026-01-15
🐛fix: Clamp nodegroup desiredSize and skip when externally managed

### LINKED ISSUE #5746: AWSManagedMachinePool sometimes unable to edit minSize/maxSize depending on MachinePool replicas managed by external-autoscaler
/kind bug
AWSManagedMachinePool is unable to effectively set minSize/maxSize depending on MachinePool replicas which is used as the desiredSize. This is especially troublesome when using cluster-autoscaler (with cluster.x-k8s.io/replicas-managed-by annotation on MachinePool).

Example:
When -
MachinePool.spec.replicas = 0
AWSManagedMachinePool.spec.scaling = {minSize: 0, maxSize: 5}

We want to update the node group to always have 2 nodes. 
Update via gitops - AWSManagedMachinePool.spec.scaling = {minSize: 2, maxSize: 5}

Reconciliation fails with 400 from AWS stating that desiredReplicas 0 cannot be less than minSize: 2. The same will probably happen if we try to lower maxSize < MachinePool.spec.replicas.

Workaround: 
1) remove `cluster.x-k8s.io/replicas-managed-by` annotation
2) set MachinePool replicas manually, set minSize on AWSManagedMachinePool.

```
// ReplicasManagedByAnnotation is an annotation that indicates external (non-Cluster API) management of infra scaling.
	// The practical effect of this is that the capi "replica" count should be passively derived from the number of observed infra machines,
	// instead of being a source of truth for eventual consistency.
	// This annotation can be used to inform MachinePool status during in-progress scaling scenarios.
	ReplicasManagedByAnnotation = "cluster.x-k8s.io/replicas-managed-by"
```

Setting minSize > observed replica count will bump desiredSize = minSize. Could be some push back on this on the opposite end where setting maxSize < observed replica count will drain nodes when setting desiredSize = maxSize.

It would be truer to definition if desiredSize was set with a clamp based on configured minSize/maxSize.

**Environment:**

- Cluster-api-provider-aws version: latest
- Kubernetes version: (use `kubectl version`): n/a

Minh-Ng:
**What type of PR is this?**
/kind bug

When cluster autoscaler manages node scaling for AWSManagedMachinePool, we should not call UpdateNodeGroupConfig with desired replica counts. Similarly if we update a min/max size, it should change the desired size appropriately. Same goes for AWSMachinePool

Fixes #5746

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
Fix AWSManagedMachinePool and AWSMachinePool to skip desired replica updates when externally managed, and clamp desiredSize within min/max bounds when scaling limits are updated.
```

### DIFF
--- a/pkg/cloud/services/autoscaling/autoscalinggroup.go
+++ b/pkg/cloud/services/autoscaling/autoscalinggroup.go
@@ -277,16 +277,28 @@ func (s *Service) UpdateASG(machinePoolScope *scope.MachinePoolScope) error {
 		return fmt.Errorf("getting subnets for ASG: %w", err)
 	}
 
+	minSize := machinePoolScope.AWSMachinePool.Spec.MinSize
+	maxSize := machinePoolScope.AWSMachinePool.Spec.MaxSize
+
 	input := &autoscaling.UpdateAutoScalingGroupInput{
 		AutoScalingGroupName: aws.String(machinePoolScope.Name()), // TODO: define dynamically - borrow logic from ec2
-		MaxSize:              aws.Int32(machinePoolScope.AWSMachinePool.Spec.MaxSize),
-		MinSize:              aws.Int32(machinePoolScope.AWSMachinePool.Spec.MinSize),
+		MaxSize:              aws.Int32(maxSize),
+		MinSize:              aws.Int32(minSize),
 		VPCZoneIdentifier:    aws.String(strings.Join(subnetIDs, ",")),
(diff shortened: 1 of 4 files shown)


Breee:
@AndiDog  @Minh-Ng any idea when this gets merged?

Issue #5176, opened 2024-10-23
[e2e] Review all the tests marked pending


richardcase:
/kind feature

**Describe the solution you'd like**

Review all the e2e tests marked pending (i.e. `PDescribe`). Some of these are covered by #5169  but there are others. For example, we had some tests that were testing k8s version prior to v1.23 for intree CCM/CSI etc. 

If tests are no longer needed then we need to remove the code.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

**Environment:**

- Cluster-api-provider-aws version: 
- Kubernetes version: (use `kubectl version`): 
- OS (e.g. from `/etc/os-release`):

Arpit529Srivastava:
hey, @richardcase would like to work on this. can u please assign
thanks

Arpit529Srivastava:
hey, @richardcase planning to address the pending tests in separate prs by category. will start with the simpler ones first (dedicated hosts decision, clusterclass enablement).

Arpit529Srivastava:
i looked into the pending tests and found 7 `pdescribe `tests and 2 `ginkgo.skip` tests.
- [x] dedicated hosts test (cost decision)  #5829 
- [ ] clusterclass tests  https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5838 
- [ ] gpu test (server-side apply fix)  https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5852
- [x] clusterctl upgrade test  https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5843
- [ ] multi-cluster & external infrastructure tests  
- [ ] upgrade to main branch test (env var refactor)  https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5844

PR #5844, opened 2026-01-22
🌱 test: enable k8s upgrade test 

Arpit529Srivastava:
**What type of PR is this?**
/kind cleanup

enables the “upgrade to main branch kubernetes” e2e test and removes unused environment variable handling.
the test is enabled by switching `PDescribe` to `Describe`. at the same time, dead code related to the `USE_CI_ARTIFACTS` environment variable is removed, as it was set but never read anywhere in the repository.
the test already correctly uses the `upgrade-ci-artifacts` flavor directly, so relying on an environment variable was unnecessary. this simplifies the code path and ensures the test behavior is explicit and easier to reason about.

related #5176 

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/suites/unmanaged/unmanaged_functional_test.go
+++ b/test/e2e/suites/unmanaged/unmanaged_functional_test.go
@@ -278,6 +278,7 @@ var _ = ginkgo.Context("[unmanaged] [functional]", func() {
 				searchSemVer, err := semver.Make(strings.TrimPrefix(e2eCtx.E2EConfig.MustGetVariable(shared.KubernetesVersion), tagPrefix))
 				Expect(err).NotTo(HaveOccurred())
 
+				shared.SetEnvVar("USE_CI_ARTIFACTS", "true", false)
 				shared.SetEnvVar(shared.KubernetesVersion, "v"+searchSemVer.String(), false)
 				configCluster := defaultConfigCluster(cluster1Name, namespace.Name)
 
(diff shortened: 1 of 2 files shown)


Arpit529Srivastava:
@faiq a gentle ping, thanks

test/e2e/suites/unmanaged/unmanaged_functional_test.go
```diff
@@ -275,11 +274,11 @@ var _ = ginkgo.Context("[unmanaged] [functional]", func() {
 				defer shared.DumpSpecResourcesAndCleanup(ctx, "", namespace, e2eCtx)
 				ginkgo.By("Creating first cluster with single control plane")
 				cluster1Name := fmt.Sprintf("%s-%s", specName, util.RandomString(6))
-				shared.SetEnvVar("USE_CI_ARTIFACTS", "true", false)
 				tagPrefix := "v"
 				searchSemVer, err := semver.Make(strings.TrimPrefix(e2eCtx.E2EConfig.MustGetVariable(shared.KubernetesVersion), tagPrefix))
 				Expect(err).NotTo(HaveOccurred())
 
+				shared.SetEnvVar("USE_CI_ARTIFACTS", "true", false)
```
nrb, line 281: I take it because this was added back, it's still required?

Arpit529Srivastava: yup. i initially thought it was unused and removed it but that caused the e2e tests to fail. without `USE_CI_ARTIFACTS=true` the test wasn’t picking up the correct template for the upgrade scenario which led to validation errors, specifically attempts to modify immutable fields like `vpc.id`.

i added it back so the correct artifacts are used and the test passes as expected.

PR #5850, opened 2026-01-25
🌱 [KAL] Enable jsontags linter and scope it to API directories and fix …

### LINKED ISSUE #5487: [KAL] Enable the jsontags linter
Enable the json tags linter

Zeyadzahran:
/kind cleanup

This PR enables the `jsontags` rule in Kube API Linter (KAL) and ensures it is enforced only on Kubernetes API types.

The `jsontags` rule is meant to validate the public API surface. When enabled globally, it also reports violations in internal runtime structs (scopes, helpers, controllers), which are not part of the API and should not require JSON tags.

Although the configuration already attempts to scope KAL to `api/**`, the custom `golangci-lint-kal` build does not reliably honor path-based exclusions for individual KAL rules. As a result, `jsontags` still fires outside API packages.

This change makes the intent explicit at the invocation level:

- Enable `jsontags` in `.golangci-kal.yml`.
- Update the `lint-api` target to run KAL only on known API directories.

With this approach, KAL is executed exclusively against API packages, so `jsontags` is enforced exactly where it belongson Kubernetes API types and never on internal runtime code.

In addition, missing and incorrect JSON tags in API types have been fixed to satisfy the new rule and align with Kubernetes API conventions.

**Which issue this PR fixes**:
Fixes #5487 

This intentionally avoids adding JSON tags to internal structs. The Makefile change is a workaround for current limitations in the custom KAL build with respect to path-based exclusions.

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.golangci-kal.yml
+++ b/.golangci-kal.yml
@@ -20,7 +20,7 @@ linters:
               #- "commentstart" # Ensure comments start with the serialized version of the field name.
               #- "conditions" # Ensure conditions have the correct json tags and markers.
               #- "integers" # Ensure only int32 and int64 are used for integers.
-              #- "jsontags" # Ensure every field has a json tag.
+              - "jsontags" # Ensure every field has a json tag.
               #- "maxlength" # Ensure all strings and arrays have maximum lengths/maximum items.
               #- "nobools" # Bools do not evolve over time, should use enums instead.
               #- "nofloats" # Ensure floats are not used.
@@ -46,14 +46,9 @@ linters:
 
 exclusions:
     generated: strict
-    paths:
-      - zz_generated.*\.go$
-      - ".*_test.go"  # Exclude test files.
+    # NOTE: Path-based exclusions via config don't work reliably in this golangci-lint version.
+    # Instead, we use command-line path arguments in the Makefile to limit linting to API directories only.
     rules:
-    # KAL should only run on API folders.
-    - path-except: "api//*"
-      linters:
-        - kubeapilinter
     - path: "api/v1beta1/*|api/v1alpha1/*"
       text: "Conditions field must be a slice of metav1.Condition"
       linters:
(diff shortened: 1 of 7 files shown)


Zeyadzahran:
Please Can Anyone See This PR  , this is my first Contrbution @AndiDog  , @damdo

api/v1beta1/tags.go
```diff
@@ -212,25 +212,25 @@ func ClusterAWSCloudProviderTagKey(name string) string {
 // BuildParams is used to build tags around an aws resource.
 type BuildParams struct {
```
AndiDog, line 213: In the whole PR change, there are unnecessary changes to the API. We're not using `BuildParams.*` as JSON fields, so they shouldn't be tagged as such, right?

iam/api/v1beta1/types.go
```diff
@@ -82,21 +82,23 @@ const (
...
+// Note: JSON field names must be capitalized to match AWS IAM policy format.
 type StatementEntry struct {
-	Sid          string     `json:",omitempty"`
-	Principal    Principals `json:",omitempty"`
-	NotPrincipal Principals `json:",omitempty"`
-	Effect       Effect     `json:"Effect"`
-	Action       Actions    `json:"Action"`
-	Resource     Resources  `json:",omitempty"`
-	Condition    Conditions `json:"Condition,omitempty"`
+	Sid          string     `json:"Sid,omitempty"` //nolint:kubeapilinter
```
AndiDog, line 95: This seems to be semantically the same as before: When deserializing into a Go struct, the `json` package doesn't care about lowercase or uppercase. When serializing, it takes the `Uppercase` field name of the struct if no name is given, and you changed that to be explicit.

You need to format the Go code, though, and set your editor to do so automatically.

.golangci-kal.yml
```diff
@@ -46,14 +46,9 @@ linters:
 
 exclusions:
     generated: strict
-    paths:
-      - zz_generated.*\.go$
-      - ".*_test.go"  # Exclude test files.
+    # NOTE: Path-based exclusions via config don't work reliably in this golangci-lint version.
```
AndiDog, line 49: This comment isn't actionable. Is there an issue we can link to? What exactly doesn't work?

Issue #5835, opened 2026-01-19
Upgrade to remaining v1beta2 CAPI types


nrb:
[Cluster API has decided to no longer serve their v1beta1 APIs in CAPI v1.14](https://github.com/kubernetes-sigs/cluster-api/issues/11920), and ultimately remove those versions in v1.18.

In order to stay current, we will need to adopt the v1beta2 types. Our v2.10 release was able to rely on the compatibility layer that CAPI is providing for now, but that layer will eventually be removed.

However, we do not want to upgrade the CAPI Go release and v1beta2 types at the same time; this would create a lot of code churn across our repo and ecosystem. #5834 is for tracking the CAPI release upgrade separately.

Support for CAPI releases and API versions should look like the following:

| CAPA release | CAPI release | CAPI API version |
|--------------|--------------|------------------|
| v2.10        | v1.11        | v1beta1 and some v1beta2 |
| v2.11        | v1.12        | v1beta1 and some v1beta2 |
| v2.12        | v1.12        | v1beta2          |
| v2.13        | v1.13?       | v1beta2          |

We can use this issue to discuss and coordinate our change.

Currently, what's left undecided is:

* Will we increment our own Kubernetes API version when we fully adopt the v1beta2 changes? (*Other breaking changes that might be introduced should be discussed in their respective issues*)
* Any other potential issues?

/label triage/accepted
/label priority/important-soon

Arpit529Srivastava:
@nrb would like to work on this ,thanks.
/assign

Arpit529Srivastava:
adding discussion for more context to this issue => https://kubernetes.slack.com/archives/CD6U2V71N/p1769634444859949

Arpit529Srivastava:
the plan is roughly:

1. add `V1Beta2` fields to all types (starting with `AWSCluster`)  #5854
2. update controllers to dual-write to both the old and new condition fields  
3. migrate the condition constants  
4. switch consumers over to the new fields

damdo:
Closing in favour of #5448, which now has a detailed two-phase migration plan with sub-issues.

PR #5854, opened 2026-01-29
✨ Add v1beta2 conditions infrastructure to AWSCluster

Arpit529Srivastava:
**What type of PR is this?**
/kind api-change
/kind feature

This pr prepares `AWSCluster` for migrating conditions to the new `metav1.Condition` format, following the same explicit dual-write pattern used in CAPI.

it adds a new `V1Beta2` field to `AWSClusterStatus` to hold the next-generation condition set alongside the existing one, and implements `GetV1Beta2Conditions` and `SetV1Beta2Conditions` on `AWSCluster`. the conversion logic is updated so the `V1Beta2` field is preserved across version conversions (v1beta2 ↔ v1beta1).

like capi, this keeps the dual-write logic explicit rather than introducing helper abstractions. that way, deprecated v1beta1 logic stays clearly separated from the newer v1beta2+ logic, making it easier to remove the legacy path later.

the actual v1beta2 condition constants will be introduced in a follow-up pr alongside the controller changes that will consume them.

related : https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5835

PR *only* includes API/Status changes and constants. No controller logic is changed in this PR. The controller implementation (dual-writing) will be handled in a follow-up PR.
**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
 Added v1beta2 conditions infrastructure (Status field and Constants) to `AWSCluster` to support future migration.
```


Arpit529Srivastava:
@damdo a gentle ping, thanks :)

api/v1beta1/awscluster_conversion.go
```diff
@@ -150,6 +150,9 @@ func (src *AWSCluster) ConvertTo(dstRaw conversion.Hub) error {
 		}
```
damdo, line 150: I don't think we should touch v1beta1 types at all.
Only v1beta2 ones

api/v1beta2/awscluster_types.go
```diff
@@ -281,6 +281,23 @@ type AWSClusterStatus struct {
...
+
+	// v1beta3 groups all the fields that will be added or modified in AWSCluster's status with the V1Beta3 version.
+	// +optional
+	V1Beta3 *AWSClusterV1Beta3Status `json:"v1beta3,omitempty"`
+}
+
+// AWSClusterV1Beta3Status groups all the fields that will be added or modified in AWSClusterStatus with the V1Beta3 version.
+// See https://github.com/kubernetes-sigs/cluster-api/blob/main/docs/proposals/20240916-improve-status-in-CAPI-resources.md for more context.
+type AWSClusterV1Beta3Status struct {
+	// conditions represents the observations of an AWSCluster's current state.
```
damdo:
```suggestion
	// Conditions represents the observations of an AWSCluster's current state.
```

api/v1beta2/conditions_consts.go
```diff
@@ -199,3 +199,98 @@ const (
 	// DedicatedHostReleaseFailedReason used when the dedicated host release fails.
 	DedicatedHostReleaseFailedReason = "DedicatedHostReleaseFailed"
 )
+
+// AWSCluster's v1beta2 conditions and corresponding reasons.
```
damdo:
```suggestion
// AWSCluster's v1beta3 conditions and corresponding reasons.
```

damdo: Also we should probably move these to a different place, how are other CAPI providers doing this during their transition?

api/v1beta2/conditions_consts.go
```diff
@@ -199,3 +199,98 @@ const (
...
+// These will be used with the V1Beta2 API version.
```
damdo: Will be or, are for CAPA v1beta2?

api/v1beta2/conditions_consts.go
```diff
@@ -199,3 +199,98 @@ const (
...
+
+// AWSCluster's Ready condition and corresponding reasons that will be used in v1Beta2 API version.
```
damdo: Same here

api/v1beta2/awscluster_types.go
```diff
@@ -281,6 +281,23 @@ type AWSClusterStatus struct {
...
+	// Known condition types are Ready, VpcReady, SubnetsReady, InternetGatewayReady, NatGatewaysReady,
+	// RouteTablesReady, ClusterSecurityGroupsReady, BastionHostReady, LoadBalancerReady, and Paused.
+	// +optional
+	// +listType=map
+	// +listMapKey=type
+	// +kubebuilder:validation:MaxItems=32
+	Conditions []metav1.Condition `json:"conditions,omitempty"`
```
damdo, line 300: Are we only dealing with AWSCluster and Conditions here?
If we do it this granular we'll potentially end up with (number of API types) x (number of fields to add) PRs

damdo (commented):
Hey @Arpit529Srivastava thanks a lot for your PR!
This is a good start :)

Left some comments

api/v1beta2/conditions_consts.go
Arpit529Srivastava: i took another look and separating them does seem like the cleaner approach. i’ll move the new v1beta3 constants into a separate file (`api/v1beta2/conditions_consts_v1beta3.go`).

api/v1beta2/awscluster_types.go
Arpit529Srivastava: my bad, i was initially planning to split these into separate prs to keep things smaller,. i agree it’s better to avoid creating too many prs for closely related changes.

Arpit529Srivastava:
> Hey @Arpit529Srivastava thanks a lot for your PR!
> This is a good start :)
> 
> Left some comments

thanks for the review 😃, taking a look.

api/v1beta1/awscluster_conversion.go
Arpit529Srivastava: @damdo just to make sure i’m understanding correctly about the `v1beta1` conversion changes, should i remove the `V1Beta3` restoration logic from `awscluster_conversion.go`? i followed the same pattern used for existing fields like `ControlPlaneLoadBalancer`, but i’m totally fine removing it if that’s not the direction we want to take.
wanted to confirm this part before i move on to the rest of the feedback. thanks

api/v1beta2/awscluster_types.go
Arpit529Srivastava: rethinking again, once we settle on the approach for v1beta1 conversions, https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5854#discussion_r2767975567 i’ll add the same `V1Beta3` status field to the other types as well (`AWSMachine`, `AWSMachinePool`, `AWSManagedMachinePool`, etc ...) in this same pr. just wanted to make sure we’re aligned on the pattern with `AWSCluster` first before rolling it out everywhere.

Arpit529Srivastava:
@clebs, can you please take a look.
Thanks

api/v1beta2/conditions_consts_v1beta3.go
```diff
@@ -0,0 +1,114 @@
+/*
```
clebs: Why are these constants introduced? They are not being used anywhere and most values are duplicating existing constants.

Arpit529Srivastava: hmm, make sense.. i initially added these constants as groundwork for the next step, where the `AWSClusterController` will dual-write conditions into both `status.conditions` and `status.v1beta3.conditions`. however, rightg now since nothing in this pr actually consumes them yet, ill remove them.

Arpit529Srivastava: will introduce them together with the controller changes in the follow-up pr where they’ll actually be used.

clebs: Great thanks! Makes sense to introduce them in the next step. Also only introducing what is necessary.

Arpit529Srivastava: also @clebs, could you please advise me on this [comment](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5854#discussion_r2768114046)? I’m a little stuck here. Thanks, I really appreciate your help.

clebs: Looking at the initial comment from @damdo, he is saying you should not update files under `api/v1beta1`. Only files under `api/v1beta2`.

The comment is not asking to change CAPI conversion, just to not include it in the older CAPA v1beta1 API.

Please bear in mind that CAPI v1beta1 and CAPA v1beta1 are 2 different things.

Arpit529Srivastava: thanks @clebs for the review will keep this in mind.

api/v1beta1/awscluster_conversion.go
richardcase: As we are adding new fields in v1beta2 then we need to preserve roundtrip compatibility. So it will require that you update the conversions in v1beta1 like you have done here to restore the value.

richardcase: A lot of people make the mistake of adding the new field to old api versions as well to make the conversions work. But restoring the value from the saved state is the correct way (like you are doing).

api/v1beta2/awscluster_types.go
richardcase: I'm happy either way. Smaller PRs always helps with review though.

api/v1beta2/awscluster_types.go
```diff
@@ -355,6 +372,22 @@ func (r *AWSCluster) SetConditions(conditions clusterv1beta1.Conditions) {
 	r.Status.Conditions = conditions
 }
 
+// GetV1Beta3Conditions returns the set of conditions for this object.
+func (r *AWSCluster) GetV1Beta3Conditions() []metav1.Condition {
```
richardcase: This is where the naming of things is going to get weird 😉 

We need to adopt `paused.EnsurePausedCondition` in our controllers. This means that `AWSCluster` (and our other resource types) need to implement the `v1beta2conditions.Setter/Getter` interfaces . So these methods need to be named:

```go
	// SetV1Beta2Conditions sets conditions for an API object.
	// Note: SetV1Beta2Conditions will be renamed to SetConditions in a later stage of the transition to V1Beta2.
	SetV1Beta2Conditions([]metav1.Condition)

	// GetV1Beta2Conditions returns the list of conditions for a cluster API object.
	// Note: GetV1Beta2Conditions will be renamed to GetConditions in a later stage of the transition to V1Beta2.
	GetV1Beta2Conditions() []metav1.Condition
```

Or we need additional functions. 

It will be very messy/weird until we drop the old fields.

richardcase (changes requested):

api/v1beta1/awscluster_conversion.go
damdo: > As we are adding new fields in v1beta2 then we need to preserve roundtrip compatibility. So it will require that you update the conversions in v1beta1 like you have done here to restore the value.

Good point! Hadn't thought about that!

api/v1beta2/awscluster_types.go
Arpit529Srivastava: thanks for clarifying 😄 
i’ve updated the implementation to use `GetV1Beta2Conditions()` and `SetV1Beta2Conditions()` along with the `Note:` comments you mentioned.

it does feel a bit awkward during the transition, but i agree it’s better to align exactly with capi’s interface expectations.

Arpit529Srivastava:
@richardcase @damdo made all corrections, PTAL .
thanks :)

api/v1beta2/awscluster_types.go
```diff
@@ -281,6 +281,23 @@ type AWSClusterStatus struct {
 	FailureDomains clusterv1beta1.FailureDomains `json:"failureDomains,omitempty"`
 	Bastion        *Instance                     `json:"bastion,omitempty"`
 	Conditions     clusterv1beta1.Conditions     `json:"conditions,omitempty"`
+
+	// v1beta2 groups all the fields that will be added or modified in AWSCluster's status with the V1Beta2 version.
+	// +optional
+	V1Beta2 *AWSClusterV1Beta2Status `json:"v1beta2,omitempty"`
+}
+
+// AWSClusterV1Beta2Status groups all the fields that will be added or modified in AWSCluster with the V1Beta2 version.
```
richardcase:
```suggestion
// AWSClusterV1Beta2Status groups all the fields that will be added or modified in AWSCluster with the V1Beta2 CAPI contract version.
```

api/v1beta2/awscluster_types.go
```diff
@@ -281,6 +281,23 @@ type AWSClusterStatus struct {
 	FailureDomains clusterv1beta1.FailureDomains `json:"failureDomains,omitempty"`
 	Bastion        *Instance                     `json:"bastion,omitempty"`
 	Conditions     clusterv1beta1.Conditions     `json:"conditions,omitempty"`
+
+	// v1beta2 groups all the fields that will be added or modified in AWSCluster's status with the V1Beta2 version.
```
richardcase:
```suggestion
	// v1beta2 groups all the fields that will be added or modified in AWSCluster's status with the V1Beta2 CAPI contract version.
```

api/v1beta2/awscluster_types.go
```diff
@@ -355,6 +372,24 @@ func (r *AWSCluster) SetConditions(conditions clusterv1beta1.Conditions) {
 	r.Status.Conditions = conditions
 }
 
+// GetV1Beta2Conditions returns the set of conditions for this object.
+// Note: GetV1Beta2Conditions will be renamed to GetConditions in a later stage of the transition to V1Beta2.
```
richardcase:
```suggestion
// Note: GetV1Beta2Conditions will be renamed to GetConditions in a later stage of the transition to CAPI contract V1Beta2.
```

api/v1beta2/awscluster_types.go
```diff
@@ -355,6 +372,24 @@ func (r *AWSCluster) SetConditions(conditions clusterv1beta1.Conditions) {
...
+func (r *AWSCluster) GetV1Beta2Conditions() []metav1.Condition {
+	if r.Status.V1Beta2 == nil {
+		return nil
+	}
+	return r.Status.V1Beta2.Conditions
+}
+
+// SetV1Beta2Conditions sets conditions for an API object.
+// Note: SetV1Beta2Conditions will be renamed to SetConditions in a later stage of the transition to V1Beta2.
```
richardcase:
```suggestion
// Note: SetV1Beta2Conditions will be renamed to SetConditions in a later stage of the transition to CAP contract V1Beta2.
```

richardcase (commented):
Looking good @Arpit529Srivastava . I made some suggestions around the comments to make it obvious that we are talking about "CAPI contract" v1beta2 and not CAPA v1beta2.

Would you now plan to add the same changes for the other resource kinds?

Arpit529Srivastava:
> Looking good @Arpit529Srivastava . I made some suggestions around the comments to make it obvious that we are talking about "CAPI contract" v1beta2 and not CAPA v1beta2.
> 
> Would you now plan to add the same changes for the other resource kinds?

for the other resource kinds - yes, i’m planning to add them as well. since you mentioned that smaller prs are easier to review, i started with `AWSCluster` first and will follow up with separate prs for the remaining types.

is there anything specific i should watch out for when applying the same pattern to the other resources? any heads-up before i proceed?
thanks for reviewing :)

api/v1beta2/awscluster_types.go
Arpit529Srivastava: addressed

richardcase:
@Arpit529Srivastava - it might be nice to add a task list to #5835 so that progress can be tracked and people can help out. Once the conditions are added to the types, i'm guessing that the next stage is to then dual right conditions to the new fields?

Arpit529Srivastava:
sure that is much better way, added a task list to #5835 for tracking. thanks 
if everything looks good, we can proceed with merging. awaiting your review. :)

api/v1beta2/awscluster_types.go
Arpit529Srivastava: done

richardcase:
Sorry @Arpit529Srivastava this fell off my radar. Is there anything from your side that you think needs doing with this?

clebs:
This has been neglected a bit because of all the issues we faced for CAPI 1.12... I do not see any overlap that would make an already huge PR harder to merge.
It is probably safe to get this in.

damdo:
/hold

I'm drafting a new detailed plan with new understanding. Let's see how this fits

PR #5856, opened 2026-02-02
fix: return correct error in NAT gateway creation

atharrva01:
## Summary

This PR fixes a silent error propagation bug during NAT gateway creation that could cause clusters to appear **successfully provisioned while private subnet networking was actually broken**.

While working on this, I noticed that failures from concurrent NAT gateway creation goroutines were not being returned correctly. As a result, reconciliation could continue even when one or more NAT gateways failed to be created.

---

## What was happening

During NAT gateway creation, Elastic IPs are allocated first, followed by concurrent NAT gateway provisioning.

If any NAT gateway creation failed, the error was detected but **not returned correctly**. The function ended up returning `(nil, nil)`, silently swallowing the failure.

```go
if ngwResult.error != nil {
    return nil, err // err was already nil here
}
```

At this point, `err` referred to the named return variable, which had already been set to `nil` after successful Elastic IP allocation.

---

## Why this matters (Impact)

This bug could lead to **“Ready but broken” clusters**:

* Workloads in private subnets lose internet access
* Nodes fail to pull container images or bootstrap correctly
* Cluster status reports success despite missing NAT gateways
* Elastic IPs may be allocated but left orphaned
* Users only discover the issue later through failing workloads

Because the failure is silent, it’s especially hard to debug in production.

---

## The fix

The fix is intentionally small and focused: **return the actual goroutine error** so it propagates correctly to the reconciler.

```diff
if ngwResult.error != nil {
-    return nil, err
+    return nil, ngwResult.error
}
```

With this change, NAT gateway creation failures correctly surface, allowing reconciliation to fail fast and report accurate cluster status.

---

## Result after this change

* Infrastructure failures are no longer silently ignored
* Cluster status correctly reflects networking failures
* NAT gateway errors trigger proper requeues
* Prevents orphaned Elastic IP allocation
* Users get clear, actionable error messages

I’ve kept the fix minimal to avoid changing any success-path behavior, while restoring correct error handling during reconciliation.

---

### DIFF
--- a/pkg/cloud/services/network/natgateways.go
+++ b/pkg/cloud/services/network/natgateways.go
@@ -255,11 +255,12 @@ func (s *Service) getNatGatewayTagParams(id string) infrav1.BuildParams {
 	}
 }
 
-func (s *Service) createNatGateways(subnetIDs []string) (natgateways []*types.NatGateway, err error) {
+func (s *Service) createNatGateways(subnetIDs []string) ([]*types.NatGateway, error) {
 	eips, err := s.getOrAllocateAddresses(len(subnetIDs), infrav1.CommonRoleTagValue, s.scope.VPC().GetElasticIPPool())
 	if err != nil {
 		return nil, errors.Wrapf(err, "failed to create one or more IP addresses for NAT gateways")
 	}
+
 	type ngwCreation struct {
 		natGateway *types.NatGateway
 		error      error
@@ -273,14 +274,18 @@ func (s *Service) createNatGateways(subnetIDs []string) (natgateways []*types.Na
 		}(c, sn, eips[i])
 	}
 
+	var natgateways []*types.NatGateway
+	var errs []error
 	for range subnetIDs {
 		ngwResult := <-c
 		if ngwResult.error != nil {
-			return nil, ngwResult.error
+			errs = append(errs, ngwResult.error)
+			continue
 		}
 		natgateways = append(natgateways, ngwResult.natGateway)
 	}
-	return natgateways, nil
+
+	return natgateways, kerrors.NewAggregate(errs)
 }
 
 func (s *Service) createNatGateway(subnetID, ip string) (*types.NatGateway, error) {


atharrva01:
hi @damdo @faiq,  I’m returning the actual NAT gateway creation error here instead of the named return value, which was silently swallowing failures.

pkg/cloud/services/network/natgateways.go
```diff
@@ -276,7 +276,7 @@ func (s *Service) createNatGateways(subnetIDs []string) (natgateways []*types.Na
 	for range subnetIDs {
 		ngwResult := <-c
 		if ngwResult.error != nil {
-			return nil, err
+			return nil, ngwResult.error
```
faiq: I don't think it makes sense to return just one error here and nil. What if some nat gateways were created?

What if we did something like this?

```golang
var errs []error
for range subnetIDs {
		ngwResult := <-c
		if ngwResult.error != nil {
			errs=append(errs, ngwResult.error)
            continue
         }
         natgateways = append(natgateways, ngwResult.natGateway)
   }
return natgateways, errors.Join(errs)
```

richardcase:
@atharrva01 - can you confirm how you came about the issue that this pr fixes? How did you replicate the issue (I.e the steps) and what cluster definitions did you use?

pkg/cloud/services/network/natgateways.go
atharrva01: hi @faiq, Thanks for the suggestion, I’ve updated the implementation to aggregate errors using kerrors.NewAggregate and return successfully created NAT gateways alongside any errors. This aligns with the existing pattern in deleteNatGateways and preserves reconciliation semantics.

atharrva01:
Hi @richardcase, good question.

I actually found this while walking through the NAT gateway creation flow and trying to understand how errors were handled across the goroutines.

When I traced it carefully, I noticed that inside the result loop we were returning the named err variable instead of ngwResult.error. At that point err had already been set to nil after the EIP allocation step, so if a NAT gateway creation failed, the function would still return (nil, nil).

I didn’t hit this from a real cluster failure it was more from stepping through the control flow and reasoning about what would happen if createNatGateway returned an error.

Happy to add a small test around this if that would be useful.

PR #5861, opened 2026-02-07
🌱 fix(test): resolve flaky TestROSANetworkReconciler_Reconcile by using Eventually

VijetaPriya47:
### Description
This PR fixes a flaky test failure in [TestROSANetworkReconciler_Reconcile](cci:1://file:///home/vijetapriya/cluster-api-provider-aws/exp/controllers/rosanetwork_controller_test.go:45:0-526:1) related to race conditions between the controller-runtime cache and the API server.

Previously, the test would assert conditions immediately after an API call. However, the controller-runtime client uses a cache that might not have been updated yet, leading to false positives or failures when reading the status.

### Changes
- Replaced immediate `g.Expect` assertions with `g.Eventually(...).Should(Succeed())` blocks for status checks. This ensures the test waits for the cache to update instead of failing on stale data.
- Scoped `g := NewWithT(t)` inside each `t.Run` subtest. This ensures that test failures are reported to the correct subtest instance and prevents panics when a subtest fails.

### Verification
Verified by running the test multiple times locally:
`go test -v -count=5 ./exp/controllers/... -run TestROSANetworkReconciler_Reconcile`

**What type of PR is this?**

- [X] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title 
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/exp/controllers/rosanetwork_controller_test.go
+++ b/exp/controllers/rosanetwork_controller_test.go
@@ -158,24 +158,13 @@ func TestROSANetworkReconciler_Reconcile(t *testing.T) {
 		g.Expect(reqReconcile.RequeueAfter).To(Equal(time.Duration(0)))
 		g.Expect(errReconcile).To(MatchError(ContainSubstring("failed to start CF stack creation:")))
 
-		g.Eventually(func() error {
+		g.Eventually(func(g Gomega) {
 			cnd, err := getROSANetworkReadyCondition(reconciler, rosaNetwork)
-			if err != nil {
-				return err
-			}
-			if cnd == nil {
-				return fmt.Errorf("condition is nil")
-			}
-			if cnd.Reason != expinfrav1.ROSANetworkFailedReason {
-				return fmt.Errorf("reason mismatch: got %s, want %s", cnd.Reason, expinfrav1.ROSANetworkFailedReason)
-			}
-			if cnd.Severity != clusterv1beta1.ConditionSeverityError {
-				return fmt.Errorf("severity mismatch: got %s, want %s", cnd.Severity, clusterv1beta1.ConditionSeverityError)
-			}
-			if cnd.Message != "test-error" {
-				return fmt.Errorf("message mismatch: got %s, want %s", cnd.Message, "test-error")
-			}
-			return nil
+			g.Expect(err).NotTo(HaveOccurred())
+			g.Expect(cnd).ToNot(BeNil())
+			g.Expect(cnd.Reason).To(Equal(expinfrav1.ROSANetworkFailedReason))
+			g.Expect(cnd.Severity).To(Equal(clusterv1beta1.ConditionSeverityError))
+			g.Expect(cnd.Message).To(Equal("test-error"))
 		}).Should(Succeed())
 	})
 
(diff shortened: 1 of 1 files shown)


VijetaPriya47:
I think making an helper function for this repetitive code would be much better. WDYT? @richardcase @AndiDog @serngawy 
Also I hope the g.Expect are not intentional, as in such case, I would close my PR.

richardcase:
@VijetaPriya47 - how did you come across this issue? Are you using ROSA and in what situation did you see this?

VijetaPriya47:
@richardcase I came across this issue while running the full test suite locally (make test) during my development work on the project. I noticed that 
TestROSANetworkReconciler_Reconcile
 was failing intermittently
Then I searched for the unit test and saw that it uses mocked AWS clients.
I'm not currently using ROSA in a specific production scenario; I just wanted to ensure the test suite remains stable and reliable for everyone.

serngawy:
Thanks @VijetaPriya47 for fixing this , would you fix the PR title. You need to add emoji for support/bug to let the PR verify job pass.

VijetaPriya47:
@serngawy I'd like to update the title to 🐛  but I don't have permission to edit it. Could a maintainer please update it? Thanks for reviewing it though.

serngawy:
> @serngawy I'd like to update the title to 🐛 but I don't have permission to edit it. Could a maintainer please update it? Thanks for reviewing it though.

you are the PR owner, you should be able to change it. Just be sure you logged in with the same github user

VijetaPriya47:
@serngawy Done....could you again give it /ok-to-test

VijetaPriya47:
could you approve this @serngawy

exp/controllers/rosanetwork_controller_test.go
```diff
@@ -106,6 +106,8 @@ func TestROSANetworkReconciler_Reconcile(t *testing.T) {
 	g.Expect(err).NotTo(HaveOccurred())
 
 	t.Run("Empty result when ROSANetwork object not found", func(t *testing.T) {
+		g := NewWithT(t)
```
serngawy, line 109: no need to redefine the g (internal.Gomega) its already created above

exp/controllers/rosanetwork_controller_test.go
```diff
@@ -117,6 +119,8 @@ func TestROSANetworkReconciler_Reconcile(t *testing.T) {
 	})
 
 	t.Run("Error result when CF stack GET returns error", func(t *testing.T) {
+		g := NewWithT(t)
```
serngawy, line 122: same in all changes, no need to re define g

exp/controllers/rosanetwork_controller_test.go
```diff
@@ -152,15 +158,30 @@ func TestROSANetworkReconciler_Reconcile(t *testing.T) {
 		g.Expect(reqReconcile.RequeueAfter).To(Equal(time.Duration(0)))
 		g.Expect(errReconcile).To(MatchError(ContainSubstring("failed to start CF stack creation:")))
 
-		cnd, err := getROSANetworkReadyCondition(reconciler, rosaNetwork)
-		g.Expect(err).NotTo(HaveOccurred())
-		g.Expect(cnd).ToNot(BeNil())
-		g.Expect(cnd.Reason).To(Equal(expinfrav1.ROSANetworkFailedReason))
-		g.Expect(cnd.Severity).To(Equal(clusterv1beta1.ConditionSeverityError))
-		g.Expect(cnd.Message).To(Equal("test-error"))
+		g.Eventually(func() error {
```
serngawy: better to pass g Gomega as variable to the func and use it to check the expected condition instead of returning error. Example is [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosaroleconfig_controller_test.go#L372C15-L372C29)

exp/controllers/rosanetwork_controller_test.go
```diff
@@ -182,14 +203,27 @@ func TestROSANetworkReconciler_Reconcile(t *testing.T) {
 		g.Expect(reqReconcile.RequeueAfter).To(Equal(time.Duration(0)))
 		g.Expect(errReconcile).ToNot(HaveOccurred())
 
-		cnd, err := getROSANetworkReadyCondition(reconciler, rosaNetwork)
-		g.Expect(err).NotTo(HaveOccurred())
-		g.Expect(cnd).ToNot(BeNil())
-		g.Expect(cnd.Reason).To(Equal(expinfrav1.ROSANetworkCreatingReason))
-		g.Expect(cnd.Severity).To(Equal(clusterv1beta1.ConditionSeverityInfo))
+		g.Eventually(func() error {
```
serngawy: same as previous comment. For all changes as well

exp/controllers/rosanetwork_controller_test.go
VijetaPriya47: @serngawy what I was thinking is that, if we don't redefine g, failures in subtests would be reported to the parent test context instead of the specific subtest, which can cause incorrect test reporting.
And I noticed  that you can see this same pattern in 
controlplane/rosa/controllers/rosacontrolplane_controller_test.go
 at line 618, where g := NewWithT(t) is also redefined inside t.Run()." but it's inconsistent across the codebase. So WDYT?

serngawy: not sure about "incorrect test reporting" but we want to avoid the test been flaky, so we need to check that. Run the test without redefining the g inside the test run scope and see if it is flaky or not. Something like run the test with -count=10 as you mentioned in the PR description.

serngawy:
@damdo looks good to me , thanks @VijetaPriya47 for fixing that

PR #5062, opened 2024-07-18
:sparkles: Spot instances fallback - WaitingTimeout

Prucek:
**What type of PR is this?**

/kind feature
/kind api-change

**What this PR does / why we need it**:
This feature is adding `WaitingTimeout` for `SpotMarketOptions` that is optional. The `WaitingTimeout` serves as a timeout when no Spot instances are available and after this timeout is reached a regular ec2 instance will be acquired instead.

**Special notes for your reviewer**:

**Checklist**:
- [X] squashed commits
- [X] includes documentation
- [X] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [X] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:
```release-note
feat: add WaitingTimeout for Spot instances to fallback to a regular ec2 instance
```

PR #5868, opened 2026-02-11
:sparkles: spot instance fallback: add WaitingTimeout

Prucek:
**What type of PR is this?**
/kind feature
/kind api-change

This feature is adding WaitingTimeout for SpotMarketOptions that is optional. The WaitingTimeout serves as a timeout when no Spot instances are available and after this timeout is reached a regular ec2 instance will be acquired instead.

2nd attempt of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5062

**Checklist**:
- [X] squashed commits
- [X] includes documentation
- [X] includes emoji in title 
- [X] adds unit tests

**Release note**:
```release-note
feat: add WaitingTimeout for Spot instances to fallback to a regular ec2 instance
```


Prucek:
Could someone help me understand the e2e failures? And if they are valid or just flakes? I see the e2es have a very high error/flake rate.

Arpit529Srivastava (commented):
> Could someone help me understand the e2e failures? And if they are valid or just flakes? I see the e2es have a very high error/flake rate.

hey @Prucek i was looking into the e2e flakiness, agree with you none of the failures is related to this Pr.

some findings from my side
1. Clusterctl upgrade test failure - the CAPI test framework at `v1.11.4-0.20251125201037-d322ff6baa2f` (pinned in go.mod) changed how it builds the local filesystem repository. the old clusterctl v1.2.0 binary expects provider labels like `cluster-api`, but the new framework creates `core-cluster-api`. this is a compatibility break in the CAPI test framework, not this PR. 

2. I checked GPU test code and it doesn't use `SpotMarketOptions` at all. GPU tests are consistently flaky due to `p3.2xlarge` capacity.

3. The EKS auth test uses [cluster-template-eks-auth-bootstrap-disabled.yaml](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/data/eks/cluster-template-eks-auth-bootstrap-disabled.yaml), which creates only `AWSManagedCluster` + `AWSManagedControlPlane` (0 workers, no spot config). The DumpMachines panic is a pre-existing bug at [common.go:117](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/test/e2e/shared/common.go#L117) where `namespace` isn't nil-checked.

that being said,
could you check if commit 8fad5a160 (the base for this PR) passes these same tests on main? If main is also fails, then these are pre-existing issues. If main passes, then there's something more subtle I'm missing.

pkg/cloud/services/ec2/instances.go
```diff
@@ -736,7 +736,13 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
...
 	out, err := s.EC2Client.RunInstances(context.TODO(), input)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to run instance")
+		if !isSpotCapacityError(err) || i.SpotMarketOptions == nil || i.SpotMarketOptions.WaitingTimeout == nil {
+			return nil, errors.Wrap(err, "failed to run instance")
+		}
+		out, err = s.runInstancesWithSpotFallback(input, i.SpotMarketOptions)
+		if err != nil {
+			return nil, errors.Wrap(err, "failed to run instance with spot fallback")
+		}
```
damdo, line 747: I'd invert this condition for easier reading and control flow
```
		if isSpotCapacityError(err) && isSpotWaitingTimeoutDefined(i) {
		  out, err = s.runInstancesWithSpotFallback(input, i.SpotMarketOptions)
  		  if err != nil {
  		    return nil, errors.Wrap(err, "failed to run instance with spot fallback")
  		  }
		} else {
		  return nil, errors.Wrap(err, "failed to run instance")
        }
```

pkg/cloud/services/ec2/instances.go
```diff
@@ -746,6 +752,65 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
...
+func (s *Service) runInstancesWithSpotFallback(input *ec2.RunInstancesInput, spotOptions *infrav1.SpotMarketOptions) (*ec2.RunInstancesOutput, error) {
+	timeout := spotOptions.WaitingTimeout.Duration
+	deadline := time.Now().Add(timeout)
+
+	s.scope.Info("Attempting to acquire spot instance with fallback timeout",
+		"timeout", timeout.String())
+
+	for {
+		time.Sleep(spotRetryInterval)
+
```
damdo: I had a look at trying to understand if we can use the aws-sdk-go-v2's `SpotMarketOptions.ValidUntil` option (from [here](https://github.com/aws/aws-sdk-go-v2/blob/97845c0d6e4fc8e9af220fb5c623e77c1d9ebad8/service/ec2/types/types.go#L21391-L21400)) as it would significantly simplify the reconciliation logic

But my understanding is that to use it we'd need to set
```
spotOpts.SpotInstanceType = types.SpotInstanceTypePersistent                   // This is the only way to set ValidUntil for one-time requests.
spotOpts.InstanceInterruptionBehavior = types.InstanceInterruptionBehaviorStop // This is the only way to set InstanceInterruptionBehavior for one-time requests.
```

which might be problematic as it may conflict with "CAPA's current 1:1 Machine-to-Instance model".

Have you looked into this yourself?

Curious to see what others think of this

pkg/cloud/services/ec2/instances.go
```diff
@@ -746,6 +752,65 @@ func (s *Service) runInstance(role string, i *infrav1.Instance) (*infrav1.Instan
 	return s.SDKToInstance(out.Instances[0])
 }
 
+// spotRetryInterval is the interval between spot instance capacity checks.
+const spotRetryInterval = 30 * time.Second
```
damdo: We should probably make this configurable

damdo (commented):
Hey @Prucek I have a question about whether we could implement this with ValidUntil.

pkg/cloud/services/ec2/instances.go
Prucek: I added new fields to the status. This would solve the issue of not having the 1:1Machine-to-Instance.

pkg/cloud/services/ec2/instances.go
Prucek: I rather added the logic to the reconcile loop.

Prucek:
@damdo Could you take another look? Thanks!

PR #5869, opened 2026-02-11
🌱 Bump AWS CCM to v1.32.5 and add variable for it in E2Es

damdo:
Stacked on top of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5894

**What type of PR is this?**

/kind support

Bump AWS CCM to v1.32.5 and add variable for it in E2Es

**Release note**:

```release-note
templates: bump AWS CCM to v1.32.5 and add variable to control it
```

### DIFF
--- a/templates/cluster-template-flatcar-machinepool.yaml
+++ b/templates/cluster-template-flatcar-machinepool.yaml
@@ -252,7 +252,7 @@ data:
           serviceAccountName: cloud-controller-manager
           containers:
             - name: aws-cloud-controller-manager
-              image: registry.k8s.io/provider-aws/cloud-controller-manager:v1.28.3
+              image: registry.k8s.io/provider-aws/cloud-controller-manager:${KUBERNETES_AWS_CCM_VERSION:-v1.32.5}
               args:
                 - --v=2
                 - --cloud-provider=aws
--- a/templates/cluster-template-flatcar.yaml
+++ b/templates/cluster-template-flatcar.yaml
@@ -235,7 +235,7 @@ data:
           serviceAccountName: cloud-controller-manager
           containers:
             - name: aws-cloud-controller-manager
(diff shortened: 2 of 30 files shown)


damdo:
/hold

For e2e testing to finish

test/e2e/shared/suite.go
```diff
@@ -262,7 +262,14 @@ func AllNodesBeforeSuite(e2eCtx *E2EContext, data []byte) {
 				return
 			case <-e2eCtx.Environment.ResourceTicker.C:
 				for k := range e2eCtx.Environment.Namespaces {
-					DumpSpecResources(resourceCtx, e2eCtx, k)
+					func() {
```
clebs: What is the reason for wrapping this inside an anonymous function?

damdo: The anonymous function creates a new scope for the defer.
In Go, defer is bound to the enclosing function, not the enclosing block.

Without the wrapper, the defer would stack up and only execute when the goroutine's outer function returns, so the recover() would never fire when you actually need it.

clebs: Of course... Thanks for clarifying.

test/e2e/shared/suite.go
```diff
@@ -262,7 +262,12 @@ func AllNodesBeforeSuite(e2eCtx *E2EContext, data []byte) {
 				return
 			case <-e2eCtx.Environment.ResourceTicker.C:
 				for k := range e2eCtx.Environment.Namespaces {
-					DumpSpecResources(resourceCtx, e2eCtx, k)
+					// Warn instead of error when dumping resources fails due to a cluster being deleted.
```
nrb, line 270: Looks like this change is already merged into main at https://github.com/kubernetes-sigs/cluster-api-provider-aws/commit/30186658e2a369ede1c56051f194cd5c69fa94a4

PR #5885, opened 2026-03-03
⚠️ Move webhooks out of API packages

mdbooth:
**What type of PR is this?**

/kind cleanup

The primary purpose of this change is to reduce the impact of incompatible changes to controller-runtime on packages which import CAPA API packages. This change removes webhooks from the api package, and their associated c-r dependencies.

This change deliberately follows the webhook pattern used in the cluster-api repo. In general this means all webhooks move from api packages into a common webhooks package. Each API type which implements a webhook gets a corresponding type in the webhooks package which implements SetupWebhookWithManager. Internal callers are updated to use the new webhook types.

The impact on exported symbols is minimal. The api packages no longer export any SetupWebhookWithManager types. Additionally, the AWSMachineTemplateWebhook and AWSMachinePoolWebhook types have been removed from their corresponding api packages. Apart from that, there should be no changes in exported symbols.

Corresponding webhook tests have been moved to the webhooks package.

An explicit goal of this change is to minimise code change during code motion. Where possible, code has not been changed at all, just moved.

Fixes #

Technically a breaking change (hence the release note and the warning emoji), but in practise I don't expect any caller to be using any of the affected symbols.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:

```release-note
SetupWebhookWithManager methods have been removed from all api packages for all exported types. Additionally, the AWSMachineTemplateWebhook and AWSMachinePoolWebhook types are no longer exported. Callers importing CAPA as a Go library can no longer use these methods and types.
```


nrb:
/lgtm

This follows CAPI's lead, and should hopefully make future version bumps easier.

controlplane/rosa/webhooks/rosacontrolplane_webhook.go
```diff
@@ -29,13 +29,17 @@ import (
 	ctrl "sigs.k8s.io/controller-runtime"
 	"sigs.k8s.io/controller-runtime/pkg/webhook"
 	"sigs.k8s.io/controller-runtime/pkg/webhook/admission"
+
+	rosacontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/rosa/api/v1beta2"
 )
 
+// ROSAControlPlane implements a custom validation webhook for ROSAControlPlane.
+type ROSAControlPlane struct{}
```
serngawy, line 37: In general looks good to me. However, not sure, if removing the webhook from all webhook structs names is mandatory ? 
IMO better to keep the struct name with webhook eg; rosaControlPlaneWebhook not to confuse with the API struct ROSAControlPlane
```suggestion
type ROSAControlPlaneWebhook struct{}
```

mdbooth: This was my first thought too. However, I noticed that `cluster-api` hasn't done this, e.g. https://github.com/kubernetes-sigs/cluster-api/blob/bad0c7eb2467864bc3712786639cb8f4f5bb04e9/internal/webhooks/cluster.go#L72-L80

And on reflection I think this makes sense. If you look at where it's actually used, it's only where webhooks are initialised: `main.go` and test suites. And there the usage looks like:

```Go
	if err := (&capawebhooks.AWSClusterRoleIdentity{}).SetupWebhookWithManager(mgr); err != nil {
```

In that context, a `Webhook` suffix would stutter.

damdo:
It seems to me e2e's are failing on this PR https://prow.k8s.io/job-history/gs/kubernetes-ci-logs/pr-logs/directory/pull-cluster-api-provider-aws-e2e for the usual e2e troubles.
So I'd not worry about those.
Do you agree @nrb ?

controlplane/rosa/webhooks/rosacontrolplane_webhook.go
damdo: The pkgs is already called `webhooks` and Go's general recommendation is to avoid stuttering/repetition: https://go.dev/doc/effective_go#package-names

I think it makes sense to update it the way Matt is suggesting

damdo:
The two tests failing in the [e2e here](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5885/pull-cluster-api-provider-aws-e2e/2029591100276084736)

- capa-e2e: [It] [unmanaged] [Cluster API Framework] Clusterctl Upgrade Spec [from latest v1beta1 release to v1beta2] Should create a management cluster and then upgrade all the providers
  - https://kubernetes.slack.com/archives/CD6U2V71N/p1771401351581489?thread_ts=1770134623.098539&cid=CD6U2V71N 

- capa-e2e: [It] [unmanaged] [functional] GPU-enabled cluster test should create cluster with single worker:
  - https://kubernetes.slack.com/archives/CD6U2V71N/p1771083063899809?thread_ts=1770134623.098539&cid=CD6U2V71N 

Are both known and unrelated issues.

I'm inclined to unhold this

PR #5900, opened 2026-03-11
🌱 ci: bump image-builder to include MachinePool cloud-init fix

damdo:
**What type of PR is this?**

/kind support

The v1.33/v1.34 CAPA AMIs were built with image-builder v0.1.48, which includes a custom cloud-init datasource (DataSourceEc2Kubernetes) that unconditionally reads from /etc/secret-userdata.txt. This file is only created by the SecretsManager boothook for individual Machine nodes (control plane, MachineDeployment), but not for MachinePool/ASG nodes where bootstrap data is passed directly as EC2 user-data. The missing file causes a FileNotFoundError, crashing cloud-init and preventing MachinePool worker nodes from joining the cluster.

Pin the image-builder version to commit 990737c (post-v0.1.48) which adds a graceful fallback: when /etc/secret-userdata.txt is absent, the datasource uses the original EC2 metadata user-data instead.

The affected AMIs (ubuntu v1.33.4 and v1.34.3) need to be rebuilt with this version for the fix to take effect.

Ref: https://github.com/kubernetes-sigs/image-builder/commit/990737c

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/build-ami-varsfile.yml
+++ b/.github/workflows/build-ami-varsfile.yml
@@ -6,7 +6,7 @@ on:
       image_builder_version:
         description: "Image builder version"
         required: true
-        default: 'v0.1.48'
+        default: '990737c79abb37fd56faca64f46e5a385dd1a982' # post-v0.1.48, includes fix https://github.com/kubernetes-sigs/image-builder/pull/1919
       target:
         description: "target os"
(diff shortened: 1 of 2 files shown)


damdo (commented):
/approve

PR #5906, opened 2026-03-16
✨ Add ExistingLaunchTemplate (BYO) support for EKS managed node groups

### LINKED ISSUE #5896: Support BYO (Bring Your Own) launch template for EKS managed node groups
/kind feature

EKS allows creating a node group with an existing launch template (by ID + version). CAPA today only supports launch templates that it creates and manages; there is no way to reference a pre-existing launch template for use with an EKS managed node group.

**Proposed behavior:** Add an `id` field to the existing `AWSLaunchTemplate` struct. When `awsLaunchTemplate.id` is set on an `AWSManagedMachinePool`, CAPA treats the launch template as BYO (Bring Your Own):

- The ID is passed directly to the EKS `CreateNodegroup`/`UpdateNodegroupVersion` API
- CAPA does **not** create, update, or delete the launch template
- `awsLaunchTemplate.versionNumber` can be set to pin a specific version, and is mutable to allow rolling out new versions
- On pool deletion, only the node group is deleted; the launch template is left unchanged
- CAPA-managed fields (AMI, instanceType, rootVolume, sshKeyName, imageLookup*, etc.) are rejected by webhook validation when `id` is set, since the user manages the template contents externally

*Use case:* Users who already have a launch template (e.g. from another tool, console, or Terraform) want to use it with CAPA-managed node groups without CAPA creating or modifying that template.

*Current behavior:* When `spec.awsLaunchTemplate` is set, the controller always creates and manages a launch template by a derived name (`{control-plane-name}-{pool-name}`). There is no way to reference an existing launch template by ID.

*Example spec:*
```yaml
spec:
  awsLaunchTemplate:
    id: "lt-0abcd1234efgh5678"
    versionNumber: 3
```

AmitSahastra:
**What type of PR is this?**

/kind feature

Adds BYO (bring-your-own) launch template support for EKS managed node groups by introducing an `id` field on `AWSLaunchTemplate` (in both `v1beta1` and `v1beta2`). When `spec.awsLaunchTemplate.id` is set, CAPA references the pre-existing AWS launch template directly instead of creating or managing one.

This enables use cases where organizations manage launch templates externally (Terraform, CloudFormation, the AWS console, etc.) and want CAPA to adopt them as-is for EKS managed node groups.

Highlights:
- **API**: new `id` field and immutable `name` field on `AWSLaunchTemplate` (v1beta1 + v1beta2), with CEL `self == oldSelf` immutability markers and regenerated CRDs/conversions.
- **Controller**: when `id` is set, the reconciler skips LT create/update/delete and uses the referenced template directly; status (`launchTemplateID` / `launchTemplateVersion`) is populated from the spec. CAPA-managed (no-id) behaviour is unchanged.
- **Webhook**: BYO-specific gates — `versionNumber` required; LT-level fields that only make sense for CAPA-managed LTs (`ami`, `instanceType`, `rootVolume`, `nonRootVolumes`, `sshKeyName`, image-lookup fields, `additionalSecurityGroups`) are forbidden; `id`/`name` immutability. Plus cross-cutting gates aligned with the AWS `CreateNodegroup` contract: `spec.instanceType` and `spec.amiType` are allowed alongside a BYO LT, while `spec.diskSize` and `spec.remoteAccess` are rejected whenever any LT is specified.
- **Nodegroup service**: BYO-LT-aware `CreateNodegroup` path — `InstanceTypes` and `AmiType` flow through for BYO LTs, `RemoteAccess` and `DiskSize` are omitted when a LT is in use; symbolic LT versions (`$Latest`, `$Default`) are handled to avoid endless reconcile loops; guards added for nil `Version`/`ReleaseVersion` on failed node groups.

**Which issue(s) this PR fixes** *(optional, in `fixes # (, fixes #, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #5896

- `VersionNumber` is intentionally mutable on `AWSLaunchTemplate` so users can roll out a new LT version to the node group; `id` and `name` are immutable (enforced via CEL and the webhook).
- For BYO LTs, CAPA cannot introspect the referenced template, so conflicts between node-group-level fields (e.g. `amiType`, `amiVersion`) and a LT that already pins a custom AMI are surfaced by the EKS API at create time rather than at admission.
- How to exercise locally:
  1. Create an EC2 launch template in AWS.
  2. Reference it on an `AWSManagedMachinePool`:
     ```yaml
     spec:
       awsLaunchTemplate:
         id: lt-xxxxxxxxxxxx
         versionNumber: 1
     ```
  3. The node group is created using the referenced launch template; CAPA does not delete the LT on pool deletion.

**AI Usage**:

AI assistance (Warp Agent Mode) was used for drafting webhook validation and writing unit tests. All generated code was reviewed, tested, and signed off by the author.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
Add BYO launch template support for EKS managed node groups via the new `spec.awsLaunchTemplate.id` field on `AWSManagedMachinePool`. When set, CAPA uses the referenced pre-existing AWS launch template directly and does not create, update, or delete it.
```


AmitSahastra:
Hi @fiunchinho  @serngawy the bot suggested I consider joining the kubernetes-sigs org given my contributions. Would either of you be willing to sponsor me for membership? I'm happy to open the formal request at https://github.com/kubernetes/org/issues/new?template=membership.md once I have two sponsors confirmed. Thanks!

exp/api/v1beta1/types.go
```diff
@@ -62,6 +62,11 @@ type BlockDeviceMapping struct {
 
 // AWSLaunchTemplate defines the desired state of AWSLaunchTemplate.
 type AWSLaunchTemplate struct {
+	// ID is the ID of an existing launch template (e.g. lt-xxxx). When set, CAPA will
+	// not create or manage the launch template and will use the referenced one directly.
+	// +optional
+	ID *string `json:"id,omitempty"`
```
serngawy: I think this one need to be immutable, What will happen if the user create the template without ID then set the ID ?

exp/api/v1beta2/types.go
```diff
@@ -61,6 +61,11 @@ type BlockDeviceMapping struct {
 
 // AWSLaunchTemplate defines the desired state of AWSLaunchTemplate.
 type AWSLaunchTemplate struct {
+	// ID is the ID of an existing launch template (e.g. lt-xxxx). When set, CAPA will
+	// not create or manage the launch template and will use the referenced one directly.
+	// +optional
+	ID *string `json:"id,omitempty"`
```
serngawy, line 69: same

exp/api/v1beta1/types.go
AmitSahastra: Good catch, thanks! Fixed in the latest commit, added both a CEL validation marker (`self == oldSelf`) on the `ID` and `Name` fields in v1beta1 and v1beta2, and a dedicated `validateLaunchTemplateImmutability` function in the webhook for belt-and-suspenders coverage. The CRDs are regenerated too so the `x-kubernetes-validations` rules show up there as well. `VersionNumber` is intentionally left mutable since that's the knob users turn to roll out a new LT version to the nodegroup.

AmitSahastra: Also added test case around it

exp/api/v1beta2/types.go
AmitSahastra: fixed

exp/webhooks/awsmanagedmachinepool_webhook.go
```diff
@@ -138,8 +138,46 @@ func (w *AWSManagedMachinePool) validateLaunchTemplate(r *expinfrav1.AWSManagedM
...
+	if lt.ID != nil && *lt.ID != "" {
+		if lt.VersionNumber == nil {
+			allErrs = append(allErrs, field.Required(ltPath.Child("versionNumber"), "versionNumber is required when using a BYO launch template (id is set)"))
+		}
+		if lt.AMI.ID != nil || lt.AMI.EKSOptimizedLookupType != nil {
+			allErrs = append(allErrs, field.Forbidden(ltPath.Child("ami"), "ami cannot be specified with a BYO launch template (id is set)"))
+		}
+		if lt.InstanceType != "" {
+			allErrs = append(allErrs, field.Forbidden(ltPath.Child("instanceType"), "instanceType cannot be specified with a BYO launch template (id is set)"))
+		}
```
cPu1, line 164: The `CreateNodegroup` API allows setting `InstanceTypes` if the supplied launch template does not have instance type set, but the validation above in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5906/changes#diff-e8d34ee221031d69c3d59088b98d174901f49ca8329dc82b88738bafa008741dL134 disallows usage of a pre-existing launch template that does not have an instance type set. CAPA should allow specifying both `spec.instanceType` and `spec.launchTemplate.id` if the launch template lacks an instance type. Without this, pre-existing launch templates that lack an instance type cannot be adopted.

AmitSahastra: Good catch, fixed. `spec.instanceType` is now allowed alongside `spec.awsLaunchTemplate.id`, it gets passed as InstanceTypes in CreateNodegroup. The restriction is kept for CAPA-managed LTs (no id) where instance type must be set inside the template.

exp/webhooks/awsmanagedmachinepool_webhook.go
```diff
@@ -178,6 +183,11 @@ func (w *AWSManagedMachinePool) validateLaunchTemplate(r *expinfrav1.AWSManagedM
 		if len(lt.AdditionalSecurityGroups) > 0 {
 			allErrs = append(allErrs, field.Forbidden(ltPath.Child("additionalSecurityGroups"), "additionalSecurityGroups cannot be specified with a BYO launch template (id is set)"))
 		}
+		// amiType is silently ignored at the service layer for BYO LTs; the AMI type is
+		// determined entirely by the referenced launch template.
+		if r.Spec.AMIType != nil {
+			allErrs = append(allErrs, field.Forbidden(field.NewPath("spec", "amiType"), "amiType cannot be specified with a BYO launch template; the AMI type is determined by the launch template"))
+		}
```
cPu1: Like `instanceTypes`, `amiType` is another field that can be set alongside a launch template ID if the launch template does not specify an AMI type.

AmitSahastra: fixed

AmitSahastra:
> /test pull-cluster-api-provider-aws-e2e-eks /test pull-cluster-api-provider-aws-e2e

Friendly ping @fiunchinho @serngawy @cPu1 ok-to-test is set and all CI checks are green. Could you take another look when you have a chance? Previous review feedback has been addressed. Thanks! 🙏

exp/api/v1beta1/conversion.go
```diff
@@ -50,6 +50,10 @@ func (src *AWSMachinePool) ConvertTo(dstRaw conversion.Hub) error {
 	if restored.Spec.AWSLaunchTemplate.EnclaveOptions != nil {
 		dst.Spec.AWSLaunchTemplate.EnclaveOptions = restored.Spec.AWSLaunchTemplate.EnclaveOptions
 	}
+	// ID is a v1beta2-only field; restore it from the annotation.
+	if restored.Spec.AWSLaunchTemplate.ID != nil {
```
richardcase: We can probably just remove the `if`.

richardcase: That's a nit and you are welcome to ignore :)

exp/api/v1beta1/conversion.go
```diff
@@ -121,6 +125,10 @@ func (src *AWSManagedMachinePool) ConvertTo(dstRaw conversion.Hub) error {
 		if dst.Spec.AWSLaunchTemplate == nil {
 			dst.Spec.AWSLaunchTemplate = restored.Spec.AWSLaunchTemplate
 		}
+		// ID is a v1beta2-only field (BYO launch template); restore it from the annotation.
+		if restored.Spec.AWSLaunchTemplate.ID != nil {
```
richardcase: Same comment as above

exp/api/v1beta2/types.go
```diff
@@ -61,7 +61,16 @@ type BlockDeviceMapping struct {
...
 type AWSLaunchTemplate struct {
+	// ID is the ID of an existing launch template (e.g. lt-xxxx). When set, CAPA will
+	// not create or manage the launch template and will use the referenced one directly.
+	// +optional
+	// +immutable
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ID is immutable"
+	ID *string `json:"id,omitempty"`
+
 	// The name of the launch template.
+	// +immutable
```
richardcase: This is a change of behaviour for this field. We are changing an existing field which is mutable to immutable and this is considered a breaking change according to the Kubernetes API conventions: https://github.com/kubernetes/community/blob/main/contributors/devel/sig-architecture/api_changes.md#on-compatibility

We'll need to remove these new field validation markers.

If we want to make this change then it will have to be done as part of an API version bump to v1beta3 (see #3853 )

exp/controllers/awsmanagedmachinepool_controller.go
```diff
@@ -279,7 +288,8 @@ func (r *AWSManagedMachinePoolReconciler) reconcileDelete(
 		return errors.Wrapf(err, "failed to reconcile machine pool deletion for AWSManagedMachinePool %s/%s", machinePoolScope.ManagedMachinePool.Namespace, machinePoolScope.ManagedMachinePool.Name)
 	}
 
-	if machinePoolScope.ManagedMachinePool.Spec.AWSLaunchTemplate != nil {
+	// Only delete the launch template when it is CAPA-managed (not BYO).
+	if machinePoolScope.ManagedMachinePool.Spec.AWSLaunchTemplate != nil && !machinePoolScope.IsBYOLaunchTemplate() {
```
richardcase: Could the `machinePoolScope.ManagedMachinePool.Spec.AWSLaunchTemplate != nil` be incorporated into the `IsBYOLaunchTemplate` function?

exp/webhooks/awsmanagedmachinepool_webhook.go
```diff
@@ -131,15 +131,69 @@ func (w *AWSManagedMachinePool) validateLaunchTemplate(r *expinfrav1.AWSManagedM
...
+	if lt.IamInstanceProfile != "" {
+		allErrs = append(allErrs, field.Invalid(ltPath.Child("iamInstanceProfile"), lt.IamInstanceProfile, "IAM instance profile in launch template is prohibited in EKS managed node group"))
 	}
 
-	if r.Spec.AWSLaunchTemplate.IamInstanceProfile != "" {
-		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "AWSLaunchTemplate", "IamInstanceProfile"), r.Spec.AWSLaunchTemplate.IamInstanceProfile, "IAM instance profile in launch template is prohibited in EKS managed node group"))
+	// When using a BYO launch template (ID is set), versionNumber is required and
+	// CAPA-managed fields must not be specified.
+	if isBYO {
+		if lt.VersionNumber == nil {
```
richardcase: Might be nice to extra all these checks into a separate function.

exp/webhooks/awsmanagedmachinepool_webhook.go
```diff
@@ -131,15 +131,69 @@ func (w *AWSManagedMachinePool) validateLaunchTemplate(r *expinfrav1.AWSManagedM
...
+			allErrs = append(allErrs, field.Required(ltPath.Child("versionNumber"), "versionNumber is required when using a BYO launch template (id is set)"))
+		}
+		if lt.AMI.ID != nil || lt.AMI.EKSOptimizedLookupType != nil {
```
richardcase: Theres 2 fields being checked here but only 1 error type. We should probably have 2 checks and 2 errors.

exp/webhooks/awsmanagedmachinepool_webhook.go
```diff
@@ -278,8 +332,31 @@ func (w *AWSManagedMachinePool) validateImmutable(r *expinfrav1.AWSManagedMachin
...
+	if oldLT == nil || newLT == nil {
+		return allErrs
+	}
+
+	ltPath := field.NewPath("spec", "awsLaunchTemplate")
+
+	if !reflect.DeepEqual(oldLT.ID, newLT.ID) {
+		allErrs = append(allErrs, field.Forbidden(ltPath.Child("id"), "id is immutable"))
+	}
+	if !reflect.DeepEqual(oldLT.Name, newLT.Name) {
```
richardcase: This check on `Name` will need to be removed as its changing the behaviour of the field (see previous comment)

exp/webhooks/awsmanagedmachinepool_webhook.go
```diff
@@ -278,8 +332,31 @@ func (w *AWSManagedMachinePool) validateImmutable(r *expinfrav1.AWSManagedMachin
...
+	oldLT := old.Spec.AWSLaunchTemplate
+	newLT := r.Spec.AWSLaunchTemplate
+
+	if oldLT == nil || newLT == nil {
+		return allErrs
+	}
+
+	ltPath := field.NewPath("spec", "awsLaunchTemplate")
+
+	if !reflect.DeepEqual(oldLT.ID, newLT.ID) {
```
richardcase: We generally use `cmp.Equals` instead of the reflect.DeepEqual.

pkg/cloud/scope/managednodegroup.go
```diff
@@ -413,13 +413,26 @@ func (s *ManagedMachinePoolScope) GetLaunchTemplate() *expinfrav1.AWSLaunchTempl
 	return s.ManagedMachinePool.Spec.AWSLaunchTemplate
 }
 
+// IsBYOLaunchTemplate returns true when the pool uses an existing (user-provided) launch template.
+func (s *ManagedMachinePoolScope) IsBYOLaunchTemplate() bool {
+	lt := s.ManagedMachinePool.Spec.AWSLaunchTemplate
```
richardcase, line 419: I wonder if this would work and be clearer?

```go
ltID := ptr.Deref(s.ManagedMachinePool.Spec.AWSLaunchTemplate.ID, "")
return ltID != ""
```

(not tested it myself though)

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -222,16 +222,32 @@ func (s *NodegroupService) createNodegroup(ctx context.Context) (*ekstypes.Nodeg
 		NodeRole:      roleArn,
 		Labels:        managedPool.Labels,
 		Tags:          tags,
-		RemoteAccess:  remoteAccess,
 		UpdateConfig:  updatedConfig,
 	}
-	if managedPool.AMIType != nil && (managedPool.AWSLaunchTemplate == nil || managedPool.AWSLaunchTemplate.AMI.ID == nil) {
-		input.AmiType = converters.AMITypeToSDK(*managedPool.AMIType)
+	useLaunchTemplate := managedPool.AWSLaunchTemplate != nil
+	isBYO := useLaunchTemplate && managedPool.AWSLaunchTemplate.ID != nil && *managedPool.AWSLaunchTemplate.ID != ""
```
richardcase: Could this be on the scope instead?

exp/api/v1beta1/conversion.go
AmitSahastra: Done, removed the redundant if guards and assign ID directly.

exp/api/v1beta1/conversion.go
AmitSahastra: Done, removed the redundant if guards and assign ID directly.

exp/api/v1beta2/types.go
AmitSahastra: Good catch, thanks. Removed the +immutable and XValidation:self == oldSelf markers from AWSLaunchTemplate.Name and regenerated the CRDs.

exp/webhooks/awsmanagedmachinepool_webhook.go
AmitSahastra: Done, dropped the reflect.DeepEqual helper and folded the checks back into validateImmutable using the existing appendErrorIfMutated, which uses cmp.Equal.

exp/webhooks/awsmanagedmachinepool_webhook.go
AmitSahastra: Fixed, now it split into two separate checks/errors: ami.id and ami.eksLookupType.

exp/webhooks/awsmanagedmachinepool_webhook.go
AmitSahastra: moved the BYO-specific validations into validateBYOLaunchTemplate(lt, ltPath).

exp/webhooks/awsmanagedmachinepool_webhook.go
AmitSahastra: fixed

pkg/cloud/scope/managednodegroup.go
AmitSahastra: Done, with a nil guard kept since Spec.AWSLaunchTemplate can be nil: return lt != nil && ptr.Deref(lt.ID, "") != "".

pkg/cloud/services/eks/nodegroup.go
AmitSahastra: done

exp/controllers/awsmanagedmachinepool_controller.go
AmitSahastra: done

test/e2e/data/eks/cluster-template-eks-managed-machinepool-with-byo-launch-template-only.yaml
```diff
@@ -0,0 +1,32 @@
...
+kind: AWSManagedMachinePool
+metadata:
+  name: "${CLUSTER_NAME}-pool-byo-lt-0"
+spec:
+  eksNodegroupName: "${CLUSTER_NAME}-pool-byo-lt-0"
+  # instanceType must not be set when using a BYO launch template — EKS uses
+  # the type configured inside the launch template itself.
+  awsLaunchTemplate:
+    id: "${BYO_LAUNCH_TEMPLATE_ID}"
+    versionNumber: ${BYO_LAUNCH_TEMPLATE_VERSION}
```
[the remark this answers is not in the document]
AmitSahastra: The flavor now sets instanceType: `${AWS_NODE_MACHINE_TYPE}` and `amiType: AL2023_x86_64_STANDARD`, and the misleading comment is updated. This also resolves the `pull-cluster-api-provider-aws-e2e-eks failure`.

exp/api/v1beta2/types.go
```diff
@@ -61,6 +61,13 @@ type BlockDeviceMapping struct {
...
 type AWSLaunchTemplate struct {
+	// ID is the ID of an existing launch template (e.g. lt-xxxx). When set, CAPA will
+	// not create or manage the launch template and will use the referenced one directly.
+	// +optional
+	// +immutable
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf",message="ID is immutable"
+	ID *string `json:"id,omitempty"`
+
 	// The name of the launch template.
 	Name string `json:"name,omitempty"`
```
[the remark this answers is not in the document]
AmitSahastra: Fixed this in commit d64083388,

exp/webhooks/awsmanagedmachinepool_webhook.go
```diff
@@ -130,16 +130,84 @@ func (w *AWSManagedMachinePool) validateLaunchTemplate(r *expinfrav1.AWSManagedM
 		return allErrs
 	}
 
-	if r.Spec.InstanceType != nil {
-		allErrs = append(allErrs, field.Invalid(field.NewPath("spec", "InstanceType"), r.Spec.InstanceType, "InstanceType cannot be specified when LaunchTemplate is specified"))
+	lt := r.Spec.AWSLaunchTemplate
+	ltPath := field.NewPath("spec", "awsLaunchTemplate")
+	isBYO := lt.ID != nil && *lt.ID != ""
```
richardcase: Could potentially do something like this:
```suggestion
	isBYO := ptr.Deref(lt.ID, "") != ""
```

pkg/cloud/services/eks/nodegroup.go
```diff
@@ -317,6 +333,33 @@ func (s *NodegroupService) deleteNodegroupAndWait(ctx context.Context) (reterr e
 	return nil
 }
 
+// isSymbolicLaunchTemplateVersion returns true for AWS symbolic version aliases
+// ("$Latest", "$Default") that resolve to a concrete version number at apply time.
+// When comparing a symbolic spec version against the nodegroup's resolved version,
+// we skip the comparison to avoid an endless update loop.
+func isSymbolicLaunchTemplateVersion(v string) bool {
+	return v == "$Latest" || v == "$Default"
```
richardcase: Could we just check for a `$` prefix?

richardcase (commented):
Generally look good @AmitSahastra .

The next statement isn't a reflection on this change but how we handle launch templates is quite convoluted with a number of different options. It feels like in a future API version we should revisit launch templates and how we specify them.

pkg/cloud/services/eks/nodegroup.go
AmitSahastra: Good idea, thanks — done. Simplified to `strings.HasPrefix(v, "$")`, which covers both current aliases (`$Latest`, `$Default`) and is safer for any future/unrecognised `$`-prefixed value (treating it as symbolic avoids a doomed, endless update loop rather than misreading it as a mismatched concrete version). Updated the unit test accordingly.

richardcase:
@AmitSahastra - thanks for this. Lets check that the e2e tests pass.

Also, can you squash your commits? It doesn't have to be down to 1 commit but if there are multiple commits they need to tell a story.  A smaller number of commits is always good though.

AmitSahastra:
> @AmitSahastra - thanks for this. Lets check that the e2e tests pass.
> 
> Also, can you squash your commits? It doesn't have to be down to 1 commit but if there are multiple commits they need to tell a story. A smaller number of commits is always good though.

@richardcase I squashed all commits and fixed the commit msg.

AmitSahastra:
@richardcase tests are passing now. Had to remove the eksNodegroupName line from test/e2e/data/eks/cluster-template-eks-managed-machinepool-with-byo-launch-template-only.yaml so that it matches new auto genrated name.

PR #5928, opened 2026-04-02
✨ Add AWS Nitro Enclave support to AWSLaunchTemplate

### LINKED ISSUE #5927: Add AWS Nitro Enclave support to AWSLaunchTemplate
/kind feature

  `AWSLaunchTemplate` does not expose `EnclaveOptions`, which means CAPA users
  cannot enable AWS Nitro Enclaves on managed machine pool nodes. The field is
  available in the EC2 launch template API (`LaunchTemplateEnclaveOptionsRequest`)
  but has never been wired into the CAPA types or launch template builder.

  The fix appears straightforward: add an `EnclaveOptions` struct to `api/v1beta2`
  and a corresponding field on `AWSLaunchTemplate`, then map it through
  `createLaunchTemplateData()`, `SDKToLaunchTemplate()`, and
  `LaunchTemplateNeedsUpdate()` — following the same pattern as
  `InstanceMetadataOptions`.

  This was called out as a motivating example in the original launch template
  proposal (docs/proposal/20211210-launch-templates-managedmachinepools.md).

  I can work on a PR with the implementation, tests, and documentation.

  **Environment:**

  - Cluster-api-provider-aws version: main
  - Kubernetes version: N/A
  - OS: N/A

raykrueger:
**What type of PR is this?**

/kind api-change

`AWSLaunchTemplate` does not expose `EnclaveOptions`, preventing CAPA users from enabling AWS Nitro Enclaves on managed machine pool nodes. The field is available in the EC2 launch template API but was never wired into the CAPA types or launch template builder.

This adds an `EnclaveOptions` struct to `api/v1beta2` and a corresponding field on `AWSLaunchTemplate`, mapped through `createLaunchTemplateData()`, `SDKToLaunchTemplate()`, and `LaunchTemplateNeedsUpdate()` — following the same pattern as `InstanceMetadataOptions`. Nitro Enclaves were called out as a motivating example in the original launch template proposal (`docs/proposal/20211210-launch-templates-managedmachinepools.md`).

Fixes #5927

`EnclaveOptions` is a v1beta2-only field. The annotation-based restore pattern in `exp/api/v1beta1/conversion.go` has been updated for both `AWSMachinePool` and `AWSManagedMachinePool`, following the same approach as `InstanceMetadataOptions`, `PrivateDNSName`, and other v1beta2-only `AWSLaunchTemplate` fields.

This PR was developed with AI and my direct supervision.

**Checklist**:
- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:
```release-note
Add `enclaveOptions` field to `AWSLaunchTemplate` to enable AWS Nitro Enclave support on EC2 instances via managed machine pools.
```


richardcase (commented):
This looks good to me.

Do you think you'd be able to add a new e2e test to cover this option?

raykrueger:
Is it a must have? It looks like a good chunk of work and I don't have the means to run it myself. Since we're just setting a field to "true" on a LaunchTemplate are we testing AWS at that point?

Also, as a heads up we can't use the t3 instances, c5.xlarge is the cheapest we can get away with it looks like.

raykrueger:
Despite my respectful push back, I am working on this :)

raykrueger:
I've added an e2e test for the feature here. I've left it as a separate commit for it's own review. I can squash them at your call :)

raykrueger:
Disregard, I've squashed and cleaned up the commit.

raykrueger:
I should add the obvious... The new E2E test passes. I was able to run the other machine pool test as well as a regression check.

richardcase:
> Despite my respectful push back

Sorry @raykrueger i hadn't seen your response. I'm more than happy to get push back on things. Feel free to also raise via slack.

raykrueger:
Looks like I need to rebase this one too?

test/e2e/data/infrastructure-aws/withoutclusterclass/kustomize_sources/nitro-enclave/patches/limit-az.yaml
```diff
@@ -0,0 +1,9 @@
+---
+apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
+kind: AWSCluster
+metadata:
+  name: "${CLUSTER_NAME}"
+spec:
+  network:
+    vpc:
+      availabilityZoneUsageLimit: 1
```
AndiDog, line 12: Is a single AZ needed?

test/e2e/suites/unmanaged/helpers_test.go
```diff
@@ -389,6 +389,44 @@ func terminateInstance(instanceID string) {
 	Expect(utils.ToInt64Value(result.TerminatingInstances[0].CurrentState.Code)).To(Equal(termCode))
 }
 
+func assertNitroEnclaveEnabled(ctx context.Context, clusterName string) {
+	ginkgo.By(fmt.Sprintf("Verifying Nitro Enclave enabled on instances in cluster: %s", clusterName))
+	ec2Client := ec2.NewFromConfig(*e2eCtx.AWSSession)
+	// Filter only worker (node) instances — control plane uses AWSMachineTemplate, not AWSLaunchTemplate,
```
AndiDog, line 395: With this, we unfortunately only get supported for machine pools. Does it make sense to restrict it like that, and not offer it for control plane nodes or `AWSMachineDeployment`?

api/v1beta2/types.go
```diff
@@ -405,6 +405,13 @@ const (
 	PrimaryIPv6AssignmentStateDisabled = PrimaryIPv6AssignmentState("disabled")
 )
 
+// EnclaveOptions defines the options for Nitro Enclave support on the instance.
+type EnclaveOptions struct {
+	// Enabled enables the instance for AWS Nitro Enclaves.
+	// +optional
+	Enabled bool `json:"enabled,omitempty"`
```
AndiDog: @richardcase should we use `*bool` here?

raykrueger: I think you're right, I can work on that change and revert if @richardcase says we're wrong :)

test/e2e/suites/unmanaged/helpers_test.go
raykrueger: I don't think it's restricted, is two different paths and two different implementations. This PR is enabling LaunchTemplate support. The Machine Deployment uses RunInstances directly. I think a separate PR can work on that.

test/e2e/data/infrastructure-aws/withoutclusterclass/kustomize_sources/nitro-enclave/patches/limit-az.yaml
raykrueger: This is just for the e2e tests, a single AZ keeps NAT Gateway costs down.

raykrueger: Which was my compensation for the more expensive instance type required :)

api/v1beta2/types.go
raykrueger: This also includes a bunch of `&infrav1.EnclaveOptions{Enabled: aws.Bool(true)}` changes.

docs/book/src/topics/nitro-enclaves.md
```diff
@@ -0,0 +1,32 @@
...
+Nitro Enclaves are enabled per-instance via the `enclaveOptions` field on the launch template.
+
+## Prerequisites
+
+- The instance type must support Nitro Enclaves. See the [AWS documentation](https://docs.aws.amazon.com/enclaves/latest/user/nitro-enclave.html#nitro-enclave-reqs) for supported instance types.
+- The instance must use a Nitro-based AMI.
+
+## Configuration
+
+Nitro Enclave support is configured via the `enclaveOptions` field in `AWSManagedMachinePool.spec.awsLaunchTemplate`:
```
richardcase: It also applies to `AWSMachinePool` as it's on the AWS launch template options.

exp/api/v1beta2/types.go
```diff
@@ -130,6 +130,10 @@ type AWSLaunchTemplate struct {
 	// +optional
 	InstanceMetadataOptions *infrav1.InstanceMetadataOptions `json:"instanceMetadataOptions,omitempty"`
 
+	// EnclaveOptions defines the options for Nitro Enclave support on the instance.
+	// +optional
+	EnclaveOptions *infrav1.EnclaveOptions `json:"enclaveOptions,omitempty"`
```
richardcase, line 135: We should also add it to `AWSMachine` in the future as well so that enclave support can also be used with MachineDeployment. But thats a different PR for sure 😄

api/v1beta2/types.go
```diff
@@ -405,6 +405,13 @@ const (
 	PrimaryIPv6AssignmentStateDisabled = PrimaryIPv6AssignmentState("disabled")
 )
 
+// EnclaveOptions defines the options for Nitro Enclave support on the instance.
+type EnclaveOptions struct {
```
richardcase, line 409: How likely do you think it is that there will be other options for enclave support?

raykrueger: Hard to say, but this mirrors the API for it, so if AWS adds stuff later we're fine.

raykrueger:
/retest

This looks really unrelated to this PR.

AndiDog (commented):
/lgtm

raykrueger:
I think this is because EKS 1.32 slipped into Extended Support on March 23rd, and the test thinks it's starting with Standard support.

I ran this in my environment and Claude found the following, which makes sense given the expiration dates.

```
Clear root cause:

  InvalidParameterException: The requested Kubernetes version "1.32" is only supported
  under extended support. Please change the supportType to extended support to create the cluster.

  K8s 1.32 requires extended support in this AWS account — but the test is deploying with STANDARD support policy first (that's the whole point of the test: start STANDARD, then switch to EXTENDED). AWS is rejecting the cluster creation outright.

```

api/v1beta2/types.go
JoelSpeed, line 409: How do we configure the link between an enclave instance and its parent instance? Or is this creating the parent instance?

Are we validating that hibernation cannot be enabled at the same time this option is configured?

Do we support local or wavelength zones? If so, how do we prevent this option being configured in conjunction with those zones?

raykrueger: 1) In the scope of the larger pull request, the enclave options struct is pulled into a LaunchTemplate. It's never used directly against an instance.

2 & 3) I had not considered these limitations. How do we handle this? Do we prevent the user doing these things, or do we rely on them to understand AWS current feature positioning. Following our discussion on future expansion, AWS does not support nitro enclaves in wavelength or local zones *today* and AWS does not support hibernation with enclaves *today*. if we add code stopping them from doing something AWS has documented, we may have to just remove it again later.

Thoughts?

raykrueger: I have logic to reject a Enclave Enabled LaunchTemplate that targets WaveLength or LocalZones in the AWSMachinePoolReconciler.

We have no logic to enable Hibernation via LaunchTemplates today, so that's not a scenario I can defend against.

raykrueger:
Same test (EKS upgrade policy test) fails for the same reason in PR #5992 which is unrelated to the work here.

richardcase:
As the failure is unrelated i'm happy to:

/override pull-cluster-api-provider-aws-e2e-eks

damdo:
/hold

We are adding a new e2e test that has not been exercised in CI. Let's try it out.

On this one, @raykrueger @richardcase as I pointed out during the call, I think we need to understand if the added time and cost of running this new extra e2e test for a simple field being plumbed exceeds the benefit of the signal it gives.

richardcase:
> I think we need to understand if the added time and cost of running this new extra e2e test for a simple field being plumbed exceeds the benefit of the signal it gives.

For me personally, enclaves are an important feature for CAPA for secure compute and as such it would be good to be able to test the feature. I do get the cost/benefit argument though. Would a compromise be that we add a new grouping of e2e tests, lets call it "extended", and these are only run 1 time per week (or on-demand). So we end up with 3 groups of e2e tests:

- PR blocking (quick)
- Normal e2e (eks and non-eks)
- Extended

@damdo @raykrueger - if you'd be happy with this compromise then how about we do this:

- @raykrueger if you change your test definition to: `ginkgo.PDescribe("Workload cluster with Nitro Enclave enabled", func() {` to mark it as a pending test for the time being
- When this PR merges i will then setup the "extended" suit and unmark the test
- I will create a page for contributors with a decision matrix on which group tests should live in

damdo:
> PR blocking (quick)
Normal e2e (eks and non-eks)
Extended

Hey @richardcase thanks for the proposal, this is a great idea! And something we also do downstream :)
I'm 💯 in favour of this.

> I will create a page for contributors with a decision matrix on which group tests should live in

Nice! I think this will be super useful!

api/v1beta2/types.go
JoelSpeed: I'm not sure if 1 really answers what I was asking. There's some element of enclave on AWS side where it has an parent child relationship right? Do you have to create an instance to be the parent, and then create children of it inside the enclave? 

Can we document the limitations in the godoc on the field so that users have some idea about the validations being imposed?

raykrueger:
I have updated the documentation and added some significant new e2e tests that test out what's documented. They are currently **not** pending. I will run them here and when they succeed I'll make them Pending and squash everything down.

AndiDog (commented):
/lgtm

/hold
because test still fails

raykrueger:
It seems I've fallen victim to a new infrastructure thing that isn't my fault?
https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5928/pull-cluster-api-provider-aws-build-docker/2051746383563591680

richardcase:
This should pass after a rebase, thanks to your other PR 🙇

raykrueger:
@richardcase The new managed pool nitro test passed and has been moved to pending. This is thoroughly documented and tested. I think we're done here :)

Special note: My new managed test, which tests what is in the nitro book, is overriding things to be AL2023 and Kube 1.35. This is to ensure the book is current and accurate, and is being tested. I can come back around and remove those when Andreas finishes his work. In any case, both tests are pending right now.

richardcase:
Lets get this merged as soon as the checks pass. So:

/unhold

richardcase:
@AndiDog  could you lgtm again?

PR #5857, opened 2026-02-03
🌱 Bump CAPI to v1.12.2, k8s to v1.34 and controller-runtime to v0.22.5

clebs:
**What type of PR is this?**
/kind support

**What this PR does / why we need it**:
This PR bumps CAPI to v1.12.2 and with it, k8s to v1.34.3 and controller-runtime to v0.22.5.
Needed to be up to date with CAPI releases.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #5834 

**Special notes for your reviewer**:
None

**TODO**
This bump is affected by the following issues which needs to be addressed before we can successfully see this bump passing regular e2es:
- [x] https://github.com/kubernetes-sigs/image-builder/pull/1951
- [x] https://github.com/kubernetes/kubeadm/issues/3294

**Checklist**:
- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:
```release-note
Bump CAPI to v1.12.2, k8s to v1.34 and controller-runtime to v0.22.5
```

PR #5992, opened 2026-04-27
🌱 test(e2e): bump EKS addons (VPC v1.21.1, kube-proxy v1.34.6)

AndiDog:
**What type of PR is this?**

/kind failing-test

Bumps EKS addon versions to be compatible with Kubernetes v1.34.8:
- VPC addon: v1.21.1-eksbuild.7 (from v1.19.5-eksbuild.3)
- kube-proxy addon: v1.34.6-eksbuild.2 (from v1.33.3-eksbuild.6)

Also adds `nodeDrainTimeout: 2m` to machine deployment templates to prevent EKS-managed coredns PDB from stalling MachineDeployment teardown during e2e tests.

**AI Usage**: None

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [ ] adds unit tests

**Release note**:

\`\`\`release-note
NONE
\`\`\`


damdo (commented):
/lgtm

damdo:
^ for EKS testing to finish

raykrueger:
Yes, I've confirmed this in my PR #5928. EKS 1.32 went into Extended support on March 23. The test attempts to deploy a cluster with STANDARD support set, and then changes it to EXTENDED support from there. EKS won't let us deploy 1.32 with STANDARD, so the test fails.

We should probably update the UPGRADE_FROM_VERSION as well.  On March 23 2027 our upgrade tests will start failing when we cannot create 1.32 clusters any more :)

raykrueger:
It looks like these tests are failing because we're still using AL2, and there are no AMIs for that after 1.32
https://docs.aws.amazon.com/eks/latest/userguide/al2023.html

So bumping it to 1.35 or any other version above 1.32 won't matter.

In [the logs](https://storage.googleapis.com/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5992/pull-cluster-api-provider-aws-e2e-eks/2049036664151478272/build-log.txt) we can see the error

```
failed to create AWSMachine instance: failed to get ami SSM parameter:
"/aws/service/eks/optimized-ami/1.35/amazon-linux-2/recommended/image_id": operation error SSM
```

test/e2e/data/eks/cluster-template-eks-managedmachinepool.yaml
```diff
@@ -51,5 +51,6 @@ spec:
 apiVersion: infrastructure.cluster.x-k8s.io/v1beta2
 kind: AWSManagedMachinePool
 metadata:
+  amiType: AL2023_x86_64_STANDARD
```
raykrueger: Looks like this landed in metadata and should be in spec.

AndiDog:
Maybe coredns pods can't be drained because there's only one worker node. Retrying:

AndiDog:
Cleaned up commits, backing out `Replicas: 2` which likely isn't needed. Trying if this succeeds again:

AndiDog:
That worked fine. I squashed into a single commit.

test/e2e/data/e2e_eks_conf.yaml
```diff
@@ -134,7 +134,7 @@ providers:
         targetName: "cluster-template-eks-nitro-enclave-managedmachinepool.yaml"
 
 variables:
-  KUBERNETES_VERSION: "v1.32.0"
+  KUBERNETES_VERSION: "v1.35.3"
```
damdo: Can we use v1.34.8 here and below in KUBE_PROXY_ADDON_VERSION?
I think we want to use that as the test reference for this dev cycle/release.

AndiDog: done

test/e2e/suites/managed/machine_pool.go
```diff
@@ -75,16 +73,6 @@ func MachinePoolSpec(ctx context.Context, inputGetter func() MachinePoolSpecInpu
 	configCluster.Flavor = input.Flavor
 	configCluster.WorkerMachineCount = ptr.To[int64](1)
 	workloadClusterTemplate := shared.GetTemplate(ctx, configCluster)
-	if input.UsesLaunchTemplate {
```
richardcase: Whats the reason for removing this block? And how does it relate to using newer versions?

AndiDog: It's dead code as the USER_DATA placeholder was only used for older Amazon Linux variants. See also [this more scoped commit](https://github.com/clebs/cluster-api-provider-aws/pull/10/changes) where the removed `USER_DATA` env and related code removal can be seen more obviously than in my large PR.

stefanSpectro:
Just checking is this PR still blocked on something?  I see [pull-cluster-api-provider-aws-e2e-blocking](https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5992/pull-cluster-api-provider-aws-e2e-blocking/2063909484702470144) is green.  Asking as this PR seems to block E2E testing on: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5808

damdo:
@AndiDog given most of the things that this PR was doing has merged via #5857 I rebased this to only include the actual diff between the two

damdo (commented):
/lgtm

PR #5999, opened 2026-05-03
🐛 fix: fully bypass CloudFormation when -skip-cloudformation-creation is set

### LINKED ISSUE #5944: skip-cloudformation-creation flag does not fully bypass CloudFormation stack dependency
/kind bug

Deploy the CAPA CloudFormation bootstrap stack in `us-east-1`, then run e2e tests in a different region with the skip flag:

```bash
AWS_REGION=us-east-2 make test-e2e \
  GINKGO_FOCUS="..." \
  E2E_ARGS="-skip-cloudformation-creation -skip-cloudformation-deletion"
```

The suite panics/fails in two places not guarded by `-skip-cloudformation-creation`:

**1. `ensureStackTags` — `test/e2e/shared/suite.go:155`**

Calls `CloudFormation:DescribeStacks` unconditionally after the creation block. Fails with:
```
ValidationError: Stack with id cluster-api-provider-aws-sigs-k8s-io does not exist
```

**2. `newUserAccessKey` — `test/e2e/shared/aws.go:883`**

This is related, but different. Our existing stack was deployed without the BootstrapUser.
This causes a nil-pointer panic rather than a handled error. 

`IAM:ListAccessKeys` error is silently discarded (`keyOuts, _ := ...`), leaving `keyOuts` nil and causing a nil-pointer panic:
```
runtime error: invalid memory address or nil pointer dereference
    test/e2e/shared/aws.go:883
```

`-skip-cloudformation-creation` should bypass all operations that assume the CF stack exists in the current region, not just the stack creation call itself.

`CloudFormation:DescribeStacks` is regional. A stack deployed in us-east-1 is not visible in us-east-2. IAM is global, so a bootstrap user created by the stack should exist cross-region, but the flag guards only the creation step, leaving two unconditional calls that require the stack to exist locally. Additionally, if the stack was deployed without `BootstrapUser.Enable = true`, the bootstrap user does not exist at all, making the newUserAccessKey failure inevitable regardless of region.

The silent error discard in `newUserAccessKey` (`keyOuts, _ := ...`) also independently deserves fixing. It turns a real API error into a confusing nil-pointer panic with no context.

Suggested fix: guard `ensureStackTags` and `newUserAccessKey` inside the existing `if !e2eCtx.Settings.SkipCloudFormationCreation` block, falling back to the caller's session credentials when skipping.

The ListAccessKeys nil pointer is more complicated. My cloudformation stack was deployed without the BootstrapUser. This is in a shared environment, so it's not something I can let e2e tests do dynamically.

My local work around was to use `&creds` and `&iamtypes.AccessKey` to populate the `e2eCtx.Environment.BootstrapAccessKey`.

**Environment:**

- Cluster-api-provider-aws version: main
- Kubernetes version: v1.32.0
- OS: macOS

I fully admit to, and apologize for, violating the new AI Policy and having Claude open this Issue. I didn't see the policy till just now, an hour after opening this issue. I do stand by the issue. I had to work around this with local hacks in order to run the e2e tests. -Ray

SAY-5:
**What type of PR is this?**

/kind bug

`-skip-cloudformation-creation` only guarded the `createCloudFormationStack` call. Two follow-on calls assumed the stack existed in the current region:

- `ensureStackTags` (suite.go:155) calls `CloudFormation:DescribeStacks` and fails with `ValidationError: Stack ... does not exist` when the stack lives in a different region (CloudFormation is regional).
- `newUserAccessKey` (aws.go:877) calls `IAM:ListAccessKeys` for the bootstrap user. If the existing stack was deployed without `BootstrapUser.Enable = true`, the user does not exist; the error from `ListAccessKeys` was being silently discarded (`keyOuts, _ := ...`) and the next loop iteration nil-derefs `keyOuts.AccessKeyMetadata`.

This PR moves both calls inside the existing `if !SkipCloudFormationCreation` guard. When skipping, it derives `BootstrapAccessKey` from the calling session's resolved credentials so downstream consumers (`encodeCredentials`, `BootstrapUserAWSSession`, `EnsureServiceQuotas`, etc.) keep working. It also surfaces the previously discarded `ListAccessKeys` error so any future failure produces a real diagnostic instead of a nil-pointer panic.

Fixes #5944

The skip path assumes the calling session uses long-lived IAM credentials (which is the typical setup when the CF stack has been pre-deployed). `iamtypes.AccessKey` has no `SessionToken` field, so STS temporary credentials would lose their token; this matches the behavior of `NewAWSSessionWithKey` already in the file.

**AI Usage**:

Claude Code (Sonnet/Opus) assisted with locating the affected call sites and drafting the patch. I reviewed every line before commit, validated the control flow against the issue reporter's repro, and confirmed `go build -tags=e2e ./test/e2e/shared/...` and `go vet` pass cleanly.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [ ] adds unit tests

**Release note**:

```release-note
Fix `-skip-cloudformation-creation` so it also bypasses the CloudFormation `DescribeStacks` tag check and the bootstrap user access-key creation, allowing e2e tests to run against a pre-deployed stack in a different region or one without `BootstrapUser` enabled. Also surface a previously discarded `ListAccessKeys` error that masked the real failure as a nil-pointer panic.
```

### DIFF
--- a/test/e2e/shared/aws.go
+++ b/test/e2e/shared/aws.go
@@ -872,14 +872,31 @@ func encodeCredentials(accessKey *iamtypes.AccessKey, region string) string {
 	return encCreds
 }
 
+// accessKeyFromSession returns an iamtypes.AccessKey populated from the
+// caller's resolved AWS credentials. It is used when -skip-cloudformation-creation
+// is set so that downstream consumers of BootstrapAccessKey (e.g. encodeCredentials
+// and NewAWSSessionWithKey) keep working without needing the CloudFormation
+// bootstrap user to exist.
+func accessKeyFromSession(ctx context.Context, cfg *aws.Config) *iamtypes.AccessKey {
+	creds, err := cfg.Credentials.Retrieve(ctx)
+	Expect(err).NotTo(HaveOccurred(), "should be able to resolve AWS credentials from the calling session when skipping CloudFormation creation")
+	Expect(creds.AccessKeyID).NotTo(BeEmpty(), "calling session must have a non-empty access key id when skipping CloudFormation creation")
+	Expect(creds.SecretAccessKey).NotTo(BeEmpty(), "calling session must have a non-empty secret access key when skipping CloudFormation creation")
+	return &iamtypes.AccessKey{
+		AccessKeyId:     aws.String(creds.AccessKeyID),
+		SecretAccessKey: aws.String(creds.SecretAccessKey),
+	}
+}
+
 // newUserAccessKey generates a new AWS Access Key pair based off of the
 // bootstrap user. This tests that the CloudFormation policy is correct.
 func newUserAccessKey(ctx context.Context, cfg *aws.Config, userName string) *iamtypes.AccessKey {
 	iamSvc := iam.NewFromConfig(*cfg)
 
-	keyOuts, _ := iamSvc.ListAccessKeys(ctx, &iam.ListAccessKeysInput{
+	keyOuts, err := iamSvc.ListAccessKeys(ctx, &iam.ListAccessKeysInput{
 		UserName: aws.String(userName),
 	})
+	Expect(err).NotTo(HaveOccurred(), fmt.Sprintf("listing access keys for bootstrap user %q", userName))
 	for i := range keyOuts.AccessKeyMetadata {
 		By(fmt.Sprintf("Deleting an existing access key: user-name=%s", userName))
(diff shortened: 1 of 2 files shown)


SAY-5:
CLA pending — will link the commit email to the GitHub account shortly to clear EasyCLA.

SAY-5:
rebased on main. the only conflict was in suite.go, upstream moved the access key wait into waitForAccessKeyPropagation, so the skip branch now falls through to that call as well.

PR #6007, opened 2026-05-09
🐛 Prevent access entries delete+recreate cycle on EKS reconciliation

### LINKED ISSUE #6003: accessEntries reconciliation causes continuous delete+recreate cycle, breaking auth
## What happened

When `accessEntries` is defined in `AWSManagedControlPlane`, CAPA deletes and recreates the access entry on every reconcile loop (~7 seconds). The access policy association is lost during this cycle because the next delete occurs before the policy is successfully re-associated.

This results in intermittent `Unauthorized` and `Forbidden` errors for all users authenticating via the affected access entry, making the cluster unusable for human operators.

## What I expected

CAPA should detect that the access entry already exists with the correct configuration and skip reconciliation (no delete+create).

## How to reproduce

1. Create an EKS cluster via CAPA with the following `accessEntries` in `AWSManagedControlPlane`:

```yaml
kind: AWSManagedControlPlane
metadata:
  name: my-cluster
spec:
  eksClusterName: my-cluster
  region: eu-central-1
  version: v1.35.1
  accessConfig:
    authenticationMode: api_and_config_map
  accessEntries:
    - principalARN: "arn:aws:iam::123456789012:role/aws-reserved/sso.amazonaws.com/eu-central-1/AWSReservedSSO_AdministratorAccess_abc123def456"
      type: standard
      accessPolicies:
        - policyARN: "arn:aws:eks::aws:cluster-access-policy/AmazonEKSClusterAdminPolicy"
          accessScope:
            type: cluster
  endpointAccess:
    private: true
    public: false
```

2. Wait for the cluster to become Ready.
3. Observe that `kubectl get ns` intermittently returns `Unauthorized` or `Forbidden`.

## Evidence

**Access entry `createdAt` changes every ~7 seconds** (entry is being deleted and recreated):

```
$ for i in 1 2 3; do aws eks describe-access-entry ... --query 'accessEntry.createdAt'; sleep 7; done
"2024-01-01T10:00:00.044000+00:00"
"2024-01-01T10:00:07.355000+00:00"
"2024-01-01T10:00:14.683000+00:00"
```

**Access policy is empty after recreation** — CAPA creates the entry but fails to associate the policy before the next reconcile deletes it:

```
$ aws eks list-associated-access-policies --cluster-name my-cluster --principal-arn "arn:aws:iam::..."
{
    "associatedAccessPolicies": []
}
```

**Kubernetes events** showed 707 reconciliations before event deduplication kicked in (count stopped incrementing while `createdAt` continued to change every 7s):

```
$ kubectl get events -n my-cluster --field-selector reason=SuccessfulReconcileAccessEntries
707    2024-01-01T10:00:00Z    2024-01-01T11:39:00Z
```

**Stopping CAPA controller confirms it is the cause** — after scaling the deployment to 0, `createdAt` stops changing and (once policy is manually re-associated) auth works stably.

**CAPA controller logs do not explicitly log access entry create/delete operations at Info level**, making this issue difficult to diagnose. The only indication is the Kubernetes event count and the changing `createdAt` timestamp.

## Suspected root cause

The access entry reconciler detects drift on every cycle even when the entry matches the desired state, triggering a delete+create. The sequence is: detect drift → delete entry (removes policy association) → create entry → (should associate policy, but next reconcile deletes before association completes).

## Environment

- CAPA version: v2.11.1
- EKS version: v1.35.1
- Kubernetes (management cluster): v1.35
- Auth mode: `API_AND_CONFIG_MAP`
- Endpoint: private only
- Principal ARN: SSO role with full IAM path (`/aws-reserved/sso.amazonaws.com/...`)

When `accessEntries` is not defined in the spec, CAPA logs `"no access entries defined, skipping reconcile"` and does not touch existing entries.

raykrueger:
**What type of PR is this?**

/kind bug

When `accessEntries` with an unspecified `type` field are defined in `AWSManagedControlPlane`, the reconciler detects drift on every cycle because an empty `Type` is compared as `""` against `"STANDARD"` (the EKS default) from the AWS API. This triggers a continuous delete-and-recreate loop every ~7s, with each deletion removing the access policy association and the subsequent recreation racing against the next reconcile before the policy can be re-associated.

This change:

1. Normalizes empty `Type` to `AccessEntryTypeStandard` before comparison in `updateAccessEntry`, so unspecified types no longer appear as drift.
2. Skips re-associating access policies that already match the desired scope to avoid unnecessary API calls.

Fixes #6003

The fix touches two functions in `pkg/cloud/services/eks/accessentry.go`:
- `updateAccessEntry` — normalizes empty type before drift detection
- `reconcileAccessPolicies` — skips already-matching policy associations; new helper `policyScopeMatches` compares desired vs existing scope

**AI Usage**:

This PR benefited from AI assistance (Qwen3.6-27B via Opencode) for:
- Initial code exploration and understanding the drift detection logic
- Drafting the PR description and release note
- Reviewing case-sensitivity of the `AccessEntryType.APIValue()` comparison

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
Fix infinite delete-and-recreate loop for EKS access entries when the type field is left unspecified (defaults to STANDARD).
```

### DIFF
--- a/pkg/cloud/services/eks/accessentry.go
+++ b/pkg/cloud/services/eks/accessentry.go
@@ -167,13 +167,19 @@ func (s *Service) updateAccessEntry(ctx context.Context, accessEntry ekscontrolp
 		return errors.Wrapf(err, "failed to describe access entry for principal %s", accessEntry.PrincipalARN)
 	}
 
+	// Use the actual type from AWS when not specified in spec (defaults to STANDARD on EKS)
+	desiredType := accessEntry.Type
+	if desiredType == "" {
+		desiredType = ekscontrolplanev1.AccessEntryTypeStandard
+	}
+
 	// EKS requires recreate when changing type or removing username
 	existingUsername := ""
 	if describeOutput.AccessEntry.Username != nil {
 		existingUsername = *describeOutput.AccessEntry.Username
 	}
 
-	if *accessEntry.Type.APIValue() != *describeOutput.AccessEntry.Type || accessEntry.Username != existingUsername {
+	if *desiredType.APIValue() != *describeOutput.AccessEntry.Type || accessEntry.Username != existingUsername {
 		if err = s.deleteAccessEntry(ctx, accessEntry.PrincipalARN); err != nil {
 			return errors.Wrapf(err, "failed to delete access entry for principal %s during recreation", accessEntry.PrincipalARN)
 		}
@@ -220,7 +226,12 @@ func (s *Service) deleteAccessEntry(ctx context.Context, principalArn string) er
 }
 
 func (s *Service) reconcileAccessPolicies(ctx context.Context, accessEntry ekscontrolplanev1.AccessEntry) error {
-	if accessEntry.Type == ekscontrolplanev1.AccessEntryTypeEC2Linux || accessEntry.Type == ekscontrolplanev1.AccessEntryTypeEC2Windows {
+	// Normalize type: empty means STANDARD, which needs policy reconciliation
+	entryType := accessEntry.Type
+	if entryType == "" {
+		entryType = ekscontrolplanev1.AccessEntryTypeStandard
+	}
+	if entryType == ekscontrolplanev1.AccessEntryTypeEC2Linux || entryType == ekscontrolplanev1.AccessEntryTypeEC2Windows {
 		s.scope.Info("Skipping access policy reconciliation for EC2 access type", "principalARN", accessEntry.PrincipalARN)
 		return nil
 	}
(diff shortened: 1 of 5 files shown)


richardcase:
This looks good to me, thanks @raykrueger

richardcase:
When the e2e passes feel free to unhold.

/hold

But from my side:

/approve

raykrueger:
This is failing on the upgrade policy test. So we'll still need #5992

raykrueger:
I’ll rebase this after some breakfast :)

damdo:
/hold

I think the failure here might be related PTAL @raykrueger

raykrueger:
@Ankitasw @damdo @richardcase This looks good to go now.

dlipovetsky:
The fix from https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6176 has resolved the e2e failure, so we can remove the hold.

/hold cancel

pkg/cloud/services/eks/accessentry.go
```diff
@@ -167,13 +167,16 @@ func (s *Service) updateAccessEntry(ctx context.Context, accessEntry ekscontrolp
 		return errors.Wrapf(err, "failed to describe access entry for principal %s", accessEntry.PrincipalARN)
 	}
 
-	// EKS requires recreate when changing type or removing username
```
AndiDog, line 170: Seems the previous code allowed removing the username. Does that make any sense to support? EKS docs say the username field [is optional](https://docs.aws.amazon.com/eks/latest/APIReference/API_AccessEntry.html). How does your PR currently behave in the case of removal?

raykrueger: Having no username is one of the two problem triggers here. If it is left blank AWS assigns a value and we end up in the loop. We'd have to have a way to differentiate "unset". I'm not sure it's valuable.

AndiDog (approved):
/approve

PR #6018, opened 2026-05-13
🐛 Allow AWSMachine instance creation when control-plane LB is disabled

### LINKED ISSUE #6017: AWSMachine instance creation fails with "APIServer ELB not available" when loadBalancerType is disabled
/kind bug

When setting `AWSCluster.spec.controlPlaneLoadBalancer.loadBalancerType: disabled`
to manage the control-plane load balancer externally and providing a custom
`spec.controlPlaneEndpoint`, control-plane `AWSMachine` objects never reach the
`Running` state. Instance creation fails immediately with:

```shell
failed to create AWSMachine instance: failed to run controlplane, APIServer ELB not available
```

Example `AWSCluster` configuration:
```yaml
spec:
  controlPlaneLoadBalancer:
    loadBalancerType: disabled
  controlPlaneEndpoint:
    host: myCustomDnsRecord
    port: 6443
```

The AWSCluster controller correctly stops managing the LB and marks
`LoadBalancerReady` as true once `spec.controlPlaneEndpoint` is populated.
However, the AWSMachine controller still checks
`status.network.apiServerELB.dnsName` in the instance creation path, which is
never populated when the LB is disabled, causing all control-plane machine
creation to fail permanently.

When `loadBalancerType: disabled` and `spec.controlPlaneEndpoint` is fully
populated, instance creation should proceed without requiring
`status.network.apiServerELB.dnsName` to be set.

The root cause is in `instances.go`. The guard:

```go
if !scope.IsControlPlaneExternallyManaged() && !scope.IsExternallyManaged() &&
    !scope.IsEKSManaged() && s.scope.Network().APIServerELB.DNSName == "" {
```

does not account for `loadBalancerType: disabled`. The disabled type is only
handled by the `AWSCluster` reconciler; the `AWSMachine` instance creation path
was never updated when disabled was introduced.

Environment:

Cluster-api-provider-aws version: v2.11.1
Kubernetes version (use kubectl version): 1.34
OS (e.g. from /etc/os-release): ubuntu 24.04

arno-pons:
**What type of PR is this?**

/kind bug

When `AWSCluster.spec.controlPlaneLoadBalancer.loadBalancerType` is set to `disabled`, the AWSCluster controller correctly delegates endpoint management to the user and waits for `spec.controlPlaneEndpoint` to be populated externally.

However, the AWSMachine instance creation code still blocked on `status.network.apiServerELB.dnsName` being non-empty, which is never populated when the LB is disabled. This caused all control-plane machine creation to fail with:

  "failed to run controlplane, APIServer ELB not available"

Add `IsControlPlaneLBDisabled()` to `MachineScope` and include it in the ELB availability guard so that instance creation proceeds when the user has opted out of CAPA-managed load balancing and provided their own `spec.controlPlaneEndpoint`.

**AI Usage**:

Copilot usage to fix the issue.

Fixes #6017

None

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/pkg/cloud/scope/machine.go
+++ b/pkg/cloud/scope/machine.go
@@ -408,6 +408,19 @@ func (m *MachineScope) IsExternallyManaged() bool {
 	return annotations.IsExternallyManaged(m.InfraCluster.InfraCluster())
 }
 
+// IsControlPlaneLBDisabled returns true when the AWSCluster spec sets the control-plane
+// load balancer type to "disabled".  In that case the caller is responsible for
+// providing spec.controlPlaneEndpoint and CAPA must not block instance creation
+// because APIServerELB.DNSName is empty.
+func (m *MachineScope) IsControlPlaneLBDisabled() bool {
+	awsCluster, ok := m.InfraCluster.InfraCluster().(*infrav1.AWSCluster)
+	if !ok {
+		return false
+	}
+	return awsCluster.Spec.ControlPlaneLoadBalancer != nil &&
+		awsCluster.Spec.ControlPlaneLoadBalancer.LoadBalancerType == infrav1.LoadBalancerTypeDisabled
+}
+
 // SetInterruptible sets the AWSMachine status Interruptible.
 func (m *MachineScope) SetInterruptible() {
 	if m.AWSMachine.Spec.SpotMarketOptions != nil {
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -197,7 +197,7 @@ func (s *Service) CreateInstance(ctx context.Context, scope *scope.MachineScope,
 		input.PublicIPOnLaunch = ptr.To(false)
 	}
 
-	if !scope.IsControlPlaneExternallyManaged() && !scope.IsExternallyManaged() && !scope.IsEKSManaged() && s.scope.Network().APIServerELB.DNSName == "" {
+	if !scope.IsControlPlaneExternallyManaged() && !scope.IsExternallyManaged() && !scope.IsEKSManaged() && !scope.IsControlPlaneLBDisabled() && s.scope.Network().APIServerELB.DNSName == "" {
 		record.Eventf(s.scope.InfraCluster(), "FailedCreateInstance", "Failed to run controlplane, APIServer ELB not available")
 		return nil, awserrors.NewFailedDependency("failed to run controlplane, APIServer ELB not available")
 	}
--- a/pkg/cloud/services/ec2/instances.go
+++ b/pkg/cloud/services/ec2/instances.go
@@ -203,7 +203,7 @@ func (s *Service) CreateInstance(ctx context.Context, scope *scope.MachineScope,
 		input.PublicIPOnLaunch = ptr.To(false)
 	}
 
-	if !scope.IsControlPlaneExternallyManaged() && !scope.IsExternallyManaged() && !scope.IsEKSManaged() && s.scope.Network().APIServerELB.DNSName == "" {
+	if !scope.IsControlPlaneExternallyManaged() && !scope.IsExternallyManaged() && !scope.IsEKSManaged() && !scope.IsControlPlaneLBDisabled() && s.scope.Network().APIServerELB.DNSName == "" {
 		record.Eventf(s.scope.InfraCluster(), "FailedCreateInstance", "Failed to run controlplane, APIServer ELB not available")
 		return nil, awserrors.NewFailedDependency("failed to run controlplane, APIServer ELB not available")
 	}
--- a/pkg/cloud/services/ec2/instances_test.go
+++ b/pkg/cloud/services/ec2/instances_test.go
@@ -6031,6 +6031,118 @@ func TestCreateInstance(t *testing.T) {
 				}
 			},
 		},
+		{
+			name: "disabled load balancer with custom control plane endpoint should allow instance creation",
+			machine: &clusterv1.Machine{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: map[string]string{"set": "node"},
+				},
+				Spec: clusterv1.MachineSpec{
+					Bootstrap: clusterv1.Bootstrap{
+						DataSecretName: ptr.To[string]("bootstrap-data"),
+					},
+				},
+			},
+			machineConfig: &infrav1.AWSMachineSpec{
+				AMI: infrav1.AMIReference{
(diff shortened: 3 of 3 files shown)


arno-pons:
Hell @richardcase,

Tests seems ok now. Is there anything else I need to do here for PR to be merged?

Thanks and best regards.
Arnaud.

damdo:
> However, the AWSMachine instance creation code still blocked on status.network.apiServerELB.dnsName being non-empty, which is never populated when the LB is disabled. This caused all control-plane machine creation to fail with:

Was this because of a mistake in the logic or do you see a reason for this to be in place?

arno-pons:
Good question — I'm not 100% sure of the original author's intent, but based on tracing the code (IA assisted), my guess is this wasn't a deliberate decision to block the `disabled` case, but rather that this condition was probably just never revisited when it was added.

The check isn't required by the EC2 `RunInstances` call itself — nothing in the instance input depends on `APIServerELB.DNSName`. It's more a fail-fast dependency check. Probably because if a control-plane node is bootstrapped before the API server endpoint is actually known, then failures could be more difficult to diagnose (bad/empty cert SANs, kubeadm join failures, etc...).

When `loadBalancerType: disabled` was added, the AWSCluster controller got its own readiness path for it (`checkForExternalControlPlaneLoadBalancer`, checking the user defined `spec.controlPlaneEndpoint` instead of the ELB), but it looks like this check in `instances.go` just wasn't updated.

So, the intent (don't create control-plane instances before the endpoint is known) is still valid and worth keeping, but checking `APIServerELB.DNSName` specifically mean CAPA always owns the LB, which became false once `disabled` was introduced.

damdo (commented):
/lgtm

Will leave the approval to @richardcase

arno-pons:
Hello @richardcase,

Any chance of having this PR merged for next release?

Thanks and regards.

lmilbaum:
**Disclosure:** I'm new to this project. I used Claude Code (AI) to read this PR and its discussion and to draft this comment, and the conclusions below come from that AI-assisted read of the code. I have not run the tests or reproduced anything myself, so please treat this as a set of questions and possible mistakes to check, not as authoritative findings. Apologies in advance for any noise. This is a plain comment, not a review.

I see the PR already has `/ok-to-test`, passing tests and an LGTM from @damdo, and is waiting on @richardcase. I don't want to hold it up, but there is one thing I'd like a maintainer to confirm before merge (point 1).

Thanks for the fix, @arno-pons. The diagnosis looks right to me: the AWSCluster controller already handles `loadBalancerType: disabled` (`controllers/awscluster_controller.go:316`), and the guard at `pkg/cloud/services/ec2/instances.go:206` was the part still blocking on `APIServerELB.DNSName`. I also agree with your reply to @damdo that the check is a fail-fast "endpoint must be known" guard whose intent is still valid, and that it was simply not updated when `disabled` was introduced.

**1. Possible follow-on failure in LB attachment (the most important point, and unverified)**
After the instance is created, `reconcileLBAttachment` (`controllers/awsmachine_controller.go:1040`) loops over `ControlPlaneLoadBalancers()`, which returns the spec as-is, including a `disabled` one. For a running control-plane instance it calls `registerInstanceToLBs`, which has no `disabled` case and returns `unknown load balancer type "disabled"` (lines ~1074-1083). The caller at ~line 727 only filters out `InstanceNotRunning`, so it looks like the AWSMachine reconcile could keep returning an error and `ELBAttached` would never go healthy, even though the EC2 instance now gets created. The deregister path may need the same handling. I didn't see this raised in the discussion so far. Could someone confirm this end to end (e2e or envtest)? If it holds, skipping `LoadBalancerTypeDisabled` in that loop (or adding a no-op case) may be needed, possibly as a follow-up PR if you'd rather not widen this one.

**2. PR checklist**
"adds or updates e2e tests" and "includes documentation" are ticked, but the diff touches only `machine.go`, `instances.go` and `instances_test.go`. Maybe those boxes were ticked by mistake?

**3. Test coverage**
- The new test machine is labelled `set: node`, so it isn't a control-plane machine. It still passes because the guard doesn't check the machine's role, but it isn't quite the scenario described.
- There's no negative case showing that a nil or `classic` LB with an empty DNS name still returns the failed-dependency error.
- There's no direct unit test for `IsControlPlaneLBDisabled()` (nil spec, non-`AWSCluster` infra cluster, `disabled`).

**4. Minor nits**
- Since the intent is "don't create control-plane instances before the endpoint is known", the guard could also check that `spec.controlPlaneEndpoint.Host` is non-empty when the LB is disabled. This is probably covered by the cluster controller gating readiness, so it may not be needed.
- Double space after the period in the new doc comment.
- The error message says "failed to run controlplane" although the guard also runs for worker machines (this is pre-existing).

Overall the core change looks sensible to me. Point 1 is the one I'd most like a maintainer's eyes on. Please correct anything I've got wrong.

arno-pons:
Thanks for the detailed read, @lmilbaum. I dug into it and pushed follow-up commits:

1. LB attachment for disabled type. You were right: ControlPlaneLoadBalancers() returns the spec as-is (never nil for disabled), and registerInstanceToLBs/reconcileLBAttachment had no case for it, so it fell through to unknown load balancer type "disabled" on every control-plane reconcile once instance creation was unblocked. Fixed by skipping disabled load balancer specs in reconcileLBAttachment before they reach the register/deregister functions, since CAPA doesn't manage them. Added a unit test asserting no ELB calls happen in that case.

2. PR checklist. You're right, those two boxes were ticked by mistake. This PR only touches machine.go, instances.go, and their tests.

3. Test coverage:

- Relabeled the CreateInstance disabled-LB test machine with the actual control-plane label instead of set: node, so it matches the scenario it's meant to exercise.
- Added a negative test asserting CreateInstance still returns the FailedDependency error when the load balancer is not disabled and APIServerELB.DNSName is empty.
- Added direct unit tests for IsControlPlaneLBDisabled(): nil spec, disabled, classic, and a non-AWSCluster infra cluster (EKS), which previously fell through the type assertion untested.

4. Minor nits:

- Fixed the double space in the doc comment.
- Left the "failed to run controlplane" message and the lack of a spec.controlPlaneEndpoint.Host check as-is. Both are pre-existing, and the latter is already covered by AWSCluster's LoadBalancerReadyCondition (checkForExternalControlPlaneLoadBalancer), so a duplicate check here seemed redundant.
- All existing + new tests pass locally (go test ./controllers/... ./pkg/cloud/scope/... services.).

PR #6034, opened 2026-05-28
✨ feat(awsmanagedcontrolplane): support resource marked as externally managed

aidan-canva:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:
Adds support for `AWSManagedControlPlane` resources to be marked as externally managed using the standard `cluster.x-k8s.io/managed-by` annotation.  This allows operators to use CAPI/CAPA to manage machines with pre-existing EKS clusters managed by external-systems (eg, Terraform).

**Which issue(s) this PR fixes**:
Fixes #6031

**Special notes for your reviewer**:
The second test added is a little redundant with the existing `TestAWSManagedControlPlaneReconcilerIntegrationTests` test but IMO a positive/negative check sitting close together will make identifying a regression easier in the future.

I'm not sure how detailed to go in the docs.  Too detailed and it'll likely just become out of date quickly as the status/conditions required will change as the internal controllers logic changes. IMO it's a pretty 'power-user' feature so some thing can be left up to the operator to determine for their environment, like the `kubeconfig` credential generation.  Alternatively we could to a longer/more detailed guide.

**AI Usage**:
My usage of CAPA doesn't extend to all of its features, so AI was used as an additional review before opening the PR and to try and capture any missed status/conditions or fields required for a minimal working `AWSManagedControlPlane` resource that's externally managed.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes AI generated content
- [x] includes emoji in title
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Support for externally managed `AWSManagedControlPlane` resources using the `cluster.x-k8s.io/managed-by` annotation.
```

PR #5700, opened 2025-10-07
:sparkles:  Implement nodeadm bootstrapping type

faiq:
**What type of PR is this?**
/kind feature

**What this PR does / why we need it**:

This PR implements the nodeadm config type outlined by the KEP #5678 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests
- [x] adds or updates e2e tests

**Release note**:

```release-note
Implement nodeadm bootstrapping type
```

PR #6036, opened 2026-05-29
✨ feat(eks-hybrid): add support for EKS Hybrid Node bootstrap configuration

### LINKED ISSUE #6032: Support for EKS Hybrid Node bootstrap generation
/kind feature

Add support for generating bootstrap configuration (and credentials) for CAPI managed nodes to join an EKS cluster using the [Hybrid Node](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-overview.html) feature.

EKS has support for enabling non-EC2 instances to join an EKS cluster as Nodes using the [Hybrid Node](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-overview.html) feature.  This process involves enabling some configuration on the [control-plane itself](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-cluster-update.html), as well as generating credentials for the Machine to use when executing the [nodeadm](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-join.html) bootstrapper. 

Supporting this feature provides a path for cluster-api providers to be used to create a CAPI-managed hybrid cluster using EKS across multiple providers.

This feature would leverage the work laid out in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5700, ideally extending the `NodeadmConfig` CRD to support additional parameters to support the creation of short-lived [SSM activation](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-creds.html#hybrid-nodes-ssm) on-demand as part of the Machine Bootstrap generation. 

I am keen to contribute this feature if there is appetite to extend CAPA to support it.

**Environment:**

- Cluster-api-provider-aws version: n/a
- Kubernetes version: (use `kubectl version`): 1.32+
- OS (e.g. from `/etc/os-release`): All

aidan-canva:
**What type of PR is this?**
/kind feature

Extends `NodeadmConfig` to support [EKS Hybrid Node](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-overview.html) config generation with per-Machine [SSM activation](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-creds.html#hybrid-nodes-ssm) credentials.

This allows CAPA to handle the life-cycling of join-credentials while other infrastructure providers in the CAPI ecosystem handle the Machine life-cycling, allowing for multi-cloud/provider EKS usage.

As part of this, adds a more generic option for user-data generation using a user provided template that is rendered/interpolated with well-defined machine/cluster values.  This allows more flexibility with other CAPI providers consuming this bootstrap data which may not handle the rigidity of the existing approaches structure (unfortunately hit this in practice).

Fixes #6032

This PR only covers the bootstrap generation through `NodeadmConfig` which is already pretty big with the generated changes.  There is additional configuration required (though minimal) on the [EKS control plane](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-cluster-update.html#hybrid-nodes-cluster-enable-existing) side that I think makes sense to be done in a separate PR.

The PR is lacking updates to docs (and E2E tests) at the moment - but I wanted to get an early set of eyes on the approach before updating them (I hope that's okay).

Some specific areas that I think could use some discussion are:
- Does this warrant a separate CRD from `NodeadmConfig` ? I think the current implementation of a `Hybrid` field holds this unique configuration nicely but the added rendering approach 
(`NodeadmConfig.Hybrid.CustomUserData`) might justify it being separate.
- The new rendering approach (`NewCustomHybridUserdata`) is to enable broader flexibility for consumers (other CAPI providers) and operators. The existing stricter cloud-init rendering caused that compatibility issues with some providers that were challenging to work around directly at infra time.
- `reconcileHybridNode` and `joinWorker` are kept as separate code paths as it makes status/log tracing easier to follow IMO but I can see it being consolidated if preferred.
- The `nodeadm` binary used in hybrid situations is [different](https://github.com/aws/eks-hybrid) to the standard EC2 `nodeadm`, this will be clarified in docs but wanted to call it out in case there is confusion in the fields being added.
- EKS also supports hybrid node activations using [AWS IAM Roles Anywhere](https://docs.aws.amazon.com/eks/latest/userguide/hybrid-nodes-creds.html#hybrid-nodes-iam-roles-anywhere) - this PR does not add that support (yet?) in an attempt to reduce the scope and focus on what I can easily test.
- We have this functioning quite nicely in real environments based off a pre-merged version of https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5700 (and leveraging https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6034) and will do more testing with this rebased version.

**AI Usage**:
AI was used to support a refactor after #5700 was merged as this feature was originally implemented using a pre-merged commit.  It was also used to review the implementation for consistency across other parts of the codebase, remove non-useful tests I added and to clean up wording in in comments.  It was also used for general code reviews before opening this PR.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
Adds support for EKS Hybrid Node bootstrap creation using generated or BYO SSM activation credentials.
```


bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
```diff
@@ -65,6 +65,137 @@ type NodeadmConfigSpec struct {
 	// Mounts specifies a list of mount points to be setup.
 	// +optional
 	Mounts []MountPoints `json:"mounts,omitempty"`
+
+	// Hybrid contains configuration for EKS Hybrid Nodes.
+	// When specified, the NodeadmConfig generates userdata for hybrid node
+	// bootstrapping instead of standard EC2 node bootstrapping.
+	// Hybrid mode is mutually exclusive with EC2-specific options.
```
richardcase: What EC2 specific options are these? Do we need to enforce this via validation webhook or cel?

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
```diff
@@ -65,6 +65,137 @@ type NodeadmConfigSpec struct {
 	// Mounts specifies a list of mount points to be setup.
 	// +optional
 	Mounts []MountPoints `json:"mounts,omitempty"`
+
+	// Hybrid contains configuration for EKS Hybrid Nodes.
```
richardcase, line 70: To be clear, when using "hybrid" nodeadm is still used (but a different version specifically for hybrid )?

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
```diff
@@ -65,6 +65,137 @@ type NodeadmConfigSpec struct {
...
+	// +kubebuilder:validation:Required
+	SSM *HybridSSMOptions `json:"ssm"`
+
+	// CustomUserData allows providing a custom userdata template with variable interpolation.
+	// When specified, the default nodeadm MIME multipart userdata generation is completely
+	// replaced with the user-provided template. The template uses Go text/template syntax.
+	// This is useful for hybrid nodes in environments where cloud-init or the standard
+	// bootstrap approach is not suitable.
+	// +optional
+	CustomUserData *CustomUserDataOptions `json:"customUserData,omitempty"`
```
richardcase, line 92: Is the expectation that the custom user data would still use nodeadm to do the bootstrapping?

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
```diff
@@ -65,6 +65,137 @@ type NodeadmConfigSpec struct {
...
+}
+
+// HybridSSMOptions configures SSM activation-based authentication for hybrid nodes.
+// Either ActivationRef or ActivationConfig must be specified, but not both.
+type HybridSSMOptions struct {
+	// ActivationRef references an existing Secret containing SSM activation credentials.
+	// The Secret must contain 'activationId' and 'activationCode' keys.
+	// When specified, the controller will not create or manage SSM activations.
+	// +optional
+	ActivationRef *SSMActivationReference `json:"activationRef,omitempty"`
```
richardcase: Could probably use a `LocalObjectReference` instead of our own struct for the ref

bootstrap/eks/controllers/nodeadmconfig_controller.go
```diff
@@ -126,10 +150,15 @@ func (r *NodeadmConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reques
 
 	// set up defer block for updating config
 	defer func() {
+		conditions := []clusterv1beta1.ConditionType{
+			eksbootstrapv1.DataSecretAvailableCondition,
+		}
+		// Include SSM condition in summary for hybrid nodes
+		if config.Spec.Hybrid != nil && config.Spec.Hybrid.SSM != nil && config.Spec.Hybrid.SSM.ActivationConfig != nil {
```
richardcase: This test is done in 2 places, it would be worth extracting to a `isHybrid` function

bootstrap/eks/controllers/nodeadmconfig_controller.go
```diff
@@ -284,6 +330,458 @@ func (r *NodeadmConfigReconciler) joinWorker(ctx context.Context, cluster *clust
...
+		case err == nil:
+			return ctrl.Result{}, nil
+		case !apierrors.IsNotFound(err):
+			log.Error(err, "unable to check for existing bootstrap secret")
+			return ctrl.Result{}, err
+		}
+	}
+
+	// Validate control plane reference
+	if cluster.Spec.ControlPlaneRef.Kind != "AWSManagedControlPlane" {
```
richardcase: You'd probably want to do this before checking the secret?

bootstrap/eks/controllers/nodeadmconfig_controller.go
```diff
@@ -284,6 +330,458 @@ func (r *NodeadmConfigReconciler) joinWorker(ctx context.Context, cluster *clust
...
+	}
+
+	// Get control plane
+	controlPlane := &ekscontrolplanev1.AWSManagedControlPlane{}
+	if err := r.Get(ctx, client.ObjectKey{Name: cluster.Spec.ControlPlaneRef.Name, Namespace: cluster.Namespace}, controlPlane); err != nil {
+		return ctrl.Result{}, errors.Wrap(err, "failed to get control plane")
+	}
+
+	// Check if control plane is ready
+	if !v1beta1conditions.IsTrue(controlPlane, ekscontrolplanev1.EKSControlPlaneReadyCondition) {
```
richardcase, line 387: I'm not sure we need this check if we are testing `cluster.Status.Initialization.ControlPlaneInitialized` above

richardcase: In your testing did you see something that made this a required test?

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
```diff
@@ -65,6 +65,137 @@ type NodeadmConfigSpec struct {
...
+}
+
+// HybridOptions defines configuration for EKS Hybrid Nodes.
+// When specified, the NodeadmConfig generates userdata for hybrid node
+// bootstrapping instead of standard EC2 node bootstrapping.
+type HybridOptions struct {
+	// SSM configures SSM-based authentication for hybrid nodes.
+	// This is required for hybrid node support.
+	// +kubebuilder:validation:Required
+	SSM *HybridSSMOptions `json:"ssm"`
```
richardcase: If `SSM` is required then this doesn't need to be a pointer

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
```diff
@@ -65,6 +65,137 @@ type NodeadmConfigSpec struct {
...
+	//   /usr/local/bin/nodeadm init -c file:///etc/nodeadm/config.yaml
+	//
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MinLength=1
+	Template string `json:"template"`
+}
+
+// HybridSSMOptions configures SSM activation-based authentication for hybrid nodes.
+// Either ActivationRef or ActivationConfig must be specified, but not both.
+type HybridSSMOptions struct {
```
richardcase, line 152: We should probably enforce the "one of ActivationRef or ActivationConfig must be supplied". Either in a validating webhook or using `kubebuilder:validation:XValidation`

bootstrap/eks/internal/userdata/nodeadm.go
```diff
@@ -137,25 +151,51 @@ type NodeadmInput struct {
...
+	ActivationCode string
+
+	// MIME boundary
+	Boundary string
+}
+
+// isHybridMode returns true if the input is configured for hybrid node generation.
+// Hybrid mode is indicated by the presence of Region or SSM activation fields.
+func (input *NodeadmInput) isHybridMode() bool {
+	return input.Region != "" || input.ActivationID != "" || input.ActivationCode != ""
```
richardcase: Could we pass in an explicit field to say its hybrid? Like `Hybrid bool` instead of implying it from field values.

On the NodeAdmConfig is explicit that its Hybrid because `config.Spec.Hybrid` is not nil and we can use the nillability of the field to pass into this.

bootstrap/eks/internal/userdata/nodeadm.go
```diff
@@ -137,25 +151,51 @@ type NodeadmInput struct {
...
-	if input.APIServerEndpoint == "" {
-		return fmt.Errorf("API server endpoint is required for nodeadm")
-	}
-	if input.CACert == "" {
-		return fmt.Errorf("CA certificate is required for nodeadm")
-	}
 	if input.ClusterName == "" {
 		return fmt.Errorf("cluster name is required for nodeadm")
 	}
+	if input.isHybridMode() {
```
richardcase: Its ok to have this validation here as backup, but we need this validation as early as possible in the webhook or via cel.

bootstrap/eks/internal/userdata/nodeadm_hybrid_custom.go
```diff
@@ -0,0 +1,174 @@
...
+	// It will be converted to a YAML string before template rendering.
+	KubeletConfig *runtime.RawExtension
+
+	// ContainerdConfig contains the containerd configuration (optional).
+	ContainerdConfig string
+}
+
+// customTemplateFuncMap provides template functions for custom userdata templates.
+// These functions enable common operations within user-provided templates.
+var customTemplateFuncMap = template.FuncMap{
```
richardcase: It might be worth moving these template functions to a new `template` package, as I can see them being reused in the future.

bootstrap/eks/internal/userdata/nodeadm_hybrid_custom.go
```diff
@@ -0,0 +1,174 @@
...
+	"replace": strings.ReplaceAll,
+	// lower converts a string to lowercase
+	"lower": strings.ToLower,
+	// upper converts a string to uppercase
+	"upper": strings.ToUpper,
+}
+
+// validateCustomHybridInput validates the required fields for custom hybrid userdata generation.
+func validateCustomHybridInput(input *CustomHybridInput) error {
+	if input == nil {
```
richardcase, line 30: Similar comment here regarding validation.

pkg/cloud/services/ssm/activation.go
```diff
@@ -0,0 +1,198 @@
...
+
+	// Set optional fields if provided
+	if params.DefaultInstanceName != "" {
+		input.DefaultInstanceName = aws.String(params.DefaultInstanceName)
+	}
+	if params.Description != "" {
+		input.Description = aws.String(params.Description)
+	}
+
+	output, err := s.SSMClient.CreateActivation(ctx, input)
```
richardcase, line 161: For the new SSM service calls we may need to update the cloudformation stack that gives these permissions to the CAPA service account. It would be worth checking in: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/cmd/clusterawsadm/cloudformation/bootstrap/cluster_api_controller.go

richardcase: (this applies to any new AWS calls)

richardcase: Seems to be just CreateActivation and DeleteActivation

richardcase (commented):
Thanks for this @aidan-canva .

Theres a few comments but nothing major imho. Feel free to ping me here or on slack if you want to discuss any of them.

aidan-canva:
Thank you for the review/feedback @richardcase - I'll start chipping away at these over the next couple days.

pkg/cloud/services/ssm/activation.go
aidan-canva: Fixed in https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6036/commits/cfa2f2b7d34153cc39e6023993746b9d3aeb3044

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
aidan-canva: Yeah, from a user POV it's quite confusing.  Both binaries are called `nodeadm` and they are both driven by a `nodeconfigs.node.eks.aws/v1alpha` manifest, but they use separate definitions that have diverged.

[Hybrid nodeconfigs definition](https://raw.githubusercontent.com/aws/eks-hybrid/refs/heads/main/crds/node.eks.aws_nodeconfigs.yaml) vs [regular nodeconfigs](https://raw.githubusercontent.com/awslabs/amazon-eks-ami/refs/heads/main/nodeadm/crds/node.eks.aws_nodeconfigs.yaml) 

Rather than having that confusion bleed into CAPA, I am thinking that we support interpolating all values from the `NodeadmConfigSpec` struct in the custom user data.  It's a free text template, so the user can handle boilerplate around the unsupported nodeconfig fields themselves.

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
aidan-canva: See comment above; I think this constraint would be better removed so all fields are consistently supported in the interpolation.

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
aidan-canva: Yep, that is correct.  Though depending on the user, they will likely have a custom image build and may have bootstrap scripts that abstract the execution of `nodeadm` directly. 
A simple userdata example would just be the `nodeadm init ...` but more complex environments could have a large amount of unrelated bootstrapping around that binary being executed.

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
aidan-canva: Good idea, I will add that in 👍

bootstrap/eks/controllers/nodeadmconfig_controller.go
aidan-canva: I borrowed this from the existing `joinWorker` method which checks both

bootstrap/eks/internal/userdata/nodeadm.go
aidan-canva: I think adding the `XValidation` rules in d6cc7c853e16878fb369999b4f80734643ef5b4b is the best we can do.  The other fields are are pulled in from other resources/objects and are not known at CRD mutation time of the nodeadm template.

bootstrap/eks/api/v1beta2/nodeadmconfig_types.go
aidan-canva: Alright, I've implemented this in 15f0ffddb8ee049a2e45ea94f9aa343ab48e3843 - basically removed the need for [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6036#discussion_r3333467208) comment by making all fields in the `NodeadmConfigSpec` supported within the hybrid custom uerdata rendering.  That removes a fair bit of redundancy I added in earlier and I've also centralized the required fields checks which helps with another comment.

ChaoHuang2018:
I noticed a possible retry/cache-lag race in the managed SSM activation path.

If reconcile creates an SSM activation and then writes the activation Secret/status, a subsequent reconcile may still observe the old cached state where the activation Secret/status is absent. In that schedule, the branch can call `CreateHybridActivation` again before the cache observes the write.

The current Secret create-or-update behavior is helpful for the Kubernetes Secret write, but it would not prevent a duplicate external activation, because `CreateHybridActivation` has already happened before the Secret is created or updated.

Would it make sense to guard the external `CreateHybridActivation` call with a live read of the activation Secret/status, or make activation creation idempotent by looking up an existing activation keyed by the Machine/Cluster/NodeadmConfig before creating a new one?

I can try to contribute a small regression test if that would be helpful.

PR #6038, opened 2026-05-29
🐛 ROSA Fix cross account creation issues-6037 

serngawy:
**What type of PR is this?**
bug 

All ROSA-HCP controllers were sharing a single cached AWS session. As a result, when multiple AWS accounts were configured and mapped to different AWSClusterController identities, AWS resources were incorrectly created using the client associated with the first AWS account initialized in the cache.

The fix ensures that the AWS client session is re-established during each reconcile loop whenever the currently cached session belongs to a different target AWS account. This guarantees that all AWS operations are executed against the correct account associated with the reconciled AWSClusterController identity.

Fixes #6073

**AI Usage**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
Fix ROSA-HCP usage with aws cross accounts. 
```

### DIFF
--- a/controllers/rosacluster_controller.go
+++ b/controllers/rosacluster_controller.go
@@ -24,7 +24,6 @@ import (
 	"github.com/pkg/errors"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/client-go/tools/record"
@@ -41,9 +40,7 @@ import (
 	rosacontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/rosa/api/v1beta2"
 	expinfrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/exp/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/exp/utils"
-	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope"
-	stsservice "sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/sts"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/logger"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/rosa"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/util/paused"
@@ -58,7 +55,6 @@ type ROSAClusterReconciler struct {
 	client.Client
 	Recorder         record.EventRecorder
 	WatchFilterValue string
-	NewStsClient     func(cloud.ScopeUsage, cloud.Session, logger.Wrapper, runtime.Object) stsservice.STSClient
 	NewOCMClient     func(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope) (rosa.OCMClient, error)
 }
 
@@ -137,7 +133,6 @@ func (r *ROSAClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 		ControlPlane:   controlPlane,
 		ControllerName: "",
 		Logger:         log,
-		NewStsClient:   r.NewStsClient,
 	})
 	if err != nil {
 		return ctrl.Result{}, fmt.Errorf("failed to create rosa controlplane scope: %w", err)
@@ -156,7 +151,6 @@ func (r *ROSAClusterReconciler) Reconcile(ctx context.Context, req ctrl.Request)
 func (r *ROSAClusterReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Manager, options controller.Options) error {
 	log := logger.FromContext(ctx)
 	r.NewOCMClient = rosa.NewWrappedOCMClient
-	r.NewStsClient = scope.NewSTSClient
 
 	rosaCluster := &expinfrav1.ROSACluster{}
 
--- a/controlplane/rosa/controllers/rosacontrolplane_controller.go
+++ b/controlplane/rosa/controllers/rosacontrolplane_controller.go
@@ -29,7 +29,10 @@ import (
 	"strings"
 	"time"
 
-	stsv2 "github.com/aws/aws-sdk-go-v2/service/sts"
+	awsv2 "github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/aws/arn"
+	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
+	stsv2sdk "github.com/aws/aws-sdk-go-v2/service/sts"
 	"github.com/google/go-cmp/cmp"
 	idputils "github.com/openshift-online/ocm-common/pkg/idp/utils"
 	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
@@ -40,7 +43,6 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	kerrors "k8s.io/apimachinery/pkg/util/errors"
 	"k8s.io/apiserver/pkg/storage/names"
@@ -59,9 +61,7 @@ import (
 	rosacontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/rosa/api/v1beta2"
 	expinfrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/exp/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/annotations"
-	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope"
-	stsiface "sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/sts"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/logger"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/rosa"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/utils"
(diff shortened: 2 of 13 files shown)


controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1361,6 +1366,90 @@ func operatorIAMRoles(rolesRef rosacontrolplanev1.AWSRolesRef) []ocm.OperatorIAM
...
 
+// resolveCreatorForTargetAccount handles cross-account deployments: when the management-cluster
+// identity is in a different AWS account than the target account (where the InstallerRoleARN and
+// other role ARNs reside), it assumes the InstallerRoleARN using the current session credentials
+// and calls GetCreator() on the resulting client to obtain a real STS identity in the target account.
+//
+// OCM requires that creator.AccountID, the account parsed from creator.ARN, and every role ARN
+// account are all the same (target) account. Assuming the InstallerRole gives a genuine
+// sts:assumed-role ARN in the target account, satisfying both OCM validations.
+func (r *ROSAControlPlaneReconciler) resolveCreatorForTargetAccount(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope, creator *rosaaws.Creator) (*rosaaws.Creator, error) {
```
andclt, line 1379: can we add some tests for:
1. same account scenario - original creator
2. cross account scenario - return new creator with target account
3. missing `InstallerRoleARN` - original creator
4. invalid ARN
etc.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1361,6 +1366,90 @@ func operatorIAMRoles(rolesRef rosacontrolplanev1.AWSRolesRef) []ocm.OperatorIAM
...
+	installerRoleARN := rosaScope.ControlPlane.Spec.InstallerRoleARN
+	if rosaScope.ControlPlane.Spec.RosaRoleConfigRef != nil {
+		rosaRoleConfig := &expinfrav1.ROSARoleConfig{}
+		if err := r.Client.Get(ctx, client.ObjectKey{
+			Name:      rosaScope.ControlPlane.Spec.RosaRoleConfigRef.Name,
+			Namespace: rosaScope.ControlPlane.Namespace,
+		}, rosaRoleConfig); err == nil {
+			installerRoleARN = rosaRoleConfig.Status.AccountRolesRef.InstallerRoleARN
```
andclt: should we return the error here if the error is thrown because the `Get` call failed due to a transient error and not because it was not found?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1361,6 +1366,90 @@ func operatorIAMRoles(rolesRef rosacontrolplanev1.AWSRolesRef) []ocm.OperatorIAM
...
+		return creator, nil // same account — no cross-account assumption needed
+	}
+
+	// Assume the InstallerRole in the target account using the current session credentials.
+	// This yields a real sts:AssumeRole identity whose ARN and AccountID are both in the
+	// target account, satisfying all OCM creator validations.
+	session := rosaScope.Session()
+	stsSvc := stsv2sdk.NewFromConfig(session)
+	assumeRoleProvider := stscreds.NewAssumeRoleProvider(stsSvc, installerRoleARN, func(o *stscreds.AssumeRoleOptions) {
+		o.RoleSessionName = "capa-session"
```
andclt: The duration is not set here, the default one is 15 mins, should we bump it?
Should we also log when we assume the InstallerRole?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1361,6 +1366,90 @@ func operatorIAMRoles(rolesRef rosacontrolplanev1.AWSRolesRef) []ocm.OperatorIAM
...
+	}
+
+	return targetCreator, nil
+}
+
+// accountIDFromRoleARN parses the AWS account ID from a role ARN using the AWS ARN parser.
+func accountIDFromRoleARN(roleARN string) string {
+	parsed, err := arn.Parse(roleARN)
+	if err != nil {
+		return ""
```
andclt: should we log here if we cannot parse the roleARN?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1361,6 +1366,90 @@ func operatorIAMRoles(rolesRef rosacontrolplanev1.AWSRolesRef) []ocm.OperatorIAM
...
+// account are all the same (target) account. Assuming the InstallerRole gives a genuine
+// sts:assumed-role ARN in the target account, satisfying both OCM validations.
+func (r *ROSAControlPlaneReconciler) resolveCreatorForTargetAccount(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope, creator *rosaaws.Creator) (*rosaaws.Creator, error) {
+	installerRoleARN := rosaScope.ControlPlane.Spec.InstallerRoleARN
+	if rosaScope.ControlPlane.Spec.RosaRoleConfigRef != nil {
+		rosaRoleConfig := &expinfrav1.ROSARoleConfig{}
+		if err := r.Client.Get(ctx, client.ObjectKey{
+			Name:      rosaScope.ControlPlane.Spec.RosaRoleConfigRef.Name,
+			Namespace: rosaScope.ControlPlane.Namespace,
+		}, rosaRoleConfig); err == nil {
```
andclt: Could there be a scenario where `ROSARoleConfig` exists but its status is not fully populated yet (example: still reconciling), so that we unconditionally overwrite `installerRoleARN` with an empty string from the status even if `spec.InstallerRoleARN` had a valid value?

andclt (commented):
Could we add an integration test verifying that when `identityRef` changes between reconciliations, a new AWS client is created with the new identity?

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: There is a webhook validation to prevent setting the spec.*Roles* and spec.RosaRoleConfigRef at same time

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: done

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: why we would need to bump it ?

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: done

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: added new test case with AWSClusterRoleIdentity

controlplane/rosa/controllers/rosacontrolplane_controller.go
andclt: Oh good point, thanks!

controlplane/rosa/controllers/rosacontrolplane_controller.go
andclt: In the scenario where cluster creation takes more than 15 mins.
Still think the logging would be valuable here though - we could also make the `RoleSessionName` more descriptive.

controlplane/rosa/controllers/rosacontrolplane_controller.go
andclt: The new test case verifies identity resolution, however, should we include also the tests to validate the cross-account functionality?
examples:
1. cross account scenario: management cluster identity in account A, `InstallerRoleARN` in account B, verify `resolveCreatorForTargetAccount` assumes the role and returns creator with AccountID B.
2. missing `InstallerRoleARN`: validating that the original creator is returned when `InstallerRoleARN` is not present.
3. when using `spec.RosaRoleConfigRef`, verify the installer role ARN is correctly read from `ROSARoleConfig.Status`

controllers/rosacluster_controller.go
andclt: Do we still need this?

andclt: Do we still need this?

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -63,7 +64,14 @@ type ROSARoleConfigReconciler struct {
 	WatchFilterValue string
 	NewStsClient     func(cloud.ScopeUsage, cloud.Session, logger.Wrapper, runtime.Object) stsiface.STSClient
```
andclt: Do we still need this?

exp/controllers/rosaroleconfig_controller.go
andclt: Do we still need this?

andclt (commented):
Overall the PR LGTM, just left some comments

controllers/rosacluster_controller.go
serngawy: ops, good point. No need as the session established in the controlplane now

serngawy: removed

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: I added cluster name to the session name. Not sure, what we should log ? usually controller log cluster phase eg; initialize, create (waiting to be ready), upgrade and delete.

controlplane/rosa/controllers/rosacontrolplane_controller.go
serngawy: I added TestResolveCreatorForTargetAccount that test resolveCreatorForTargetAccount with multiple aws accounts

exp/controllers/rosaroleconfig_controller.go
serngawy: removed

andclt (approved):
/approve
/lgtm

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1361,6 +1366,103 @@ func operatorIAMRoles(rolesRef rosacontrolplanev1.AWSRolesRef) []ocm.OperatorIAM
...
+
+	stsSvc := stsv2sdk.NewFromConfig(session)
+	assumeRoleProvider := stscreds.NewAssumeRoleProvider(stsSvc, installerRoleARN, func(o *stscreds.AssumeRoleOptions) {
+		o.RoleSessionName = fmt.Sprintf("%s-%s", "capa-session", rosaScope.ControlPlane.Spec.RosaClusterName)
+	})
+
+	// Build targetSession based on old session default values and load the new credential.
+	targetSession := session.Copy()
+	targetSession.Credentials = awsv2.NewCredentialsCache(assumeRoleProvider)
+
```
tinaafitz, line 1424: The `err != nil `check can be removed here.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1361,6 +1366,103 @@ func operatorIAMRoles(rolesRef rosacontrolplanev1.AWSRolesRef) []ocm.OperatorIAM
...
+		return creator, nil // same account — no cross-account assumption needed
+	}
+
+	session := rosaScope.Session()
+	// Fall back to the control plane spec region when the session carries no region
+	// (e.g. when the scope is constructed without a full AWS session in tests).
+	if session.Region == "" {
+		session.Region = rosaScope.ControlPlane.Spec.Region
+	}
+
```
tinaafitz, line 1413: It would be helpful to add a log statement before the cross-account InstallerRole assumption:

`rosaScope.Info("Assuming InstallerRole for cross-account deployment", "targetAccount", targetAccountID, "installerRoleARN", installerRoleARN)`

richardcase:
When you are happy with this just unhold

/hold

richardcase:
Looks like @andclt was happy with these changes. So adding the:

/approve

Unhold when the Rosa folk are happy with this

serngawy:
Hi @damdo would you review/lgtm this PR

PR #6046, opened 2026-06-11
feat: add support for OCM role creation

andclt:
**What type of PR is this?**

Fixes #

**AI Usage**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title
- [ ] adds unit tests

**Release note**:

```release-note

```


exp/controllers/ocmroleconfig_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	}()
+
+	if err := r.setUpRuntime(ctx, scope); err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to set up runtime: %w", err)
+	}
+
+	// Note: We do NOT perform any cleanup on deletion because OCM roles are
+	// singleton resources shared across clusters. Automated deletion could break
+	// all clusters using this role.
+	if !scope.OCMRoleConfig.DeletionTimestamp.IsZero() {
```
robpblake: @andclt So how does a customer delete an `OCM Role` once created?

I think delete of the resource should unlink and delete the `OCM Role`? Customers will be prompted to recreate it when they try to interact with Cluster Service

exp/api/v1beta2/ocmroleconfig_types.go
```diff
@@ -0,0 +1,150 @@
...
+	// PermissionsBoundaryARN is the ARN of the policy that is used to set the permissions boundary for the OCM role.
+	// +optional
+	PermissionsBoundaryARN string `json:"permissionsBoundaryARN,omitempty"`
+
+	// Path is the IAM path for the OCM role.
+	// +optional
+	// +kubebuilder:default="/"
+	Path string `json:"path,omitempty"`
+
+	// ManagedPolicies indicates whether to use AWS-managed policies (true) or customer-managed policies (false).
```
robpblake: @andclt No need to expose this at this time. We only support customer managed policies on the OCM Role today. We can revisit if/when they move to managed policies.

exp/api/v1beta2/rosaocmroleconfig_types.go
```diff
@@ -0,0 +1,150 @@
...
+	Profile OCMRoleProfile `json:"profile"`
+
+	// PermissionsBoundaryARN is the ARN of the policy that is used to set the permissions boundary for the OCM role.
+	// +optional
+	PermissionsBoundaryARN string `json:"permissionsBoundaryARN,omitempty"`
+
+	// Path is the IAM path for the OCM role.
+	// +optional
+	// +kubebuilder:default="/"
+	Path string `json:"path,omitempty"`
```
robpblake, line 85: @andclt Can you validate this starts with a `/`?

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,268 @@
...
+	// Convert profile from CRD enum to ROSA CLI enum
+	rosaProfile, err := r.convertProfile(ocmRoleConfig.Spec.Profile)
+	if err != nil {
+		v1beta1conditions.MarkFalse(ocmRoleConfig, expinfrav1.OCMRoleConfigReadyCondition,
+			expinfrav1.OCMRoleConfigReconciliationFailedReason, clusterv1beta1.ConditionSeverityError,
+			"Invalid profile: %s", err.Error())
+		return ctrl.Result{}, err
+	}
+
+	roleARN, created, err := ocmrole.GetOrCreateOCMRole(
```
robpblake, line 163: @andclt This is where the mutex is required due to the nature of the API:

- read
- append
- write back

Two reconciles happening at the same time can interleave for the same organization. You need to protect the whole read, create, update flow to ensure it happens sequentially.

robpblake: @andclt As we discussed in Slack. This is wrong and where you have the current mutex is correct.

Apologies for the confusion.

cheers,

Rob

andclt: Thanks!

exp/controllers/ocmroleconfig_controller.go
andclt: As discussed on Slack, deleting the `OCM role` when deleting the CR instance is the right approach here since the OCM role deletion won't break running clusters - will fix it.

exp/api/v1beta2/rosaocmroleconfig_types.go
andclt: sure - fixed in latest commit

exp/api/v1beta2/ocmroleconfig_types.go
andclt: sure - fixed in latest commit

exp/api/v1beta2/ocmroleconfig_types.go
```diff
@@ -0,0 +1,146 @@
...
+	OCMRoleConfigLinkedReason = "Linked"
+)
+
+// OCMRoleConfigSpec defines the desired state of OCMRoleConfig
+type OCMRoleConfigSpec struct {
+	// RolePrefix is the user-defined prefix for the OCM role name.
+	// The final role name will be: {RolePrefix}-OCM-Role-{ExternalID}
+	// where ExternalID is the organization's external identifier from OCM.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength:=32
```
serngawy: I would make the max 4-6 to avoid long name case

exp/api/v1beta2/ocmroleconfig_types.go
```diff
@@ -0,0 +1,146 @@
...
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="profile is immutable"
+	Profile OCMRoleProfile `json:"profile"`
+
+	// PermissionsBoundaryARN is the ARN of the policy that is used to set the permissions boundary for the OCM role.
+	// +optional
+	PermissionsBoundaryARN string `json:"permissionsBoundaryARN,omitempty"`
+
+	// Path is the IAM path for the OCM role.
+	// +optional
+	// +kubebuilder:default="/"
```
serngawy: no need for default as its an option

exp/api/v1beta2/ocmroleconfig_types.go
```diff
@@ -0,0 +1,146 @@
...
+// OCMRoleConfigStatus defines the observed state of OCMRoleConfig
+type OCMRoleConfigStatus struct {
+	// RoleARN is the ARN of the created OCM role.
+	RoleARN string `json:"roleARN,omitempty"`
+
+	// OrganizationID is the OCM organization ID that this role is linked to.
+	OrganizationID string `json:"organizationID,omitempty"`
+
+	// Linked indicates whether the role has been successfully linked to the OCM organization.
+	Linked bool `json:"linked,omitempty"`
```
serngawy: I don't think there is a need to show this flag. We should  show error in link condition if we couldn't link it

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,284 @@
...
+)
+
+// OCMRoleConfigReconciler reconciles an OCMRoleConfig object.
+type OCMRoleConfigReconciler struct {
+	client.Client
+	Recorder         record.EventRecorder
+	WatchFilterValue string
+	NewStsClient     func(cloud.ScopeUsage, cloud.Session, logger.Wrapper, runtime.Object) stsiface.STSClient
+	NewOCMClient     func(ctx context.Context, scope caparosa.OCMSecretsRetriever) (caparosa.OCMClient, error)
+	Runtime          *rosacli.Runtime
```
serngawy, line 57: this need to change as we are moving to cross account creation. The runtime need to be established with the reconciling. check my PR [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6038/changes)

exp/api/v1beta2/ocmroleconfig_types.go
```diff
@@ -0,0 +1,146 @@
...
+	// Conditions specifies the OCMRoleConfig conditions
+	Conditions clusterv1beta1.Conditions `json:"conditions,omitempty"`
+}
+
+// OCMRoleConfig is the Schema for the ocmroleconfigs API
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=ocmroleconfigs,scope=Cluster,categories=cluster-api,shortName=ocmrole
+// +kubebuilder:storageversion
+// +kubebuilder:subresource:status
+type OCMRoleConfig struct {
```
serngawy: lets name it as ROSAOCMRoleConfig to follow all other ROSA APIs naming. Also for all struct, types and constant let them all start with ROSA

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,284 @@
...
+		scope.Info("Using existing OCM role", "roleARN", roleARN)
+	}
+
+	// Update status with role ARN
+	ocmRoleConfig.Status.RoleARN = roleARN
+
+	// Acquire per-organization mutex to prevent concurrent LinkOrgToRole races
+	// Multiple OCMRoleConfigs with different AWS accounts can link to the same org,
+	// causing read-modify-write races on the org label. Serialize linking per org.
+	linkMu := r.getLinkMutex(orgID)
```
serngawy, line 192: can you elaborate more why this is needed ?  how this statement " Multiple OCMRoleConfigs with different AWS accounts can link to the same org" could happen ? I think it should be; first role linked wine others will raise error cannot link.

exp/controllers/ocmroleconfig_controller.go
serngawy: I believe the right behavior is to unlink then delete the ocmRole AND provide optional field eg; `skipDelete`  to let the user skip deleting Role just delete CR.

serngawy: Also we should have finalizer to handle any error happen during delete

serngawy (commented):
some comments and please add release note in the PR desc.

exp/controllers/rosaocmroleconfig_controller.go
robpblake: The OCM Role is used to link an OCM Org to the AWS Account(s) that they use for cluster provisioning. An OCM org can use many AWS accounts, so it should be possible to link an OCM Role as long as there is not one already linked for the same AWS account.

robpblake: This is the same functionality as offered today by the CLI in `rosa create ocm-role`

andclt: Yes, as Rob said, we can have multiple AWS accounts associated to the same OCM organization by defining multiple `OCMRoleConfig` with different `IdentityRef`.

exp/controllers/ocmroleconfig_controller.go
andclt: yes I will re-add the finalizers as well, I removed them because we didn't have an actual ReconcileDelete function.

exp/api/v1beta2/ocmroleconfig_types.go
andclt: Makes sense, the `Ready` condition already captures whether the role was successfully linked or not.

exp/api/v1beta2/ocmroleconfig_types.go
andclt: I initially wanted to keep it consistent with the ROSA CLI (which allows up to 32 chars for all role prefixes), but I see that `ROSARoleConfig` already limits both `AccountRoleConfig` and `OperatorRoleConfig`
prefixes to 4 characters, so I am fine to change it to 4 chars.

andclt:
> I believe the right behavior is to unlink then delete the ocmRole AND provide optional field eg; skipDelete to let the user skip deleting Role just delete CR.


@serngawy, could you share more about the optional field use case? What's the scenario where someone wants to delete the CR but keep the IAM role?

andclt:
Closing in favor of: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6055
We could not update the ROSA dependency directly due to Go directive version incompatibility between ROSA and CAPA, so the logic was ported and adapted to work within CAPA's architecture.
The comments have been solved in the new PR.

PR #6049, opened 2026-06-12
✨ feat(ROSA): add optional sts:ExternalId support for ROSA account roles and control plane

### LINKED ISSUE #6048: Add STS external ID support for ROSA
/kind feature

Add support for `sts:ExternalId` to CAPA's ROSA (HCP) implementation, aligned with the behavior of the ROSA CLI and Terraform provider. Neither `ROSARoleConfig` (for account role creation) and `ROSAControlPlane` (for cluster creation) accept an external ID today, even though STS external ID is supported by the OCM API.

This is opt-in only: an External ID is used only when it is set explicitly. 

Proposed behavior is:

Add an optional, immutable field on `AccountRoleConfig` called `trustPolicyExternalID`:
```yaml
spec:
  accountRoleConfig:
    prefix: capa
    version: "4.20.11"
    trustPolicyExternalID: "example-external-id"
```

And add an optional, immutable field on `ROSAControlPlane` called `trustPolicyExternalID`:

```yaml
spec:
  rosaRoleConfigRef:
    name: role-config
  trustPolicyExternalID: "example-external-id"
```

**Environment:**

- Cluster-api-provider-aws version: `main`
- Kubernetes version: (use `kubectl version`): `any supported k8s version`
- OS (e.g. from `/etc/os-release`): `any supported OS`

michaelryanmcneill:
**What type of PR is this?**
/kind feature
/kind api-change
/kind documentation

Adds an optional `trustPolicyExternalID` field to `ROSARoleConfig` and `ROSAControlPlane` CRDs. When set on `ROSARoleConfig`, the controller embeds an `sts:ExternalId` condition in installer and support role trust policies after role creation, and the value is automatically propagated to OCM during cluster creation. When set directly on `ROSAControlPlane` (without `rosaRoleConfigRef`), the value is passed to OCM but trust policies are not modified and the user is responsible for configuring their pre-existing roles.

Fixes #6048

- Trust policy injection is implemented directly via the IAM API (`GetRole` + `UpdateAssumeRolePolicy`) rather than relying on upstream `openshift/rosa` library support, which would have required bumping the Go version from 1.24.0 to 1.25.8. This means that no dependency changes are needed to support this feature. There is a small downside to this: between role creation and the trust policy update, there is a brief window where the role exists without the `sts:ExternalId` condition. This is less than a second and, as the role was just created, I believe the exposure is minimal.
- When using `rosaRoleConfigRef`, the external ID is automatically read from the `ROSARoleConfig` and passed to OCM, so users only need to declare it once.
- When providing role ARNs directly on `ROSAControlPlane` (no `rosaRoleConfigRef`), CAPA only passes the value to OCM; it does not modify externally-created roles' trust policies.
- The field is immutable once set, enforced via CEL validation.
- The worker and Shared VPC account roles are intentionally excluded from the external ID condition.

**AI Usage**:

Cursor with Claude was used to assist with test authoring, implementation of the trust policy injection helper, and evaluating approaches to avoid the Go version bump.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
Add optional `trustPolicyExternalID` field to ROSARoleConfig and ROSAControlPlane for embedding sts:ExternalId conditions in ROSA HCP account role trust policies. When using ROSARoleConfig, the controller manages the trust policies automatically. When providing role ARNs directly, the value is passed to OCM and users are responsible for properly configuring their pre-existing roles.
```

### DIFF
--- a/controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
+++ b/controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
@@ -226,6 +226,23 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// +optional
 	WorkerRoleARN string `json:"workerRoleARN,omitempty"`
 
+	// TrustPolicyExternalID is an optional STS external ID that OCM will use when assuming
+	// the installer and support account roles.
+	// When using RosaRoleConfigRef, this field is ignored, the value is always read from
+	// the ROSARoleConfig, which owns the trust policies.
+	// When providing role ARNs directly (no RosaRoleConfigRef), the user is responsible for
+	// including the sts:ExternalId condition in the roles' trust policies; this field only
+	// tells OCM which external ID to present when assuming the roles.
+	// Worker roles are not affected.
+	// Must be 2–1224 characters matching [a-zA-Z0-9=,.@:/-]+ per AWS STS requirements.
+	//
+	// +kubebuilder:validation:MinLength=2
+	// +kubebuilder:validation:MaxLength=1224
+	// +kubebuilder:validation:Pattern=`^[a-zA-Z0-9=,.@:\/-]+$`
+	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="trustPolicyExternalID is immutable"
+	// +optional
+	TrustPolicyExternalID string `json:"trustPolicyExternalID,omitempty"`
+
 	// BillingAccount is an optional AWS account to use for billing the subscription fees for ROSA HCP clusters.
 	// The cost of running each ROSA HCP cluster will be billed to the infrastructure account in which the cluster
 	// is running.
--- a/controlplane/rosa/controllers/rosacontrolplane_controller.go
+++ b/controlplane/rosa/controllers/rosacontrolplane_controller.go
@@ -405,6 +405,7 @@ func (r *ROSAControlPlaneReconciler) reconcileRosaRoleConfig(ctx context.Context
 		rosaRoleConfig.Status.AccountRolesRef.SupportRoleARN = rosaScope.ControlPlane.Spec.SupportRoleARN
 		rosaRoleConfig.Status.AccountRolesRef.WorkerRoleARN = rosaScope.ControlPlane.Spec.WorkerRoleARN
 		rosaRoleConfig.Status.OperatorRolesRef = rosaScope.ControlPlane.Spec.RolesRef
+		rosaRoleConfig.Spec.AccountRoleConfig.TrustPolicyExternalID = rosaScope.ControlPlane.Spec.TrustPolicyExternalID
 	}
 
 	return rosaRoleConfig, nil
@@ -1199,6 +1200,7 @@ func buildOCMClusterSpec(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpe
 		AuditLogRoleARN:              ptr.To(controlPlaneSpec.AuditLogRoleARN),
 		ExternalAuthProvidersEnabled: controlPlaneSpec.EnableExternalAuthProviders,
 		FIPS:                         controlPlaneSpec.FIPS == rosacontrolplanev1.FIPSEnabled,
+		ExternalID:                   roleConfig.Spec.AccountRoleConfig.TrustPolicyExternalID,
 	}
 
 	if controlPlaneSpec.EndpointAccess == rosacontrolplanev1.Private {
--- a/controlplane/rosa/controllers/rosacontrolplane_controller_test.go
+++ b/controlplane/rosa/controllers/rosacontrolplane_controller_test.go
@@ -963,4 +963,96 @@ func TestBuildOCMClusterSpec(t *testing.T) {
 		g.Expect(ocmSpec.FIPS).To(BeFalse())
 		g.Expect(ocmSpec.Name).To(Equal("test-cluster-zero-fips"))
 	})
+
+	// Test case 4: TrustPolicyExternalID set on ROSARoleConfig (takes precedence)
+	t.Run("TrustPolicyExternalID Set", func(t *testing.T) {
+		g := NewWithT(t)
+		roleConfigWithExtID := mockRoleConfig.DeepCopy()
+		roleConfigWithExtID.Spec.AccountRoleConfig.TrustPolicyExternalID = "223B9588-36A5-ECA4-BE8D-7C673B77CEC1"
+
+		controlPlaneSpec := rosacontrolplanev1.RosaControlPlaneSpec{
+			RosaClusterName:   "test-cluster-extid",
+			Region:            "us-west-2",
+			Version:           "4.14.5",
+			Subnets:           []string{"subnet-1", "subnet-2"},
+			AvailabilityZones: []string{"us-west-2a"},
+			DefaultMachinePoolSpec: rosacontrolplanev1.DefaultMachinePoolSpec{
+				InstanceType: "m5.xlarge",
+			},
+		}
+
+		ocmSpec, err := buildOCMClusterSpec(controlPlaneSpec, roleConfigWithExtID, nil, mockCreator)
+
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(ocmSpec.ExternalID).To(Equal("223B9588-36A5-ECA4-BE8D-7C673B77CEC1"))
+		g.Expect(ocmSpec.Name).To(Equal("test-cluster-extid"))
+	})
+
+	// Test case 5: TrustPolicyExternalID empty (should remain empty)
+	t.Run("TrustPolicyExternalID Empty", func(t *testing.T) {
+		g := NewWithT(t)
+		controlPlaneSpec := rosacontrolplanev1.RosaControlPlaneSpec{
+			RosaClusterName:   "test-cluster-no-extid",
+			Region:            "us-west-2",
+			Version:           "4.14.5",
+			Subnets:           []string{"subnet-1", "subnet-2"},
+			AvailabilityZones: []string{"us-west-2a"},
+			DefaultMachinePoolSpec: rosacontrolplanev1.DefaultMachinePoolSpec{
+				InstanceType: "m5.xlarge",
+			},
+		}
+
+		ocmSpec, err := buildOCMClusterSpec(controlPlaneSpec, mockRoleConfig, nil, mockCreator)
+
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(ocmSpec.ExternalID).To(BeEmpty())
+	})
+
+	// Test case 6: TrustPolicyExternalID on controlPlane is ignored when ROSARoleConfig is present
+	t.Run("TrustPolicyExternalID ControlPlane Ignored With RoleConfig", func(t *testing.T) {
+		g := NewWithT(t)
+		controlPlaneSpec := rosacontrolplanev1.RosaControlPlaneSpec{
+			RosaClusterName:       "test-cluster-cp-extid",
+			Region:                "us-west-2",
+			Version:               "4.14.5",
+			TrustPolicyExternalID: "should-be-ignored",
+			Subnets:               []string{"subnet-1", "subnet-2"},
+			AvailabilityZones:     []string{"us-west-2a"},
+			DefaultMachinePoolSpec: rosacontrolplanev1.DefaultMachinePoolSpec{
+				InstanceType: "m5.xlarge",
+			},
+		}
+
+		ocmSpec, err := buildOCMClusterSpec(controlPlaneSpec, mockRoleConfig, nil, mockCreator)
+
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(ocmSpec.ExternalID).To(BeEmpty())
+	})
+
+	// Test case 7: TrustPolicyExternalID from controlPlane propagated via synthesized roleConfig
+	// (simulates direct-role-ARN flow where reconcileRosaRoleConfig copies the value)
+	t.Run("TrustPolicyExternalID ControlPlane Without RoleConfig", func(t *testing.T) {
+		g := NewWithT(t)
+		controlPlaneSpec := rosacontrolplanev1.RosaControlPlaneSpec{
+			RosaClusterName:       "test-cluster-direct",
+			Region:                "us-west-2",
+			Version:               "4.14.5",
+			TrustPolicyExternalID: "direct-external-id",
+			Subnets:               []string{"subnet-1", "subnet-2"},
+			AvailabilityZones:     []string{"us-west-2a"},
+			DefaultMachinePoolSpec: rosacontrolplanev1.DefaultMachinePoolSpec{
+				InstanceType: "m5.xlarge",
+			},
+		}
+
+		// Simulate what reconcileRosaRoleConfig does in the else branch:
+		// it copies TrustPolicyExternalID from the control plane spec onto the roleConfig
+		synthesizedRoleConfig := mockRoleConfig.DeepCopy()
+		synthesizedRoleConfig.Spec.AccountRoleConfig.TrustPolicyExternalID = controlPlaneSpec.TrustPolicyExternalID
+
+		ocmSpec, err := buildOCMClusterSpec(controlPlaneSpec, synthesizedRoleConfig, nil, mockCreator)
+
+		g.Expect(err).ToNot(HaveOccurred())
+		g.Expect(ocmSpec.ExternalID).To(Equal("direct-external-id"))
+	})
 }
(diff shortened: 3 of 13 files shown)


michaelryanmcneill:
This has been fixed in #6051

~~I noticed in the docs preview that the CRD reference wasn't being properly updated. Working with Claude, it identified:~~
>The docs/book/Makefile defines the dependency tracking for `src/crd/index.md at` line 41:
>```
>src/crd/index.md: $(API_FILTERED_SRCS) src/crd
>```
>`API_FILTERED_SRCS` is derived from `API_DIRS` (line 28), which only lists `v1beta1` directories:
>```
>API_DIRS := cmd/clusterawsadm/api/bootstrap/v1beta1 api/v1beta1 exp/api/v1beta1 controlplane/eks/api/v1beta1 bootstrap/eks/api/v1beta1 iam/api/v1beta1
>```
>The v1beta2 directories (`exp/api/v1beta2/`, `controlplane/rosa/api/v1beta2/`) are not in this list. So Make sees no source files have changed and skips regeneration.

~~Running the full generator target directly:~~
```
make -C docs/book gen_crd_docs API_DIR="$(go list -m)" OUT_FILE=src/crd/index.md
```
~~results in a significant number of changes to that file specifically that appear to have been previously left out. I am going to leave them in for now since they should be in there, but am glad to remove it or place it in a separate issue and PR.~~

go.mod
```diff
@@ -1,6 +1,6 @@
 module sigs.k8s.io/cluster-api-provider-aws/v2
 
-go 1.24.0
+go 1.25.8
```
damdo: We can't do this ATM.

See: https://kubernetes.slack.com/archives/CD6U2V71N/p1781190298231099

michaelryanmcneill: Quite timely conversation from yesterday! Thanks @damdo, I'm going to think about a separate implementation plan.

michaelryanmcneill: I have gone ahead and implemented an alternative path that doesn't bump dependencies (including go). Will resolve this, but would love to get your thoughts @damdo.

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -226,6 +226,16 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// +optional
 	WorkerRoleARN string `json:"workerRoleARN,omitempty"`
 
+	// TrustPolicyExternalID is an optional STS external ID that OCM uses when assuming the installer
+	// and support account roles. Must match the sts:ExternalId condition in those roles' trust policies.
+	//
+	// +kubebuilder:validation:MinLength=2
```
serngawy, line 239: not sure if min=2 is right ? can you add example of the expected ID

docs/book/src/topics/rosa/creating-a-cluster.md
```diff
@@ -96,6 +97,7 @@ The CAPA controller requires service account credentials to provision ROSA HCP c
       accountRoleConfig:
         prefix: "rosa"
         version: "4.20.11"
+        # trustPolicyExternalID: "my-external-id"  # optional
```
serngawy:
```suggestion
       # trustPolicyExternalID: "my-external-id"   optional 
```

docs/book/src/topics/rosa/creating-a-cluster.md
```diff
@@ -256,6 +258,7 @@ The CAPA controller requires service account credentials to provision ROSA HCP c
       domainPrefix: rosa-hcp
       rosaRoleConfigRef:
         name: role-config  # reference to the ROSARoleConfig created above
+      # trustPolicyExternalID: "my-external-id"  # optional, must match the ROSARoleConfig value
```
serngawy: Why do I need to set the ROSAControlPlane.spec.trustPolicyExternalID while I'm referring to rosaRoleConfigRef that is already has ref to trustPolicyExternalID ?

exp/api/v1beta2/rosaroleconfig_types.go
```diff
@@ -141,6 +141,17 @@ type AccountRoleConfig struct {
 	// SharedVPCConfig is used to set up shared VPC.
 	// +optional
 	SharedVPCConfig SharedVPCConfig `json:"sharedVPCConfig,omitempty"`
+
+	// TrustPolicyExternalID is an optional STS external ID to embed in installer and support
```
serngawy: Would you make the desc for TrustPolicyExternalID field similar at ROSARoleConfig and ROSACOntrolPlane

pkg/rosa/trust_policy.go
serngawy, line 1: please move this logic to https://github.com/kubernetes-sigs/cluster-api-provider-aws/tree/main/pkg/rosa so it can be re-used in other controllers if needed

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -134,7 +134,7 @@ func (r *ROSARoleConfigReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 		return ctrl.Result{}, err
 	}
 
-	if err := r.reconcileAccountRoles(scope); err != nil {
+	if err := r.reconcileAccountRoles(ctx, scope); err != nil {
```
serngawy: please keep the reconcileAccountRoles as it is cause accountRoles need to be created first anyway. Then check if there is  AccountRoleConfig.TrustPolicyExternalID exist so we can apply the logic of applyTrustPolicyExternalID

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -332,10 +332,24 @@ func (r *ROSARoleConfigReconciler) reconcileAccountRoles(scope *scope.RosaRoleCo
...
-	return accountroles.CreateHCPRoles(r.Runtime, scope.RosaRoleConfig.Spec.AccountRoleConfig.Prefix, true, scope.RosaRoleConfig.Spec.AccountRoleConfig.PermissionsBoundaryARN,
+	if err := accountroles.CreateHCPRoles(r.Runtime, scope.RosaRoleConfig.Spec.AccountRoleConfig.Prefix, true, scope.RosaRoleConfig.Spec.AccountRoleConfig.PermissionsBoundaryARN,
 		rosa.GetOCMClientEnv(r.Runtime.OCMClient), policies, scope.RosaRoleConfig.Spec.AccountRoleConfig.Version, scope.RosaRoleConfig.Spec.AccountRoleConfig.Path,
 		scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.IsSharedVPC(), scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.RouteRoleARN,
-		scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.VPCEndpointRoleARN)
+		scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.VPCEndpointRoleARN); err != nil {
+		return err
+	}
+
+	if externalID := scope.RosaRoleConfig.Spec.AccountRoleConfig.TrustPolicyExternalID; externalID != "" {
```
serngawy: lets move this logic from reconcileAccountRoles and call applyTrustPolicyExternalID after  reconcileAccountRoles is done.

serngawy (changes requested):
Having some comments AND I have question; for ROSAControlPlane only (no ref to ROSARoleConfig) when setting the `ROSAControlPlane.spec.InstallerRoleARN` AND `ROSAControlPlane.spec.TrustPolicyExternalID` what should happen ? should we applyTrustPolicyExternalID ?

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
michaelryanmcneill: Per the [AWS documentation](https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_common-scenarios_third-party.html#:~:text=The%20ExternalId%20value%20must%20have%20a%20minimum%20of%202%20characters%20and%20a%20maximum%20of%201%2C224%20characters.%20The%20value%20must%20be%20alphanumeric%20without%20white%20space.%20It%20can%20also%20include%20the%20following%20symbols%3A%20plus%20(%2B)%2C%20equal%20(%3D)%2C%20comma%20(%2C)%2C%20period%20(.)%2C%20at%20(%40)%2C%20colon%20(%3A)%2C%20forward%20slash%20(/)%2C%20and%20hyphen%20(%2D).), minimum 2 characters is correct. It could be any value, many are something along the lines of `12345-ABCDE-678910`.

docs/book/src/topics/rosa/creating-a-cluster.md
michaelryanmcneill: That's a good catch. We shouldn't make a user have to declare it twice. Updated.

michaelryanmcneill:
> I have question; for ROSAControlPlane only (no ref to ROSARoleConfig) when setting the ROSAControlPlane.spec.InstallerRoleARN AND ROSAControlPlane.spec.TrustPolicyExternalID what should happen ? should we applyTrustPolicyExternalID ?

This was a great call out @serngawy. With the latest changes, if a user sets `ROSAControlPlane.spec.InstallerRoleARN` _and_ `ROSAControlPlane.spec.TrustPolicyExternalID` (without a `rosaRoleConfigRef` set), CAPA only passes the external ID value to OCM in the cluster creation spec. It does not modify the roles' trust policies.

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
michaelryanmcneill: @serngawy re-reading this: did you want me to add an example to the comment in the code? Just want to make sure I didn't misunderstand!

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1316,6 +1317,18 @@ func buildOCMClusterSpec(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpe
 	return ocmClusterSpec, nil
 }
 
+// trustPolicyExternalID returns the effective STS external ID, preferring the value
+// from the ROSAControlPlane spec, falling back to the ROSARoleConfig when available.
```
serngawy: The logic should check the rosaroleconfig.externalId first then rosacontroplane.externalId . When the rosaroleConfig is available the  rosacontroplane.externalId will be ignored, right ?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1316,6 +1317,18 @@ func buildOCMClusterSpec(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpe
...
+func trustPolicyExternalID(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpec, roleConfig *expinfrav1.ROSARoleConfig) string {
```
serngawy: Something like below.

```suggestion
func trustPolicyExternalID(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpec, roleConfig *expinfrav1.ROSARoleConfig) string {

if RosaControlPlaneSpec.RosaRoleConfigRef != nil {
    return roleConfig.TrustPolicyExternalID
 }
 
 return  controlPlaneSpec.TrustPolicyExternalID

}
```

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -226,6 +226,23 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// +optional
 	WorkerRoleARN string `json:"workerRoleARN,omitempty"`
 
+	// TrustPolicyExternalID is an optional STS external ID that OCM will use when assuming
+	// the installer and support account roles.
+	// When using RosaRoleConfigRef, this field is optional; the value is read from the
```
serngawy:
```suggestion
	// When setting RosaRoleConfigRef, this filled will be ignored as the referral RosaRoleConfig.Spec.TrustPolicyExternalID will be in use.
```

serngawy (commented):
couple of comments

controlplane/rosa/controllers/rosacontrolplane_controller.go
michaelryanmcneill: Yeah, that makes sense. I went ahead and changed that, and also added a warning in the controller logs if they're both defined, to say "ROSAControlPlane.spec.trustPolicyExternalID is ignored when rosaRoleConfigRef is set; using value from ROSARoleConfig".

michaelryanmcneill:
~~It appears the pull-cluster-api-provider-aws-e2e-blocking prow job is failing to find a host, not failing on the e2e itself. Will try a retest tomorrow.~~ Tests passed. I think we're good for final review.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -346,6 +346,14 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		}
 	}
 
+	if rosaScope.ControlPlane.Spec.RosaRoleConfigRef != nil &&
+		rosaScope.ControlPlane.Spec.TrustPolicyExternalID != "" &&
+		rosaRoleConfig != nil && rosaRoleConfig.Spec.AccountRoleConfig.TrustPolicyExternalID != "" {
```
serngawy: better to prevent this from happening at webhook add check [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/webhooks/rosacontrolplane_webhook.go#L233) as 
` || r.Spec.TrustPolicyExternalID != ""`

controlplane/rosa/controllers/rosacontrolplane_controller_test.go
```diff
@@ -963,4 +963,93 @@ func TestBuildOCMClusterSpec(t *testing.T) {
...
+			Version:               "4.14.5",
+			TrustPolicyExternalID: "direct-external-id",
+			Subnets:               []string{"subnet-1", "subnet-2"},
+			AvailabilityZones:     []string{"us-west-2a"},
+			DefaultMachinePoolSpec: rosacontrolplanev1.DefaultMachinePoolSpec{
+				InstanceType: "m5.xlarge",
+			},
+		}
+
+		// In the real flow, reconcileRosaRoleConfig never returns nil;
```
serngawy: Oh, so the logic need to be simpler. No need for this function [trustPolicyExternalID](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6049/changes#diff-693a964b12104ea3c2432ded8cbb95b43bf89782beb8d202e32dcf6422413e7cR1331) . At the [reconcileRosaRoleConfig](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/controllers/rosacontrolplane_controller.go#L371C38-L371C61) func [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/controllers/rosacontrolplane_controller.go#L402) set
`rosaRoleConfig.Spec.AccountRoleConfig.TrustPolicyExternalID = rosaScope.ControlPlane.Spec.TrustPolicyExternalID`

So you always have the TrustPolicyExternalID if ever defined.

What do you think ?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1199,6 +1207,7 @@ func buildOCMClusterSpec(controlPlaneSpec rosacontrolplanev1.RosaControlPlaneSpe
 		AuditLogRoleARN:              ptr.To(controlPlaneSpec.AuditLogRoleARN),
 		ExternalAuthProvidersEnabled: controlPlaneSpec.EnableExternalAuthProviders,
 		FIPS:                         controlPlaneSpec.FIPS == rosacontrolplanev1.FIPSEnabled,
+		ExternalID:                   trustPolicyExternalID(controlPlaneSpec, roleConfig),
```
serngawy: Based on my comment below, should be simple as this
```suggestion
		ExternalID:                   roleConfig.Spec.AccountRoleConfig.TrustPolicyExternalID,
```

controlplane/rosa/controllers/rosacontrolplane_controller_test.go
michaelryanmcneill: Totally makes sense, should be simpler. Fixed!

controlplane/rosa/controllers/rosacontrolplane_controller.go
michaelryanmcneill: Perfect, I'll do that. Thanks!

serngawy (approved):
Looks good to me, thanks for doing this @michaelryanmcneill

PR #6055, opened 2026-06-16
✨ feat: add support for OCM role creation

andclt:
**What type of PR is this?**
/kind feature

This PR adds support for managing OCM roles in AWS through a new `ROSAOCMRoleConfig` controller. This controller automates the creation, validation, and lifecycle management of OCM roles required for ROSA cluster operations.

**Key additions**:
-  New `ROSAOCMRoleConfig` CRD with validation rules (role prefix patterns, profile immutability, max lengths)
- Controller that reconciles OCM role resources and links them to OCM organizations
- Support for multiple role profiles (Admin, Standard, NoConsole)
- Self-healing capabilities for role tags and policies

The core `GetOrCreateOCMRole` logic and related helper methods are adapted from the ROSA CLI (https://github.com/openshift/rosa/pull/3262). We could not update the ROSA dependency directly due to Go directive version incompatibility between ROSA and CAPA, so the logic was ported and adapted to work within CAPA's architecture.

 **Manual Testing**:
Comprehensive manual testing was performed against real AWS and OCM infrastructure.
 
**Test Environment:**
  - Kind cluster v1.35.0 (capa-test)
  - CAPA controller built with Go 1.24 and ROSA feature gate enabled

  **tests:**
  - Standard & Admin profile creation with OCM linking
  - Idempotency (role reuse across multiple ROSAOCMRoleConfigs)
  - Self-healing (auto-restoration of missing admin tags)
  - Permissions boundary attachment validation
  - Custom IAM path support
  - Field immutability enforcement (profile, rolePrefix)
  - Singleton constraint (one OCM role per AWS account)
  - Auto-reconciliation with exponential backoff
  - Deletion flow with proper OCM unlink → AWS role delete sequence
  - Deletion flow with `deletePolicy=Retain`
  - Credential secret validation and error handling

Fixes
https://redhat.atlassian.net/browse/ROSAENG-8292

The `GetOrCreateOCMRole` implementation is adapted from ROSA CLI PR #3262 rather than using the ROSA dependency directly due to Go version incompatibility. The logic has been adapted to fit CAPA's controller pattern and error handling conventions.

**AI Usage**:

Claude code was used for test infrastructure refactoring and boilerplate elimination in test files.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
Add ROSAOCMRoleConfig controller for automated OCM role management in AWS, supporting Admin, Standard, and NoConsole role profiles.
```


pkg/rosa/awshelpers/constants.go
```diff
@@ -0,0 +1,63 @@
...
+	TagAdminRole = prefix + "admin_role"
+
+	// TagNoConsoleRole tags the role as no-console role.
+	TagNoConsoleRole = prefix + "no_console_role"
+
+	// TagRedHatManaged tags the role as red_hat_managed.
+	TagRedHatManaged = "red-hat-managed"
+)
+
+// JumpAccounts are the various AWS accounts used for the installer jump role in the various OCM environments.
```
serngawy: this info shouldn't be here, better to refer the rosa package  JumpAccounts

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,357 @@
...
+	client.Client
+	Recorder         record.EventRecorder
+	WatchFilterValue string
+	NewStsClient     func(cloud.ScopeUsage, cloud.Session, logger.Wrapper, runtime.Object) stsiface.STSClient
+	NewOCMClient     func(ctx context.Context, scope caparosa.OCMSecretsRetriever) (caparosa.OCMClient, error)
+	// runtimeFactory overrides runtime creation per reconciliation. Used in tests to inject mock clients.
+	runtimeFactory func(ctx context.Context, scope *scope.ROSAOCMRoleConfigScope) (*rosacli.Runtime, error)
+
+	// Per-organization linking mutex
+	orgLinkMutexes sync.Map // map[orgID]*sync.Mutex
```
serngawy, line 57: As we discussed, We may have no need for this shared Map. Please remove the mutex map logic and test it first with the cases we discussed.

serngawy (commented):
few comments

test/e2e/suites/unmanaged/unmanaged_functional_test.go
```diff
@@ -547,7 +547,8 @@ var _ = ginkgo.Context("[unmanaged] [functional]", func() {
 				}, e2eCtx.E2EConfig.GetIntervals("", "wait-machine-status")...).Should(BeTrue(), "Eventually failed waiting for AWSMachine to be labelled as terminated")
 
 				ginkgo.By("Waiting for machine to reach Failed state")
-				statusChecks := []framework.MachineStatusCheck{framework.MachinePhaseCheck(string(clusterv1.MachinePhaseFailed))} //nolint:staticcheck
+				//nolint:staticcheck // SA1019: MachinePhaseFailed is deprecated but still needed for v1beta1 compatibility
+				statusChecks := []framework.MachineStatusCheck{framework.MachinePhaseCheck(string(clusterv1.MachinePhaseFailed))}
```
andclt, line 551: unrelated to our changes but `make lint` was failing

pkg/rosa/awshelpers/constants.go
andclt: Agree, fixed!

exp/api/v1beta2/rosaocmroleconfig_types.go
```diff
@@ -0,0 +1,162 @@
...
+	ROSAOCMRoleConfigLinkedReason = "Linked"
+)
+
+// ROSAOCMRoleConfigSpec defines the desired state of ROSAOCMRoleConfig
+type ROSAOCMRoleConfigSpec struct {
+	// RolePrefix is the user-defined prefix for the OCM role name.
+	// The final role name will be: {RolePrefix}-OCM-Role-{ExternalID}
+	// where ExternalID is the organization's external identifier from OCM.
+	// +kubebuilder:validation:Required
+	// +kubebuilder:validation:MaxLength:=4
```
robpblake: @andclt I think ROSA CLI supports a prefix of up to 30 characters.

pkg/rosa/awshelpers/helpers.go
```diff
@@ -0,0 +1,112 @@
...
+func GetOCMRoleName(prefix, role, postfix string) string {
+	name := fmt.Sprintf("%s-%s-Role-%s", prefix, role, postfix)
+	return truncateRoleName(name)
+}
+
+// truncateRoleName truncates a role name to 64 characters (IAM role name limit).
+func truncateRoleName(name string) string {
+	const maxRoleNameLength = 64
+	if len(name) > maxRoleNameLength {
+		name = name[0:maxRoleNameLength]
```
robpblake: @andclt We should truncate the prefix but leave the `-OCM-Role-$EXTERNAL_ORG_ID` part. I see this is a carry over from the ROSA CLI. It looks like the CLI works around this by enforcing a max prefix of 30 characters, but lets be safe

exp/api/v1beta2/rosaocmroleconfig_types.go
andclt: Yes ROSA CLI supports up to 32 chars for all role prefixes but `ROSARoleConfig` already limits both `AccountRoleConfig` and `OperatorRoleConfig` prefixes to 4 characters.

More info why this was changed to 4 here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6046#discussion_r3397070501

robpblake: Works for me. Consistency across this provider. Thanks.

pkg/rosa/awshelpers/helpers.go
robpblake: Given we're enforcing the `4` character max prefix, then I don't think this will ever be an issue, but the same comment applies that we should only ever truncate the prefix

andclt: edit: written before seeing the comment above

Good catch! However, this code path is actually unreachable in CAPA since we limit role prefixes to 4 chars (max 22 char total role name vs 64 char AWS limit). We could simply remove the truncation entirely, what do you think?

exp/controllers/rosaocmroleconfig_controller.go
robpblake: @serngawy This is required due to the design choice of the OCM API back when the OCM Role concept was first introduced in ~2021

The flow requires that each OCM Role a customer is using is appended to the value of a Label on their OCM Organization. Each OCM Role is separated in the list by `,`. So the label value looks like:

`ocm_role_1, ocm_role_2`

The OCM API does not have append semantics, it only works with PUT. So in the scenario where two clients are updating the label at the same time, the following is possible:

- client 1 read
- client 2 read
- client 1 append value locally
- client 2 append value locally
- client 1 write value
- client 2 write value (client 2 write does not include client 1 value)

This can end up with lost writes. The cli works around this by only allowing you to link one OCM role at a time. Terraform we have applied the same mutex concept when multiple roles are being reconciled so that the provider does them sequentially. We need the same here in the CAPA provider to avoid the potential for lost writes.

We can't rely on this being eventually consistent via a follow on reconcile loop. If an OCM Role value is over-written and removed from the list, ROSA is no longer able to validate the association between OCM org and AWS account and will block any operations on clusters until the OCM Role link is restored. This will result in a customer experience where they are unable to perform operations on certain clusters whilst everything appears correctly configured. It also feels wrong that the resource reports `ready` but the link may actually not be in place.

The mutex here (combined with leadership election) is a simple approach to prevent competing writes, which further prevents a bad customer experience.

pkg/rosa/awshelpers/helpers.go
robpblake: I am fine with that. Personally, it would be better to fail if the role name is too long rather than randomly truncating it as customers don't end up with what they asked for.

andclt: I agree - removed in the latest commit.

exp/controllers/rosaocmroleconfig_controller.go
robpblake: The mutex is at the granularity of per-organization, meaning that reconciles over resources that target different orgs still happen concurrently.

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,368 @@
...
+	// Update status with role ARN and organization ID
+	ocmRoleConfig.Status.RoleARN = roleARN
+	ocmRoleConfig.Status.OrganizationID = orgID
+
+	// Acquire per-organization mutex to prevent concurrent LinkOrgToRole races.
+	// Multiple ROSAOCMRoleConfigs with different AWS accounts can link to the same org,
+	// causing read-modify-write races on the org label. Serialize linking per org.
+	linkMu := r.getLinkMutex(orgID)
+	linkMu.Lock()
+	defer linkMu.Unlock()
```
serngawy, line 230: better to move this to line 205, where link the role happen.

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,368 @@
...
+	roleName, err := aws.GetResourceIdFromARN(roleARN)
+	if err != nil {
+		return fmt.Errorf("failed to parse role ARN: %w", err)
+	}
+
+	// Acquire per-organization mutex to prevent concurrent unlink/link races.
+	// Multiple ROSAOCMRoleConfigs with different AWS accounts can link to the same org,
+	// and we need to prevent a race where one CR is unlinking while another is linking.
+	linkMu := r.getLinkMutex(orgID)
+	linkMu.Lock()
```
serngawy:
```suggestion
	linkMu := r.getLinkMutex(orgID)
	linkMu.Lock()
	defer linkMu.Unlock()
```

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,368 @@
...
+
+		linked, err := rt.OCMClient.LinkOrgToRole(orgID, roleARN)
+		if err != nil {
+			v1beta1conditions.MarkFalse(ocmRoleConfig, expinfrav1.ROSAOCMRoleConfigReadyCondition,
+				expinfrav1.ROSAOCMRoleConfigReconciliationFailedReason, clusterv1beta1.ConditionSeverityError,
+				"Failed to link OCM role: %s", err.Error())
+			return ctrl.Result{}, fmt.Errorf("failed to link OCM role to organization: %w", err)
+		}
+
+		if linked {
```
serngawy: Is it possible to have `linked` as false with no error ?

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,368 @@
...
+	v1beta1conditions "sigs.k8s.io/cluster-api/util/deprecated/v1beta1/conditions"
+	"sigs.k8s.io/cluster-api/util/predicates"
+)
+
+// ROSAOCMRoleConfigReconciler reconciles an ROSAOCMRoleConfig object.
+type ROSAOCMRoleConfigReconciler struct {
+	client.Client
+	Recorder         record.EventRecorder
+	WatchFilterValue string
+	NewStsClient     func(cloud.ScopeUsage, cloud.Session, logger.Wrapper, runtime.Object) stsiface.STSClient
```
serngawy: no need for stsClient

exp/controllers/rosaocmroleconfig_controller.go
serngawy: ignore this

exp/controllers/rosaocmroleconfig_controller.go
andclt: Unlike the `reconcileOCMRole` method which uses `defer`, I'd prefer explicit unlock here because we have sequential AWS and OCM operations that don't require mutex protection. In the worst case (role needs deletion), we make 4 API calls:

  1. `CheckIfAWSAccountExists` (OCM) - needs mutex
  2. `UnlinkOCMRoleFromOrg` (OCM) - needs mutex
  3. `CheckRoleExists` (AWS IAM) - does not need mutex
  4. `DeleteOCMRole` (AWS IAM) - does not need mutex

Only operations 1-2 need the org-level mutex since they interact with the OCM org label. Holding the lock through the subsequent IAM operations (3-4) would unnecessarily block other reconciliations for the same organization during potentially slow AWS API calls.

exp/controllers/rosaocmroleconfig_controller.go
andclt: Good catch! Yes, `LinkOrgToRole` can return `(false, nil)` when the requested role is already linked. This should not happen because we check it first with `CheckIfAWSAccountExists` under the same lock,
but for completeness I'll add a log statement for that case.

exp/controllers/rosaocmroleconfig_controller.go
andclt: removed

robpblake (approved):
The changes look good to me. Thanks @andclt

exp/api/v1beta2/rosaocmroleconfig_types.go
serngawy: I'm re-thinking this, as we discussed could be a customer with 30 accounts "30 ocmRoles" having prefix with only 4 chars could limit there naming way. Might be better to increase it to 12 chars prefix. What do you think ?

exp/controllers/rosaocmroleconfig_controller.go
serngawy: okay,  better then to move `CheckIfAWSAccountExists` AND `UnlinkOCMRoleFromOrg` to inline function or separate function. something like below  

```suggestion
if err := func() error {
      linkMu.Lock()
      defer linkMu.Unlock()
      existsOnOCM, _, linkedARN, err := rt.OCMClient.CheckIfAWSAccountExists(orgID, rt.Creator.AccountID)
      .... 
      if existsOnOCM && linkedARN == roleARN {
          rt.OCMClient.UnlinkOCMRoleFromOrg(orgID, roleARN)
          scope.Info("Successfully unlinked role", "roleARN", roleARN, "orgID", orgID)
      }
     return nil
  }(); err != nil {
      return err
  }
```

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,343 @@
...
+	rt, err := r.setUpRuntime(ctx, scope)
+	if err != nil {
+		return ctrl.Result{}, fmt.Errorf("failed to set up runtime: %w", err)
+	}
+
+	if !scope.ROSAOCMRoleConfig.DeletionTimestamp.IsZero() {
+		scope.Info("Deleting ROSAOCMRoleConfig.")
+		v1beta1conditions.MarkFalse(scope.ROSAOCMRoleConfig, expinfrav1.ROSAOCMRoleConfigReadyCondition, expinfrav1.ROSAOCMRoleConfigDeletionStarted, clusterv1beta1.ConditionSeverityInfo, "Deletion of ROSAOCMRoleConfig started")
+		err = r.reconcileDelete(ctx, scope, rt)
+		if err == nil {
```
serngawy: better to check if `err != nil` then set `ROSAOCMRoleConfigDeletionFailedReason` and return with error. Move the `controllerutil.RemoveFinalizer` to be after the err check

serngawy (commented):
looks good, few more comments

exp/controllers/rosaocmroleconfig_controller.go
andclt: Good point, changed!

exp/controllers/rosaocmroleconfig_controller.go
andclt: Yeah inline functions is better IMO

exp/api/v1beta2/rosaocmroleconfig_types.go
andclt: I think 12 chars would be a better option considering this use case.

pkg/rosa/awshelpers/helpers.go
```diff
@@ -0,0 +1,102 @@
...
+	return fmt.Sprintf("%s-Admin-Policy", name)
+}
+
+// GetNoConsolePolicyName returns the no-console policy name for a role.
+func GetNoConsolePolicyName(name string) string {
+	return fmt.Sprintf("%s-NoConsole-Policy", name)
+}
+
+// GetPolicyArn returns the ARN for a policy with a given name.
+func GetPolicyArn(partition, accountID, name, path string) string {
```
serngawy: Not sure, why having the function twice ? can we just change `getPolicyARN` to `GetPolicyARN` and use it every where instead of having this function

pkg/rosa/awshelpers/helpers.go
```diff
@@ -0,0 +1,102 @@
...
+func GetNoConsolePolicyARN(partition, accountID, name, path string) string {
+	return getPolicyARN(partition, accountID, GetNoConsolePolicyName(name), path)
+}
+
+// getPolicyARN constructs an IAM policy ARN.
+func getPolicyARN(partition, accountID, name, path string) string {
+	str := fmt.Sprintf("arn:%s:iam::%s:policy", partition, accountID)
+	if path != "" {
+		str = fmt.Sprintf("%s%s", str, path)
+		return fmt.Sprintf("%s%s", str, name)
```
serngawy: one comment by claude below, could be fixed by setting regexp at spec.path to end with "/"  instead of using TrimSuffix

```
When a user sets spec.path: /custom/path (valid per CRD regex ^\/.*$), the name is glued directly to the last path segment, producing
  arn:aws:iam::123:policy/custom/pathMyPolicy instead of arn:aws:iam::123:policy/custom/path/MyPolicy. Every IAM call using this ARN (EnsurePolicy,
  AttachRolePolicy) returns NoSuchEntity; the role is created in IAM but policy attachment permanently fails. The CRD pattern does not require a trailing
  slash, so this is user-reachable. Fix: strings.TrimSuffix(path, "/") + "/" before concatenating.
```

exp/api/v1beta2/rosaocmroleconfig_types.go
```diff
@@ -0,0 +1,162 @@
...
+	// OrganizationID is the OCM organization ID that this role is linked to.
+	OrganizationID string `json:"organizationID,omitempty"`
+
+	// Conditions specifies the ROSAOCMRoleConfig conditions
+	Conditions clusterv1beta1.Conditions `json:"conditions,omitempty"`
+}
+
+// ROSAOCMRoleConfig is the Schema for the rosaocmroleconfigs API
+// +kubebuilder:object:root=true
+// +kubebuilder:resource:path=rosaocmroleconfigs,scope=Cluster,categories=cluster-api,shortName=rosaocmrole
```
serngawy, line 131: As the implementation changes from our initial thoughts, may be better to change the scope to namespace instead of cluster.  We may have issue with aws cross account creation and cluster scope resources.

serngawy (commented):
Looks good in general, just few small comments but most important is we need to test the rosaocmrole creation with aws cross account creation. As we discussed, users will have 4-5 aws accounts to work with. Most likely using AWSClusterRoleIdentity with IAM role arn setup to authorize capa with different aws accounts.  
Below is doc for how to setup CAPA with IAM roles and cross account creation, let me know if you need help. 
https://github.com/stolostron/cluster-api-installer/blob/main/doc/Enable_iam_roles_capa.md 

https://github.com/stolostron/cluster-api-installer/blob/main/doc/multi-account-rosa-hcp.md

pkg/rosa/awshelpers/helpers.go
andclt: These helper methods were initially added to maintain consistency with the ROSA CLI. However, most of them already exist in the vendored ROSA version, so I've removed the duplicates and now import directly from github.com/openshift/rosa/pkg/aws.

pkg/rosa/awshelpers/helpers.go
andclt: good point, I prefer setting regexp at spec.path to end with "/" since now we are calling that method from the vendored ROSA

exp/api/v1beta2/rosaocmroleconfig_types.go
andclt: Before I change the scope, can you clarify what you mean by "we may have issue with aws cross account creation and cluster scope resources"?
 
I want to make sure I understand the specific issue:
  * Is it about allowedNamespaces validation on identities?
  * Is it about cluster-scoped resources not being able to reference namespace-scoped secrets?
  * Something else?

Also, one thing I'm trying to reconcile: the OCM role is AWS account-level (shared across all clusters in that account), not cluster-specific. If it's namespace-scoped and `namespace-a` deletes it, clusters in `namespace-b` lose OCM API functionality even though they didn't delete anything.
Does that concern you, or is that expected behavior?

Just want to make sure I understand the design intent before making the change!

exp/api/v1beta2/types.go
```diff
@@ -27,6 +27,17 @@ const (
 	KindMachinePool string = "MachinePool"
 )
 
+// SecretReference references a secret by name and namespace.
+type SecretReference struct {
```
serngawy: corev1 "k8s.io/api/core/v1" 

already exist under corev1.SecretReference  better to use it from there

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,348 @@
...
+
+	roleARN, orgID, created, err := ocmrole.GetOrCreateOCMRole(
+		rt,
+		&scope.Logger,
+		ocmRoleConfig.Spec.RolePrefix,
+		extractedProfile,
+		ocmRoleConfig.Spec.PermissionsBoundaryARN,
+		ocmRoleConfig.Spec.Path,
+	)
+	if err != nil {
```
serngawy, line 215: Is it possible to be created and get error at setting policy permission ?  if yes, we should set the  `ocmRoleConfig.Status.RoleARN` and `ocmRoleConfig.Status.OrganizationID`  in order when deletion happen the role get deleted.

andclt:
> Looks good in general, just few small comments but most important is we need to test the rosaocmrole creation with aws cross account creation. As we discussed, users will have 4-5 aws accounts to work with. Most likely using AWSClusterRoleIdentity with IAM role arn setup to authorize capa with different aws accounts. Below is doc for how to setup CAPA with IAM roles and cross account creation, let me know if you need help. https://github.com/stolostron/cluster-api-installer/blob/main/doc/Enable_iam_roles_capa.md
> 
> https://github.com/stolostron/cluster-api-installer/blob/main/doc/multi-account-rosa-hcp.md

tested and the `identityPermission` field was introduced to allow cross role account creation

exp/api/v1beta2/types.go
andclt: Thanks, changed it in the latest commit.

exp/controllers/rosaocmroleconfig_controller.go
andclt: This is actually a good point! It made me think that we should probably self-heal the role (in case of missing policies) instead of just returning an error as the CLI does.

This would be similar to the pattern we follow for `ROSANetwork` and `ROSARoleConfig`, and I'd say it provides a better experience since the controller automatically reconciles to the desired state rather than leaving resources in an incomplete state and simply return an error.

I implemented the self healing logic in the latest commit: if the role exists but has missing policies, the controller will automatically attach them during reconciliation.

exp/controllers/rosaocmroleconfig_controller.go
```diff
@@ -0,0 +1,410 @@
...
+		if err != nil {
+			v1beta1conditions.MarkFalse(ocmRoleConfig, expinfrav1.ROSAOCMRoleConfigReadyCondition,
+				expinfrav1.ROSAOCMRoleConfigReconciliationFailedReason, clusterv1beta1.ConditionSeverityError,
+				"Failed to detect role profile: %s", err.Error())
+			return ctrl.Result{}, fmt.Errorf("failed to detect role profile: %w", err)
+		}
+
+		requestedProfile := string(ocmRoleConfig.Spec.Profile)
+		if existingProfile != requestedProfile {
+			v1beta1conditions.MarkFalse(ocmRoleConfig, expinfrav1.ROSAOCMRoleConfigReadyCondition,
```
serngawy: in order to self heal, I think we should try to apply the `requestedProfile` before setting the condition. Otherwise the OCMRole will stuck with the `existingProfile` . Right ?

serngawy (commented):
looks good, I have comment to clarify self healing

exp/controllers/rosaocmroleconfig_controller.go
andclt: this is actually intended. It's mainly a safety check for when the AWS role already exists but has the wrong profile. This happens when someone pre-created the role manually with `rosa create ocm-role` or if someone modified the role directly in AWS.

Since `profile` is immutable in the CRD spec, the user can't just update the CR to match the existing role. Better to error out so they delete and recreate the resource, which triggers our deletion logic that properly removes the role and all attached policies. Otherwise we'd have to handle detaching old policies and attaching new ones ourselves, which is error-prone and invasive. `rosa CLI` does the same thing, it
errors out instead of trying to mutate an existing role's profile.

Do you think we should inform the user to delete and re-create the CR in the error message when this happens?

serngawy: No, we just log the profile miss match

serngawy (approved):
looks good to me

nrb (approved):
/lgtm
/approve

PR #6064, opened 2026-06-24
🐛 fix(e2e): use targeted patch for EKS upgrade policy test

damdo:
**What type of PR is this?**

/kind bug
/kind flake

The `[managed] [upgrade-policy]` EKS e2e test flakes because it re-applies the entire cluster template (via `ManagedClusterSpec()`) to change a single field (`upgradePolicy`). The CAPI framework's `CreateOrUpdate` performs a full object `Update()`, which wipes `ownerReferences`, finalizers, and status from the existing `AWSManagedControlPlane`. This creates a race where the controller may never reconcile within the 5-minute timeout.

This PR replaces the second `ManagedClusterSpec()` call with a targeted patch using the CAPI `patch.NewHelper`, matching the pattern already used by `UpgradeControlPlaneVersionSpec()`. This preserves all controller-managed metadata and triggers a reliable reconciliation. The `Get` and `Patch` calls are wrapped in `Eventually` blocks for resilience against transient API errors.

Fixes #

Confirmed from Prow CI logs that in failing runs, after the second `ManagedClusterSpec()` call the webhook accepts the update but the `awsmanagedcontrolplane` controller never reconciles the resource again (ownerRef wiped). The test passed in ~4/19 runs on PR #5857 (timing-dependent).

**AI Usage**:

Claude Code (Opus) was used to investigate the root cause of the flake via Prow CI artifact analysis, and to draft the initial fix. The final code was reviewed and adjusted by the author.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/test/e2e/suites/managed/eks_upgrade_policy_test.go
+++ b/test/e2e/suites/managed/eks_upgrade_policy_test.go
@@ -29,6 +29,7 @@ import (
 	"github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	corev1 "k8s.io/api/core/v1"
+	crclient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	ekscontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/eks/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/awserrors"
@@ -36,6 +37,7 @@ import (
 	"sigs.k8s.io/cluster-api-provider-aws/v2/test/e2e/shared"
 	"sigs.k8s.io/cluster-api/test/framework"
 	"sigs.k8s.io/cluster-api/util"
+	"sigs.k8s.io/cluster-api/util/patch"
 )
 
 // EKS upgrade policy test.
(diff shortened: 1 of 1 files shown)


damdo:
/hold

for test to finish

damdo:
## CI Failure Analysis

The EKS e2e failure is **not caused by this PR's changes**. The panic occurs at line 83 (`ManagedClusterSpec()`) during the initial cluster creation — before our patch code is reached.

**Root cause**: EKS now rejects creating a v1.32 cluster with `supportType: standard`. From the controller logs:

```
InvalidParameterException: The requested Kubernetes version "1.32" is only supported under extended support.
Please change the supportType to extended support to create the cluster.
```

Kubernetes v1.32 has moved out of standard support in EKS. This test creates the cluster with `UpgradePolicyStandard` + `KUBERNETES_VERSION: "v1.32.0"` (from `e2e_eks_conf.yaml`), which EKS now rejects. The cluster never provisions, hitting the 40-minute timeout.

All other EKS e2e tests in this run passed, this is the only one that uses `UpgradePolicyStandard` for cluster creation.

This is a pre-existing issue on `main` unrelated to the patch fix in this PR. Will be fixed by https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/5857

damdo (commented):
/approve

damdo:
Ok so the EKS e2e with this fix stacked on top of #5857 passed: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5857/pull-cluster-api-provider-aws-e2e-eks/2069759758755172352

I'd say we merge this, as this fixes the first issue, the second issue (the one that makes the e2e fail here) will be fixed by #5857

PR #6073, opened 2026-06-25
🐛 fix: clean up orphaned ENIs before subnet deletion at teardown

### LINKED ISSUE #6067: [flake] Orphaned ENIs block security group and subnet deletion during cluster teardown
## Summary

During cluster deletion, orphaned ENIs (created by NLBs for cross-AZ connectivity) block security group and subnet deletion. After CAPA deletes the NLB, AWS asynchronously cleans up the NLB's ENIs — a process that can take up to 1 hour. CAPA proceeds immediately to SG/subnet deletion with no ENI cleanup step, hitting `DependencyViolation` errors until the test times out.

## Observed Behavior

In the ignition/S3 e2e test, after the API server NLB and all instances were successfully deleted, 2 security groups and 1 subnet remained stuck with `DependencyViolation` errors for 29+ minutes until the controller context was canceled:

- `sg-0118fb1f614c1f100` (`*-lb` security group) — "has a dependent object"
- `sg-0acda437b46107e73` (`*-node` security group) — "has a dependent object"
- `subnet-0a39593d76536bcd2` (private subnet) — "has dependencies"

Meanwhile, `sg-04e1e99280fab2e28` (`*-controlplane`) and `sg-042725e438cafe649` (`*-apiserver-lb`) were successfully deleted ~4 minutes after instance termination, once EC2 finished cleaning up instance ENIs.

## Detailed Timeline

| Time | Event |
|------|-------|
| 14:40:22-24 | 4 SGs created: `apiserver-lb`, `lb`, `controlplane`, `node` |
| 14:40:28 | API server NLB created (uses `apiserver-lb` SG, creates cross-AZ ENIs in subnets) |
| 14:43:52 | EC2 instances launched with SGs: `node` + `lb` (all machines) + `controlplane` (CP only). Instance ENIs have `deleteOnTermination: true` |
| 14:47:31-33 | EC2 instances terminated |
| 14:49:54 | CAPA deletes the NLB. Waits for LB resource to disappear via `describeLB`, but **not** for NLB ENIs to be cleaned up |
| 14:49:56 | All 4 SGs fail deletion with `DependencyViolation` (ENIs still exist) |
| 14:51:34-35 | `controlplane` + `apiserver-lb` SGs deleted (~4 min after instance termination — EC2 finished cleaning up instance ENIs) |
| 14:51 → 15:18 | `lb` + `node` SGs stuck with `DependencyViolation`, retrying every ~2-7 seconds |
| 15:07:31 | **E2e test times out** (1200s) |
| 15:18:50 | Controller context canceled |
| CloudTrail: 14:50-15:47 | AWS ELB service role (`elasticloadbalancing.amazonaws.com`) deletes NLB ENIs asynchronously over ~1 hour |

### Why `lb` and `node` SGs are stuck but `controlplane` and `apiserver-lb` are not

The `lb` SG is attached to **every EC2 instance** (both control-plane and worker nodes) — see [`pkg/cloud/services/ec2/instances.go:488`](pkg/cloud/services/ec2/instances.go#L488). It's a "container for the cloud provider to inject its load balancer ingress rules" (comment at [`securitygroups.go:719`](pkg/cloud/services/securitygroup/securitygroups.go#L719)). The `node` SG is similarly attached to all instances.

When instances are terminated, their primary ENIs (`deleteOnTermination: true`) are cleaned up by EC2 within ~4 minutes. That's why `controlplane` and `apiserver-lb` SGs freed up at 14:51:34.

But the `lb` and `node` SGs remain blocked because the NLB's cross-AZ ENIs (which reside in the same subnets) are cleaned up asynchronously by AWS's ELB service — and this takes much longer (observed: up to 1 hour in CloudTrail). The subnet `subnet-0a39593d76536bcd2` can't be deleted because it still has these NLB ENIs in it.

### What CAPA is missing

1. **No ENI cleanup step in `reconcileDelete`** ([`controllers/awscluster_controller.go:204-282`](controllers/awscluster_controller.go#L204)): The deletion order is S3 → ELB → bastion → SGs → GC → network, with no ENI enumeration or cleanup at any point.

2. **NLB deletion doesn't wait for ENI cleanup** ([`pkg/cloud/services/elb/loadbalancer.go:881-887`](pkg/cloud/services/elb/loadbalancer.go#L881)): After NLB deletion, CAPA polls `describeLB` until the LB resource is gone, but the LB disappears immediately — the ENIs linger for up to 1 hour.

3. **No `DeleteNetworkInterface` call anywhere in production code**: The only reference is in the e2e IAM policy ([`test/e2e/shared/template.go:84`](test/e2e/shared/template.go#L84)). The EC2 client interface has `DescribeNetworkInterfaces` and `ModifyNetworkInterfaceAttribute` but no `DeleteNetworkInterface`.

4. **GC service skips ENIs**: The GC service discovers ENIs via the Resource Groups Tagging API but has no handler for them — they are logged as "Resource not a load balancer for deletion" or "Resource not a security group for deletion".

## Suggested Fix

**Add ENI cleanup before SG/subnet deletion in `reconcileDelete`**: After LB deletion, enumerate ENIs in the VPC via `DescribeNetworkInterfaces` filtered by VPC ID + cluster ownership tag. For any ENI in `available` state (detached from instances/LBs), call `DeleteNetworkInterface`. This actively cleans up orphaned NLB ENIs instead of waiting for AWS's async cleanup.

This is the most robust approach because:
- It handles both NLB ENIs and any other orphaned ENIs (e.g., from cloud provider LBs)
- The reconciler already retries on error, so transient failures (ENI still in-use) are handled naturally
- It doesn't require changes to the GC service or SG deletion logic
- The `ec2:DeleteNetworkInterface` permission is already in the IAM policy

Alternative/complementary approaches:
- **Add an ENI handler to the GC service**: Extend `pkg/cloud/services/gc/cleanup.go` to recognize `network-interface` resources
- **Handle `DependencyViolation` with ENI introspection**: When SG/subnet deletion fails, query the blocking ENIs and clean them up before retrying
- **Increase e2e deletion timeout**: Band-aid — the reconciler already retries, so a longer timeout (60+ min) would eventually succeed, but this is not a real fix

## Environment

- Observed in: e2e CI (`pull-cluster-api-provider-aws-e2e`)
- Prow job: https://prow.k8s.io/view/gs/kubernetes-ci-logs/pr-logs/pull/kubernetes-sigs_cluster-api-provider-aws/5857/pull-cluster-api-provider-aws-e2e/2069788635669991424
- Pre-existing issue on `main`, not introduced by any specific PR

/kind bug

damdo:
**What type of PR is this?**

/kind bug

After NLB deletion during cluster teardown, AWS asynchronously cleans up the NLB's cross-AZ ENIs, a process that can take up to 1 hour. CAPA proceeds immediately to security group and subnet deletion with no ENI cleanup step, causing `DependencyViolation` errors that block cluster deletion until the e2e test times out.

This PR adds a `deleteOrphanedENIs()` step in `DeleteNetwork()`, right before subnet deletion. It discovers ENIs in the VPC that are:
- in `available` state (not attached to any instance or service)
- referencing one of the cluster's security groups (scoped to only this cluster's ENIs)

This targets exactly the ENIs blocking SG/subnet deletion without touching other clusters' ENIs sharing the same VPC.

Fixes #6067

The ENI cleanup is non-fatal: if it fails, the error is logged and subnet deletion proceeds (which will fail with `DependencyViolation` and trigger a requeue, as before). On the next reconcile, the ENIs that have since become `available` are cleaned up, unblocking both SG and subnet deletion.

No new service factory, interface, or controller wiring is needed, the implementation lives entirely in the existing network service's `DeleteNetwork()` flow.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [x] includes emoji in title 
- [x] adds unit tests

**Release note**:
```release-note
fix: clean up orphaned network interfaces (ENIs) before subnet deletion during cluster teardown, preventing DependencyViolation errors caused by NLB ENIs that AWS cleans up asynchronously after load balancer deletion.
```

### DIFF
--- a/pkg/cloud/services/network/enis.go
+++ b/pkg/cloud/services/network/enis.go
@@ -30,7 +30,7 @@ import (
 // in "available" state (not attached to any instance or service) and reference
 // one of the cluster's security groups. These are typically left behind by NLBs
 // whose cross-AZ ENIs are cleaned up asynchronously by AWS after LB deletion.
-func (s *Service) deleteOrphanedENIs() error {
+func (s *Service) deleteOrphanedENIs(ctx context.Context) error {
 	if s.scope.VPC().IsUnmanaged(s.scope.Name()) {
 		s.scope.Trace("Skipping orphaned ENI cleanup in unmanaged mode")
 		return nil
@@ -59,7 +59,7 @@ func (s *Service) deleteOrphanedENIs() error {
 		},
 	}
 
-	out, err := s.EC2Client.DescribeNetworkInterfaces(context.TODO(), input)
+	out, err := s.EC2Client.DescribeNetworkInterfaces(ctx, input)
 	if err != nil {
 		return fmt.Errorf("describing orphaned network interfaces: %w", err)
 	}
--- a/pkg/cloud/services/network/enis.go
+++ b/pkg/cloud/services/network/enis.go
@@ -75,7 +75,7 @@ func (s *Service) deleteOrphanedENIs(ctx context.Context) error {
 		eniID := aws.ToString(eni.NetworkInterfaceId)
 		s.scope.Debug("Deleting orphaned network interface", "eni-id", eniID)
 
-		if _, err := s.EC2Client.DeleteNetworkInterface(context.TODO(), &ec2.DeleteNetworkInterfaceInput{
+		if _, err := s.EC2Client.DeleteNetworkInterface(ctx, &ec2.DeleteNetworkInterfaceInput{
 			NetworkInterfaceId: eni.NetworkInterfaceId,
 		}); err != nil {
 			errs = append(errs, fmt.Errorf("deleting network interface %q: %w", eniID, err))
--- a/pkg/cloud/services/network/enis.go
+++ b/pkg/cloud/services/network/enis.go
@@ -1,0 +1,100 @@
+/*
+Copyright 2025 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package network
+
+import (
+	"context"
+	"fmt"
+
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/ec2"
+	"github.com/aws/aws-sdk-go-v2/service/ec2/types"
+	kerrors "k8s.io/apimachinery/pkg/util/errors"
+)
+
+// deleteOrphanedENIs deletes network interfaces in the cluster's VPC that are
+// in "available" state (not attached to any instance or service) and reference
+// one of the cluster's security groups. These are typically left behind by NLBs
+// whose cross-AZ ENIs are cleaned up asynchronously by AWS after LB deletion.
+func (s *Service) deleteOrphanedENIs() error {
+	if s.scope.VPC().IsUnmanaged(s.scope.Name()) {
+		s.scope.Trace("Skipping orphaned ENI cleanup in unmanaged mode")
+		return nil
+	}
+
+	sgIDs := s.clusterSecurityGroupIDs()
+	if len(sgIDs) == 0 {
+		s.scope.Debug("No cluster security groups found, skipping orphaned ENI cleanup")
+		return nil
+	}
+
+	input := &ec2.DescribeNetworkInterfacesInput{
+		Filters: []types.Filter{
+			{
+				Name:   aws.String("vpc-id"),
+				Values: []string{s.scope.VPC().ID},
+			},
+			{
+				Name:   aws.String("group-id"),
+				Values: sgIDs,
+			},
+			{
+				Name:   aws.String("status"),
+				Values: []string{"available"},
+			},
+		},
+	}
+
+	out, err := s.EC2Client.DescribeNetworkInterfaces(context.TODO(), input)
+	if err != nil {
+		return fmt.Errorf("describing orphaned network interfaces: %w", err)
+	}
+
+	if len(out.NetworkInterfaces) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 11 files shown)


damdo:
full e2es passed 5 times in a row: https://prow.k8s.io/pr-history/?org=kubernetes-sigs&repo=cluster-api-provider-aws&pr=6073

clebs:
Failing `pull-cluster-api-provider-aws-apidiff-main` is expected since new interface methods were added to delete ENIs.

/lgtm

pkg/cloud/services/network/enis.go
```diff
@@ -0,0 +1,100 @@
...
+				Values: sgIDs,
+			},
+			{
+				Name:   aws.String("status"),
+				Values: []string{"available"},
+			},
+		},
+	}
+
+	out, err := s.EC2Client.DescribeNetworkInterfaces(context.TODO(), input)
```
nrb: Should we pass one of our own contexts in to this function?

nrb (approved):
Thank you for tracking this down! The dependency violation flakes have been a headache for a while.

We may want to go through the AWS resources we're creating and double check that they have explicit deletions; it seems that recently we're uncovering a lot of async/timing issues within the AWS platform. An agent should be able to do this relatively easily.

I have a minor comment, feel free to unhold if you don't want to address it.

/hold
/approve

pkg/cloud/services/network/enis.go
damdo: Yeah it needs a bit of refactor so I was holding off on that. But I had a look at trying that and it is not too bad. Will add.

pkg/cloud/services/network/enis.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	s.scope.Info("Deleting orphaned network interfaces", "count", len(out.NetworkInterfaces))
+
+	var errs []error
+	for _, eni := range out.NetworkInterfaces {
+		eniID := aws.ToString(eni.NetworkInterfaceId)
+		s.scope.Debug("Deleting orphaned network interface", "eni-id", eniID)
+
+		if _, err := s.EC2Client.DeleteNetworkInterface(context.TODO(), &ec2.DeleteNetworkInterfaceInput{
```
nrb:
```suggestion
		if _, err := s.EC2Client.DeleteNetworkInterface(ctx, &ec2.DeleteNetworkInterfaceInput{
```

nrb (commented):
One missed `context.TODO()`, otherwise 👍

damdo:
build-docker is ooming, I've created: https://github.com/kubernetes/test-infra/pull/37376

clebs:
/lgtm

Re-tagging after fixing docker OOM issue.

PR #6076, opened 2026-06-26
iamauth: fix aws-iam-authenticator reconcile for clusters using instance profiles from AWSMachinePool / AWSMachineTemplate

### LINKED ISSUE #6075: iamauth: GetRole called with instance profile names from AWSMachinePool / AWSMachineTemplate, breaking aws-iam-authenticator reconcile on EKS clusters
/kind bug

When the `aws-iam-authenticator` reconciler in `pkg/cloud/services/iamauth` discovers worker roles from an `AWSManagedControlPlane`'s referenced `MachineDeployment`s and `MachinePool`s, it conflates two distinct concepts:

- `AWSMachineTemplate.Spec.Template.Spec.IAMInstanceProfile` — an **instance profile name**
- `AWSMachinePool.Spec.AWSLaunchTemplate.IamInstanceProfile` — an **instance profile name**
- `AWSManagedMachinePool.Spec.RoleName` — an **IAM role name**

All three are collected into the same `map[string]struct{}` in `getRolesForWorkers` ([reconcile.go#L107-L116](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/iamauth/reconcile.go#L107-L116)) and then passed to `getARNForRole`, which unconditionally calls `iam:GetRole` ([reconcile.go#L93-L105](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/iamauth/reconcile.go#L93-L105)). For the first two cases this is wrong — `GetRole` is being called with an instance profile name, not a role name.

Result on any EKS cluster (`AWSManagedControlPlane`) whose worker nodes come from `AWSMachinePool` or `MachineDeployment` + `AWSMachineTemplate`:

- `GetRole` returns `NoSuchEntity` (when an IAM role with the same name as the instance profile happens not to exist) — reconcile fails on every loop, the `aws-auth` ConfigMap never gets the node role mapping, and the AMCP condition `IAMAuthenticatorConfigured` stays `False` indefinitely.
- Or, `GetRole` returns a role ARN for an *unrelated* IAM role that happens to share the name (instance profile and a separate role coincidentally have the same name) — `aws-auth` is silently populated with the wrong ARN, and node bootstrap fails downstream.

We hit the `NoSuchEntity` flavor across our fleet. Reproducing:

1. Create an `AWSManagedControlPlane`-backed cluster.
2. Add a worker `AWSMachinePool` whose `AWSLaunchTemplate.IamInstanceProfile` is an instance profile name **that does not match any IAM role of the same name** (normal for instance profiles created with a `-instance-profile` suffix or any non-trivial naming convention).
3. Watch `capa-controller-manager` logs:
   ```
   getting roles for remote workers: failed to get ARN for role <instance-profile-name>: ... NoSuchEntity: The role with name <instance-profile-name> cannot be found.
   ```
4. `kubectl get awsmanagedcontrolplane <name> -o jsonpath='{.status.conditions[?(@.type=="IAMAuthenticatorConfigured")]}'` shows `status: False` permanently.

The reconciler should resolve **instance profile names via `iam:GetInstanceProfile`** (reading the ARN of the attached role from the response) and resolve **role names via `iam:GetRole`** as it does today. The dispatch should be based on which CRD field the name came from, since the CRD schemas already encode the distinction unambiguously.

- The bug is in `pkg/cloud/services/iamauth/reconcile.go`, in `getRolesForWorkers` and the helpers it calls (`getRolesForMachineDeployments`, `getRolesForAWSMachinePool`, `getRolesForAWSManagedMachinePool`). They write all three name sources into a single `map[string]struct{}`, losing the kind distinction by the time `getARNForRole` is called.
- Proposed shape of fix:
  - Track each discovered name's kind (`asInstanceProfile` vs `asRole`) at collection time via a `map[string]nameKind`.
  - Add a `getARNForInstanceProfile(ctx, name)` helper that calls `iam:GetInstanceProfile` and returns `Roles[0].Arn` (AWS limits an instance profile to one attached role; an instance profile with no attached role is treated as an error).
  - Add a `resolveRoleARN(ctx, name, kind)` dispatcher and call it from `ReconcileIAMAuthenticator`.
- Permissions: the IAM policy attached to the CAPA controller will need `iam:GetInstanceProfile` in addition to the existing `iam:GetRole`. Any shipped IAM policy templates (e.g. `AWSIAMManagedPolicyControllers.json`) need a corresponding update.

**Environment:**

Affected code path is version-independent — present on `main` HEAD and in every released version that has the current `iamauth/reconcile.go` shape (predates v2.5). Not OS- or Kubernetes-version-specific.

Kesav531:
**What this PR does / why we need it:**

Fixes a bug in `pkg/cloud/services/iamauth` where the `aws-iam-authenticator` reconciler conflates IAM **instance profile names** (sourced from `AWSMachineTemplate.Spec.Template.Spec.IAMInstanceProfile` and `AWSMachinePool.Spec.AWSLaunchTemplate.IamInstanceProfile`) with IAM **role names** (sourced from `AWSManagedMachinePool.Spec.RoleName`). All three are collected into a single `map[string]struct{}` and then unconditionally resolved via `iam:GetRole`, which is wrong for the first two cases.

Result on EKS clusters whose workers come from `AWSMachinePool` or `MachineDeployment` + `AWSMachineTemplate`:

- `GetRole` returns `NoSuchEntity` (when no IAM role of the same name as the instance profile exists) — reconcile fails permanently, the `aws-auth` ConfigMap never gets the node role mapping, AMCP condition `IAMAuthenticatorConfigured` stays `False`.
- Or, `GetRole` returns the ARN of an *unrelated* IAM role that happens to share the name — `aws-auth` is silently populated with the wrong ARN.

**Fix:**

- Track each discovered name's kind (`asInstanceProfile` vs `asRole`) at collection time via a `map[string]nameKind`.
- Add `getARNForInstanceProfile` calling `iam:GetInstanceProfile` and returning `Roles[0].Arn` (AWS limits an instance profile to one attached role; empty `Roles` is treated as an error).
- Add `resolveRoleARN(ctx, name, kind)` dispatcher; replace the single `getARNForRole` call site in `ReconcileIAMAuthenticator`.
- Grant `iam:GetInstanceProfile` in the EKS controllers policy generated by `clusterawsadm`, so the controller can actually call the new API. Regenerated all 13 affected bootstrap fixtures.

**Which issue(s) this PR fixes:**

Fixes #6075

**Special notes for your reviewer:**

- Unit tests added in `pkg/cloud/services/iamauth/reconcile_test.go` cover: instance-profile-name dispatch, role-name dispatch, error path for instance profile with no attached role, error path for unknown kind.
- Mock regenerated for the new `GetInstanceProfile` call.
- The bootstrap fixture test (`cmd/clusterawsadm/cloudformation/bootstrap`) passes — all 13 fixtures match the regenerated policy.
- Validated across a fleet of EKS clusters using `AWSMachinePool`; AMCP `IAMAuthenticatorConfigured` flipped from `False` to `True` after the fix landed alongside the IAM grant.

**Release note:**

```release-note
Fix `aws-iam-authenticator` reconcile for EKS clusters whose worker nodes come from `AWSMachinePool` or `MachineDeployment` + `AWSMachineTemplate`. Names from `IAMInstanceProfile` CRD fields are now resolved via `iam:GetInstanceProfile`; only `AWSManagedMachinePool.Spec.RoleName` is resolved via `iam:GetRole`. The EKS controllers IAM policy generated by `clusterawsadm` now grants `iam:GetInstanceProfile`.
```

### DIFF
--- a/pkg/cloud/services/iamauth/reconcile.go
+++ b/pkg/cloud/services/iamauth/reconcile.go
@@ -32,6 +32,18 @@ import (
 	clusterv1 "sigs.k8s.io/cluster-api/api/core/v1beta2"
 )
 
+// nameKind tracks whether a name discovered from a CRD field is an IAM
+// instance profile name or an IAM role name. AWSMachinePool and
+// AWSMachineTemplate expose an instance profile name; AWSManagedMachinePool
+// exposes a role name. The two require different IAM API calls to resolve to
+// a role ARN, so we tag each entry at discovery time and dispatch on the tag.
+type nameKind int
+
+const (
+	asInstanceProfile nameKind = iota
+	asRole
+)
+
 // ReconcileIAMAuthenticator is used to create the aws-iam-authenticator in a cluster.
 func (s *Service) ReconcileIAMAuthenticator(ctx context.Context) error {
 	s.scope.Info("Reconciling aws-iam-authenticator configuration", "cluster", klog.KRef(s.scope.Namespace(), s.scope.Name()))
@@ -51,10 +63,10 @@ func (s *Service) ReconcileIAMAuthenticator(ctx context.Context) error {
 		s.scope.Error(err, "getting roles for remote workers")
 		return fmt.Errorf("getting roles for remote workers: %w", err)
 	}
-	for roleName := range nodeRoles {
-		roleARN, err := s.getARNForRole(ctx, roleName)
+	for name, kind := range nodeRoles {
+		roleARN, err := s.resolveRoleARN(ctx, name, kind)
 		if err != nil {
-			return fmt.Errorf("failed to get ARN for role %s: %w", roleName, err)
+			return fmt.Errorf("failed to resolve role ARN for %s: %w", name, err)
 		}
 		nodesRoleMapping := ekscontrolplanev1.RoleMapping{
 			RoleARN: roleARN,
@@ -90,6 +102,23 @@ func (s *Service) ReconcileIAMAuthenticator(ctx context.Context) error {
 	return nil
 }
 
+// resolveRoleARN returns the ARN of the IAM role represented by `name`,
+// dispatching on `kind` to call the matching IAM API. Names sourced from
+// AWSMachinePool / AWSMachineTemplate are instance profile names and are
+// resolved via GetInstanceProfile (the resulting role ARN is taken from the
+// instance profile's attached role). Names sourced from AWSManagedMachinePool
+// are role names and are resolved via GetRole.
+func (s *Service) resolveRoleARN(ctx context.Context, name string, kind nameKind) (string, error) {
+	switch kind {
+	case asRole:
+		return s.getARNForRole(ctx, name)
+	case asInstanceProfile:
+		return s.getARNForInstanceProfile(ctx, name)
+	default:
+		return "", fmt.Errorf("unknown name kind %d for %s", kind, name)
+	}
+}
+
 func (s *Service) getARNForRole(ctx context.Context, role string) (string, error) {
 	input := &iam.GetRoleInput{
 		RoleName: aws.String(role),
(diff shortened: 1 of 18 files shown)


asifdxtreme:
Thanks @Kesav531 for submitting this change, this was one of the blocking issue for us.

asifdxtreme (approved):

pkg/cloud/services/iamauth/reconcile.go
```diff
@@ -90,6 +102,23 @@ func (s *Service) ReconcileIAMAuthenticator(ctx context.Context) error {
 	return nil
 }
 
+// resolveRoleARN returns the ARN of the IAM role represented by `name`,
+// dispatching on `kind` to call the matching IAM API. Names sourced from
+// AWSMachinePool / AWSMachineTemplate are instance profile names and are
+// resolved via GetInstanceProfile (the resulting role ARN is taken from the
+// instance profile's attached role). Names sourced from AWSManagedMachinePool
+// are role names and are resolved via GetRole.
+func (s *Service) resolveRoleARN(ctx context.Context, name string, kind nameKind) (string, error) {
```
AndiDog, line 111:
```suggestion
// dispatching on `kind` to call the matching IAM API.
func (s *Service) resolveRoleARN(ctx context.Context, name string, kind nameKind) (string, error) {
```

(too repetitive)

pkg/cloud/services/iamauth/reconcile.go
```diff
@@ -104,8 +133,30 @@ func (s *Service) getARNForRole(ctx context.Context, role string) (string, error
...
 }
 
-func (s *Service) getRolesForWorkers(ctx context.Context) (map[string]struct{}, error) {
-	allRoles := map[string]struct{}{}
+// getARNForInstanceProfile returns the ARN of the IAM role attached to the
+// named instance profile. AWS limits an instance profile to one attached role
+// (the Roles list in the GetInstanceProfile response is a schema artifact),
+// so we read Roles[0]. An instance profile with no role attached is treated
+// as an error.
+func (s *Service) getARNForInstanceProfile(ctx context.Context, name string) (string, error) {
```
AndiDog, line 141: The name makes it sound like it returns the instance profile's ARN, but it returns the role ARN. Please rename accordingly.

AndiDog (commented):
LGTM, just minor stuff

PR #6080, opened 2026-06-26
🐛 fix: allow cluster-scoped resources to bypass allowedNamepsaces checks

### LINKED ISSUE #6087: Cluster-scoped resources cannot use identities with allowedNamespaces restrictions
/kind bug

When a cluster-scoped resource attempts to use an `AWSClusterRoleIdentity` with `allowedNamespaces` restrictions, the permission check in `isClusterPermittedToUsePrincipal()` fails.

Cluster-scoped resources do not have a `.metadata.namespace` field (they return `""`), so there is no namespace to validate against the `allowedNamespaces` list. This causes the permission check to reject the identity even though cluster-scoped resources should be allowed to use it.

Cluster-scoped resources should be able to use AWS identities with `allowedNamespaces` restrictions. The permission check should recognize that cluster-scoped resources don't have a namespace and handle
them appropriately rather than failing validation.

**Root cause:**
The `isClusterPermittedToUsePrincipal()` function assumes all resources have a namespace and checks if that namespace is in the `allowedNamespaces` list. Cluster-scoped resources return an empty namespace (`""`), which never matches any entry in the allowed list.

**Impact:**
 - Blocks cluster-scoped resources from using namespace-restricted identities for cross-account access
- Forces use of unrestricted identities (`allowedNamespaces: {}`), defeating multi-tenancy
- Affects any future cluster-scoped resources that need to use AWS identities

**Workaround:**
Set the identity's `allowedNamespaces: {}` (empty object) to allow all namespaces. However this removes namespace restrictions entirely, defeating the multi-tenancy security boundary.

**Environment:** 
- Cluster-api-provider-aws version: v2.x
- Kubernetes version: (use kubectl version): v1.35.0
- OS (e.g. from /etc/os-release): OS-agnostic

andclt:
**What type of PR is this?**

  /kind bug

 This PR fixes a bug where cluster-scoped resources cannot use AWS identities that have `allowedNamespaces` restrictions.

  The permission check in `isClusterPermittedToUsePrincipal()` assumes all resources have a namespace, but cluster-scoped resources return an empty namespace (`""`), causing them to fail
  `allowedNamespaces` validation even when they should be allowed.

  This fix allows cluster-scoped resources to bypass `allowedNamespaces` checks since:
  1. Cluster-scoped resources don't belong to a namespace
  2. The `allowedNamespaces` field is designed for namespace-scoped resources to enforce multi-tenancy
  3. Cluster-scoped resources are already protected by cluster-level RBAC (only cluster admins can create them)

  Changes:
  - Added `isClusterScopedResource()` helper that uses `RESTMapper` to query resource scope metadata from CRD definitions.
  - Cluster-scoped resources bypass `allowedNamespaces` validation

  Fixes #6087
https://redhat.atlassian.net/browse/ROSAENG-60948

  **AI Usage**:

 None

  **Checklist**:

  - [x] squashed commits
  - [x] includes documentation
  - [x] includes emoji in title
  - [x] adds unit tests

  **Release note**:

```release-note
fix: allow cluster-scoped resources to bypass allowedNamepsaces checks
```

### DIFF
--- a/pkg/cloud/scope/session.go
+++ b/pkg/cloud/scope/session.go
@@ -31,8 +31,10 @@ import (
 	"github.com/google/go-cmp/cmp"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud"
@@ -260,7 +262,7 @@ func buildProvidersForRef(
 			return providers, err
 		}
 		log.Trace("Principal retrieved")
-		canUse, err := isClusterPermittedToUsePrincipal(k8sClient, roleIdentity.Spec.AllowedNamespaces, clusterScoper.Namespace())
+		canUse, err := isClusterPermittedToUsePrincipal(k8sClient, roleIdentity.Spec.AllowedNamespaces, clusterScoper.Namespace(), clusterScoper.InfraCluster())
 		if err != nil {
 			return providers, err
 		}
@@ -337,7 +339,7 @@ func buildAWSClusterStaticIdentity(ctx context.Context, identityObjectKey client
 		return nil, errors.Wrapf(err, "failed to patch secret name:%s namespace:%s", secret.Name, secret.Namespace)
 	}
 
-	canUse, err := isClusterPermittedToUsePrincipal(k8sClient, staticPrincipal.Spec.AllowedNamespaces, clusterScoper.Namespace())
+	canUse, err := isClusterPermittedToUsePrincipal(k8sClient, staticPrincipal.Spec.AllowedNamespaces, clusterScoper.Namespace(), clusterScoper.InfraCluster())
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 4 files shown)


pkg/cloud/scope/session.go
```diff
@@ -386,7 +387,26 @@ func getProvidersForCluster(ctx context.Context, k8sClient client.Client, cluste
...
-func isClusterPermittedToUsePrincipal(k8sClient client.Client, allowedNs *infrav1.AllowedNamespaces, clusterNamespace string) (bool, error) {
+func isClusterScopedResource(k8sClient client.Client, obj client.Object) bool {
+	gvk := obj.GetObjectKind().GroupVersionKind()
+
+	restMapper := k8sClient.RESTMapper()
+	mapping, err := restMapper.RESTMapping(gvk.GroupKind(), gvk.Version)
+	if err != nil {
+		// If RESTMapper fails, fall back to checking if the object has an empty namespace.
+		// Cluster-scoped resources have no namespace field.
+		return obj.GetNamespace() == ""
```
serngawy: I believe we should return the error not ignoring it. Otherwise may be just check if the namespace is empty without checking the `Scope.Name()`

pkg/cloud/scope/session.go
```diff
@@ -386,7 +387,26 @@ func getProvidersForCluster(ctx context.Context, k8sClient client.Client, cluste
...
+	}
+
+	return mapping.Scope.Name() == meta.RESTScopeNameRoot
```
serngawy: we should check for nil/empty values 
```suggestion
if mapping != nil && mapping.Scope != nil 
	return mapping.Scope.Name() == meta.RESTScopeNameRoot
```

pkg/cloud/scope/session.go
andclt: Good point about error handling, my original thinking was that the fallback would provide resilience against transient issues but if we're in the reconcile loop with an object, its CRD must already be registered for that object to exist in the API server. So if `RESTMapper` fails that's likely a real problem that we should surface rather than hide. I'll update it to return the error instead.

serngawy (approved):
looks good to me

damdo:
/hold

for testing

PR #6099, opened 2026-07-07
🐛 fix: create CAPI-contract kubeconfig for external auth ROSA clusters

judexzhu:
## Summary

When `EnableExternalAuthProviders: true` is set on a ROSAControlPlane, CAPA only created a `<cluster>-bootstrap-kubeconfig` secret (24h TTL, no refresh) but never the CAPI-contract `<cluster>-kubeconfig` secret. In CAPI v1.11, `RemoteConnectionProbe` became load-bearing — it gates `ClusterAvailableV1Beta2Condition`. Without a valid `<cluster>-kubeconfig`, all external-auth ROSA clusters report `Available=False`, and CAPI loses MachinePool status reporting.

### Root Cause

In `rosacontrolplane_controller.go`, the reconcile logic splits on `EnableExternalAuthProviders`:
- Standard path → `reconcileKubeconfig()` creates `<cluster>-kubeconfig` (CAPI contract) via OAuth token
- External auth path → `reconcileExternalAuthBootstrapKubeconfig()` creates only `<cluster>-bootstrap-kubeconfig` via break-glass credential

The standard path can't work for external OIDC because there is no OpenShift OAuth server.

### Fix

- Create **both** `<cluster>-bootstrap-kubeconfig` (for user RBAC setup) and `<cluster>-kubeconfig` (for CAPI/RemoteConnectionProbe) from the same break-glass credential
- Track credential expiry via `controlplane.cluster.x-k8s.io/rosacontrolplane-credential-expiry` annotation
- Auto-refresh break-glass credential before the 24h TTL expires (1h threshold)
- Use `kubeconfig.GenerateSecretWithOwner()` for the CAPI secret to ensure contract compliance (labels, owner refs, secret type)

### Changes

- **`rosacontrolplane_controller.go`**: New annotation constant, modified `reconcileExternalAuthBootstrapKubeconfig` to check expiry and write dual secrets, new `needsCredentialRefresh` helper
- **`rosacontrolplane_controller_test.go`**: Table-driven tests for `needsCredentialRefresh` (5 cases) + envtest integration tests for dual secret create/update

## Test plan

- [x] `make generate` — no generated file changes
- [x] `make lint` — 0 issues
- [x] `make test` — 49/49 passed (full controller package)
- [ ] E2E validation on a ROSA cluster with `EnableExternalAuthProviders: true` (requires OCM credentials)

```release-note
fix(rosa): RemoteConnectionProbe failure for external auth clusters by creating and auto-refreshing the CAPI-contract kubeconfig secret
```

### DIFF
--- a/controlplane/rosa/controllers/rosacontrolplane_controller.go
+++ b/controlplane/rosa/controllers/rosacontrolplane_controller.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"maps"
 	"net"
 	"net/url"
 	"reflect"
@@ -85,6 +86,12 @@ const (
 
 	// ExternalAuthProviderLastAppliedAnnotation annotation tracks the last applied external auth configuration to inform if an update is required.
 	ExternalAuthProviderLastAppliedAnnotation = "controlplane.cluster.x-k8s.io/rosacontrolplane-last-applied-external-auth-provider"
+
+	// ROSAControlPlaneCredentialExpiryAnnotation tracks when the break-glass credential in a kubeconfig secret expires.
+	ROSAControlPlaneCredentialExpiryAnnotation = "controlplane.cluster.x-k8s.io/rosacontrolplane-credential-expiry"
+
+	// credentialRefreshThreshold is how long before expiry a break-glass credential should be refreshed.
+	credentialRefreshThreshold = 1 * time.Hour
 )
 
 // ROSAControlPlaneReconciler reconciles a ROSAControlPlane object.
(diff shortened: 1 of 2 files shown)


controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -969,23 +976,31 @@ func (r *ROSAControlPlaneReconciler) reconcileExternalAuthProviders(ctx context.
...
+		bootstrapSecret = nil
+	}
+
+	clusterRef := client.ObjectKeyFromObject(rosaScope.Cluster)
+	capiSecret, err := secret.GetFromNamespacedName(ctx, r.Client, clusterRef, secret.Kubeconfig)
+	if err != nil && !apierrors.IsNotFound(err) {
+		return fmt.Errorf("failed to get kubeconfig secret: %w", err)
+	}
+
+	if !needsCredentialRefresh(bootstrapSecret) && !needsCredentialRefresh(capiSecret) {
```
serngawy, line 996: Just to confirm we changing behavior for bootstrapSecret. Before it was never  rotated after creation. Only capiSecret was rotated every 24h. We would expect that new behavior, right ?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1001,14 +1016,69 @@ func (r *ROSAControlPlaneReconciler) reconcileExternalAuthBootstrapKubeconfig(ct
...
+		return err
 	}
 
-	return nil
+	return r.reconcileCAPIKubeconfigSecret(ctx, rosaScope, kubeconfigData, expiryAnnotation, capiSecret)
+}
+
+func (r *ROSAControlPlaneReconciler) reconcileBootstrapKubeconfigSecret(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope, kubeconfigData string, annotations map[string]string, existing *corev1.Secret) error {
+	if existing != nil {
+		existing.Data = map[string][]byte{"value": []byte(kubeconfigData)}
```
serngawy: better to use ref dataname
```suggestion
		existing.Data = map[string][]byte{secret.KubeconfigDataName: []byte(kubeconfigData)}
```

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1001,14 +1016,69 @@ func (r *ROSAControlPlaneReconciler) reconcileExternalAuthBootstrapKubeconfig(ct
...
+		if existing.Annotations == nil {
+			existing.Annotations = make(map[string]string)
+		}
+		maps.Copy(existing.Annotations, annotations)
+		return r.Client.Update(ctx, existing)
+	}
+
+	newSecret := rosaScope.ExternalAuthBootstrapKubeconfigSecret()
+	newSecret.Data = map[string][]byte{"value": []byte(kubeconfigData)}
```
serngawy: same
```suggestion
	newSecret.Data = map[string][]byte{secret.KubeconfigDataName: []byte(kubeconfigData)}
```

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -1001,14 +1016,69 @@ func (r *ROSAControlPlaneReconciler) reconcileExternalAuthBootstrapKubeconfig(ct
...
+	newSecret.Annotations = annotations
+	return r.Client.Create(ctx, newSecret)
+}
+
+func (r *ROSAControlPlaneReconciler) reconcileCAPIKubeconfigSecret(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope, kubeconfigData string, annotations map[string]string, existing *corev1.Secret) error {
```
serngawy: For better coding the `reconcileBootstrapKubeconfigSecret` AND `reconcileBootstrapKubeconfigSecret` are very similar. Can we check/create the secret before passing to the function and make it 1 function instead.

serngawy (commented):
looks good in general, few comments

judexzhu:
@serngawy Updated, Please kindly help review again. Thanks

damdo:
cc. @andclt

serngawy (commented):
Thanks @judexzhu , looks good to me. Would you squash commits so we can merge it.

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -85,6 +86,12 @@ const (
 
 	// ExternalAuthProviderLastAppliedAnnotation annotation tracks the last applied external auth configuration to inform if an update is required.
 	ExternalAuthProviderLastAppliedAnnotation = "controlplane.cluster.x-k8s.io/rosacontrolplane-last-applied-external-auth-provider"
+
+	// ROSAControlPlaneCredentialExpiryAnnotation tracks when the break-glass credential in a kubeconfig secret expires.
+	ROSAControlPlaneCredentialExpiryAnnotation = "controlplane.cluster.x-k8s.io/rosacontrolplane-credential-expiry"
```
andclt, line 91: should we also consider storing the credential expiry as a status field instead of an annotation? That would give us type-safety and better observability (would be visible in `kubectl get`)

serngawy: The annotation stored at the kubeconfig secret CR, Do you mean in the ROSAControlPlane status ?

andclt: yes exactly, I meant storing it in the `ROSAControlPlane.Status` field.
I think the current approach works fine too but I was wondering if storing it as a typed field in `ROSAControlPlane.Status` could be more idiomatic because it's an observable state and it's type safe so there's no need to parse the time strings.

serngawy: hmm, that will be API changes, better to have it in another PR as enhancement. We need to cherry-pick this to release-2.11

andclt: makes sense, that works for me thanks!

serngawy (approved):

nrb (approved):
/approve

PR #6122, opened 2026-07-13
✨ feat(ROSA): add delete protection support to ROSA

### LINKED ISSUE #6121: Add delete protection support for ROSA
/kind feature

Add OCM delete protection support to `ROSAControlPlane` (ROSA HCP only).

ROSA HCP clusters managed by CAPA have no way to enable OCM delete protection at create time or to manage it as a day-2 setting on the `ROSAControlPlane`. Accidental deletion of the control plane CR can trigger OCM cluster deletion without a guard that matches other ROSA tooling.

Proposed solution: 

1. **API (`ROSAControlPlane`)**
   - Add optional `spec.deleteProtection` (`*bool`) to request delete protection for the OCM cluster.
   - Add optional `status.deleteProtection` (`*bool`) to report the observed OCM value after reconciliation.

2. **Reconciler behavior**
   - **Create:** After `CreateCluster()`, if `spec.deleteProtection` is `true`, PATCH the OCM delete protection sub-resource to enable protection. Persist `status.id` even if the PATCH fails.
   - **Update:** Reconcile spec vs live OCM state (inline cluster field with fallback GET on `/delete_protection`); PATCH when they differ. Update `status.deleteProtection`.
   - **Delete:** Block deletion when `spec.deleteProtection` is `true` or live OCM protection is enabled. Surface a clear condition/message instructing the user to set `spec.deleteProtection: false` and wait for reconciliation. The `rosacontrolplane-force-delete` annotation must **not** bypass this check.

**Environment:**

- Cluster-api-provider-aws version: `main`
- Kubernetes version: (use `kubectl version`): `any supported k8s version`
- OS (e.g. from `/etc/os-release`): `any supported OS`

michaelryanmcneill:
**What type of PR is this?**

/kind feature
/kind api-change

ROSA provides support for deletion protection, but CAPA did not expose or reconcile that setting on `ROSAControlPlane`. This PR adds optional `spec.deleteProtection` and `status.deleteProtection` fields and wires them through the controller.

Changes include:

- Add `spec.deleteProtection` and `status.deleteProtection` to `ROSAControlPlane` (`v1beta2`) and regenerate CRDs/deepcopy.
- Reconcile delete protection during normal reconciliation via `pkg/rosa/delete_protection.go`, including syncing spec to OCM and updating status from the cluster object or OCM sub-resource.
- Enable delete protection immediately after cluster creation when requested in spec.
- When a `ROSAControlPlane` is deleted while protection is enabled, remove the finalizer and allow the CR to be deleted without cascading deletion to the ROSA cluster in OCM.
- Extend `OCMClient` with `UpdateClusterDeletionProtection` and add `NewOCMClientFromRosaClient` for tests that use a real `*ocm.Client`.
- Document the disable-before-delete workflow in the ROSA cluster creation guide and add a commented example to `templates/cluster-template-rosa.yaml`.
- Add unit tests in `pkg/rosa/` and update controller tests for the new reconcile path.

Transient OCM failures during delete-protection reconciliation log the error and requeue after one minute instead of returning a reconcile error alongside `RequeueAfter`.

Fixes #6121 

- When delete protection is enabled and the CR is deleted, the finalizer is removed and the CR is garbage-collected, but the ROSA cluster in OCM is left intact. No machine pool cascade or OCM delete call occurs.
- No ROSA-specific e2e coverage exists today; validation is unit/controller tests plus manual checks below.

**AI Usage**:

Cursor was used to assist with test authoring. CodeRabbit was used to review the change before opening the PR. All changes have been reviewed and validated locally before submission.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**How to test**:

- [x] `make generate`
- [x] `make lint`
- [x] `make lint-api`
- [x] `make test`
- [x] `make binaries`
- [x] `make compile-e2e`
- [x] `go test ./pkg/rosa/... -run 'DeleteProtection|ReconcileDeleteProtection|IsDeleteProtectionBlocking' -v`
- [x] Manual: set `spec.deleteProtection: true`, confirm `status.deleteProtection` becomes `true`
- [x] Manual: set `spec.deleteProtection: false`, confirm status follows and OCM protection is cleared
- [x] Manual: attempt delete while protected and confirm deletion is blocked with the documented message
- [x] Manual: disable protection, wait for reconcile, then delete successfully

**Release note**:

```release-note
ROSAControlPlane now supports optional `spec.deleteProtection` and reports live protection state in `status.deleteProtection`. When enabled, CAPA syncs the setting to OCM. Deleting the ROSAControlPlane CR while protection is enabled removes the CR without cascading deletion to the ROSA cluster in OCM.
```

### DIFF
--- a/controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
+++ b/controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
@@ -82,6 +82,17 @@ const (
 	FIPSDisabled FIPSState = "Disabled"
 )
 
+// DeleteProtectionState represents whether delete protection is enabled for the ROSA cluster.
+type DeleteProtectionState string
+
+const (
+	// DeleteProtectionEnabled indicates delete protection is enabled.
+	DeleteProtectionEnabled DeleteProtectionState = "Enabled"
+
+	// DeleteProtectionDisabled indicates delete protection is disabled.
+	DeleteProtectionDisabled DeleteProtectionState = "Disabled"
+)
+
 // AutoNodeMode specifies the AutoNode mode for the ROSA Control Plane.
 type AutoNodeMode string
 
@@ -201,6 +212,15 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// +kubebuilder:validation:XValidation:rule="self == oldSelf", message="enableExternalAuthProviders is immutable"
 	// +optional
 	EnableExternalAuthProviders bool `json:"enableExternalAuthProviders,omitempty"`
+
+	// DeleteProtection prevents accidental ROSA cluster deletion.
+	// When set to "Enabled", the ROSA cluster cannot be deleted through OCM.
+	// Defaults to "Disabled".
+	//
+	// +kubebuilder:default=Disabled
+	// +kubebuilder:validation:Enum=Enabled;Disabled
+	// +optional
+	DeleteProtection DeleteProtectionState `json:"deleteProtection,omitempty"`
 
 	// ExternalAuthProviders are external OIDC identity providers that can issue tokens for this cluster.
 	// Can only be set if "enableExternalAuthProviders" is set to "True".
--- a/controlplane/rosa/controllers/rosacontrolplane_controller.go
+++ b/controlplane/rosa/controllers/rosacontrolplane_controller.go
@@ -286,6 +286,11 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		rosaScope.ControlPlane.Status.Ready = false
 		rosaScope.ControlPlane.Status.Version = rosa.RawVersionID(cluster.Version())
 
+		if err := rosa.ReconcileDeleteProtection(rosaScope, ocmClient, cluster); err != nil {
+			rosaScope.Error(err, "failed to reconcile delete protection")
+			return ctrl.Result{RequeueAfter: time.Minute}, nil
+		}
+
 		switch cluster.Status().State() {
 		case cmv1.ClusterStateReady:
 			v1beta1conditions.MarkTrue(rosaScope.ControlPlane, rosacontrolplanev1.ROSAControlPlaneReadyCondition)
@@ -386,6 +391,19 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 	rosaScope.Info("cluster created", "state", cluster.Status().State())
 	rosaScope.ControlPlane.Status.ID = cluster.ID()
 
+	if rosaScope.ControlPlane.Spec.DeleteProtection == rosacontrolplanev1.DeleteProtectionEnabled {
+		if err := rosa.UpdateClusterDeletionProtection(ocmClient, cluster.ID(), true); err != nil {
+			v1beta1conditions.MarkFalse(rosaScope.ControlPlane,
+				rosacontrolplanev1.ROSAControlPlaneReadyCondition,
+				rosacontrolplanev1.ReconciliationFailedReason,
+				clusterv1beta1.ConditionSeverityError,
+				"cluster '%s' was created but delete protection could not be enabled: %s",
+				cluster.ID(),
+				err.Error())
+			return ctrl.Result{}, fmt.Errorf("failed to enable delete protection for cluster '%s': %w", cluster.ID(), err)
+		}
+	}
+
 	return ctrl.Result{}, nil
 }
 
@@ -457,6 +475,13 @@ func (r *ROSAControlPlaneReconciler) reconcileDelete(ctx context.Context, rosaSc
 	}
 	if cluster == nil {
 		// cluster and machinepools are deleted, removing finalizer.
+		controllerutil.RemoveFinalizer(rosaScope.ControlPlane, ROSAControlPlaneFinalizer)
+
+		return ctrl.Result{}, nil
+	}
+
+	if rosa.IsDeleteProtectionBlocking(rosaScope, cluster) {
+		rosaScope.Info("Delete protection is enabled, removing finalizer without deleting the ROSA cluster")
 		controllerutil.RemoveFinalizer(rosaScope.ControlPlane, ROSAControlPlaneFinalizer)
 
 		return ctrl.Result{}, nil
--- a/docs/book/src/topics/rosa/creating-a-cluster.md
+++ b/docs/book/src/topics/rosa/creating-a-cluster.md
@@ -345,6 +345,12 @@ The CAPA controller requires service account credentials to provision ROSA HCP c
 
 ## Deleting a ROSA HCP cluster
 
+When `spec.deleteProtection` is set to `Enabled` on the `ROSAControlPlane`, deleting the `ROSAControlPlane` CR will remove the Kubernetes resource but leave the ROSA cluster intact in OCM. No cascade deletion of NodePools or the cluster itself will occur. To fully delete a protected cluster:
+
+1. Set `spec.deleteProtection` to `Disabled` on the `ROSAControlPlane`.
+2. Wait for reconciliation to complete.
+3. Delete the `Cluster` and `ROSAControlPlane` resources.
+
 To delete a ROSA HCP cluster, delete the `Cluster` and `ROSAControlPlane` resources. This will also clean up the associated `ROSACluster`, `MachinePool`, and `ROSAMachinePool` resources:
 
 ```shell
--- a/exp/controllers/rosaocmroleconfig_controller_test.go
+++ b/exp/controllers/rosaocmroleconfig_controller_test.go
@@ -520,7 +520,7 @@ func TestROSAOCMRoleConfigSetUpRuntimeWithExpiredAWSCredentials(t *testing.T) {
 		Client: testEnv.Client,
 		NewOCMClient: func(ctx context.Context, scope caparosa.OCMSecretsRetriever) (caparosa.OCMClient, error) {
 			ocmClientCallCount++
-			return ocmClient, nil
+			return caparosa.NewOCMClientFromRosaClient(ocmClient), nil
 		},
 		runtimeFactory: nil, // Use default setUpRuntime (not the mock factory)
 	}
--- a/pkg/rosa/delete_protection.go
+++ b/pkg/rosa/delete_protection.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+	http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package rosa
+
+import (
+	"fmt"
+
+	cmv1 "github.com/openshift-online/ocm-sdk-go/clustersmgmt/v1"
+
+	rosacontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/rosa/api/v1beta2"
+	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/scope"
+)
+
+// DeleteProtectionEnabledFromCluster reads delete protection from the inline cluster field.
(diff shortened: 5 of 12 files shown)


controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -202,6 +202,10 @@ type RosaControlPlaneSpec struct { //nolint: maligned
 	// +optional
 	EnableExternalAuthProviders bool `json:"enableExternalAuthProviders,omitempty"`
 
+	// DeleteProtection prevents accidental ROSA cluster deletion when true.
+	// +optional
+	DeleteProtection *bool `json:"deleteProtection,omitempty"`
```
serngawy: K8s API design recommend not to use boolean, better to make it as enum enabled/disbaled and set default to disabled.

controlplane/rosa/api/v1beta2/rosacontrolplane_types.go
```diff
@@ -925,6 +929,10 @@ type RosaControlPlaneStatus struct {
 
 	// Available channels for the ROSA hosted control plane.
 	AvailableChannels []string `json:"availableChannels,omitempty"`
+
+	// DeleteProtection reports whether delete protection is enabled for the ROSA cluster.
+	// +optional
+	DeleteProtection *bool `json:"deleteProtection,omitempty"`
```
serngawy: No need to show DeleteProtection in status its already in the spec

exp/controllers/rosaocmroleconfig_controller_test.go
```diff
@@ -520,7 +520,7 @@ func TestROSAOCMRoleConfigSetUpRuntimeWithExpiredAWSCredentials(t *testing.T) {
 		Client: testEnv.Client,
 		NewOCMClient: func(ctx context.Context, scope caparosa.OCMSecretsRetriever) (caparosa.OCMClient, error) {
 			ocmClientCallCount++
-			return ocmClient, nil
+			return caparosa.NewOCMClientFromRosaClient(ocmClient), nil
```
serngawy, line 523: please scope the changes only for the rosacontrolplane.

pkg/rosa/ocmclient.go
```diff
@@ -191,11 +205,7 @@ func NewMockOCMClient(ctx context.Context, rosaScope *scope.ROSAControlPlaneScop
 func ConvertToRosaOcmClient(i OCMClient) (*ocm.Client, error) {
 	c, ok := i.(*ocmclient)
 	if !ok {
-		c, ok := i.(*ocm.Client)
-		if !ok {
-			return nil, fmt.Errorf("failed to convert to Rosa OCM Client")
-		}
-		return c, nil
+		return nil, fmt.Errorf("failed to convert to Rosa OCM Client")
```
serngawy, line 208: I believe failing safe with ocm.Client cast was for a reason.  Pleas keep it

pkg/rosa/delete_protection.go
serngawy, line 1: Why do we have delete protection with http client API request ? 
Would better to include the filed with [ocmClusterBuild](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/controllers/rosacontrolplane_controller.go#L1208) function and [ocmClusterUpdate](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controlplane/rosa/controllers/rosacontrolplane_controller.go#L718) function instead ?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -435,16 +454,6 @@ func (r *ROSAControlPlaneReconciler) reconcileRosaRoleConfig(ctx context.Context
 func (r *ROSAControlPlaneReconciler) reconcileDelete(ctx context.Context, rosaScope *scope.ROSAControlPlaneScope, creator *rosaaws.Creator) (res ctrl.Result, reterr error) {
 	rosaScope.Info("Reconciling ROSAControlPlane delete")
 
-	// Deleting MachinePools first.
```
serngawy, line 438: The order of deletion need to be kept other wise the ROSAMachienPool CR will remain not deleted.  Better to modify the ROSAMachinePool controller [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosamachinepool_controller.go#L320)  check for the owner ROSAControlPlane `rosaControlPlaneScope` if the spec.DeleteProtection is enabled then skip delete and just remove the finalizer.

serngawy (changes requested):

pkg/rosa/delete_protection.go
michaelryanmcneill: I dug into this and it turns out `ocm.Spec` (from the github.com/openshift/rosa library) doesn't have a `DeleteProtection` field at all. OCM treats delete protection as its own sub-resource (`/clusters/{id}/delete_protection`) rather than a field on the cluster create/update payload, and the rosa library mirrors that with a dedicated `UpdateClusterDeletionProtection` method. So there's no way to thread it through `buildOCMClusterSpec`/`updateOCMClusterSpec` without changing the upstream library (which we can't do anyways because it would require bumping go). I did remove the raw HTTP read path entirely and instead use the inline field now.

exp/controllers/rosaocmroleconfig_controller_test.go
michaelryanmcneill: I tried to figure this one out, but I couldn't. I asked Claude and it said:

> Adding `UpdateClusterDeletionProtection` to the `OCMClient` interface means `*ocm.Client` no longer satisfies it (different method signature: `*cmv1.DeleteProtection` vs `bool`). The test must use `NewOCMClientFromRosaClient` to wrap it. This is a necessary consequence of extending the interface.

Any suggestions on how to scope this change down?

pkg/rosa/ocmclient.go
michaelryanmcneill: I don't see a way to fail safe in this because of the [same issue above](https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6122#discussion_r3575285261). 

I also asked Claude and it said: 

> `*ocm.Client` no longer implements `OCMClient` because of the signature mismatch on `UpdateClusterDeletionProtection`. Go's compiler enforces this at the type assertion level — if the asserted-to type can't possibly satisfy the interface, `i.(*ocm.Client)` is a compile error ("impossible type assertion"), not just a runtime false. So we literally cannot write this code; it won't compile.
> Before our change, the fallback worked because `*ocm.Client` happened to implement every method on `OCMClient`. Once we added a method whose signature differs between our interface and the concrete type, that relationship broke permanently. The fallback was already somewhat fragile — it relied on `*ocm.Client` coincidentally matching our interface rather than being explicitly wrapped — and extending the interface exposed that fragility. The `NewOCMClientFromRosaClient` wrapper is the correct long-term pattern: callers explicitly wrap `*ocm.Client`, and `ConvertToRosaOcmClient` only needs to unwrap `*ocmclient`.

Any suggestions on how to keep this and have it actually compile?

pkg/rosa/delete_protection.go
serngawy: hmm, the golang bump is required :(

exp/controllers/rosaocmroleconfig_controller_test.go
serngawy: okay

pkg/rosa/delete_protection_reconcile_test.go
serngawy, line 1: better to have this unit test at rosacontrolplane_controller_test.go OR may move the delete_protection_reconcile_test.go file under controlplane/rosa/controllers/

pkg/rosa/ocmclient.go
serngawy: okay, same as above

pkg/rosa/delete_protection_reconcile_test.go
michaelryanmcneill: Per discussion, consolidated this into `delete_protection_test.go‎`. Thanks!

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -286,6 +286,11 @@ func (r *ROSAControlPlaneReconciler) reconcileNormal(ctx context.Context, rosaSc
 		rosaScope.ControlPlane.Status.Ready = false
 		rosaScope.ControlPlane.Status.Version = rosa.RawVersionID(cluster.Version())
 
+		if err := rosa.ReconcileDeleteProtection(rosaScope, ocmClient, cluster); err != nil {
+			rosaScope.Error(err, "failed to reconcile delete protection")
+			return ctrl.Result{RequeueAfter: time.Minute}, nil
+		}
```
andclt, line 292: On this line we don't know if the cluster is ready yet or not. For a cluster still provisioning, this will attempt to patch delete-protection on a cluster that may bot be fully created in OCM. Could there be any scenarios where this would cause the OCM API to reject it or behaved unexpectedly?

docs/book/src/topics/rosa/creating-a-cluster.md
```diff
@@ -345,6 +345,14 @@ The CAPA controller requires service account credentials to provision ROSA HCP c
 
 ## Deleting a ROSA HCP cluster
 
+When `spec.deleteProtection` is enabled on the `ROSAControlPlane`, OCM blocks cluster deletion until protection is explicitly disabled and reconciled. Before deleting a protected cluster:
+
+1. Set `spec.deleteProtection` to `false` on the `ROSAControlPlane`.
+2. Wait for reconciliation to complete and confirm `status.deleteProtection` is `false`.
```
andclt: we should update this since we are no longer using a boolean

templates/cluster-template-rosa.yaml
```diff
@@ -50,3 +50,4 @@ spec:
   installerRoleARN: "arn:aws:iam::${AWS_ACCOUNT_ID}:role/${ACCOUNT_ROLES_PREFIX}-HCP-ROSA-Installer-Role"
   supportRoleARN: "arn:aws:iam::${AWS_ACCOUNT_ID}:role/${ACCOUNT_ROLES_PREFIX}-HCP-ROSA-Support-Role"
   workerRoleARN: "arn:aws:iam::${AWS_ACCOUNT_ID}:role/${ACCOUNT_ROLES_PREFIX}-HCP-ROSA-Worker-Role"
+  # deleteProtection: true
```
andclt: this uses true instead of Enabled - furthermore, is this comment still needed?

exp/controllers/rosaocmroleconfig_controller_test.go
```diff
@@ -5,7 +5,7 @@ Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
 
-    http://www.apache.org/licenses/LICENSE-2.0
+	http://www.apache.org/licenses/LICENSE-2.0
```
andclt: nit: tab-vs-space change

pkg/rosa/ocmclient.go
andclt: Instead of adding `UpdateClusterDeletionProtection` to the `OCMClient` interface, I believe we could use `rosa.ConvertToRosaOcmClient(ocmClient)` to get the underlying `*ocm.Client` and call its native method directly as we are doing for `ROSARoleConfigController` here: https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosaroleconfig_controller.go#L504.

However this approach has drawbacks too

controlplane/rosa/controllers/rosacontrolplane_controller.go
michaelryanmcneill: Yes, this is intended. The OCM delete protection sub-resource works as soon as the cluster exists (even before it is ready). This also matches the behavior of the ROSA CLI and Terraform provider as well.

docs/book/src/topics/rosa/creating-a-cluster.md
michaelryanmcneill: Good catch, updated the docs!

templates/cluster-template-rosa.yaml
michaelryanmcneill: Fixed! I do think it is worth keeping the comment there as an optional field.

exp/controllers/rosaocmroleconfig_controller_test.go
michaelryanmcneill: Thanks for catching this, fixed.

michaelryanmcneill:
@andclt @serngawy requested changes have been addressed. Ready for re-review!

serngawy (approved):

andclt:
I don't have permissions but LGTM

richardcase:
/approve

Someone else will have to lgtm

PR #6124, opened 2026-07-15
:bug: fix: Clear credential cache on InvalidClientTokenId to allow retry on next reconcile

### LINKED ISSUE #6123: AWSRolePrincipalTypeProvider caches InvalidClientTokenId error after SDK v2 migration, preventing recovery on next reconcile
## What happened?

When using `AWSClusterStaticIdentity` + `AWSClusterRoleIdentity` with freshly-created IAM access keys (a common pattern for multi-tenant setups), the CAPA controller fails with `InvalidClientTokenId` due to AWS IAM eventual consistency for new principals.

After the AWS SDK v1 to v2 migration, this transient error becomes persistent because `aws.NewCredentialsCache` caches the failed result and `AWSRolePrincipalTypeProvider.Retrieve()` never clears `p.credentials` after a failure:

```go
// pkg/cloud/identity/identity.go
func (p *AWSRolePrincipalTypeProvider) Retrieve(ctx context.Context) (aws.Credentials, error) {
    if p.credentials == nil {  // once set, never cleared even on error
        // ...
        p.credentials = creds
    }
    return p.credentials.Retrieve(ctx)  // cached error persists
}
```

In SDK v1, `credentials.Credentials` had `IsExpired()` which returned true after a failure, causing the next `Retrieve()` to attempt a fresh `AssumeRole`. SDK v2's `CredentialsCache` does not have this behavior -- it caches errors for the cache duration.

## What did you expect to happen?

The controller should recover from `InvalidClientTokenId` on the next reconcile loop, as it did with SDK v1.

## How to reproduce

1. Create an IAM user + access key
2. Immediately create `AWSClusterStaticIdentity` + `AWSClusterRoleIdentity` CRs referencing that key
3. CAPA's first `AssumeRole` call fails with `InvalidClientTokenId`
4. Subsequent reconcile loops keep failing because the credentials cache is poisoned

## Observed behavior

~50 minutes of continuous `InvalidClientTokenId` errors until `KubeadmControlPlane` times out. The controller never recovers without a restart.

## Environment

- CAPA version: v2.9+ (any version after the SDK v2 migration)
- Kubernetes version: N/A (affects the management cluster controller)

## Additional context

CAPA's own e2e test suite already works around this with a `time.Sleep(10 * time.Second)` after creating access keys (`test/e2e/shared/suite.go`), acknowledging the propagation delay exists.

### Proposed fix

Clear `p.credentials` when `InvalidClientTokenId` is returned, so the next reconcile creates a fresh credentials cache:

```go
creds, err := p.credentials.Retrieve(ctx)
if err != nil && isInvalidClientTokenIdError(err) {
    p.credentials = nil
}
return creds, err
```

This restores the SDK v1 behavior with minimal change.

omargallob:
Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged): 

Fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/6123

## What this PR does

Clears the cached `*aws.CredentialsCache` in `AWSRolePrincipalTypeProvider.Retrieve()` when the error is `InvalidClientTokenId`, allowing the next controller-runtime reconcile loop to create a fresh credentials cache and retry the `AssumeRole` call.

## Why this is needed

After the AWS SDK v1→v2 migration, `aws.NewCredentialsCache` caches failed `AssumeRole` results. When a freshly-created IAM access key has not yet propagated through AWS IAM eventual consistency, the first `AssumeRole` call fails with `InvalidClientTokenId`. Unlike SDK v1 where `IsExpired()` returned true after failure (triggering a fresh attempt on the next reconcile), SDK v2's `CredentialsCache` retains the error.

This causes the CAPA controller to be unable to recover from a transient IAM propagation delay for the lifetime of the provider instance (~50 minutes observed in production until KubeadmControlPlane timeout).

## How it works

```go
result, err := p.credentials.Retrieve(ctx)
if err != nil && IsInvalidClientTokenIDError(err) {
    p.credentials = nil  // next reconcile will recreate the cache
}
return result, err
```

This restores the pre-SDK-migration behavior with minimal change.

## Testing

- Unit test added verifying cache is cleared on `InvalidClientTokenId` and subsequent `Retrieve()` succeeds
- Unit test for `IsInvalidClientTokenIDError` helper with typed and string-match cases

/kind bug

AI Usage:

1. Research & Analysis — Investigated CAPA source code (<v2.9 vs v2.11.1), identified the AWS SDK v1→v2 credential caching regression as root cause of the IAM key propagation issue
2. Code — Implemented the fix in pkg/cloud/identity/identity.go (added IsInvalidClientTokenIDError helper, modified Retrieve() to clear cache on transient error) and wrote unit tests in identity_test.go

Checklist:

- [x]  squashed commits
- [x]  includes documentation
- [x]  includes AI generated content
- [x]  includes emoji in title
- [x]  adds unit tests
- [ ]  adds or updates e2e tests

```release-note
Fix credential cache poisoning in AWSRolePrincipalTypeProvider when AssumeRole fails with InvalidClientTokenId due to IAM eventual consistency. The credential cache is now cleared on this transient error, allowing the controller to retry on the next reconcile instead of remaining stuck until restart.
```

### DIFF
--- a/pkg/cloud/identity/identity.go
+++ b/pkg/cloud/identity/identity.go
@@ -22,6 +22,8 @@ import (
 	"context"
 	"crypto/sha256"
 	"encoding/gob"
+	"errors"
+	"strings"
 	"time"
 
 	"github.com/aws/aws-sdk-go-v2/aws"
@@ -29,12 +31,19 @@ import (
 	"github.com/aws/aws-sdk-go-v2/credentials"
 	"github.com/aws/aws-sdk-go-v2/credentials/stscreds"
 	"github.com/aws/aws-sdk-go-v2/service/sts"
+	"github.com/aws/smithy-go"
 	corev1 "k8s.io/api/core/v1"
 
 	infrav1 "sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2"
 	awsmetrics "sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/metrics"
 	stsservice "sigs.k8s.io/cluster-api-provider-aws/v2/pkg/cloud/services/sts"
 	"sigs.k8s.io/cluster-api-provider-aws/v2/pkg/logger"
+)
+
+const (
+	// invalidClientTokenIDCode is the AWS error code returned when an access key
+	// has not yet propagated through IAM's eventual consistency.
+	invalidClientTokenIDCode = "InvalidClientTokenId"
 )
 
 // AWSPrincipalTypeProvider defines the interface for AWS Principal Type Provider.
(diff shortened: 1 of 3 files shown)


omargallob:
what are the next steps?

pkg/cloud/identity/identity.go
```diff
@@ -180,5 +183,30 @@ func (p *AWSRolePrincipalTypeProvider) Retrieve(ctx context.Context) (aws.Creden
 		// Update credentials
 		p.credentials = creds
 	}
-	return p.credentials.Retrieve(ctx)
+	result, err := p.credentials.Retrieve(ctx)
+	if err != nil && IsInvalidClientTokenIDError(err) {
```
richardcase, line 193: So we'd still get an error in the logs but at least the cache is cleared and on the next reconcile it would work.

Schuorz:
~~I would like to propose a little adjustment: https://github.com/omargallob/cluster-api-provider-aws/pull/2
This change should reduce log pollution, by using aws-sdk-go-v2 failure handling, in CAPA while keeping the overall cache invalidation logic.~~

Giving it a little bit more thought, I realized, while related my proposal is independent of this PR. I will create my own, to not polute this one.

PR #6128, opened 2026-07-16
🐛 fix reconciler loop at status patch update for ROSA controllers

### LINKED ISSUE #6133: ROSA controller making rapid OCM API client calls
/kind bug

The ROSAMachinePool and ROSAControlPlane controllers could enter an infinite OCM API request loop whenever a create or update operation against the OCM API failed with a persistent error.

The ROSAMachinePool and ROSAControlPlane controllers could should make ocm client API request only when the CR spec change.

**Environment:**

- Cluster-api-provider-aws version: 1.12, 1.11, 1.10

serngawy:
**What type of PR is this?**
/kind bug

The ROSAMachinePool and ROSAControlPlane controllers could enter an infinite OCM API request loop whenever a create or update operation against the OCM API  failed with a persistent error.

Fixes #6133

**AI Usage**:

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title
- [ ] adds unit tests

**Release note**:

```release-note
Fix ROSA infinite OCM API request loop when failed with a persistent error.
```

### DIFF
--- a/exp/controllers/rosamachinepool_controller.go
+++ b/exp/controllers/rosamachinepool_controller.go
@@ -23,7 +23,9 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
 	"sigs.k8s.io/controller-runtime/pkg/controller"
 	"sigs.k8s.io/controller-runtime/pkg/controller/controllerutil"
+	"sigs.k8s.io/controller-runtime/pkg/event"
 	"sigs.k8s.io/controller-runtime/pkg/handler"
+	"sigs.k8s.io/controller-runtime/pkg/predicate"
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	rosacontrolplanev1 "sigs.k8s.io/cluster-api-provider-aws/v2/controlplane/rosa/api/v1beta2"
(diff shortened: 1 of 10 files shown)


exp/controllers/rosamachinepool_controller.go
```diff
@@ -64,6 +66,25 @@ func (r *ROSAMachinePoolReconciler) SetupWithManager(ctx context.Context, mgr ct
...
+					oldPool := e.ObjectOld.(*expinfrav1.ROSAMachinePool).DeepCopy()
+					newPool := e.ObjectNew.(*expinfrav1.ROSAMachinePool).DeepCopy()
+					oldPool.Status = expinfrav1.RosaMachinePoolStatus{}
+					newPool.Status = expinfrav1.RosaMachinePoolStatus{}
+					oldPool.ObjectMeta.ResourceVersion = ""
+					newPool.ObjectMeta.ResourceVersion = ""
+					return !cmp.Equal(oldPool, newPool)
+				},
+			},
+		).
```
andclt, line 92: Could it be worth adding tests for this predicate? Creating an `UpdateEvent` with status-only changes and verifying the predicate returns false would work.

Asking this because I am worried that `e.ObjectOld.GetObjectKind().GroupVersionKind().Kind` might return an empty string and this condition would always be `true`.

serngawy: Add unit test. CR kind exist all the time, its key field for the CR.

andclt: Isn't `GetObjectKind().GroupVersionKind().Kind()` reading from `TypeMeta` struct on the object instance and not from the scheme?

IIRC when I tested the changes to allow cluster-scoped resources to bypass namespace permission checks I saw that the K8s decoder clears `TypeMeta` after deserialization so that `TypeMeta` fields are left empty for objects coming from the informer cache.

I believe the unit test is passing because we are manually setting `TypeMeta` in the test objects

serngawy: hmm, okay will change it to cast the object struct instead.
done

andclt: thank you!

PR #5248, opened 2024-12-11
🐛 fix(awscluster): update with secondary control plane load balancer

alexeysofin:
**What type of PR is this?**
/kind bug

**What this PR does / why we need it**:

The PR fixes https://github.com/kubernetes-sigs/cluster-api-provider-aws/issues/5247.

Adding a secondary load balancer to existing aws cluster.

**Special notes for your reviewer**:

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] includes [emojis](https://github.com/kubernetes-sigs/kubebuilder-release-tools?tab=readme-ov-file#kubebuilder-project-versioning)
- [x] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:
```release-note
1. fixed adding a secondary load balancer to existing aws cluster
```

Issue #5247, opened 2024-12-11
Update aws cluster with secondary control plane load balancer crashes capa controller manager


alexeysofin:
/kind bug

**What steps did you take and what happened:**
When updating AWSCluster with

```
secondaryControlPlaneLoadBalancer:
  name: internal-apiserver
  scheme: internal
```

which didn't have it previously, capa controller manager crashes with 

```
2024/12/10 17:53:00 http: panic serving 10.244.0.1:15292: runtime error: invalid memory address or nil pointer dereference
goroutine 47708 [running]:
sigs.k8s.io/cluster-api-provider-aws/v2/api/v1beta2.(*AWSCluster).validateControlPlaneLoadBalancerUpdate(0xc0020442d0?, 0x0, 0xc000913a40)
	/workspace/api/v1beta2/awscluster_webhook.go:188 +0x2c2
```

**What did you expect to happen:**

A new secondary load balancer is added successfully.

**Note**

Btw is there any documentation on how to make use of a secondary load balancer, e.g. so that kubelets use it for communication with api-server?

**Environment:**

- Cluster-api-provider-aws version: v2.7.1
- Kubernetes version: (use `kubectl version`): v1.30.5
- OS (e.g. from `/etc/os-release`): ubuntu linux 22.04

PR #6130, opened 2026-07-18
🐛 Reject removal of a secondary control plane load balancer instead of panicking

### LINKED ISSUE #6129: Removing a secondary control plane load balancer panics the AWSCluster validating webhook
/kind bug

Updating an `AWSCluster` to remove a previously configured `spec.secondaryControlPlaneLoadBalancer` by setting the field to `null` causes the validating webhook handler to panic with a nil pointer dereference:

```text
panic: runtime error: invalid memory address or nil pointer dereference
...validateControlPlaneLoadBalancerUpdate(..., <non-nil oldlb>, 0x0)
    webhooks/awscluster_webhook.go:205
...ValidateUpdate(...)
    webhooks/awscluster_webhook.go:128
```

`ValidateUpdate` constructs old/new pairs for the primary and secondary control plane load balancer specifications. It skips validation only when both values are nil.

When an existing secondary load balancer configuration is removed:

```text
oldlb != nil
newlb == nil
```

the pair is passed to `validateControlPlaneLoadBalancerUpdate`. The function enters the `oldlb != nil` branch and dereferences `newlb`, for example through `newlb.LoadBalancerType`, without first checking whether `newlb` is nil.

The Go HTTP server recovers the panic for the individual request, so the CAPA manager process remains running. However, the admission request is aborted and the `AWSCluster` update fails with an internal webhook error rather than a normal validation response.

This is the mirror case of #5247, fixed by #5248:

* #5247 handled adding a secondary load balancer: `oldlb == nil`, `newlb != nil`.
* This issue concerns removing one: `oldlb != nil`, `newlb == nil`.

The removal case is not currently covered by the update validation tests.

The validating webhook must not panic for this update.

If removing an existing secondary control plane load balancer is supported, the update should be handled safely and the corresponding AWS resources should be reconciled appropriately.

If removal is not supported, the webhook should reject the update with a clear validation error explaining that the field cannot be removed.

This can be reproduced at the webhook validation level by calling `ValidateUpdate` with:

* an old `AWSCluster` containing a non-nil `SecondaryControlPlaneLoadBalancer`; and
* a new `AWSCluster` where `SecondaryControlPlaneLoadBalancer` is nil.

A regression test should verify both that the call does not panic and that it returns the intended result: either successful validation or an explicit validation error.

I would be happy to open a PR once the intended removal semantics are confirmed. A minimal fix could explicitly reject removal without panicking; fully supporting removal may additionally require reconciliation and cleanup of the existing AWS load balancer resources.

**Environment:**

* Cluster API Provider AWS version: current `main`
* Kubernetes version: n/a; this occurs in AWSCluster webhook update validation

thc1006:
**What type of PR is this?**

/kind bug

This PR prevents the `AWSCluster` validating webhook from panicking when an update attempts to remove a previously configured `spec.secondaryControlPlaneLoadBalancer`.

`ValidateUpdate` validates old/new pairs for the primary and secondary control plane load balancer specifications. It skips a pair only when both values are nil.

Before this change, when an existing secondary load balancer configuration was removed, the validation helper received:

```text
oldlb != nil
newlb == nil
```

The helper entered the existing-load-balancer branch and dereferenced `newlb`, for example through `newlb.LoadBalancerType`, causing a nil pointer panic. The HTTP server recovered the individual request, but the admission request failed with an internal webhook error instead of returning a normal validation response.

This change:

* detects an attempted removal before dereferencing `newlb`;
* returns a clear, field-scoped validation error instead of panicking;
* passes the field path for each load balancer into the validation helper, so validation errors are attributed to the correct primary or secondary load balancer field instead of always being reported under `controlPlaneLoadBalancer`; this also corrects the existing behavior where immutable-field errors for the secondary load balancer were incorrectly reported against the primary load balancer path; and
* adds a regression test covering the `oldlb != nil, newlb == nil` case.

This is the mirror of #5247, fixed by #5248, which covered adding a secondary control plane load balancer where `oldlb == nil` and `newlb != nil`.

This PR intentionally does not add support for deleting an existing secondary control plane load balancer. Supporting removal would also require controller-side reconciliation and cleanup of the corresponding AWS resources, which is outside the scope of this webhook robustness fix.

Fixes #6129

The change is limited to update validation and its regression coverage. It does not modify the API schema, defaulting behavior, or AWS resource reconciliation.

The regression test verifies that attempting to remove an existing secondary control plane load balancer:

* does not panic;
* returns a structured `apierrors.Invalid` error; and
* includes a status cause whose `Field` is exactly `spec.secondaryControlPlaneLoadBalancer`, rather than relying only on substring matching of the rendered error message.

No e2e test is added because the failure is fully reproducible in webhook validation and does not require an AWS environment.

Testing performed:

```text
make test
```

**AI Usage**:

Claude Code was used to help analyze the failure mode, adversarially review the proposed behavior, and refine the issue and pull request descriptions. All code changes and tests were manually reviewed and validated by the author.

**Checklist**:

* [x] squashed commits
* [ ] includes documentation
* [x] includes AI generated content
* [x] includes emoji in title
* [x] adds unit tests
* [ ] adds or updates e2e tests

**Release note**:

```release-note
Prevented the AWSCluster validating webhook from panicking when removal of an existing secondary control plane load balancer is attempted. The update now returns a clear validation error.
```

### DIFF
--- a/webhooks/awscluster_webhook.go
+++ b/webhooks/awscluster_webhook.go
@@ -114,18 +114,24 @@ func (w *AWSCluster) ValidateUpdate(_ context.Context, oldObj, newObj runtime.Ob
 		)
 	}
 
-	// Validate the control plane load balancers.
-	lbs := map[*infrav1.AWSLoadBalancerSpec]*infrav1.AWSLoadBalancerSpec{
-		oldC.Spec.ControlPlaneLoadBalancer:          r.Spec.ControlPlaneLoadBalancer,
-		oldC.Spec.SecondaryControlPlaneLoadBalancer: r.Spec.SecondaryControlPlaneLoadBalancer,
+	// Validate the control plane load balancers. Each entry carries its own field
+	// path so validation errors (including the immutability checks below) are
+	// attributed to the correct load balancer instead of always to the primary.
+	lbs := []struct {
+		old  *infrav1.AWSLoadBalancerSpec
+		new  *infrav1.AWSLoadBalancerSpec
+		path *field.Path
+	}{
+		{oldC.Spec.ControlPlaneLoadBalancer, r.Spec.ControlPlaneLoadBalancer, field.NewPath("spec", "controlPlaneLoadBalancer")},
+		{oldC.Spec.SecondaryControlPlaneLoadBalancer, r.Spec.SecondaryControlPlaneLoadBalancer, field.NewPath("spec", "secondaryControlPlaneLoadBalancer")},
 	}
 
-	for oldLB, newLB := range lbs {
-		if oldLB == nil && newLB == nil {
+	for _, lb := range lbs {
+		if lb.old == nil && lb.new == nil {
 			continue
 		}
 
-		allErrs = append(allErrs, w.validateControlPlaneLoadBalancerUpdate(oldLB, newLB)...)
+		allErrs = append(allErrs, w.validateControlPlaneLoadBalancerUpdate(lb.old, lb.new, lb.path)...)
 	}
 
 	if !cmp.Equal(oldC.Spec.ControlPlaneEndpoint, clusterv1beta1.APIEndpoint{}) &&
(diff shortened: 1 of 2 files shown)


damdo (commented):
/approve

Thanks!! Looks reasonable to me

/assign @nrb @richardcase @AndiDog 

For a second review

nrb (approved):
/lgtm

PR #6132, opened 2026-07-20
🐛 Persist ProviderID immediately after instance creation

### LINKED ISSUE #6131: AWSMachine controller can create duplicate EC2 instances due to deferred ProviderID persistence
/kind bug

The AWSMachine controller can create two EC2 instances for a single Machine. Only one instance ID is persisted to the AWSMachine CR; the other instance leaks (runs unmanaged, incurring cost).

The race condition occurs because `ProviderID` and `InstanceID` are set in memory after `CreateInstance` returns ([instances.go#L299-L300](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/pkg/cloud/services/ec2/instances.go#L299-L300)), but are only persisted to the Kubernetes API server by the **deferred** `Close()` → `PatchObject()` at the end of `Reconcile` ([awsmachine_controller.go#L221-L225](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/controllers/awsmachine_controller.go#L221-L225)).

If that deferred patch fails (e.g. optimistic locking conflict from a concurrent modification to the AWSMachine), the ProviderID is lost. On the next reconcile:

1. `findInstance` sees an empty ProviderID → falls back to `GetRunningInstanceByTags`
2. The EC2 `DescribeInstances` API [follows an eventual consistency model](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html) — even though tags were applied atomically via `TagSpecifications` in `RunInstances`, the tag filter query may not return the instance yet
3. `GetRunningInstanceByTags` returns nil → the controller creates a **second** instance

The deferred `PatchObject` can fail because `reconcileNormal` already calls `PatchObject` explicitly twice before instance creation (for the finalizer and for the `InstanceProvisionStarted` condition). Each call bumps the `ResourceVersion` on the API server. Any external modification to the AWSMachine between the last explicit `PatchObject` and the deferred `Close()` causes a conflict.

A single EC2 instance is created per AWSMachine, regardless of patch conflicts or EC2 API eventual consistency.

**Environment:**

- Cluster-api-provider-aws version: current `main` (also observed in the field on v2.6)
- Kubernetes version: any
- OS: any

stefanonardo:
**What type of PR is this?**

/kind bug

After `createInstance` returns, ProviderID and InstanceID are stored only in memory. They are persisted to the API server by the deferred `Close()` at end-of-reconcile. If that patch fails (e.g. optimistic locking conflict), the next reconcile sees an empty ProviderID, falls back to `GetRunningInstanceByTags`, and — because the EC2 `DescribeInstances` API [follows an eventual consistency model](https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_DescribeInstances.html) — may not find the instance yet, creating a duplicate.

This adds an explicit `PatchObject()` call immediately after `createInstance` returns, following the same pattern already used for the finalizer and condition patches earlier in `reconcileNormal`.

Fixes #6131

The deferred `PatchObject` can fail because `reconcileNormal` already calls `PatchObject` explicitly twice before instance creation (for the finalizer and for the `InstanceProvisionStarted` condition). Each call bumps the `ResourceVersion` on the API server. Any external modification to the AWSMachine between the last explicit `PatchObject` and the deferred `Close()` causes a conflict.

**AI Usage**:

Claude Code was used for codebase exploration, root cause analysis, reproduction, and drafting the fix.

**Checklist**:

- [X] squashed commits
- [ ] includes documentation
- [X] includes emoji in title
- [X] adds unit tests

```release-note
Fix duplicate EC2 instance creation caused by deferred ProviderID persistence combined with EC2 API eventual consistency
```

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -580,6 +580,11 @@ func (r *AWSMachineReconciler) reconcileNormal(ctx context.Context, machineScope
 			v1beta1conditions.MarkFalse(machineScope.AWSMachine, infrav1.InstanceReadyCondition, infrav1.InstanceProvisionFailedReason, clusterv1beta1.ConditionSeverityError, "%s", err.Error())
 			return ctrl.Result{}, err
 		}
+
+		if patchErr := machineScope.PatchObject(); patchErr != nil {
+			machineScope.Error(patchErr, "failed to patch providerID")
+			return ctrl.Result{}, patchErr
+		}
 	}
 
 	// BYO Public IPv4 Pool feature: allocates and associates an EIP to machine when PublicIP and
--- a/controllers/awsmachine_controller_test.go
+++ b/controllers/awsmachine_controller_test.go
@@ -174,6 +174,11 @@ func TestAWSMachineReconcilerIntegrationTests(t *testing.T) {
 			{infrav1.ELBAttachedCondition, corev1.ConditionTrue, "", ""},
 		})
 		g.Expect(ms.AWSMachine.Finalizers).Should(ContainElement(infrav1.MachineFinalizer))
+
+		persisted := &infrav1.AWSMachine{}
+		g.Expect(testEnv.Get(ctx, client.ObjectKeyFromObject(ms.AWSMachine), persisted)).To(Succeed())
+		g.Expect(persisted.Spec.ProviderID).ToNot(BeNil())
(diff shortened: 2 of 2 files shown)


damdo (commented):
/approve

PR #6141, opened 2026-07-22
🐛 Add ROSAOCMRoleConfig CRD to kustomization

### LINKED ISSUE #6140: ROSAOCMRoleConfig CRD not deployed when installing CAPA 2.12.1 with ROSA=true
/kind bug

  1. Created a kind cluster
  2. Ran `clusterctl init --infrastructure aws:v2.12.1` with `EXP_ROSA=true` and `EXP_MACHINE_POOL=true`
  3. The `capa-controller-manager` pod starts but logs errors every 10 seconds:
 ```
E0722 13:54:20.549002 1 kind.go:75] "if kind is a CRD, it should be installed before calling Start"
err="no matches for kind "ROSAOCMRoleConfig" in version "infrastructure.cluster.x-k8s.io/v1beta2""
logger="controller-runtime.source.Kind" kind="ROSAOCMRoleConfig.infrastructure.cluster.x-k8s.io"
```
  4. Attempting to create a `ROSAOCMRoleConfig` resource fails:
  ```
error: resource mapping not found for name: "test-ocm-role" namespace: "default" from "STDIN":
no matches for kind "ROSAOCMRoleConfig" in version "infrastructure.cluster.x-k8s.io/v1beta2" ensure CRDs 
are installed first
```

The root cause is that `config/crd/bases/infrastructure.cluster.x-k8s.io_rosaocmroleconfigs.yaml` exists and the `ROSAOCMRoleConfigReconciler` is registered in `main.go` under the ROSA feature gate, but the CRD is missing from `config/crd/kustomization.yaml`. Since `clusterctl` generates manifests via `kustomize build config default`, the CRD is not included in the release's `infrastructure-components.yaml`.

This was not caught because unit tests use `envtest` which loads CRDs directly from `config/crd/bases/`, and manual e2e tests also applied the CRD directly from that directory, both bypass the kustomization.

  The `ROSAOCMRoleConfig` CRD should be deployed alongside all other ROSA CRDs when installing CAPA with `ROSA=true`, and users should be able to create `ROSAOCMRoleConfig` resources.

N/A
  
**Environment:**

  - Cluster-api-provider-aws version: v2.12.1
  - Kubernetes version: v1.35.0 (kind)
  - OS: macOS (darwin/arm64)

andclt:
**What type of PR is this?**

  /kind bug

The `ROSAOCMRoleConfig` CRD file exists in `config/crd/bases/` and the controller is registered in `main.go` under the ROSA feature gate, but the CRD was never added to `config/crd/kustomization.yaml`. This means deployments via `clusterctl init` with `ROSA=true` (e.g. v2.12.1) don't install the CRD, and the controller logs errors every 10 seconds:

"if kind is a CRD, it should be installed before calling Start" kind="ROSAOCMRoleConfig.infrastructure.cluster.x-k8s.io"
So users cannot create `ROSAOCMRoleConfig` resources.

This was not caught because both unit tests (envtest) and manual e2e tests load CRDs directly from `config/crd/bases/`, bypassing the kustomization.

  Fixes #6140

  Reproduced on a kind cluster with `clusterctl init --infrastructure aws:v2.12.1` and `EXP_ROSA=true`.

  **AI Usage**:

  None

  **Checklist**:

  - [x] squashed commits
  - [ ] includes documentation
  - [x] includes emoji in title
  - [ ] adds unit tests

```release-note
 fix(ROSA): ROSAOCMRoleConfig CRD not being deployed when using clusterctl with ROSA=true
```

### DIFF
--- a/PROJECT
+++ b/PROJECT
@@ -64,3 +64,6 @@ resources:
 - group: infrastructure
   kind: ROSANetwork
   version: v1beta2
+- group: infrastructure
+  kind: ROSAOCMRoleConfig
+  version: v1beta2
--- a/config/crd/kustomization.yaml
+++ b/config/crd/kustomization.yaml
@@ -29,6 +29,7 @@ resources:
 - bases/infrastructure.cluster.x-k8s.io_rosaclusters.yaml
 - bases/infrastructure.cluster.x-k8s.io_rosamachinepools.yaml
 - bases/infrastructure.cluster.x-k8s.io_rosaroleconfigs.yaml
+- bases/infrastructure.cluster.x-k8s.io_rosaocmroleconfigs.yaml
 - bases/infrastructure.cluster.x-k8s.io_rosanetworks.yaml
 # +kubebuilder:scaffold:crdkustomizeresource
 


damdo (commented):
/approve
/lgtm

damdo:
@andclt @serngawy 
feel free to unhold when happy with the fix

andclt:
Thanks @damdo!
/unhold

PR #6142, opened 2026-07-22
🌱 Add test to verify all CRD bases are listed in kustomization

andclt:
**What type of PR is this?**
/kind cleanup

Adds a test that verifies every CRD file in `config/crd/bases/` is listed in `config/crd/kustomization.yaml`. This prevents a recurring class of bug where `controller-gen` generates a CRD but it's not wired into the kustomization, resulting in the CRD being absent from `clusterctl` deployments while the controller still tries to watch it.

This gap is not caught by existing tests because `envtest` loads CRDs directly from `config/crd/bases/`, bypassing the kustomization entirely.

  Ref #6140

This test would have caught the `ROSAOCMRoleConfig` bug (#6140) and the earlier `ROSACluster`/`ROSAControlPlane` omission fixed in `7744a4b76`.

```
bases/infrastructure.cluster.x-k8s.io_rosaocmroleconfigs.yaml is not listed in kustomization.yaml — it will be missing from clusterctl deployments
```

  **AI Usage**:

Claude as code assistent

  None

  **Checklist**:

  - [x] squashed commits
  - [ ] includes documentation
  - [x] includes emoji in title
  - [x] adds unit tests

```release-note
  NONE
```


damdo (commented):
/approve
/lgtm

PR #6157, opened 2026-07-24
🐛 fix: authenticate GitHub API calls in auto-publish-ami workflow

richardcase:
**What type of PR is this?**

/kind support

The 'Determine target K8s versions' step runs find-missing-ami without GITHUB_TOKEN in its environment, so the kubernetes/kubernetes tag listing is unauthenticated and hits the shared per-IP rate limit on hosted runners, failing the workflow with a 403. The inline comment claiming the token is injected automatically by the Actions runner was incorrect; it is only available via the secrets context.

Set GITHUB_TOKEN from secrets on that step and on the matrix step, and make find-ami-publishing-inputs.sh send an Authorization header for its crictl release lookup when the token is present, avoiding the same rate limit silently degrading crictl version detection.

Fixes #

**AI Usage**:

Claude Code with Fable helped investigate and make the change.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [ ] includes emoji in title
- [ ] adds unit tests

**Release note**:

```release-note
NONE
```

### DIFF
--- a/.github/workflows/auto-publish-ami.yml
+++ b/.github/workflows/auto-publish-ami.yml
@@ -60,6 +60,10 @@ jobs:
       - name: Determine target K8s versions
         id: versions
         env:
+          # Used by find-missing-ami to query the Kubernetes release API;
+          # unauthenticated requests share a per-IP rate limit across all
+          # jobs on the runner and get throttled with 403s.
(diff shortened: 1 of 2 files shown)


richardcase:
Its only a GHA change so:

/override pull-cluster-api-provider-aws-e2e-blocking

damdo (commented):
/approve
/lgtm

PR #6182, opened 2026-08-11
🐛 Rebuild AWS session per controller when principal credentials change

UgurTheG:
**What type of PR is this?**

/kind bug

`sessionForClusterWithRegion` decides whether it can reuse a cached AWS session by checking whether every principal provider was already present in `providerCache`:

```go
isChanged := false
for i, provider := range providers {
    ...
    cachedProvider, ok := providerCache.Load(providerHash)
    if ok {
        provider = cachedProvider.(identity.AWSPrincipalTypeProvider)
    } else {
        isChanged = true
        providerCache.Store(providerHash, provider)
    }
    ...
}

if !isChanged {
    if s, ok := sessionCache.Load(getSessionName(region, clusterScoper)); ok {
        entry := s.(*sessionCacheEntry)
        return entry.session, entry.serviceLimiters, nil
    }
}
```

`providerCache` is global, but `sessionCache` is keyed per controller:

```go
func getSessionName(region string, clusterScoper cloud.SessionMetadata) string {
	return fmt.Sprintf("%s-%s-%s-%s", region, clusterScoper.ControllerName(), clusterScoper.InfraClusterName(), clusterScoper.Namespace())
}
```

So when the credentials behind an identity change, the first controller to reconcile misses `providerCache`, rebuilds its session correctly, and repopulates `providerCache`. Every *other* controller then hits `providerCache`, leaves `isChanged` false, and early-returns its own `sessionCache` entry, which was built from the previous credentials. The `sessionCache.Delete()` recovery path further down is never reached, because the early return happens first, so the stale session is served until the process restarts.

This PR records the provider hashes a session was built from in `sessionCacheEntry` and reuses the cached session only when those hashes still match. The decision becomes local to each controller's own cache entry instead of depending on a global cache that another controller may have already populated.

We hit this in production: a cluster was deleted and re-provisioned reusing the same name and namespace, with freshly minted static credentials (the previous IAM access key was removed). `AWSCluster` reconciled first and came up fully healthy (VPC, subnets, NAT gateways, LB all ready) while `AWSMachine` kept failing for ~10 hours with:

```
failed to query AWSMachine instance by tags: ... operation error EC2: DescribeInstances,
get identity: get credentials: failed to refresh cached credentials,
operation error STS: AssumeRole, https response error StatusCode: 403,
api error InvalidClientTokenId: The security token included in the request is invalid.
```

Same identity, same cluster, different controller: the machine controller was pinned to the pre-rotation session. Restarting `capa-controller-manager` was the only remedy, and the problem returned on the next re-provisioning.

None filed. Happy to open one first if maintainers prefer that.

All tests pass on this branch (`go test ./pkg/cloud/scope/...` is green).

To confirm the new test is a genuine regression test rather than one that would pass either way, I reverted just the `session.go` change and re-ran it. With the fix reverted it fails as expected, because the second controller keeps serving the pre-rotation access key:

```
--- FAIL: TestSessionRebuiltPerControllerAfterCredentialRotation
    session_cache_test.go:160:
        Expected
            <string>: AKIAIOSFODNN7EXAMPLE
        to equal
            <string>: AKIAI44QH8DHBEXAMPLE
```

With the fix in place it passes. The test drives `sessionForClusterWithRegion` through two `cloud.SessionMetadata` stubs differing only in `ControllerName()`, rotates the secret behind an `AWSClusterStaticIdentity`, and asserts both controllers observe the new access key. It uses a static identity so no STS or network calls are involved.

This PR closes two distinct paths to the same symptom, because fixing only the first leaves the second reachable:
1. **Cross-controller session reuse** (`session.go`). Covered above.
2. **Role provider hash did not cover the source credentials** (`identity.go`). `AWSRolePrincipalTypeProvider.Hash()` gob-encodes the provider, and gob only encodes exported fields. `sourceProvider` is unexported, so the hash was determined solely by the `AWSClusterRoleIdentity` object. Rotating the source secret while leaving that object untouched produced an identical hash, `providerCache` returned the provider built from the previous credentials, and every caller kept using them. The source provider hash is now mixed in.
Both are needed in practice. Our provisioning applies the identity objects and the credentials secret with create-or-update semantics, so a full teardown and recreate goes through path 1 (new object UIDs), while an in-place credential refresh goes through path 2 (object unchanged, secret rotated). Each has its own test that fails without the corresponding change.
Also worth noting separately: neither `sessionCache` nor `providerCache` is ever evicted, and `Hash()` computes `sha256.New()` then `hash.Sum(gobBytes)`, which returns `gobBytes || sha256("")` rather than a digest of the input, so cache keys are full gob payloads that accumulate per identity revision.

**AI Usage**:

This PR was produced with AI assistance: GitHub Copilot in agent mode (Claude Sonnet 4.5) via the JetBrains IDE integration. The AI performed the production incident investigation, read the CAPA source to identify the cache interaction, wrote the fix and the regression test, and drafted this description. All of it was verified against a real cluster and by running the test against both patched and unpatched code. Reviewed by me before submission.

**Checklist**:

- [ ] squashed commits (two commits, one per defect; happy to squash on request)
- [x] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
Fix AWS credentials going stale after an identity's credentials are rotated. Session reuse was gated on the global provider cache, so once one controller repopulated it every other controller kept serving a session built from the previous credentials. Separately, a role principal's cache hash did not cover the source credentials used to assume the role, so rotating them left the cached provider in place. Both required a manager restart to recover.
```

### DIFF
--- a/pkg/cloud/identity/identity.go
+++ b/pkg/cloud/identity/identity.go
@@ -152,6 +152,17 @@ func (p *AWSRolePrincipalTypeProvider) Hash() (string, error) {
 	if err != nil {
 		return "", err
 	}
+	// gob only encodes exported fields, so the source provider holding the
+	// credentials used to assume this role is not covered by the encoding above.
+	// Without mixing it in, rotating those credentials leaves the hash unchanged
+	// and a cached provider built from the previous credentials is reused.
+	if p.sourceProvider != nil {
+		sourceHash, err := p.sourceProvider.Hash()
+		if err != nil {
+			return "", err
+		}
+		roleIdentityValue.WriteString(sourceHash)
+	}
 	hash := sha256.New()
 	return string(hash.Sum(roleIdentityValue.Bytes())), nil
 }
--- a/pkg/cloud/identity/identity_hash_test.go
+++ b/pkg/cloud/identity/identity_hash_test.go
@@ -44,7 +44,7 @@ func TestRolePrincipalHashCoversSourceCredentials(t *testing.T) {
 		sourceProvider := NewAWSStaticPrincipalTypeProvider(
 			&infrav1.AWSClusterStaticIdentity{
 				ObjectMeta: metav1.ObjectMeta{Name: "source-identity"},
-				Spec:       infrav1.AWSClusterStaticIdentitySpec{SecretRef: "source-credentials"},
+				Spec:       infrav1.AWSClusterStaticIdentitySpec{SecretRef: "source-secret-ref"},
 			},
 			&corev1.Secret{
 				Data: map[string][]byte{
--- a/pkg/cloud/identity/identity_hash_test.go
+++ b/pkg/cloud/identity/identity_hash_test.go
@@ -1,0 +1,87 @@
+/*
+Copyright 2026 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package identity
+
+import (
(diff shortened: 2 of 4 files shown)


UgurTheG:
anything missing here? Or OK to merge?

PR #6188, opened 2026-08-19
📖 Document EKS cross-account role assumption with IRSA

lmilbaum:
**What type of PR is this?**

/kind documentation

**What this PR does / why we need it**:

- Existing multitenancy documentation was outdated and difficult to discover.
- This adds an EKS guide for IRSA-based cross-account role assumption.
- It documents the IAM trust chain, CAPA identities, and AWSManagedControlPlane.identityRef.

**Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:
Fixes #2105 

**Special notes for your reviewer**:
In response to the review feedback asking for validation of the documented workflow, I added a local test that validates the parts of the documentation we can verify without access to AWS accounts.

The test validates that:

* The documented CAPA configuration is consistent with the current CAPA API types.
* The referenced fields and identity configuration are valid and represented in the CAPA manifests.
* The management-account and workload-account configuration described in the guide remains internally consistent with the implementation.
* Changes to CAPA APIs or manifests that make the documented configuration stale or invalid can be detected locally.

The test is intentionally self-contained and requires **no AWS credentials or AWS accounts**, so it can run locally and in CI without provisioning cloud resources.

There is an important limitation: this test validates the **documentation contract against the CAPA codebase**, not the actual AWS behavior. It does **not** prove that cross-account IAM role assumption or IRSA works end to end between two real AWS accounts.

Full validation of the cross-account IAM/IRSA flow still requires testing the documented procedure against real AWS infrastructure.

**AI Usage**:

Codex

**Checklist**: 

- [x] squashed commits
- [x] includes documentation
- [x] includes AI generated content
- [x] includes emoji in title
- [x] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
Added EKS documentation for cross-account role assumption using IRSA. There is no user action required. And the link to the new guild is - docs/book/src/topics/eks/cross-account-role-assumption.md
```

PR #6194, opened 2026-08-26
🌱 Extend management cluster provisioning for AWS-hosted testing

lmilbaum:
**What type of PR is this?**

/kind feature

This PR explores extending the existing e2e environment provisioning flow so CAPA tests can run against a self-managed CAPA management cluster in AWS. The AWS cluster is provisioned with the same CAPA templates and infrastructure used for workload clusters, rather than through a separate bootstrap implementation.

The proposed workflow has three independently invoked stages:

1. `make provision-self-hosted-management-cluster`
   - creates the normal temporary kind management cluster;
   - provisions an AWS workload cluster with the existing `remote-management-cluster` flavor;
   - installs the providers under test on the AWS cluster;
   - moves the Cluster API objects from kind to AWS, making the AWS cluster self-managed;
   - retains both clusters for the following stages.
2. `make test-self-hosted-management-cluster`
   - runs selected e2e specs against the retained AWS management cluster without provisioning or deleting the management environment.
3. `make teardown-self-hosted-management-cluster`
   - moves the management-cluster objects back to kind;
   - deletes the AWS cluster;
   - deletes the temporary kind bootstrap cluster.

The normal e2e workflow remains unchanged and continues to use kind as its management cluster.

None.

This is currently a prototype for discussing the design direction, not a finished implementation ready for merge.

Separating provisioning, test execution, and teardown exposed that the existing e2e environment lifecycle is closely coupled to Ginkgo's synchronized suite setup and cleanup hooks. Supporting stages that can run as separate processes may require a broader refactor of the shared e2e framework than originally expected.

Guidance would be appreciated on whether:

- the existing e2e framework should be refactored to make these three stages independently reusable;
- the lifecycle should remain inside the Ginkgo suite hooks; or
- dedicated lifecycle commands around the existing e2e helpers would fit the project better.

The current prototype includes the shared lifecycle implementation, Make targets, focused unit tests, suite integration, and Developer Guide documentation. Focused tests, e2e package compilation, lint for the changed shared package, Make dry runs, and diff checks pass. A real AWS end-to-end lifecycle has not been run.

**AI Usage**:

Codex was used to assist with this contribution and the PR description.

**Checklist**:

- [x] squashed commits
- [x] includes documentation
- [x] includes emoji in title
- [x] adds unit tests
- [x] adds or updates e2e framework code

**Release note**:

```release-note
NONE
```


AndiDog:
This is about bootstrapping a management cluster for testing purposes, but it only adds Make targets for manual use, if I understand correctly. I honestly don't see us able to maintain that, given that each contributing company usually has the knowledge and infra to bootstrap MCs in some "official" way (not that there's one official way, hence your PR 😉), that is, we would not run the script or ever fix problems in it. I'd like to know: which problem does this PR solve? Was the documentation of CAPI/CAPA not good enough to get going with a fresh MC that runs these projects? Would you have expected a command built into `clusterawsadm` that helps create and tear down an MC?

raykrueger:
I personally understand this PR. Right now creating a management cluster on AWS requires a bit of jumping around between docs and tools. Having a process for creating an EKS based management cluster in the AWS repo here makes sense to me.

However, I'm not sure the Makefile is the right place. Make is primarily there for development purposes, not operational tasks. Perhaps this should live in docs?

lmilbaum:
@AndiDog 

> This is about bootstrapping a management cluster for testing purposes, but it only adds Make targets for manual use, if I understand correctly. I honestly don't see us able to maintain that, given that each contributing company usually has the knowledge and infra to bootstrap MCs in some "official" way (not that there's one official way, hence your PR 😉), that is, we would not run the script or ever fix problems in it. I'd like to know: which problem does this PR solve? Was the documentation of CAPI/CAPA not good enough to get going with a fresh MC that runs these projects? Would you have expected a command built into `clusterawsadm` that helps create and tear down an MC?

Thanks, this makes sense, and I agree with the maintainability concern.

Some additional context on why I opened this PR: it originated from the documentation PR I am working on: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6188. I want to validate the documented flow against a real AWS environment. The scenario documented there requires the management cluster to run in one AWS account while the workload cluster is provisioned into a different AWS account.

My immediate need is therefore not to introduce another officially supported way of provisioning management clusters. It is to have a reproducible way to create an AWS-hosted management cluster so I can validate that cross-account workflow end to end.

After looking further into the existing CAPA e2e machinery, I think the better direction is to reuse the existing self-hosted e2e provisioning path rather than introduce independently maintained provisioning scripts.

The workflow would start with kind, initialize CAPI/CAPA, provision the AWS management cluster, initialize CAPI/CAPA there, move the management objects to it, and then remove kind. This keeps the provisioning logic aligned with something CAPA already exercises and maintains.

As far as I understand, the current e2e infrastructure assumes a single AWS account, so the complete cross-account scenario cannot currently be automated through the existing tests. Longer term, it would be interesting to see whether the e2e framework could support multiple AWS accounts, or whether the workload account side could somehow be virtualized/emulated so that this scenario could become part of automated testing.

For now, my goal is much narrower: enable manual validation of the documentation against the real cross-account setup without introducing a separate provisioning implementation that CAPA would need to maintain.

lmilbaum:
@raykrueger 

> I personally understand this PR. Right now creating a management cluster on AWS requires a bit of jumping around between docs and tools. Having a process for creating an EKS based management cluster in the AWS repo here makes sense to me.
> 
> However, I'm not sure the Makefile is the right place. Make is primarily there for development purposes, not operational tasks. Perhaps this should live in docs?

Thanks, Ray. I agree that the Makefile should not contain the operational lifecycle itself.

Based on your feedback and AndiDog's, I am planning to refactor the PR to reuse CAPA's existing self-hosted e2e provisioning machinery rather than maintain a separate shell-based implementation.

The existing e2e flow already has most of the lifecycle I need: create a kind bootstrap cluster, provision an AWS cluster, initialize CAPI/CAPA there, move the management objects to it, and remove kind.

The use case and the manual workflow can then be documented, including the lifecycle, AWS costs, and teardown behavior. If a Make target is useful at all, I would keep it as a thin entry point rather than putting the implementation there.

This should also leave the existing `make test-e2e` behavior unchanged.

PR #6198, opened 2026-08-31
🐛 Fix wrong use of object kind in update predicates, leading to reconciliation when it could be skipped

AndiDog:
**What type of PR is this?**

/kind bug

The UPDATE request predicate didn't work as expected because the object kind might be empty. Use type check instead. This brings down reconciliation rate a little.

n/a

**AI Usage**:

This was indeed a side effect finding from troubleshooting an unrelated issue using Claude Code. I manually deployed and tested this fix, ensuring the kind value is really empty, and also confirmed a reduction in a `workqueue_*` metric with this fix.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [ ] adds unit tests

**Release note**:

```release-note
fix: wrong use of object kind in update predicates leads to unnecessary reconciliation
```

(we could list the affected objects in the release note if you want – feel free to edit 😉)

### DIFF
--- a/controllers/awsmachine_controller.go
+++ b/controllers/awsmachine_controller.go
@@ -279,18 +279,29 @@ func (r *AWSMachineReconciler) SetupWithManager(ctx context.Context, mgr ctrl.Ma
 				// Avoid reconciling if the event triggering the reconciliation is related to incremental status updates
 				// for AWSMachine resources only
 				UpdateFunc: func(e event.UpdateEvent) bool {
-					if e.ObjectOld.GetObjectKind().GroupVersionKind().Kind != kindAWSMachine {
+					// Objects in watch events have no `TypeMeta`, so the kind can only be told from the Go type
+					oldAWSMachine, ok := e.ObjectOld.(*infrav1.AWSMachine)
+					if !ok {
+						return true
+					}
+					newAWSMachine, ok := e.ObjectNew.(*infrav1.AWSMachine)
+					if !ok {
 						return true
 					}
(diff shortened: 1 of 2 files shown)


damdo (commented):
Thanks!

/approve
/lgtm

damdo:
@serngawy @andclt five other controllers have the same latent bug in the ROSA area.
None have the GroupVersionKind().Kind bug (they already use safe type assertions), but all five zero Status + ResourceVersion and then cmp.Equal without nulling ManagedFields:

```
│                               File                               │       Type        │
│ controlplane/rosa/controllers/rosacontrolplane_controller.go:142 │ ROSAControlPlane  │
│ exp/controllers/rosamachinepool_controller.go:89                 │ ROSAMachinePool   │
│ exp/controllers/rosanetwork_controller.go:352                    │ ROSANetwork       │
│ exp/controllers/rosaocmroleconfig_controller.go:91               │ ROSAOCMRoleConfig │
│ exp/controllers/rosaroleconfig_controller.go:104                 │ ROSARoleConfig    │
```

The fix each needs (mirroring awsmachine_controller.go:301-304):
```
// A status write refreshes the timestamp of the writer's `managedFields` entry, so the metadata
// would otherwise always differ.
oldXX.ObjectMeta.ManagedFields = nil
newXX.ObjectMeta.ManagedFields = nil
```

@serngawy @andclt Could you create a similar PR to fix them? Thanks!

PR #6220, opened 2026-09-08
🌱  chore: add remove-duplicates command to release-tool for AMI cleanup

richardcase:
**What type of PR is this?**

/kind feature

CAPA's AMI build pipeline can leave stale duplicate AMIs behind in a region: same `distribution`/`distribution_version`/`kubernetes_version` tags, different `build_timestamp`. There was no way to find or clean these up.

This PR adds a `release-tool ami remove-duplicates` command (in `hack/tools/release-tools`) that:
- Only considers AMIs tagged with `kubernetes_version` (every CAPA AMI build sets one), so an unrelated image in the same AWS account is never in scope.
- Groups those AMIs per region by their `distribution`, `distribution_version`, and `kubernetes_version` tags, and keeps only the one with the highest `build_timestamp` in each group.
- Flags AMIs that can't be grouped, missing tags or an invalid `build_timestamp`, as ungroupable instead of quietly skipping them.
- Runs as a dry run by default. Pass `--delete` to actually remove AMIs; each one is deregistered with `DeleteAssociatedSnapshots: true`, so its backing EBS snapshots go with it.
- Only targets `DUPLICATE` AMIs unless you pass `--include-ungroupable`. Kept AMIs are never removed.

Also included is `.github/workflows/ami-housekeeping-remove-duplicates.yml`, a `workflow_dispatch` workflow (matching the existing `ami-housekeeping-*` workflows) with `region`, `include_ungroupable`, and `delete` inputs, all defaulted, so it can be run from the Actions UI without local AWS credentials.

Fixes #

This has been run against the CAPA AWS account used for AMIs and duplicates have been removed.

This PR went through a review round and the following came out of it:
- Discovery was scoped down to only AMIs tagged `kubernetes_version`. Before this, `--delete --include-ungroupable` could remove an unrelated AMI in the same account, since anything missing the CAPA tags was classified `UNGROUPABLE` rather than excluded outright.
- A group with a single AMI (no duplicates) used to be dropped from the report entirely instead of showing up as `KEEP`. Fixed.
- The AWS-facing orchestration (target selection, deletion, per-region client reuse, partial-failure handling, table/JSON/YAML output) now has unit test coverage via an injectable EC2 interface, on top of the existing grouping-logic tests.
- Replaced a delimiter-joined string used as a map key with a proper struct, matching the `amiLookupKey` pattern already used elsewhere in this package.
- Removed the duplicated 15-region default from the workflow input, the CLI's own default is now the only copy.
- Documented the `x86_64`/`available`/`hvm` discovery filters (they match CAPA's own AMI build properties) and fixed a stale workflow filename in the README.

**AI Usage**:

Built with Claude Code with Fable 5. I directed the design (the dry-run/`--delete` safety model, command behavior) and reviewed it throughout, and verified each step with build/vet/test.

**Checklist**:

- [X] squashed commits
- [X] includes documentation
- [ ] includes emoji in title
- [X] adds unit tests

**Release note**:

```release-note
NONE
```


richardcase:
Whilst i look at some potential issues

/hold

PR #6231, opened 2026-09-10
🐛 fix: apply defaults in NodePoolToRosaMachinePoolSpec to prevent phantom diff

andclt:
**What type of PR is this?**

/kind bug

`NodePoolToRosaMachinePoolSpec` only set `NodeDrainGracePeriod` and `UpdateConfig` when OCM explicitly returned those fields. Since OCM uses bitmap-based field presence, default values are often absent from the JSON response. Meanwhile, `Default()` always sets these fields, so `computeSpecDiff` detected a phantom diff on every reconciliation, causing NO-OP PATCH requests to the OCM API.

This PR sets defaults unconditionally in the conversion and overrides with OCM values when present, matching what `Default()` produces.

Fixes #

The conversion output is used in two places:
1. `computeSpecDiff` — comparison only, never used to build PATCH payloads
2. `buildROSAMachinePool` — creates a K8s object that goes through webhook defaulting, which would set the same values anyway

The PATCH payload is always built from the desired spec (user's CR + `Default()`), so this change cannot alter what gets sent to OCM.

**AI Usage**:

Claude Code (claude-opus-4-6) was used implementation and test writing.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes emoji in title
- [x] adds unit tests

**Release note**:

```release-note
fix(rosa): phantom spec diffs in ROSAMachinePool reconciliation caused by NodePoolToRosaMachinePoolSpec not applying defaults when OCM omits fields from the response
```

### DIFF
--- a/exp/utils/rosa_helper.go
+++ b/exp/utils/rosa_helper.go
@@ -68,15 +68,18 @@ func NodePoolToRosaMachinePoolSpec(nodePool *cmv1.NodePool) expinfrav1.RosaMachi
 		}
 		spec.Taints = rosaTaints
 	}
+	spec.NodeDrainGracePeriod = &metav1.Duration{}
 	if nodePool.NodeDrainGracePeriod() != nil {
-		spec.NodeDrainGracePeriod = &metav1.Duration{
-			Duration: time.Minute * time.Duration(nodePool.NodeDrainGracePeriod().Value()),
-		}
+		spec.NodeDrainGracePeriod.Duration = time.Minute * time.Duration(nodePool.NodeDrainGracePeriod().Value())
+	}
+
+	spec.UpdateConfig = &expinfrav1.RosaUpdateConfig{
+		RollingUpdate: &expinfrav1.RollingUpdate{
+			MaxUnavailable: ptr.To(intstr.FromInt32(0)),
+			MaxSurge:       ptr.To(intstr.FromInt32(1)),
+		},
 	}
 	if nodePool.ManagementUpgrade() != nil {
-		spec.UpdateConfig = &expinfrav1.RosaUpdateConfig{
-			RollingUpdate: &expinfrav1.RollingUpdate{},
-		}
 		if nodePool.ManagementUpgrade().MaxSurge() != "" {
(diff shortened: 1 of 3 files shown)


exp/utils/rosa_helper.go
```diff
@@ -68,15 +68,18 @@ func NodePoolToRosaMachinePoolSpec(nodePool *cmv1.NodePool) expinfrav1.RosaMachi
...
 	}
+	spec.NodeDrainGracePeriod = &metav1.Duration{}
 	if nodePool.NodeDrainGracePeriod() != nil {
-		spec.NodeDrainGracePeriod = &metav1.Duration{
-			Duration: time.Minute * time.Duration(nodePool.NodeDrainGracePeriod().Value()),
-		}
+		spec.NodeDrainGracePeriod.Duration = time.Minute * time.Duration(nodePool.NodeDrainGracePeriod().Value())
+	}
+
+	spec.UpdateConfig = &expinfrav1.RosaUpdateConfig{
```
serngawy, line 76: We already calling default [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosamachinepool_controller.go#L428) before calling the `NodePoolToRosaMachinePoolSpec` func so these values should be there, right ?

andclt: `Default()` there (on line 428) runs on the desired spec (the user's CR), not on the OCM conversion output. `NodePoolToRosaMachinePoolSpec` produces the current spec independently from the OCM response:

```
machinePool.Default()                                      // ← defaults desiredSpec
desiredSpec := machinePool.Spec


currentSpec := utils.NodePoolToRosaMachinePoolSpec(nodePool) // ← from OCM, no Default()
```

So `desiredSpec` has `NodeDrainGracePeriod = &Duration{0s}` and `UpdateConfig = {MaxSurge:1, MaxUnavailable:0}`, but `currentSpec` has them as nil when OCM omits those fields from the response. That's the phantom diff this fixes.

serngawy: okay, good

serngawy (approved):
/lgtm

damdo (commented):
/approve
/lgtm

PR #6235, opened 2026-09-11
🐛 Fix ROSAMachinePool version range underflow on 5.0/5.1 control planes

### LINKED ISSUE #6234: All ROSAMachinePools stuck in Provisioning on a 5.0/5.1 control plane (uint64 underflow)
/kind bug

 On an OpenShift 5.0 or 5.1 control plane, ROSA HCP cluster provisioning never completes. Every ROSAMachinePool is rejected by an internal version check and stays in Provisioning indefinitely — including the pools CAPA imports itself — even though the control plane is healthy and the worker nodes are up and running at the provider. Nothing is logged and no status is written, so there is no indication of why.

  The cause is an integer underflow: a uint64 minor version minus 2, which wraps instead of clamping when the minor version is 0 or 1. It is deterministic on every X.0 and X.1 release and affects nothing else.

 MachinePoolSupportedVersionsRange in pkg/rosa/versions.go computes the lowest allowed machine pool version by subtracting 2 from the control plane's minor version:

  minVersion := semver.Version{
      Major: maxVersion.Major,
      Minor: max(0, maxVersion.Minor-minorVersionsAllowedDeviation),
      Patch: 0,
  }

  semver.Version.Minor is a uint64, so when the minor version is 0 or 1 the subtraction wraps around instead of clamping. max(0, ...) can't help — the number is already huge by the time it runs.

  control plane 4.22.9  -> minSupported = 4.20.0                    ok
  control plane 5.0.0   -> minSupported = 5.18446744073709551614.0  broken
  control plane 5.1.0   -> minSupported = 5.18446744073709551615.0  broken
  control plane 5.2.0   -> minSupported = 5.0.0                     ok

  The minimum ends up higher than the control plane's own version, so the range is empty and validateMachinePoolSpec rejects every machine pool — including one requesting the same version
  as its control plane.

  It also fails silently. reconcileNormal returns on the failureMessage without setting Status.FailureMessage, without a condition, without a log line, and without requeueing. The ROSAMachinePool keeps an empty status and core CAPI waits on it forever with "Infrastructure provider is not yet ready", so the MachinePool never leaves Provisioning even though the node pool is healthy at the provider.

  A machine pool requesting the same version as its control plane should always be accepted. For a control plane at X.0 or X.1 the floor should clamp to X.0 rather than underflowing.

 Secondary issue (same code path): semver sorts a prerelease below its own release, so 5.0.0-rc.0 < 5.0.0. Even with the underflow fixed, a prerelease control plane falls outside a range
  derived from itself. Since the skew policy is expressed purely in minor versions, the bounds and the candidate need to be compared as core major.minor.patch.

  Why this hasn't been hit before: 
The code was introduced in 238b63b6a (2024-02-05) and every OpenShift release for the last decade has had a minor version of at least 2, so the subtraction always worked. This is reachable for the first time on the 5.x line, and only for minor 0 and 1 — 5.2 onward is unaffected. There is currently no pkg/rosa/versions_test.go,  which is likely why it went unnoticed.

**Environment:**

- Cluster-api-provider-aws version:  reproduced on `main` (81730dae4) and release `v2.13.0`

tinaafitz:
**What type of PR is this?**
/kind bug

Fixing a bug in the CAPA controller that blocks worker-node management on any ROSA HCP cluster running OpenShift 5.0 or 5.1. A version check subtracts two from the cluster's minor version using an unsigned integer, which wraps around instead of clamping at zero — so CAPA rejects every machine pool, including the ones it created itself, and they hang in Provisioning forever with nothing logged.

 - Every machine pool on 5.0/5.1 is rejected, including ones matching the control plane version and ones CAPA imports itself. No configuration avoids it.
 - It fails silently — no status, condition, log or requeue. The only symptom is a 30-minute "did not become ready" timeout, which is expensive to diagnose and easy to misattribute (we initially logged it as slow 5.0 readiness).
 - It blocks 5.0 release validation — no machine-pool suite can pass, so that whole functional area is untestable and any other defect in the path is masked.
 - It does not resolve itself — still present in upstream main and v2.13.0, latent since 238b63b6a (2024-02-05), and reachable for the first time on 5.0.
 - Control planes on minor 2 or greater are unaffected — every 4.x version produces a byte-for-byte identical range to today's, and the new tests pin that genuinely unsupported versions are still rejected.

Fixes #6234

Three things worth a close look.

The failure is reported through a condition, not Status.FailureMessage. Writing that field panics CAPI's MachinePool controller. reconcileExternal reads status.failureMessage off the infra object and then dereferences a Status.Deprecated it never initialized — the nil guards there test != nil where the surrounding code uses == nil, so the lazy allocation is skipped exactly when it is needed. RecoverPanic turns that into an endless requeue rather than a crash, so the owning MachinePool silently stops updating nodeRefs and replicas until the spec is corrected, with the stack trace landing in capi-controller-manager's logs rather than CAPA's. Present since cluster-api v1.11.0 (introduced in 2cfe66bae, which got the same migration right a few lines further down) and still on main; I have a fix and regression tests to send upstream separately. The condition and log line carry the same detail without that side effect. 

**The cross-major window is deliberately permissive.** Clamping the subtraction alone fixes the underflow but floors a 5.0 control plane at 5.0.0, which then rejects the 4.22 machine pools a 5.0 control plane is expected to run — an over-strict range in place of a broken one. The exact floor for a cross-major window depends on the last minor of the previous major (4.22 today), which is not derivable from the control plane version and is not exposed anywhere on the `OCMClient` interface. Rather than hardcode that, the cross-major case admits the whole previous major series down to `MinSupportedVersion` (4.14.0). Red Hat documents the skew rule only in terms of minor y-stream versions and publishes nothing about major-boundary behaviour, so a tighter bound would be encoding an inference. Happy to tighten this if someone can point at a source of truth for the boundary.

The principled fix is to enumerate available versions from OCM and clip that list positionally, the way `rosa list versions --hosted-cp --channel-group <cg>` does, instead of computing a bound arithmetically. That needs a new `OCMClient` method and a plumbing change through `validateMachinePoolSpec`, so I've left it for a follow-up issue rather than grow this one at RC time. Worth noting that upstream `openshift/rosa` has the same arithmetic bug in `GetMinimalHostedMachinePoolVersion` (unclamped `segments[1] - 2`, producing an unparsable `"5.-2.0"`), and `GetFilteredVersionList` then fails *open* on the bad bound.

**This range check is UX, not a safety gate.** `ValidateHypershiftVersion(versionRawID, channelGroup)` takes no cluster and no control-plane version, so it structurally cannot enforce skew — it only confirms a version exists in a channel group and supports HCP. Whether the OCM *server* rejects an out-of-skew node pool create is not something I was able to confirm; a live test failed earlier in the pipeline on AWS AssumeRole and never reached version validation. I'd rather the local check be loose and let the provider be authoritative than have it reject valid configurations.

No webhook change here on purpose. A range check in `exp/webhooks/rosamachinepool_webhook.go` would need the control plane version, which isn't resolvable at admission — the owner MachinePool ref isn't set on create, which is exactly why the controller has its own `"MachinePool Controller has not yet set OwnerRef"` early return. `ROSAControlPlane`'s webhook omits its version validation for the same reason: both ROSA webhooks validate syntax, both controllers validate anything needing cluster context.

**AI Usage**:

Claude Code (Claude Opus 5) was used for implementation and test writing. I reviewed every change and verified it independently: the underflow reproduces in a standalone Go program with no cluster, the resulting failure mode was confirmed on a live 5.0 management cluster running a released image, the CAPI panic described above reproduces against an unmodified checkout of cluster-api main, the new reconcile test fails if the `Status.FailureMessage` write is restored, and `make generate` (no drift), `make lint` (0 issues) and `make test` all pass locally.

**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [X] includes emoji in title
- [X] adds unit tests

**Release note**:

```release-note
fix(ROSA): a uint64 underflow in `MachinePoolSupportedVersionsRange` that caused every ROSAMachinePool to be rejected on control planes with a minor version of 0 or 1, leaving MachinePools stuck in `Provisioning` with no status, condition or log line. Control planes whose skew window crosses a major boundary (5.0, 5.1) now admit machine pools from the previous major series. Rejected machine pool specs now surface the reason through the Ready condition and   propagate the error to the controller logger instead of failing silently.
```


tinaafitz:
@serngawy, please review.

serngawy (commented):
Can we add test case for v4.23/22 and v5.0/1 , ex; controlPlan at v5.1 and upgrade nodePool from v 4.22 to v4.23

richardcase:
Ping me when you need approval (and there is a lgtm)

tinaafitz:
Hi @serngawy, 

Changes made. Please review.

tinaafitz:
The test fails the same way when running in my local Minikube environment with CAPA `v2.13.0`. 

Here's the full evidence:

*Versions:*
```
clusterctl:                   v1.13.4
CAPA (infrastructure-aws):    v2.13.0
CAPI (cluster-api core):      v1.14.2
OCP / RosaControlPlane:       5.0.0-rc.0
ROSAMachinePool spec version:  5.0.0-rc.0
```

*The bug line:*
https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/b22c01d334bb0bf567dda43502cb8d891332fec5/pkg/rosa/versions.go#L116

```go
Minor: max(0, maxVersion.Minor-minorVersionsAllowedDeviation),
```

`Minor` is `uint64`. On a `5.0.0-rc.0` control plane, `Minor = 0`. `0 - 2` wraps to `18446744073709551614` *before* `max()` is called — so `max(0, enormous_number)` returns the enormous number, not 0.

*Test output (run against upstream code, no modifications):*
```
=== RUN   TestBugReproduction/5.0_release_candidate,_pool_at_the_same_version
[upstream] control plane minor=0  raw subtraction 0 - 2 = 18446744073709551614 (uint64)
    underflow_test.go:161: machine pool 5.0.0-rc.0 REJECTED on control plane 5.0.0-rc.0
            computed supported range: >= 5.18446744073709551614.0 and <= 5.0.0-rc.0
            a machine pool matching its own control plane must always be accepted
--- FAIL: TestBugReproduction (0.00s)
    --- FAIL: TestBugReproduction/5.0_release_candidate,_pool_at_the_same_version (0.00s)
```

*Side-by-side across all 5.x versions:*
```
control plane | upstream minimum             | upstream   | with fix
5.0.0-rc.0   | 5.18446744073709551614.0     | REJECTED   | accepted
5.0.0        | 5.18446744073709551614.0     | REJECTED   | accepted
5.1.0        | 5.18446744073709551615.0     | REJECTED   | accepted
5.2.0        | 5.0.0                        | accepted   | accepted
```

*Call stack with live values for `"5.0.0-rc.0"`:*
```
rosamachinepool_controller.go:92   Reconcile(req={ns-rosa-hcp/mi5-rosa-hcp})
rosamachinepool_controller.go:199    reconcileNormal(...)
rosamachinepool_controller.go:223      validateMachinePoolSpec(machinePoolScope)
rosamachinepool_controller.go:456        rosa.MachinePoolSupportedVersionsRange("5.0.0-rc.0")
pkg/rosa/versions.go:109                   semver.Parse("5.0.0-rc.0")
                                             → maxVersion = {Major:5, Minor:0, Patch:0}
pkg/rosa/versions.go:116                   max(0, 0 - 2)
                                           = max(0, 18446744073709551614)  ← uint64 underflow
                                           = 18446744073709551614
                                             → minVersion = {Major:5, Minor:18446744073709551614, Patch:0}
rosamachinepool_controller.go:461        5.0.0-rc.0 < 5.18446744073709551614.0 → REJECTED
                                             → failureMessage set
rosamachinepool_controller.go:227        return ctrl.Result{}, nil  ← silent exit, no status written
```

*Why it's silent:* when `failureMessage != nil`, the controller returns `nil` with no requeue and never writes the failure to `ROSAMachinePool.status` — so the resource stays stuck on `WaitingForRosaControlPlane` indefinitely with no error surfaced.

*Live reproduction:* provisioned a ROSA HCP cluster with OCP `5.0.0-rc.0` on CAPA `v2.13.0`. Control plane reached `ready: true` at `17:41:11Z`. ROSAMachinePool stayed on `WaitingForRosaControlPlane` for 31+ minutes with no conditions, no errors, no requeue — while OCM console confirmed the cluster was fully running with workers:
```
ROSAMachinePool/mi5-rosa-hcp   False   WaitingForRosaControlPlane   54m   0 of 1 completed
```

exp/controllers/rosamachinepool_controller.go
```diff
@@ -250,6 +250,30 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 		return ctrl.Result{}, fmt.Errorf("failed to validate ROSAMachinePool.spec: %w", err)
 	}
 	if failureMessage != nil {
+		// Surface the reason. Without this the reconcile returns silently -- no
```
serngawy: good, also please fix the func `validateMachinePoolSpec` [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosamachinepool_controller.go#L480)  cause it does not make sense to return message without error . This why we end up don't see any error.

exp/controllers/rosamachinepool_controller.go
```diff
@@ -491,7 +515,11 @@ func validateMachinePoolSpec(machinePoolScope *scope.RosaMachinePoolScope) (*str
 		return nil, fmt.Errorf("failed to get supported machinePool versions range: %w", err)
 	}
 
-	if version.GT(*maxSupportedVersion) || version.LT(*minSupportedVersion) {
+	// The bounds are core versions, so the candidate has to be compared as one
```
serngawy: This func is really strange, please remove the return message and just return error

serngawy (changes requested):

exp/controllers/rosamachinepool_controller.go
tinaafitz: Hi @serngawy, Good catch, thanks. Fixed.

exp/controllers/rosamachinepool_controller.go
tinaafitz: Hi @serngawy, Thanks. Fixed.

pkg/rosa/versions.go
```diff
@@ -103,18 +103,46 @@ func ScheduleNodePoolUpgrade(client OCMClient, clusterID string, nodePool *cmv1.
...
+	parsed, err := semver.Parse(controlPlaneVersion)
 	if err != nil {
 		return nil, nil, err
 	}
-
-	minVersion := semver.Version{
-		Major: maxVersion.Major,
-		Minor: max(0, maxVersion.Minor-minorVersionsAllowedDeviation),
-		Patch: 0,
+	maxVersion := CoreVersion(parsed)
```
andclt: Is stripping the prerelease from the upper bound maybe a little too permissive here? With a `5.0.0-rc.0` control plane this makes the max `5.0.0`, so a `5.0.0` (or later prerelease such as `5.0.0-rc.1`) machine pool is considered in range.

[CAPA's upgrade docs](https://cluster-api-aws.sigs.k8s.io/topics/rosa/upgrades#machinepool-upgrade) say the machine pool version can't be greater than the control plane version, so would it make sense to preserve the original parsed control-plane version for the upper-bound comparison, and only use the core version when calculating the lower skew bound?

I believe that should still fix the RC lower-bound issue while retaining normal semver ordering for "machine pool must not be ahead of control plane".

pkg/rosa/versions.go
```diff
@@ -103,18 +103,46 @@ func ScheduleNodePoolUpgrade(client OCMClient, clusterID string, nodePool *cmv1.
...
+	// When the control plane minor is below the allowed deviation the skew window
+	// crosses a major boundary (a 5.0 or 5.1 control plane can still run 4.x
+	// pools). The exact floor there depends on the last minor of the previous
+	// major, which is not derivable from the control plane version and is not
+	// recorded anywhere in the OCM client. So admit the whole previous major
+	// series: deliberately permissive, because wrongly rejecting a valid pool is
+	// worse than admitting one the node-pool create then fails on.
+	var minVersion semver.Version
+	if maxVersion.Minor < minorVersionsAllowedDeviation && maxVersion.Major > 0 {
+		minVersion = semver.Version{Major: maxVersion.Major - 1, Minor: 0, Patch: 0}
```
andclt: I understand why we need a fallback here since the actual previous Y-streams aren't derivable from `5.0/5.1` alone, and I agree that hardcoding a guessed `4.x`. floor would be worse.

One thing I'm wondering about is whether representing the fallback as the whole previous major is too broad.

Would it be better to make the "exact lower bound is unknown across a major boundary" case explicit rather than returning a range that looks authoritative? Longer term, querying the available versions from OCM as mentioned in the PR seems like the cleanest solution.

I'm mostly concerned that callers/tests may start treating `4.14 -> 5.x` as intentionally supported rather than as a fail-open validation fallback.

exp/controllers/rosamachinepool_controller.go
andclt, line 400: Since the cross-major validation is intentionally permissive and relies on OCM to reject an unsupported version, should we surface this failure through `RosaMachinePoolReadyCondition` before returning the error?

At this point the existing NodePool has already been marked `Ready=True`, so if `ScheduleNodePoolUpgrade` rejects the requested version, users may only see the useful error in the controller logs while the `ROSAMAachinePool` still reports ready.

Maybe we should mark `Ready=False` with `RosaMachinePoolReconciliationFailedReason` here, similar to the create/update failure paths

exp/controllers/rosamachinepool_controller.go
```diff
@@ -245,12 +245,32 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
...
 	}
 
-	failureMessage, err := validateMachinePoolSpec(machinePoolScope)
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to validate ROSAMachinePool.spec: %w", err)
-	}
-	if failureMessage != nil {
-		// dont' requeue because input is invalid and manual intervention is needed.
+	if err := validateMachinePoolSpec(machinePoolScope); err != nil {
+		// Surface the reason through a condition. Without this the reconcile returns
```
serngawy: No need for long explanation. something simple like "Surface the error message to condition.

exp/controllers/rosamachinepool_controller.go
```diff
@@ -245,12 +245,32 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
...
+		machinePoolScope.RosaMachinePool.Status.Ready = false
+		v1beta1conditions.MarkFalse(machinePoolScope.RosaMachinePool,
+			expinfrav1.RosaMachinePoolReadyCondition,
+			expinfrav1.RosaMachinePoolReconciliationFailedReason,
+			clusterv1beta1.ConditionSeverityError,
+			"%s", err)
+		machinePoolScope.Info("Invalid ROSAMachinePool spec", "error", err)
+
+		// Don't requeue because input is invalid and manual intervention is needed.
 		return ctrl.Result{}, nil
```
serngawy: return the error so see it in the logs 
```suggestion
		return ctrl.Result{}, err
```

pkg/rosa/versions.go
```diff
@@ -103,18 +103,66 @@ func ScheduleNodePoolUpgrade(client OCMClient, clusterID string, nodePool *cmv1.
...
+
+	// Minor is a uint64, so subtracting the allowed deviation from a minor
+	// version below it wraps around instead of clamping -- max() cannot undo
+	// that, because the value is already enormous by the time it is applied.
+	// Clamp the subtrahend instead, which keeps a 0.x control plane at 0.0.
+	//
+	// When the control plane minor is below the allowed deviation the skew
+	// window crosses a major boundary; admit the whole previous major series.
+	var minVersion semver.Version
+	if coreVersion.Minor < minorVersionsAllowedDeviation && coreVersion.Major > 0 {
```
serngawy, line 133: not sure if this logic is correct, for max v5.0.0 this will return min v4.0.0, right ?

pkg/rosa/versions.go
tinaafitz: Thanks, @andclt. Good catch. Modified.

serngawy (commented):
The tricky case is when creating MP with v 4.18.0 and default (CP) version is 5.0.0; In the current code logic this case will return min version as 4.0.0 while we should raise error as min accepted version should be 4.20/4.21
I think we should ignore validating the version in this case and submit the version as it is then raise the coming back error from OCM.
We validate the version based on the MP.status.availableUpgrade if exist  otherwise submit the version as the user set it and raise OCM error if any
@andclt @tinaafitz , what do you think ?

pkg/rosa/versions.go
tinaafitz: Hi @andclt,

Yes. I struggled with this one. I added a comment.

exp/controllers/rosamachinepool_controller.go
tinaafitz: Thanks, @andclt. Modified.

exp/controllers/rosamachinepool_controller.go
tinaafitz: Shortened.

pkg/rosa/versions.go
tinaafitz: Hi @serngawy, yes that's correct — for a 5.0 control plane the cross-major branch sets the initial floor to 4.0.0, but two lines below it gets clamped to MinSupportedVersion (4.14.0), so the effective minimum returned is 4.14.0. The intent is to admit the whole previous major series that ROSA actually supports, without hardcoding a specific minor (4.22 today, but that will change). OCM is the authoritative gate and will reject anything genuinely unsupported.

exp/controllers/rosamachinepool_controller.go
tinaafitz: Hi @serngawy, the nil return here is intentional — this path is for an invalid spec that won't self-heal without manual intervention. Returning err would cause controller-runtime to requeue indefinitely, which is noisy and wasteful for input that won't change on its own. The error is already surfaced via the RosaMachinePoolReadyCondition condition and the log line. Happy to change the log level to Error if you'd prefer more visibility without the requeue side effect.

pkg/rosa/versions.go
serngawy: agree, we may just submit the version and surface back the OCM API error is the version upgrade path is not supported

pkg/rosa/versions.go
```diff
@@ -103,18 +103,66 @@ func ScheduleNodePoolUpgrade(client OCMClient, clusterID string, nodePool *cmv1.
...
+		minVersion = semver.Version{Major: coreVersion.Major - 1, Minor: 0, Patch: 0}
+	} else {
+		minVersion = semver.Version{
+			Major: coreVersion.Major,
+			Minor: coreVersion.Minor - min(coreVersion.Minor, minorVersionsAllowedDeviation),
```
serngawy, line 138: Also this logic is not valid ex; coreVersion v5.3.0 the minVersion will be 5.2.0 as `minorVersionsAllowedDeviation = 2` while minVersion should be v5.1.0

tinaafitz:
Hi @serngawy, 
You're right that the 4.14.0 floor is imprecise for the cross-major case — a 4.18.0 pool against a 5.0 CP would pass local validation even though OCM would reject it.

That said, the end-user experience is the same either way: OCM rejects the unsupported version and the error surfaces as a RosaMachinePoolReadyCondition failure with the OCM message, so users aren't silently blocked.

  Skipping local validation entirely for cross-major would also introduce new risks:
  - The upper-bound check would be lost too — a 6.0.0 pool against a 5.0 CP would pass local validation and only fail at OCM
  - Any OCM API outage or connectivity issue would leave users with no local feedback at all during the cross-major window
  - Preserving the upper bound separately adds complexity and new test surface at RC time
  - A new IsCrossMajorBoundary helper would need its own tests and could diverge from the logic in MachinePoolSupportedVersionsRange over time
  - It makes OCM load-bearing for correctness in the cross-major case rather than a safety net
  
  For those reasons I think keeping the permissive local floor is actually the safer option here — it adds zero new code, the fail-open intent is documented in the comment, and the proper fix using status.availableUpgrades or querying OCM for the available version list is already called out in the PR description as a follow-up.

pkg/rosa/versions.go
```diff
@@ -103,18 +103,41 @@ func ScheduleNodePoolUpgrade(client OCMClient, clusterID string, nodePool *cmv1.
 // machinepools can be created with a minimal of two minor versions from the control plane.
 const minorVersionsAllowedDeviation = 2
 
-// MachinePoolSupportedVersionsRange returns the supported range of versions
-// for a machine pool based on the control plane version.
+// CoreVersion strips any prerelease and build qualifiers, leaving major.minor.patch.
+// The machine pool skew policy is expressed purely in terms of minor versions, and
+// semver sorts a prerelease below its own release ("5.0.0-rc.0" < "5.0.0"), so the
+// qualifiers have to be dropped before any range comparison -- otherwise a
```
andclt: nit: now that the upper bound intentionally preserves prerelease ordering, "before any range "comparison" seems slightly misleading. Maybe "before the lower-bound/skew comparison" instead?

exp/controllers/rosamachinepool_controller.go
```diff
@@ -306,6 +311,15 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 			rosaMachinePool.Status.Ready = true
 
 			if err := r.reconcileMachinePoolVersion(machinePoolScope, ocmClient, nodePool); err != nil {
```
andclt, line 311: would it be worth adding a regression case for this path as well?

andclt (commented):
Overall looks good to me, I've just added 2 non blocking comments

exp/controllers/rosamachinepool_controller.go
```diff
@@ -245,15 +245,20 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
...
+		// intentionally as writing it panics CAPI's MachinePool controller.
+		machinePoolScope.RosaMachinePool.Status.Ready = false
+		v1beta1conditions.MarkFalse(machinePoolScope.RosaMachinePool,
+			expinfrav1.RosaMachinePoolReadyCondition,
+			expinfrav1.RosaMachinePoolReconciliationFailedReason,
+			clusterv1beta1.ConditionSeverityError,
+			"%s", err)
+		machinePoolScope.Info("Invalid ROSAMachinePool spec", "error", err)
+
+		// Don't requeue because input is invalid and manual intervention is needed.
```
serngawy: please return the error , so we can see it in the CAPA deployment logs

exp/controllers/rosamachinepool_controller.go
serngawy: when upgrade schedule fail , it doesn't mean nodePool is not ready . The NodePool could be still ready for workloads

serngawy: if you want surface the upgrade schedule failure then we should use the `RosaMachinePoolUpgradingCondition` , this is another PR not in this one

exp/controllers/rosamachinepool_controller.go
```diff
@@ -245,15 +245,19 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
...
-		return ctrl.Result{}, nil
+	if err := validateMachinePoolSpec(machinePoolScope); err != nil {
+		// Surface the error to the condition; Status.FailureMessage is avoided
+		// intentionally as writing it panics CAPI's MachinePool controller.
+		machinePoolScope.RosaMachinePool.Status.Ready = false
+		v1beta1conditions.MarkFalse(machinePoolScope.RosaMachinePool,
+			expinfrav1.RosaMachinePoolReadyCondition,
+			expinfrav1.RosaMachinePoolReconciliationFailedReason,
+			clusterv1beta1.ConditionSeverityError,
+			"%s", err)
```
serngawy, line 255:
```suggestion
		v1beta1conditions.MarkFalse(machinePoolScope.RosaMachinePool,
			expinfrav1.RosaMachinePoolUpgradingCondition,
			expinfrav1.RosaMachinePoolReconciliationFailedReason,
			clusterv1beta1.ConditionSeverityError,
			"%s", err)
```

exp/controllers/rosamachinepool_controller.go
```diff
@@ -245,15 +245,19 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
...
-	if err != nil {
-		return ctrl.Result{}, fmt.Errorf("failed to validate ROSAMachinePool.spec: %w", err)
-	}
-	if failureMessage != nil {
-		// dont' requeue because input is invalid and manual intervention is needed.
-		return ctrl.Result{}, nil
+	if err := validateMachinePoolSpec(machinePoolScope); err != nil {
+		// Surface the error to the condition; Status.FailureMessage is avoided
+		// intentionally as writing it panics CAPI's MachinePool controller.
+		machinePoolScope.RosaMachinePool.Status.Ready = false
```
serngawy: I don't think we should set the` ready = false ` keep it as it is , and below use the `RosaMachinePoolUpgradingCondition` instead.

tinaafitz: Done

exp/controllers/rosamachinepool_controller.go
tinaafitz: Done

serngawy (approved):

tinaafitz:
Hi @richardcase, It's been approved. :-)

PR #6217, opened 2026-09-07
🐛 fix(rosa): update predicates to nil ManagedFields before comparison

andclt:
**What type of PR is this?**

/kind bug

**What this PR does / why we need it**:

Five ROSA controllers (`ROSAControlPlane`, `ROSAMachinePool`, `ROSANetwork`, `ROSAOCMRoleConfig`, `ROSARoleConfig`) have update predicates that zero `Status` and `ResourceVersion` before comparing old and new objects with `cmp.Equal`, intending to skip reconciliation on status-only updates. However, a status-subresource write also refreshes the `lastTransitionTime` in the writer's `managedFields` entry, so the comparison always finds a difference and the predicate never filters anything out.

This PR nils `ManagedFields` on both copies before comparing, matching the fix already applied to `AWSMachine` and `AWSMachinePool` in #6198.

**Which issue(s) this PR fixes** _(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)_:
Fixes #

**Special notes for your reviewer**:

The same two-line fix is applied identically in all five files. See `controllers/awsmachine_controller.go:301-304` for the reference pattern from #6198.

**AI Usage**:

Claude Code (Opus) was used to locate the affected files, apply the fix across all five controllers, and verify with lint and tests.

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [x] includes AI generated content
- [x] includes emoji in title
- [ ] adds unit tests
- [ ] adds or updates e2e tests

**Release note**:

```release-note
fix(rosa): update predicates to nil ManagedFields before comparison, preventing unnecessary reconciliations on status-only updates
```

PR #6243, opened 2026-09-15
🐛 Fix ROSA controllers stalling after resource creation due to status-update predicate

andclt:
**What type of PR is this?**

/kind bug
/kind regression

ROSA controllers stall permanently after resource creation because they rely on status-patch watch events to re-trigger reconciliation, but the `UpdateFunc` predicate (added in #6128, tightened in #6217) now drops those events.

Four controllers are affected:
- **ROSARoleConfig**: `reconcileAccountRoles` creates IAM roles via `CreateHCPRoles` but doesn't re-lookup ARNs after creation — `reconcileOperatorRoles` finds `InstallerRoleARN` empty and returns nil with no requeue
- **ROSAControlPlane**: returns `ctrl.Result{}, nil` after `CreateCluster`
- **ROSAMachinePool**: returns `ctrl.Result{}, nil` after `CreateNodePool`
- **ROSANetwork**: returns `ctrl.Result{}, nil` after `CreateStackWithParamsTags`

The fix follows the established CAPA pattern (used by `AWSMachine` and `AWSMachinePool`): never rely on status-patch watch events for re-reconciliation. Either complete work in one pass or use explicit `RequeueAfter`.

- **ROSARoleConfig**: re-lookup ARNs after `CreateHCPRoles` and set status in one pass (matching `reconcileOIDC` in the same file)
- **ROSAControlPlane/ROSAMachinePool/ROSANetwork**: return `RequeueAfter: 60s` after resource creation

Fixes #

PRs #6128 and #6217 were individually correct, #6128 added the predicate to prevent `PatchObject()` from bypassing error-retry backoff, and #6217 fixed unnecessary reconciliation from managedFields timestamp noise. But together they removed the accidental safety valve (managedFields timestamp differences) that kept status-patch events flowing through the predicate.

The predicate itself is fine and serves a legitimate purpose. The issue was that these four controllers depended on status-patch events as their only re-trigger mechanism instead of using explicit `RequeueAfter`. `ROSAOCMRoleConfig` is not affected because it completes all work in one pass.

**AI Usage**:
Used to confirm the bug and the proposed resolution.

**Testing**:

Run full e2e suite test locally: complete provision → upgrade CP → upgrade MP → add machinepool → delete machinepool → delete cluster flow

**Checklist**:

- [x] squashed commits
- [ ] includes documentation
- [X] includes emoji in title
- [X] adds unit tests

**Release note**:

```release-note
fix(ROSA): ROSA controllers stalling permanently after resource creation in RequeueAfter loop
```


exp/controllers/rosaroleconfig_controller.go
tinaafitz, line 210: Hi @andclt,

Should we add a requeue here at line 217 too?

       return ctrl.Result{RequeueAfter: time.Second * 60}, nil

andclt: Hey @tinaafitz,
I was thinking about that too but the `reconcileAccountRoles` fix re-lookups ARNs right after `CreateHCPRoles`, so in practice we shouldn't reach this line with not-ready status. IAM control plane operations (`CreateRole`/`GetRole`) have read-after-write consistency, so the re-lookup is guaranteed to find the roles so it would be unnecessary. That said, adding a `RequeueAfter` here is harmless, it would just be an additional reconciliation trigger, do you have a strong preference here?

tinaafitz: Hi @andclt,

Thanks for the explanation. :-)  We don't need to change anything.

exp/controllers/rosaroleconfig_controller.go
```diff
@@ -375,10 +375,20 @@ func (r *ROSARoleConfigReconciler) reconcileAccountRoles(scope *scope.RosaRoleCo
...
-	return accountroles.CreateHCPRoles(rt, prefix, true, scope.RosaRoleConfig.Spec.AccountRoleConfig.PermissionsBoundaryARN,
+	if err := accountroles.CreateHCPRoles(rt, prefix, true, scope.RosaRoleConfig.Spec.AccountRoleConfig.PermissionsBoundaryARN,
 		rosa.GetOCMClientEnv(rt.OCMClient), policies, scope.RosaRoleConfig.Spec.AccountRoleConfig.Version, scope.RosaRoleConfig.Spec.AccountRoleConfig.Path,
 		scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.IsSharedVPC(), scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.RouteRoleARN,
-		scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.VPCEndpointRoleARN)
+		scope.RosaRoleConfig.Spec.AccountRoleConfig.SharedVPCConfig.VPCEndpointRoleARN); err != nil {
+		return err
+	}
+
+	accountRolesRef, err = r.lookupAccountRolesRef(rt, prefix)
```
serngawy, line 395: not sure, if every time IAM account Role created we get it right away, we may end up [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosaroleconfig_controller.go#L259)  if installer role not available

andclt: That's correct, fixed it also for `reconcileOperatorRoles`

serngawy: I mean keep this part as it is no need to change it and requeue [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosaroleconfig_controller.go#L210)  after eg; 30s . Because the func  `lookupRoleARNs` [here](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosaroleconfig_controller.go#L399) will not raise error if the role not available yet. Then we will reconile operatorRole and hitting [this](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/controllers/rosaroleconfig_controller.go#L258)  installerRole not exist . So we endup not requeuing again.  Is that make sense ?

controlplane/rosa/controllers/rosacontrolplane_controller.go
```diff
@@ -210,8 +210,10 @@ func (r *ROSAControlPlaneReconciler) Reconcile(ctx context.Context, req ctrl.Req
 	}
 
 	log = log.WithValues("cluster", klog.KObj(cluster))
-	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, rosaControlPlane); err != nil || isPaused || conditionChanged {
+	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, rosaControlPlane); err != nil || isPaused {
 		return ctrl.Result{}, err
+	} else if conditionChanged {
```
serngawy, line 218: Would you elaborate more, what conditionChanged that will require requeue in 10s ?

exp/controllers/rosamachinepool_controller.go
```diff
@@ -144,8 +144,10 @@ func (r *ROSAMachinePoolReconciler) Reconcile(ctx context.Context, req ctrl.Requ
 		return ctrl.Result{}, nil
 	}
 
-	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, rosaMachinePool); err != nil || isPaused || conditionChanged {
+	if isPaused, conditionChanged, err := paused.EnsurePausedCondition(ctx, r.Client, cluster, rosaMachinePool); err != nil || isPaused {
 		return ctrl.Result{}, err
+	} else if conditionChanged {
```
serngawy, line 149: same here

exp/controllers/rosamachinepool_controller.go
```diff
@@ -340,7 +342,7 @@ func (r *ROSAMachinePoolReconciler) reconcileNormal(ctx context.Context,
 	}
 
 	machinePoolScope.RosaMachinePool.Status.ID = nodePool.ID()
-	return ctrl.Result{}, nil
+	return ctrl.Result{RequeueAfter: time.Second * 60}, nil
```
serngawy: I believe better to have this `RequeueAfter: time.Second * 60` as variable and use it in all places. eg `DefaultRequeueInterval := time.Second * 60`

controlplane/rosa/controllers/rosacontrolplane_controller.go
andclt: `EnsurePausedCondition` patches the `Paused` condition onto the resource's status (even when the resource is not paused, the condition is set with `Status: False`). On the first reconcile, the condition doesn't exist yet, so it gets created and patched, returining `conditionChanged=true`.
Before this fix, the code fell through to `return ctrl.Result{}, nil` when `conditionChanged` was `true` (since `isPaused` is `false` and err is `nil`). The subsequent status patch event from setting the condition was filtered by the update predicate (which drops status-only changes), so the controller stalled until the next informer resync (~10 minutes).

The `RequeueAfter: 10s` ensures the controller re-reconciles promptly after the condition is first set, bypassing the predicate (`RequeueAfter` is queue-based, not event-based). 10 seconds is short because this is a one-time initialization step, subsequent reconciles won't trigger `conditionChanged` since the condition already exists and `HasSameState` returns `true`.

exp/controllers/rosamachinepool_controller.go
andclt: Same explanation as above: https://github.com/kubernetes-sigs/cluster-api-provider-aws/pull/6243#discussion_r4026857106

exp/controllers/rosaroleconfig_controller.go
andclt: I looked into the IAM consistency model a bit more. The docs are somewhat confusing here: the generic [IAM troubleshooting docs](https://docs.aws.amazon.com/IAM/latest/UserGuide/troubleshoot.html#troubleshoot_general_eventual-consistency) mention eventual consistency, but AWS's [more specific control-plane documentation ](https://aws.amazon.com/blogs/security/how-to-monitor-and-query-iam-resources-at-scale-part-1/)states the IAM control-plane writes are read-after-write consistent.

Since `CreateHCPRoles`/`EnsureRole` followed by `GetRoleByName` stays within the IAM control plane (`CreateRole` -> `GetRole`), I'd expect the role to be immediately readable after creation.

That said, I agree we shouldn't risk getting stuck if `GetRoleByName` unexpectedly returns `NoSuchEntity`. Instead of unconditionally requeueing after creation, how about requeueing only if the post-create lookup is incomplete? That would keep the normal path immediate while still safely handling any propagation edge case.

Something along the lines of:

```
if !r.accountRolesReady(scope.RosaRoleConfig.Status.AccountRolesRef) {
    return ctrl.Result{RequeueAfter: x}, nil
}
```

and similarly for the operator roles after their post-create lookup.

exp/controllers/rosamachinepool_controller.go
andclt: Makes sense, I've extracted these into named constants:
```
const (
    conditionRequeueInterval = 10 * time.Second
    defaultRequeueInterval   = 60 * time.Second
)
```

For the `exp/controllers` package (`rosamachinepool`, `rosanetwork`, `rosaroleconfig`), they live in a new `rosa_constants.go` file to avoid implicit cross-file coupling. The `rosacontrolplane` controller is in a separate package so it has its own copy in the existing const block.

exp/controllers/rosaroleconfig_controller.go
serngawy: okay, that is fine if we will check the accountRole and requeue if it is not ready similarly the operatorRole.

andclt: Looking at this again I think your suggestion is simpler and easier since `rosaRolesConfigRead()` already checks the account roles and operator roles readiness, I'll go with that!

exp/controllers/rosa_constants.go
```diff
@@ -0,0 +1,8 @@
+package controllers
```
serngawy, line 17: missing license header

serngawy (commented):
looks good, just the license header for the new file

exp/controllers/rosa_constants.go
andclt: Good catch, added thanks!

serngawy (approved):
/lgtm

serngawy:
@damdo would you  review/merge this

damdo (commented):
/approve

PR #6252, opened 2026-09-18
🐛 Draft Stop unnecessary OCM API calls on every reconcile for every node pool where subnet wasn't explicitly set.

tinaafitz:
**What type of PR is this?**
/kind bug

Ignore OCM-auto-assigned subnet in ROSAMachinePool spec diff.  

 When a ROSAMachinePool is created without a Subnet field, OCM auto-assigns one and always returns it in subsequent responses. This caused computeSpecDiff to detect a diff on every reconcile (desiredSpec.Subnet="" vs currentSpec.Subnet="subnet-xxx"), triggering a no-op UpdateNodePool OCM API call every loop.

  Fix: when desiredSpec.Subnet is empty, append "Subnet" to the ignored fields list so OCM-defaulted subnets are treated the same as other immutable/OCM-managed fields (Version, AdditionalTags, AdditionalSecurityGroups).

  Add TestComputeSpecDiff_SubnetOmitted to cover the missing case — the existing NoPhantomDiff tests never set Subnet on the OCM node pool builder, so this scenario was untested.

Fixes #

**AI Usage**:

Claude Code (Claude Opus 5) was used for implementation and test writing. I reviewed every change and verified it independently: the underflow reproduces in a standalone Go program with no cluster, the resulting failure mode was confirmed on a live 5.0 management cluster running a released image, the CAPI panic described above reproduces against an unmodified checkout of cluster-api main, the new reconcile test fails if the `Status.FailureMessage` write is restored, and `make generate` (no drift), `make lint` (0 issues) and `make test` all pass locally.
**Checklist**:

- [ ] squashed commits
- [ ] includes documentation
- [X] includes emoji in title
- [X] adds unit tests

**Release note**:

```release-note
```


serngawy (commented):
Oh wasn't aware that MP.spec.subnet is optional. In that case we should then show the subnet-id in the MP.status.subnet-id . If the user didn't specify the subnet and rely on OCM to assign it , we should show this info . What do you think ? 
Also please add some desc to the subnet-id [field](https://github.com/kubernetes-sigs/cluster-api-provider-aws/blob/main/exp/api/v1beta2/rosamachinepool_types.go#L54)

tinaafitz:
Hi @serngawy,

Good suggestion, though in practice, when using ROSARoleConfig, the subnet gets auto-generated as part of the network setup, so the user never needs to know which subnet OCM picked — it's an implementation detail. 
The status field would mainly help in the manual case where someone intentionally omits a subnet and wants to inspect what OCM chose.

serngawy:
> so the user never needs to know which subnet OCM picked

That is not true, It is important info for endUser to know which subnet is the machinePool attached to it so we can understand how many nodePools available in the subnet and availability zone.

